PR #326, opened 2014-09-26
feat(build): copy src sass files to dist for bower-material

ocombe:
Fixes https://github.com/angular/bower-material/issues/4


breandr:
I assumed this meant that scss files will would be available in bower-material, but I'm not finding it in my angular-material directory (version 0.7.0-rc3). Have I misunderstood the purpose of this PR? How do I go about getting .scss files if I use bower? :)

PR #479, opened 2014-10-25
fix (slider) - fixing slider default value in ng-repeat

sonnylazuardi:
md-slider does not work within ng-repeat.
when i have used md-slider without ng-repeat the slider sets on its proper position and aria-valuenow is also set with its proper value, but when i have used md-slider within ng-repeat the aria-valuenow="NaN" and slider is not set with its proper position.

Plunker demo of problem - http://plnkr.co/edit/PslB2p?p=preview

With this pull request the default value of md-slider in ng-repeat will be correct.

### DIFF
--- a/src/components/slider/slider.js
+++ b/src/components/slider/slider.js
@@ -258,6 +258,9 @@ function SliderController(scope, element, attr, $$rAF, $window, $mdEffects, $mdA
       ngModelCtrl.$setViewValue( minMaxValidator(stepValidator(value)) );
     }
     function ngModelRender() {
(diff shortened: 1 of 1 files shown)


johnrivelt:
It still seems not working correctly in my case.
Here is my case from plunker.

https://embed.plnkr.co/FCizUTQnOTzsklhxSIry/

click SSD 1 -> SSD 10 direction set min and value correctly.
But SSD 10 -> SSD 1  backward case, md-slider does not update correctly.
Even min-value is same with valuenow but slider cursor indicate middle of somewhere.

PR #565, opened 2014-11-07
Multiline input 

erbear:
I have added multiline option to mdInputDirective. Expanding option works very smoth but shrinking has no animation. I can write tests if you like it


ThomasBurleson:
@ajoslin - I think we should review this PR for an upcoming work on multi-line textareas with `<md-input-container>`.

Issue #586, opened 2014-11-09
src/components/textField/textField.spec.js out of date?


mlibby:
As the spec file stands in `master`, the tests all pass. 

But line 240 has a template like this
```
    md_input_group: '<div class="md-input-group" tabindex="-1">' +
                ' <label>{{label}}</label>' +
                ' <md-input id="{{id}}" type="{{type}}" ng-model="{{model}}"></md-input>' +
                '</div>'
```
This does not match the template in `textField.js` (line 74)
```
    '<md-input-group ng-disabled="isDisabled" tabindex="-1">' +
    ' <label for="{{fid}}" >{{label}}</label>' +
    ' <md-input id="{{fid}}" ng-model="value" type="{{inputType}}"></md-input>' +
    '</md-input-group>'
```
Using the second template causes two test to fail.

marcysutton:
Closing as this now references deprecated code.

PR #587, opened 2014-11-09
fix(mdTextField): set tabindex on input if set on mdTextFloat

mlibby:
This allows tabindex to be set on the initial mdTextFloat element and have it properly applied to the input element.

Closes: angular#583.

Because the template is changed, I can write tests once https://github.com/angular/material/issues/586 is resolved (or I understand why the current spec's template is correct).


ThomasBurleson:
mdTextFloat is now a deprecated component; consider `<md-input-container>` usages.

PR #615, opened 2014-11-11
enhance(ripple): rewrites ripple effect to match the spec

robertmesserle:


aeroxy:
Does this means the current version already fix the scrolling issue? I still observe this problem. @robertmesserle @ajoslin

robertmesserle:
It should

aeroxy:
@robertmesserle Well, I found the ripple's position (the center isn't where the mouse click is) incorrect after the page being scrolled.

Here is how to reproduce the situation:

1. Create a list of small cards, using `layout="row" layout-wrap` in its wrapper, and apply `md-ink-ripple` to the cards.

2. Create a `md-tabs` element with some `md-tab` inside. Make `.md-header`'s position `fixed`.

I am looking into the 2nd situation's codes right now, it looks like `.md-ripple.md-ripple-placed.md-ripple-scaled.md-ripple-visible`'s top position is incorrectly calculated when the position is set to `fixed` but I am not sure what the reason is for the 1st situation.

aeroxy:
@robertmesserle I see. The Y location is depending the page, it should be depending on the window when position is set to fixed. Still looking to solve the first situation.

aeroxy:
For those who are interested, here is a quick and dirty fix (jquery required) for fixed ripple elements (this one is for tab):

    $(window).mousemove(function(e){
      winY = e.pageY - $(window).scrollTop(),
      winH = $(window).height();
      if ($('.md-header.fixed').length == 0) {
        $('#ripplefix').html('');
      } else {
        $('#ripplefix').html('md-tab .md-ripple{top:' + winY + 'px !important}');
      }
    });

aeroxy:
@robertmesserle The first situation is due to the miscalculation of margin-top. I don't have a fix yet.

PR #627, opened 2014-11-12
fix(mdDialog): fix scrollbar going behind dialog header

Awk34:
This simply adds z-index:1 to md-content of the dialog, so that the scrollbar stays on top of the header.


Awk34:
Looks good in 0.10.0-RC2 :+1:

PR #800, opened 2014-12-01
fix(slider): slider repositions itself when min or max is updated

### LINKED ISSUE #799: md-slider doesn't update its position when max or min is updated
When the `max` value of an `md-slider` is updated, the horizontal position of the slider isn't updated:

``` html
<!-- index.html -->
<md-slider flex min="0" max="{{max}}" ng-model="value"></md-slider>
<md-button ng-click="doubleMax()">Double max!</md-button>
```

``` javascript
// index.js
$scope.doubleMax = function () {
  $scope.max *= 2
}
```

See [this Plunker](http://plnkr.co/edit/Cd95q4Un0UmNgA2xxS1t?p=preview). If I double the maximum value for the slider, it doesn't move to reflect the modification. However, when I try to move it (using the keyboard for instance), then it moves in the right position.
untitled

Tested on Chrome, Firefox and Safari.

PaulMougel:
Closes #799

### DIFF
--- a/src/components/slider/slider.js
+++ b/src/components/slider/slider.js
@@ -161,10 +161,12 @@ function SliderController($scope, $element, $attrs, $$rAF, $window, $mdAria, $md
     function updateMin(value) {
       min = parseFloat(value);
       $element.attr('aria-valuemin', value);
+      updateAll();
     }
     function updateMax(value) {
       max = parseFloat(value);
       $element.attr('aria-valuemax', value);
+      updateAll();
     }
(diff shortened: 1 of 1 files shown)


johnrivelt:
When min value changed, there is still some issue.
[https://embed.plnkr.co/tdyYHl/](url)
In above case, when click ssd 1 to 5, top-to-bottom direction works as I expect.
But, bottom-to-top is not updating correctly.

PR #824, opened 2014-12-02
layout-fill firefox fix

rem4ik4ever:
Fixed layout-fill for Firefox so it only fills

### DIFF
--- a/src/core/style/layout.scss
+++ b/src/core/style/layout.scss
@@ -43,6 +43,13 @@
   min-height: 100%;
   width: 100%;
 }
+@-moz-document url-prefix() { 
+  [layout-fill] {
+    margin: 0;
+    width: 100%;
+    min-height: auto;
+  }
+}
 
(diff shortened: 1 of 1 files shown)


rschmukler:
Merged, thank you!

src/core/style/layout.scss
```diff
@@ -43,6 +43,13 @@
   min-height: 100%;
   width: 100%;
 }
+@-moz-document url-prefix() { 
+  [layout-fill] {
+    margin: 0;
+    width: 100%;
+    min-height: auto;
```
jeppebemad: I'm not sure this does as intended. In FF, fill-layout no longer makes a container full height, and overflow breaks in e.g. md-content. Works in Chrome.

dnadoba:
min-height: auto
dosen't work as excepted
i have tried 
min-height: 0
and it worked like in chrome

Issue #814, opened 2014-12-02
md-text-float and "name" attribute


OlivierD911:
The md-text-float directive generate a template with md-input-group,label and input. But if i put a "name" attribute in the md-text-float (for POST method when i Submit the form), i don't found it in the input element generated. 
is this a bug ? or i'm wrong in the use of md-text-float ?

ajoslin:
Fixed via 60fcd6f4d8b895162b37a940c3f33164d8044382

PR #776, opened 2014-11-26
docs(textField): changes eMail to Email

jesselpalmer:

PR #829, opened 2014-12-02
fix(mdTextFloat): Adds support for name attribute

marcysutton:
With this change, the `name` attribute is now supported on mdTextFloat.

Closes issue #814 and PR https://github.com/angular/material/pull/776


ajoslin:
Fixed via 60fcd6f4d8b895162b37a940c3f33164d8044382

PR #835, opened 2014-12-03
Accept flex="100"

myagoo:
Now able to do things like `<div flex="50" flex-sm="100"/>`


ajoslin:
This is already in master, I think I added it without seeing this PR. Thanks though!

Issue #583, opened 2014-11-09
tabindex and mdTextFloat


mlibby:
The `tabindex` attribute is not being set correctly on text inputs if it is set on a `mdTextFloat`. Some of what I'm seeing is partially due to including ngAria, but I believe the problem is in `tabindex` propagation inside `mdTextFloat`.

In my sample code I am attempting to set the `tabindex` on the `mdButton` so that it is last in the tab order. That way the keyboard user can just tab into the form at the first field, enter values, tab thru the form, and finally tab to the `Save` button and save.

However, it does not work as expected. I have set up this sample code:

```html
<style>
  input[type=text]:focus{ background-color: red; }
  md-input-group:focus{ background-color: blue; }
</style>

<form novalidate>
  <md-toolbar class='md-toolbar-tools'>
    <h1>New Account</h1>
    <span flex></span>
    <md-button tabindex='2' class='md-button' type='submit' ng-click='ctrl.submit()'>Save</md-button>
  </md-toolbar>
  <md-content layout-padding>
    <md-text-float tabindex='1' label='Text One' ng-model='$rootScope.foo'></md-text-float>
    <md-text-float tabindex='1' label='Text Two' ng-model='$rootScope.bar'></md-text-float>
  </md-content>
</form>
```
which behaves as follows as shown in the animated GIF (one frame per Tab keypress):

![tabindex](https://cloud.githubusercontent.com/assets/36049/4968063/ea43b7b4-6830-11e4-81e8-67239063044e.gif)
The markup that was generated was:

```html
<main class="md-padding clearfix ng-scope" layout="vertical" role="main" ng-view="">
<style class="ng-scope">
  input[type=text]:focus{ background-color: red; }
  md-input-group:focus{ background-color: blue; }
</style>

<form class="ng-pristine ng-valid ng-scope ng-submitted" novalidate="">
  <md-toolbar class="md-toolbar-tools md-default-theme">
    <h1>New Account</h1>
    <span flex=""></span>
    <button tabindex="2" type="submit" ng-click="ctrl.submit()" class="md-button md-default-theme" ng-transclude=""><span class="ng-scope">Save</span></button>
  </md-toolbar>
  <md-content class="md-default-theme" layout-padding="">
    <md-input-group aria-invalid="false" class="ng-isolate-scope md-default-theme ng-valid ng-touched" label="Text One" ng-model="$rootScope.foo" ng-disabled="isDisabled" tabindex="1 -1"> 
[...]

Issue #654, opened 2014-11-14
md-text-float does not work with ng-disabled, version 0.5.1


RobScottDEED:
Markup:
```
<md-text-float fid="fedTaxId" name="fedTaxId" label="Federal Tax ID" 
  ng-model="revenue.service.data.revenue.fedTaxId"
  ng-disabled="registrant.service.isRegistrantInputDisabled()" type="text" required>
</md-text-float>
```
Generated:
```
<md-input-group ng-disabled="registrant.service.isRegistrantInputDisabled() isDisabled" 
  tabindex="-1" fid="fedTaxId" name="fedTaxId" label="Federal Tax ID" 
  ng-model="revenue.service.data.revenue.fedTaxId" 
  class="ng-isolate-scope md-default-theme ng-invalid ng-invalid-required" 
  aria-required="true" aria-invalid="true"> 
  <label for="fedTaxId" class="ng-binding">Federal Tax ID</label> 
  <input id="fedTaxId" ng-model="value" type="text" tabindex="0" aria-disabled="false" 
    aria-invalid="false" class="ng-valid">
</md-input-group>
```
Error:
```
Error: [$parse:syntax] Syntax Error: Token 'isDisabled' is an unexpected token at column 48 of the expression [registrant.service.isRegistrantInputDisabled() isDisabled] starting at [isDisabled].
http://errors.angularjs.org/1.3.1/$parse/syntax?p0=isDisabled&p1=is%20an%20unexpected%20token&p2=48&p3=registrant.service.isRegistrantInputDisabled()NaNsDisabled&p4=isDisabled
    at http://localhost:8080/bower_components/angular/angular.js:80:12
    at Parser.throwError (http://localhost:8080/bower_components/angular/angular.js:11808:11)
    at Parser.parse (http://localhost:8080/bower_components/angular/angular.js:11759:12)
    at $parse (http://localhost:8080/bower_components/angular/angular.js:12398:39)
    at Scope.$watch (http://localhost:8080/bower_components/angular/angular.js:13544:19)
    at http://localhost:8080/bower_components/angular-aria/angular-aria.js:100:15
    at invokeLinkFn (http://localhost:8080/bower_components/angular/angular.js:8081:9)
    at nodeLinkFn (http://localhost:8080/bower_components/angular/angular.js:7593:11)
    at compositeLinkFn (http://localhost:8080/bower_components/angular/angular.js:6959:13)
[...]

Issue #562, opened 2014-11-07
disable/enable mt-text-float at runtime


celestinoramalho:
Although it is possible to do it with a wrapper directive to add or remove the "disabled" attribute, it would be very helpful to have a bindable property to do that, like ng-disabled.

PR #845, opened 2014-12-03
feat(mdTextFloat): Adds support for custom tabindex

marcysutton:
With this change, developers can provide a custom tabindex on `mdTextFloat` and have it successfully copied to the child input. Here are the expectations for supporting this feature:

 * TabIndex supplied by developer should be copied to `md-input` or set to `0`
 * TabIndex on `md-input-group` should be set to `-1` for scripting focus

Known issue: `disabled` is not actually disabling textfields. Related: https://github.com/angular/material/issues/654, https://github.com/angular/material/issues/562

Closes Issues #583, #586 and PR https://github.com/angular/material/pull/587


ajoslin:
Fixed via 60fcd6f4d8b895162b37a940c3f33164d8044382

PR #848, opened 2014-12-04
md-input directive

ajoslin:
Old problem: We didn't know what attributes to pass down to the input directive from the input container.

Solution: We have a directive with priority 100 and `terminal: true`. `terminal` means that no directive with lower priority will run on this element - basically, directive compilation will stop at `md-input`.

With `priority: 100`, `md-input` still runs after `ng-if`, `ng-repeat`, etc.

When `md-input` runs, it takes all attributes on `<md-input>` container and copies them down to a child `<input>` element - EXCEPT attributes that are directives with higher priority (like ng-if).

**PROBLEMS**

- ng-show is actually 0 priority, so it will be passed down to the input element. That means the input element will be hidden, but not the parent.

### DIFF
--- a/src/components/input/_input.scss
+++ b/src/components/input/_input.scss
@@ -1,0 +1,101 @@
+
+$tff-font-size: 0.75em;
+$tff-line-height:26px;
+$tff-transition: all 0.15s $swift-ease-in-out-timing-function;
+// - `label` element (aka hint)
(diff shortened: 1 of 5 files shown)


ajoslin:
Went with a different approach. 60fcd6f4d8b895162b37a940c3f33164d8044382

PR #858, opened 2014-12-05
fix(dialog): fix dialog alignment in IE

### LINKED ISSUE #790: Dialog on IE11
The dialog shown on the demo page is aligned right in IE11 while centered in Chrome.

jcruz2us:
Closes #790

I ran into this issue in IE10 and IE11 and I found a simple solution; Remove margin auto from the md-dialog element. After removing the margin, the dialog is properly aligned in Chrome, Firefox and IE10/11


marcysutton:
This is no longer applicable.

PR #881, opened 2014-12-09
fix tab error where contentContainer is gone before onRemove is called

thebigredgeek:
I received the error `Cannot read property 'then' of undefined` when using famous-angular due to the line edited below.  This appears to be due to the element being removed before onRemove is being called, and this appears to cause `$animate.leave(self.contentContainer)` to return undefined instead of a promise.

### DIFF
--- a/src/components/tabs/js/tabItemController.js
+++ b/src/components/tabs/js/tabItemController.js
@@ -5,7 +5,7 @@
 angular.module('material.components.tabs')
   .controller('$mdTab', TabItemController);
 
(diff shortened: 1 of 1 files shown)


epayet:
+1

ThomasBurleson:
Which version of Angular Material is your code based upon?

> It references `self.hammertime.destroy();` which is no longer used. And Tabs has be rearchitected with the lastest code in master (to be formally released as 0.9.0)

Please rebase your code off the latest codebase in master and re-test.

PR #909, opened 2014-12-11
docs(mdTextField): add md-padding and remove inline padding

Stroemgren:
Apply md-padding to content instead of using inline padding


ThomasBurleson:
Component has been deprecated.

PR #939, opened 2014-12-14
Fix layout-padding css

epelc:
Adds a `,` which was missing in 5caa22b2c444485b3d340cda662d8808b1fc381d when `layout-padding` was readded. Also removes a comment saying `layout-padding` is depreciated since it has been readded.


ajoslin:
Thanks, but this code is outdated now. We've removed the + rule as @gkalpak said.

PR #952, opened 2014-12-16
fix(layout): fix 'layout-padding'

TheHippo:
make sure css properties of layout-padding are applied correctly to elements with 'layout-padding' and 'flex' elements below


ajoslin:
Merged, thanks.

PR #977, opened 2014-12-18
fix(mdUtil): fix `throttle()` delay check

gkalpak:
Previously, the check was `recent - now > delay`, but it would never evaluate to `true`, since `now > recent`.


ThomasBurleson:
@gkalpak - I will close for next commit. But... it would be great to have a unit test to confirm this. ;-)

gkalpak:
Tests for `throttle()` added with #1236.

/ping @ThomasBurleson

PR #990, opened 2014-12-19
feat(mdRadioGroup): Radio submits on keydown/enter

### LINKED ISSUE #577: Focused radio buttons don't submit form on Enter
To reproduce:
Construct a `<form>` with `md-radio-group` any positive number of `md-radio-button` elements and a submit button.
Use the keyboard to focus the radio button group.
Press Enter.
Notice that the submit button is not activated.

Expected result:
The submit button should be activated.

Rationale:
Typical user agent behavior for forms is that pressing Enter while focused on a form element triggers an "implicit" submit (activate the submit button, if one is available). This is the behavior of the native control in Firefox and Chrome. http://www.w3.org/TR/html5/forms.html#implicit-submission

marcysutton:
With this change, mdRadioGroup will submit a form on a press of the enter key. The directive checks for a parent node with the tagName of `form` and if it finds a match, it will trigger a submit event.

Closes #577

### DIFF
--- a/src/components/radioButton/radioButton.js
+++ b/src/components/radioButton/radioButton.js
@@ -65,13 +65,25 @@ function mdRadioGroupDirective($mdUtil, $mdConstant, $mdTheming) {
     var ngModelCtrl = ctrls[1] || $mdUtil.fakeNgModel();
 
     function keydownListener(ev) {
-      if (ev.keyCode === $mdConstant.KEY_CODE.LEFT_ARROW || ev.keyCode === $mdConstant.KEY_CODE.UP_ARROW) {
-        ev.preventDefault();
-        rgCtrl.selectPrevious();
-      }
-      else if (ev.keyCode === $mdConstant.KEY_CODE.RIGHT_ARROW || ev.keyCode === $mdConstant.KEY_CODE.DOWN_ARROW) {
-        ev.preventDefault();
-        rgCtrl.selectNext();
+      switch(ev.keyCode) {
+        case $mdConstant.KEY_CODE.LEFT_ARROW:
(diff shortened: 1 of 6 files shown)


src/components/radioButton/radioButton.js
```diff
@@ -73,6 +73,12 @@ function mdRadioGroupDirective($mdUtil, $mdConstant, $mdTheming) {
         ev.preventDefault();
         rgCtrl.selectNext();
       }
+      else if (ev.keyCode === $mdConstant.KEY_CODE.ENTER) {
+        var form = angular.element($mdUtil.getClosest(element[0], 'form'));
+        if (form.length > 0) {
+          form.triggerHandler('submit');
+        }
+      }
     }
```
ThomasBurleson: I think a `switch` is better here:

```js
switch(ev.keyCode)
{
  case $mdConstant.KEY_CODE.LEFT_ARROW:
  case $mdConstant.KEY_CODE.UP_ARROW:
    ev.preventDefault();
    rgCtrl.selectPrevious();
    break;

  case $mdConstant.KEY_CODE.RIGHT_ARROW:
  case $mdConstant.KEY_CODE.DOWN_ARROW:
    ev.preventDefault();
    rgCtrl.selectNext();
    break;

  case $mdConstant.KEY_CODE.ENTER:
    var form = angular.element($mdUtil.getClosest(element[0], 'form'));
    if (form.length > 0) {
      form.triggerHandler('submit');
    }
    break;
}

```

marcysutton:
Closed with 03c7592798f904ac7a59b4a1c580672ca7c4789f

PR #1012, opened 2014-12-20
fix(radioButton): fix keyboard navigation

alirezamirian:
#### Bug description:
If there are disabled `md-radio-button`s inside a `md-radio-group`, (like the second
example of the [demo](https://material.angularjs.org/#/demo/material.components.radioButton)), it's not possible to select the next enable 'md-radio-button' with arrow keys.
#### Fix:
In the `changeSelectedButton` function of `mdRadioGroupDirective`,
instead of `'md-radio-button'`, use `'md-radio-button:not([disabled])'` as
selector, to exclude disabled items.


ThomasBurleson:
Fixed in SHA f520d50

PR #1032, opened 2014-12-22
fix(mdButton): fab vertical positioning

### LINKED ISSUE #1031: mdButton: md-fab-top-* are not properly aligned when using tabs
Not sure if this is the intended behavior, since the class isn't documented yet.

Here's how it looks like by default:
image
and with a little CSS fix to reposition the button:
image

Source:
```html
<body ng-app="YourApp">
  <div layout="column" layout-fill>
    <md-tabs>
      <md-tab>Tab 1</md-tab>
      <md-tab>Tab 2</md-tab>
    </md-tabs>
    
    <md-button class="md-accent md-fab md-fab-top-right">Yo</md-button>
  </div>
</body>
```
```css
/* Quick & dirty positioning fix */
.md-fab-top-right {
  top: 20px !important;
}
```
[Plunker](http://plnkr.co/edit/6Gr6ZOezfnnAYFcgJ6ci?p=preview)

By default, those buttons have a `top: 28px` rule applied to them, which comes from [here](https://github.com/angular/material/blob/master/src/components/button/button.scss#L78-L81).

PaulMougel:
Affects the following CSS classes:
md-fab-bottom-right
md-fab-bottom-left
md-fab-top-right
md-fab-top-left

Closes #1031.

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -75,10 +75,10 @@ $button-fab-toast-offset: $toast-height + $toast-margin;
 
   &.md-fab {
 
-    @include fab-position(bottom-right, auto, $button-fab-width/2, $button-fab-height/2, auto);
-    @include fab-position(bottom-left, auto, auto, $button-fab-height/2, $button-fab-width/2);
-    @include fab-position(top-right, $button-fab-height/2, $button-fab-width/2, auto, auto);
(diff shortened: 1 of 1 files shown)


ajoslin:
Thanks again @PaulMougel! Merged.

I also added a tweak to the commit to make the fab with toast's offset look good with this change.

Issue #1072, opened 2014-12-30
md-tabs, object is not a function error


jamm:
On load:
TypeError: object is not a function
    at angular-material.js:6866
    at angular.js:16205
    at completeOutstandingRequest (angular.js:4902)
    at angular.js:5282

and this line points to function:
```js
function waitForVisible() {
        return watcher || scope.$watch(
            function () {
              $timeout(function () {
                if (element[0].offsetParent) {
                  watcher(); // <-- here is error
                  debouncedUpdatePagination();
                  watcher = null;
                }
              }, 0, false);
            }
        );
      }
```
v0.7.0-rc1-master-84842ff

palmani:
Hi,

I'm also getting these below error, while using md-tabs inside ionic modal popup.
Please help me to fix this.
```
  TypeError: object is not a function
    at http://localhost:8100/lib/ionic/js/angular-material/angular-material.js:6866:19
    at http://localhost:8100/lib/ionic/js/angular/angular.js:16174:28
    at completeOutstandingRequest (http://localhost:8100/lib/ionic/js/angular/angular.js:4889:10)
    at http://localhost:8100/lib/ionic/js/angular/angular.js:5269:7
```
Note: Also, I have checked md-tabs working without using ionic modal popup.
I'm using the below the versions:
angularjs: v1.3.6
angular-material: v0.7.0-rc1-master-98247bc

Thanks,
Palmani. K

ThomasBurleson:
Which version of Ionic are you using? - This is more than likely an integration issue...
Please provide a GitHub, CodePen, or Plunkr sample.

jamm:
it's not integration issue, error is in your code and pretty obvious, see
my PR please.

ThomasBurleson:
@jamm - Where is the PR ? Do you have a PR link ?

jamm:
Link to PR: https://github.com/jamm/material/commit/59379b4e65f78cfe50e86b13c4ae80f7e5e2d0dc
Sorry, I thought it's visible for all: 
![screenshot 2015-01-04 20 52 25](https://cloud.githubusercontent.com/assets/526352/5606838/ae964ea8-9453-11e4-86e8-1465fd265717.png)

PaulMougel:
Correct link to PR: https://github.com/angular/material/pull/1073

ThomasBurleson:
Fixed e952ab4

PR #1073, opened 2014-12-30
Only call watcher() when it's a function

jamm:
Fix for #1072, maybe not most elegant, but pretty simple and works fine.


ThomasBurleson:
@jamm - Let's add this for now. 
> But in the future, we will accept PRs for source changes only if additional unit tests (*.spec.js) have been added.

PR #1074, opened 2014-12-30
fix error when md-button is disabled anchor

nikita-yaroshevich:
attr.ngDisabled can't be trackable by $watch because it is not inside the scope.

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -79,7 +79,7 @@ function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria) {
     // For anchor elements, we have to set tabindex manually when the 
     // element is disabled
     if (isAnchor(attr)) {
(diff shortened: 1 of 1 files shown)


gkalpak:
@ThomasBurleson: Are you sure this patch should have been merged ?
`ngDisabled` is supposed to hold an expression ([docs][1]). It is this expression that needs to be evaluated and watched and not `attrs.ngDisabled` (which will always return the expression string).

For an example of what I mean, take a look at this [demo][2].

[1]: https://docs.angularjs.org/api/ng/directive/ngDisabled#usage
[2]: http://jsfiddle.net/ExpertSystem/p38hck11/2/

ThomasBurleson:
@gkalpak - thank you for the recheck. Somehow I merged this without testing myself. I will resolve today.

Issue #962, opened 2014-12-17
Responsive design can be improved for small devices


MCBL:
Toolbar title, breadcrumb and directive name are poorly displayed:
![capture d ecran 2014-12-17 a 13 33 49](https://cloud.githubusercontent.com/assets/6067349/5471094/886c9b76-85f1-11e4-8779-9be9afee7159.png)

A proposed solution would be to switch to menu when width is smaller than 771 pixels instead of 600.

marcysutton:
Closed with bfc947f66b2a2568dc76ca3278eb9b5f83424a2f

MCBL:
It is still poorly displayed. Juste resize your browser on page with a lot of text. It's ugly!
![capture d ecran 2015-01-12 a 11 46 24](https://cloud.githubusercontent.com/assets/6067349/5701800/d74eea6e-9a50-11e4-8dc6-d722e8c50f5d.png)

marcysutton:
Are you looking at material.angularjs.org or a local build from master? We are still working to get the site updated, which won't reflect these changes.

MCBL:
At material.angularjs.org. Ok all right, thanks.

ashishpatel02:
@ThomasBurleson  as you said "When the content width [box area with Basic Usage] is great than 900px, then the SideNav component is locked into the open position. Less than that and it auto-hides."

How can i fix to open for all the resolutions?

ThomasBurleson:
@ashishpatel02 - try `md-is-locked-open="true"`

Issue #916, opened 2014-12-11
Improve UX for Material API documentation


ThomasBurleson:
The current API section of the Angular Material documentation needs improvements. We should consider:

 - the API presentation in [UI Router](http://angular-ui.github.io/ui-router/site/#/api/ui.router.util.$templateFactory)
 - the [API documentation](https://docs.angularjs.org/api/ng/directive/ngClick) for AngularJS 1.x
 - the planned API docs for Angular 2.x. (talk with @petebacondarwin )

 > We may need to enhance our jsDocs, update the gulp and CSS, etc.

<h2>Checklist:</h2>

- [x] Docs menu items in more common Guides/Tutorials/Resource structure (IA), restyled to use [expand/collapse list controls](http://www.google.com/design/spec/components/list-controls.html#list-controls-types-of-list-controls)  https://github.com/angular/material/pull/1116
- [x] Add real-world examples https://github.com/angular/material/issues/1102
- [x] Restyle API services & directives into more readable format https://github.com/angular/material/issues/984
- [ ] Simplify API toolbar buttons, add Search https://github.com/angular/material/issues/917
- [ ] Add version picker https://github.com/angular/material/issues/918

epelc:
Please do not base it too heavily on `ui-router`'s docs. I find them extremely lacking and I always have to jump between them and their wiki pages.

marcysutton:
@epelc the only part of UI-Router's docs that was considered was how methods were displayed. But I will keep that in mind.

Issue #772, opened 2014-11-26
The styles for `hide/show[-xx[-yy]]` are wrong


gkalpak:
I think that the current implementation of `hide[-xx[-yy]]`/`show[-xx[-yy]]` is kind of broken and confusing (not sure if it is so by design or as a result of a bug in style declarations).

Based on the current styles it seems that `show-xx-yy` can negate `hide-xx-yy`, which seems to have little point (why would I want to define `show-xx-yy` and `hide-xx-yy` on the same element). Furthermore it comes to contradiction with **[the docs](https://material.angularjs.org/#/layout/options)**, which state that `show[-xx[-yy]]` _"negates hide"_ (not `hide[-xx[-yy]]`).

It is makes much more sense (imo) anyway:
Either show the element "by default" and use appropriate `hide-xx[-yy]` attributes to hide it on specific devices or hide the element "by default" (using `hide`) and use appropriate `show-xx[-yy]` attributes to negate `hide` and make the element visible on specific devices.
Mixing both approaches won't help.

--
That said, I believe that:

1. `show` is unnecessary (since elements are shown (i.e. not hidden)) without needing any attribute (and overriding `hide` across all devices doesn't make sense).

2. `show[-xx[-yy]]` should **only** negate `hide` (on appropriate screen-sizes).

According to the above, here are the selectors that should have a `display: none;` style per screen-size:

**any:**      0 < size

  * [hide]:not([show])

**sm:**       0 < size <  600

  * [hide-sm]
  * [hide]:not([show]):not([show-sm])

**gt-sm:**  600 < size

  * [hide-gt-sm]
  * [hide]:not([show]):not([show-gt-sm])

**md:**     600 < size <  960

  * [hide-md]
  * [hide]:not([show]):not([show-md]):not([show-gt-sm])

**gt-md:**  960 < size

  * [hide-gt-md]
  * [hide]:not([show]):not([show-gt-md]):not([show-gt-sm])

**lg:**     960 < size < 1200

  * [hide-lg]
  * [hide]:not([show]):not([show-lg]):not([show-gt-md]):not([show-gt-sm])

**gt-lg:** 1200 < size

  * [hide-gt-lg]
  * [hide]:not([show]):not([show-gt-lg]):not([show-gt-md]):not([show-gt-sm])

--
BTW, this is related to the problem described in #764.

---
<sub>I hope all this makes sense, because I used `make sense` too many times...</sub>

ajoslin:
[...]

PR #1086, opened 2014-12-31
Responsive layout fix

marcysutton:
To address issues in https://github.com/angular/material/issues/962, this PR includes a style refactor for mdToolbar and markup changes in the docs.

An update to the toolbar icons will come later as part of https://github.com/angular/material/issues/916 (removing icon text and updating artwork to be more semantic).

Closes https://github.com/angular/material/issues/962

### DIFF
--- a/docs/app/css/style.css
+++ b/docs/app/css/style.css
@@ -145,6 +145,7 @@ code:not(.highlight) {
 .menu-icon {
   background: none;
   border: none;
+  margin-right: 16px;
 }
 .app-toolbar .md-toolbar-tools h3 {
   -webkit-margin-before: 0;
--- a/docs/config/template/index.template.html
+++ b/docs/config/template/index.template.html
@@ -21,8 +21,8 @@
               md-component-id="left"
               md-is-locked-open="$media('gt-sm')">
 
-    <md-toolbar style="min-height: 64px; max-height:64px;" layout="column">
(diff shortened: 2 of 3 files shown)


gkalpak:
The implementation and semantics of `hide-lt-md` are wrong.
If nothing else it sounds more like a `hide-lt-lg`.

This again brings up the issue raised in #772 (which has been long ignored/"neglected"), namely:
`show-/hide-{size}` attributes are implemented in a way that is (more or less) wrong.

marcysutton:
Yes, I definitely had trouble figuring out what attributes to use. The combination I *thought* would accomplish what I wanted (hide something when the screen is less than 960w) was `[hide][show-gt-md]` but it didn't didn't hide anything, if I remember correctly. If there's an attribute combination that accomplishes what I'm after without adding a new attribute, sweet! If we can keep this issue moving forward without being delayed by another outstanding one, that would be ideal.

mzbyszynski:
@marcysutton `hide-sm hide-md` should work, if you don't mind the verbosity...

marcysutton:
That works–I've updated the code to work without adding a new attribute. They could definitely be more intuitive.

marcysutton:
Closed with bfc947f66b2a2568dc76ca3278eb9b5f83424a2f

PR #1122, opened 2015-01-06
fix(switch): adds tabindex to switches and checkboxes according to

### LINKED ISSUE #1038: mdSwitch: Weird tab behavior on the demo page
On [the demo page](https://material.angularjs.org/#/demo/material.components.switch), the user can tab through switches, including a disabled switch (the 4th one). A non-disabled switch (the 5th) can't be selected though.

untitled

Affects Chrome, Firefox and Safari.

robertmesserle:
disabled state

Closes #1038.

### DIFF
--- a/src/components/checkbox/checkbox.js
+++ b/src/components/checkbox/checkbox.js
@@ -46,7 +46,7 @@ angular.module('material.components.checkbox', [
  * </hljs>
  *
  */
-function MdCheckboxDirective(inputDirective, $mdInkRipple, $mdAria, $mdConstant, $mdTheming, $mdUtil) {
+function MdCheckboxDirective(inputDirective, $mdInkRipple, $mdAria, $mdConstant, $mdTheming, $mdUtil, $parse) {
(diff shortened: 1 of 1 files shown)


src/components/checkbox/checkbox.js
```diff
@@ -68,11 +68,22 @@ function MdCheckboxDirective(inputDirective, $mdInkRipple, $mdAria, $mdConstant,
...
 
+    var disabledParsed = $parse(tAttrs.ngDisabled) || angular.noop;
+
     tAttrs.type = 'checkbox';
     tAttrs.tabIndex = 0;
     tElement.attr('role', tAttrs.type);
 
     return function postLink(scope, element, attr, ngModelCtrl) {
+      scope.$watch(function () {
+        return typeof disabledParsed === 'function' && disabledParsed(scope);
```
ThomasBurleson: @robertmesserle - 
*  why are you check `typeof` here? L71 will always set a function, right?
*  the title to this PR does not match. Thoughts ?

PR #1123, opened 2015-01-06
fix(mdRadioButton): Loosen equality check to allow numerical values as well as strings

### LINKED ISSUE #1112: radioButton: integer values are not working because of strict comparison
If the `md-radio-group` model is defined as integer, it doesn't display checked status (`md-checked` class) because of the strict `===` value comparison [here](https://github.com/angular/material/blob/master/src/components/radioButton/radioButton.js#L237):
```javascript
var checked = (rgCtrl.getViewValue() === attr.value);
```

Bug reproduce suite: http://codepen.io/artch/pen/jEVKYe?editors=101

marcysutton:
Closes #1112

### DIFF
--- a/src/components/radioButton/demoBasicUsage/script.js
+++ b/src/components/radioButton/demoBasicUsage/script.js
@@ -9,8 +9,8 @@ angular.module('radioDemo1', ['ngMaterial'])
   };
 
   $scope.radioData = [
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
Closed with SHA  ca3e4c3

PR #1124, opened 2015-01-06
style(mdButton): Changes to medium font weight

### LINKED ISSUE #1108: md-button: bolder font
Hi,

according to http://www.google.com/design/spec/style/typography.html#typography-standard-styles

> Button style (Medium 14sp, all caps) is used for **all** buttons, whether they are ink or material.

(Emphasis on ALL)

However, current implementation of md-button's font is not as bold as the examples from the page.

Do you think the style for the font should be bold by default? I find `font-weight: 500` looks very similar with the examples.

Thanks

marcysutton:
Closes #1108

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -36,7 +36,7 @@ $button-fab-toast-offset: $toast-height + $toast-margin;
 
   // Always uppercase buttons
   text-transform: uppercase;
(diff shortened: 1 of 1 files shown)


marcysutton:
Closed with 5e489e96b56a998655433aa43086e76e04336552

PR #1130, opened 2015-01-06
mdRadioButton: added support for rtl

### LINKED ISSUE #1068: mdRadioButton: Add support for rtl in inner text
Hi well it's simple..
all you have to do is adding ```margin-right:10px``` like this:

```css
.md-label {
    (...)
    margin-right: 10px;
}
```

in [here](https://github.com/angular/material/blob/master/src/components/radioButton/radio-button.scss#L63)
here's a [plunker](http://embed.plnkr.co/CCNyosIMWpUcIndhwaHv)

should i PR?

EladBezalel:
Closes #1068 
Adding support for bi-directional

### DIFF
--- a/src/components/radioButton/radio-button.scss
+++ b/src/components/radioButton/radio-button.scss
@@ -61,6 +61,7 @@ md-radio-button,
     position: relative;
     display: inline-block;
     margin-left: 10px;
+	margin-right: 10px;
     vertical-align: middle;
     white-space: normal;
     pointer-events: none;


src/components/radioButton/radio-button.scss
```diff
@@ -61,6 +61,7 @@ md-radio-button,
     position: relative;
     display: inline-block;
     margin-left: 10px;
+	margin-right: 10px;
```
gkalpak: The indentation seems to be off...

EladBezalel: what do you mean?

PaulMougel: This file is indented using spaces, but your change uses a tab character; which is why your line is more indented than it should be.

PR #1151, opened 2015-01-10
fixed sticky translate X

dafunkk:
In the case were sticky elements had a left margin to the container, the sticky elements were being translated to the right just before the next element arrives. But the sticky element gets a margin-left, when translating element up from the top as it scrolls into view, it was also translated to the right. (I don't know if I am clear, put subheader elements in a scrolling container where they are not aligned to the left but centered)

[Review on Reviewable](https://reviewable.io/reviews/angular/material/1151)

### DIFF
--- a/src/components/sticky/sticky.js
+++ b/src/components/sticky/sticky.js
@@ -244,7 +244,7 @@ function MdSticky($document, $mdConstant, $compile, $$rAF, $mdUtil) {
        item.translateY = amount;
        item.clone.css(
          $mdConstant.CSS.TRANSFORM, 
-         'translate3d(' + item.left + 'px,' + amount + 'px,0)'
+         'translate3d(0,' + amount + 'px,0)'
        );
      }
    }


dafunkk:
Just signed the CLA.

ThomasBurleson:
@dafunkk - do you have a plunkr that demonstrates the original bug and the fix?

ThomasBurleson:
@robertmesserle - seem reasonable ?

robertmesserle:
@ThomasBurleson Seems reasonable, but I'm not familiar enough to say that with much confidence.  I'd also like to see a Plunker sample first.

ThomasBurleson:
@robertmesserle - my thoughts also.

dafunkk:
I will put a plunkr together to demonstrate the problem.

marcysutton:
I got around something similar in the Subheader demo by removing padding and margins from the parent container, and instead applying it to the children (md-list). I'd try insetting it on a parent container. `$mdSticky` really needs a refactor though.

EladBezalel:
@dafunkk can you please check if it's still happening on HEAD, closing for now, if it's still happening please rebase and i will reopen.

PR #1158, opened 2015-01-11
refactor(tabs): use md-input-container as input container

patrykkopycinski:


src/components/tabs/demoDynamicTabs/index.html
```diff
@@ -16,13 +16,25 @@
 
   <form ng-submit="addTab(tTitle,tContent)" flex>
     <div layout="row" layout-sm="column" layout-margin layout-align="left center">
-      <md-text-float label="Active Index" ng-model="selectedIndex" disabled></md-text-float>
-      <md-text-float label="Active Title" ng-model="tabs[selectedIndex].title"></md-text-float>
+      <md-input-container>
+        <label>Active Index</label>
+        <input type="text" ng-model="selectedIndex">
```
gkalpak: I think this should be disabled.

patrykkopycinski:
I thought the same, but when i compared it with current docs:
https://material.angularjs.org/#/demo/material.components.tabs

it seems it doesn't work, so i decided to not disable the input, because i didn't want to change the behavior after refactor, but maybe i missed something.

What do you think? Should i add the disabled attr?

src/components/tabs/demoDynamicTabs/index.html
patrykkopycinski: I thought the same, but when i compared it with current docs:
https://material.angularjs.org/#/demo/material.components.tabs

it seems it doesn't work, so i decided to not disable the input, because i didn't want to change the behavior after refactor, but maybe i missed something.

What do you think? Should i add the disabled attr?

gkalpak: I definitely think you should make it disabled. The fact that it does not work in the current docs, is a bug (and one of the main reasons for implementing `md-input-container`; i.e. the fact that `md-text-float` did not propagate the attributes to the `input`).

patrykkopycinski: Thanks @gkalpak 

I added disabled attribute and also when i was checking it on the browsers i found bug that on IE11 disabled input hadn't dotted background, so i fixed it also. I hope it'll be helpful.

ThomasBurleson:
@patrykkopycinski - plz sign the CLA to allow this PR to be merged.

patrykkopycinski:
Hi @ThomasBurleson i signed CLA before my first PR to angular-material has been merged. Could you check it again? 

Thanks

ThomasBurleson:
@patrykkopycinski - please rebase your PR with the latest changes from `master`. Afterwards, I will be able to easily merge your PR. Thank you.

ThomasBurleson:
Already fixed.

PR #1168, opened 2015-01-12
fix(button): better support for ngDisabled and tabindex attributes

ThomasBurleson:
Refs #1074.

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -54,34 +54,76 @@ function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria) {
     link: postLink
   };
 
-  function isAnchor(attr) {
-    return angular.isDefined(attr.href) || angular.isDefined(attr.ngHref);
+  function postLink(scope, element, attr) {
+    var node = element[0];
+    var elementHasText = node.textContent.trim();
+
+    $mdTheming(element);
+    $mdInkRipple.attachButtonBehavior(scope, element);
+
+    if (!elementHasText) {
+      $mdAria.expect(element, 'aria-label');
+    }
+
+    updateTabIndex(scope, element, attr);
+
   }
-  
+
+  // ************************************************
+  // Internal Methods
+  // ************************************************
+
+  /**
+   * Publish either an <a/> or <button/> template
+   * @returns {string}
+   */
   function getTemplate(element, attr) {
     return isAnchor(attr) ?
            '<a class="md-button" ng-transclude></a>' :
            '<button class="md-button" ng-transclude></button>';
   }
 
-  function postLink(scope, element, attr) {
-    var node = element[0];
-    $mdTheming(element);
-    $mdInkRipple.attachButtonBehavior(scope, element);
+  /**
+   * Should we use an <a/> template
+   * @returns {boolean|*}
+   */
+  function isAnchor(attr) {
+    return angular.isDefined(attr.href) || angular.isDefined(attr.ngHref);
+  }
 
-    var elementHasText = node.textContent.trim();
-    if (!elementHasText) {
-      $mdAria.expect(element, 'aria-label');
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@marcysutton - can you review this for ARIA correctness. Thx.

src/components/button/button.js
```diff
@@ -54,34 +54,76 @@ function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria) {
...
+  function isAnchor(attr) {
+    return angular.isDefined(attr.href) || angular.isDefined(attr.ngHref);
+  }
+
+  /**
+   * Is this a simulate button ?
+   * @returns {boolean}
+   */
+  function isButton(attr) {
+    return !isAnchor(attr) && (attr.role == "button");
```
marcysutton: @ThomasBurleson what is the use-case for this? We use native button elements that won't have a role attribute.

src/components/button/button.js
```diff
@@ -54,34 +54,76 @@ function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria) {
...
+  }
+
+  /**
+   * Update the accessibility `tabindex` property
+   */
+  function updateTabIndex(scope, element, attr) {
+    if ( isAnchor(attr) ) {
+
+      // For anchor elements, we have to set tabindex manually
```
marcysutton: Shouldn't `<button>` get the same treatment (change tabindex to -1 when disabled)? To reset to the default, we can either remove the attribute (since they are interactive by default), or reset to the original tabIndex provided by the developer.

src/components/button/button.js
```diff
@@ -54,34 +54,76 @@ function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria) {
...
+
+        scope.$watch(attr.ngDisabled, function(isDisabled) {
+          element.attr('tabindex', isDisabled ? -1 : 0);
+        });
+
+      }
+
+    } else if ( isButton(attr) ) {
+
+      $mdAria.expect(element, 'tabindex', "0" );
```
ajoslin: We don't have to expect a tabIndex on buttons.

src/components/button/button.js
```diff
@@ -54,34 +54,76 @@ function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria) {
...
-  
-  function getTemplate(element, attr) {
-    return isAnchor(attr) ?
-           '<a class="md-button" ng-transclude></a>' :
-           '<button class="md-button" ng-transclude></button>';
-  }
-
   function postLink(scope, element, attr) {
     var node = element[0];
+    var elementHasText = node.textContent.trim();
```
ajoslin: I still don't like this approach of checking if an element has text  - since it may get its text later.  We should reconsider this in the future.

src/components/button/button.js
ajoslin: Doesn't button handle that natively for disabled?

marcysutton: Yep! You're totally right. It does get removed from the tab order.

marcysutton: Expanding on this a little further, I created a demo that shows what happens with links vs. buttons when ngAria and ngDisabled are present: http://codepen.io/marcysutton/pen/Lsfhk

TL:DR; Buttons get removed from the tab order unless `tabIndex="0+"` has been added manually. Links do not get removed from the tab order even with `ngDisabled`.

src/components/button/button.js
marcysutton: +1

marcysutton:
@ThomasBurleson what's the status of this PR? I just noticed disabled checkboxes have `tabIndex="0"` even though they shouldn't, probably because of ngAria.

ThomasBurleson:
Will not merge.

PR #1185, opened 2015-01-13
Fix typos in CHANGELOG for md-textfloat

Splaktar:
The instructions say to change md-textfloat to the new implementation, but it was actually called md-text-float.

Fixing this to help others who are migrating to the new version so they won't be confused like I was.


ajoslin:
Merged. thanks.

PR #1200, opened 2015-01-15
md-input-container: add dotted line on disabled input for IE10

yarl:
It looks like IE10 is not fully supporting `linear-gradient`. This fix resolve it.

### DIFF
--- a/src/components/input/input-theme.scss
+++ b/src/components/input/input-theme.scss
@@ -61,5 +61,6 @@ md-input-container.md-THEME_NAME-theme {
     border-bottom-color: transparent;
     color: '{{foreground-3}}';
     background-image: linear-gradient(to right, '{{foreground-4}}' 0%, '{{foreground-4}}' 33%, transparent 0%);
+    background-image: -ms-linear-gradient(left, transparent 0%, '{{foreground-4}}' 100%);
   }
 }


yarl:
@googlebot: Done.

Frank3K:
Are you sure this is the issue? [caniuse](http://caniuse.com/#search=css%20gradient) reports full support for CSS gradients in IE 10+. Autoprefixer otherwise would have added the necessary `-ms`-prefix.

Note: I haven't tested it in IE10 myself.

yarl:
@Frank3K: yes, but somehow it's not working, no line at all.

yarl:
OK, so after inspecting it on IE11 (where dev tools are actually working) and taking a look at [IE guidelines](http://msdn.microsoft.com/en-us/library/ie/hh673532(v=vs.85).aspx) it looks that `transparent 0%` is the problem. So basically next percentage cannot be lower than previous.

After checking `background-image` as below:
`linear-gradient(to right, rgba(0,0,0,0.12) 0%, rgba(0,0,0,0.12) 33%, transparent 34%)`
it looks good on Chrome, IE10 and IE11, so probably it resolves problem.

marcysutton:
Closed with be87c3453bab2148931aba8f55576e7f1bbae080.

Issue #536, opened 2014-11-04
Demo's not working on IE11


Martinspire:
I was just checking the progress and noticed somebody saying on Google+ that some things weren't working in IE11. So here's a list of demo's not working properly:

* ~~Bottom Sheet - List: not showing items~~
* ~~Dialog - Simple: wrong postion and coming out of its container~~
* ~~Dialog - Complex: wrong position, z-index wrong and coming out of its container~~
* Divider: lists have scrolling (margin issue? without md-list padding it works)
* List: contents has scrolling like divider
* ~~Progress Linear: bars have wrong position and go outside container (wrong width)~~
* ~~Slider: width issue (container to big)~~
* ~~Tabs - Pagination: clicking titles doesn't show that page, only arrow works~~
* ~~Text Field: When clicking textfield with input it shifts the input vertically (but goes to original position)~~
* ~~Toast: No toast is shown, not at all~~

* ~~Grid System - layout padding, layout padding with offsets, layout order (albeit right order): wrong height of elements~~

So there is quite some work to be done to render it correctly in IE. Which you should do because it is needed for webapps to work on desktop or WP8.1.

Still, nice progress and hope to see more

epelc:
The grid system page is also really broken in ie11
https://material.angularjs.org/#/layout/grid
![image](https://cloud.githubusercontent.com/assets/5204642/5742050/48a282c4-9bdd-11e4-9f3d-9861c27a85ec.png)

epelc:
The following appears to be only in ie I checked chrome and it worked as expected
https://material.angularjs.org/#/demo/material.components.button
![image](https://cloud.githubusercontent.com/assets/5204642/5742383/9f8ce212-9bdf-11e4-8fcb-ccf7cbe91f73.png)

PR #1227, opened 2015-01-19
fix(layout): fix IE11 layout

patrykkopycinski:
Fixes for #536


patrykkopycinski:
Hi @marcysutton,

could you explain me why you closed my PR without merge commits? And instead of this you created your own branch with the same changes and then merged your branch?

marcysutton:
There was an issue with the patch where the changes wouldn't apply, so I had to recreate them in a new branch. We have a specific merge process to keep our changelog clean. I'll try to find a way around it next time.

PR #1235, opened 2015-01-20
style(tabs): md-tabs-content is now flex'd

PaulMougel:
At the moment, the content of an `<md-tabs>` component doesn't take all the vertical height of its parent:
capture d ecran 2015-01-20 a 09 54 07
which creates an ugly transition between tabs that don't have the same height ([Plunker](http://plnkr.co/edit/rmmE7aNWc6GyP0me5XcM?p=preview
)):
untitled

This commit simply adds a `flex` attribute to the `<section class="md-tabs-content">` element. So if you have `<md-tabs layout="column" layout-fill>`, the transition should look better:

untitled

### DIFF
--- a/src/components/tabs/js/tabsDirective.js
+++ b/src/components/tabs/js/tabsDirective.js
@@ -127,7 +127,7 @@ function TabsDirective($mdTheming) {
         '</button>' +
 
       '</section>' +
-      '<section class="md-tabs-content"></section>',
+      '<section class="md-tabs-content" flex></section>',
     link: postLink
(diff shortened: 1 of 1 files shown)


PaulMougel:
Forget about that, adding `flex` removes the ability to scroll inside the tab.

PaulMougel:
Reopening this PR.

At the moment, I am using this CSS to mock the functionality:
```css
.md-tabs-content {
  flex: 1;
  overflow: auto;
}
```

PR #1239, opened 2015-01-20
Wip icon

programmist:
Initial md-icon branch changes.


src/components/icon/icon.js
```diff
@@ -20,25 +21,55 @@ angular.module('material.components.icon', [
...
+      mdIconColor: '@'
+    },
     restrict: 'E',
-    template: '<object class="md-icon"></object>',
-    compile: function(element, attr) {
-      var object = angular.element(element[0].children[0]);
-      if(angular.isDefined(attr.icon)) {
-        object.attr('data', attr.icon);
+    replace: true,
+    template: '<span class="icon"></span>',
```
marcysutton: This should probably have `role="img"` since it represents an image. It also needs a text alternative, either with a child node containing offscreen text or an `aria-label` attribute.

src/components/icon/icon.js
```diff
@@ -20,25 +21,55 @@ angular.module('material.components.icon', [
...
+    link: function(scope, element, attr) {
+      if ($mdIcon[scope.mdIConKey]) {
+        // TODO: use from config.
+      } else {
+        element.append(angular.element('<i>').addClass(scope.mdIconKey));
```
marcysutton: To prevent the character from being read aloud in a screen reader, we can add `aria-hidden="true"` on the `<i>` element.

src/components/icon/icon.js
```diff
@@ -20,25 +21,55 @@ angular.module('material.components.icon', [
...
       }
+      if(angular.isDefined(attr.mdIconKey)) {
```
ThomasBurleson: Perhaps you meant `attr.mdIconSize` ?

src/components/icon/icon.js
```diff
@@ -20,25 +21,55 @@ angular.module('material.components.icon', [
...
+      attr.$observe('mdIconColor', function(color) {
+        if(color) {
+          element.css('color', scope.mdIconColor);
+        }
+      }, true);
     }
   };
 }
+
+function MdIconProvider() {
```
ThomasBurleson: Need comments and docs here to describe the intent of a Provider. How will provider access support extension or customization?

src/components/icon/icon.js
ThomasBurleson: @marcysutton  - so this:

```html
<span class="icon" role="img" arial-label=""></span>
```
And I would how we would  embed or associate a label with the key/alias?

programmist:
Thanks for the comments. My apologies for the shabby state of this branch.  I'm still new to the material-design SCM process and thought "wip" branches could be in an unfinished state. Had I known it needed to be in a "ready for review" state I would have waited on the PR until I felt it was ready.

src/components/icon/icon.js
marcysutton: The [Polymer core-icon](https://www.polymer-project.org/docs/elements/core-elements.html#core-icon) handles it by allowing the developer to provide an `alt` attribute, copying that content to `aria-label` and setting `role="img"`. If alt is provided as an empty string, it is hidden from the accessibility layer. If neither of those are present, it copies the key to the `aria-label`. This sounds like it could work well for Angular Material, too.

marcysutton:
Sorry to bombard you with feedback prematurely @programmist!

programmist:
No worries at all.  All good comments.  I'll make sure to incorporate them in my next PR.

src/components/icon/icon.js
programmist: I agree.  I don't like the verbose md-icon-*, but I thought that was required.  I'm ok with md-alias and md-size.  

Not sure what you mean by "using the require:"^mdIcon" rule."

ThomasBurleson: I was thinking the md-alias and md-size were directives that could require themselves to be children of md-icon; but they are **attributes** not directives.

src/components/icon/icon.js
ThomasBurleson: @marcysutton  - I like this, :+1:

ThomasBurleson:
@programmist - whenever the team has a PR; we use that as trigger to review, comment, and improve.
The PR becomes a great venue to collaborate.

programmist:
And I welcome that. I just misunderstood the process.

PR #1246, opened 2015-01-21
style(tabs): md-tabs-content is now flex'd

PaulMougel:
This makes transitions between tabs smoother, as the full tab's height is animated.

See my previous (botched) PR: #1235 for an explanation of what this does.

Note, in my app I am using this CSS to mock the functionality:

```css
.md-tabs-content {
  flex: 1;
  overflow: auto;
}
```

### DIFF
--- a/src/components/tabs/js/tabsDirective.js
+++ b/src/components/tabs/js/tabsDirective.js
@@ -127,7 +127,7 @@ function TabsDirective($mdTheming) {
         '</button>' +
 
       '</section>' +
-      '<section class="md-tabs-content"></section>',
+      '<section class="md-tabs-content" flex></section>',
     link: postLink
   };
 


ThomasBurleson:
@PaulMougel  - above you are using `overflow : auto` yet the PR has removed overflow (so the default is `overflow:visible`).

ThomasBurleson:
@robertmesserle - can you review this PR?

PaulMougel:
Note that I'm not really sure about the code, especially with respect to 74fe3eb19b097611ed17f2f1459a5682b043387a :-/

robertmesserle:
This change breaks the demo page by causing scrollbars to show up during content transitions.  That `overflow: hidden;` that was removed was important.  Is there a way that we can accomplish this without removing that?

PaulMougel:
Indeed. I removed the change about `overflow`.

However, using only `flex` (as per my current commit in this PR) doesn't work. One also has to add `layout="column" flex` to the `<md-tabs>` element. The, then lateral transitions work… but the vertical scroll doesn't.

I'm not much of a CSS guru however; I don't really understand the root cause of that.

robertmesserle:
I opened an issue and this will hopefully be addressed in 0.8.  Closing for now.

PR #1257, opened 2015-01-21
feat(gestures): add built in gesture system, remove HammerJS

ajoslin:
Merge this too: https://github.com/angular/bower-material/pull/16

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -10,57 +10,47 @@ angular.module('material.core', ['material.core.theming'])
   .config(MdCoreConfigure);
 
 function MdCoreInitialize() {
-  if (typeof Hammer === 'undefined') {
-    throw new Error(
-      'ngMaterial requires HammerJS to be preloaded.'
-    );
-  }
-  // By default, Hammer disables user selection on desktop if swipe is enabled.
-  // We don't want this, so we make sure Hammer doesn't set a user-select: none.
-  Hammer.defaults.cssProps.userSelect = '';
 }
 
 function MdCoreConfigure($provide, $mdThemingProvider) {
-  $provide.decorator('$$rAF', ['$delegate', '$rootScope', rAFDecorator]);
+  $provide.decorator('$$rAF', rAFDecorator);
 
   $mdThemingProvider.theme('default')
     .primaryPalette('indigo')
     .accentPalette('pink')
     .warnPalette('red')
     .backgroundPalette('grey');
+}
 
-  function rAFDecorator($$rAF, $rootScope) {
-    /**
-     * Use this to debounce events that come in often.
-     * The debounced function will always use the *last* invocation before the
-     * coming frame.
-     *
-     * For example, window resize events that fire many times a second:
-     * If we set to use an raf-debounced callback on window resize, then
-     * our callback will only be fired once per frame, with the last resize
-     * event that happened before that frame.
-     *
-     * @param {function} callback function to debounce
-     */
-    $$rAF.debounce = function(cb) {
-      var queueArgs, alreadyQueued, queueCb, context;
-      return function debounced() {
-        queueArgs = arguments;
-        context = this;
-        queueCb = cb;
-        if (!alreadyQueued) {
-          alreadyQueued = true;
-          $$rAF(function() {
-            queueCb.apply(context, queueArgs);
-            alreadyQueued = false;
-          });
-        }
-      };
+function rAFDecorator($delegate, $rootScope) {
+  /**
+   * Use this to throttle events that come in often.
+   * The throttled function will always use the *last* invocation before the
+   * coming frame.
+   *
+   * For example, window resize events that fire many times a second:
+   * If we set to use an raf-throttled callback on window resize, then
+   * our callback will only be fired once per frame, with the last resize
+   * event that happened before that frame.
+   *
+   * @param {function} callback function to debounce
+   */
+  $delegate.throttle = function(cb) {
+    var queueArgs, alreadyQueued, queueCb, context;
+    return function debounced() {
+      queueArgs = arguments;
+      context = this;
+      queueCb = cb;
+      if (!alreadyQueued) {
+        alreadyQueued = true;
+        $delegate(function() {
+          queueCb.apply(context, queueArgs);
+          alreadyQueued = false;
+        });
+      }
     };
-    return $$rAF;
-
-  }
-
+  };
+  return $delegate;
 }
 
 })();
--- a/src/core/services/gesture/gesture.js
+++ b/src/core/services/gesture/gesture.js
@@ -1,0 +1,375 @@
+(function() {
+'use strict';
+
+/*
+ * TODO: Add support for multiple fingers on the `pointer` object (enables pinch gesture)
+ */
+
+var START_EVENTS = 'mousedown touchstart pointerdown';
+var MOVE_EVENTS = 'mousemove touchmove pointermove';
+var END_EVENTS = 'mouseup mouseleave touchend touchcancel pointerup pointercancel';
+var HANDLERS;
+
+document.addEventListener('click', function(ev) {
+  // Prevent clicks unless they're sent by material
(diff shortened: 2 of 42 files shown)


ajoslin:
Tested on:

- [ ] Android 4.4
- [ ] Android 5.0
- [ ] iOS 8
- [ ] Firefox Desktop
- [ ] Safari Desktop
- [ ] Chrome Desktop

ilanbiala:
@ajoslin what's the benefit of this vs. HammerJS?

ajoslin:
HammerJS doesn't work on touch+click devices (Chrome Pixel).

HammerJS has lots of bugs with tapping on iOS and Android 4.3-.

ilanbiala:
@ajoslin does it make sense to extract the work in this PR to a touch library that can be used by others independent of angular-material? Is this something that might make sense to upstream to AngularJS itself?

ajoslin:
Yes, that's definitely something to think about.  Ionic may be using this
in the future, as well.

src/core/services/gesture/gesture.js
```diff
@@ -0,0 +1,359 @@
...
+  //         holdTriggered = true;
+  //       }, self.options.delay);
+  //     }
+  //   }
+  // });
+
+  addHandler('drag', /* @ngInject */ function($$rAF) {
+    var dragState;
+    var dragCancelled;
+    if (!$$rAF.throttle) {
```
ajoslin: remove

src/core/services/gesture/gesture.js
```diff
@@ -0,0 +1,359 @@
...
+
+    return {
+      reset: function() {
+        dragState = dragCancelled = null;
+      },
+      options: {
+        minDistance: 6,
+      },
+      shouldPreventMove: function(ev, pointer) {
+        var farEnoughY = Math.abs(pointer.distanceY) > this.options.minDistance;
```
ajoslin: Figure out vertical dragging.

src/core/services/gesture/gesture.js
```diff
@@ -0,0 +1,359 @@
...
+        if (!dragState) {
+          if (farEnoughY) {
+            dragCancelled = true;
+            return false;
+          }
+          return true;
```
ajoslin: Only need one return here.

src/core/services/gesture/gesture.js
```diff
@@ -0,0 +1,359 @@
...
+        } else {
+          return true;
+        }
+      },
+      onMove: function(ev, pointer) {
+        if (dragCancelled) return;
```
ajoslin: Built in way to cancel?

src/core/services/gesture/gesture.js
```diff
@@ -0,0 +1,359 @@
...
+          .filter(function(handler) {
+            return handler && handler.onMove !== angular.noop;
+          });
+
+        if (onMoveHandlers.length) {
+          element.on('touchmove', function preventDefaultListener(ev) {
+            var shouldPrevent = onMoveHandlers.some(function(handler) {
+              return handler.shouldPreventMove(ev, pointer);
+            });
+        console.log('shouldPreventMove', shouldPrevent);
```
ajoslin: remove test console

src/core/services/gesture/gesture.js
```diff
@@ -0,0 +1,359 @@
...
+          .map(function(handlerName) {
+            return HANDLERS[ handlerName.replace(/^\$md./, '') ];
+          })
+          .filter(function(handler) {
+            return handler && handler.onMove !== angular.noop;
+          });
+
+        if (onMoveHandlers.length) {
+          element.on('touchmove', function preventDefaultListener(ev) {
+            var shouldPrevent = onMoveHandlers.some(function(handler) {
```
ajoslin: Use something more efficient than `.some`

src/core/services/gesture/gesture.js
```diff
@@ -0,0 +1,359 @@
...
+        var onMoveHandlers = (handlers || '')
+          .split(' ')
+          .map(function(handlerName) {
+            return HANDLERS[ handlerName.replace(/^\$md./, '') ];
+          })
+          .filter(function(handler) {
+            return handler && handler.onMove !== angular.noop;
+          });
+
+        if (onMoveHandlers.length) {
```
ajoslin: All of this code should be moved into the handler, if possible.

src/core/services/gesture/gesture.js
```diff
@@ -0,0 +1,359 @@
...
+            element.off('touchmove', preventDefaultListener);
+          };
+        }
+        return angular.noop;
+      }
+    };
+
+  }
+
+  function preventDefaultListener(ev) {
```
ajoslin: outdated

ajoslin:
Improvements needed:

- Have each handler allow an `elements` array. These are elements registered with `$mdGesture.register`.
- Any element registered with `$mdGesture.register` can set options (minDistance, drag direction etc)
- Use the list of registered elements with their options (vertical vs horizontal, minDistance, etc) to decide whether a drag's touchmove event should prevent scroll or not.

ajoslin:
At the end of the day, handlers need to be more sophisticated than just a JSON object, for easier use.

pandaiolo:
Not sure if my question is appropriate in the context of this change : is this also impacting the tapping delay on mobile web ? I am currently using [Fastclick](https://github.com/ftlabs/fastclick) to mitigate it. Any thoughts / advice about it ? Thanks

gustavohenke:
@ajoslin,
I haven't read the full code, but with this patch, am I able to remove Hammer from my app while migrating existing code?
I'm using Hammer to handle some pan events related to the [Material Design gestures pattern](http://www.google.com/design/spec/patterns/gestures.html).

ajoslin:
@pandaiolo: FastClick isn't needed anymore.

@gustavohenke: Yes. Unless you multitouch gestures like pinching (not supported yet).

ajoslin:
This is ready to merge now.

src/core/core.js
```diff
@@ -10,57 +10,47 @@ angular.module('material.core', ['material.core.theming'])
...
+   * coming frame.
+   *
+   * For example, window resize events that fire many times a second:
+   * If we set to use an raf-throttled callback on window resize, then
+   * our callback will only be fired once per frame, with the last resize
+   * event that happened before that frame.
+   *
+   * @param {function} callback function to debounce
+   */
+  $delegate.throttle = function(cb) {
```
gkalpak: Why rename this to `throttle` ? It sounds more like a `debounce` (at least in the sense of `$mdUtils`' debounce/throttle).

In any case, if we stick with "throttle" there are several comments and variable names that be better updated to `throttle...` (instead of `debounce...`).

src/core/services/gesture/gesture.js
```diff
@@ -0,0 +1,375 @@
...
+    },
+    onCancel: function() {
+      $timeout.cancel(this.state.timeout);
+    },
+    onStart: function(ev, pointer) {
+      // For hold, require a parent to be registered with $mdGesture.register()
+      // Because we prevent scroll events, this is necessary.
+      if (!this.state.registeredParent) return this.cancel();
+
+      this.state.pos = {x: pointer.x, y: pointer.x};
```
CasiOo: Is it intentional that y is set to pointer.x ?
Or should it be
```
this.state.pos = {x: pointer.x, y: pointer.y};
```

ThomasBurleson: @CasiOo - looks like a great bug catch. :+1:

ajoslin: Thanks. Will fix and add a unit test for holding and moving in the y direction.

ajoslin:
Merged via 8364fb57a9ac1b211c09ff564fea6ad0dea94e61

PaulMougel:
For some reason, this commit breaks all ripples.

I haven't been able to reproduce the issue on a Plunker however, so this may be a mistake on my side.

PR #1275, opened 2015-01-23
0.7.1 fixes

robertmesserle:
These are some fixes that were left out of master during 0.7.0 release.

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -5,6 +5,8 @@ $button-padding: 2px 6px 3px !default;
 $button-fab-width: 56px !default;
 $button-fab-height: 56px !default;
 $button-fab-padding: 16px !default;
+$button-fab-mini-width: 40px !default;
+$button-fab-mini-height: 40px !default;
 
 $button-fab-toast-offset: $button-fab-height * 0.75;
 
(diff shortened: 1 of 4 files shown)


mzbyszynski:
I put in a PR for a mini fab (#1174) a while ago. I considered adding an `md-mini` class, but after playing around with it for a bit that approach started to seem kind of verbose and clunky to me. There are already several `md-*` classes that are relevant only to `.md-button.md-fab` and having to string together a bunch of classes on a custom widget does not seem as elegant an API as many of the other Angular Material components have. For example:

```html
<md-button class="md-fab md-fab-bottom-right md-mini md-primary md-hue-1">
```
I guess it isn't that big of deal, but I also started to think that having a class with such a specific function that is only applicable to one widget that has such a generic name (`md-mini` that is) might be confusing for users--sort of how there are frequent issues/forum posts from new users that are confused about why they can't just stick `md-hue-1` on their `div` and get it to change color. 

Even though it might be more work to maintain, I think it might be cooler to use sass inheritance to make an `md-mini-fab` class that you could use instead of `md-fab`.

Thoughts?


ps- I also added some mini fab buttons to the button demo page in my PR... It might be a good idea to put some in there as well regardless of the implementation.

ThomasBurleson:
@robertmesserle - should I merge this now ?

robertmesserle:
Yes.

ThomasBurleson:
Closed with dde9ab7987c8df787ff72c3ce46b9247ffdf7aad

Issue #1042, opened 2014-12-23
mdIcon: Icons based on SVG, font-faces, and CSS 


ThomasBurleson:
A suite of icons should be support within Angular Material. Many components (BottomSheet, Lists, Cards, Chips, FABs, etc.) will use `md-icon` as a child element.

![screen shot 2014-12-23 at 1 24 39 pm](https://cloud.githubusercontent.com/assets/210413/5542046/00641582-8aa7-11e4-92b6-bae178fccaf0.png)
![toolbar_icons](https://cloud.githubusercontent.com/assets/210413/5541945/b332b274-8aa5-11e4-9d5e-ba1c4a35a47a.png)
![keyboard_icons](https://cloud.githubusercontent.com/assets/210413/5541946/b3361d1a-8aa5-11e4-9ad6-67a77e72feaa.png)

#### Icon Types

While Material Design icons come with SVG and CSS sprites for various resolutions, the full set of icons is enumerated below:

* SVG
* Animated SVG
* Color SVG
* Font Faces
* PNG
* JPEG

A live preview of all icons in the [Material-Icons](https://github.com/google/material-design-icons) repository is shown here:  [Live Preview](http://google.github.io/material-design-icons/)

![screen shot 2014-12-23 at 1 26 37 pm](https://cloud.githubusercontent.com/assets/210413/5542065/4bae1a7e-8aa7-11e4-8e2f-f84cfb978fd9.png)

<h2>Visual requirements:</h2>
 * Font-icons only
   *  class name used to specify desired icon; e.g. `<md-icon icon="refresh" />`
 * Hover effects to animate, colorize, etc.
 * Media support for responsive changes
 * Support for Dark themes

<h2>Functional requirements:</h2>
 * Ability to use remote SVG, PNG, or JPEG resources
 * Ability to use embedded (aka inlined) resources

<h2>Questions:</h2>
 * How can developers easily select a subset of icons for their needs?
 * How can developers easily embedded their own SVGs ?
   * Can we provide a template and a batch gulp tool to easily embedded a first-pass rendition of 1..n remote SVGs ?
 * How will the icons be deployed as part of Angular Material releases ?
 * Will Theming impact icon choices (colors) ?

### Additional Resources

*  [FlatIcon](http://www.flaticon.com/packs/material-design)
*  [SVG Animations and CSS Transforms](http://css-tricks.com/svg-animation-on-css-transforms/)
*  http://www.filamentgroup.com/lab/bulletproof_icon_fonts.html
[...]

PR #1288, opened 2015-01-25
feat(icon): svg icons with styles and simple img elements

justindujardin:
Update the md-icon directive to support CSS styling and loading of .svg file icons directly.  Add documentation, tests, and examples.

Questions:
- the directive currently defaults to render icons as `svg` so that styling may be applied to svg elements.  is it better to default to the `img` rendering, which would pollute the dom less but not svg element styling?
- sizing of the svg/img elements that are rendered seems to be unintuitive.  e.g. if you use CSS to size the md-icon element, the svg or img child elements won't always size as expected.   does it make sense to size the child elements to fit the parent md-icon container?
- what is the desired behavior when an icon url is invalid?  should this case be handled, and how?


marcysutton:
`mdIcon` is currently being rewritten, so this might all be irrelevant very soon: https://github.com/angular/material/issues/1042 

@ThomasBurleson can you provide some input?

Issue #1292, opened 2015-01-26
Prevent md-button for submit action


masseuro:
If a md-button is inside a form, its action will submit the form

To prevent this, add this in the template button :
```
type="button"
```
http://www.w3.org/TR/html-markup/button.html

masseuro:
http://jsfiddle.net/3ojbauoa/1/

yarl:
#1291 related

marcysutton:
This has already been discussed and was determined not to be an issue: https://github.com/angular/material/issues/710

PR #1293, opened 2015-01-26
add type button on button to prevent defaut submit action

masseuro:
This will fix this #1292

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -62,7 +62,7 @@ function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria) {
     if (isAnchor(attr)) {
       return '<a class="md-button" ng-transclude></a>';
     } else {
-      return '<button class="md-button" ng-transclude></button>';
+      return '<button class="md-button" type="button" ng-transclude></button>';
     }
   }
 


src/components/button/button.js
```diff
@@ -62,7 +62,7 @@ function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria) {
     if (isAnchor(attr)) {
       return '<a class="md-button" ng-transclude></a>';
     } else {
-      return '<button class="md-button" ng-transclude></button>';
+      return '<button class="md-button" type="button" ng-transclude></button>';
```
marcysutton: This won't work because it won't allow users to include submit buttons. A better approach would be to provide an attribute on `<md-button>` that will pass the desired type to the transcluded button element.

ajoslin:
We'll stay with the HTML default for buttons.

gkalpak:
@ajoslin: Why not provide a way to set the button type ? Having all buttons be "submit" buttons (as is the HTML default) is kind of limiting. @marcysutton's suggestion is reasonable.

gkalpak:
Hm...I just realized that the `type` attribute of `md-button` is passed to the created `button` element, so my previous comment is irrelevant (since there **is** a way to set the button type).

PR #1308, opened 2015-01-27
Fix(slider): cancel onPressUp when disabled

### LINKED ISSUE #1307: md-slider: responds to click events when disabled
Md-slider is calling onPressUp even when slider is disabled, causing it to update the UI and ng-model on user click. Tested on https://material.angularjs.org/#/demo/material.components.slider with Chrome 41.

nhayflick:
Closes #1307


ajoslin:
So the whole disabled checking was actually running on the wrong scope, since slider has an isolate scope.

Added unit tests for it and fixed the problem.  Thanks @nhayflick!

PR #1310, opened 2015-01-27
fix(themes): update theming specs for foreground-3 so tests pass

### LINKED ISSUE #1309: mdTheming: tests fail on foreground #3
Background-color opacity needs to be updated from 0.3 to 0.35 in Theming.spec.js line 201.

nhayflick:
Closes #1309


ajoslin:
@rschmukler reverted this because I made a mistake.

Issue #1281, opened 2015-01-24
layout-margin/layout-padding won't work with flex variants


gustavohenke:
When using `layout-margin` or `layout-padding`, child elements with any of the `flex` attribute variants (`flex-sm`, `flex-gt-sm`, etc) won't receive the margin/padding, respectively.

Example plunker: http://plnkr.co/edit/bRLK0Dk8JCXsMydFvrqv?p=preview

PR #1322, opened 2015-01-28
layout-margin/layout-padding won't work with flex variants #1281

MContagious:
Added different padding and margins for flex-sm, flex-md and flex-lg
To fix issue #1281 (layout-margin/layout-padding won't work with flex variants)

flex-lg = $layout-gutter-width
flex-md = $layout-gutter-width /2
flex-sm = $layout-gutter-width/3

### DIFF
--- a/src/components/sidenav/sidenav.scss
+++ b/src/components/sidenav/sidenav.scss
@@ -1,4 +1,5 @@
 $sidenav-default-width: 304px !default;
+$sidenav-min-space: 56px !default;
 
 md-sidenav {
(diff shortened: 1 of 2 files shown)


gustavohenke:
I don't think dividing the margin/padding is the correct way of handling it.
What if I have two rows of elements, one of which uses `[flex]` elements and the other uses `[flex-sm]`?

MContagious:
Yes, you are correct. We need to use the margin/padding for flex-sm, and other variants as flex.

PR #1324, opened 2015-01-28
feat(popover): add popover component and tests

stevenh77:
Online demo:  http://stevenhollidge.com/docs/#/demo/material.components.popover

Gif of demo:

demo

This pull request replaces the previous PR: https://github.com/angular/material/pull/1132

### DIFF
--- a/src/components/popover/demoBasicUsage/index.html
+++ b/src/components/popover/demoBasicUsage/index.html
@@ -1,0 +1,105 @@
+<section ng-controller="AppCtrl">
+
+    <md-toolbar class="md-accent">
+        <h1 class="md-toolbar-tools">
+            <span flex>Awesome Md App</span>
+        </h1>
+    </md-toolbar>
+    <md-content class="md-padding">
+
+        <p>
+            Popovers provide rich tooltip style content with directional indicators.  <br /> <br />Click the button to display the popover, clicking the button again to close.
+        </p>
+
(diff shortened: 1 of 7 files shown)


src/components/popover/demoBasicUsage/index.html
```diff
@@ -0,0 +1,105 @@
...
+
+                    <!-- rich content starts here -->
+                    <div style=" margin: 8px;">
+                        <div layout="row">
+                            <div>
+                                <h3 style="color: white; margin: 0 10px 10px 0;">Douglas Crockford</h3>
+                            </div>
+                            <div layout="row" layout-align="end start" style="width:100%">
+                                <div>
+                                    <a href="https://github.com/douglascrockford">
```
marcysutton: These links need text in them. It can be accomplished using [accessible SVG](http://www.sitepoint.com/tips-accessible-svg/) or an `aria-label` attribute.

marcysutton:
Great job with the accessibility support! 

One thing we discussed as a team is that this style of popover does not blend in with Material Design. For example: it doesn't resemble paper as a surface, the animation to show and hide isn't delightful and meaningful, and the visual style inside of the popover doesn't fit in with the theme. Further discussion is needed (and forthcoming), but I can point out the closest thing in the design spec, point-of-origin animation: http://www.google.com/design/spec/animation/responsive-interaction.html#responsive-interaction-material-response

PR #1334, opened 2015-01-29
docs(layout): subheaders over the code section

EladBezalel:
As you can see here
sample

I marked the subheader labels so it will be easy to see,
the subheader z-index is prior to the layout-panel-parent so i added z-index: 2


marcysutton:
:+1:

Issue #1340, opened 2015-01-29
mdDialog: user should not be able to interact outside of dialog


marcysutton:
So that keyboard and screen reader users cannot get lost while a dialog is open, focus should be limited to controls within the dialog (Cancel button, Accept button, etc.). Modal dialogs should not allow the user to interact with the page behind it.

Related: clicking on the background behind an alert or confirmation dialog should not do anything.

marcysutton:
This also might require us to remove the `parent` option for alert/confirm dialogs. Custom dialogs could still use it.

ThomasBurleson:
@marcysutton - the dialog mask currently traps the outside-dialog clicks right ?

marcysutton:
@ThomasBurleson by default, yes...clicking the background is the same action as clicking the "cancel" button. But any other links on the page shouldn't be accessible. 

The `parent` option is necessary to keep so that people can provide a different wrapper as a dialog target. I'm doing two things to fix our docs:

 1. Pass `parent: document.body` to our demos to deal with multiple `ng-apps` and make modals cover the entire screen.
 1. Write docs indicating that `parent` should only be used on nodes that cover the entire screen.

For accessibility, I still have to disable every node that isn't the dialog so it won't be accessible from a screen reader using `aria-hidden`. I have focus handled already in #1350.

yarl:
One more thing to add here: there should be no possibility to focus element behind dialog window using `[tab]` key. So basically, tab focus order should be looped inside dialog.

marcysutton:
Exactly–that's what I meant by interaction.

kburson:
Hmmm,   Now the default behavior is that nothing outside the dialog is clickable.  It used to be that clicking outside the dialog would 'hide' (close) the dialog, but that is no longer the case.  This was functionality I was depending on.  Was it a mistake to remove it?  

I understand wanting to keep the focus inside the dialog, but it is also nice to close the dialog by clicking on the mask outside... now I do not have that anymore.

Can you please put this back?

Also, are you planning on a separate popover service where I can open a dialog and keep it open while I interact with the backend? This is now a requirement from the designers on my project.  The 'modeless' dialog would display a list of data that would be referenced for data entry somewhere on the main page.  I thought about using 'toastr' or 'mdToast', and even simply rolling my own hidden/floating div.  Seems like a sort of functionality that should be included in the base framework doesn't it?

[...]

PR #1350, opened 2015-01-30
mdDialog: improved interaction and browser support

### LINKED ISSUE #1914: CRITICIAL BLOCKER: md-Select within md-Dialog
Szenario:

In the following demo http://plnkr.co/edit/Bs7vTflGrI7H9dkmk0z4?p=preview a list of entries is generated.
Each item has a dialog attached. Within the dialog a md-select item is provided.

Issue:

a) Scroll through the list and open a dialog.
b) Open the select box : The dialog gets re-positioned and drops of the visible area and can not be closed anymore
c) Press ESC which triggers to close the unseen selectbox. Now the Dialog appears again on the screen.
d) Select the md-select and the dialog stays on the screen

What I have observed is that the md-select component does not proper recognize the scrollTo position. This behaviour does not occur with the autocomplete component !

marcysutton:
Fixes for #1340 

Closes #1914

### DIFF
--- a/src/components/dialog/demoBasicUsage/script.js
+++ b/src/components/dialog/demoBasicUsage/script.js
@@ -4,18 +4,24 @@ angular.module('dialogDemo1', ['ngMaterial'])
   $scope.alert = '';
 
   $scope.showAlert = function(ev) {
+    // Appending dialog to document.body to cover sidenav in docs app
+    // Modal dialogs should fully cover application
+    // to prevent interaction outside of dialog
     $mdDialog.show(
       $mdDialog.alert()
+        .parent(angular.element(document.body))
         .title('This is an alert title')
         .content('You can specify some description text in here.')
-        .ariaLabel('Password notification')
+        .ariaLabel('Alert Dialog Demo')
         .ok('Got it!')
         .targetEvent(ev)
     );
   };
 
   $scope.showConfirm = function(ev) {
+    // Appending dialog to document.body to cover sidenav in docs app
     var confirm = $mdDialog.confirm()
+      .parent(angular.element(document.body))
       .title('Would you like to delete your debt?')
       .content('All of the banks have agreed to forgive you your debts.')
       .ariaLabel('Lucky day')
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -275,7 +275,8 @@ function MdDialogDirective($$rAF, $mdTheming) {
  *   - `targetEvent` - `{DOMClickEvent=}`: A click's event object. When passed in as an option,
  *     the location of the click will be used as the starting point for the opening animation
  *     of the the dialog.
- *   - `scope` - `{object=}`: the scope to link the template / controller to. If none is specified, it will create a new isolate scope.
+ *   - `scope` - `{object=}`: the scope to link the template / controller to. If none is specified,
+ *     it will create a new isolate scope.
  *     This scope will be destroyed when the dialog is removed unless `preserveScope` is set to true.
  *   - `preserveScope` - `{boolean=}`: whether to preserve the scope when the element is removed. Default is false
  *   - `disableParentScroll` - `{boolean=}`: Whether to disable scrolling while the dialog is open.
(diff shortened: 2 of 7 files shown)


src/components/dialog/dialog.js
```diff
@@ -272,8 +272,8 @@ function MdDialogProvider($$interimElementProvider) {
   function advancedDialogOptions($mdDialog) {
     return {
       template: [
-        '<md-dialog aria-label="{{ dialog.ariaLabel }}">',
-          '<md-content>',
+        '<md-dialog aria-label="{{ dialog.ariaLabel }}" tabIndex="-1">',
+          '<md-content role="document" tabIndex="0">',
```
gkalpak: Should this really be `document` ?

marcysutton: Yes. It creates a context for the message part of a dialog, within the `alertdialog` role. Multiple sources say so:
 * http://www.w3.org/TR/wai-aria-practices/#alert
 * https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_alertdialog_role

gkalpak: I thought it was applied to all dialogs. I didn't realize it was only applied to `alert` and `confirm` dialogs.
Sorry for the noise :)

src/components/dialog/dialog.js
```diff
@@ -336,6 +342,18 @@ function MdDialogProvider($$interimElementProvider) {
...
+          elementToFocus = closeButton;
+
+      if (options.$type === 'alert') {
+        role = 'alertdialog';
+        elementToFocus = element.find('md-content');
+      }
+
+      configureAria(element.find('md-dialog'), role);
+
+      document.addEventListener('focus', trapFocus, true);
```
marcysutton: The native event listener was required to enable the capture phase. Relevant source (see "Trapping focus"): http://www.nczonline.net/blog/2013/02/12/making-an-accessible-dialog-box/

ThomasBurleson:
@marcysutton  - can you update this PR with the rebase for current HEAD on master?

src/components/dialog/demoBasicUsage/script.js
```diff
@@ -4,24 +4,32 @@ angular.module('dialogDemo1', ['ngMaterial'])
   $scope.alert = '';
 
   $scope.showAlert = function(ev) {
-    $mdDialog.show(
-      $mdDialog.alert()
```
ThomasBurleson: @marcysutton - why did you replace the 'API-chaining' approach with the 'configuration object' approach?

src/components/dialog/dialog.js
```diff
@@ -384,7 +384,7 @@ function MdDialogProvider($$interimElementProvider) {
       isolateScope: true,
       onShow: onShow,
       onRemove: onRemove,
-      clickOutsideToClose: true,
+      clickOutsideToClose: false,
```
ThomasBurleson: @marcysutton - Why did you disable the `clickOutsideToClose` flag?

src/components/dialog/dialog.js
```diff
@@ -393,6 +393,14 @@ function MdDialogProvider($$interimElementProvider) {
       }
     };
 
+    function trapFocus(ev) {
+      var dialog = document.querySelector('md-dialog');
+
+      if (dialog && !dialog.contains(ev.target)) {
```
ThomasBurleson: nice

src/components/dialog/dialog.js
```diff
@@ -393,6 +393,14 @@ function MdDialogProvider($$interimElementProvider) {
...
+        ev.stopPropagation();
```
ThomasBurleson: Do we not want to call `stopImmediatePropagation()` ?

src/components/dialog/demoBasicUsage/script.js
marcysutton: I needed to supply a `parent` element. Dialogs must cover the whole screen, and because our docs have multiple `ng-app`s, they couldn't do that without an override. Can you supply a parent with the API-chain approach?

src/components/dialog/dialog.js
marcysutton: Dialogs shouldn't be that easy to dismiss–users are supposed to use the buttons inside of a modal dialog to either cancel or proceed. Although, this change could cause a negative impact on our existing users, so we could change it back. It's also arguable that clicking outside to close is the same as hitting cancel. I just figured by defaulting `clickOutsideToClose` to false, it would make our dialogs behave more like desktop ones but still allow the option to override.

src/components/dialog/dialog.js
marcysutton: I didn't think it needed to stop other listeners on the same element, just prevent the listener triggering this function from bubbling/capturing.

src/components/dialog/dialog.js
ThomasBurleson: I see. Agreed.

src/components/dialog/demoBasicUsage/script.js
ThomasBurleson: I will check. Meanwhile we should show a both approaches... I will do that with the merge.

src/components/dialog/dialog.js
ThomasBurleson: If you are resetting focus, I think we should stop **immediately**.

src/components/dialog/dialog.js
ThomasBurleson: :+1:

src/components/dialog/demoBasicUsage/script.js
marcysutton: Scratch that, figured it out. :)

src/components/dialog/demoBasicUsage/script.js
```diff
@@ -4,18 +4,26 @@ angular.module('dialogDemo1', ['ngMaterial'])
   $scope.alert = '';
 
   $scope.showAlert = function(ev) {
+    // Appending dialog to document.body to cover sidenav in docs app
+    // Modal dialogs should fully cover application
+    // to prevent interaction outside of dialog
     $mdDialog.show(
-      $mdDialog.alert()
+      $mdDialog.alert({
+          parent: angular.element(document.body)
```
ThomasBurleson: @marcysutton - I just added support for `.parent(angular.element(document.body))` in master.
Also I think you should be using ` $rootElement.find('body') || $rootElement` instead of `document.body`

marcysutton:
Ok @ThomasBurleson, I think this one is ready. It fixes quite a few things.

robertmesserle:
:+1:

marcysutton:
Closed with 0b0ed2339e1ec32e34d20daa18cadd4ef89d5f86.

PR #1367, opened 2015-02-02
fix(gesture): Implement proper event dispatching when using JQuery

### LINKED ISSUE #1359: mdGestures breaks Ripple effect when using JQuery
Only when using JQuery, ripple effect is broken. It only occurs using `angular-material>=0.7.1`.

[plunkr with JQuery](http://plnkr.co/edit/ZWhlHe9rzwP7eFzkObPd)
[plunkr without JQuery](http://plnkr.co/edit/4fOuM3UfFgdcN5PqOmhb)

Both jquery 2.x and 1.x cause the issue.

Toilal:
Close #1359

For [on](http://api.jquery.com/on/) and [off](http://api.jquery.com/on/), jqLite use native events, but JQuery doesn't . So to trigger events properly when JQuery is loaded, the JQuery [trigger](http://api.jquery.com/trigger/) method must be used instead of native [dispatchEvent](https://developer.mozilla.org/fr/docs/DOM/element.dispatchEvent) method.

### DIFF
--- a/src/core/services/gesture/gesture.js
+++ b/src/core/services/gesture/gesture.js
@@ -302,7 +302,7 @@ angular.module('material.core')
     onCancel: angular.noop,
     options: {},
 
-    dispatchEvent: dispatchEvent,
+    dispatchEvent: typeof jQuery !== 'undefined' && angular.element === jQuery ? jQueryDispatchEvent : nativeDispatchEvent,
 
     start: function(ev, pointer) {
       if (this.state.isRunning) return;
@@ -361,10 +361,11 @@ angular.module('material.core')
     bubbles: true,
     cancelable: true
   };
+
   /*
-   * NOTE: dispatchEvent is very performance sensitive. 
+   * NOTE: native and jquery event dispatchers are very performance sensitive.
    */
-  function dispatchEvent(srcEvent, eventType, eventPointer, /*original DOMEvent */ev) {
+  function nativeDispatchEvent(srcEvent, eventType, eventPointer, /*original DOMEvent */ev) {
     eventPointer = eventPointer || pointer;
     var eventObj;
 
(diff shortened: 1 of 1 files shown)


src/core/services/gesture/gesture.js
```diff
@@ -361,10 +361,11 @@ angular.module('material.core')
     bubbles: true,
     cancelable: true
   };
+
   /*
    * NOTE: dispatchEvent is very performance sensitive.
```
Splaktar: The name here probably needs to be changed as well
OR 
I'm not sure that the name needs to specifically contain jqLite or if it would be enough to just add that to the comment here.

src/core/services/gesture/gesture.js
```diff
@@ -302,7 +302,7 @@ angular.module('material.core')
     onCancel: angular.noop,
     options: {},
 
-    dispatchEvent: dispatchEvent,
+    dispatchEvent: angular.element === window.jQuery ? jQueryDispatchEvent : jqLiteDispatchEvent,
```
ajoslin: We shouldn't check for jQuery on `window`, this won't work in all situations.

This would work better:

```js
dispatchEvent: typeof jQuery !== 'undefined' && angular.element === jQuery ? 
  jQueryDispatchEvent : 
  jqLiteDispatchEvent,
```

Toilal: Updated

Splaktar:
LGTM! Thanks!

src/core/services/gesture/gesture.js
```diff
@@ -302,7 +302,7 @@ angular.module('material.core')
     onCancel: angular.noop,
     options: {},
 
-    dispatchEvent: dispatchEvent,
+    dispatchEvent: typeof jQuery !== 'undefined' && angular.element === jQuery ? jQueryDispatchEvent : nativeDispatchEvent,
```
gkalpak: Beware that this will probably break once https://github.com/angular/angular.js/pull/10761 lands (and I believe it's pretty close to landing).

Toilal: Maybe we would need to check if jQuery object is really jQuery, but it should work even after merge from what I can read.

ajoslin:
Merged, with a small addition to be sure jQuery users get metaKey/ctrlKey/ etc: https://github.com/angular/material/commit/88813b78cae831b28e2e8d0eb37c32269034bfad#diff-a82fbe888db156ef2e09e7fcfb85f189R370

Thanks:

src/core/services/gesture/gesture.js
gkalpak: It has been merged, btw. The problem is that `typeof jQuery !== 'undefined' && angular.element === jQuery` might evaluate to false (thus using `nativeDispatchEvent` although jQuery **is** used (leading to the wrong behaviour again)

Toilal: You are right, jQuery may be defined as another name in global namespace, and then properly register using the new `ngJq` directive.

Maybe we should only check if `angular.element` has `trigger` function ? jqLite doesn't have this function.

https://github.com/angular/angular.js/blob/master/src/Angular.js#L1532-L1536

ThomasBurleson: @rschmukler - Can you follow-up on this issue?

PR #1372, opened 2015-02-02
fix(tabs): Fix tabs swipping when using JQuery

Toilal:
When using dots in JQuery events name, an [event namespace](http://api.jquery.com/event.namespace/) is created and Event.type doesn't give the same name as using jqLite. JQuery removes the namespace from Event.type, but JQlite doesn't support namespace and keep the whole dotted value.

This cause an issue in tabs swipping, because onSwipe listener is checking Event.type property value.

For consistency with jqLite and jQuery, it's better to avoid checking the Event.type value, and use two distinct listeners for right and left swipping.

[Review on Reviewable](https://reviewable.io/reviews/angular/material/1372)

### DIFF
--- a/src/components/tabs/js/tabItemController.js
+++ b/src/components/tabs/js/tabItemController.js
@@ -65,7 +65,8 @@ function TabItemController($scope, $element, $attrs, $compile, $animate, $mdUtil
         'aria-selected': true,
         'tabIndex': 0
       })
-      .on('$md.swipeleft $md.swiperight', onSwipe);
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@robertmesserle - do we still support content swiping in Tabs ?

ThomasBurleson:
No longer valid since tab re-architectures; @see https://github.com/angular/material/blob/master/src/components/tabs/js/tabsDirective.js#L178-L179

Issue #1370, opened 2015-02-02
gulp karma releaseMode option has no effect


Toilal:
I've noticed that gulp karma task should execute tests against minified files, but it doesn't work because `releaseMode` parameter is always undefined in [karma.conf.js](https://github.com/angular/material/blob/master/config/karma.conf.js#L26).

`releaseMode` is set properly in [gulpfile.js](https://github.com/angular/material/blob/master/gulpfile.js#L158), but is not passed to karma.conf.js

It could be fixed using an [environment variable](http://stackoverflow.com/questions/20034256/big-js-app-testing-avoiding-multiple-karma-conf-js-files) instead.

ThomasBurleson:
Fixed in SHA 532e1533cdb93526bbe08ab62fb5fd3d40bff2a8

PR #1373, opened 2015-02-02
chore(gulp): Run gulp karma with minified build and JQuery dependency

Toilal:
This fix an issue (#1370) causing minified build test to run against non minified sources, and adds a 3rd test configuration with JQuery dependency.

Others PRs (#1367, #1371, #1372) are there to make the tests to pass with JQuery.

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -14,6 +14,7 @@
     "angular-route": "1.3.x",
     "angular-mocks": "1.3.x",
     "angularytics": "^0.3.0",
-    "angular-messages": "~1.3.8"
+    "angular-messages": "~1.3.8",
+    "jquery": "~2.1.x"
   }
 }
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -155,8 +155,22 @@ gulp.task('karma', function(done) {
   function testMinified() {
     gutil.log('Running unit tests on minified source.');
     buildJs(true);
-    karmaConfig.releaseMode = true;
-    karma.start(karmaConfig, done);
+    process.env['MATERIAL_RELEASE_MODE'] = 'true';
+    karma.start(karmaConfig, testMinifiedJQuery);
+  }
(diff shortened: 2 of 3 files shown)


bower.json
```diff
@@ -14,6 +14,7 @@
     "angular-route": "1.3.x",
     "angular-mocks": "1.3.x",
     "angularytics": "^0.3.0",
-    "angular-messages": "~1.3.8"
+    "angular-messages": "~1.3.8",
+    "jquery": "~2.1.x"
```
ThomasBurleson: 1) Since the `/material/bower.json` dependencies are copied to `/bower-material/bower.json`, we cannot include the jQuery as a dependency here; since we only want to install jQuery for `/angular/material/` testing purposes.

2) Similar to (1), we do not have a dependency upon angular-messages at deploy time; only for testing purposes.\

> Perhaps we need a solution as discussed for [Bower-Ignore-Dependencies](https://github.com/bower/bower/pull/1394)

gulpfile.js
```diff
@@ -155,8 +155,22 @@ gulp.task('karma', function(done) {
...
     buildJs(true);
-    karmaConfig.releaseMode = true;
-    karma.start(karmaConfig, done);
+    process.env['MATERIAL_RELEASE_MODE'] = 'true';
+    karma.start(karmaConfig, testMinifiedJQuery);
+  }
+
+  function testMinifiedJQuery() {
+    gutil.log('Running unit tests on minified source using JQuery.');
+    buildJs(true);
```
rschmukler: This second `buildJs` is probably unnecessary since the previous `testMinified()` already built it.

rschmukler:
Merged in a slightly modified version with 532e1533cdb93526bbe08ab62fb5fd3d40bff2a8. Thanks!

PR #1386, opened 2015-02-04
Update toast.scss

ChristopheBoucaut:
I propose replace `absolute` with `fixed`, because the toast doesn't move when user scroll the page and it finds himself in the middle of the page (it's ugly).

Currently => 
Start: http://img4.hostingpics.net/pics/767337good.png
Scroll: http://img4.hostingpics.net/pics/503064bad.png

With fixed =>
Scroll : http://img4.hostingpics.net/pics/992306withfixed.png

[Review on Reviewable](https://reviewable.io/reviews/angular/material/1386)

### DIFF
--- a/src/components/toast/toast.scss
+++ b/src/components/toast/toast.scss
@@ -1,7 +1,7 @@
 // See height set globally, depended on by buttons
 md-toast {
   display: flex;
-  position:absolute;
+  position:fixed;
   box-sizing: border-box;
   align-items: center;
 


ChristopheBoucaut:
i signed

ilanbiala:
Has to do with https://github.com/angular/material/issues/1106.

ChristopheBoucaut:
Okey :) thanks.

ilanbiala:
It doesn't need to be closed, I'm just referencing the issue that it would fix.

ThomasBurleson:
@rschmukler - I think this change must be rejected; as we do not want `position:fixed`. Do you agree ?

birdcw:
If not fixed, could you just set the toast to close on a swipe/scroll event? To me, it seems like a reasonable assumption that the user is dismissing the toast when they scroll

rschmukler:
Closing this issue - position fixed would violate [the spec](https://www.google.com/design/spec/components/snackbars-toasts.html#snackbars-toasts-usage) on blocking FABs. Ultimately we need a better solution that resolves both. @birdcw please open a new issue if you still think we should have the close on scroll

PR #1391, opened 2015-02-04
fix(slider): adds padding inside of slider to fit the thumb

### LINKED ISSUE #1343: md-slider width increases when slide to right
<b>md-slider</b> width increases when sliding

and this <b>problem</b> is started from the <b>version 0.7.1-rc1</b>

check problem - <b><a href="http://plnkr.co/edit/hDPPg1UQZk7veyvPvNbJ?p=preview">http://plnkr.co/edit/hDPPg1UQZk7veyvPvNbJ?p=preview </a></b>

Same behavior is seen on angular material slider demo page.

robertmesserle:
Closes #1343

### DIFF
--- a/src/components/slider/slider.js
+++ b/src/components/slider/slider.js
@@ -53,20 +53,22 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
     scope: {},
     require: '?ngModel',
     template:
-      '<div class="md-track-container">' +
-      '<div class="md-track"></div>' +
-      '<div class="md-track md-track-fill"></div>' +
-      '<div class="md-track-ticks"></div>' +
-      '</div>' +
-      '<div class="md-thumb-container">' +
-      '<div class="md-thumb"></div>' +
-      '<div class="md-focus-thumb"></div>' +
-      '<div class="md-focus-ring"></div>' +
-      '<div class="md-sign">' +
-      '<span class="md-thumb-text"></span>' +
-      '</div>' +
-      '<div class="md-disabled-thumb"></div>' +
-      '</div>',
+      '<div class="md-slider-wrapper">\
+        <div class="md-track-container">\
+          <div class="md-track"></div>\
+          <div class="md-track md-track-fill"></div>\
+          <div class="md-track-ticks"></div>\
+        </div>\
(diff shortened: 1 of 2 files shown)


robertmesserle:
@ThomasBurleson Please note that this will cause styles to change for some users.  Basically, in order for this to work, I had to add padding inside the actual slider wrapper, which causes the slider to shrink.  If this is not how we expect it to work, then comment here and we can discuss alternative solutions.

src/components/slider/slider.js
```diff
@@ -85,9 +87,10 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
     };
 
     var isDisabledParsed = attr.ngDisabled && $parse(attr.ngDisabled);
-    var isDisabledGetter = isDisabledParsed ? 
+    var isDisabledGetter = isDisabledParsed ?
       function() { return isDisabledParsed(scope.$parent); } :
       angular.noop;
+    var wrapper = angular.element(element[0].querySelector('.md-slider-wrapper'));
```
ThomasBurleson: Why are you creating a unused `wrapper` variable?

ThomasBurleson:
@robertmesserle - This effectively creates a left/right 24px padding around md-slider so the thumb shows properly at the start/end ranges. 

*  What about the ripple/focus indicators are those clipped? 
*  Have you tested with hsizing of the md-slider... does that still work ?

src/components/slider/slider.js
robertmesserle: Good question - I used it at one point while fixing this, but apparently, didn't end up needing it.  Will remove.

robertmesserle:
@ThomasBurleson The 24px padding is the size needed for the focus/held state.

Can you point me to an example of what you mean by hsizing of the slider?

sunnyshah01:
@robertmesserle This is not the right fix. size of slider became smaller because of padding.

robertmesserle:
@sunnyshah01 @ThomasBurleson Updated this PR to include a safer method that should not impact the size at all.

ThomasBurleson:
@robertmesserle - `curl https://github.com/angular/material/pull/1391.patch | git am` failed.
Can you update your PR with a `git pull --rebase origin master` first? Thx.

PR #1392, opened 2015-02-04
feat(input): adds `no-float` class to disable floating label animations

### LINKED ISSUE #201: Text field with fixed label; no animation when the input field on focus
In fact, what I want is like this:
material text field

It seems that I can not use `<label>` tag because of the animation. I can just work around with other tag like `<p>`, `<div>`.

```html
<material-input-group class="material-input-group-theme-light" layout="horizontal">
    <p flex="30">123123</p>
    <material-input type="text" 
                    ng-model="user.fullName"
                    flex="70"
                    style="display: inline-block;"
                    placeholder="please input something"></material-input>
</material-input-group>
```

For the semantic of HTML, I think it will be better if we support a way to remove the label animation.

robertmesserle:
Closes #201

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -25,7 +25,7 @@ angular.module('material.components.input', [
  * @description
  * `<md-input-container>` is the parent of any input or textarea element.
  *
- * Input and textarea elements will not behave properly unless the md-input-container 
+ * Input and textarea elements will not behave properly unless the md-input-container
  * parent is provided.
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@robertmesserle - you need to update this PR with lastest rebase from master HEAD. Thx

breandr:
Any chance of getting this for md-autocomplete and others?

PR #1407, opened 2015-02-05
Copy update to ngDoc for $mdDialog#show

johnhok:
The return value mentions a promise that can be resolved with $mdDialog.hide and mdDialog.show.. tweaked the copy to read $mdDialog.show instead.


johnhok:
Signed the CLA.

PR #1411, opened 2015-02-06
docs(mdDialog): document obj setup, fix headings

### LINKED ISSUE #1383: mdDialog docs missing object setup documentation alongside chainable methods
Why do we have to chain function calls on dialogs to set their parameters? As in:

```javascript
      .title('Would you like to delete your debt?')
      .content('All of the banks have agreed to forgive you your debts.')
      .ariaLabel('Lucky day')
      .ok('Please do it!')
      .cancel('Sounds like a scam')
      .targetEvent(ev);
```

 I feel like it would be much cleaner to just pass an object with all the parameters to $mdDialog, as is done with most API's in javascript. As in:

```javascript
$mdDialog.confirm({
	title: 'Would you like to delete your debt?',
	content: 'All of the banks have agreed to forgive you your debts.',
	ariaLabel: 'Lucky day',
	ok: 'Please do it!',
	cancel: 'Sounds like a scam',
	targetEvent: ev
}).then(function() {
    $scope.alert = 'You decided to get rid of your debt.';
}, function() {
    $scope.alert = 'You decided to keep your debt.';
});
```

 I really think the current way is counterintuitive. Can anyone explain to me why this is the way it is? Or maybe if it can be changed?

marcysutton:
Closes #1383, #1413

### DIFF
--- a/docs/config/template/ngdoc/api/directive.template.html
+++ b/docs/config/template/ngdoc/api/directive.template.html
@@ -5,7 +5,7 @@
 
   {% block usage %}
   <div class="usage">
-    <h3 id="Usage">Usage</h3>
+    <h2 id="Usage">Usage</h2>
   {% if doc.usage %}
     {$ doc.usage | marked $}
   {% else %}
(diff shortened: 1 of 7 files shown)


ThomasBurleson:
Closed with SHA c6d99a88e49c0abbc

PR #1414, opened 2015-02-06
chore: change "//" to "https://" to encourage https.

jelbourn:


IgorMinar:
looks right to me. but I don't know if there could by any other places, so if you feel like you covered everything then LGTM :)

PR #1418, opened 2015-02-06
feat(autocomplete): adds autocomplete text-field

robertmesserle:


robertmesserle:
@marcysutton Can you please confirm that I did not break anything with my re-organization of the aria code?

ThomasBurleson:
Oh my... cannot wait to review this code. #woot

robertmesserle:
I maybe jumped the gun on doing a PR - there are no unit tests yet - but I wanted to get this up for review early.

src/core/style/structure.scss
```diff
@@ -148,6 +148,18 @@ input {
   }
 }
 
+.visuallyhidden {
```
ilanbiala: Shouldn't this use a dash to be consistent?

marcysutton: This class is a common utility from HTML5 Boilerplate, so by not using the hyphen it will be easier to remember.

ThomasBurleson: perhaps something like `.hide-visually` ?

src/components/autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,161 @@
...
+      overflow: auto;
+      max-height: 41px * 5.5;
+      li {
+        border-top: 1px solid #ddd;
+        padding: 0 15px;
+        font: 400 14px/40px Roboto, sans-serif;
+        overflow: hidden;
+        height: 40px;
+        transition: background 0.15s linear;
+        cursor: pointer;
```
marcysutton: This (and `:hover`) would normally be cause for concern, since `li` is not reachable from the keyboard by default. But because the arrow keys allow you to reach list items from the input, it works in this case.

src/core/style/structure.scss
robertmesserle: Updated (and also removed redundant class in doc styles)

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -0,0 +1,145 @@
...
+        if (searchText !== $scope.searchText) return; //-- just cache the results if old request
+        promise = null;
+        self.loading = false;
+        self.matches = matches;
+      });
+    }
+
+    function keydown (event) {
+      switch (event.keyCode) {
+        case 40: //-- DOWN ARROW
```
marcysutton: These keyCodes are all stored in `$mdConstant`. You could use `$mdConstant.KEYCODE.DOWN_ARROW`.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -0,0 +1,145 @@
...
+    //-- return init
+    return init();
+
+    //-- start method definitions
+    function init () {
+      configureWatchers();
+      attachAriaAttributes();
+    }
+
+    function attachAriaAttributes () {
```
marcysutton: Should this function be named `configureAria` for consistency?

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -0,0 +1,145 @@
...
+      .module('material.components.autocomplete')
+      .controller('MdAutocompleteCtrl', MdAutocompleteCtrl);
+
+  function MdAutocompleteCtrl ($scope, $element, $timeout, $q, $mdUtil) {
+
+    //-- private variables
+    var self = this,
+        itemParts = $scope.itemsExpr.split(/\ in\ /i),
+        itemExpr = itemParts[1],
+        elements = {
```
marcysutton: :+1:  love this

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -0,0 +1,58 @@
...
+              aria-label="{{placeholder}}"\
+              aria-autocomplete="list"\
+              aria-haspopup="true"\
+              aria-activedescendant=""\
+              aria-expanded="{{!$mdAutocompleteCtrl.hidden}}"\
+              ng-keydown="$mdAutocompleteCtrl.onKeydown($event)"/>\
+          <button\
+              type="button"\
+              ng-if="searchText"\
+              ng-click="$mdAutocompleteCtrl.clear()">\
```
marcysutton: When this button has focus and you clear the field, focus is lost. We should handle it by sending focus back to the input.

src/components/autocomplete/html/autocomplete.tmpl.html
```diff
@@ -0,0 +1,12 @@
+<label>
```
marcysutton: What is this template used for? It's missing some of the accessibility stuff from the template in `autocompleteDirective.js`.

Also: I'm still not quite sure why this is wrapped in a label element, since there is no text in it to act as a label.

robertmesserle: Ah, this was ditched, but made its way into the commit.  Removed it.

src/components/autocomplete/autocomplete.scss
robertmesserle: Ah, good point.  If you think that some alternative markup would be more fitting, I'd be happy to change it.

src/components/autocomplete/js/autocompleteController.js
robertmesserle: Sure, I'll rename this to match `dialog.js`

src/components/autocomplete/js/autocompleteController.js
robertmesserle: Ah, right - this was leftover from my initial draft done on CodePen. =p

src/components/autocomplete/autocomplete.scss
marcysutton: I think it's fine, I just wanted to point it out for future components.

src/components/autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,161 @@
...
+      border: none;
+      box-shadow: none;
+      padding: 0 15px;
+      font: 400 14px/40px Roboto, sans-serif;
+      height: 40px;
+      outline: none;
+      z-index: 2;
+      background: transparent;
+    }
+    button {
```
marcysutton: This button could use a focus style. Icon buttons in toolbars get an ink circle, which might work but we'd have to do some tricky positioning stuff to not affect the layout of the directive.

Toolbar icon focus

robertmesserle: Agreed.  Right now, this button is somewhat placeholder - I was trying to wait until our icons were ready to do the PR because ideally, there should be an icon for this, but I wanted to get some eyes on it early because I anticipated having to make some more changes based on feedback.

src/components/autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,161 @@
...
+    opacity: 0;
+    height: 0;
+    animation-timing-function: linear;
+  }
+  50% {
+    opacity: 1;
+    height: 40px;
+    animation-timing-function: ease-out;
+  }
+  100% {
```
gkalpak: (Sorry if it is obvious, but) what is the purpose of these keyframes having identical values for 50% and 100% ?

src/components/autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,161 @@
...
+    opacity: 1;
+    height: 40px;
+  }
+}
+body {
```
gkalpak: Is styling `body` really necessary in autocomplete.scss ?

src/components/autocomplete/demoBasicUsage/script.js
```diff
@@ -0,0 +1,33 @@
...
+    var deferred = $q.defer();
+    var regex = new RegExp('^' + query.replace(/[\*\[\]\(\)\{}\\\^\$]/g, function (match) {
+          return '\\' + match;
+        }), 'i'),
+        results = [];
+    angular.forEach(self.states, function (state) {
+      if (state.value.match(regex)) {
+        results.push(state);
+      }
+    });
```
gkalpak: Not that it matters much (since it is just a demo), but I believe that lines 21 - 29 could be simplified as:

```js
var lowercaseQuery = angular.lowercase(query);
var results = self.states.filter(function (state) { return state.value.indexOf(lowercaseQuery) === 0; }); 
```

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -0,0 +1,145 @@
...
+      $scope.searchText = '';
+      select(-1);
+    }
+
+    function isHidden () {
+      return self.matches.length === 1 && $scope.searchText === getValue(self.matches[0]);
+    }
+
+    function getValue (item) {
+      return item ? $scope.itemText ? item[$scope.itemText] : item : item;
```
gkalpak: This looks intriguing, but `(item && $scope.itemText) ? item[$scope.itemText] : item` would be more simple :smiley:

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -0,0 +1,145 @@
...
+    function clearValue () {
+      $scope.searchText = '';
+      select(-1);
+    }
+
+    function isHidden () {
+      return self.matches.length === 1 && $scope.searchText === getValue(self.matches[0]);
+    }
+
+    function getValue (item) {
```
gkalpak: `getDisplayValue()` or something similar would be more discriptive.

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -0,0 +1,58 @@
...
+          <input type="text"\
+              ng-model="searchText"\
+              ng-keydown="$mdAutocompleteCtrl.keydown($event)"\
+              placeholder="{{placeholder}}"\
+              aria-label="{{placeholder}}"\
+              aria-autocomplete="list"\
+              aria-haspopup="true"\
+              aria-activedescendant=""\
+              aria-expanded="{{!$mdAutocompleteCtrl.hidden}}"\
+              ng-keydown="$mdAutocompleteCtrl.onKeydown($event)"/>\
```
gkalpak: This line should be removed.

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -0,0 +1,58 @@
...
+              type="button"\
+              ng-if="searchText"\
+              ng-click="$mdAutocompleteCtrl.clear()">\
+              <span aria-hidden="true">X</span>\
+              <span class="visually-hidden">Clear</span>\
+              </button>\
+          <md-progress-linear ng-if="$mdAutocompleteCtrl.loading" md-mode="indeterminate"></md-progress-linear>\
+        </md-autocomplete-wrap>\
+        <ul role="presentation">\
+          <li ng-repeat="(index, item) in $mdAutocompleteCtrl.matches"\
```
gkalpak: You could also drop `index` and use `ngRepeat`'s `$index` instead.

src/components/autocomplete/js/highlightController.js
```diff
@@ -0,0 +1,25 @@
...
+          html = text.replace(regex, function (match) {
+            return '<span class="highlight">' + match + '</span>';
+          });
+      $element.html(html);
+    });
+
+    function sanitize (term) {
+      if (!term) return term;
+      return term
+          .replace(/[\*\[\]\(\)\{\}\\\^\$]/g, function (match) { return '\\' + match; });
```
gkalpak: The function could be replaced by `'\\$&'`

src/components/autocomplete/js/highlightDirective.js
```diff
@@ -0,0 +1,14 @@
...
+  'use strict';
+  angular
+      .module('material.components.autocomplete')
+      .directive('mdHighlightText', MdHighlight);
+
+  function MdHighlight () {
+    return {
+      terminal: true,
+      controller: 'MdHighlightCtrl',
+      controllerAs: 'mdCtrl'
```
gkalpak: Since `MdHighlightCtrl` does not expose any public API and since there is no template, what is the purpose of `controllerAs` ? If it is not used then it just creates an unneeded object reference on the scope.

src/components/autocomplete/js/highlightDirective.js
```diff
@@ -0,0 +1,14 @@
+(function () {
+  'use strict';
+  angular
+      .module('material.components.autocomplete')
+      .directive('mdHighlightText', MdHighlight);
+
+  function MdHighlight () {
+    return {
+      terminal: true,
```
gkalpak: Why terminal ?

src/components/autocomplete/js/highlightController.js
```diff
@@ -0,0 +1,25 @@
+(function () {
+  'use strict';
+  angular
+      .module('material.components.autocomplete')
+      .controller('MdHighlightCtrl', MdHighlightCtrl);
+
+  function MdHighlightCtrl ($scope, $element, $interpolate) {
+    var term = $element.attr('md-highlight-text'),
+        text = $interpolate($element.text())($scope.$parent);
```
gkalpak: Since this does not create a new scope, why use `$scope.$parent` ? Wouldn't it reference the wrong scope (taking into account that the element will be transcluded) ?

gkalpak: Hm...and why interpolate the text manually ?

src/components/autocomplete/autocomplete.scss
robertmesserle: Actually, that was a bug - opacity was supposed to be 0 at 50%.  Good catch!

src/components/autocomplete/autocomplete.scss
robertmesserle: Nope, removed.

src/components/autocomplete/js/autocompleteDirective.js
robertmesserle: This is more personal preference than anything, but I like to avoid using variables that I don't explicitly name.  It's not super important in this case, but I still find it cleaner to specify my own variables.

src/components/autocomplete/js/highlightController.js
robertmesserle: To be honest, I'm not 100% sure where the extra scope is coming from, but `$scope.$parent` is absolutely necessary.  As for interpolating manually, this is because I am modifying the HTML manually, and allowing Angular to control it resulted in issues.

ThomasBurleson:
@robertmesserle  - is this ready to merge ?

robertmesserle:
I have to add documentation and unit tests - should have that in tomorrow.

robertmesserle:
@ThomasBurleson Ready to merge.

src/components/autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,155 @@
...
+  50% {
+    opacity: 0;
+    height: 40px;
+  }
+  100% {
+    opacity: 1;
+    height: 40px;
+  }
+}
+md-content {
```
marcysutton: Should this be global?

src/components/autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,155 @@
...
+    li {
+      border-top: 1px solid #ddd;
+      padding: 0 15px;
+      font: 400 14px/40px Roboto, sans-serif;
+      overflow: hidden;
+      height: 40px;
+      transition: background 0.15s linear;
+      cursor: pointer;
+      margin: 0;
+      color: #212121;
```
marcysutton: These colors should be in variables.

src/components/autocomplete/autocomplete.scss
```diff
@@ -0,0 +1,155 @@
...
+  input {
+    position: absolute;
+    left: 0;
+    top: 0;
+    width: 100%;
+    box-sizing: border-box;
+    border: none;
+    box-shadow: none;
+    padding: 0 15px;
+    font: 400 14px/40px Roboto, sans-serif;
```
marcysutton: Do you need to set the font face here? Shouldn't it cascade?

src/components/autocomplete/js/listItemDirective.js
```diff
@@ -0,0 +1,22 @@
+(function () {
+  'use strict';
+  angular
+      .module('material.components.autocomplete')
+      .directive('mdListItem', MdListItem);
+
+  function MdListItem ($compile, $mdUtil) {
+    return {
+      require: '^mdAutocomplete',
```
marcysutton: Does this share any functionality with the `mdItem` which already exists as a child of `mdList`? We talked about renaming `mdItem` to `mdListItem` for consistency. Would that cause a conflict with this directive?

marcysutton:
@robertmesserle sorry for the late feedback. Nothing major, just a few little things!

src/components/autocomplete/autocomplete.scss
robertmesserle: Renamed in latest commit

robertmesserle:
@marcysutton No worries, addressed your feedback and update PR

PR #1447, opened 2015-02-10
fix(mdInput): CSS cascades from disabled fieldsets

### LINKED ISSUE #895: Disabled fieldset won't apply disabled styles to child elements
If using a disabled fieldset, styles won't be applied in its child elements.

Current results:

Note that the button is raised.
I expected the button to look like [the disabled one from the Angular Material demo](https://material.angularjs.org/#/demo/material.components.button).

I think inputs don't need any special styles when they're disabled, except for `cursor: default` (which already is applied).

I'm using Angular Material v0.6.1.

marcysutton:
Closes #895

### DIFF
--- a/src/components/input/input-theme.scss
+++ b/src/components/input/input-theme.scss
@@ -57,11 +57,13 @@ md-input-container.md-THEME_NAME-theme {
       color: '{{warn-500}}';
     }
   }
-
-  .md-input[disabled] {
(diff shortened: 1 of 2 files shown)


marcysutton:
Merged with 66fa1e3e276ddf33d669da0ef4d13c73c668e654.

PR #1465, opened 2015-02-11
update(icon): added unit test to ensure error when icon or set not found

programmist:


programmist:
Not sure how this got closed, but it won't allow me to reopen.

PR #1466, opened 2015-02-11
Use mdIcon instead of data-URIs

### LINKED ISSUE #1464: mdTabs pagination arrows have dissapeared
Just as the title says, with the latest commit (bbbec18e5a) the ui changed, and actually uses the accent color now, but the pagination arrows have disappeared, which is a very bad thing.

mdtabs issue

As you can see in the picture, there are no pagination arrows, which on smaller screens, leads the user to think there are only 5 tabs, when in fact there are 10. The buttons are still there, and can be clicked, but there is no icon to indicate that. Occurs on Firefox 35.0.1, Chrome 41 beta, and Chrome Mobile version 40, which makes me think it isn't a browser thing.

marcysutton:
A few places in the Angular Material docs still use the interim solution for icons: hard-coded data-URIs. We need to replace these icons with the new mdIcon component.

Updated:
 * Tabs pagination arrows (needs theming support)
 * Docs menu button
 * Docs demo source button

Closes #1464


src/components/tabs/js/tabsDirective.js
```diff
@@ -110,6 +110,7 @@ function TabsDirective($mdTheming) {
           'ng-if="pagination.active && pagination.hasPrev" ' +
           'ng-click="pagination.clickPrevious()" ' +
           'aria-hidden="true">' +
+          '<md-icon md-svg-icon="img/icons/tabs-arrow.svg"></md-icon>' +
```
marcysutton: @ThomasBurleson do these paths need to be configurable?

PR #1477, opened 2015-02-12
update(icon): added unit test to ensure error when icon or set not found

programmist:


programmist:
Sorry.  Somehow I managed to close my other PR.

programmist:
Updated PR with more tests.  Also split directive and service tests into separate files.

PR #1488, opened 2015-02-12
fix(gridlist): Prevents media from being unwatched immediately

shyndman:


ThomasBurleson:
Fixed with SHA a4104215be8c3aa902095dcb182d28b05ff3b79e

PR #1503, opened 2015-02-13
Couple of gridList fixes

shyndman:


src/components/gridList/gridList.js
```diff
@@ -164,15 +164,23 @@ function GridListDirective($interpolate, $mdConstant, $mdGridLayout, $mdMedia, $
...
-                      styles: getStyles(ps.position, ps.spans,
-                          colCount, rowCount,
-                          gutter, rowMode, rowHeight)
+                      grid: {
+                        element: element,
+                        style: getGridStyle(colCount, rowCount, gutter, rowMode, rowHeight)
+                      },
+                      tiles: tilePositions.map(function(ps, i) {
+                        var element = angular.element(tiles[i]);
+                        element.scope().$mdGridPosition = ps; // for debugging
```
gkalpak: Besides being a bad idea to leave debug data in production code, `scope()` can return `undefined`, in which case this line would throw an error.

shyndman: Good point. Completely forgot about the ability to disable debug data.

I was actually thinking this object would be useful for application developers, but I'll pull it until I have something that's more appropriate.

shyndman: Resolved.

src/components/gridList/gridList.js
```diff
@@ -375,7 +412,8 @@ function GridLayoutFactory($mdUtil) {
       reflow: function(customAnimatorFn) {
         reflowTime = $mdUtil.time(function() {
           var animator = customAnimatorFn || defaultAnimator;
-          tiles.forEach(function(it) {
+          animator(gridStyles.grid.element, gridStyles.grid.style);
```
ThomasBurleson: What is the intention to animate the grid ?

src/components/gridList/gridList.js
```diff
@@ -375,7 +412,8 @@ function GridLayoutFactory($mdUtil) {
...
+          gridStyles.tiles.forEach(function(it) {
```
ThomasBurleson: This approach feeds the animator each tile individually to the animator.
Normally the animator will want the SET of animations and will decide how to animate.

```js
function animator( targets ) {
 targets = angular.isArray(targets) ? targets : [targets];

// First collapse all targets to center
// collapseToCenter(targets);

// Then explode to destination positions
 angular.forEach(function(it) {
   angular.element(it.element).attr(it.styles);
});

} 

}
```

shyndman: Good call.

src/components/gridList/gridList.js
shyndman: I thought it might be interesting to have that flexibility, and even more so when I do what you're suggesting below. Maybe you want to have the grid flip over and reveal the new layout or something. :)

bcbroussard:
@shyndman Nice work on the height fix. Will this make it into the official v0.8.0 release?

shyndman:
Hope so. Maybe @ThomasBurleson can weigh in.

PR #1527, opened 2015-02-16
fix(autocomplete) select item on mouse click

rrharvey:
The ng-if on the list item was removing the item on the md-autocomplete blur due to $mdAutocompleteCtrl.hidden being set to true. The ng-click handler for the list item was never called.

### DIFF
--- a/src/components/autocomplete/js/autocompleteDirective.js
+++ b/src/components/autocomplete/js/autocompleteDirective.js
@@ -58,8 +58,8 @@
           <li ng-repeat="(index, item) in $mdAutocompleteCtrl.matches"\
               ng-class="{ selected: index === $mdAutocompleteCtrl.index }"\
               ng-if="searchText && !$mdAutocompleteCtrl.hidden"\
-              ng-click="$mdAutocompleteCtrl.select(index)"\
               ng-transclude\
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@robertmesserle - FYI

robertmesserle:
Thanks for the PR - I resolved this by switching to `ng-show` in e088f6aceb108449b7e6786ef3f1329d805a8001.  That felt cleaner than adding a manual click event.

rrharvey:
Looks good. Thanks for the fix.

FUT:
@rrharvey @robertmesserle It seems to has the same problem still. It works OK with arrows and enter button but not with mouse. If I break the `ng-show` and make the options always be visible, it works just fine:

`ng-show="searchText && !$mdAutocompleteCtrl.hidden"\`

But that is definitely not the solution. Checked in chrome & ff. Version 0.8.2

robertmesserle:
@FUT I am not able to see this error anymore in master - can you create a plunker or codepen demonstrating this bug against the latest code?

FUT:
Just a second I will check master branch

FUT:
@robertmesserle Thanks! It works just fine on master.

Despite some crazy errors I love material. Guys, you are the best!

PR #1529, opened 2015-02-16
refactor(iterator): refactor `findSubsequentItem()` to avoid recursion

gkalpak:
(As per https://github.com/angular/material/commit/124466e71945a4515a7b5742310594e8753c4314#commitcomment-9204854.)

### DIFF
--- a/src/core/util/iterator.js
+++ b/src/core/util/iterator.js
@@ -190,41 +190,40 @@
     }
 
     /**
-     * Find the next item. If reloop is true and at the end of the list, it will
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@gkalpak - I believe these changes break fundamental assumptions about the iterator; a noticeable change the loss of `validate( item )` to skip the next/prevsiou item(s) if **not** a valid selection.

gkalpak:
Not sure what assumptions you are talking about. The `validate()` check is still in place.
AFAIKT the functionality is exactly the same. (Also, note that the tests still pass.)

Issue #1536, opened 2015-02-16
Clicking non-submit button causes form to be submitted


demisx:
Using the latest angular material build `v0.8.0-rc1-master-f704dda`. Have two buttons in my form like this.

```html
<form  ng-submit="vm.createAd(newAdForm)" name="newAdForm" novalidate>
...
<section class="photos" layout="row" layout-align="start center">
  <label>Photos</label>
  <md-button class="md-raised">Add photos</md-button>
</section>
...
<md-button class="md-accent md-raised md-primary"
  type="submit"
  ng-disabled="newAdForm.$invalid">
  Publish
</md-button>
</form>
```

The problem is that clicking `Add photos` button causes the form also being submitted, though I have only one Publish button defined with `type="submit"`.

gustavohenke:
That's the default behavior of the tag `<button>`.
If you don't specify `type` attribute, the default is `submit`, so you have to specify `type="button"` for your "Add photos" button

demisx:
@gustavohenke I see. Thanks!

ThomasBurleson:
@gustavohenke - this issue has occurred more than 2x now... perhaps the `<md-button>` should set the `type="button"` if **type** is not specified ?

marcysutton:
@ThomasBurleson we discussed this a while back, it just never got done. See this closed PR: https://github.com/angular/material/pull/1293

vivekimsit:
Probably its fixed in this PR:  #1554

gkalpak:
This more of a feature than a fix (unless by "fix" you mean "fix the HTML spec :laughing:) !

red010b37:
also related - #3127

Cosmos-it:
I have the same issues too! What is the solution for this?

gkalpak:
This issue has been fixed. Maybe you are using an old version or your problem is different.

altsanz:
lol, this is happening right now

Splaktar:
@altsanz if you need some help debugging an issue, please post a reproduction to [Stack Overflow](https://stackoverflow.com/), [Gitter](https://gitter.im/angular/material), or the [AngularJS Material forums](https://groups.google.com/forum/?pli=1#!forum/ngmaterial).

If you find a bug, please file a new issue with a reproduction.

PR #1554, opened 2015-02-18
fix:(mdbutton): allow default 'type' attribute

vivekimsit:
Prevent default submit action  #1536


rschmukler:
See my comment on #1555. It doesn't make sense to have our buttons any different than normal buttons.

chrislambe:
Is there a good reason to have deviated from the HTML spec here? I get that people may not know or like the behavior, but it is clearly documented part in the spec.

The relevant section, for reference:
http://www.w3.org/TR/html5/forms.html#the-button-element

> The missing value default is the Submit Button state.

Issue #1473, opened 2015-02-11
Autocomplete is not Accessable (Tested on NVDA)


zamamohammed:
I have tested Autocomplete with NVDA and it reads "blank" instead of the list items when you use arrows to scroll through the list

marcysutton:
@zamamohammed we worked on a fix and just pushed it to master in a787e2262fd805e1edd5fc5fd412a332ef673657. Can you try again?

zamamohammed:
@marcysutton I am still hearing the 'blank'

Please listen to the recording:
https://www.dropbox.com/s/q9ezng1yocskjqg/Autocomplete.swf?dl=0

marcysutton:
I see what you're saying. It's really slow to announce "There is one match available" or the selection on arrow down in Firefox/NVDA despite having `aria-live="assertive"` on the text to be announced. Sometimes it doesn't read at all. It's great in VoiceOver, but not so good in JAWS and NVDA (which have higher usage). I'll see if I can make it better.

zamamohammed:
Most of the custom dropdown plugins in the Community (Chosen, Select2 etc) also has the same issue with it, the only implementation I liked is of Jquery UI (http://jqueryui.com/selectmenu/)

marcysutton:
Good to know. I think we can get it working by improving our implementation. The inspiration for our autocomplete was actually Twitter's Typeahead/draft tweet functionality.

robertmesserle:
@marcysutton I believe you were looking into this, assigned to you.

zamamohammed:
@marcysutton when can I test this on material.angular.com ?

marcysutton:
@zamamohammed it is up on the docs site now! Just not out in a release yet.

PR #1556, opened 2015-02-18
fix(autocomplete): set tabindex 1000 for the clear button

tsdaemon:
Autocomplete tabulation stops on the clear button instead go to the next form element. It is definitely wrong as clear button is not pretend to be next input element.

### DIFF
--- a/src/components/autocomplete/js/autocompleteDirective.js
+++ b/src/components/autocomplete/js/autocompleteDirective.js
@@ -51,7 +51,8 @@
           <button\
               type="button"\
               ng-if="searchText"\
-              ng-click="$mdAutocompleteCtrl.clear()">\
+              ng-click="$mdAutocompleteCtrl.clear()"\
+			  tabindex="1000">\
               <md-icon md-svg-icon="cancel"></md-icon>\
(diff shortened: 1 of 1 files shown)


marcysutton:
A better solution would be to set `tabIndex="-1"` on the clear button by default, then update it to `tabIndex="0"` or `element.removeAttr('tabIndex')` when the autocomplete has results in it. That way we're not just putting an arbitrary tab index in the page and changing the tab order, but rather changing the behavior of the component.

tsdaemon:
Thanks, I didn't knew about a tabIndex="-1".
When I selected a result of autocomplete, I certainly do not want tab to the clear button, I want tab on the next form element. Why not set tabIndex="-1" at all?

ThomasBurleson:
won't fix

marcysutton:
I can work on this as part of https://github.com/angular/material/issues/1473.

marcysutton:
Actually the button already had ng-if, so you were right to say "won't fix" @ThomasBurleson. Not a good idea at all to remove the button from the keyboard when it's visible.

PR #1562, opened 2015-02-19
feat(mdSelect): add select component

### LINKED ISSUE #276: mdSelect: Add select component
The mdSelect is a type of [Material Design Menu](https://www.google.com/design/spec/components/menus.html#menus-behavior) that behaves like and resembles a `<select>`. 

Menu behavior

<h2>Functional requirements</h2>
 * Menu should be positioned over their "emitting" elements, establishing context
 * The selected menu item should not be shown as a duplicate when expanded
 * Dismiss menu by selecting an item (unless selecting multiple), tapping outside the menu, or using the Escape key
 * Menu animates out from emitting item when expanding

<h2>Accessibility requirements</h2>
 * **mdSelect element**
* `[aria-live="polite"]` makes sure selections are announced in a screen reader. Copy list item text to button only when selection is made with enter/space key, or click (not arrow keys). That way it won't read aloud twice.
 * **Button element**
* `[aria-haspopup="true"]` indicates button launches associated input
* `[aria-controls="_listID_"]` links button to options list
* Needs unique ID for pairing with list
 * **Unordered list**
* [`role="menu"`](http://www.w3.org/TR/wai-aria/roles#menu) to communicate purpose of list
* `aria-labelledby="_buttonID_"` to pair with button element
* `aria-expanded="true | false"` to indicate when associated panel is expanded
* Unique ID for pairing with button
* `tabindex="-1"`to script focus but remove from tab order
 * **Child list items**
* [`role="menuitem"`](http://www.w3.org/TR/wai-aria/roles#menuitem) on each child item
* `tabIndex="-1"` for scripting
 
<h2>Resources</h2>
 * [WAI-ARIA Listbox](http://www.w3.org/TR/wai-aria/roles#listbox)
 * [Using "listbox" and "option" roles vs. "menu" and "menuitem"](http://www.marcozehe.de/2014/03/11/easy-aria-tip-7-use-listbox-and-option-roles-when-constructing-autocomplete-lists/)

marcysutton:
This PR adds a more complete version of the mdSelect component, including keyboard and screen reader accessibility and unit test coverage. Some items outstanding:

 * Ensure component has an accessibility label
 * Documentation and demos outlined in select.js

Closes #276

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -1,0 +1,621 @@
+(function() {
+'use strict';
+
+/***************************************************
+
+### TODO ###
+
+**IMPLEMENTATION**
+
+- [x] Async loading of md-options (eg no options for one second, then they all show up).
+- [x] Style select button to match spec
+- [x] Implement CSS for an md-progress-circular in the select while the options are loading. 
+- [x] Implement CSS for md-optgroup
+- [ ] Implement keyboard interaction
+- [ ] Implement ARIA & accessibility
+- [x] Finish theming - theme color
+- [x] Add element option to interimElement to avoid recompiling every time (?)
+
+### TODO - POST RC1 ###
+- [ ] Abstract placement logic in $mdSelect service to $mdMenu service
+
+**DOCUMENTATION AND DEMOS**
+
+- [ ] ng-model with child mdOptions (basic)
+- [ ] ng-model="foo" ng-model-options="{ trackBy: '$value.id' }" for objects
+- [ ] mdOption with ng-value
+- [ ] mdOption with value
+- [ ] Multiple repeaters full of mdOptions
+- [ ] md-optgroups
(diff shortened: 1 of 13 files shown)


src/components/select/select.js
```diff
@@ -0,0 +1,733 @@
...
+  };
+
+  function compile(element, attr) {
+    // The user is allowed to provide a label for the select as md-select-label child
+    var labelEl = element.find('md-select-label').remove();
+
+    // If not provided, we automatically make one
+    if (!labelEl.length) {
+      // Use the input as a label if there's an input inside.
+      if ( (labelEl = element.find('input')).length ) {
```
marcysutton: @rschmukler any idea what this is supposed to be doing? Why would there be an input inside of `md-select`?

Issue #1269, opened 2015-01-23
colors and hues should be available everywhere.


SanderElias:
I created [this sample on plunker](http://plnkr.co/edit/M22sy0AI7Bax3Cop3OmI?p=preview) to demonstrate a few issues I'm having with forms. 

If you look at like 24 of index.html in my sample, you see there is a heading there. This heading is not visible in the page, because I don't know how to set the background color according to the current selected theme/colorpallete. I added an md-hue-1 to the class-list as well as an attribute, but it is not giving the result I need.

This is also mentioned in #537, but I still could not find a solution to it.

SanderElias:
I took my solution from #537. and rewrote it a bit so it will work with the new $mdThemingProvider.
For anybody that need something similar, have a look at my mdColors module. It's available in [this plunker](http://plnkr.co/edit/SHkXcAQDNPHWloBnVxrw?p=preview)

This will serve my needs for now, but it still feels awkward that I have to go trough those lengths just to be able to get my hands on the selected color palette. 
I still feel like I somehow miss the obvious!

iRouge:
Sander, I share your feelings, your concerns... and your solution !

Thanks a lot

SanderElias:
@iRouge I began to believe I was to only one ;)

ThomasBurleson:
@SanderElias - I think we should include a version of your **colors.js** and publish a service `$mdColors` service for easy color lookups.
@rschmukler - thoughts ?

SanderElias:
@ThomasBurleson I would like that! Should I add support for md-hue-x and so on too?

ThomasBurleson:
@SanderElias - let's start with a simple, clean API for the initial work.
Can you create the $mdColors service and submit a PR ? Thx is advance.

jonnysamps:
I believe I have a pretty solution for color lookups: checkout my plunkr: http://plnkr.co/edit/gBDV2ABghTMqqs5YaElB?p=preview

SanderElias:
@marcysutton I'm preparing an PR, as it looks now, I will be done by the end of the day.

SanderElias:
@marcysutton I created a new branch, its in PR #1599. Somehow it would not let me update my other PR

mtraynham:
This is awesome! +1

rschmukler:
@SanderElias finally got to reviewing the PR. I want to move it back a bit as I want to extend the API to make the following possible:

[...]

PR #1566, opened 2015-02-19
Create colors.js for $mdColor service.

SanderElias:
A service to create css rules for foreground and background colors a talked about in #1269

### DIFF
--- a/src/core/services/theming/colors.js
+++ b/src/core/services/theming/colors.js
@@ -1,0 +1,129 @@
+(function () {
+    "use strict";
+
+    //module global to keep an reference to the dynamic style sheet
+    var customSheet;
+    //module global to available color pallettes
+    var colorStore     = {};
+    var defaultPalette = '';
+
+/**
+ * @ngdoc service
+ * @name $mdColors
+ * @module material.core.color
+ *
+ * @description
+ * `$mdColors` makes it possible to use the colors of any palette to be available
+ *  as CSS rules in the form of md-bg-[colorname] and md-fg-[colorname] (as foreground and background)
+ *  If you want to activate another palette you can use the $mdColor.loadPalette(palletteName) method.
+ *
+ *  the $mdColors.palettes array lists all available palettes
+ *  the $mdColors.palette property lists the current selected palette
+ *  the $mdColors.colors array lists all the available colors with hex encoded color values for the current selected palette.
+ *
+ **/
+
+
+    angular
+        .module('material.core.colors',['material.core.theming'])
+        .service('$mdColors', mdColors)
+        .config(configColors)
+        .run(loadDefaults);
+
+    configColors.$inject = ['$mdThemingProvider'];
+    function configColors($mdThemingProvider) {
+        // fetch the colors out of the theming provider
+        Object.keys($mdThemingProvider._PALETTES).forEach(parsePalette);
+
+        // I need access to the default 'default', hardcoded for now as I miss access to it.
+        defaultPalette = $mdThemingProvider._THEMES['default'].colors.primary.name;
+        return;
+
+        // clone the palette colors to the colorStore var
+        function parsePalette(paletteName) {
+            var palette = $mdThemingProvider._PALETTES[paletteName];
+            var colors  = [];
+            colorStore[paletteName] = colors;
+            Object.keys(palette).forEach(copyColors);
+            return ;
+
+            function copyColors(colorName) {
+                // use an regex to look for hex colors, ignore the rest
+                if (/#[0-9A-Fa-f]{6}|0-9A-Fa-f]{8}\b/.exec(palette[colorName])) {
+                    colors.push({color:colorName,value:palette[colorName]});
+                }
+            }
+        }
+    }
+
+    loadDefaults.$inject=['$mdColors'];
+    function loadDefaults($mdColors) {
+        // this sets the default that is stored in the config-fase into the service!
+        $mdColors.palettes = Object.keys(colorStore);
+        $mdColors.loadPalette(defaultPalette);
+    }
+
+    mdColors.$inject = ['$interpolate'];
+    function mdColors ($interpolate) {
+        // wrap all of the above up in a reusable service.
+        var service        = this;
+        service.palette    = defaultPalette;
+        service.colors     = [];
+        service.palettes   = [];
+        service.loadPalette = loadPalette;
+
+        return service;
+
+        function loadPalette(newPalette) {
+            if(service.palettes.indexOf(newPalette) === -1) {
+                throw new Error('$mdColors.loadPalette: Unknown palette name ' + newPalette);
+            }
+            service.palette = newPalette;
+            service.colors  = colorStore[newPalette];
+            createStyleSheet();
+        }
+
+        function createStyleSheet () {
+            var colors = service.colors;
+            var fg, bg;
+            if (typeof customSheet === 'undefined') {
+                // use closure for caching the styleSheet
+                newStyleSheet();
+            } else {
+                // remove existing rules
+                // TODO: look into disabling/enabling pre-build style-guides
+                //       in stead of delete and recreate!
+                while (customSheet.cssRules.length > 0 ) {
+                    customSheet.deleteRule(0);
+                }
+            }
+
+            // set up interpolation functions to build css rules.
+            fg  = $interpolate('.md-fg-{{color}} { color:{{value}};}');
+            bg  = $interpolate('.md-bg-{{color}} { background-color:{{value}};}');
+
+            colors.forEach(function (color) {
+                // insert foreground color rule
+                customSheet.insertRule(fg(color));
+                // insert background color rule
+                customSheet.insertRule(bg(color));
+            });
+        }
+    }
+
+    function newStyleSheet() {
+        // function to ad an dynamic style-sheet to the document
+        var style   = document.createElement("style");
+        style.title = 'Dynamic Generated my Angular-Material';
+        // WebKit hack... (not sure if still needed)
+        style.appendChild(document.createTextNode(""));
+
+        document.head.appendChild(style);
+        // store the sheet in the closure for reuse
+        // creating a new sheet is a 'costly' operation, and I
+        // just need one.
+        customSheet = style.sheet;
+        return style.sheet;
+    }
+
+}());
--- a/src/core/services/theming/colors.js
+++ b/src/core/services/theming/colors.js
@@ -1,8 +1,6 @@
 (function () {
     "use strict";
 
-    //module global to keep an reference to the dynamic style sheet
-    var customSheet;
     //module global to available color pallettes
     var colorStore     = {};
     var defaultPalette = '';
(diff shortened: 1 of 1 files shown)


src/core/services/theming/colors.js
```diff
@@ -0,0 +1,129 @@
...
+
+/**
+ * @ngdoc service
+ * @name $mdColors
+ * @module material.core.color
+ *
+ * @description
+ * `$mdColors` makes it possible to use the colors of any palette to be available
+ *  as CSS rules in the form of md-bg-[colorname] and md-fg-[colorname] (as foreground and background)
+ *  If you want to activate another palette you can use the $mdColor.loadPalette(palletteName) method.
```
gkalpak: _`$mdColor.loadPalette(paletteName)`_ ? (typo + backticks)

src/core/services/theming/colors.js
```diff
@@ -0,0 +1,129 @@
...
+ **/
+
+
+    angular
+        .module('material.core.colors',['material.core.theming'])
+        .service('$mdColors', mdColors)
+        .config(configColors)
+        .run(loadDefaults);
+
+    configColors.$inject = ['$mdThemingProvider'];
```
gkalpak: This is not necessary - the build process uses `ngAnnotate`.
It's better to keep the codebase consistent.

src/core/services/theming/colors.js
```diff
@@ -0,0 +1,129 @@
...
+        function parsePalette(paletteName) {
+            var palette = $mdThemingProvider._PALETTES[paletteName];
+            var colors  = [];
+            colorStore[paletteName] = colors;
+            Object.keys(palette).forEach(copyColors);
+            return ;
+
+            function copyColors(colorName) {
+                // use an regex to look for hex colors, ignore the rest
+                if (/#[0-9A-Fa-f]{6}|0-9A-Fa-f]{8}\b/.exec(palette[colorName])) {
```
gkalpak: This RegExp seems wrong... Not sure what the second part is checking for, though.

gkalpak: Since you are not capturing anything, `test` might be a better choice (instead of `exec`).

src/core/services/theming/colors.js
```diff
@@ -0,0 +1,129 @@
...
+                    colors.push({color:colorName,value:palette[colorName]});
+                }
+            }
+        }
+    }
+
+    loadDefaults.$inject=['$mdColors'];
+    function loadDefaults($mdColors) {
```
gkalpak: Maybe you could get rid of the module-globals and this run block by using a provider.
This would have the added benefit that the service wouldn't be instantiated if the user didn't need it.

src/core/services/theming/colors.js
```diff
@@ -0,0 +1,129 @@
...
+    mdColors.$inject = ['$interpolate'];
+    function mdColors ($interpolate) {
+        // wrap all of the above up in a reusable service.
+        var service        = this;
+        service.palette    = defaultPalette;
+        service.colors     = [];
+        service.palettes   = [];
+        service.loadPalette = loadPalette;
+
+        return service;
```
gkalpak: This is not necessary (since you are using `.service(...)`).

src/core/services/theming/colors.js
```diff
@@ -0,0 +1,129 @@
...
+                }
+            }
+        }
+    }
+
+    loadDefaults.$inject=['$mdColors'];
+    function loadDefaults($mdColors) {
+        // this sets the default that is stored in the config-fase into the service!
+        $mdColors.palettes = Object.keys(colorStore);
+        $mdColors.loadPalette(defaultPalette);
```
gkalpak: This doesn't seem to be necessary, since you are using `service.palette = defaultPalette` below.
(The above comment still applies, though.)

src/core/services/theming/colors.js
```diff
@@ -0,0 +1,129 @@
...
+    }
+
+    mdColors.$inject = ['$interpolate'];
+    function mdColors ($interpolate) {
+        // wrap all of the above up in a reusable service.
+        var service        = this;
+        service.palette    = defaultPalette;
```
gkalpak: Why `palette` but not `colors` not creating a stylesheet ?

src/core/services/theming/colors.js
```diff
@@ -0,0 +1,129 @@
...
+                // insert background color rule
+                customSheet.insertRule(bg(color));
+            });
+        }
+    }
+
+    function newStyleSheet() {
+        // function to ad an dynamic style-sheet to the document
+        var style   = document.createElement("style");
+        style.title = 'Dynamic Generated my Angular-Material';
```
gkalpak: _Dynamically Generated by Angular-Material_ ?

src/core/services/theming/colors.js
```diff
@@ -0,0 +1,129 @@
...
+        // WebKit hack... (not sure if still needed)
+        style.appendChild(document.createTextNode(""));
```
gkalpak: `""`  -->  `''`

src/core/services/theming/colors.js
SanderElias: @gkalpak I have thought about that, but the colors are only available in the form I need them, after MD finishes bootstrapping. I looked into use the material.core.themeing.palette constant, but that is rewritten somewhere (error in my eye's its a CONST!), to use rgb colors, and for this I needed the hex notation. I can reconvert them tough, or paste them in as RGB.
What are you toughts about that?

src/core/services/theming/colors.js
```diff
@@ -0,0 +1,129 @@
...
+                customSheet.insertRule(fg(color));
+                // insert background color rule
+                customSheet.insertRule(bg(color));
+            });
+        }
+    }
+
+    function newStyleSheet() {
+        // function to ad an dynamic style-sheet to the document
+        var style   = document.createElement("style");
```
gkalpak: It is better to use single-quotes for consistency.

src/core/services/theming/colors.js
```diff
@@ -0,0 +1,129 @@
...
+        style.title = 'Dynamic Generated my Angular-Material';
+        // WebKit hack... (not sure if still needed)
+        style.appendChild(document.createTextNode(""));
+
+        document.head.appendChild(style);
```
gkalpak: Please don't use `document` directly :smiley:
Use it via an DI-ed service instead (`$window`, `$document`).

src/core/services/theming/colors.js
```diff
@@ -0,0 +1,129 @@
+(function () {
+    "use strict";
+
+    //module global to keep an reference to the dynamic style sheet
+    var customSheet;
```
gkalpak: This doesn't need to be a "module global" (with a minor 2-lines refactoring).
Module globals aren't that much better than globals anyway :smile:

src/core/services/theming/colors.js
SanderElias: I forgot to replace them by /* @ngInject */. sorry about htat.

src/core/services/theming/colors.js
SanderElias: This regex checks if a string is a valid hex-color, it accepts multiple notations. I ran it trough an regex validator, and that gave it's OK, but I'm not an regex crack.
You are right about the test, I will change that right away.

gkalpak:
This is a good start. Thx @SanderElias 

Left some comments inline :)
We should also have tests of course.

I was thinking if it would make sense to either have more classes for setting stuff like `borderColor`, `outlineColor` etc or someway to expose the hex values themselves, in order to be able to set other stuff as well (e.g. `boxShadow` color).

src/core/services/theming/colors.js
SanderElias: In lieu of an better title!

src/core/services/theming/colors.js
gkalpak: If you tell me what you are trying to match, I can help fix it up.
I think you should also match `#XYZ` (3 instead of 6 chars).

In any case the second part (`0-9A-Fa-f]{8}\b`) wouldn't match anything reasonable.
It would match the literal string `0-9A-Fa-f]]]]]]]]` plus some word character.
It is more than obvious that a leading `[` is missing after the `|`, but I suspect that there are more problems with it.

I've never seen this 8-chars long form, so I am not sure what it's supposed to match, but you probably meant: `#([0-9A-Fa-f]{6}|[0-9A-Fa-f]{8})\b`

src/core/services/theming/colors.js
gkalpak: Actually, I didn't mean load the palette in the provider. What I meant was not load a palette until the user instantiates the `$mdColors` service explicitly.
Thinking about it again, there is a design desicion to be made:

1. Automatically instantiate `$mdColors` and load the default palette.
2. Have the user explicitly load a palette if they want to use `$mdColors`.

With (1) there is less boilerplate for people that do want to use `$mdColors`, but there is unnecessary code run for people that don't care to use it (e.g. creating a stylesheet that is never ging to be used).
With (2) we avoid unnecessary initialization code, but require the user to "manually" load a palette if thy want to use it.

(Not sure which one I prefer. It would be good to somehow benchmark the initialization code).

--
BTW, with the current implementation, can there only be one palette active at a time ?
Wouldn't it be nice to be able to use more than one ?

src/core/services/theming/colors.js
gkalpak: No, I meant there are some typos:
_Dymanic_ -> _Dynamically_
_my_ -> _by_

(The title is fine by me otherwise :smile:)

src/core/services/theming/colors.js
SanderElias: As this was driven out of my own need, I opted for 1.  I'll think this true, and will do some performance tests.

Yes, it would be nice to have more then 1 palette active, as it also would be nice to make the md-hue-x available. 
I proposed this to @ThomasBurleson, and hos comment was go with this for a first version.

src/core/services/theming/colors.js
SanderElias: @gkalpak I prefer it to do it this way, because it is more in line with the John Papas styleguide. Also I like to explicitly return something. 
I'm not aware of any downsides/performance issues on this approach, aside from a few extra tokens in the JS.

gkalpak: I don't think John recommends return from a `.service`. (In fact, he recommends using `.factory` for consistency - and in that case you **need** to return the service, but that's another story).

I prefer using `.service` (and aliasing `this`). I was just referring to the `return service` line (which is unnecessry and possibly confusing, but not a big deal - just a matter of style anyway.)

src/core/services/theming/colors.js
```diff
@@ -0,0 +1,137 @@
...
+                customSheet.deleteRule(0);
+            }
+
+            // set up interpolation functions to build css rules.
+            fg  = $interpolate('.md-fg-{{color}} { color:{{value}};}');
+            bg  = $interpolate('.md-bg-{{color}} { background-color:{{value}};}');
+
+            colors.forEach(function (color) {
+                // insert foreground color rule
+                customSheet.insertRule(fg(color), index);
```
ajoslin: Each new rule inserted or deleted will cause a DOM repaint. This is why we do it all in one string and just set `style.innerHTML = rulesString`

SanderElias: @ajoslin If you insist I will change this, as it is not a big issue to me, however, your statement is only partly true. Modern browsers will not do a DOM repaint while JS is running, unless you are request any form of coordinates back from the DOM.
The work the browser has to do stays the same either way, so It's your call.

marcysutton:
@SanderElias, can you rename your branch to not use a `$`? I want to check it out and run the code, but I'm getting an error due to the special character.

ThomasBurleson:
@SanderElias - can you close this PR; since we have another, more recent version ?

PR #1578, opened 2015-02-19
amend(select): fix race condition when loadingAsync is done

ajoslin:


marcysutton:
Closed with 3cbd3e0e11777547bb47176942a7a69e6f47bfed.

PR #1579, opened 2015-02-19
perf(select): use live nodeList, dont create array on arrow press

ajoslin:
- Fixes DOM repaint from querySelectorAll every key press by using a live nodelist
- Fixes having to allocate a new node list and new array every key press (which could cause unknown garbage collcetion burps later)


marcysutton:
Closed with 541a01ef0b20733b111f69f9029ec80e3974ec73.

PR #1582, opened 2015-02-20
enhance(dialog): add option focusCloseBtn=true

wuv1982:
alter this option to support customize focus.
e.g. let input text get the focus by default

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -385,6 +385,7 @@ function MdDialogProvider($$interimElementProvider) {
       escapeToClose: true,
       targetEvent: null,
       disableParentScroll: true,
+      focusCloseBtn: true,
       transformTemplate: function(template) {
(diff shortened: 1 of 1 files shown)


wuv1982:
i have just signed the CLA

marcysutton:
I think it would be better to supply an element to focus rather than opting out entirely. Focus management is very important to the dialog, so it shouldn't be so easy to turn off.

wuv1982:
I tried to use a customize directive `my-focus` to set the focus on my element , but `closeButton.focus();` always force the focus on the `CloseBtn`. 
I just want a way to avoid this behavior and set the focus on my element in this case.

Also, I think it is a good idea for setting the focus on `CloseBtn` by default for almost the using cases.
So I use a option to opt the choice.
If u don't like the `option` , i think remove the force focus behavior maybe better.

wuv1982:
thanks for accepting

PR #1592, opened 2015-02-20
fix(tabs): add component.icon to module

rgthree:
The tabs directive is using md-icon but its module was not requiring material.component.icon.


rgthree:
Ha. I've signed the CLA.

Issue #142, opened 2014-08-19
Buttons require a visual focus state


marcysutton:
To make flat and raised buttons accessible from the keyboard, they should have visual contrast upon user interaction. In the [Material Design](http://www.google.com/design/spec/components/buttons.html#buttons-main-buttons) spec, focus states are illustrated with a ripple effect: 

![Flat button states](https://cloud.githubusercontent.com/assets/1045233/3960549/63f99d06-2742-11e4-93e9-5959e48ce9ef.png)
![Raised Button states](https://cloud.githubusercontent.com/assets/1045233/3960518/c1e58e08-2741-11e4-9369-d127feec41ea.png)

A similar ripple effect is needed to add focus states to all buttons.

sambryant:
It seems like some of you are suggesting that the problem is that the element is remaining in a focus state. I think it's supposed to do that: http://jsfiddle.net/nv3rvru1/8/

albertboada:
:+1:

skribbz14:
Don't you think there could be some sort of option applied to the <md-button> that would choose whether we want the :focus to stay on or be removed after clicking? This seems to be a personal preference that not everyone agrees on, so allowing the developer to choose would make everyone happy.

marcysutton:
Buttons actually have a focus state now, so I'm closing this issue.

@skribbz14 to answer your question, consistency is more important in this case. Supporting both of those options would add a lot of complexity to maintain, and in my opinion would fragment the Material Design user experience. I have a question out to the UX team about how controls should handle visual focus in a way that works with the browser and looks intentional. Work on that can be tracked here: https://github.com/angular/material/pull/1594

There is also a separate issue open for the button pressed state: #2089

chanakaDe:
Yeah all the buttons are ok with now.

Issue #556, opened 2014-11-06
md-button stays focused after a click


PaulMougel:
After clicking or touching on a `<md-button>`, it stays focused, both on latest Firefox, Chrome, mobile Chrome and mobile Safari (see the  [demo page](https://material.angularjs.org/#/demo/material.components.button)). Only by clicking somewhere else will the button loose focus. Is it the intended behavior?

Screenshot in latest Chrome stable, on Mac Os Yosemite, on the demo page:
![untitled](https://cloud.githubusercontent.com/assets/1788596/4944296/49b32bbc-65fc-11e4-91d7-873f0f6a3c95.gif)

caok2709:
Does anyone know how to disable md-button hover effect?

vas1468:
.md-button:not([disabled]):hover {
    background-color: transparent;
  }

ananyaojha:
What's the exact solution to this issue? I'm still getting this in my application. When switching tab and coming back to application, button is still showing active.

simhnna:
@marcysutton I think this issue should still be open. The linked issue has been closed without a resolution to this one...

marcysutton:
@simhnna I don't work on Angular Material anymore, but I'd recommend looking into the proposed <a href="https://github.com/WICG/focus-visible"><code>:focus-visible</code> CSS selector</a> or the [What Input](https://github.com/ten1seven/what-input) JavaScript library. I lovingly refer to this problem as <a href="https://marcysutton.com/button-focus-hell/">Button focus hell</a> because it isn't easy to solve in a way that supports all users. However, the aforementioned projects can help provide more styling hooks based on the user's input modality.

Splaktar:
@AngelQuirogaM Please submit Angular Material questions [here](https://groups.google.com/forum/#!forum/angular-material2) and issues [here](https://github.com/angular/material2/issues). This repo is for AngularJS Material.

marcysutton:
@AngelQuirogaM the problem with that CSS is you're hiding the visible focus state, which means keyboard users can't see where we are on the screen (a huge accessibility/usability problem). It doesn't get at the core issue, which is that mouse users don't like the persistent focus state. See my above comment for a better way forward.

PR #1594, opened 2015-02-20
fix(mdButton): focus styles only on keyboard

### LINKED ISSUE #1423: <md-button>: cancelling button tap/click does not remove :focus style
Hi, first thanks for this project, great stuff.

Noticing this behavior with md-buttons.. please let me know if this is expected behavior or if an enhancement that might create a better UX.

If user clicks a button but drags their finger/cursor off the button and cancels the tap/click, the :focus style is not removed from the button. So like in the examples on your demo page, the button will remain with darker box shadow and darker button shade.

marcysutton:
This change limits focus styles to the keyboard and not on mouse clicks by setting a flag on mouseover/mouseout and checking for that flag when a focus event happens. The goal is to make custom focus styles act more like native buttons, where clicking doesn't persist a focus style.

Related prototype: http://codepen.io/marcysutton/pen/GgxboO

Closes #1423, #142

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -58,7 +58,7 @@ angular.module('material.components.button', [
  *  </md-button>
  * </hljs>
  */
-function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria) {
+function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria, $timeout) {
 
   return {
     restrict: 'EA',
@@ -95,6 +95,20 @@ function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria) {
         element.attr('tabindex', isDisabled ? -1 : 0);
       });
     }
+
+    // restrict focus styles to the keyboard
+    scope.mouseActive = false;
+    element
+      .on('mousedown', function() {
+        scope.mouseActive = true;
+        $timeout(function(){
+          scope.mouseActive = false;
+        }, 100);
+      })
+      .on('focus', function() {
+        if(scope.mouseActive === false) element.addClass('focus');
+      })
+      .on('blur', function() { element.removeClass('focus'); });
   }
 
 }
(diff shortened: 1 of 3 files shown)


marcysutton:
Alternative approach uses CSS and ARIA tricks instead of events: http://codepen.io/marcysutton/pen/OPveqd

gkalpak:
It is a (rather rare) corner, but what if the user places the mouse over the element and then uses the keyboard to press the button ?

marcysutton:
@gkalpak I don't think it matters on buttons. On checkboxes and switches, however, this is an issue.

marcysutton:
@ThomasBurleson can you take a look at this one? It will solve the problem with focus styles persisting on mouse click. For buttons, I think handling focus with events is better than creating and managing multiple elements. But they are both pretty hacky, to be honest.

ThomasBurleson:
@marcysutton - let's review this together after ng-conf, etc.

albertboada:
So, do Material specs not have a unique style for *focus* (instead of having to share one with *hover*)?
This looks like a very dodgy patch (probably from my ignorant point of view).

marcysutton:
@albertboada every solution to this problem is dodgy, quite frankly. It's a major and known limitation of browser focus/hover styles. Refer to this thread: https://github.com/angular/material/issues/556#issuecomment-62072801

src/components/button/button.js
```diff
@@ -86,6 +86,16 @@ function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria) {
         element.attr('tabindex', isDisabled ? -1 : 0);
       });
     }
+
+    // restrict focus styles to the keyboard
+    scope.mouseActive = false;
+    element
+      .on('mouseover', function() { scope.mouseActive = true; })
```
ThomasBurleson: var mouseActive;
element.on('moveover', function onMouseOver() {
   mouseActive = true; 
  element.on('mouseleave', onMouseOut );
   element.off('mouseover', onMouseOver );
}

ThomasBurleson:
@marcysutton - like it.

PR #1605, opened 2015-02-21
fix(md-input-container): Correctly style invalid input

alexgirdler:
Fixes styling on invalid inputs by adding touched class to md-input-container and only applying error style once the input has been touched to prevent style from being applied on page load. More detailed information with line comments.

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -64,6 +64,9 @@ function mdInputContainerDirective($mdTheming, $parse) {
     self.setFocused = function(isFocused) {
       $element.toggleClass('md-input-focused', !!isFocused);
     };
+    self.setTouched = function(isTouched) {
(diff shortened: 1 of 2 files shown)


src/components/input/input.js
```diff
@@ -170,7 +173,11 @@ function inputTextareaDirective($mdUtil, $window) {
     }
 
     var isErrorGetter = containerCtrl.isErrorGetter || function() {
-      return ngModelCtrl.$invalid && ngModelCtrl.$touched;
+      if (!ngModelCtrl.$touched) {
```
alexgirdler: Return true if input has not been touched, this causes the last value of $scope.watch to be true when the first digest is ran when the page loads. After the input has been touched then it will return the value of if the input is invalid.

Issue #1607, opened 2015-02-21
md-button fab with ng-href change icon position


luclanet:
![image](https://cloud.githubusercontent.com/assets/3264432/6316611/2506f1a6-ba2d-11e4-8f3c-85fc114d9ed9.png)
First with "ng-href", second one without.

marcysutton:
That means anchors and buttons inside of `md-button` have slightly different styles...`ng-href` creates an child `<a>` element instead of `<button>`. Should be an easy fix, want to create a PR?

luclanet:
What's a PR?

ruiaraujo:
@luclanet pull request, ie, fix it. ;)

iainmcgin:
I've been playing with this in codepen (http://codepen.io/iainmcgin/pen/wBjqyB), I have a proposed fix using absolute positioning for the icon within fabs, but it's kind of ugly and I hope someone finds a better way. Will create a pull request and let everyone discuss it there.

iainmcgin:
The fix from @ruiaraujo looks good, and is included in my codepen for comparison.

PR #1619, opened 2015-02-23
fix(button) center contents in case of anchor tag.

ruiaraujo:
Possible fix for #1607

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -5,8 +5,10 @@ $button-padding: 2px 6px 3px !default;
 $button-fab-width: 56px !default;
 $button-fab-height: 56px !default;
 $button-fab-padding: 16px !default;
+$button-fab-anchor-line-height: 40px !default;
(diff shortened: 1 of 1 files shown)


iainmcgin:
I don't think this works, just tried it in my codepen: 
http://codepen.io/iainmcgin/pen/wBjqyB
The vertical alignment still seems off compared to md-button without href.

iainmcgin:
Sorry, let me retract that: my codepen was not setting "vertical-align: middle" on the svg sub-elements like md-icon does. Your solution looks good.

ruiaraujo:
Great, I also did a local test like yours on my own app but I didn't bother with a codepen example. Thanks for that!

I am quite interested in other solutions that don't involve settings fixed margins or something like that.
Let's see if any CSS wizard shows up and delivers.

marcysutton:
Fixes were committed in a936e1edc964b69ffe5bf96905e348fc6f4c6b4d.

PR #1629, opened 2015-02-23
fix(md-input-container): Properly set and unset when input is invalid

### LINKED ISSUE #1485: md-input-container: ng-touched added on page load
I think [this](https://github.com/angular/material/commit/e0f53ddda7d080e80c08d5106b5586b697ea8e87) commit has broken the styles for `md-input-container`. It's adding the invalid styles when the page is loaded instead of after the user first touches the input. Also I believe `ng-touched` is not supposed to be added on page load.

alexgirdler:
This is to prevent isErrorGetter being called twice and allow the watch
function only to control if the input has an error.
Fixes #1485


ThomasBurleson:
Closed with SHA 747eb9c

PR #1642, opened 2015-02-24
feat(select): set label value based on md-option text

### LINKED ISSUE #1601: md-select: Make the display of the selected value configurable
> Seperated this issue from #1563 as suggested by @gkalpak (https://github.com/angular/material/issues/1563#issuecomment-75307693)

Currently there is no way to configure the display of the selected value in an `md-select`. 

[This Plunker](http://plnkr.co/edit/FzvqHazia1AAmTt5BjNb?p=preview) shows the differences between a native select (with support for ng-options `select as label for value in array` syntax) and `md-select`. Where the native select-element can have a label on both the select and its options, md-select currently only supports labels on the options.

Demo using [v0.8.0-rc1-master-eda14e9](https://github.com/angular/bower-material/commit/5d70169b6147dc15144d3f85a929a9ac3f429584). Change comments in source to change to master.

Note that in ng-options you can even apply a filter to the label in `select as label for value in array`, e.g. `select as label | translate for value in array` using the excellent [angular-translate](https://github.com/angular-translate/).

@jadrake75: can you update this issue with your comment?

rschmukler:
closes #1601

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -65,7 +65,7 @@ angular.module('material.components.select', [
  *   </md-select>
  * </hljs>
  */
-function SelectDirective($mdSelect, $mdUtil, $mdTheming, $interpolate) {
+function SelectDirective($mdSelect, $mdUtil, $mdTheming, $interpolate, $compile, $parse) {
   var intStart = $interpolate.startSymbol();
   var intEnd = $interpolate.endSymbol();
 
@@ -81,11 +81,11 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $interpolate) {
 
     // If not provided, we automatically make one
     if (!labelEl.length) {
-      labelEl = angular.element('<md-select-label>').html('<span>' + intStart + attr.ngModel + ' !== undefined ? ' + attr.ngModel + ': \'' + attr.placeholder + '\''+ intEnd + '</span>');
+      labelEl = angular.element('<md-select-label><span></span></md-select-label>');
     }
     labelEl.append('<span class="md-select-icon" aria-hidden="true"></span>');
     labelEl.addClass('md-select-label');
-    labelEl.addClass(intStart + attr.ngModel + ' ? \'\' : \'md-placeholder\'' + intEnd);
+    labelEl.addClass(intStart + attr.ngModel + ' !== undefined ? \'\' : \'md-placeholder\'' + intEnd);
     labelEl.attr('id', 'select_label_' + $mdUtil.nextUid());
 
     // There's got to be an md-content inside. If there's not one, let's add it.
@@ -105,12 +105,12 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $interpolate) {
     }
 
     // Use everything that's left inside element.contents() as the contents of the menu
-    var selectTemplate = '' +
+    var selectTemplate = angular.element(
       '<div class="md-select-menu-container">' +
         '<md-select-menu ' +
         (angular.isDefined(attr.multiple) ? 'multiple' : '') + '>' +
           element.html() +
-        '</md-select-menu></div>';
+        '</md-select-menu></div>');
 
     element.empty().append(labelEl);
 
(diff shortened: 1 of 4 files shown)


src/components/select/select.js
```diff
@@ -149,6 +154,30 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $interpolate) {
         }
       });
 
+      // Create a fake select to find out the label value
+      function setInitialLabelValue() {
+        if (ngModel && scope[attr.ngModel] !== undefined) {
```
ajoslin: `$parse(attr.ngMode)(scope)` for weird expressions like `ng-model="foo.bar"`

src/components/select/select.js
```diff
@@ -149,6 +154,30 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $interpolate) {
...
+          var fakeSelectEl = angular.element(selectTemplate).find('md-select-menu');
+          fakeSelectEl.data('$ngModelCtrl', ngModel);
+          fakeSelectEl = $compile(fakeSelectEl)(scope.$new());
+          var fakeSelectCtrl = fakeSelectEl.controller('mdSelectMenu');
+          var fakeSelectScope = fakeSelectEl.scope();
```
ajoslin: Don't use `.scope()`, it's broken when `$compileProvider.debugInfoEnabled(false)` is set, which many people do during production.

Let's just create a scope property on the controller.

src/components/select/select.js
```diff
@@ -160,15 +189,17 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $interpolate) {
 
       function openSelect() {
         scope.$evalAsync(function() {
+          var selectEl = selectTemplate.clone();
+          selectEl.find('md-select-menu').data('$ngModelCtrl', ngModel);
```
ajoslin: Just set `element.data('$ngModelController')` and the require will work as intended on line 216.

src/components/select/select.js
```diff
@@ -198,6 +227,7 @@ function SelectMenuDirective($parse, $mdUtil, $mdTheming) {
     $mdTheming(element);
     element.on('click', clickListener);
     element.on('keypress', keyListener);
+    ngModel = ngModel || element.data('$ngModelCtrl');
```
ajoslin: If you set `element.data('$ngModelController') this won't have to be done.

src/components/select/select.js
```diff
@@ -296,6 +326,15 @@ function SelectMenuDirective($parse, $mdUtil, $mdTheming) {
       }
     };
 
+    self.selectedLabels = function() {
+      var selectedOptionEls = Array.prototype.slice.call($element[0].querySelectorAll('md-option[selected]'));
+      if (selectedOptionEls.length) {
+        return selectedOptionEls.map(function(el) { return el.innerText; }).join(', ');
+      } else {
+        return undefined;
```
ajoslin: This should always return an array, don't like the two-types thing. The user of the API should check length.

ajoslin: Nevermind.

ajoslin: Instead, this should return empty string. So it's always a string at least (Empty string is falsy remember)

src/components/select/select.js
```diff
@@ -81,11 +81,11 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $interpolate) {
 
     // If not provided, we automatically make one
     if (!labelEl.length) {
-      labelEl = angular.element('<md-select-label>').html('<span>' + intStart + attr.ngModel + ' !== undefined ? ' + attr.ngModel + ': \'' + attr.placeholder + '\''+ intEnd + '</span>');
+      labelEl = angular.element('<md-select-label><span></span></md-select-label>');
     }
     labelEl.append('<span class="md-select-icon" aria-hidden="true"></span>');
     labelEl.addClass('md-select-label');
-    labelEl.addClass(intStart + attr.ngModel + ' ? \'\' : \'md-placeholder\'' + intEnd);
+    labelEl.addClass(intStart + attr.ngModel + ' !== undefined ? \'\' : \'md-placeholder\'' + intEnd);
```
ajoslin: Wee could use ng-bind so you don't have to intStart and intEnd.

```js
labelEl.attr('ng-bind', attr.ngModel + ' !== undefined ? "" : "md-placeholder"');
```

PR #1685, opened 2015-02-26
fix(autocomplete): trigger input event when clearing value

Toilal:
input event should be triggered when clearing value with (X) button.

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -298,6 +298,9 @@
     function clearValue () {
       $scope.searchText = '';
       select(-1);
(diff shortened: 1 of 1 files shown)


robertmesserle:
@Toilal Which event are you referring to?

Toilal:
When user is changing value of an input field, an `input` event is raised. It should be raised when clearing value with the clear button.

http://www.w3schools.com/jsref/event_oninput.asp

nlarche:
is there a way to catch this event in the parent controller ($scope.$on())? Shouldn't there be $scope.$emit() ? Or may be there is another way to now when the input is clearing in the parent controller ?

PR #1688, opened 2015-02-26
fix(button): Use md-raised colors for md-fab included in md-toolbar

### LINKED ISSUE #1687: md-fab button in md-toolbar has white background
I thing there's a problem with styling in [button-theme.scss](https://github.com/angular/material/blob/04feeb836ae5508d4c0349f125d13b75dd63b7b3/src/components/button/button-theme.scss#L4-L6). It cause md-fab button embeded in md-toolbar to have white background, regardless the theme.

[Plunkr showing the issue](http://plnkr.co/edit/2Jj73NUPERrih2siCuwM?p=preview)

Toilal:
Instead of using a white background, use the same styles as md-raised buttons for md-fab buttons included in md-toolbar.

It allows md-fab buttons included in md-toolbar to support all rendering effects properly.

Close #1687

### DIFF
--- a/src/components/button/button-theme.scss
+++ b/src/components/button/button-theme.scss
@@ -1,8 +1,15 @@
 $button-border-radius: 3px !default;
 $button-fab-border-radius: 50% !default;
 
-md-toolbar .md-button.md-THEME_NAME-theme.md-fab {
-  background-color: white;
+.md-button.md-THEME_NAME-theme.md-fab {
+  color: '{{background-contrast}}';
+  background-color: '{{background-50}}';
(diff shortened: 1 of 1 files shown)


Toilal:
I'm not expert in SCSS, but maybe it could be included [here](https://github.com/Toilal/material/blob/master/src/components/button/button-theme.scss#L45-L54) using a parent SCSS selector, to avoid duplicate code ...

ThomasBurleson:
Will not merge. See existing fixes and features for md-fab and toolbar.

PR #1709, opened 2015-02-27
feat(script): Set ngModule to Closure namespace.

wesalvaro:
When using Closure, it is common to reference Angular module dependencies by the name property of the module.
This looks like:
```
goog.provide('bar.module');
goog.require('foo.module');
bar.module = angular.module('bar.module', [foo.module.name]);
```
This change makes it possible for this kind of reference instead of having to refer to the dependent module as a string.


jelbourn:
Look good to me.

PR #1724, opened 2015-02-28
docs(): improved keyboard/focus management

### LINKED ISSUE #1723: docs(): focus sent to main content area on page load
It's a bad keyboard and screen reader experience to have focus sent to `[role=main]` automatically on the docs website on page load. This functionality was added to redirect focus on click of a docs sidenav link, but is happening on initial page load as well. Focus is currently triggered on route change without a hook on the links to differentiate clicks from page loads. One idea to fix: add a click handler to those links and set some kind of flag telling the `openPage` function to only redirect focus if the flag is true, and reset it to false. Open to other ideas.

Also adding a skip-to-main-content link as part of this work.

marcysutton:
With this change, focus is only sent to the main content area on click of a nav link. There is also a "Skip to Content" link that allows users to skip past the sidebar navigation.

Skip to content link

Closes #1723

### DIFF
--- a/docs/app/css/style.css
+++ b/docs/app/css/style.css
@@ -51,6 +51,31 @@ ul li:first-child {
   margin-top: 0;
 }
 
+ul.skip-links li {
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+ul.skip-links li a {
+  background-color: #fff;
+  display: block;
+  font-size: 1.25em;
+  margin: 0.5em 0 0.5em 0.5em;
+  opacity: 0;
+  left: 0;
+  padding: 0.75em 0 0.75em 0.5em;
+  position: absolute;
+  top: 0;
+  width: 96%;
+  -webkit-transition: opacity   0.15s linear;
+  -moz-transition: opacity  0.15s linear;
+  transition: opacity   0.15s linear;
+}
+ul.skip-links li a:focus {
+  opacity: 1;
+  z-index: 2;
+}
+
 pre {
   white-space: pre;
   white-space: pre-wrap;
--- a/docs/app/js/app.js
+++ b/docs/app/js/app.js
@@ -287,6 +287,12 @@ function(COMPONENTS, DEMOS, PAGES, $location, $rootScope) {
       $scope.isSelected = function() {
         return controller.isSelected($scope.section);
       };
+
+      $scope.focusSection = function() {
+        // set flag to be used later when
+        // $locationChangeSuccess calls openPage()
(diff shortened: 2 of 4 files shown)


docs/app/css/style.css
```diff
@@ -51,6 +51,31 @@ ul li:first-child {
...
+  list-style: none;
+  margin: 0;
+  padding: 0;
+}
+ul.skip-links li a {
+  background-color: #fff;
+  display: block;
+  font-size: 1.25em;
+  margin: 0.5em;
+  opacity: 0;
```
ThomasBurleson: So by using opacity:0, it is not visible... but still rendered. 
Would `display:none;` be a better choice?

docs/app/js/app.js
```diff
@@ -287,6 +287,10 @@ function(COMPONENTS, DEMOS, PAGES, $location, $rootScope) {
       $scope.isSelected = function() {
         return controller.isSelected($scope.section);
       };
+
+      $scope.itemClicked = function() {
+        controller.navItemClicked = true;
```
ThomasBurleson: Instead of `navItemClicked`, I recommend that we call this variable `autoFocusContent`; since we do not care **how** the value was set to true, but rather **why** the variable is used.

docs/app/partials/menu-link.tmpl.html
```diff
@@ -1,5 +1,5 @@
 <md-button ng-class="{'active' : isSelected()}"
-  ng-href="#{{section.url}}">
+  ng-href="#{{section.url}}" ng-click="itemClicked()">
```
ThomasBurleson: I assume that `ng-click` has higher priority that `ng-href`?

docs/app/css/style.css
marcysutton: The pattern for skip links is to have them always render but show visibly on focus. That way they are in the page for screen reader users at all times.

docs/app/js/app.js
marcysutton: Sounds good!

docs/app/partials/menu-link.tmpl.html
marcysutton: We can name it whatever we want. The purpose of `ng-click` here was to provide a hook differentiating link clicks from route changes. There wasn't a way to do that with `ng-href`.

marcysutton:
@ThomasBurleson I incorporated your changes and fixed a horizontal scrollbar issue in the sidenav. Going to merge to master and we can tweak it later if needed.

PR #1752, opened 2015-03-02
Docs toolbar buttons: Version/SHA and Skip link

marcysutton:
@ThomasBurleson I added a version button to the docs toolbar on the Home and Getting Started pages. There is a bit of a flash when you go from either of those pages to a Demo page, probably because there wasn't an easy way to target both without a complex `ng-if`. If you have suggestions on how to improve performance while still having it show up in both places, I'm all ears.

Font size changed for `md-button` inside of `md-toolbar` from 14px to 16px to match the new global type style.

I also converted the Skip to Content link to an md-button on @robertmesserle's suggestion. 

new SHA button in Angular Material toolbar

### DIFF
--- a/docs/app/css/style.css
+++ b/docs/app/css/style.css
@@ -59,19 +59,19 @@ ul.skip-links li {
 ul.skip-links li a {
   background-color: #fff;
   display: block;
-  font-size: 1.25em;
   margin: 0.5em 0 0.5em 0.5em;
   opacity: 0;
   left: 0;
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
@marcysutton - I will review later today. Thx for this PR.

ilanbiala:
We can probably remove the text at the bottom stating the SHA and time and move the time to somewhere else? Also it would be nice if the time were ET or PT or `moment()`'ed to be localized.

PR #1759, opened 2015-03-03
fix(dialog): correct opening position for showing a dialog

### LINKED ISSUE #1401: mdDialog: md-dialog-container problem on IE10
Normally, dialog container has CSS `top` parameter when you scroll down page, eg.:

    <div class="md-dialog-container ng-scope" style="top: 390px;">

On IE10, `top` is always set to `0`, so when you scroll page, you cannot see dialog window. After quick search, it looks problem is [here](https://github.com/angular/material/blob/master/src/components/dialog/dialog.js#L332).  `$("body").prop("scrollTop");` returns `0` on IE10.

yarl:
fixes fix #1401

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -407,7 +407,7 @@ function MdDialogProvider($$interimElementProvider) {
       if (options.hasBackdrop) {
         // Fix for IE 10
         var computeFrom = (options.parent[0] == $document[0].body && $document[0].documentElement 
-                           && $document[0].scrollTop) ? angular.element($document[0].documentElement) : options.parent;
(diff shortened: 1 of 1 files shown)


mramanauskas:
:+1:

kschwidder:
Checked the code by copying the change into the current build. It is fixed !!!
BUT the backdrop is not properly painted :-) Does not matter for me. When can you release the fix ?

marcysutton:
@kschwidder what do you mean by "backdrop is not properly painted"? Is that in a specific browser or OS?

kschwidder:
When dialog is opened it looks like this
backdrop ok when not scolled

when browser has been scrolled and dialog is opened it looks this this
backdrop wrong after scrolling

PR #1768, opened 2015-03-03
fix(docs: Added NPM Install to build instructions)

hanford:
You must "npm install" to have all of the correct dependencies as well as bower installing


hanford:
Signed CLA with according GitHub name

PR #1784, opened 2015-03-04
fix(removed duplicate rule for button element)

mobelio:


ilanbiala:
Why was this closed?

marcysutton:
@ilanbiala it won't be applicable any more after a CSS refactor, open in #1985. There was also no CLA.

PR #1838, opened 2015-03-09
fix(icon):  fix fill attribute for svg icons who do have a fill attribute in their path object

clue42:
For svg icons having a fill attribute in their path object the current
color settings do not work. To override this, the setting needs to be
applied directly to the path fill attribute, rather than just to
md-icon.

### DIFF
--- a/src/components/icon/icon.scss
+++ b/src/components/icon/icon.scss
@@ -8,6 +8,10 @@ md-icon {
   width: 3 * $baseline-grid;
 }
 
+md-icon path {
+    fill: currentcolor;
+}
+
 //
 //@font-face {
 //  font-family:"material";


clue42:
CLA signed and submitted

clue42:
Yes, I'm the author of this pull request.

clue42:
Yes, I'm the author of this commit and pull request on top of that.

clue42:
CLA signed and submitted

ThomasBurleson:
@clue42 - What about scenarios where where the fill is none and we have a custom style set.
Consider the [online input demos](https://material.angularjs.org/#/demo/material.components.input):
The last demo has icons associated with the inputs. 

Your changes will completely break the rendering of those icons.

clue42:
Hi Thomas,

that's right. The question is however, how do we deal with svg-icons having a fill attribute set. I stumbled across the issue while trying to use the icomoon svg-icons which do have a fill attribute set (at least the ones I have been using). Many svg editors are setting the fill attribute from the stroke color. In that case the online input demo would not work either.

Pajn:
Can't `md-icon path {` be changed to `md-icon[color-fill] path {` and thus allowing the user of `<md-icon>` to opt-in to this styling when it's needed?

ThomasBurleson:
Will not merge.

Issue #1842, opened 2015-03-09
ngClick and checkboxes receiving double clicks on touch devices when jquery is included


FrankMeyer:
md-checkbox and ng-click are behaving strange on touch-devices when jQuery is included.
Without jQuery all works fine.
Here is an example with angularJS 1.3.14 and angular-material 0.8.3:
http://codepen.io/anon/pen/qEMJXo
The Button clicks will count twice and the md-checkbox is barely checkable.
Testet on Galaxy S4 with android standard browser and on a Motorola Tablet with Chrome and Firefox.
I also tried to include ngTouch,hammer.js and jQuery-mobile. Neither helped.
It seems that angular elements receive double clicks all the time. 

Dropping jQuery is not an option, as it will be used widely in the underlying project.
Anyone have a clue why jquery brings such behaviour?
Or am I missing something?

MilosStanic:
Hey, this is the same issue I'm having on desktop Windows Chrome with mdSidenav. When I click open the button gets clicked twice so my sidenav closes immediately after opening. Didn't try without JQuery.

ducva:
I got same issue, and fixed by removing window.jQuery at line 1058 in angular-material.js
<code>
if (window.jQuery || isKeyClick || ev.$material) return;
</code>

MilosStanic:
I can also confirm doubleclicks happening on buttons in mdSidenav and mdBottomSheet.

ThomasBurleson:
Closed with SHA 81bcf7f

gharel:
I have the same problem on : md-sidenav and md-switch.
Like @mathiasmoeller said maybe the same problem than #1841.
I use jQuery UI and it need jQuery to be load before Angular, maybe this could help.
Thanks.

kennethcachia:
@gharel @mathiasmoeller 

Can you test http://codepen.io/kennethcachia/pen/dPBPxq and let me know if you're still experiencing this issue?

It works for me on both Android and iOS and I want to confirm if it's resolved or if it's a possible conflict with ionic.

Thanks!

ThomasBurleson:
Fixed in 2de0cf26a4e7fc985b6247039537311a7e7861c1

gharel:
It works for me on iOS. Thank you.
So it solved the problem for md-switch, what about md-sidenav ?
Thanks in advance.

mathiasmoeller:
Did not have the time to test it completely but it seems to work! Thank you very much!

PR #1869, opened 2015-03-11
fix(gesture): don't stop hijacking clicks on ios&android when jquery is loaded

FrankMeyer:
Remove window.jQuery from check for hijacking clicks on mobile devices.
	Touch clicks were emitted twice when jQuery was loaded which made checkboxes and buttons unusable on mobile devices.

	Addressing Issue https://github.com/angular/material/issues/1842 Utilizing proposed fix


MilosStanic:
+1 to merge this asap! Totally breaking android usability. Affects also non-angular-material components.

PR #1871, opened 2015-03-11
fix(md-select): Track disabled state

troch:
Track the component disabled state in order to avoid a double
    registration of click and key events on the md-select element in the
    disabled $observe listener.


rschmukler:
Added with dc93bffe25cd5b2c28e6eac7adfe954c69c672c6. Thank you!

Novashock:
@troch Thanks for this!

PR #1887, opened 2015-03-12
update peerDependencies to permit beta

### LINKED ISSUE #1848: Fix peerDependency
Thanks for using `peerDependencies`. That's awesome. However, when I try to install this via `npm` with the beta version of angular, it fails the `peerDependencies`. I actually don't know of a good solution to this as I'm also having issues with it in [my library](https://github.com/formly-js/angular-formly)... Would love it if we could come up with a solution.

linclark:
Fixes #1848 

This allows users to use the package with the beta version of Angular 1.4.0.

I have a version of the bower-material package up to demo this. You can get it by running:

```
npm install --save git://github.com/linclark/bower-material.git#semver-beta-test
```

Then you should be able to run the following without errors:

```
npm install angular@beta
```

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -7,7 +7,7 @@
     "dist/angular-material.css"
   ],
   "dependencies": {
-    "angular": "^1.3.0",
+    "angular": "^1.3.0 || >1.4.0-beta.0",
     "angular-animate": "^1.3.0",
     "angular-aria": "^1.3.0"
   },


ThomasBurleson:
See comments @ https://github.com/angular/material/issues/1848#issuecomment-78599035

kentcdodds:
@linclark I think you meant to make the changes to the `package.json` file, not the `bower.json` file. Right? `bower.json` doesn't have a concept of `peerDependencies`

ThomasBurleson:
@kentcdodds  If you modify the npm dependencies:

*  angular/material does **not** need this because those packages are for builds not deployment
* angular/bower-material should **not** have these because we do not want to explicitly support unreleased builds.

Does that make sense? Would you and @linclark  agree ?

kentcdodds:
I think I may be misunderstanding you, but this is what I'm experiencing:

```
~/Desktop
🐑  $ npm install angular@1.4.0-beta.5
angular@1.4.0-beta.5 node_modules/angular
~/Desktop
🐑  $ npm install angular-material
npm ERR! Darwin 14.1.0
npm ERR! argv "node" "/usr/local/bin/npm" "install" "angular-material"
npm ERR! node v1.0.2
npm ERR! npm  v2.2.0
npm ERR! code EPEERINVALID

npm ERR! peerinvalid The package angular does not satisfy its siblings' peerDependencies requirements!
npm ERR! peerinvalid Peer angular-material@0.8.3 wants angular@1.3.x

npm ERR! Please include the following file with any support request:
npm ERR!     /Users/kentcdodds/Desktop/npm-debug.log
```

I want to move an app over to use angular 1.4 right now, but I'm unable to do so because of this issue.

(and yes, those are sheep in my terminal).

linclark:
@kentcdodds I responded in #1848, but I'm pretty sure the way the Angular projects do it is to copy dependencies from `bower.json` to `package.json` as part of the release process.

@ThomasBurleson I responded in #1848. Currently, anyone trying to use `angular@beta` with `angular-material` will get the error that @kentcdodds posted.

jgoux:
This fix is perfect for my NPM installation ! :+1:

ThomasBurleson:
@linclark - 

* both `bower.json` and `package.json` in **bower-material** are updated with version information. 
* only `bower-material/bower.json` is updated with the dependencies defined in `material/bower.json`. 

So I think we need to update the semver information in [bower-material/package.json](https://github.com/angular/bower-material/blob/master/package.json#L23-L25):

```json
"peerDependencies": {
      "angular": "^1.3.0 || >1.4.0-beta.0",
      "angular-animate": "^1.3.0 || >1.4.0-beta.0",
      "angular-aria": "^1.3.15 || >1.4.0-beta.0"
  }
```

kentcdodds:
Don't if it matters at this point, but I think what that version specifies
excludes the first beta version... Needs `>=`

- Kent C. Dodds

(Sent from my mobile device, please forgive typos or brevity)

ThomasBurleson:
@kentcdodds - true and true. Thx for the review note.

Issue #1893, opened 2015-03-13
mdSelect: accessibility label lost when item is selected


marcysutton:
Because of the way `mdSelect` dynamically binds `aria-labelledby` to the `<md-select-label>`, when an option is selected by the user or programmatically by default, there is no descriptive label telling a screen reader user what the select is for. See the "select label test" portion of this demo: http://codepen.io/marcysutton/full/ZYqjPj/

In VoiceOver, the native select is announced as "Animal Muppets pop-up button", combining the first option and the label of the component. The mdSelect component should announce something similar.

Issue #918, opened 2014-12-11
Provide `versioning` of Online Docs


ThomasBurleson:
Similar to doc versioning in AngularJS 1.x:

![doc_versioning](https://cloud.githubusercontent.com/assets/210413/5399741/8656fed4-8133-11e4-905e-ab1f0edbfa78.png)

> Will need to use the Select menu also planned for 0.8 Milestone release.

epelc:
@ThomasBurleson If we decide to implement #1103  then I think [hyperboot](https://github.com/substack/hyperboot) would go really well with it.

Hyperboot allows you to save several versions of a webapp for offline viewing with the help of appcache. So if you view a version of the docs while you are online it will be available when you are offline. You can also add js to download all versions and to set defaults for which version to load.

marcysutton:
Discussed: 
1. How far back to include picker: old versions of docs will have different links, structure, etc.
1. UI for picker: select? dropdown? archive page?
1. Where to host snapshots? Angularjs snapshots hosted on code.angularjs.org
1. Picker should include `master` but default to latest tag

epelc:
@Marcus tutor did you take a look at hyperboot by any chance? It seems like a really good fit for this + you get offline support.

marcysutton:
@epelc I'll assume "Marcus tutor" was supposed to be me? :stuck_out_tongue_winking_eye: 

We haven't yet, but that's a great idea. I'll have a look.

epelc:
@marcystutton yes I think auto correct got it. Andoird is so aggressive with its spell checking.

ThomasBurleson:
Feature Summary:  

*  Lock the documentation with its associated release. 
*  Provide a version picker to choose the associated documentation with download zip link.
*  Continue provide development documentation linked to current #master development branch.

PR #1904, opened 2015-03-13
docs(versions): provide version menu for navigating to versioned docs

radamant:
This was created in collaboration with @marcysutton.

Refs #918 provides version menu (with customizable minimum-version cutoff)
that navigates to docs based on a simple string format.

This approach was based on that provided in the base angular.js repo,
and some of the code was ported from that repo.

It looks like this:

screenshot 2015-03-13 15 05 24

screenshot 2015-03-13 14 26 39

As aforementioned, this works pretty much identically to the one in the angular.js repo, using much of the same code. It gets previous version from the list of tags, and adds a CDN and snapshotted version if available. The doc url generation is controlled by two variables, one to determine the minimum version for which doc urls will be generated, the other to format of the URL for versioned docs (none of the URLs currently exist.)

I do not have access to either material's public-facing servers nor its CI configuration(s), so there are still several items that must be done before this can be rolled out:
- determine where snapshotted versions will be hosted, angular currently uses https://code.angularjs.org/:version:/ which is why this PR uses https://code.angularjs.org/material/:version/.
- provide documentation for older versions (or just set the min-version to >=0.8.3 and start with the current version)
- add the version processor, controller and template to the docs for any older versions that should go in the version list
- test with a real snapshotted version (I have only tested with a "-local" version on my machine)
- @marcysutton to work on fixing the accessibility for the labels of md-select -> #1893

### DIFF
--- a/config/version-info.js
+++ b/config/version-info.js
@@ -1,0 +1,205 @@
+'use strict';
+
+var fs = require('fs');
+var path = require('path');
+var shell = require('shelljs');
+var semver = require('semver');
+var _ = require('lodash');
+
+var currentPackage, previousVersions, cdnVersion, gitRepoInfo;
+
+
+/**
+ * Load information about this project from the package.json
+ * @return {Object} The package information
+ */
+var getPackage = function() {
+  // Search up the folder hierarchy for the first package.json
+  var packageFolder = path.resolve('.');
+  while (!fs.existsSync(path.join(packageFolder, 'package.json'))) {
+    var parent = path.dirname(packageFolder);
+    if (parent === packageFolder) { break; }
+    packageFolder = parent;
+  }
+  return JSON.parse(fs.readFileSync(path.join(packageFolder,'package.json'), 'UTF-8'));
+};
+
+
+/**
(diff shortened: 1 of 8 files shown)


radamant:
@googlebot CLA is signed.

ThomasBurleson:
@marcysutton - this PR is incomplete due to dependent features:

- determine where snapshotted versions will be hosted, angular currently uses https://code.angularjs.org/:version:/ which is why this PR uses https://code.angularjs.org/material/:version/.
- provide documentation for older versions (or just set the min-version to >=0.8.3 and start with the current version)
- add the version processor, controller and template to the docs for any older versions that should go in the version list
- test with a real snapshotted version (I have only tested with a "-local" version on my machine)
- @marcysutton to work on fixing the accessibility for the labels of md-select -> #1893 

@robertmesserle - can you and I work with BrianF to prepare these above items for 0.9 release?

ThomasBurleson:
Closed. See SHA 57e9ea47698c3ca517246a9c8b84d382ad196bd0

PR #1931, opened 2015-03-17
fix(button): Support ui-sref attribute natively.

### LINKED ISSUE #1922: mdButton does not play nicely with uiSref
`mdButton` transforms into an `<a>` tag if the `href` or `ng-href` attribute is present. This doesn't work for third party directives which adds an `href` attribute dynamically, such as [`uiSref`](https://github.com/angular-ui/ui-router/wiki/Quick-Reference#ui-sref) from `angular-ui-router`.

A workaround is to add an empty `href` attribute to the element.

ruiaraujo:
closes #1922.

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -60,7 +60,7 @@ function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria) {
   };
 
   function isAnchor(attr) {
-    return angular.isDefined(attr.href) || angular.isDefined(attr.ngHref);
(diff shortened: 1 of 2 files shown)


Frank3K:
Nice one. Faced the same problem and chose to wrap the element in an `a`-tag as a workaround. This is a good solution.

gustavohenke:
Nice and simple patch!

marcysutton:
Closed with 7b743ed4c62e2680c350d56daf57b202f15e63bf.

PR #1936, opened 2015-03-17
fix(dialog): fix backdrop position

yarl:
After recent fixes by @marcysutton backdrop is not following alert (tested on Chrome and IE10). 
Screen

This commit should fix it.

PR #1937, opened 2015-03-17
fix(backdrop): fix backdrop for dialog

yarl:
CSS backdrop fix for #1936.


marcysutton:
Can this go in the same PR as #1936?

PR #1944, opened 2015-03-17
feat(tabs): refactors tabs to function closer to spec

### LINKED ISSUE #1087: Scrollbar appears when right paginator for tabs is visible.
Check the [Plunker](http://embed.plnkr.co/8kzEkzAAbqABNx6kAgyv/preview).

If you reduce the browser width which enables the paginators for the tabs. A scrollbar shows up on the bottom whenever the right paginator is visible.

I've fixed it using this code.

```css
.md-paginator{
    margin-right:0px;
}
```

### LINKED ISSUE #1107: mdTabs inside ng-repeat not working properly
Well i tried to put some tabs inside ng-repeat, i tried many combinations and finally understood this is a bug.
I can't point where the problem occurs but here's an [example](http://embed.plnkr.co/xfCQCUL4tNSJo8axaPfg/) of whats happening.

thanks ahead :)

### LINKED ISSUE #1140: md-tabs: add refactored tabs code
Key improvements:
- Consolidate code into a single manageable controller
- Reduce the amount of code contained inside directives
- Remove communication between directives and controllers via `$broadcast`
- Rely more heavily on data-bindings and do less DOM manipulation manually
- Consolidate all tab logic to a single watcher on `selectedIndex`
- Remove performance hit of repeatedly restyling the tabs to calculate ink bar position by using a hidden copy of the tabs for all calculations
- Remove complex logic required to handle pagination on keyboard tabbing by using hidden tabs for keyboard interaction (ie. pagination can happen immediately on focus, rather than waiting for pagination to complete before allowing focus)

### LINKED ISSUE #1247: mdTabs: Should the the pagination arrows change tabs?
On a mobile display, if tab labels are long, it is pretty common to display only one tab at a time:
image
In that case, I have found disturbing that clicking on the arrow doesn't actually select the next (or previous) tab, but only changes the pagination. I understand the goal of this feature, however I wonder if it is a strict design requirement. The [design specification](http://www.google.com/design/spec/components/tabs.html) states:

> Scrollable tabs display a subset of tabs at any given moment, and can contain longer tab labels and a larger number of tabs. They are best used for browsing contexts in touch interfaces when users don’t need to directly compare the tab labels.

> To navigate between scrollable tabs, touch the tab, swipe the tabs left or right, or swipe the content area left or right.

It doesn't mention using the arrow buttons to navigate between tab pagination. In that case, can it be an option? `md-pagination-arrows="paginate"` or `md-pagination-arrows="changeTab"` for instance?

Can anyone shed some light on this issue?

### LINKED ISSUE #1261: mdTabs: with inner content, first swipe fails
In an `<md-tabs>` with inner tab content, the first swipe to change tabs doesn't work. However, if we click on another tab, and then try to swipe, it works.

untitled

I was able to pinpoint the bug to 3ee83a645b9e4da8f4c0f2e6cbf772f504d8e9a9.

Here is the stacktrace:
image

Maybe #1257 will fix this bug, but it would be nice to check anyway…

### LINKED ISSUE #1380: Tab animation falls apart when switching quickly between 3+ tabs.
On OSX with chrome 41.0.2272.35 beta (64-bit), clicking on tabs in the demo
https://material.angularjs.org/#/demo/material.components.tabs

Tab animations falls apart if you switch too quickly between 3+ tabs. To reproduce, click on tab "One" then tab "Two" then tab "Three" quickly one after the other. You'll see the tab body jump around on the screen. Transitioning back and forth between two tabs quickly seems to be ok.

### LINKED ISSUE #1387: tabs scroll with content
See plunker http://plnkr.co/edit/gpawIEjFSx8AeaBA3KyU?p=preview 

For the two fist cards there are tabs above scrolling lists. For both cards the tabs scroll away

Expected behaviour is shown in the bottom card where both toolbar and subheader stays put

### LINKED ISSUE #1403: md-tabs Does not work on mobile using touch, works on desktop
Nothing I do can get md-tabs to work on mobile, works fine in browser.

I can't remember what the last version I used where it works on mobile, I'd like to say it was 0.7.0-rc2...

### LINKED ISSUE #1443: md-tabs weird pop in
The demo shows a weird pop in on the tabs component, probably a messed up css transition... Here is a [demo video](http://files.slingingcode.com/200S3A2r3h0B)

### LINKED ISSUE #1505: md-tabs: First tab cutoff after window resize with `md-stretch-tabs="always"`
If your `md-tabs` has `md-stretch-tabs="always"` then if load the page with a big window and resize it to a small window the first tab(possibly more) becomes hidden. Resizing the page a second time fixes this though.

Heres a few pictures to demonstrate this.
image

### LINKED ISSUE #1506: md-tabs: Scrollbar after making window small
If you load the page with a wide window then resize it down to a small window the tabs are showing a scroll bar.

image

Tested on the docs demo and in my app with `0.8.0-rc1`
https://material.angularjs.org/#/demo/material.components.tabs

### LINKED ISSUE #1516: Swipe gesture does not work anymore on tab control
The swipe gesture does not change tabs on the dynamic tabs demo. 
I've also recently upgraded to 0.8.0 from 0.7.0 and noticed the swipe gesture is not changing tabs anymore.

The static tabs demo swipe gesture is changing tabs however.

### LINKED ISSUE #1518: IE 11 Flicker when switching tabs in demo's
https://material.angularjs.org/#/demo/material.components.tabs

Every time you click a new tab the window seems to resize briefly causing a quick flicker at the end of the slide transition.

I did also find that in general IE11 suffers from performance issues (i'm seeing this in my work) as well as the demo's.  For example when on the link above clicking on the Button demo causes a very severe lag in the css animations.  Firefox and Chrome don't have either of these issues.

### LINKED ISSUE #1564: mdTabs: pagination arrows invisible when using md-primary
Relevant issue: #1464.

As the arrows are using the primary color, when using `class="md-primary"` for tabs, the pagination arrows become invisible. See for instance this pen, which has 6 tabs: http://codepen.io/PaulMougel/pen/PwRbgQ

### LINKED ISSUE #1570: md-tabs: Single tab w/ stretch is not centered
`<md-tabs md-stretch-tabs="always">` does not center a single-tab correctly (see that last example below, which should have the label "one" directly below the title):
dy2jkvpzwk 2
_(#1506 has the scrollbar issue)_

Looks like it has a `width` style _that is correct_ but also a `min-width` that keeps it shorter. Here's a live jsbin: http://jsbin.com/guweto/2/

### LINKED ISSUE #1620: md-tabs: a lot of watchers generated
md-tabs component creates a lot of watchers on window resize when it's hidden.
Material v0.7.1 and v0.8.0-rc1-master-ffbea7d
All browsers are affected.
Test: http://codepen.io/tedbeer/pen/bNvMmw
The test uses ng-stats to show number of watchers. You can also inspect $$watchers in the md-tabs scope to see it's growing.

Proposed fix:
```
*** /MaterialDesign/src/components/tabs/js/paginationDirective.js
***************
*** 65,73 ****
        onTabFocus(newTab);
      }
  
      function updatePagination() {
        if (!element.prop('offsetParent')) {
!         var watcher = waitForVisible();
          return;
        }
  
--- 65,74 ----
        onTabFocus(newTab);
      }
  
+     var watcher = null;
      function updatePagination() {
        if (!element.prop('offsetParent')) {
!         watcher = waitForVisible();
          return;
        }
```

### LINKED ISSUE #1626: Single tab in material design
Hi,

I'm working on a tabs container that starts as a single tab and as you keep interacting with the application, you keep adding more. Problem is that when you have only one tab, the ink bar does not show up. Is that by design? I could not find reference to this case in the material guidance for when you'd use dynamic tabs. The code in Angular Material checks the following:
```
var hideInkBar = !selected || tabsCtrl.count() < 2 || mdNoBar;
```
Should this behavior be at least configurable?

Thanks
-Gabriel

### LINKED ISSUE #1698: bug(tabs): disconnect causing issues
As posted by @mxdubois:

>This causes a problem for me where the scope is disconnected after it is reconnected by onSelect during the initial load. The bug results in ng-model bindings in the initially selected tab not working until the user switches to another tab and switches back.

>In what case is the $timeout necessary? If we need to keep the $timeout, we should save the promise returned by $timeout and cancel it in onSelect where we call reconnectScope. If that sounds good, I can submit a PR.

Link: https://github.com/angular/material/commit/78ba497e443ca31e8a8c97f11db281f743f6aca0#commitcomment-9946067

### LINKED ISSUE #1777: Upon tab reapearing selection is not taken into account
First of all, I never mentioned how I LOVE this angular-material project.
Here's the codepen example:
http://codepen.io/yccteam/pen/GgGbzO
The issue is this:
1) Tabs are created and work fine.
2) When I click on "Change IF" the tabs are supposed to disappear due to ```ng-if``` - works great.
The problem is:
3) When  "Change IF" is clicked again, the ``ng-if``` is satisfied and shows the tabs again. but ```md-select``` isn't taken into account once the tabs reappear.

I've also added the "ng-show" for sports (just click  "Change SHOW"  to play with it).
It works, but now I need to take care of the ```md-bar```, which is external to the tab itself.

I currently sove this using a hack with a watcher that sorts things out, but using a watcher while I already have a watcher (```ng-if```) seems very... hackish...

Is this behavior intentional? Any way to solve this within the API?

Added later:
I've added another functionality that recreates a bug in the tabs:
1) Click on the "Change SHOW" in order to make the tabs to disappear.
2) Then click on the "Change Tab" until the tab index changes (the number below the button)
3) Click on "Change SHOW" in order to make the tabs reappear.
4) Note the tabs appear wrong

### LINKED ISSUE #1788: md-tabs fail when used with ng-include
I have a page that adds a number of templates using ng-include. One of the templates includes an md-tabs component.

The tabs are not switching correctly, when selecting the second tab, the first tab remains visible. I have tried the template outside of an ng-include and it works as to should do.

### LINKED ISSUE #1850: md-tab compiled() and then digest() throws an error in unit tests
Hi guys,

I'm using Material Design in my app and now I'm writing unit tests using Jasmine & Karma. 
When I try to compile and digest a `md-tab` component, it throws an error: 

`'undefined' is not a function (evaluating 'element.addClass.bind(element, 'md-tab-themed')')`

Directive example: (I voluntary removed tab contents)
```javascript
'use strict';

angular.module('ria.dev').directive('myDirective', function () {
  return {
    template: '<md-content style="height: 600px;">' +
      '<md-tabs md-selected="selectedIndex">' +
        '<md-tab label="hello"></md-tab>' +
      '</md-tabs>' +
  '</md-content>',
    restrict: 'E',
    controller: function($scope, $element) {
      $scope.selectedIndex = 0;
    }
  };
});
``` 

The Jasmine unit test is:

```javascript
describe('MyDirective', function() {
  var $compile, $rootScope;

  beforeEach(module('ria.dev'));
  beforeEach(inject(function(_$compile_, _$rootScope_){
    $compile = _$compile_;
    $rootScope = _$rootScope_;
  }));

  describe('Fake Unit Test', function() {    
    beforeEach(function() {
      element = $compile(angular.element('<my-directive></my-directive>'))($rootScope);
      $rootScope.$digest();
      // the digest() method fails
    });
    
    it('should be true', function() {
      expect(true).toBe(true);
    });
  });
});
```

The error is: (this is the `bind` method that's undefined)
```
	TypeError: 'undefined' is not a function (evaluating 'element.addClass.bind(element, 'md-tab-themed')')
	    at postLink (/xxx/bower_components/angular-material/angular-material.js:11071)
	    at /xxx/bower_components/angular/angular.js:8213
```

And the corresponding line is:
```javascript
$timeout(element.addClass.bind(element, 'md-tab-themed'), 0, false);
````

When I comment this line, everything goes better.

I use:
* Angular.js v1.3.14
* Angular-Material v0.8.3

Does anyone have an idea? :-)

PS: It works correctly in the browser.

### LINKED ISSUE #1959: Paginated tabs missing arrow icons
Hello,

I am having an issue with tabs that no longer fit the parent width and become paginated. The issue is that the left/right buttons do not show icons or any sort of outline and as such are basically invisible. 

The image below illustrates the issue. This is a group of about 10 tabs. Here the screen is shrunk to show only three, however, the left/right buttons are missing their icons.
capture

As I adjust the screen size, I receive the following 3 errors from angular / material. Note, I only receive these errors on pages with paginated tabs. 
* angular.js:11607 TypeError: Cannot read property 'tagName' of undefined
    at Object.Icon (angular-material.js:5926)
    at angular-material.js:5851
    at angular.js:13189
    at l.$get.l.$eval (angular.js:14401)
    at l.$get.l.$digest (angular.js:14217)
    at l.$get.l.$apply (angular.js:14506)
    at l (angular.js:9659)
    at S (angular.js:9849)
    at XMLHttpRequest.D.onload (angular.js:9790)angular.js:11607 (anonymous function)angular.js:8557 $getangular.js:13197 (anonymous function)angular.js:14401 $get.l.$evalangular.js:14217 $get.l.$digestangular.js:14506 $get.l.$applyangular.js:9659 langular.js:9849 Sangular.js:9790 D.onload

* angular.js:11607 TypeError: undefined is not a function
    at loadFromIconSet (angular-material.js:5860)
    at angular.js:13189
    at l.$get.l.$eval (angular.js:14401)
    at l.$get.l.$digest (angular.js:14217)
    at l.$get.l.$apply (angular.js:14506)
    at l (angular.js:9659)
    at S (angular.js:9849)
    at XMLHttpRequest.D.onload (angular.js:9790)angular.js:11607 (anonymous function)angular.js:8557 $getangular.js:13197 (anonymous function)angular.js:14401 $get.l.$evalangular.js:14217 $get.l.$digestangular.js:14506 $get.l.$applyangular.js:9659 langular.js:9849 Sangular.js:9790 D.onload

* angular.js:11607 undefined is not a function

I hope I've described the issue properly and that someone can either explain what I'm doing wrong, or get the code corrected. Please let me know if I can provide further feedback.

### LINKED ISSUE #1986: Potential Memory Leak With Inactive md-tab
In relation to #1982, I'm including code in my directives' controllers that prints out when controllers are destroyed.

For example:
```javascript
       $scope.$on('$destroy', function () {
		$log.debug('Destroying: file-grid-ctrl');

		// Add any cleanup/tear-down logic here
	});
```

Based on my exploration of the code and the documentation, internal **non-active** tab content is removed from the `$scope\$digest` cycle:

https://material.angularjs.org/#/api/material.components.tabs/directive/mdTabs:
```javascript
As a performance bonus, if the tab content is managed internally then the non-active (non-visible) tab
contents are temporarily disconnected from the $scope.$digest() processes; which restricts and optimizes
DOM updates to only the currently active tab.
```

Since Material Design detaches inactive Tab content from the $scope/$digest cycle, **none** of my controllers for all my "inactive" tabs print out `Destorying: file-grid-ctrl`.  This leads me to believe that any resources or other things I may be "destroying/tearing down/disposing" of in my `$destroy` callback are not being called, leading to potential memory issues.

Perhaps instead when tabs are removed or destroyed, they should be "re-connected" to the $scope/$digest cycle right before destruction?

robertmesserle:
Closes #1087
Closes #1107
Closes #1140
Closes #1247
Closes #1261
Closes #1380
Closes #1387
Closes #1403
Closes #1443
Closes #1505
Closes #1506
Closes #1516
Closes #1518
Closes #1564
Closes #1570
Closes #1620
Closes #1626
Closes #1698
Closes #1777
Closes #1788
Closes #1850
Closes #1959
Closes #1986


src/components/tabs/js/tabScroll.js
```diff
@@ -0,0 +1,22 @@
+(function () {
+  'use strict';
+  angular.module('material.components.tabs')
+      .directive('mdTabScroll', MdTabScroll);
+
+  function MdTabScroll () {
```
ThomasBurleson: Can you add comments why this is needed ?

robertmesserle: There's no angular directive for mousewheel to my knowledge.  I wrote this to be a generic re-usable mousewheel directive.  I named it mdTabScroll for now, but it could easily be reused in other components.

gkalpak: The implementation seems weird (creating a new scope every time an event happens etc).
Not ot mention sub-optimal from a performance perspective.

Why not follow the paradigm of the rest of the `ngEvent` directives ?
(Reference: https://github.com/angular/angular.js/blob/master/src/ng/directive/ngEventDirs.js)

Issue #1450, opened 2015-02-10
$mdThemingProvider backgroundPalette does not theme the background


jaderabbit:
I configure the theme provider as follows: 
```
$mdThemingProvider
            .theme('default')
            .primaryPalette('deep-orange', {
                'default': '900',
                'hue-2': '500'
            })
            .backgroundPalette('grey', {
                'default': '200'
            })
            .accentPalette('orange');
```

But the background remain's white

nicbon:
It would be really great to have this working.

ThomasBurleson:
@rschmukler - any interim thoughts while we resolve this for 0.9.x ?

blackendstudios:
it takes the '50' value when you set the backgroundPalette, so maybe an easy workaround will be that you can put the background color on the '50' while this get fix... what do u think?

gustavohenke:
@blackendstudios that trick helped me, but I rather use something less "hacky", so I'll just :+1: this issue.

blackendstudios:
ohhhh glad that helped...it was my firts contribution on git hubjajajaja ...great...hope to help more in the future

Hendrixer:
:+1:

arshabh:
This issue is similar to #1321 . The reason why it takes the 'A100' value always is because it is the default color for 'hue-3'. And after digging in the code i found that the theme sass files for components have hard coded background color as:

```javascript
md-content.md-THEME_NAME-theme {
  background-color: '{{background-hue-3}}';
}
```
The above example is from `content-theme.scss`. Similar mention is also found in other theme files like `dialog-theme.scss`, `sidenav-theme.scss`, `slider-theme.scss` and `subheader-theme.scss`.

The solution that i propose for this is something like this

```javascript
md-content.md-THEME_NAME-theme {
  background-color: '{{background-color}}';
}
```

This has started working for me in my local build. But i guess the solution would be more comprehensive - providing background capabilities in a more generic manner. Just putting it out there. Hope it helps

dpipkin:
@blackendstudios Do you change the 'A100' value on `primaryPalette` or `accentPalette`?

arshabh:
@dpipkin it  is on `backgroundPalette`

blackendstudios:
backgroundPalette :)

dpipkin:
Awesome! Thank you.  
The hacky way that I made it work:
```
var background = $mdThemingProvider.extendPalette('grey', {
  'A100': 'f2f2f2'
});
[...]

PR #1950, opened 2015-03-18
fix(theme): change hardcoded hue-3 values for background color to backgr...

arshabh:
fix(theme): change hardcoded hue-3 values for background color to color

Closes: #1450

### DIFF
--- a/src/components/content/content-theme.scss
+++ b/src/components/content/content-theme.scss
@@ -1,3 +1,3 @@
 md-content.md-THEME_NAME-theme {
-  background-color: '{{background-hue-3}}';
+  background-color: '{{background-color}}';
 }
--- a/src/components/dialog/dialog-theme.scss
+++ b/src/components/dialog/dialog-theme.scss
@@ -2,7 +2,7 @@ $dialog-border-radius: 4px !default;
 
 md-dialog.md-THEME_NAME-theme {
   border-radius: $dialog-border-radius;
-  background-color: '{{background-hue-3}}';
+  background-color: '{{background-color}}';
 
   &.md-content-overflow .md-actions {
     border-top-color: '{{foreground-4}}';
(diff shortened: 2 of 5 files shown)


rschmukler:
@arshabh this looks good, but what is actually changing `background-color` into the proper color? Could you add that functionality?

arshabh:
Hey @rschmukler, take a look at the file `theming.js` - line no. 363 - `function parseRules()` takes care of parsing these css rules. It creates a regex by the name `simpleVariableRegex` which replaces specific hues that have been specified explicitly like `{{background-hue-3}}` or `{{primary-100}}`. It also defines a regex by the name `hueRegex` which looks for eg `{{background-color}}` or `{{primary-contrast}}` etc.. and puts in proper color with line no 397.

rschmukler:
@arshabh Right, but usually for `background-color` it will put in the `default` hue for that palette. In this case `500`. That ends up being a different color than `hue-3 (A100)`, which results in the background color being changed to not white, right?

arshabh:
@rschmukler No that does not seem to be the case - `background-color` would enable the parser to put in all hues in place i.e. `hue-1`, `hue-2` and `hue-3`. That is what happens after line no. 397. So the resulting class would look something like this - `md-default-theme .md-hue-1`, `md-default-theme .md-hue-2` and `md-default-theme .md-hue-3` - with their background color as per the specified hue. Only `.md-default-theme` would get the `default` hue for itself.

rschmukler:
@arshabh the issue is that the `default` hue color for the background palette is currently `500` which would cause most people's expected white background to turn grey. I've changed the default background palette and will now merge this in.

arshabh:
@rschmukler yes, pardon me for not confirming to that in my previous statement - yes the default color would not be white and some thing greyish.

Jonybang:
Now the same problem on [sidenav theme](https://github.com/angular/material/blob/e3b0cd8a143086bb3e5173b8d30f2cc8e8e42eca/src/components/sidenav/sidenav-theme.scss).
Please, fix this.

PR #1981, opened 2015-03-20
WIP - comments only - chips

typotter:
use ng-model to set the items instead of directly via attrs

add a default chip template

### DIFF
--- a/src/components/chips/chips.scss
+++ b/src/components/chips/chips.scss
@@ -1,0 +1,28 @@
+md-chips {
+  md-chips-wrap {
+    li {
+      border: 1px solid #DCDCDC;
+      border-radius: 11px;
+      background: #F5F5F5;
+      padding-left:5px;
+      padding-top: 2px;
+      margin-left: 4px;
+      display: inline-block;
+      color: rgba(0,0,0,0.8);
+      font-family: roboto;
+      font-size: 13px;
+      height:22px;
+    }
+    ul {
+    }
+    ul * {
+      vertical-align: middle;
+    }
+  }
+  md-chips-wrap.readonly {
+    li {
+      padding-right: 5px;
+    }
+  }
+}
+
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -1,0 +1,86 @@
+(function () {
+  'use strict';
+  angular
+      .module('material.components.chips')
+      .controller('MdChipsCtrl', MdChipsCtrl);
+
+  function MdChipsCtrl ($scope, $element, $q, $mdUtil, $mdConstant, $log) {
+    //-- private variables
+    var self = this;
+    this.$mdConstant = $mdConstant;
+
+    /** @type {!angular.NgModelController} */
+    this.ngModelCtrl = null;
+
+    /** @type {Array.<Object>} */
+    this.items = [];
+
+    //-- public variables
+    self.scope = $scope;
+
+    return self.init();
+
+    // How to update the model:
+    // self.ngModelCtrl.$setViewValue(new Date(Number(this.dataset.timestamp)));
+    // self.ngModelCtrl.$render();
+
+  }
+
+  /** 
+   * The model used by the input element.
+   * @type {!String}
+   */
+  MdChipsCtrl.prototype.itemBuffer = '';
+
+  MdChipsCtrl.prototype.workerInputKeydown = function(event) { 
+    switch (event.keyCode) {
+      case this.$mdConstant.KEY_CODE.DOWN_ARROW:
+          break;
+      case this.$mdConstant.KEY_CODE.UP_ARROW:
+          break;
+      case this.$mdConstant.KEY_CODE.ENTER:
+          this.appendBuffer();
+          break;
+      case 8:
+          if (this.itemBuffer == '') {
+            this.removeItem(this.items.length - 1);
+          }
+          break;
+      case this.$mdConstant.KEY_CODE.ESCAPE:
+          break;
+      case this.$mdConstant.KEY_CODE.TAB:
+          break;
+      default:
+    }
+  };
+
+
+  MdChipsCtrl.prototype.appendBuffer = function() {
+   this.ngModelCtrl.$viewValue.push(angular.copy(this.itemBuffer));
+   this.resetBuffer();
+  }
+
+  MdChipsCtrl.prototype.resetBuffer = function() {
+    this.itemBuffer = '';
+  }
+
+  MdChipsCtrl.prototype.removeItem = function(idx) {
+    this.items.splice(idx, 1);
+  }
+
+  MdChipsCtrl.prototype.init = function() {
+  }
+
+
+  MdChipsCtrl.prototype.configureNgModel = function(ngModelCtrl) {
+    this.ngModelCtrl = ngModelCtrl;
+
+    var self = this;
+    ngModelCtrl.$render = function() {
+      // model is updated. do something.
+      // self.ngModelCtrl.$viewValue
+      self.items = self.ngModelCtrl.$viewValue;
+    }; 
+    //this.items = this.ngModelCtrl.$viewValue;
+  };
+})();
--- a/src/components/chips/js/chipsDirective.js
+++ b/src/components/chips/js/chipsDirective.js
@@ -1,0 +1,86 @@
+(function () {
+  'use strict';
+  angular
+      .module('material.components.chips')
+      .directive('mdChips', MdChips);
+
+  /**
+   * @ngdoc directive
+   * @name mdChips
+   * @module material.components.chips
+   *
+   * @description
+   * `<md-chips>` 
+   *
+   */
+
+  function MdChips ($compile) {
+    return {
+      template: function(element, attrs) {
+        var userTemplate = element.clone();
+        attrs['$mdUserTemplate'] = userTemplate;
+        return '\
+          <md-chips-wrap ng-class="{readonly : $mdChipsCtrl.readonly}">\
+          <ul role="presentation">\
+            <li ng-repeat="(index, $chip) in $mdChipsCtrl.items"\
+                class="md-chip-list-item">\
+            </li>\
+            <li ng-if="!$mdChipsCtrl.readonly" class="md-chip-worker"></li>\
+          </ul>\
+          </md-chips-wrap>';
+      },
+      require: ['ngModel', 'mdChips'],
+      controller:   'MdChipsCtrl',
+      controllerAs: '$mdChipsCtrl',
+      bindToController: true,
+      compile: compile,
+      scope: {
+        readonly: '=readonly'
+      }
+    };
+    function compile(element, attr) {
+      var userTemplate = attr['$mdUserTemplate'];
+      var chipEl = userTemplate.find('md-chip');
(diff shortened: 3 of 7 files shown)


src/components/chips/chips.scss
```diff
@@ -0,0 +1,28 @@
+md-chips {
+  md-chips-wrap {
+    li {
```
jelbourn: Try to avoid using element selectors for styles and just add a class. This makes it easier for people to override the style and makes it easier to create a CSS API.

src/components/chips/chips.scss
```diff
@@ -0,0 +1,28 @@
...
+      border-radius: 11px;
+      background: #F5F5F5;
+      padding-left:5px;
+      padding-top: 2px;
+      margin-left: 4px;
+      display: inline-block;
+      color: rgba(0,0,0,0.8);
+      font-family: roboto;
+      font-size: 13px;
+      height:22px;
```
jelbourn: Lots of magic-value numbers in here. Can you extract some into SCSS variables with some semantically meaningful name?

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,86 @@
...
+    var self = this;
+    this.$mdConstant = $mdConstant;
+
+    /** @type {!angular.NgModelController} */
+    this.ngModelCtrl = null;
+
+    /** @type {Array.<Object>} */
+    this.items = [];
+
+    //-- public variables
```
jelbourn: If you want to distinguish public vs. private variables on the controller, I'd suggest using the `@private` annotation on private things (and maybe ending with an underscore, but that's not strictly necessary).

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,86 @@
...
+    self.scope = $scope;
```
jelbourn: Why does scope need to be public?

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,86 @@
...
+    // self.ngModelCtrl.$setViewValue(new Date(Number(this.dataset.timestamp)));
+    // self.ngModelCtrl.$render();
+
+  }
+
+  /** 
+   * The model used by the input element.
+   * @type {!String}
+   */
+  MdChipsCtrl.prototype.itemBuffer = '';
```
jelbourn: This shouldn't be on the prototype; doing this would make one reference for all Chips instances.

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,86 @@
...
+
+  MdChipsCtrl.prototype.workerInputKeydown = function(event) {
```
jelbourn: Try to add function descriptions as you go. As it is now, I'm not sure what "worker" in the function name refers to.

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,86 @@
+(function () {
+  'use strict';
+  angular
+      .module('material.components.chips')
+      .controller('MdChipsCtrl', MdChipsCtrl);
+
+  function MdChipsCtrl ($scope, $element, $q, $mdUtil, $mdConstant, $log) {
```
jelbourn: Add a class comment like
```js
  /**
   * Controller for the MdChips component. Something about responsibilities. 
   * @ngInject @constructor
   */
function MdchipsCtrl() { }
```

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,86 @@
...
+
+  MdChipsCtrl.prototype.resetBuffer = function() {
+    this.itemBuffer = '';
+  }
+
+  MdChipsCtrl.prototype.removeItem = function(idx) {
+    this.items.splice(idx, 1);
+  }
+
+  MdChipsCtrl.prototype.init = function() {
```
jelbourn: I know it's still early on, but try to keep any initialization functions well-scoped to a single responsibility.

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,86 @@
...
+  'use strict';
+  angular
+      .module('material.components.chips')
+      .controller('MdChipsCtrl', MdChipsCtrl);
+
+  function MdChipsCtrl ($scope, $element, $q, $mdUtil, $mdConstant, $log) {
+    //-- private variables
+    var self = this;
+    this.$mdConstant = $mdConstant;
+
```
jelbourn: Also define any properties that _will_ be deifned on `this` here, even if they get initialized later (such as `ngModelCtrl`)

src/components/chips/js/chipsDirective.js
```diff
@@ -0,0 +1,86 @@
...
+    function compile(element, attr) {
+      var userTemplate = attr['$mdUserTemplate'];
+      var chipEl = userTemplate.find('md-chip');
+      if (chipEl.length == 0) {
+        chipEl = angular.element('\
+            <md-chip>\
+              <span>{{$chip}}</span>\
+              <button ng-if="!$mdChipsCtrl.readonly"\
+                  md-chip-remove>x</button>\
+            </md-chip>');
```
jelbourn: Break this template string out into a constant outside the function.

src/components/chips/js/chipsDirective.js
```diff
@@ -0,0 +1,86 @@
...
+   *
+   * @description
+   * `<md-chips>` 
+   *
+   */
+
+  function MdChips ($compile) {
+    return {
+      template: function(element, attrs) {
+        var userTemplate = element.clone();
```
jelbourn: Here's a comment I wrote a while back explaining this technique:
```js
      // Specifying templateUrl as a function gives us the opportunity to
      // store the user-provided template for the directive instance, which
      // we can manually parse (rather than having angular compile them).
      // The attrs object is a conventient place to store this template because
      // the same attrs are passed to the appropriate link function invocation
      // for this directive instance. Prefix the property name with a $ so that
      // angular will not actually append the property to the rendered element.
      // The element is cloned because this element reference actually has its
      // contents replaced by the directive's rendered template.
```

src/components/chips/js/chipsDirective.js
```diff
@@ -0,0 +1,86 @@
...
+      controllerAs: '$mdChipsCtrl',
+      bindToController: true,
+      compile: compile,
+      scope: {
+        readonly: '=readonly'
+      }
+    };
+    function compile(element, attr) {
+      var userTemplate = attr['$mdUserTemplate'];
+      var chipEl = userTemplate.find('md-chip');
```
jelbourn: I suggest just working with raw DOM elements here instead of going through `angular.element`. The DOM APIs are quite usable for the browsers we support today.

src/components/chips/js/chipsDirective.js
```diff
@@ -0,0 +1,86 @@
...
+        }
+      }
+      var listNode = angular.element(element[0].querySelector('li.md-chip-list-item'));
+      listNode.append(chipEl);
+
+      // Input Element: Look for an autocomplete or an input.
+      var inputEl = userTemplate.find('md-autocomplete');
+      var hasAutocomplete = inputEl.length > 0;
+
+      if (!hasAutocomplete) {
```
jelbourn: Also need to check for an `<input>` and possibly a `<textarea>`

src/components/chips/js/xgeneDirective.js
```diff
@@ -0,0 +1,25 @@
...
+  'use strict';
+  angular
+      .module('material.components.chips')
+      .directive('mdXgene', MdXgene);
+
+  /**
+   * A directive for lazily binding a single property of aparent scope to a
+   * transcluded chunk. ngRepeats typically need this late-bound property.
+   */
+  function MdXgene ($compile, $mdUtil) {
```
jelbourn: wat

src/components/chips/chips.scss
marcysutton: The font and typography should be inherited wherever possible, as well. PR #1985 will help with this.

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,86 @@
...
+  /** 
+   * The model used by the input element.
+   * @type {!String}
+   */
+  MdChipsCtrl.prototype.itemBuffer = '';
+
+  MdChipsCtrl.prototype.workerInputKeydown = function(event) { 
+    switch (event.keyCode) {
+      case this.$mdConstant.KEY_CODE.DOWN_ARROW:
```
marcysutton: :+1: nice use of `$mdConstant`!

src/components/chips/js/chipsDirective.js
```diff
@@ -0,0 +1,86 @@
...
+        // Default element.
+        inputEl = angular.element('\
+            <input type="text" \
```
marcysutton: This input needs a localizable label, either using `aria-label` or an offscreen `<label>` element.

src/components/chips/js/xgeneDirective.js
```diff
@@ -0,0 +1,25 @@
...
+      require: '^?mdChips',
+      terminal: true,
+      link: link,
+      scope: false
+    };
+    function link (scope, element, attr, ctrl) {
+      var newScope = scope.$new(false, scope);
+      newScope[attr.mdXgene] = scope.$eval(attr.mdXgene);
+      $compile(element.contents())(newScope);
+      element.attr({ 'role': 'option', 'id': 'item_' + $mdUtil.nextUid() });
```
marcysutton: Is the `listbox` role being added to the parent elsewhere? http://www.w3.org/TR/2010/WD-wai-aria-20100916/roles#option

src/components/chips/js/chipsDirective.js
```diff
@@ -0,0 +1,86 @@
...
+      }
+    };
+    function compile(element, attr) {
+      var userTemplate = attr['$mdUserTemplate'];
+      var chipEl = userTemplate.find('md-chip');
+      if (chipEl.length == 0) {
+        chipEl = angular.element('\
+            <md-chip>\
+              <span>{{$chip}}</span>\
+              <button ng-if="!$mdChipsCtrl.readonly"\
```
marcysutton: The `mdAutocomplete` remove button is probably a good pattern to look at: https://github.com/angular/material/blob/master/src/components/autocomplete/js/autocompleteDirective.js#L79

src/components/chips/js/chipsController.js
robertmesserle: Since `itemBuffer` is a string, this shouldn't be able to cause any issues since changing its value will assign the new value to the instance, not the prototype.

jelbourn: Yeah, you're right, it won't cause any issues, but it has no reason to be on the prototype either way.

typotter: This is probably a side-effect of me coming from projects that heavily
employed Closure where having public (and private) members on the prototype
is the rule.

jelbourn: Even in Closure-land, instance properties don't go on the prototype- only methods.

typotter: can confirm - checked it out in the JSCompiler styleguide... Interesting...
Social team should probably be told this. We did this all over the place in
Glass, after stealing some of their engineers and basing our entire stack
off theirs...

PR #1985, opened 2015-03-20
CSS Refactor: Minimize global styles, support high contrast mode, new typography feature

marcysutton:
Includes: 
 * Framework typography stylesheet
 * SVG icons in the docs (menu, list toggle)
 * Flex styles for `layout.scss` in Safari
 * Docs menu & demo source button bug fixes (Safari, Firefox, IE)
 * Media queries for Windows high contrast mode
 * Relegates `text-rendering` to specific elements concerned with typography

Closes #1442, #1561, #1128, #1989, Refs #1718

PR #2021, opened 2015-03-24
Add List Controls

rschmukler:


src/components/list/demoListControls/index.html
```diff
@@ -0,0 +1,29 @@
+<md-list ng-controller="ListCtrl">
+  <md-subheader class="md-no-sticky">Single Action Checkboxes</md-subheader>
+  <md-item ng-repeat="topping in toppings">
+    <p> {{ topping.name }} </p>
+    <md-checkbox class="md-secondary" ng-model="topping.wanted" aria-label="{{ topping.name }}">
```
marcysutton: I wonder how many people are going to leave off the `aria-label` here. I think the way you have it coded will already be covered by the `$mdAria` warnings, but I'm wondering if it makes more sense to associate the `<p>` with the checkbox rather than duplicating `{{topping.name}}`. Either with `aria-labelledby` or a more clear element API. Just want to throw it out there though.

src/components/list/demoListControls/index.html
```diff
@@ -0,0 +1,29 @@
...
+  </md-item>
+  <md-subheader class="md-no-sticky">Clickable Items with Secondary Controls</md-subheader>
+  <md-item ng-click="navigateTo(setting.extraScreen)" ng-repeat="setting in settings">
+    <md-icon md-svg-icon="{{setting.icon}}"></md-icon>
+    <p> {{ setting.name }} </p>
```
marcysutton: The click on this `<md-item>` is doing something different than the `<md-switch>`, yet the have the same label of `{{setting.name}}`. How can we encourage labels that actually indicate what the two actions/controls do?

src/components/list/demoListControls/index.html
```diff
@@ -0,0 +1,29 @@
...
+    <md-switch class="md-secondary" ng-model="setting.enabled" aria-label="{{ setting.name }}">
+  </md-item>
+  <md-item ng-click="navigateTo('data usage')">
+    <md-icon md-svg-icon="cached"></md-icon>
+    <p>Data Usage</p>
+  </md-item>
+  <md-subheader class="md-no-sticky">Checkbox with Secondary Action</md-subheader>
+  <md-item ng-repeat="message in messages">
+    <md-checkbox ng-model="message.selected" aria-label="Select"></md-checkbox>
```
marcysutton: When I click on an `md-checkbox` to activate it, it gets different indentation. Seems like a weird render bug, because it goes away when I blur the browser window. 
Checkboxes

src/components/list/demoListControls/index.html
```diff
@@ -0,0 +1,29 @@
...
+    <p>{{message.title}}</p>
+    <md-icon class="md-secondary" ng-click="doSecondaryAction()", md-svg-icon="communication:message"></md-icon>
```
marcysutton: This item needs an `aria-label` if it has `ng-click` on it. Without `aria-label` present, `md-icon` gets `aria-hidden="true"` because it assumes you meant it was for presentation only. So we'll have to enforce it somehow.

src/components/list/demoListControls/index.html
rschmukler: @marcysutton completely defer to you in all things aria. How should we do it?

src/components/list/demoListControls/index.html
marcysutton: In the compiled code, the `<div role="button">` wrapper and checkbox with `tabIndex="-1"` don't currently work for accessibility because the user isn't focused on the control that's actually being toggled (they are focused on the button "wrapper"). 

Ideally, focus would go directly onto the checkbox and you'd toggle it there, so that its semantics were announced. The `<div role=button>` is also unnecessary around a switch, as I'm pretty sure you wouldn't ever nest a checkbox inside a button. I think any place we transform a list item with a switch as its primary action, we're going to run into this problem with the code in its current state. We *may* be able to try something tricky on the parent, like `aria-controls`, but we'll have to test it.

rschmukler:
Note to self to re-add https://github.com/angular/material/commit/38f04230a3064f84701a1c7eca862fd6bb2c256f

src/components/list/list.js
```diff
@@ -74,11 +77,102 @@ function mdListDirective() {
...
+      // Check for proxy controls (no ng-click on parent, and a control inside)
+      var secondaryItem = tEl[0].querySelector('.md-secondary');
+      var hasProxiedElement;
+
+      if (!tAttrs.ngClick) {
+        for (var i = 0, type; type = proxiedTypes[i]; ++i) {
+          var proxyElement;
+          if (proxyElement = tEl[0].querySelector(type)) {
+            hasProxiedElement = true;
+            proxyElement.setAttribute('tabindex', '-1');
```
marcysutton: This causes problems on checkboxes and switches, because those elements have to be focused in order for their semantics to be announced. They also shouldn't be nested in `role=button`...can we just get rid of `<div role=button>` if it's a switch?

src/components/button/button-theme.scss
```diff
@@ -5,6 +5,10 @@ md-toolbar .md-button.md-THEME_NAME-theme.md-fab {
   background-color: white;
 }
 
+button.md-no-style {
```
marcysutton: Does this actually match anything? The element is a `<div>` in JS....I'm wondering if we can make it a real `button`. I got around that Safari layout bug by wrapping its contents in a `flex` element.

src/components/list/list-theme.scss
```diff
@@ -0,0 +1,27 @@
...
+    h3, h4 {
+      color: '{{foreground-1}}';
+    }
+    p {
+      color: '{{foreground-2}}';
+    }
+  }
+  .md-clickable:focus,
+  .md-secondary:focus,
+  .md-no-style:focus {
```
marcysutton: :+1: for thinking about focus

src/components/list/list.js
```diff
@@ -74,11 +77,102 @@ function mdListDirective() {
...
+        }
+        function computeClickable() {
+          if (proxies.length || $element[0].firstElementChild.hasAttribute('ng-click')) { 
+            $element.addClass('md-clickable');
+          }
+        }
+
+        if (!$element[0].firstElementChild.hasAttribute('ng-click')) {
+          $element[0].firstElementChild.addEventListener('keypress', function(e) {
+            if (e.keyCode == 13 || e.keyCode == 32) {
```
marcysutton: This looks a lot like the code from ngAria, which has bugs related to custom elements in IE and Firefox (see https://github.com/angular/angular.js/issues/11287). At the very least, we should expand this to also include `e.which`. Like this:
```javascript
var keyCode = event.which || event.keyCode;
if (keyCode === 32 || keyCode === 13) {}
```

src/components/list/list.scss
```diff
@@ -16,63 +17,161 @@ $item-padding-top: 0px !default;
...
+$list-item-text-padding-top: $baseline-grid * 2 !default;
+$list-item-text-padding-bottom: $baseline-grid * 2.5 !default;
 
 md-list {
+  display: block;
   padding: $list-padding-top $list-padding-right $list-padding-bottom $list-padding-left;
+
+  .md-subheader {
+    font-weight: 500;
+    font-size: 0.875em;
```
marcysutton: Should this use a variable? Also, should we start using `rem` units for fonts? That's what the new Typography stylesheet uses.

src/components/list/list.scss
```diff
@@ -16,63 +17,161 @@ $item-padding-top: 0px !default;
...
+  &.md-no-style,
+  .md-no-style {
+    padding: $list-item-padding-vertical $list-item-padding-horizontal;
+    flex: 1;
+  }
+  &.md-with-secondary {
+    padding-right: $list-item-padding-horizontal;
+  }
+  &.md-clickable:hover {
+    cursor: pointer;
```
marcysutton: There is a small gap where you get the hover cursor but clicking doesn't activate anything. Looks like `md-icon` should either have padding instead of margin or be wrapped in a clickable/focusable element that has padding.

src/components/list/list.scss
```diff
@@ -16,63 +17,161 @@ $item-padding-top: 0px !default;
...
+  }
 }
 
-md-item-content {
+md-item, md-item > .md-no-style {
   display: flex;
```
marcysutton: This might need prefixed `flex` properties for broader support.

src/components/list/list.scss
```diff
@@ -16,63 +17,161 @@ $item-padding-top: 0px !default;
...
+  & > div.md-primary > md-icon,
+  & > div.md-secondary > md-icon,
+  & > md-icon:first-child, 
+  > md-icon.md-secondary {
+    width: $list-item-primary-icon-width;
+    margin-top: 12px;
+    margin-bottom: 12px;
+    box-sizing: content-box;
+    &:focus {
+      padding: 6px;
```
marcysutton: What is the intention with these focus styles?

src/components/list/list.scss
```diff
@@ -16,63 +17,161 @@ $item-padding-top: 0px !default;
...
+  .md-item-text {
+    flex: 1;
+    padding: $baseline-grid * 2 0;
+    text-overflow: ellipsis;
 
-  h3 {
-    margin: $list-h3-margin;
-    font-weight: $list-h3-font-weight;
-    font-size: $list-h3-font-size;
+    h3 {
```
marcysutton: We should revisit these after PR #1985 lands, in case we can consolidate styles.

src/components/list/demoListControls/index.html
marcysutton: I'm actually wondering if the right-aligned single-action checkbox is even a thing. After looking at the spec, I don't see that combination. We could use the existing checkbox component if we make single-line items left aligned, they they show in the spec:
```html
<md-item ng-repeat="topping in toppings">
  <md-checkbox class="md-secondary" ng-model="topping.wanted">
    {{ topping.name }}
  </md-checkbox>
</md-item>
```
That also should NOT be wrapped in `<div role=button>`. That makes the checkbox unusable.

src/components/list/list.scss
rschmukler: Autoprefixer will take care of that for us.

src/components/list/list.scss
rschmukler: So that icon buttons look good. If you tab through some of the secondary controls (icons with `ng-click`) you can see how they get applied.

src/components/list/list.scss
rschmukler: :+1:

src/components/button/button-theme.scss
rschmukler: Nope, it doesn't haha. Nice catch. I'll try the safari fix then.

ilanbiala:
Looking good so far, is this going to be part of 0.9?

rschmukler:
@ilanbiala we hope so.

rschmukler:
- [ ] Don't wrap a single `md-secondary` checkbox with a button (topping example)
- [ ] inject id on paragraph and do `aria-labelledby` on the paragraph (topping example)
- [ ] Layout fix for buttons instead of `role="button"`.  
- [ ] Auto infer aria-label for md-secondary w/ switch/checkbox as `Toggle {{p.textContent}}`.

config/build.config.js
```diff
@@ -4,7 +4,7 @@ var fs = require('fs');
 var versionFile = __dirname + '/../dist/commit';
 
 module.exports = {
-  ngVersion: '1.3.2',
+  ngVersion: '1.3.15',
```
marcysutton: :+1: :+1:  this will also upgrade ngAria to a better version

Issue #1967, opened 2015-03-19
md-dialog backdrop displays incorrectly if page is scrolled


dozingcat:
See http://codepen.io/anon/pen/ZYVpWo . Scroll down and click the button; the dialog backdrop will not cover the entire frame, and sometimes showing the dialog causes the page to unexpectedly jump.

dominikstrasser:
maybe this could help you....
md-backdrop{
  position: fixed;
}

wesalvaro:
The same is true for the bottom sheet.

I have a PR for the md-backdrop in #2023.

david-gang:
I think this is also connected to #1830 
It happens also with sidenavs and scrolled content.

ivanhtp:
Hey guys! This ticket is really important.

Just to complement: position fixed to backdrop won't solve the problem completely.
You must also change the class .md-dialog-is-showing on body ( overflow: hidden) to overflow: visible.

alexthewilde:
+1

wesalvaro:
Quick fiddle to demonstrate this is still a very present bug:

http://output.jsbin.com/yotiloniri

Just make sure to shrink the screen vertically smaller than the content (the kitten picture).

oli107:
+1

pablorsk:
Solution:

Search object causing the proble. For example a `<md-sidenav />`. Next, check the parent, for example a div. On the code...

```html
<div>
  <md-sidenav>
     <!-- this md-sidenav tag is causing a problem -->
  </md-sidenav>
</div>
```

Fixing the problem...

```html
<div style="position:fixed;">
  <md-sidenav>
     <!-- No more warning! :) -->
  </md-sidenav>
</div>
```

stephengardner:
the `top` offset of the `<md-backdrop>` element is not being properly calculated...
adding: 
```css
md-backdrop {
    top: 0;
}
```
might help you, if this is the case.

gglukmann:
I managed to fix it with setting background from hex to rgba value with rgba(243, 246, 251, .99)

tavurth:
Had the same problem, the body was getting custom styles added, a hacky fix:

```
/**
 * If we don't do this, angular material will break when loading modal windows
 * Don't remove this unless you have a better alternative
 */
body {
    top: 0 !important;
}
```

ImaginationHub:
To solve the incorrect position after the scrolling in iPad Safari, please try:
.md-dialog-container{
  position:fixed; 
[...]

PR #2023, opened 2015-03-25
bug(backdrop): Position should be fixed to account for scrolling.

wesalvaro:
If the position is absolute, after scrolling the backdrop does not cover the entire viewport.

I assume this is not an oversight, so what am I doing wrong?

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2023)

### DIFF
--- a/src/components/backdrop/backdrop.scss
+++ b/src/components/backdrop/backdrop.scss
@@ -15,7 +15,7 @@ md-backdrop {
 
   background-color: rgba(0,0,0,0);
 
-  position: absolute;
+  position: fixed;
   left: 0;
   top: 0;
   right: 0;


wesalvaro:
I've replied privately to you @ThomasBurleson

ThomasBurleson:
@wesalvaro - never got it.

wesalvaro:
I don't know why I would say that and not do it... But I don't even know how to reply privately nor why I would... That's so strange of me...

But I'm also not sure why you would close this without comment. It was a pretty straightforward demo on the ng-material website and the other issue referenced #1967 makes it obvious this is an actual issue.

However, looking at the current demo on the site, I could not reproduce this issue. The hacks in place to abate it seem rather on the crazy side, though. Including the height style values of 100%, the overflow:hidden element style and the md-dialog-is-showing class.

Anyway, I was still very easily able to reproduce the issue with a fiddle (and there is no need to PM it to you! haha): http://output.jsbin.com/yotiloniri Just make sure to shrink the screen vertically smaller than the content (the kitten picture).

barryvdh:
See http://codepen.io/anon/pen/EjeOBZ

When you need to scroll to see the button and you click it, the backdrop is wrong and the dialog isn't visible.

barryvdh:
But to fix that, you should set `.md-dialog-container` also to fixed, and perhaps `.md-scroll-mask` too.

barryvdh:
It was working in rc2: http://codepen.io/anon/pen/XbPLXv

ThomasBurleson:
Resolved with SHA b71fdfb609a9b8c5bf6e2f4b41638020e265acd2

PR #2030, opened 2015-03-25
feat(chips): initial commit

typotter:
Supports basic list mutation (add, remove items). Does not yet support
child-input elements (renders its own input element by default).
TODO/future work laid out in directive comments.

### DIFF
--- a/src/components/chips/chips.scss
+++ b/src/components/chips/chips.scss
@@ -1,0 +1,51 @@
+$chip-font-size: 13px !default;
+$chip-height: 22px !default;
+$chip-button-padding: 2px !default;
+$chip-padding: 0 8px 0 8px !default;
+$chip-margin: 0 0 0 8px !default;
+
+.md-chips {
+  box-shadow: $whiteframe-shadow-z1;
+  display: block;
+  font-family: $font-family;
+  font-size: $chip-font-size;
+  padding: 8px;
+  vertical-align: middle;
+
+  .md-chip {
+    border-radius: $chip-height / 2;
+    box-shadow: $whiteframe-shadow-z1;
+    display: inline-block;
+    height: $chip-height;
+    margin: $chip-margin;
+    line-height: $chip-height;
+    padding: $chip-padding;
+    transform: translate3d(0, 0, 0);
+
+    &:first-child {
+      margin: 0;
+    }
+
+    &.selected {
+      transform: translate3d(0, -1px, 0);
+    }
+
+    .md-button {
+      padding: $chip-button-padding;
+    }
+  }
+
+  .md-chip-worker {
+    display: inline-block;
+    line-height: $chip-height + 3px;  /* 3px are added to height of chip for...the shadow? */
+
+    &:not(:first-child) {
+      margin: $chip-margin;
+    }
+  }
+
+  .md-chip-input {
+    background-color:transparent;
+    border-width: 0px;
+  }
+}
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -1,0 +1,137 @@
+describe('<md-chips>', function() {
+
+  beforeEach(module('material.components.chips'));
+
+  function compile (str, scope) {
+    var container;
+    inject(function ($compile) {
+      container = $compile(str)(scope);
+      scope.$apply();
+    });
+    return container;
+  }
+
+  function createScope () {
+    var scope;
+    var items = ['Apple', 'Banana', 'Orange'];
+    inject(function ($rootScope) {
+      scope = $rootScope.$new();
+      scope.items = items;
+    });
+    return scope;
+  }
+
+  function getChipElements(root) {
+    return angular.element(root[0].querySelectorAll('div.md-chip'));
+  }
+
+  describe('basic functionality', function () {
+    it('should render a default input element', function() {
+      var scope = createScope();
+      var template = '<md-chips ng-model="items"></md-chips>';
+      var element = compile(template, scope);
+      var ctrl = element.controller('mdChips');
+
+      element.scope().$apply();
+      var input = element.find('input');
+      expect(input.length).toBe(1);
+      expect(input).toHaveClass('md-chip-input');
+    });
+
+    it('should render a list of chips', function() {
+      var scope = createScope();
+      var template = '<md-chips ng-model="items"></md-chips>';
+      var element = compile(template, scope);
+      var ctrl = element.controller('mdChips');
+
+      element.scope().$apply();
+      var chips = getChipElements(element);
+      expect(chips.length).toBe(3);
+      expect(chips[0].innerHTML).toContain('Apple');
+      expect(chips[1].innerHTML).toContain('Banana');
+      expect(chips[2].innerHTML).toContain('Orange');
+    });
+
+    it('should render a user-provided chip template', function() {
+      var scope = createScope();
+      var template =
+          '<md-chips ng-model="items">' +
+          '  <md-chip class="mychiptemplate"></md-chip>' +
+          '</md-chips>';
+      var element = compile(template, scope);
+      var ctrl = element.controller('mdChips');
+
+      element.scope().$apply();
+      var chip = element.find('md-chip');
+      expect(chip).toHaveClass('mychiptemplate');
+    });
+
+    it('should add a chip', function() {
+      var scope = createScope();
+      var template = '<md-chips ng-model="items"></md-chips>';
+      var element = compile(template, scope);
+      var ctrl = element.controller('mdChips');
+
+      element.scope().$apply();
+      ctrl.chipBuffer = 'Grape';
+      element.scope().$apply();
+
+      ctrl.appendChipBuffer();
+      element.scope().$apply();
+
+      var chips = getChipElements(element);
+      expect(chips.length).toBe(4);
+      expect(chips[0].innerHTML).toContain('Apple');
+      expect(chips[1].innerHTML).toContain('Banana');
+      expect(chips[2].innerHTML).toContain('Orange');
+      expect(chips[3].innerHTML).toContain('Grape');
+    });
+
+    it('should remove a chip', function() {
+      var scope = createScope();
+      var template = '<md-chips ng-model="items"></md-chips>';
+      var element = compile(template, scope);
+      var ctrl = element.controller('mdChips');
+
+      element.scope().$apply();
+
+      // Remove "Banana"
+      ctrl.removeChip(1);
+      element.scope().$apply();
+
+      var chips = getChipElements(element);
+      expect(chips.length).toBe(2);
+      expect(chips[0].innerHTML).toContain('Apple');
+      expect(chips[1].innerHTML).toContain('Orange');
+    });
+  });
+
+  describe('<md-chip-remove>', function() {
+    it('should remove a chip', function() {
+      var scope = createScope();
+      var template = '<md-chips ng-model="items"></md-chips>';
+      var element = compile(template, scope);
+      var ctrl = element.controller('mdChips');
+      element.scope().$apply();
+
+      var chips = getChipElements(element);
+      expect(chips.length).toBe(3);
+      // Remove 'Banana'
+      var db = angular.element(chips[1]).find('md-button');
+      db[0].click();
+      element.scope().$apply();
+
+      chips = getChipElements(element);
+      expect(chips.length).toBe(2);
+
+      // Remove 'Orange'
+      db = angular.element(chips[1]).find('md-button');
+      db[0].click();
+      element.scope().$apply();
+
+      chips = getChipElements(element);
+      expect(chips.length).toBe(1);
+    });
+  });
+
+});
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -1,0 +1,196 @@
+(function () {
+  'use strict';
+  angular
+      .module('material.components.chips')
+      .controller('MdChipsCtrl', MdChipsCtrl);
+
+
+
+  /**
+   * Controller for the MdChips component. Responsible for adding to and
+   * removing from the list of chips, marking chips as selected, and binding to
+   * the models of various input components.
+   *
+   * @param $mdUtil
+   * @param $mdConstant
+   * @param $log
+   * @ngInject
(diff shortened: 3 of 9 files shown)


src/components/chips/chips.scss
```diff
@@ -0,0 +1,39 @@
+$chip-bgcolor: #F5F5F5;
+$chip-border: 1px solid #DCDCDC;
```
jelbourn: These should almost certainly be in terms of the theme

src/components/chips/chips.spec.js
```diff
@@ -0,0 +1,133 @@
...
+      expect(chips.length).toBe(4);
+      expect(chips[0].innerHTML).toContain('Apple');
+      expect(chips[1].innerHTML).toContain('Banana');
+      expect(chips[2].innerHTML).toContain('Orange');
+      expect(chips[3]).toHaveClass('md-chip-worker');
+    });
+
+    it('should render a user-provided chip template', function() {
+      var scope = createScope();
+      var template = '<md-chips ng-model="items"><md-chip class="mychiptemplate"></md-chip></md-chips>';
```
jelbourn: It's more readable to format the inline templates like
```js
var template = 
  '<md-chips ng-model="items">' +
    '<md-chip class="mychiptemplate"></md-chip>' + 
  '</md-chips>';
```

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+
+  /**
+   * Handles the keydown event on the input element: <enter> appends the
+   * buffer to the chip list, while backspace removes the last chip in the list
+   * if the current buffer is empty.
+   * @param event
+   */
+  MdChipsCtrl.prototype.workerInputKeydown = function(event) {
```
jelbourn: What does the word "worder" mean?

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,196 @@
...
+    switch (event.keyCode) {
+      case this.$mdConstant.KEY_CODE.ENTER:
+        event.preventDefault();
+        this.appendChipBuffer();
+        break;
+      case this.$mdConstant.KEY_CODE.BACKSPACE: // backspace
+        event.preventDefault();
+        if (this.chipBuffer === '') {
```
jelbourn: `if (!this.chipBuffer)` ?

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,196 @@
...
+  };
+
+
+  /**
+   * Append the contents of the buffer to the chip list. This method will first
+   * call out to the md-chip-append method, if provided
+   */
+  MdChipsCtrl.prototype.appendChipBuffer = function() {
+    var newChip = this.getChipBuffer();
+    if (this.useMdChipAppend && !!this.mdChipAppend) {
```
jelbourn: Use of `!!` is not necessary in conditionals

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,196 @@
...
+   * {@code md-input-container} element.
+   * @return {Object|string}
+   */
+  MdChipsCtrl.prototype.getChipBuffer = function() {
+    if (!!this.mdAutocompleteCtrl) {
+      this.$log.error('md-autocomplete not yet supported');
+    } else if (this.hasInputElement) {
+      this.$log.error('user-provided inputs not yet supported');
+    } else {
+      return angular.copy(this.chipBuffer);
```
jelbourn: You don't need to copy the chipBuffer

src/components/chips/js/chipsDirective.js
```diff
@@ -0,0 +1,162 @@
...
+      var chipEl = userTemplate.find('md-chip');
+      if (chipEl.length === 0) {
+        chipEl = angular.element(CHIP_DEFAULT_TEMPLATE);
+      } else {
+        // Warn if no remove button is included in the template.
+        if (!chipEl[0].querySelector('[md-chip-remove]')) {
+          $log.warn('md-chip-remove attribute not found in md-chip template.');
+        }
+      }
+      var listNode = angular.element(element[0].querySelector('li.md-chip-list-item'));
```
jelbourn: Does `li` need to be part of this selector?

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,196 @@
...
+      this.chipBuffer = '';
+    }
+  };
+
+
+  /**
+   * Removes the chip at the given index.
+   * @param idx
+   */
+  MdChipsCtrl.prototype.removeChip = function(idx) {
```
jelbourn: `idx` > `index`
(yep, I'm that picky)

src/components/chips/chips.scss
```diff
@@ -0,0 +1,39 @@
...
+$chip-font-size: 13px;
+$chip-height: 22px;
+$chip-button-padding: 2px;
+$chip-padding: 1px 5px 0px 5px;
```
jelbourn: `0px` > `0`

src/components/chips/chips.scss
```diff
@@ -0,0 +1,39 @@
...
+$chip-margin: 0 0 0 4px;
+
+.md-chips {
+  * {
+    font-family: $font-family;
+    font-size: $chip-font-size;
+  }
+  li:first-child {
```
jelbourn: Try to avoid using `ul` and `li` and just use CSS classes in rules. 

Also, try to reduce the amount of nesting going on here.

src/components/chips/chips.spec.js
```diff
@@ -0,0 +1,133 @@
...
+    inject(function ($rootScope) {
+      scope = $rootScope.$new();
+      scope.items = items;
+    });
+    return scope;
+  }
+
+  describe('basic functionality', function () {
+    it('should render a default input element', function() {
+      var scope = createScope();
```
jelbourn: I'm not a fan of aligning assignment operators, but especially when it's not done consistently

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,196 @@
...
+   * Configures the required interactions with the ngModel Controller.
+   * Specifically, set {@code this.items} to the {@code NgModelCtrl#$viewVale}.
+   * @param ngModelCtrl
+   */
+  MdChipsCtrl.prototype.configureNgModel = function(ngModelCtrl) {
+    this.ngModelCtrl = ngModelCtrl;
+
+    var self = this;
+    ngModelCtrl.$render = function() {
+      // model is updated. do something.
```
jelbourn: Prefix 
`TODO(typotter): ...`

(here and for other todo-style comments)

src/components/chips/js/chipsController.js
typotter: The last chip in the list was labeled as a 'worker' chip, as a nod to the fact that it contains the input element. Since it's in the list of chips, it needs some special styling and so a distinguishing name. kept the CSS name, but switched to default in the js

typotter:
Thanks for the review, Jeremy. ptal

src/components/chips/chips.scss
```diff
@@ -0,0 +1,51 @@
+$chip-font-size: 13px;
```
jelbourn: Add !default to each of these so that someone could theoretically override them.

ThomasBurleson:
@typotter - Can you rebase this PR from latest master source and squash your commits? Then I can finish reviewing and merge. Thx.

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,196 @@
...
+
+
+  /**
+   * Resets the input buffer.
+   */
+  MdChipsCtrl.prototype.resetChipBuffer = function() {
+    if (this.mdAutocompleteCtrl) {
+      this.$log.error('md-autocomplete not yet supported');
+    } else {
+      this.chipBuffer = '';
```
ThomasBurleson: Since Line 74 checks this `if (!this.chipBuffer)`, I do not recommend using

```js
this.chipBuffer = '';
```

instead use 

```js
this.chipBuffer = null;
```

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,196 @@
...
+   * @param event
+   */
+  MdChipsCtrl.prototype.defaultInputKeydown = function(event) {
+    switch (event.keyCode) {
+      case this.$mdConstant.KEY_CODE.ENTER:
+        event.preventDefault();
+        this.appendChipBuffer();
+        break;
+      case this.$mdConstant.KEY_CODE.BACKSPACE: // backspace
+        if (!this.chipBuffer) {
```
ThomasBurleson: I think we should use `if ( this.chipBuffer.length )`

jelbourn: @ThomasBurleson why? The truthy check semantically says "Do I have no value?", which is appropriate here.

src/components/chips/js/chipsController.js
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+
+  /**
+   * Handles the keydown event on the input element: <enter> appends the
+   * buffer to the chip list, while backspace removes the last chip in the list
+   * if the current buffer is empty.
+   * @param event
+   */
+  MdChipsCtrl.prototype.defaultInputKeydown = function(event) {
```
ThomasBurleson: Our Code standard for Angular Material 1.x is to use  `<class>.prototype = { ... }`.
See [mdGridList](https://github.com/angular/material/blob/master/src/components/gridList/gridList.js#L396) and [mdIcon](https://github.com/angular/material/blob/master/src/components/icon/iconService.js#L185)

src/components/chips/js/chipsDirective.js
```diff
@@ -0,0 +1,159 @@
...
+
+  /**
+   * MDChips Directive Definition
+   * @param $mdTheming
+   * @param $log
+   * @ngInject
+   */
+  function MdChips ($mdTheming, $log) {
+    return {
+      template: function(element, attrs) {
```
ThomasBurleson: Is this a variation of using the transclude function? Not sure what trickery is happening here?

src/components/chips/js/chipsDirective.js
```diff
@@ -0,0 +1,159 @@
...
+      restrict: 'E',
+      controller:   'MdChipsCtrl',
+      controllerAs: '$mdChipsCtrl',
+      bindToController: true,
+      compile: compile,
+      scope: {
+        readonly:             '=readonly',
+        placeholder:          '@',
+        secondaryPlaceholder: '@',
+        mdChipAppend:         '&'
```
ThomasBurleson: should this just be a **mdChipChange** with the $event specify an  add, deleted, or modified indicator?

ThomasBurleson:
@typotter - With the exception of my questions, very cool looking.  :+1:

src/components/chips/js/chipsController.js
ThomasBurleson: @jelbourn - The value chipBuffer is initialized to '' and then cleared to ''. So for consistency, I thought a valid string check was appropriate.

src/components/chips/js/chipsController.js
jelbourn: Personally I'm a fan of this style, but it's not a huge difference.

ThomasBurleson: @jelbourn - when we go to Angular Material 2 with ES6, we will not be using the `<class>.prototype = { ... }` approach any longer. So perhaps it is a moot standard for Angular Material 1.x with ES5.

jelbourn: I'm fine with whatever (and technically Material 2 is typescript...)

typotter: I've written a lot of Google-Closure js, where Class.prototype.method = ... is the rule. Not sure how strongly we're adhering to a styleguide here in Angular (and if there is a written guide in the first place)... I've seen 2 or 3 different ways of defining class methods in this repo now. The style I've used here is just what I'm used to and so it looks the best to me. Ultimately, my preference is towards whatever is the most readable, so if you can read my code, and I don't have to change it, I'm not going to :)

src/components/chips/js/chipsDirective.js
typotter: @jelbourn can probably elaborate on the trickery since he showed me this trick... The template property allows for a function instead of a string which takes the original HTML of the component and its attributes as parameters. Via Jeremy's trick, I am storing the original HTML in a hidden attribute so that I can do some of my own manual trans- and in-clusion of child elements. In this case, I am only including (i.e. the hunk of HTML I take from the caller and include in the final rendering takes on the scope of its destination). In a following change, I will be doing a poor-man's transclude of an md-autocomplete or input element, by $compiling its HTML with the component scope's parent scope.

src/components/chips/js/chipsDirective.js
typotter: I don't think so - it's actually meant to be a data-transforming function and not a 'notification' method, so I don't think using an $event is appropriate. I didn't want to have this method in the first place, but haven't figured out a way around it. Specifying this method allows callers to create objects from the input string, so that an object is appended to the list of chips. I'd rather md-chips be used in conjunction with md-autocomplete when a list of objects is desired, as md-autocomplete itself works with lists of objects (or strings).
That being said, I'm not sure yet how "chip editing" is going to play out and whether it will involve this transform as well.
So: I am more than open to suggestions for a new name. md-transform-string-to-object-for-chip-append-expression is too long. And, actually, a misnomer, since callers are able to provide the method even when strings are being used - they could, for example, use it to convert string chips to UPPERCASE..

I've added a comment to the directive indicating its WIP nature and that stuff may break. This officially gives me license to change stuff, like mdChipAppend in the future :)

typotter:
Comments addressed and/or rambled on in response. rebased, squashed, and pushed. Please, take another look :)

src/components/chips/js/chipsDirective.js
ThomasBurleson: @typotter 
*  I concur that I do not like the current **mdChipAppend** name. It appears to act like a filter; but I cannot think of a better name right now.
*  For Angular Material, in general, we need to consider which components will dispatch events for which component-internal changes. This prepares use for Angular 2.x... Let's revisit the event debat in a separate issue.

PR #2062, opened 2015-03-28
style(whiteframe): add 2 layers of shadows

thienedits:
Added more depth to shadows by using 2 layers of shadows.

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2062)

### DIFF
--- a/src/components/whiteframe/demoBasicUsage/style.css
+++ b/src/components/whiteframe/demoBasicUsage/style.css
@@ -1,5 +1,5 @@
 md-whiteframe {
   background: #fff;
-  margin: 20px;
+  margin: 40px;
   padding: 20px;
 }
--- a/src/core/style/variables.scss
+++ b/src/core/style/variables.scss
@@ -31,15 +31,15 @@ $toast-margin: $baseline-grid * 1 !default;
 
 // Whiteframes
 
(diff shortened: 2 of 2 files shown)


marcysutton:
Is this intended to align with the Material design spec?

thienedits:
Hey Marcy, I couldn't find anything specific in the design spec so just used the Polymer paper-shadows as a reference for the vertical spacing and blur radius.

thienedits:
Actually just saw the Objects in 3D space section of the spec. The vertical offset should probably be updated to match the 8dp baseline grid.

marcysutton:
@thienedits was your change in response to something, like Angular Material not being close enough to the spec? Or is it just aiming for Polymer code parity?

thienedits:
@marcysutton yes a little of both. I also noticed the new google io site using 2 layers of shadows and visually, at least to me, it looks more appealing and has more depth. The only guidance I've seen in the spec is 2px(resting state), 4px(resting state of app bar), 8px(pressed button state + FAB resting state), 14px(FAB pressed state). So I guess I would assume just adding +6dp for each higher elevation. So the next would be 20px, 26px, etc. The polymer shadows do not follow this.

Issue #2072, opened 2015-03-29
gesture.js: runHandler breaks with TypeErrors if custom code extends Object


xtv:
If somewhere in your code you extend Object eg
    Object.prototype.test123 = function(x) { ...

this leads to TypeErrors in runHandlers 
TypeError: handler.cancel is not a function in gesture.js:1084 and
TypeError: handler[handlerEvent] is not a function in gesture.js:1086

Reason is that the result from the "for in" loop on HANDLERS is not checked for valid returns..

You can see this in trivial test case http://plnkr.co/edit/AC4doGt9o1bPTmRgbLgg

I prepare a pull request for this trivial fix

PR #2074, opened 2015-03-29
fix(runHandlers): check if handler returned is not undefined

xtv:
runHandlers performs a for in loop to retrieve all properties, but does not check
    it HANDLERS contains a handler with that key - which breaks if user code extends
    Object.

   fixes https://github.com/angular/material/issues/2072


xtv:
travis checks failed due to github issues:

fatal: unable to connect to github.com:
github.com[0: 192.30.252.131]: errno=Connection timed out

PR #2092, opened 2015-03-31
fix (select.js) : allow user type to focus on option

wenbolovesnz:
BREAKING CHANGE: md-select now must have a value attribute

    Closes #17040

### DIFF
--- a/angular-material.css
+++ b/angular-material.css
@@ -1,0 +1,6692 @@
+/*!
+ * Angular Material Design
+ * https://github.com/angular/material
+ * @license MIT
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
@rschmukler - Can you review this an merge if ready/appropriate?

rschmukler:
@wenbolovesnz I ended up implementing this in a slightly different way so that the same code could be used to allow both the focusing of options, as well as the selecting of them when the option menu was open.

Thanks for the contribution, your code definitely helped me get started on it.

PR #2103, opened 2015-03-31
fix(2081,2098): style changes/fixes

typotter:
convert chip remove to element directive, simplifying implementation and use an icon instead of text
use rems for style instead of px
fix min-width issue on remove button
add selected/focus styling to chips
remove crowded div.md-chip > chip

### DIFF
--- a/src/components/chips/chips-theme.scss
+++ b/src/components/chips/chips-theme.scss
@@ -1,6 +1,10 @@
 md-chips.md-THEME_NAME-theme {
   .md-chip {
-    color: '{{background-contrast}}';
+    color: '{{background-contrast-0.6}}';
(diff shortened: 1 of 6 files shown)


Frank3K:
I think the test failed since the template now renders a `md-chip`-element instead of a `div`-tag [(source)](https://github.com/angular/material/pull/2103/files#diff-9adcd92319edf96bf70db5d131a9d755L80).

The function `getChipElements` won't return these elements [(source)](https://github.com/angular/material/blob/713f7b67cd195d5abfde1540f26679d58f738bb5/src/components/chips/chips.spec.js#L24-L26).

typotter:
whoops. You're right, Frank! fixing now.

Issue #1891, opened 2015-03-12
element.focus() in $mdSidenav causes page jump


nolandubeau:
Hi,  
I am building an application which has a layout similar to the angular material design site.  At the very top of the page however, is another md-toolbar.  Below the first toolbar is the sidenav and the main content area.

With a layout like this when the sidenav is finished opening, element.focus() is called.  This causes the page to focus to the sidenav toolbar top, and causes the toolbar above the sidenav to be partially visible (since the page has scrolled).  Is it really necessary to have the element.focus() in the updateIsOpen function?  If I comment out element.focus(), the issue is fixed.

if (scope.isOpen) {
  //element.focus();
}

marcysutton:
Focus is necessary for accessibility purposes, so that screen reader users are directed into sidenav. Sounds like we may need to do some CSS wrangling to support both use cases. Can you provide a Codepen or Plunker with your markup showing the jump?

DavidVaini:
I am also having this issue, I do not want it to snap, is there a way to override this behavior? I do not want the sidenav to "snap" causing the page to jump.

DavidVaini:
I can confirm that if you comment out that line it stops the focus, however this isn't a solution for me as modifying angular itself will not work with our deployment strategy (not to mention its bad practice).

DavidVaini:
Perhaps a fix like this would work:
```
              var x = window.scrollX, y = window.scrollY;
              elemment.focus();
              window.scrollTo(x, y);
```

marcysutton:
Smooth scrolling is something I've done in the past. That makes it so you can still focus an element and while smoothing out the jump.

DavidVaini:
Except thats not the desired effect, why force scrolling? could we have it as an optional config property? I think most people would just want the side nav to slide out and thats it.

I would say if people want it to scroll smoothly down, they could just write their own javascript to do it. 

It just seems to go away from being extensible if you force scrolling. You could implement no scrolling/no snapping and then if someone wants to smooth scroll, they could add an event to handle it. Just my 2 cents.

marcysutton:
The scrolling would only be added to smooth out `focus`, which is required for the sidenav component. I'm looking into how to make this better while still keeping it accessible for keyboard and screen reader users.

DavidVaini:
I guess, I wonder why its required to scroll, to me it would make sense to have two behaviors. One that doesn't scroll but still focuses, and one that focuses and scrolls. Could bet set by a config proerpty on the slidenav.

[...]

PR #2108, opened 2015-03-31
feat(sidenav): add focus override option and demo

marcysutton:
Developers want to be able to focus on a different in the sidenav, so this PR adds a `md-no-focus` attribute that will allow you to override the default behavior (focus is otherwise sent to the sidenav wrapper). 

To specify a different element to focus on open, a handler has to be added with JavaScript (see Sidenav demo). It didn't seem like a good idea to bake a selector into the HTML (the API for sidenav), so I provided a simple opt-out method instead.

There was discussion about making it also smooth scroll to handle the "jump" on focus (see issue #1891), but this change only includes the override.

### DIFF
--- a/src/components/sidenav/demoBasicUsage/index.html
+++ b/src/components/sidenav/demoBasicUsage/index.html
@@ -26,6 +26,9 @@
         <p>
         The left sidenav will 'lock open' on a medium (>=960px wide) device.
         </p>
+        <p>
+        The right sidenav will focus on a specific child element.
+        </p>
 
         <div>
           <md-button ng-click="toggleLeft()"
@@ -52,6 +55,13 @@
         <h1 class="md-toolbar-tools">Sidenav Right</h1>
       </md-toolbar>
       <md-content ng-controller="RightCtrl" class="md-padding">
+        <form>
+          <md-input-container>
+            <label for="testInput">Test input</label>
+            <input type="text" id="testInput"
+                   ng-model="data" md-sidenav-focus>
+          </md-input-container>
+        </form>
         <md-button ng-click="close()" class="md-primary">
           Close Sidenav Right
         </md-button>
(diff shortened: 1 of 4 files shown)


dmackerman:
Yes, please!

ThomasBurleson:
@marcysutton - I do not think this is the best approach for a solution. Perhaps a `md-sidenav-focus` would be a better approach that does not require a DOM query in the controller.

src/components/sidenav/demoBasicUsage/index.html
```diff
@@ -46,12 +49,18 @@ <h1 class="md-toolbar-tools">Sidenav Left</h1>
...
+    <md-sidenav class="md-sidenav-right md-whiteframe-z2" md-no-focus="true" md-component-id="right">
 
       <md-toolbar class="md-theme-light">
         <h1 class="md-toolbar-tools">Sidenav Right</h1>
       </md-toolbar>
       <md-content ng-controller="RightCtrl" class="md-padding">
+
+        <md-input-container>
+          <label>Test input</label>
+          <input ng-model="data" id="focusInput">
```
ThomasBurleson:
```html
<input ng-model="data" md-sidenave-focus >
```

ThomasBurleson:
@marcysutton - can you rebase plz.

src/components/sidenav/sidenav.spec.js
```diff
@@ -61,6 +61,43 @@ describe('mdSidenav', function() {
       expect($document.activeElement).toBe(el[0]);
     }));
 
+    it('should focus child directive with md-sidenav-focus', inject(function($rootScope, $animate, $document, $compile) {
```
ThomasBurleson: @marcysutton - love the unit test for the new directive. :+1:

ThomasBurleson:
@marcysutton - love the demo. that input focus is very cool!

PR #2158, opened 2015-04-04
feat(mdCard): support md-card-footer

### LINKED ISSUE #1474: md-card: Add a card footer
Currently you can put your content outside of `md-card-content` and it will not have padding. But it always put(I assume with a simple ng-transclude) above the content within `md-card-content` even if you add html below it.

Example
```
<md-card>
  <h1>My header</h1>
  <md-card-content>
    <p>Lorem ipsum ....</p>
  </md-card-content>
  <span>This should be below the content but it ends up just below the h1</span>
</md-card>
```

The current hack to have a footer is to set `padding-left` and `padding-right` to 0 for `md-card-content` and then to put your footer within it and add a div with the original padding for your real content. Perhaps an `md-card-footer` directive could be added. Or elements below the `md-card-content` could "just work" like how the header works now.

marcysutton:
Closes #1474

### DIFF
--- a/docs/app/partials/menu-link.tmpl.html
+++ b/docs/app/partials/menu-link.tmpl.html
@@ -1,7 +1,7 @@
 <md-button ng-class="{'active' : isSelected()}"
   ng-href="#{{section.url}}" ng-click="focusSection()">
   {{section | humanizeDoc}}
-  <span class="visually-hidden"
+  <span class="md-visually-hidden"
     ng-if="isSelected()">
     current page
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
@marcysutton - can you rebase plz.

marcysutton:
@ThomasBurleson I think I forgot to close this one, it was already added in ff794ecad8299f343d79501b470a914a75cdb6bc.

PR #2163, opened 2015-04-04
Fix for flexgrid position in safari

amay0048:
Native calc isn't supported in safari, it requires the -webkit-calc attribute


amay0048:
I signed it!

PR #2164, opened 2015-04-04
Fix for flexgrid position in safari

amay0048:
unbalanced brackets and space between brackets means safari wont apply the CSS


ThomasBurleson:
Fixed in 33049cff96cd643e76159fc2d72cbd9fb911245c

PR #2204, opened 2015-04-08
fix(input): delegate clicks in IE10

### LINKED ISSUE #1394: IE10 Placeholder blocking input focus
With the placeholder value set it seems to be blocking the clickability of the input field. This can be seen in the demo as well [First Name](https://material.angularjs.org/#/demo/material.components.input#input_008).  I don't like the fix i have locally but i added a z-index: 0; to my placeholder as a work around.  I attempted to bind a click event to the placeholder but it seems that it was being caught somewhere on the chain.  Not sure if IE9 is a worry but similar issue is occurring with the label click, this i solved with a
```javascript
element.bind('click', function(ev) {
  containerCtrl.setFocus(true);
  containerCtrl.input.focus();
  //any other existing code
});
```

marcysutton:
`pointer-events: none` does not work in IE10, so the placeholder element was blocking clicks on the input.

Closes #1394

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -61,6 +61,9 @@ function mdInputContainerDirective($mdTheming, $parse) {
 
     self.isErrorGetter = $attrs.mdIsError && $parse($attrs.mdIsError);
 
+    self.delegateClick = function() {
+      self.input.focus();
+    };
     self.element = $element;
     self.setFocused = function(isFocused) {
       $element.toggleClass('md-input-focused', !!isFocused);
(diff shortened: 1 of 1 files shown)


src/components/input/input.js
```diff
@@ -325,7 +328,9 @@ function placeholderDirective() {
     var placeholderText = attr.placeholder;
     element.removeAttr('placeholder');
 
-    inputContainer.element.append('<div class="md-placeholder">' + placeholderText + '</div>');
+    var placeholder = '<div class="md-placeholder" ng-click="delegateClick()">' +
```
ThomasBurleson: What happens if `placeholderText == undefined`? 
What is the sizing of the `md-placeholder`.
Is usable for all our input demos on browsers ?

marcysutton: This is only a fallback for when the `placeholder` attribute is present, so it won't even be there otherwise. All it's doing is delegating a click to the input, and it worked everywhere I checked (multiple versions of IE, Firefox, Chrome, Safari, Mobile Safari). I need to check Android, but given how minor this thing is, I don't see how it would be an issue.

Issue #1984, opened 2015-03-20
layout-margin/padding: Add gt-* varients


epelc:
It would be nice if we had `layout-margin-gt-md` etc. Some things end up with to much margin/padding on smaller devices.

gustavohenke:
Related: #1281 #1322

marcysutton:
@epelc do you mean you want the amount of margin to change depending on the width of the device? Will the fix in this PR cover your use case? https://github.com/angular/material/pull/1322

Side note: the spec has specific metrics for margins independent of device width, so I don't think they should really change cross-device. If there is an actual bug with the margin/padding attributes, though, like the flex variant one...that should be fixed.

epelc:
@marcysutton I meant only setting the margin for specific media queries but I think it changing automatically is a better system. For example you probably don't need as much margin on a small device as you do a large one.

marcysutton:
I definitely think specific margin/padding values should be handled inside the framework CSS so they align with the Material Design metrics. The only cross-device adjustment I could find from looking at the [spec](http://www.google.com/design/spec/layout/metrics-keylines.html#metrics-keylines-keylines-spacing) was that margins are `16px` on mobile, `24px` on tablet, and nothing is specified for desktop.

PR #2221, opened 2015-04-09
refactor(): change .md-padding to layout-padding

### LINKED ISSUE #1304: md-content: Remove `md-padding` class in favor of `layout-padding`
I'd like to suggest we remove the `md-padding` class for `md-content` as it is essentially the same as the `layout-padding` attribute. It seems like it might confuse people and having them try to use `md-padding` on other elements when we already have `layout-padding` and `layout-margin` for that.

marcysutton:
This PR changes all `<md-content class="md-padding">` instances in the docs to use `[layout-padding]` instead. 

We discussed making `md-content` padded by default given the frequency of use in our docs, but I decided against it once I saw how many things it affected. `md-content` elements are frequently nested, so it undesirably stacked margins against each other. 

One outstanding question I have is about the default margin/padding for `md-content`: currently `[layout-padding]` uses the baseline grid of `8px`, but I see `16px` used much more often. IMO, the framework should have one attribute that adjusts its size in SCSS depending on the breakpoint (less margin on small screens). That way, spacing would be consistent across all Angular Material apps. Refer to #1984 and #1322.

Closes #1304

### DIFF
--- a/docs/config/template/index.template.html
+++ b/docs/config/template/index.template.html
@@ -120,7 +120,7 @@
 
     </md-toolbar>
 
-    <md-content ng-view md-scroll-y flex class="md-padding"></md-content>
+    <md-content ng-view md-scroll-y flex layout-padding></md-content>
 
   </div>
   <script>
--- a/docs/guides/component/_name_/demoBasicUsage/index.html
+++ b/docs/guides/component/_name_/demoBasicUsage/index.html
@@ -1,6 +1,6 @@
 
 <div ng-controller="AppCtrl">
-  <md-content class="md-padding">
+  <md-content layout-padding>
 
   </md-content>
 </div>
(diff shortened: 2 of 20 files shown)


ThomasBurleson:
@marcysutton - these now look bad:

*  http://localhost:8080/#/demo/material.components.bottomSheet
*  http://localhost:8080/#/demo/material.components.sidenav
*  http://localhost:8080/#/demo/material.components.subheader
*  http://localhost:8080/#/demo/material.components.toolbar

Can you resolve as part of this PR?

marcysutton:
@ThomasBurleson those fixes have all been included. I left `layout-padding` at 8px because it affected a bunch of things, but I do think we need a solution for other larger margins/padding units. That should be a separate discussion we have as a team.

PR #2278, opened 2015-04-12
fix(gestures): attach event listeners once when multiple ng-apps are running.

### LINKED ISSUE #2173: $mdGesture: only attach event listeners once, even if multiple ng-apps run

kennethcachia:
Closes #2173.


kennethcachia:
CLA signed.

PR #2300, opened 2015-04-13
fix(build): only uglify js files

xlc:
https://github.com/angular/material/commit/0e916bfccbd2abd05508c6bde61eb513530c3331 add images to  src/components/list folder, which was assumed (by gulpfile) only contains .js and .css files. So the gulp task was attempt to uglify .jpeg files and throw error.


ThomasBurleson:
Build systems significantly enhanced for 0.9.4.

PR #2304, opened 2015-04-14
fix(gridlist): The gridlist will now lay out everytime a tile is added

### LINKED ISSUE #2227: mdGridList: md-grid-tile elements are gone when the grid list is re-created
First of all, I am using the latest builds from: http://rawgit.com/angular/bower-material/master/angular-material.js and this issue was observed on Chrome v41.

We are dynamically creating mdGridList tiles, and refreshing those tiles from script was working as expected before. However today, when we try to refresh them as before, tiles cannot be seen anymore. I am not sure when this was broken, but I am certain that it was okay on April 7th.

Codepen for the issue: http://codepen.io/ecem/pen/myZVgv, normally it's looking good, but when we click refresh the tiles are gone.

shyndman:
Fixes #2227

### DIFF
--- a/src/components/gridList/demoDynamicTiles/index.html
+++ b/src/components/gridList/demoDynamicTiles/index.html
@@ -1,5 +1,4 @@
 <div ng-app="gridListDemo" ng-controller="gridListDemoCtrl as vm" flex>
-  <md-button ng-click="vm.rotate()">Rotate</md-button>
   <md-grid-list
         md-cols-sm="1" md-cols-md="2" md-cols-gt-md="6"
         md-row-height-gt-md="1:1" md-row-height="4:3"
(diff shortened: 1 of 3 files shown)


src/components/gridList/demoDynamicTiles/index.html
```diff
@@ -1,5 +1,4 @@
 <div ng-app="gridListDemo" ng-controller="gridListDemoCtrl as vm" flex>
-  <md-button ng-click="vm.rotate()">Rotate</md-button>
```
ThomasBurleson: Why the removal of the 'rotate' feature? Did you consider it too confusing for visitors?

shyndman: Yeah, didn't think it added much to the demo, but if you think otherwise I can add it back.

ThomasBurleson: No strong opinion either way.

shyndman: Cool. Let's leave it out.

PR #2325, opened 2015-04-15
update(docs): fix bug in input directive example

rawlink:
Fixed a bug in the input directive documentation example.


rawlink:
Signed CLA agreement.

PR #2371, opened 2015-04-16
Fix Select

vfrico:
When select is open and you choose an option wich will do
a route change, it is unable to find parent to delete.

This causes an exception and the backdrop generated never closes

EDIT: Backtrace: http://dpaste.com/3P3XWGG

The solution is destroy first the backdrop and then try to find the
parent and delete it.


vfrico:
I signed it!

marcysutton:
@rschmukler can you review?

PR #2374, opened 2015-04-17
demo: Fixed invalid fab button link

Eun:
Removed `href` attributes, so you can see the shiny animation :smile:


Eun:
I signed it!

Eun:
I signed it!

PR #2395, opened 2015-04-17
fix(list): add ripple, inherit from .md-button

marcysutton:
Updates interaction styles of list controls to attach ripple on click of list items with proxy controls, and makes child primary & secondary icon buttons into `md-button` elements. This allows for correct icon button dimensions and mouse/focus behavior to be extended to list controls.

### DIFF
--- a/src/components/list/list.js
+++ b/src/components/list/list.js
@@ -10,6 +10,7 @@
 angular.module('material.components.list', [
   'material.core'
 ])
+  .controller('MdListController', MdListController)
   .directive('mdList', mdListDirective)
   .directive('mdListItem', mdListItemDirective);
 
@@ -46,7 +47,6 @@ function mdListDirective($mdTheming) {
     }
   };
 }
-
 /**
  * @ngdoc directive
  * @name mdListItem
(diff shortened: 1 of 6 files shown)


ThomasBurleson:
The click on the switch does not ripple the row... yet the checkbox-click does. Bug ?

ThomasBurleson:
I really like this UX now.

src/components/list/list.js
```diff
@@ -189,6 +198,8 @@ function mdListItemDirective($mdAria, $mdConstant) {
         function computeClickable() {
           if (proxies.length || $element[0].firstElementChild.hasAttribute('ng-click')) {
             $element.addClass('md-clickable');
+
+            ctrl.attachRipple($scope, angular.element($element[0].querySelector('.md-no-style')));
```
ThomasBurleson: Why `.md-clickable` and `.attachRipple()` ?

marcysutton: I think the only thing `.md-clickable` does is add the pointer cursor in CSS.

marcysutton:
The "Clickable Items with Secondary Controls" demo shows the switch as a secondary item, so it wouldn't highlight the whole row on click. It contrasts with the first controls demo, where the checkbox is the only action, and it can be toggled by clicking on the row.

Issue #1242, opened 2015-01-20
mdInputContainer: Label should be enforced


marcysutton:
The new `<md-input-container>` requires developers to include native elements in their markup. This change was made to account for the many attributes and race conditions that made transclusion impossible. As a result, developers are able to implement `mdInputContainer` without a `<label>` element or `aria-label` attribute on the input, resulting in inaccessible forms. To encourage developers to create accessible inputs, Angular Material should run the `$mdAria.expect()` utility. This first checks for text to use for an [accessible name](http://www.w3.org/TR/wai-aria/roles#namecalculation_header) for the input (which is required for assistive technologies), and if no text is found, it logs a warning to the console about a missing label. 

**Prerequisites: `$mdAria` needs refactoring to work inside of `ng-repeat` and `ng-if`.**
 * Refactor to use `compile`: https://github.com/angular/material/issues/608
 * `ng-if`: https://github.com/angular/material/issues/1094
 * `ng-repeat`: https://github.com/angular/material/issues/507

matthewrfindley:
Merged and fixed in https://github.com/angular/material/pull/2397

PR #2397, opened 2015-04-17
fix(input): expect md-input-container to have aria-label

matthewrfindley:
Bug fix for https://github.com/angular/material/issues/1242.

See issue for requirements.

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -145,7 +145,7 @@ function labelDirective() {
  *
  */
 
-function inputTextareaDirective($mdUtil, $window) {
+function inputTextareaDirective($mdUtil, $window, $mdAria) {
   return {
     restrict: 'E',
     require: ['^?mdInputContainer', '?ngModel'],
@@ -163,6 +163,10 @@ function inputTextareaDirective($mdUtil, $window) {
       throw new Error("<md-input-container> can only have *one* <input> or <textarea> child element!");
     }
     containerCtrl.input = element;
+
+    if(!containerCtrl.label) {
+      $mdAria.expect(element, 'aria-label', element.attr('placeholder'));
+    }
 
     element.addClass('md-input');
     if (!element.attr('id')) {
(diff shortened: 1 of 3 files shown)


src/components/input/input.js
```diff
@@ -138,6 +140,12 @@ function labelDirective() {
  *       <div ng-message="md-maxlength">That's too long!</div>
  *     </div>
  *   </md-input-container>
+ *   <md-input-container>
+ *     <input aria-label='title' ng-model='title'>
```
marcysutton: I'd be surprised if anyone used `md-input-container` without a placeholder *or* label element. Maybe we should show it with placeholder and aria-label?

src/components/input/input.js
```diff
@@ -110,6 +110,8 @@ function labelDirective() {
  *
  * @param {number=} md-maxlength The maximum number of characters allowed in this input. If this is specified, a character counter will be shown underneath the input.<br/><br/>
  * The purpose of **`md-maxlength`** is exactly to show the max length counter text. If you don't want the counter text and only need "plain" validation, you can use the "simple" `ng-maxlength` or maxlength attributes.
+ * @param {string=} aria-label Aria-label is required when no label is present.  A warning message will be logged in the console if not present.
```
marcysutton: This should probably read "`aria-label` is required when no label element is present. `mdInput` will attempt to copy `placeholder` content to `aria-label`. If neither is present, a warning will be logged to the console."

src/components/input/input.spec.js
```diff
@@ -4,9 +4,9 @@ describe('md-input-container directive', function() {
   function setup(attrs, isForm) {
     var container;
     inject(function($rootScope, $compile) {
-      container = $compile((isForm ? '<form>' : '') + 
-          '<md-input-container><input ' +(attrs||'')+ '><label></label></md-input-container>' +
-          (isForm ? '<form>' : ''))($rootScope);
+      container = $compile((isForm ? '<form>' : '') +
```
marcysutton: Just a heads up to watch for whitespace changes in commits...we try to avoid them.

src/components/input/input.spec.js
```diff
@@ -159,4 +159,18 @@ describe('md-input-container directive', function() {
...
+
+    var el = $compile('<form name="form">' +
+                      '<md-input-container md-no-float>' +
+                      '<input placeholder="baz" md-maxlength="max" ng-model="foo" name="foo">' +
+                      '</md-input-container>' +
+                      '</form>')($rootScope);
+
+    $rootScope.$apply();
+
+    expect($mdAria.expect).toHaveBeenCalledWith(jasmine.any(Object), 'aria-label', 'baz');
```
marcysutton: An alternative assertion would be:
```js
var input = el.find('input');
expect(input.attr('aria-label')).toBe('baz');
```

PR #2418, opened 2015-04-20
Apply md-dialog styles only to direct child md-content element

breandr:
This prevents the style being applied to other md-content elements.

### DIFF
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -33,7 +33,7 @@ md-dialog {
   display: flex;
   flex-direction: column;
 
-  md-content {
+  > md-content {
     order: 1;
     padding: $baseline-grid * 3;
     overflow: auto;


marcysutton:
Why are there multiple `md-content` elements? This causes problems, and is not recommended as a solution. https://github.com/angular/material/issues/2514

breandr:
My bad. The multiple md-content is a result of me trying to make my
directives super reusable (and failing). This style should either be
removed or perhaps instead use :first-child-of-type (untested bit should
work?).

marcysutton:
Good idea, I'll investigate alternate selectors. Knowing the use case helps (reusable directives).

marcysutton:
Actually we're going to change `md-content` inside of a dialog to `md-dialog-content`, similar to `md-card` which has `md-card-content`. That will eliminate the problem entirely. I'll be sure to make note of it in commits and the Changelog!

Issue #2373, opened 2015-04-17
mdButton: icon buttons should have circular focus state


marcysutton:
Currently, `md-button` icon buttons have a rectangular focus and pressed state. They should instead have a circular shape that grows out from the center, similar to `md-checkbox`. This also applies to secondary `md-icon` buttons in `md-list`.

**Current:**

![rectangular icon buttons](https://cloud.githubusercontent.com/assets/1045233/7194109/50d8c6aa-e45e-11e4-8f43-118ecd109293.png)

**Updated:**

![toolbar icon buttons with rounded focus state](https://cloud.githubusercontent.com/assets/1045233/7194126/9f2d11a8-e45e-11e4-8dbe-6be397bae730.png)

ThomasBurleson:
Pending https://github.com/angular/material/pull/2460 PR fix.

PR #2460, opened 2015-04-21
feature(button): change md-icon-button to use round ripple effect

matthewrfindley:
Addresses issue -> https://github.com/angular/material/issues/2373.

A breaking change was introduced in https://github.com/angular/material/commit/0300c3af1bcb29ea8de5e9550a96775614de9df3.  The breaking change removes attach{insert component name}Behavior from $mdInkRipple to seperate the concern of configuring ripple effects for specific component types and the core ripple effect behavior.

This PR could be broken up into 1) implement feature 2) refactor inkRipple if this change is too extreme to bring into the project ATM.

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -57,7 +57,7 @@ angular
  *  </md-button>
  * </hljs>
  */
-function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria, $timeout) {
+function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
 
   return {
     restrict: 'EA',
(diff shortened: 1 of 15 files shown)


ThomasBurleson:
Hey Rob, @matthewrfindley created separate Ripple effect files to simulate inheritance-like solution. Not sure I like this approach.... what are your thoughts ?

src/components/button/button.js
```diff
@@ -57,7 +57,7 @@ angular
  *  </md-button>
  * </hljs>
  */
-function MdButtonDirective($mdInkRipple, $mdTheming, $mdAria, $timeout) {
+function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
```
robertmesserle: I think I like this approach as opposed to a bunch of boilerplate functions on the original `$mdInkRipple`

robertmesserle:
@ThomasBurleson @matthewrfindley I much prefer this approach to our original approach.  LGTM!

matthewrfindley:
@ThomasBurleson this has been rebased.

Issue #2442, opened 2015-04-21
mdAutoComplete: add option to to auto-oepn dropdown during focus on empty search input


ginader:
open Select Box
---
![screen shot 2015-04-20 at 8 31 05 pm](https://cloud.githubusercontent.com/assets/2269/7245219/2a10572c-e79c-11e4-8da5-65c1a214ee5b.png)

type in searchbox to filter or scroll through all options
---
![screen shot 2015-04-20 at 8 30 54 pm](https://cloud.githubusercontent.com/assets/2269/7245218/2a0ff584-e79c-11e4-9468-ad700b468836.png)

The correct behavior is shown here:
http://plnkr.co/edit/uHFSq9G9WA54adWAEupf?p=preview

ThomasBurleson:
So this is essentially an mdAutoComplete with an additional feature to "auto open on focus":

![screen shot 2015-04-21 at 11 37 23 am](https://cloud.githubusercontent.com/assets/210413/7257367/ceb29e9c-e81a-11e4-9904-168084ee7f38.png)

I think we can add a `md-auto-open` attribute and then developers can easily style for other visual aspects.

ginader:
:+1:

Panchitoboy:
is there a problem if you use md-min-length= 0?

robertmesserle:
@Panchitoboy @ginader @ThomasBurleson  This **should** have worked with `md-min-length="0"`; however, there was a very minor bug that prevented it from working properly.  I committed a fix for this just now.

ginader:
that's great! thanks a lot @robertmesserle @ThomasBurleson and @Panchitoboy :-)

ThomasBurleson:
Fixed with SHA da0a4f06c54248c6d73eba7767cdf06bc0225f57; see [Basic Usage Demo](https://material.angularjs.org/#/demo/material.components.autocomplete)

PR #2463, opened 2015-04-21
feat(autocomplete): add md-autoopen attribute to open dropdown on focus

ThomasBurleson:
Fixes to #2442

### DIFF
--- a/src/components/autocomplete/demoFloatingLabel/index.html
+++ b/src/components/autocomplete/demoFloatingLabel/index.html
@@ -8,6 +8,7 @@
           <input type="text"/>
         </md-input-container>
         <md-autocomplete flex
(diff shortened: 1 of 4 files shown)


robertmesserle:
@ThomasBurleson I don't think that this is necessary.  I think the ideal fix would be for this functionality to occur when `md-min-length="0"`

ThomasBurleson:
Fixed SHA da0a4f06c54248c6d73eba7767cdf06bc0225f57

gpothier:
Not sure it did work at some point, but it currently doesn't. I gather it is because on a focus event, handleQuery is called only if shouldHide() returns false, which it won't if there are currently no matches.

Issue #2285, opened 2015-04-13
Docs: backing to the homepage doesn't update the breadcrumbs


danwkennedy:
Do the following:
Go to the docs homepage
Navigate to another page (i.e. the autocomplete demo)
Click the browser's back button

The breadcrumbs don't update (see screenshot, notice how the title is "Demos > Autocomplete" and the selected navigation item is "Autocomplete")
![docsbug](https://cloud.githubusercontent.com/assets/501989/7116834/b3bf1fc4-e1c2-11e4-81fc-2d42eb30431d.png)

PR #2464, opened 2015-04-21
demo(): fix breadcrumb when using back button

matthewrfindley:
Fixes issue -> https://github.com/angular/material/issues/2285.

menu.currentSection and menu.currentPage were not updated when location changed.  This occurs because '/' will never match a registered section in the menu, causing the currentSection and currentPage to remain the same.

### DIFF
--- a/docs/app/js/app.js
+++ b/docs/app/js/app.js
@@ -273,6 +273,12 @@ function(SERVICES, COMPONENTS, DEMOS, PAGES, $location, $rootScope) {
   function onLocationChange() {
     var path = $location.path();
 
+    if (path == '/') {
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@matthewrfindley - where can I see this issue on http://material.angularjs.org ? How to reproduce?

matthewrfindley:
@ThomasBurleson The issue (#2285) has the repro steps.  I was able to reproduce both locally and on  http://material.angularjs.org.

ThomasBurleson:
@matthewrfindley - I see it now; I was not look at the Docs toolbar breadcrumbs... Thanks.

PR #2502, opened 2015-04-22
fix(dialog): spacing issue

### LINKED ISSUE #2472: Too much space between md-dialog buttons and bottom edge of dialog
According to the spec, there should be 8px of space between a button and the bottom of the dialog:

http://www.google.com/design/spec/components/dialogs.html#dialogs-specs

However, currently there are 16px of space vertically, because the 8px margin on the button and the 8px padding on the .md-actions div are stacking.

calebegg:
Remove bottom margin from an md-dialog's actions. The bottom margin was causing extra space that isn't in the Material spec.

Closes #2472

### DIFF
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -61,6 +61,7 @@ md-dialog {
     > * {
       margin-left: $baseline-grid / 2;
       margin-right: 0;
+      margin-bottom: 0;
     }
   }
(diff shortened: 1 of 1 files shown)


calebegg:
I am covered by the Google Corporate CLA according to that link. (I am a Google employee).

calebegg:
Issue was fixed separately. Thanks!

PR #2510, opened 2015-04-23
Update computeClickable() check for child length

mattoz0:
This now checks for child length to stop the browser error counter from going crazy.

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2510)


ThomasBurleson:
@mattoz0  - Please sign the CLA before I can merge

ThomasBurleson:
Not needed; @see https://github.com/angular/material/blob/master/src/components/list/list.js#L166.

Issue #2225, opened 2015-04-09
mdGridList: unexpected tile ordering


VilemP:
Hi,

we have a case where we would like to add tiles into the grid in two ways:
a) the first tile is always fixed and basically plays the role of "add new" button for the list items
b) any number of other tiles added via ng-repeat

Unfortunately it is not possible to achieve that as tiles added via ng-repeat have the scope.$parent.$index defined (unlike the others) and as a consequence, they are always put in front of the other tiles.

See this plunkr: http://plnkr.co/edit/uFFmoZMaWallf8M3PHRi?p=preview

I would suggest to allow an "order" attribute for a tile (similar to flexbox) to be able to specify desired order of the tile in the list.

BTW, we have run into similar issue with tabs (a mixture of fixed tabs + tabs created via ng-repeat)

VilemP:
Hi again... actually, the tab behaviour is different (still problematic). I have created another issue and plunkr for it #2226 
Sorry for mixing the two together.

ThomasBurleson:
mdGridList has been improved for 0.9; which closed many issues.
Please update your Plunkr with latest from #master and retest:

*  `//rawgit.com/angular/bower-material/master/angular-material.css`
*  `//rawgit.com/angular/bower-material/master/angular-material.js`

VilemP:
Hi, I have updated the plunkr, yet the order still seems to be wrong...

ThomasBurleson:
@shyndman - FYI

Stamy:
I have the same problem as well.

tkrotoff:
+1

shyndman:
This one is kind of tricky. I have to think about it for a bit.

PR #2553, opened 2015-04-24
fix(gridList): reorder tiles when ngRepeat array changes

tarnas14:
when I tried sorting an array bound to md-grid-list ngRepeat we found that items that didn't change place were displayed incorrectly - this here fixes it (created a demo to prove it - before fix the 4th item would be displayed on 2nd from beginning or end of the md-grid-list after sorting the underlying array)

### DIFF
--- a/src/components/gridList/gridList.js
+++ b/src/components/gridList/gridList.js
@@ -415,12 +415,12 @@ function GridListController($timeout) {
 }
 
 GridListController.prototype = {
-  addTile: function(tileElement, tileAttrs, idx) {
-    var tile = { element: tileElement, attrs: tileAttrs };
-    if (angular.isUndefined(idx)) {
+  addTile: function(tileElement, tileAttrs, tileViewModel) {
+    var tile = { element: tileElement, attrs: tileAttrs, vm: tileViewModel };
+    if (angular.isUndefined(tileViewModel)) {
       this.tiles.push(tile);
     } else {
-      this.tiles.splice(idx, 0, tile);
+      this.tiles.splice(tileViewModel.getCurrentIndex(), 0, tile);
     }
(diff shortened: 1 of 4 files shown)


tarnas14:
signed the CLA

ThomasBurleson:
@shyndman  - Do you have time to review this PR. I approved let me know and I will merge.
If no time, I will review later. Thx.

shyndman:
Yeah, I'll be able to take a look over the weekend.

src/components/gridList/gridList.js
```diff
@@ -747,8 +760,10 @@ function GridTileDirective($mdMedia) {
     if (angular.isDefined(scope.$parent.$index)) {
       scope.$watch(function() { return scope.$parent.$index; },
         function indexChanged(newIdx, oldIdx) {
-          gridCtrl.removeTile(element, attrs);
-          gridCtrl.addTile(element, attrs, newIdx);
+          if (newIdx === oldIdx) {
+            return;
+          }
+          gridCtrl.reorderTiles();
```
shyndman: Take the case where someone prepends a tile to the grid -- each of the tiles that already exists in the grid will have this watch trigger, and reorderTiles will be run as many times as there are tiles. This seems like a lot of work just to add a tile.

shyndman: I'll play around with a couple things. I want to roll in a fix for #2225, since it's related.

tarnas14: yes, it is computation intensive, I didn't have time to go deeper into the codebase at the time, will probably try to find a better solution today :)

shyndman:
@tarnas14 #2568 will fix your issue.

PR #2565, opened 2015-04-26
style(): fix some style inconsistencies and indentation

tkrotoff:
Partially fix styling and indentation.

Discovered this via documentation https://material.angularjs.org/#/demo/material.components.checkbox : the JS source for the Syncing example is not properly indented and it feels confusing.

Suggestion: go further and apply stricter code styling like `if( foo )` vs `if (foo)`, `function ( foo ){` vs `function(foo) {`, 2 spaces indentation... so it is easier to contribute.


ThomasBurleson:
Will not fix. Before the 1.0 Beta release, we will perform a global styling check and update.

tkrotoff:
At least, fix the indentation on checkbox syncing example: it is confusing.

PR #2567, opened 2015-04-26
chore(): rename radio-button to radioButton

tkrotoff:
Since progressCircular files and progressLinear files are named without - dashes, seems logical to name radioButton the same way.


ThomasBurleson:
Will not change.

PR #2568, opened 2015-04-26
fix(gridlist): Tile ordering improved

shyndman:
Closes #2553, #2225

* ng-repeated and static tiles can now be mixed
* Ordering should work properly once and for all. There's no more state
  maintained in the grid controller. Whenever the grid re-renders, the
  DOM is queried to determine the tiles involved


ThomasBurleson:
@shyndman  - really nice work here!

shyndman:
Thanks!

Issue #1161, opened 2015-01-11
scss bower-material support


itajaja:
I am using bower-material, Is there a way to provide the sass files along with the compiled css?

bogdandjukic:
We would love that too!
Looking forward to it, as we prefer scss over the already compiled css.

PaulMougel:
Duplicate of #381

nicbon:
#381 has been closed but I believe there are valid use cases for this request.

It is my understanding that the source repository is meant to be used by whoever wants to develop angular-material and the deployment repository is for developers that use angular-material in their own projects.

If that's the case, I would argue that there are valid uses cases for exposing the SCSS in the project builds repository. In these cases the developer doesn't want to mess with the original angular-material css but rather use the SCSS as a library to build his own classes. For instance I might want to access the SCSS variables, or use the handy mixins to build more *-sm *-gt classes.

I personally see the SCSS as being integral part of the library exposed to the developers rather than purely part of the library sources.

itajaja:
an example of something similar is done in this [repo for bootswatch](https://github.com/log0ymxm/bootswatch-scss). The use cases for this are several, of course the most important is building additional element that leverage existing mixins. Exposing just the [`core/style`](https://github.com/angular/material/tree/master/src/core/style) folder would be already a big improvement. Another use case would be to intercept and modify the variables before they are used in the material style sheets, or simply using the variables in your own style sheets.

paramburu:
+1

marcysutton:
@rschmukler should chime in on this, but I believe the reason CSS is provided in a compiled format is due to theming support, where theme CSS is generated dynamically.

pzurowski:
+1

johannesjo:
+1

johannesjo:
For those interested I extracted the styles and put them into their own repository:
https://github.com/johannesjo/angular-material-sass-files

marcysutton:
@johannesjo how will that be sustainable? It looks like there are specific filenames referenced and not all of them are present (ex. typography.scss). Also, what about the themes scss?

johannesjo:
@marcysutton the colors would have to be replaced with sass-variables/functions. Still working on it. Thank you for reporting the issue, but maybe the repos page is a better place for this. I'm glad for help and suggestions.

oste:
:+1: I am also lost on how to properly use the .scss files. There is the Gulp build process but that seems overkill for this use case. Looking forward to seeing how this one pans out :)

[...]

PR #2569, opened 2015-04-26
Copy the scss files to dist directory

tkrotoff:
See issue #1161. The solution includes 2 things:
- New file `src/angular-material.scss` that `@import` all angular-material scss files (like [Bootstrap does](https://github.com/twbs/bootstrap/blob/v3.3.4/less/bootstrap.less))
- All scss files are copied inside `dist\src` directory

Now users can write inside their main scss file:
`@import 'node_modules/angular-material/src/angular-material';`

This PR needs proper review since I've modified gulpfile.js without a full understanding of it.

(First commit "rename radio-button to radioButton" is from PR #2567)

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2569)


jgoux:
I hope this will be merged ! I have to rename the css file into scss every time I update angular-material to be able to include it in my sass workflow. And I don't have the flexibility of sass with just the css file.

ThomasBurleson:
Implemented in 720ed9096fbe62955a075041a1c43e62cba10471

PR #2571, opened 2015-04-27
fix(mdUtil): move comment nodes as well when disabling scroll

### LINKED ISSUE #2456: md-select moves DOM elements and puts them back in the wrong position
When `md-select` gets opened it rearranges the DOM and the layout goes crazy.
The layout of the site is `<body layout="row">` with 2 main children one being `<md-sidenav>` and second being the main `<div layout="column" layout-fill role="main" ui-view="main"></div>` where the Angular injects the view. `md-select` is located inside the `<div ui-view="main"`.

First screenshot is before `md-select` is opened
screen shot 2015-04-21 at 16 51 44

This screenshots is while its open
screen shot 2015-04-21 at 16 52 06

And finally when `md-select` is closed we can see the action moved the `<div ui-view="main">` to the top effectively moving the `<md-sidenav>` from left to right because of the flex layout ordering.
screen shot 2015-04-21 at 16 52 18

gustavohenke:
As comment nodes were not moved by `disableScrollAround()`, Angular could
misplace a few elements because of a forgotten comment node that points
to a directive.

Fixes #2456

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -76,7 +76,7 @@ angular.module('material.core')
       var disableStyle = $window.getComputedStyle(disableTarget[0]);
       var wrapperEl = angular.element('<div class="md-virtual-scroll-container"><div class="md-virtual-scroller"></div></div>');
       var virtualScroller = wrapperEl.children().eq(0);
-      virtualScroller.append(disableTarget.children());
+      virtualScroller.append(disableTarget[0].childNodes);
       disableTarget.append(wrapperEl);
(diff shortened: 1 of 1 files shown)


gustavohenke:
Signed.

gustavohenke:
Sorry, I incorrectly changed the file via GitHub, this patch will not work.
I'll fix and squash locally in a minute.

gustavohenke:
Now it'll be okay.

rschmukler:
Thanks @gustavohenke

PR #2585, opened 2015-04-27
enhancement(shadows): Updated to include Robertson shadows

shyndman:


ilanbiala:
What's a robertson shadow? How does this change the UI (before/after)?

shyndman:
Christian Robertson is a designer on the Material Design team.

It improves the appearance of the shadows dramatically, while keeping roughly the same weight of shadow at each depth.

PR #2592, opened 2015-04-28
fix(mdUtil): move comment nodes as well when reenabling scroll

gustavohenke:
The comment nodes were being moved to the virtual scroll container,
but they were not moved back to the disabled target when the scroll
was reenabled.

Ref: #2456 and 160df4cf436ff22b487e4691dfb3b8034cc0da19

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -159,7 +159,7 @@ angular.module('material.core')
       }
 
       return function restoreScroll() {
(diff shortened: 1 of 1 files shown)


gustavohenke:
Guys,
I noted that there aren't unit tests for such features.

Would you like me to create some? It's the second PR I send in 2 days to fix the same method.

PR #2608, opened 2015-04-29
fixed md-autocomplete loading indicator bug

anderson-arlen:
When selecting an item with noCache turned on, a loading indicator persists on the bottom of the control. To reproduce:

1. Go to https://material.angularjs.org/#/demo/material.components.autocomplete
2. Make sure the **Simulate query for results?** and **Disable caching of queries?** options are checked.
3. Select an autocompleted result from the list
4. Observe the loading indicator still present at the bottom of the control


ThomasBurleson:
@anderson-arlen - nice catch.

Issue #1550, opened 2015-02-17
mdCheckbox: wrong display state when using ngChecked


ngraef:
CodePen example: http://codepen.io/anon/pen/zxWrGP?editors=101

In the example, `ngChecked` and `ngClick` are used to track a list of multiple selected items. The `mdCheckbox`es display correct state when the model is manipulated by the standard checkboxes. However, after interacting with the `mdCheckbox`es, their display state is reversed. Using the standard checkboxes again fixes the `mdCheckbox` state.

(Related: `ngChecked` initially implemented for #535)

ngraef:
Related: check validity after the #1006 refactor

djgadd:
Confirmed, there's a not operator on the checked var inside the apply:

checkbox.js:123
```javascript
scope.$apply(function() {
      checked = !checked;
      ngModelCtrl.$setViewValue(checked, ev && ev.type);
      ngModelCtrl.$render();
});
```

Removing the ! has fixed it for me, don't know what the intention of it was and I unfortunately don't have the time to take a more in depth look, sorry! Hope this helps though.

Thanks guys.

ngraef:
@ThomasBurleson It doesn't look like 942d0b9 fixed [the demo in my initial post](http://codepen.io/anon/pen/zxWrGP?editors=101).

ThomasBurleson:
@ngraef - switch `ng-checked` to ` ng-model="exists(item, selected)"` and all works.

ngraef:
@ThomasBurleson Interesting, but that gives `ngModel:nonassign` errors (as expected). Thanks for reopening.

ngraef:
So the underlying issue is that the `ngChecked` attribute is watched by `mdCheckbox` and triggers a view value update, but then the `mdCheckbox` click listener also fires and toggles the view value. This explains why the display state is reversed and why the correct state is produced when manipulating the value outside of any `mdCheckbox` interaction (via the standard checkboxes).

A couple possible fixes with varying degrees of cleanliness:

1. Add a check to the click listener that prevents toggling of the value if `ngChecked` is being used (line 111).

 ```javascript
 function listener(ev) {
     if (element[0].hasAttribute('disabled')) return;
     if (attr.ngChecked) return;
     ...
 }
 ```

2. Set the correct value in the click listener (line 114).

 ```javascript
 scope.$apply(function() {
     var newVal = attr.ngChecked ? attr.checked : !ngModelCtrl.$viewValue;
[...]

Issue #535, opened 2014-11-04
bug(md-checkbox): ngChecked directive not working


breandr:
According to the doc:

> The checkbox directive is used like the normal angular checkbox.

But I am unable to get ngChecked working.

sudeshnaDatta:
How to use md-checkbox to select an entire row when used inside a grid

Elfayer:
+1, This is a real big issue. It's a regression from angularJS.

DarkMerge:
+1

kamikaze942:
+1

ThomasBurleson:
@Elfayer - are you saying that this issue has not been fixed with the latest codebase in `master`? If that is the case, please submit a new issue with a codepen demo. thank you.

PR #2619, opened 2015-04-29
feat(checkbox): handle 'checked' property + emit 'change' event

tkrotoff:
A checkbox is not a checkbox until:
- it emits the 'change' event each time a user interacts with it
- it exposes its state via the 'checked' property (true or false - never undefined) + listen to any programmatic change of 'checked'.

This is what [the regular HTML checkbox](http://www.w3.org/TR/html-markup/input.checkbox.html) (`<input type="checkbox">`) does.

This PR is somewhat related to #535 and #1550.

How others manage this?
- [Twitter Bootstrap](http://getbootstrap.com/css/#forms-example) uses directly the regular HTML checkbox: `<input type="checkbox">` => no problem: all W3C spec is therefore implemented
- [Materialize](http://materializecss.com/forms.html#checkbox) creates a new beautiful checkbox and binds it to the regular HTML checkbox which is hidden (with `position: absolute; left: -9999px`) => no problem: all W3C spec already implemented
- [React material ui](http://material-ui.com/#/components/switches) works the same way as Materialize: binds to and hides (`opacity: 0`) the regular HTML checkbox
- [awesome-bootstrap-checkbox](http://flatlogic.github.io/awesome-bootstrap-checkbox/demo/) again binds to and hides (`opacity: 0`) the regular HTML checkbox
- [Polymer](https://github.com/PolymerElements/paper-checkbox/blob/v0.8.3/paper-checkbox.html#L111) does not reuse the regular HTML checkbox and reimplements it => much like Angular Material. And they do handle the 'checked' property and fire the 'change' event (very clear implementation btw).

The way Angular Material does things is probably cleaner than hidding the regular checkbox input, but then it has the responsibility to implement the W3C specifications. This is also true whenever Angular Material reimplements a standard HTML element: md-radio-button (code factorization is probably possible). For md-button this does not apply since it wraps a regular `<button>`.

Also I've noticed that you [verify](https://github.com/angular/material/blob/v0.9.0-rc3/src/components/checkbox/checkbox.js#L139) the `checked` "attribute":
`var viewValue = attr.ngChecked ? attr.checked : !ngModelCtrl.$viewValue;`
`checked` should be a DOM node object property and thus you don't always see it in the HTML.

You also verify the `ngChecked` attribute: this is wrong. This is intended for the user not for the directive implementation. The purpose of `ng-checked` is to set `checked`.

Here a demo: http://plnkr.co/edit/yXJ7nn?p=preview

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2619)

### DIFF
--- a/src/components/checkbox/checkbox.js
+++ b/src/components/checkbox/checkbox.js
@@ -83,6 +83,21 @@ function MdCheckboxDirective(inputDirective, $mdInkRipple, $mdAria, $mdConstant,
             ngModelCtrl.$setViewValue.bind(ngModelCtrl)
         );
       }
+
+      // Initial state
+      // Like standard HTML checkbox, 'checked' property should be false by default
+      element[0].checked = false;
+
+      // Watch external changes made to 'checked' property
+      scope.$watch(
+        function() {
+          return element[0].checked;
+        },
+        function(newValue, oldValue) {
+          ngModelCtrl.$setViewValue(newValue);
+        }
+      );
+
       $$watchExpr('ngDisabled', 'tabindex', {
         true: '-1',
         false: attr.tabindex
@@ -141,10 +156,15 @@ function MdCheckboxDirective(inputDirective, $mdInkRipple, $mdAria, $mdConstant,
           ngModelCtrl.$setViewValue( viewValue, ev && ev.type);
           ngModelCtrl.$render();
         });
+
+        // Emit 'change' event
+        element.triggerHandler('change');
       }
 
       function render() {
-        if(ngModelCtrl.$viewValue) {
+        element[0].checked = ngModelCtrl.$viewValue;
+
+        if (element[0].checked) {
           element.addClass(CHECKED_CSS);
         } else {
           element.removeClass(CHECKED_CSS);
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@tkrotoff - can you rebase this PR with latest code?

tkrotoff:
@ThomasBurleson done
Keep in mind that this commit does not include fixes about all my comments (e.g. checked, ngChecked, this [comment](https://github.com/angular/material/commit/973a2fcadfc0fb786635cabf7c6674e5027117c4#commitcomment-10980270), md-radio-button, code factorization...)

src/components/checkbox/checkbox.js
```diff
@@ -141,10 +156,15 @@ function MdCheckboxDirective(inputDirective, $mdInkRipple, $mdAria, $mdConstant,
...
+
+        // Emit 'change' event
+        element.triggerHandler('change');
       }
 
       function render() {
-        if(ngModelCtrl.$viewValue) {
+        element[0].checked = ngModelCtrl.$viewValue;
+
+        if (ngModelCtrl.$viewValue) {
```
ThomasBurleson: more consistent to use `element[0].checked` here.

tkrotoff: @ThomasBurleson done

doppio:
+1 on this, would really like to see it integrated. I just tried doing it myself on my own branch, but I'm having problems building Angular (#4791) and can't test it out myself

EladBezalel:
@tkrotoff can you please rebase?

EladBezalel:
closing due to inactivity

PR #2631, opened 2015-04-30
Backdrop dialog fix (fixes #1967)

### LINKED ISSUE #1967: md-dialog backdrop displays incorrectly if page is scrolled

Kiougar:
This fixes the #1967. [Example 1](http://codepen.io/anon/pen/LVVbay) (comment the first css line to see the problem)

Also, it fixes another issue with the sidenav toggle where the backdrop element caused the body to overflow. [Example 2](http://codepen.io/anon/pen/waaoyX) (comment the second css line to see the problem)

### DIFF
--- a/src/components/backdrop/backdrop.scss
+++ b/src/components/backdrop/backdrop.scss
@@ -15,7 +15,7 @@ md-backdrop {
 
   background-color: rgba(0,0,0,0);
 
-  position: absolute;
+  position: fixed;
   height: 100%;
   left: 0;
   right: 0;
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -2,7 +2,7 @@
   display: flex;
   justify-content: center;
   align-items: center;
-  position: absolute;
+  position: fixed;
   top: 0;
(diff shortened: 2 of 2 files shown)


Kiougar:
Will remake the pull request when I fix the commit messages.

ThomasBurleson:
@Kiougar - whenever I see CSS changes to use 'position: fixed' I worry about sideaffects.
@rschmukler - what are your thoughts on this PR?

Kiougar:
@ThomasBurleson The second example I provided can be solved by setting `top: 0` instead of `position: fixed`. But that won't solve the #1967 issue.
Another solution for the #1967 could be to manually update the `top` attribute of both the `.md-dialog-container` and `md-backdrop` element when a user scrolls vertically.

Kiougar:
@ThomasBurleson, @rschmukler Reopened in #2754. We can discuss there if there are any side effects.

PR #2657, opened 2015-05-01
fix(slider): increment the value properly when step is a decimal number

### LINKED ISSUE #2015: mdSlider if set step="0.1", when slide the slider it's value will be a long float number
```html
<md-slider flex min="0" max="20" step="0.1" md-discrete
                  ng-model="component.animation.duration"
                  aria-label="transition-duretion" id="duration-slider">
</md-slider>

<md-input-container flex>
    <input type="number" min="0" max="20" step="0.1" ng-model="component.animation.duration" aria-label="transition-duretion" aria-controls="duration-slider">
</md-input-container>
```
When I slide the slider, it's value will be a long float number like those: 
`14.700000000000001`,
`0.7000000000000001`,
`1.7000000000000002`

kennethcachia:
Closes #2015

### DIFF
--- a/src/components/slider/slider.js
+++ b/src/components/slider/slider.js
@@ -269,7 +269,9 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
     }
     function stepValidator(value) {
       if (angular.isNumber(value)) {
-        return Math.round(value / step) * step;
+        var formattedValue = (Math.round(value / step) * step);
+        // Format to 3 digits after the decimal point - fixes #2015.
+        return parseFloat(formattedValue.toFixed(3));
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/components/slider/slider.js
```diff
@@ -269,7 +269,9 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
     }
     function stepValidator(value) {
       if (angular.isNumber(value)) {
-        return Math.round(value / step) * step;
+        var formattedValue = (Math.round(value / step) * step);
+        // Format to 3 digits after the decimal point - fixes #2015.
+        return parseFloat(formattedValue.toFixed(3));
```
Kiougar: You should use `return Math.round(formattedValue * 1000) / 1000`.
JSperf: http://jsperf.com/math-round-vs-tofixed2/5

kennethcachia:
Thanks @Kiougar! Pushed an update.

mprentic-harris:
It's not clear to me that this resolves the issue. Previously `Math.round(value / step) * step` was returning `14.700000000000001` or some such value. With this fix it will now return `14.700` or `14.7`? Is this really addressing the cause of `Math.round(0.15 / 0.05) * 0.05` equalling `0.15000000000000002`? What if I have `step="0.0001"`?

kennethcachia:
Hey @mprentic-harris!
It should return ``14.7`` and addresses #2015.

PR #2661, opened 2015-05-01
docs(icon): expand md-icon description with overview and context.

jelbourn:
Expanding the `md-icon` API docs with some more overview and context based on feedback that it wasn't very clear what the purpose of `md-icon` was.

Can someone else take a quick look and make sure this sounds good?

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2661)

### DIFF
--- a/src/components/icon/iconDirective.js
+++ b/src/components/icon/iconDirective.js
@@ -17,14 +17,39 @@ angular.module('material.components.icon', [
  * @restrict E
  *
  * @description
- * The `md-icon` directive is an markup element useful for showing an icon based on a font-face
- * or a SVG. Both external SVGs (via URLs) or cached SVG from icon sets can be
- * easily loaded and used.
+ * The `md-icon` directive makes it easier to use vector-based icons in your app (as opposed to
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
:+1: Do we want to discuss [in `iconDirective.js`] these facts?

* If `md-svg-src` is used, the url is considered the lookup key for searches in the internal icon registry cache. If the cache does not have the key, the key is also used as the url for the external file.
*  $mdIconProvider` is used to register icon IDs with URLs and supports preloading of icons and icon sets. 
* `<md-icon/>` uses `$mdIcon` service to both load uncached icons and to cache loaded icon data
* `<md-icon md-font-icon="" />` usages will silently fail if the fonts have not been loaded.

> I will merge this weekend.

jelbourn:
First two points seem like implementation details that we don't need to mention. $mdIconProvider has its own docs, and the last one... I don't even know what you would expect that to do other than no-op.

gkalpak:
:+1: for `mdIcon` docs expansion !

--
> and the last one... I don't even know what you would expect that to do other than no-op

Show a default WTF icon ? :stuck_out_tongue:

Issue #1932, opened 2015-03-17
Fix sticky header in complex dialog


mtraynham:
Honestly, the base dialog is pretty bland, even if it does follow the [spec design](http://www.google.com/design/spec/components/dialogs.html#dialogs-specs).  The one shown in the docs is much nicer looking, but kind of a pain to get set up.

I suggest adding a css class to md-dialog, say, `<md-dialog class="md-complex-dialog"`, would be much more useful than multiple users trying to recreate the look and feel of the demo.

I was able to re-create that look and feel with the following css:
```
md-dialog.md-complex-dialog .md-subheader {
  margin: 0;
}
md-dialog.md-complex-dialog .md-subheader.md-sticky-clone {
  box-shadow: 0 2px 4px 0 rgba(0, 0, 0, 0.16);
}
md-dialog.md-complex-dialog > md-content {
  padding: 0;
}
md-dialog.md-complex-dialog > md-content > :not(.md-subheader) {
  padding-left: 24px;
  padding-right: 24px;
}
md-dialog.md-complex-dialog > md-content > :not(.md-subheader):first-of-type {
  padding-top: 24px;
}
md-dialog.md-complex-dialog > md-content > :not(.md-subheader):last-of-type {
  padding-bottom: 24px;
}
```

And usage in a dialog would look like:
```
<md-dialog class="md-complex-dialog" aria-label="Test Header">
    <md-content>
        <md-subheader class="md-primary">Test Header</md-subheader>
        <p>Message</p>
    </md-content>
</md-dialog>
```

epelc:
+1

marcysutton:
I'm not clear on what this is trying to solve that the framework doesn't do already. I cleaned up some of the CSS in https://github.com/angular/material/pull/2662, but before adding a CSS class I need to understand the problem. Sticky headers in dialogs are a known issue.

epelc:
I just wanted the sticky header like in the custom dialog demo.

marcysutton:
The sticky header doesn't currently work in the custom dialog demo–that feature is due for a refactor. If it did work though, it should be accomplishable without adding extra CSS.

mtraynham:
@marcysutton I created this issue because I wanted to replicate the complex dialog in the documentation.  After seeing the demo, I can guarantee that I'm not the only one trying to replicate that look and feel.  The out-of-box dialog is stark with color and really bland, which does not do well for things like forms with headers.

[...]

PR #2662, opened 2015-05-01
fix(dialog): style cleanup, action bar renaming

marcysutton:
When looking into https://github.com/angular/material/issues/1932, I found that some of the dialog styles either should have been in `dialog.scss` instead of the demo CSS or they weren't being used at all. Known issue: sticky headers in dialogs are currently not working.

This PR also updates dialogs to use `<md-dialog-actions>` instead of `<div class="md-actions">`, and cards to use `<md-card-actions>`.


marcysutton:
@ThomasBurleson I'm hesitant to merge this until our docs site is locked on a release. It will show a different HTML API for dialogs than what's in 0.9.0.

PR #2670, opened 2015-05-02
fix(list): Do not ignore spaces in descendant textareas.

### LINKED ISSUE #2668: mdListItem: spaces are ignored in descendant textarea
When using a `textarea` element inside an `md-list-item` (no proxy) all spaces are ignored. This behaviour can be observed in Chrome 42 (Windows, Ubuntu) and Internet Explorer. Due to another bug in the `mdListItemDirective`, the problem does not appear in Firefox.

The culprit is inside the [following lines](https://github.com/angular/material/blob/f66e4718451b90cd27f393a2bbc46e3e6891a61e/src/components/list/list.js#L206-L216) of the `mdListItemDirective`:

``` js
        if (!hasClick && !proxies.length) {
          firstChild.addEventListener('keypress', function(e) {
            if (e.target.nodeName != 'INPUT') {
              if (e.keyCode == $mdConstant.KEY_CODE.SPACE) {
                firstChild.click();
                e.preventDefault();
                e.stopPropagation();
              }
            }
          });
        }
```

All spaces, except those on `input` elements, are prevented. I think this should be changed to:

``` js
if (e.keyCode == $mdConstant.KEY_CODE.SPACE) {
  if (e.target.nodeName != 'INPUT' && e.target.nodeName != 'TEXTAREA') {
    ...
  }
}
```

In Firefox, the above code does not trigger since `e.keyCode` returns 0. This can be fixed by also checking `e.which`, which is also done in (for example) the [radioGroupDirective](https://github.com/angular/material/blob/977a495e7d3de27bebabdebaaf5611c9d06a1e22/src/components/radioButton/radioButton.js#L72).

Frank3K:
Fixes #2668.

### DIFF
--- a/src/components/list/list.js
+++ b/src/components/list/list.js
@@ -205,8 +205,9 @@ function mdListItemDirective($mdAria, $mdConstant, $timeout) {
 
         if (!hasClick && !proxies.length) {
           firstChild.addEventListener('keypress', function(e) {
-            if (e.target.nodeName != 'INPUT') {
-              if (e.keyCode == $mdConstant.KEY_CODE.SPACE) {
+            if (e.target.nodeName != 'INPUT' && e.target.nodeName != 'TEXTAREA') {
+              var keyCode = e.which || e.keyCode;
+              if (keyCode == $mdConstant.KEY_CODE.SPACE) {
                 firstChild.click();
                 e.preventDefault();
                 e.stopPropagation();


src/components/list/list.js
```diff
@@ -205,8 +205,9 @@ function mdListItemDirective($mdAria, $mdConstant, $timeout) {
 
         if (!hasClick && !proxies.length) {
           firstChild.addEventListener('keypress', function(e) {
-            if (e.target.nodeName != 'INPUT') {
-              if (e.keyCode == $mdConstant.KEY_CODE.SPACE) {
+            if (e.target.nodeName != 'INPUT' && e.target.nodeName != 'TEXTAREA') {
+              var keyCode = e.which || e.keyCode;
```
ThomasBurleson: @Frank3K - why are you using the **deprecated** event property `e.which` ?

Frank3K: For compatibility with Firefox, which returns 0 for `e.keyCode` here. This fallback is more than once in the material source. See my comments in the corresponding issue (#2668).

ThomasBurleson:
A unit test is also missing to validate this logic change/enhancement.

Frank3K:
Never really wrote a Jasmine unit test before. I'm stuck at generating the keydown/keypress (not sure which of the two I need here) events. The textarea does not seem to pick up the value.

When looking at other material `.spec.js` files I see them using `.triggerHandler()` or calling methods directly on the corresponding controller. The latter option can't be done here, since there is no method to call. I'm not sure if the first option should be able work here; I can't get it working yet.

I tried the following:

Added method:
```js
  function keyPress(target, key) {
    var event = document.createEvent('Event');
    event.keyCode = key;
    event.initEvent('keydown');
    target.dispatchEvent(event);
    console.log(event);
  }
```

The test itself (testing with an `a` now for clearer debugging):
```js
  it('allows spaces in a nested textarea', function() {
    var listItem = setup('<md-list-item><textarea></textarea></md-list-item>');
    var textarea = listItem.children()[0];
    keyPress(textarea, 'a'.charCodeAt(0));
    expect(textarea.value).toBe('a');
  });
```

The `textarea.value` always seems to be blank when running this test.

Anyone who can give me any pointers? I'll try fixing it tomorrow.

Frank3K:
Also tried using the following code, inspired by [checkbox.spec.js](https://github.com/angular/material/blob/48ee9867cd44f7a387e78d4a323f5a9e44382550/src/components/checkbox/checkbox.spec.js#L130-L134)

```js
  function keyPress(target, code) {
    angular.element(target).triggerHandler({
      type: 'keypress',
      keyCode: code
    });
  }
```

Which does not work, since it (as the name says) probably only triggers the attached handler(s).

Currently out of ideas on how to fix this. Any tips are appreciated.

ThomasBurleson:
@Frank3K - `browserTrigger(...)` does not work either.Give the team a day or two to investigate unit tests for space keypresses. Thx

PR #2672, opened 2015-05-02
fix(select): expose and preserve aria-label

marcysutton:
Select previously used `aria-labelledby` pointing to `<md-select-label>`, which worked fine until a value was selected, wiping out the accessible name for the select. To inform screen reader users what a particular select is for, the contents from `placeholder=""` or `<md-select-label>` are copied to an `aria-label` attribute on `<md-select>` instead. A warning is logged to the console if neither value is present.

Closes https://github.com/angular/material/issues/1893


ThomasBurleson:
@marcysutton - love your PRs with unit tests added. :+1:

PR #2673, opened 2015-05-02
fix($mdCompiler): was overwriting `resolve` and `locals` option values

### LINKED ISSUE #2614: $mdCompiler breaks if called multiple times with the same options object
This is because `$mdCompiler` modifies the source object and gets confused if it receives the same object again later. Treating the options object as immutable would solve the issue at a very small performance cost (copying properties).

I will try to submit a pull request this weekend.

rochdev:
closes #2614


rochdev:
I signed it!

Issue #2704, opened 2015-05-04
Replacing $timeout with $interval for $mdToast for protractor


adamweeks:
Since the `$mdToast` service uses `$timeout` to do the hiding delay, you are unable to e2e test toasts that have a hide delay. 

An example test that fails:
`toast.js`
```
$mdToast.show(
    $mdToast.simple()
        .content('Saved')
        .position('bottom right')
);
```
`toast.spec.js`
```
var saveAlert = element(by.css('md-toast'));
browser.wait(function(){
    return browser.isElementPresent(saveAlert);
});
expect(saveAlert.getText()).toBe('Saved');
``` 

But, if you don't let it hide, the test will pass
```
$mdToast.show(
    $mdToast.simple()
        .content('Saved')
        .position('bottom right')
        .hideDelay(0)
);
```

Protractor handles $interval's better and does not wait for the interval to finish before processing.
Protractor issue reference:
https://github.com/angular/protractor/issues/169

andrew-cunliffe:
:+1:  

Just hit this issue while laying protractor, I was thinking it was due to using cucumber.js and chaiAsPromissed, but just did the hide delay 0 and it is passing now, so will have to turn of this test till it is fixed...

itrethan:
+1 for fix

sricc:
:+1:

fedelopez:
+1

pieterwillaert:
+1

AdamMcCormick:
+1

dacarley:
+1

TylerDixon:
+1

jasonmc86:
+1

stefan--:
+1

bclin2:
+1

zaknuces:
+1. Can you please change the priority of this issue. Writing code without automation backup is neither a good idea nor recommended.

mprokopowicz:
+1

BelfordZ:
:+1:

edgarchinchilla:
+1

podoler:
+1

DevVersion:
When we're replacing `$timeout` with `$interval`, errors will appear in the Karma tests. 

This is because, `$interval` is not able to not guarantee a given delay between executions.

![image](https://cloud.githubusercontent.com/assets/4987015/14500428/60424ffa-01a2-11e6-8cc4-d21f52ade14e.png)

There are several possibilities to fix that issue:
- Using async unit tests. 
[...]

PR #2731, opened 2015-05-05
feature: mdToast uses `$interval` for hiding after delay

adamweeks:
In order to be better compatiable with Protractor, the hide delays on mdToast are
switched from `$timeout` to `$interval`.

https://github.com/angular/material/issues/2704

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2731)

### DIFF
--- a/src/components/toast/toast.js
+++ b/src/components/toast/toast.js
@@ -209,7 +209,7 @@ function MdToastProvider($$interimElementProvider) {
     return $mdToast;
 
   /* @ngInject */
-  function toastDefaultOptions($timeout, $animate, $mdToast, $mdUtil) {
+  function toastDefaultOptions($interval, $animate, $mdToast, $mdUtil) {
     return {
       onShow: onShow,
       onRemove: onRemove,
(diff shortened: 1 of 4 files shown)


ThomasBurleson:
Angular Material uses $timeout for frame delays not polling.  Per Protractor docs on [Timeouts](https://github.com/angular/protractor/blob/master/docs/timeouts.md):

> You should use the $interval service (interval.js) for anything that polls continuously (introduced in Angular 1.2rc3).

BelfordZ:
@ThomasBurleson what is the harm in merging this? 

the angular service $interval was [introduced to angular](https://github.com/angular/angular.js/commit/2b5ce84fca7b41fca24707e163ec6af84bc12e83) literally for [this exact reason](https://github.com/angular/angular.js/issues/2402). 

Timeout will increment outstandingRequestCount. Think about this for a second. You open the toast, and have determined that it will auto close in say 3 seconds. cool. Once the toast appears, is the page in a 'ready' state? That is, are there any outstanding units of work to complete? functionally, no there isn't, but with $timeout you are telling angular that there are. 

Protractor waits until the timeout has finished before continuing execution. If you are trying to test the text value inside the toast, protractor will wait for the timeout, removing the toast before you can actually touch it. This makes mdToast impossible cover via protractor.

This is a major issue and should be looked at as such.

PR #2761, opened 2015-05-06
fix(gesture): fix conflicts with Ionic

### LINKED ISSUE #1528: shouldHijackClicks does not work (when associated with ionic?)
the function:

```
if (shouldHijackClicks) {
  document.addEventListener('click', function(ev) {
    // Space/enter on a button, and submit events, can send clicks
    var isKeyClick = ev.clientX === 0 && ev.clientY === 0;
    if (isKeyClick || ev.$material) return;

    // Prevent clicks unless they're sent by material
    ev.preventDefault();
    ev.stopPropagation();
  }, true);
}
```

https://github.com/angular/material/blob/6ffc53374f2a28faee24c1ec24f5cc21211e1e75/src/core/services/gesture/gesture.js#L23

does not work when using Android or IOS. It easily reproductible using Chrome dev tools. Try Nexus 5 or Iphone 6 and you will see that none of the button, link etc work..

When I comment the above code it works as expected.

Using Angular 1.3.13, Ionic beta14 and Material 8.0.1-rc1

Hope it helps :)

kennethcachia:
Closes #1528 

This is an auto-fix for Ionic conflicts. Alternatively, you can currently fix Ionic conflicts via:
`$mdGestureProvider.skipClickHijack();`

### DIFF
--- a/src/core/services/gesture/gesture.js
+++ b/src/core/services/gesture/gesture.js
@@ -444,7 +444,7 @@
        */
       document.addEventListener('click', function clickHijacker(ev) {
         var isKeyClick = ev.clientX === 0 && ev.clientY === 0;
-        if (!isKeyClick && !ev.$material) {
+        if (!isKeyClick && !ev.$material && !ev.isIonicTap) {
           ev.preventDefault();
           ev.stopPropagation();
         }
--- a/src/core/services/gesture/gesture.js
+++ b/src/core/services/gesture/gesture.js
@@ -439,8 +439,9 @@
        * The only way to know if this click is real is to prevent any normal
        * click events, and add a flag to events sent by material so we know not to prevent those.
        * 
-       * One exception to click events that should be prevented is click events sent by the
-       * keyboard (eg form submit). 
+       * Two exceptions to click events that should be prevented are:
+       *  - click events sent by the keyboard (eg form submit)
+       *  - events that originate from an Ionic app
        */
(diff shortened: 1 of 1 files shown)


src/core/services/gesture/gesture.js
```diff
@@ -439,12 +439,13 @@
...
-       * One exception to click events that should be prevented is click events sent by the
-       * keyboard (eg form submit). 
+       * Two exceptions to click events that should be prevented are:
+       *  - click events sent by the keyboard (eg form submit)
+       *  - events that originate from an Ionic app
        */
       document.addEventListener('click', function clickHijacker(ev) {
         var isKeyClick = ev.clientX === 0 && ev.clientY === 0;
-        if (!isKeyClick && !ev.$material) {
+        if (!isKeyClick && !ev.$material && !ev.isIonicTap) {
```
marcysutton: Is there a way to make this less specific to Ionic, so we don't have it so closely coupled with Angular Material internals?

kennethcachia: Unfortunately I don't know how yet. I agree though, this is not ideal but we can always push this and improve it if we find a better alternative.

ThomasBurleson: @marcysutton - great question. 
Kenneth and I considered ideas such using GestureProvider to register 'shouldIgnore` callbacks. But for this 1 special case, it seemed more trouble with little ROI.

PR #2786, opened 2015-05-08
fix(checkbox): only add margin to the first span

jacksonrayhamilton:
If there are multiple spans inside a checkbox, all of them will have a left margin. That can be undesirable.

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2786)

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -121,7 +121,7 @@ md-checkbox {
     content: '';
   }
 
-  .md-label {
+  > .md-label {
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@jacksonrayhamilton - this PR is not associated with an issue. Please provide an screencapture, or CodePen demo that demonstrates its use. Is this part of the Material Design specs?

ThomasBurleson:
Fixed with https://github.com/angular/material/commit/99ba8706311bbca48e9c8c532763ddbdb52b8828

Issue #1911, opened 2015-03-15
md-swipe-left/right: Docs "view demo" link sends you to docs home page


epelc:
The view demo link sends you to the docs home page for `md-swipe-left` and `md-swipe-left`

https://material.angularjs.org/#/api/material.components.swipe/directive/mdSwipeRight

marcysutton:
Hiding the link for now.

matthewrfindley:
@epelc We decided to add a basic demo.  In the future this page will be the home of any additional demos for md-swipe-{left | right}

PR #2813, opened 2015-05-12
fix(demo,swipe): add missing demo for swipe

matthewrfindley:
Closes: https://github.com/angular/material/issues/1911

Fixes an issue where the link on the toolbar redirects to '/' since no
demo exists.  

One solution was to hide the link, but the process to determine if an ngDoc has a demo would be more complicated.  The code already handles showing the button based on if the doc hasDemo is true.  The sticky part is given a doc, it is difficult to determine the associated examples without knowledge of the example files convention

Additional demos can be added in the future to show more advanced usage.

Currently this is the only directive without a demo.


ThomasBurleson:
Closed with SHA 2cfc6929

PR #2815, opened 2015-05-12
docs(tooltip): Fix src attribute in "Refresh" button's <md-icon> in t…

jimmyorr:
…ooltip demo

Change src attribute from "icon" to "md-svg-src" and remove leading slash from url. As is, the refresh button does not appear in the demo.

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2815)


ThomasBurleson:
Already fixed.

PR #2831, opened 2015-05-12
Fix scrollbars when backdrop is nested and pushed out of the viewport.

richarddavenport:
Absolutely positioning the backdrop will always position it based off the parent's positioning. This is a problem if the backdrop were to be nested. Positioning fixed will always give it a position based off the viewport, which is the goal of any backdrop.

rad_2015-05-12_at_1_59_53_pm

rad_2015-05-12_at_2_11_18_pm

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2831)

### DIFF
--- a/src/components/backdrop/backdrop.scss
+++ b/src/components/backdrop/backdrop.scss
@@ -15,7 +15,7 @@ md-backdrop {
 
   background-color: rgba(0,0,0,0);
 
-  position: absolute;
+  position: fixed;
   height: 100%;
   left: 0;
   right: 0;


richarddavenport:
I signed it!

ThomasBurleson:
@robertmesserle - this lgtm. Your thoughts?

robertmesserle:
@ThomasBurleson Seems fine to me, but I'm wondering if there was a reason for using `absolute`.

ThomasBurleson:
If you change the `position:fixed` then all the dialog demos will fail.

robertmesserle:
@ThomasBurleson Ah, is this because of the scenarios where we don't blackout the whole page?  We may be able to use `position: fixed;` only when `body` is the parent element.  Thoughts?

ThomasBurleson:
@robertmesserle - exactly! I am testing that code now.

Issue #2822, opened 2015-05-12
Documentation code pen link for dialogs does not load anything


mrmeku:
See https://material.angularjs.org/#/demo/material.components.dialog and click on the code pen link to replicate

matthewrfindley:
This also applies to slider, contact chips demo, progress circle.  All others appear to work correctly.

matthewrfindley:
@mrmeku This has been fixed.

PR #2833, opened 2015-05-12
fix(codepen): fix demos with special characters

matthewrfindley:
Closes -> https://github.com/angular/material/issues/2822

An undocumented change occured during the 05/12/2015 release of codepen that caused some of our demos to break.

A more robust solution could be put in place to solve this, but the edge cases make the effort rather large for what seems to be a small feature.  At this point I am unclear if the changes to the API are intended and will continue to be the standard moving forward.  I am in favor of the simpler solution until the Codepen API is stable.

### DIFF
--- a/docs/app/js/codepen.js
+++ b/docs/app/js/codepen.js
@@ -39,8 +39,8 @@
     // See http://blog.codepen.io/documentation/api/prefill
     function escapeJsonQuotes(json) {
       return JSON.stringify(json)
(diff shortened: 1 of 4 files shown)


ThomasBurleson:
@matthewrfindley - can you rebase and update this PR?

matthewrfindley:
@ThomasBurleson PR is updated.

Issue #2788, opened 2015-05-08
Incorrect syntax on $mdDialog example


nikkwong:
Wanted to note that the example for **JavaScript: object syntax** in the [$mdDialog documentaiton](https://material.angularjs.org/#/api/material.components.dialog/service/$mdDialog) is missing a `}` on the last line, it should read, `}})(angular);` instead of the current `})(angular);`. 

As a side note, it would be great to get some further information on how `$mdDialog` returns promise objects, say if I modified models in a Dialog window, is that what `$mdDialog` returns via `$mdDialog.hide`?—I feel as the documentation is vague, even after reading the `$q` angular documentation. 

Thanks.

PR #2836, opened 2015-05-13
feat(docs): Use $scope instead of scope

miensol:
To avoid confusion it would be good to keep the examples consistent.
While `scope` in `DialogController` will be the same as `$scope` it brings confusion to a reader.
I can only guess that `scope` here working is by accident - caused by https://github.com/angular/material/blob/786b0ed3652b7460c2c802efa1aa79972bd96f5d/src/core/services/interimElement/interimElement.js#L353.

The fix also includes missing `}` mentioned here:   #2788

Here's a working pen: http://codepen.io/anon/pen/ZGWOpZ.

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -128,14 +128,14 @@ function MdDialogDirective($$rAF, $mdTheming) {
  *          },
  *          controller: DialogController
  *       });
- *       function DialogController(scope, $mdDialog, items) {
- *         scope.items = items;
- *         scope.closeDialog = function() {
+ *       function DialogController($scope, $mdDialog, items) {
+ *         $scope.items = items;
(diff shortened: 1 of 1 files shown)


miensol:
I signed it!

gkalpak:
The example does not work anyway. There are several issues that need to be fixed.

BTW, @ThomasBurleson, the examples in the _Services_ section seem to be derived from comments. As a result they are not editable in CodePen and (most importantly) they are not run in tests (which may result in incorrect/outdated code => confusion).
Something should probably be done about it :)

ThomasBurleson:
@gkalpak - https://github.com/angular/material/issues/2869

gkalpak:
:+1: @ThomasBurleson

PR #2842, opened 2015-05-13
Add FAB Speed Dial and FAB Toolbar components.

### LINKED ISSUE #1416: mdSpeedDial: Implement floating-action-button menu
The "speed dial" component inspired by Google Inbox allows for a small number of items to animate out from a floating action button (FAB). In the [Material Design spec](http://www.google.com/design/spec/components/buttons.html#buttons-floating-action-button) although there isn't any detail for this specific component, there are a few patterns and requirements noted for items animating out of FABs:

 * The FAB should be attached/anchored to something and not in a random location
 * A FAB menu has to contain related items. Consider an overflow menu for unrelated items.

There are a few inspirations: 
Implementation: http://nobitagit.github.io/ng-material-floating-button/ (needs accessibility)
Design: https://dribbble.com/shots/1788016-Inbox-Speed-Dial-Animation
Google Inbox SpeedDial UX: [MP4](https://drive.google.com/file/d/0B9DJpaYQys_9R01nZzY1ZzB0TTA/view?usp=sharing)

speeddial

Similar to a Bottom Sheet, the speed dial component is a mobile-focused component, and should be noted in the docs. 

h/t to https://github.com/angular/material/issues/1368.

topherfangio:
Fixes #1416 - Adds the Floating Action Button Speed Dial and Toolbar components.

CLA has been signed under Profoundry LLC.

Please let me know if you need any changes! :-)

### DIFF
--- a/src/components/fabSpeedDial/demoBasicUsage/index.html
+++ b/src/components/fabSpeedDial/demoBasicUsage/index.html
@@ -1,0 +1,70 @@
+<div ng-controller="AppCtrl" layout="column">
+  <md-content class="md-padding" layout="column">
+    <p>
+      You may supply a direction of <code>left</code>, <code>up</code>, <code>down</code>, or
+      <code>right</code> through the <code>md-direction</code> attribute.
+    </p>
+  </md-content>
+
+  <div layout="row" layout-align="center center">
+    <md-fab-speed-dial md-open="demo.isOpen" md-direction="{{demo.selectedDirection}}"
+                       ng-class="demo.selectedMode">
+      <md-fab-trigger>
+        <md-button aria-label="menu" class="md-fab md-warn">
+          <md-icon md-svg-src="img/icons/menu.svg"></md-icon>
+        </md-button>
+      </md-fab-trigger>
+
+      <md-fab-actions>
+        <md-button aria-label="twitter" class="md-fab md-raised md-mini">
+          <md-icon md-svg-src="img/icons/twitter.svg"></md-icon>
+        </md-button>
+        <md-button aria-label="facebook" class="md-fab md-raised md-mini">
+          <md-icon md-svg-src="img/icons/facebook.svg"></md-icon>
+        </md-button>
+        <md-button aria-label="Google hangout" class="md-fab md-raised md-mini">
+          <md-icon md-svg-src="img/icons/hangout.svg"></md-icon>
+        </md-button>
+      </md-fab-actions>
+    </md-fab-speed-dial>
+  </div>
+
+  <md-content class="md-padding" layout="column">
+    <div layout="row" layout-align="space-around">
+      <div layout="column">
+        <b>Direction</b>
+
+        <md-radio-group ng-model="demo.selectedDirection">
+          <md-radio-button ng-repeat="direction in demo.availableDirections"
+                           ng-value="direction" class="text-capitalize">
+            {{direction}}
+          </md-radio-button>
+        </md-radio-group>
+      </div>
+
+      <div layout="column">
+        <b>Open/Closed</b>
+
+        <md-radio-group ng-model="demo.isOpen">
+          <md-radio-button ng-value="true">Open</md-radio-button>
+          <md-radio-button ng-value="false">Closed</md-radio-button>
+        </md-radio-group>
+      </div>
+
+      <div layout="column">
+        <b>Animation Modes</b>
+
+        <md-radio-group ng-model="demo.selectedMode">
+          <md-radio-button ng-repeat="mode in demo.availableModes" ng-value="mode">
+            {{mode}}
+          </md-radio-button>
+        </md-radio-group>
+      </div>
+    </div>
+
+    <p class="note">
+      Note that you can also hover over the directive's area or tab through each button to open and
+      activate the speed dial menu.
+    </p>
+  </md-content>
+</div>
--- a/src/components/fabSpeedDial/fabSpeedDial.js
+++ b/src/components/fabSpeedDial/fabSpeedDial.js
@@ -1,0 +1,209 @@
+(function() {
+  'use strict';
+
+  angular
+    .module('material.components.fabSpeedDial', [
+      'material.core',
+      'material.components.fabTrigger',
+      'material.components.fabActions'
+    ])
+    .directive('mdFabSpeedDial', MdFabSpeedDialDirective)
+    .animation('.md-fling', MdFabSpeedDialFlingAnimation)
+    .animation('.md-scale', MdFabSpeedDialScaleAnimation);
+
+  /**
+   * @ngdoc directive
+   * @name mdFabSpeedDial
+   * @module material.components.fabSpeedDial
+   *
+   * @restrict E
+   *
+   * @description
+   * The `<md-fab-speed-dial>` directive is used to present a series of popup elements (usually
+   * `<md-button>`s) for quick access to common actions.
+   *
+   * There are currently two animations available by applying one of the following classes to
+   * the component:
+   *
+   *  - `md-fling` - The speed dial items appear from underneath the trigger and move into their
+   *    appropriate positions.
+   *  - `md-scale` - The speed dial items appear in their proper places by scaling from 0% to 100%.
+   *
+   * @usage
+   * <hljs lang="html">
+   * <md-fab-speed-dial direction="up" class="md-fling">
+   *   <md-fab-trigger>
+   *     <md-button><md-icon icon="/img/icons/plus.svg"></md-icon></md-button>
+   *   </md-fab-trigger>
+   *
+   *   <md-fab-actions>
+   *     <md-button><md-icon icon="/img/icons/user.svg"></md-icon></md-button>
+   *     <md-button><md-icon icon="/img/icons/group.svg"></md-icon></md-button>
+   *   </md-fab-actions>
+   * </md-fab-speed-dial>
+   * </hljs>
+   *
+   * @param {string=} md-direction From which direction you would like the speed dial to appear
+   * relative to the trigger element.
+   * @param {expression=} md-open Programmatically control whether or not the speed-dial is visible.
+   */
+  function MdFabSpeedDialDirective() {
+    return {
+      restrict: 'E',
+
+      scope: {
+        direction: '@?mdDirection',
+        isOpen: '=?mdOpen'
+      },
+
+      bindToController: true,
+      controller: FabSpeedDialController,
+      controllerAs: 'vm'
+    };
+
+    function FabSpeedDialController($scope, $element, $animate) {
+      var vm = this;
+
+      // Define our open/close functions
+      vm.open = function() {
+        vm.isOpen = true;
+        $scope.$apply();
+      };
+
+      vm.close = function() {
+        vm.isOpen = false;
+        $scope.$apply();
+      };
+
+      setupDefaults();
+      setupListeners();
+      setupWatchers();
+
+      // Set our default variables
+      function setupDefaults() {
+        // Set the default direction to 'down' if none is specified
+        if (!vm.direction) {
+          vm.direction = 'down';
+        }
+
+        // Set the default to be closed
+        vm.isOpen = vm.isOpen || false;
+      }
+
+      // Setup our event listeners
+      function setupListeners() {
+        $element.on('mouseenter', vm.open);
+        $element.on('mouseleave', vm.close);
+      }
+
+      // Setup our watchers
+      function setupWatchers() {
+        // Watch for changes to the direction and update classes/attributes
+        $scope.$watch('vm.direction', function(newDir, oldDir) {
+          // Add the appropriate classes so we can target the direction in the CSS
+          $animate.removeClass($element, 'md-' + oldDir);
+          $animate.addClass($element, 'md-' + newDir);
+        });
+
+
+        // Watch for changes to md-open
+        $scope.$watch('vm.isOpen', function(isOpen) {
+          var toAdd = isOpen ? 'md-is-open' : '';
+          var toRemove = isOpen ? '' : 'md-is-open';
+
+          $animate.setClass($element, toAdd, toRemove);
+        });
+      }
+    }
+  }
+
+  function MdFabSpeedDialFlingAnimation() {
+    function runAnimation(element) {
+      var el = element[0];
+      var ctrl = element.controller('mdFabSpeedDial');
+      var items = el.querySelectorAll('.md-fab-action-item');
+
+      // Always reset the items to their natural position/state
+      angular.forEach(items, function(item, index) {
+        item.style.transform = '';
+        item.style.transitionDelay = '';
+      });
+
+      // If the control is closed, hide the items behind the trigger
+      if (!ctrl.isOpen) {
+        angular.forEach(items, function(item, index) {
+          var newPosition, axis;
+
+          switch (ctrl.direction) {
+            case 'up':
+              newPosition = item.scrollHeight * (index + 1);
+              axis = 'Y';
+              break;
+            case 'down':
+              newPosition = -item.scrollHeight * (index + 1);
+              axis = 'Y';
+              break;
+            case 'left':
+              newPosition = item.scrollWidth * (index + 1);
+              axis = 'X';
+              break;
+            case 'right':
+              newPosition = -item.scrollWidth * (index + 1);
+              axis = 'X';
+              break;
+          }
+
+          item.style.transform = 'translate' + axis + '(' + newPosition + 'px)';
+        });
+      }
+    }
+
+    return {
+      addClass: function(element, className, done) {
+        if (element.hasClass('md-fling')) {
+          runAnimation(element);
+        }
+      },
+      removeClass: function(element, className, done) {
+        runAnimation(element);
+      }
+    }
+  }
+
+  function MdFabSpeedDialScaleAnimation() {
+    function runAnimation(element) {
+      var el = element[0];
+      var ctrl = element.controller('mdFabSpeedDial');
+      var items = el.querySelectorAll('.md-fab-action-item');
+
+      // Always reset the items to their natural position/state
+      angular.forEach(items, function(item, index) {
+        item.style.transform = '';
+        item.style.transitionDelay = '';
+      });
+
+      // If it's open,
+      if (ctrl.isOpen) {
+        angular.forEach(items, function(item, index) {
+          item.style.transform = 'scale(1)';
+          item.style.transitionDelay = (index * 100) + 'ms';
+        });
+      } else {
+        angular.forEach(items, function(item, index) {
+          item.style.transform = 'scale(0)';
+          item.style.transitionDelay = (items.length - index * 100) + 'ms';
+        });
+      }
+    }
+
+    return {
+      addClass: function(element, className, done) {
+        runAnimation(element);
+      },
+
+      removeClass: function(element, className, done) {
+        runAnimation(element);
+      }
+    }
+  }
+})();
--- a/src/components/fabSpeedDial/fabSpeedDial.scss
+++ b/src/components/fabSpeedDial/fabSpeedDial.scss
@@ -1,0 +1,93 @@
+md-fab-speed-dial {
+  position: relative;
+  display: flex;
+  align-items: center;
+
+  &.md-is-open {
+    .md-fab-action-item {
+      visibility: visible;
+    }
+  }
+
+  md-fab-actions {
+    display: flex;
+
+    // Set the height so that the z-index in the JS animation
+    height: 100%;
+
+    .md-fab-action-item {
+      visibility: hidden;
+      transition: $swift-ease-in;
+    }
+  }
+
+  &.md-down {
+    flex-direction: column;
+
+    md-fab-trigger {
+      order: 1;
+    }
+
+    md-fab-actions {
+      flex-direction: column;
+      order: 2;
+    }
+  }
+
+  &.md-up {
+    flex-direction: column;
+
+    md-fab-trigger {
+      order: 2;
+    }
+
+    md-fab-actions {
+      flex-direction: column-reverse;
+      order: 1;
+    }
+  }
+
+  &.md-left {
+    flex-direction: row;
+
+    md-fab-trigger {
+      order: 2;
+    }
+
+    md-fab-actions {
+      flex-direction: row-reverse;
+      order: 1;
+
+      .md-fab-action-item {
+        transition: $swift-ease-in;
+      }
+    }
+  }
+
+  &.md-right {
+    flex-direction: row;
+
+    md-fab-trigger {
+      order: 1;
+    }
+
+    md-fab-actions {
+      flex-direction: row;
+      order: 2;
+
+      .md-fab-action-item {
+        transition: $swift-ease-in;
+      }
+    }
+  }
+
+  /*
+   * Handle the animations
+   */
+  &.md-scale {
+    .md-fab-action-item {
+      transform: scale(0);
+      transition: $swift-ease-in;
+    }
+  }
+}
--- a/src/components/fabSpeedDial/fabSpeedDial.spec.js
+++ b/src/components/fabSpeedDial/fabSpeedDial.spec.js
@@ -1,0 +1,82 @@
+describe('<md-fab-speed-dial> directive', function() {
+
+  beforeEach(module('material.components.fabSpeedDial'));
+
+  var pageScope, element, controller;
+
+  function compileAndLink(template) {
+    inject(function($compile, $rootScope) {
+      pageScope = $rootScope.$new();
+      element = $compile(template)(pageScope);
+      controller = element.controller('mdFabSpeedDial');
+
+      pageScope.$apply();
+    });
+  }
+
+  it('applies a class for each direction', inject(function() {
+    compileAndLink([
+      '<md-fab-speed-dial md-direction="{{direction}}"></md-fab-speed-dial>'
+    ].join("\n"));
+
+    pageScope.$apply('direction = "down"');
+    expect(element.hasClass('md-down')).toBe(true);
+
+    pageScope.$apply('direction = "up"');
+    expect(element.hasClass('md-up')).toBe(true);
+
+    pageScope.$apply('direction = "left"');
+    expect(element.hasClass('md-left')).toBe(true);
+
+    pageScope.$apply('direction = "right"');
+    expect(element.hasClass('md-right')).toBe(true);
+  }));
+
+  it('opens when the trigger element is focused', inject(function() {
+    compileAndLink([
+      '<md-fab-speed-dial><md-fab-trigger><button></button></md-fab-trigger></md-fab-speed-dial>'
+    ].join('\n'));
+
+    element.find('button').triggerHandler('focus');
+    expect(controller.isOpen).toBe(true);
+  }));
+
+  it('opens when the speed dial elements are focused', inject(function() {
+    compileAndLink([
+      '<md-fab-speed-dial><md-fab-actions><button></button></md-fab-actions></md-fab-speed-dial>'
+    ].join('\n'));
+
+    element.find('button').triggerHandler('focus');
+    expect(controller.isOpen).toBe(true);
+  }));
+
+  it('closes when the speed dial elements are blurred', inject(function() {
+    compileAndLink([
+      '<md-fab-speed-dial><md-fab-actions><button></button></md-fab-actions></md-fab-speed-dial>'
+    ].join('\n'));
+
+    element.find('button').triggerHandler('focus');
+    expect(controller.isOpen).toBe(true);
+
+    element.find('button').triggerHandler('blur');
+    expect(controller.isOpen).toBe(false);
+  }));
+
+  it('allows programmatic opening through the md-open attribute', inject(function() {
+    compileAndLink([
+      '<md-fab-speed-dial md-open="isOpen"></md-fab-speed-dial>'
+    ].join('\n'));
+
+    // By default, it should be closed
+    expect(controller.isOpen).toBe(false);
+
+    // When md-open is true, it should be open
+    pageScope.$apply('isOpen = true');
+    expect(controller.isOpen).toBe(true);
+
+    // When md-open is false, it should be closed
+    pageScope.$apply('isOpen = false');
+    expect(controller.isOpen).toBe(false);
+  }));
+
+});
--- a/src/components/fabToolbar/fabToolbar.js
+++ b/src/components/fabToolbar/fabToolbar.js
@@ -1,0 +1,152 @@
+(function() {
+  'use strict';
+
+  angular
+    .module('material.components.fabToolbar', [
+      'material.core',
+      'material.components.fabTrigger',
+      'material.components.fabActions'
+    ])
+    .directive('mdFabToolbar', MdFabToolbarDirective)
+    .animation('.md-is-open', MdFabToolbarOpenAnimation);
+
+  /**
+   * @ngdoc directive
+   * @name mdFabToolbar
+   * @module material.components.fabToolbar
+   *
+   * @restrict E
+   *
+   * @description
+   *
+   * The `<md-fab-toolbar>` directive is used present a toolbar of elements (usually `<md-button>`s)
+   * for quick access to common actions when a floating action button is activated (via hover or
+   * keyboard navigation).
+   *
+   * @usage
+   *
+   * <hljs lang="html">
+   * <md-fab-toolbar>
+   *   <md-fab-trigger>
+   *     <md-button><md-icon icon="/img/icons/plus.svg"></md-icon></md-button>
+   *   </md-fab-trigger>
+   *
+   *   <md-fab-actions>
+   *     <md-button><md-icon icon="/img/icons/user.svg"></md-icon></md-button>
+   *     <md-button><md-icon icon="/img/icons/group.svg"></md-icon></md-button>
+   *   </md-fab-actions>
+   * </md-fab-toolbar>
+   * </hljs>
+   *
+   * @param {expression=} md-open Programmatically control whether or not the toolbar is visible.
+   */
+  function MdFabToolbarDirective() {
+    return {
+      restrict: 'E',
+      transclude: true,
+      template: [
+        '<div class="md-fab-toolbar-wrapper">',
+        '  <div class="md-fab-toolbar-background"></div>',
+        '  <div class="md-fab-toolbar-content" ng-transclude></div>',
+        '</div>'
+      ].join('\n'),
+
+      scope: {
+        isOpen: '=?mdOpen'
+      },
+
+      bindToController: true,
+      controller: FabToolbarController,
+      controllerAs: 'vm',
+
+      link: link
+    };
+
+    function FabToolbarController($scope, $element, $animate) {
+      var vm = this;
+
+      // Set the default to be closed
+      vm.isOpen = vm.isOpen || false;
+
+      vm.open = function() {
+        vm.isOpen = true;
+        $scope.$apply();
+      };
+
+      vm.close = function() {
+        vm.isOpen = false;
+        $scope.$apply();
+      };
+
+      // Setup some mouse events so the hover effect can be triggered
+      // anywhere over the toolbar
+      $element.on('mouseenter', vm.open);
+      $element.on('mouseleave', vm.close);
+
+      // Watch for changes to md-open and toggle our class
+      $scope.$watch('vm.isOpen', function(isOpen) {
+        var toAdd = isOpen ? 'md-is-open' : '';
+        var toRemove = isOpen ? '' : 'md-is-open';
+
+        $animate.setClass($element, toAdd, toRemove);
+      });
+    }
+
+    function link(scope, element, attributes) {
+      // Don't allow focus on the trigger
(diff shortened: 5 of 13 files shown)


topherfangio:
I signed it!

Signed under company Profoundry LLC with email topher@profoundry.us.

ThomasBurleson:
@topherfangio - this looks very impressive.

topherfangio:
@ThomasBurleson Thanks for the kind words!

Have you found any issues that need to be addressed?

If there aren't any changes, and I have time before the 0.10.0 release, I may try to add a more generic method that supports both the toolbar and any other "generic material" as shown in the video demonstrating a pop-up menu.

marcysutton:
This looks great, just a few comments! 

On the FAB Speed Dial, can we limit it to one direction at a time and slow down the animation? The current animation sequence with all 4 speed dials at once is a bit overwhelming visually, and it also complicates labeling from a screen reader perspective (plus they probably shouldn't be used that way). 

For labeling the speed dial, the triggering FAB needs some kind of toggle language to indicate what it does and whether it's open or closed. `aria-haspopup`, `aria-expanded` and `aria-controls` are a few attributes to consider. If there were more than one on a page, just for context, each instance of the menu would need unique button labels in it to differentiate one from the other.

For the FAB toolbar, obviously we need to fix the icon button colors. I'd look at a similar pattern as I described for the speed dial: associating the triggering button with its associated content with ARIA, and semantically indicating whether it's open or closed. I am totally available to help you work through this, I just wanted to capture some ideas here.

For keyboard focus of both demos, should the FAB behave more like a toggle where the child items are removed from the tab order until explicitly toggled? It's a bad idea to force keyboard users through items that are otherwise accessed through hover. I don't think this level of accessibility detail was reached for the Inbox version (but it needs it).

topherfangio:
Hi @marcysutton, thanks for the feedback!

I can definitely update the speed dial demo to only allow one direction at a time, I think this definitely makes a lot more sense. I should be able to update this tonight or tomorrow.

**Re: Labeling** - I can certainly add the `aria` controls, but I could definitely use a bit of direction in this area as I made the assumption it would be up to the developer who uses the component to add the proper aria attributes. Do I just need to update the demo to show how this could be done, or were you suggesting that the speed dial/toolbar actually update the child elements automatically?

**Re: Icon Colors** - I just realized that this didn't follow the same behavior as the standard toolbar. Do I just need to add the `md-focused` class to the currently focused button to fix this?

Lastly regarding keyboard focus, I'm not quite sure I understand the scenario. If we assume a complete lack of a mouse, then the only way to toggle/activate the child elements would be to tab to the FAB at which point it would activate and the children would then be added to the tab index, so you would still have to tab through the children (i.e. there would still be no way to skip them). I admittedly have not done much with keyboard support, so perhaps there is a scenario I am missing regarding screen readers?

Thanks again! Looking forward to making this as awesome as possible :-)

topherfangio:
@marcysutton Quick update: I just realized that in Gmail you can't even tab to the compose FAB. Is that the intended behavior? Since it is a button and you can click it, I just assumed you could tab to it, but if that is not the case, we can definitely remove it and it's children from the tabindex.

topherfangio:
@marcysutton Sorry for the slew of updates; just noticed what you were talking about regarding the toolbar icon color. I think that broke when I merged in the latest master changes. I'll obviously fix that :-)

ThomasBurleson:
*  The toolbar slideout is too fast and [perhaps] should be below the button as the trigger button fades; e.g. an improved transition would be good.
*  To apply a change and trigger a digest/binding updates, use:
```js
vm.focus = function() { $scope.$apply('isOpen = true'); };
```
*  Rename to `md-is-open` since it is a value/state and not an expression
*  The lookup of `backgroundElement` and `triggerElement` are too dependent upon DOM order and will easily fail. These lookups should not be in `$watch()`.

topherfangio:
@ThomasBurleson First three points sound good; I'll make the changes.

Regarding DOM lookup, I totally agree but can you suggest a better way? Without jqLite supporting CSS selectors, the only thing I can think of is to create full elements so I can do `$element.find('md-fab-toolbar-background')`. That seemed like a bit of overkill to create a whole element just so I could find it more easily.

ThomasBurleson:
How about something like $element.querySelector('div.md-fab-toolbar-background')` as an example.
And if you do this 1x in the postLink, the perf impact is minimal.

src/components/fabToolbar/fabToolbar.js
```diff
@@ -0,0 +1,134 @@
...
+        '  <div class="md-fab-toolbar-background"></div>',
+        '  <div ng-transclude></div>',
+        '</div>'
+      ].join('\n'),
+
+      scope: {
+        isOpen: '=?mdOpen',
+      },
+
+      controller: controller,
```
ThomasBurleson: Rename to `FabToolbarController`

src/components/fabToolbar/fabToolbar.js
```diff
@@ -0,0 +1,134 @@
...
+
+      link: link
+    };
+
+    function controller($scope, $element) {
+      var vm = this;
```
ThomasBurleson: No need for this approach since you are not using **controllerAs** syntax now using such notations in your template.

src/components/fabToolbar/fabToolbar.js
```diff
@@ -0,0 +1,134 @@
...
+
+      // Set the default to be closed
+      if (typeof $scope.isOpen === 'undefined') {
+        $scope.isOpen = false;
+      }
+
+      vm.focus = function() {
+        // TODO: Why does this need to be inside a $timeout???
+        $timeout(function() {
```
ThomasBurleson: Use `$scope.$apply('isOpen = true')` to update the scope variable and trigger databinding updates.

src/components/fabToolbar/fabToolbar.js
```diff
@@ -0,0 +1,134 @@
...
+          // Get our variables
+          var width = $element[0].offsetWidth;
+          var height = $element[0].offsetHeight;
+
+          // Make a square
+          var scaleWidth = width * 2;
+          var scaleHeight = width * 2;
+
+          // Set the positioning and border
+          var borderRadius = width;
```
ThomasBurleson: Not needed.

src/components/fabSpeedDial/fabSpeedDial.scss
```diff
@@ -0,0 +1,167 @@
...
+
+    &.md-right {
+      .md-fab-action-item {
+        &:nth-child(1) { transform: translateX(-$trigger-height * 1 - $trigger-offset - 5); z-index: 6; }
+        &:nth-child(2) { transform: translateX(-$trigger-height * 2 - $trigger-offset - 5); z-index: 5; }
+        &:nth-child(3) { transform: translateX(-$trigger-height * 3 - $trigger-offset - 5); z-index: 4; }
+        &:nth-child(4) { transform: translateX(-$trigger-height * 4 - $trigger-offset - 5); z-index: 3; }
+        &:nth-child(5) { transform: translateX(-$trigger-height * 5 - $trigger-offset - 5); z-index: 2; }
+        &:nth-child(6) { transform: translateX(-$trigger-height * 6 - $trigger-offset - 5); z-index: 1; }
+      }
```
ThomasBurleson: So this styling effectively limits the animations to max of 6 action items ?

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,138 @@
...
+      scope: {
+        direction: '@?mdDirection',
+        isOpen: '=?mdOpen'
+      },
+
+      controller: controller,
+
+      compile: function($element) {
+        // Wrap every child in a new div and add a class that we can scale independently
+        var children = $element.find('md-fab-actions').children(),
```
ThomasBurleson: Perhaps better to create a new `md-fab-actions` directive which registers its children elements with the **FabSpeedDialController**. This gives us more control to manipulate and sanity check action items.

ThomasBurleson:
@topherfangio - Really excellent first draft. After you make these changes, I will ask @jelbourn to provide a final review. 

> With these changes I think we will have something solid.

src/components/fabToolbar/fabToolbar.js
topherfangio: Can you explain why the border radius is not needed? In the animation video on the design spec, it clearly shows (okay, semi-clearly shows; I had to play/pause the animation about 20 times to see it) a rounded edge as the toolbar appears.

src/components/fabToolbar/fabToolbar.js
topherfangio: Yeah, I was originally using `controllerAs` but switched when I was having trouble with the scope and wound up using `$timeout`. I'll remove this.

src/components/fabSpeedDial/fabSpeedDial.scss
topherfangio: Indeed; I tried to refactor this using an SCSS for-loop so I could easily support 100 or more, but it wouldn't recognize a variable inside of the `:nth-child()` pseudo-selector. If there is a better way, I am all for fixing this as it seems stupid to limit it to 6 even if that is what the spec suggests.

src/components/fabSpeedDial/fabSpeedDial.js
topherfangio: Sounds good; I'll get that implemented.

src/components/fabSpeedDial/fabSpeedDial.scss
ThomasBurleson: Maybe we just predefine for  <= 10 items and then note it in the docs.
@robertmesserle - any thoughts here?

src/components/fabToolbar/fabToolbar.js
ThomasBurleson: Since you only use it in 
```js 
backgroundElement.style.borderRadius = borderRadius + 'px';
```
I was suggesting just:
```js 
backgroundElement.style.borderRadius = width + 'px';
```

topherfangio: Ah gotcha, will do.

src/components/fabSpeedDial/demoBasicUsage/index.html
```diff
@@ -0,0 +1,69 @@
+<div ng-controller="AppCtrl" layout="column">
+    <md-content class="md-padding" layout="column">
+        <p>
+            You may supply a direction of <code>left</code>, <code>up</code>, <code>down</code>, or <code>right</code>
```
jelbourn: Can you keep your lines to 100 characters (here and everywhere else)?

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,138 @@
...
+    };
+
+    function controller($scope, $element) {
+      var vm = this;
+
+      vm.scope = $scope;
+      vm.element = $element;
+
+      // Set the default direction to 'down' if none is specified
+      if (typeof $scope.direction === 'undefined') {
```
jelbourn: Why not just use a falsy check?

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,138 @@
...
+   *     <md-button><md-icon icon="/img/icons/group.svg"></md-icon></md-button>
+   *   </md-fab-actions>
+   * </md-fab-speed-dial>
+   * </hljs>
+   *
+   * @param {string=} md-direction From which direction you would like the speed dial to appear relative to the
+   * trigger element.
+   * @param {expression=} md-open Programmatically control whether or not the speed-dial is visible.
+   */
+  function MdFabSpeedDialDirective($timeout) {
```
jelbourn: Can you refactor the code such that the "directive" is just a simple configuration object, where the `compile`, `link`, and `controller` properties are set as references to names functions?

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,138 @@
...
+
+    return {
+      restrict: 'E',
+
+      scope: {
+        direction: '@?mdDirection',
+        isOpen: '=?mdOpen'
+      },
```
jelbourn: Use `bindToController: true` to bind to the controller instance instead of the `$scope` instance.

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,138 @@
...
+    function controller($scope, $element) {
+      var vm = this;
+
+      vm.scope = $scope;
+      vm.element = $element;
+
+      // Set the default direction to 'down' if none is specified
+      if (typeof $scope.direction === 'undefined') {
+        $timeout(function() {
+          $scope.direction = 'down';
```
jelbourn: Can you make an enum for the direction values?

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,138 @@
...
+
+        // Return our link function
+        return {
+          post: link
+        }
+      }
+    };
+
+    function controller($scope, $element) {
+      var vm = this;
```
jelbourn: Can you use `self` instead of `vm` ?

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,138 @@
...
+        $timeout(function() {
+          vm.scope.isOpen = true;
+        });
+      };
+
+      vm.blur = function() {
+        $timeout(function() {
+          vm.scope.isOpen = false;
+        });
+      };
```
jelbourn: `focus` and `blur` here would be much more accurately named `open` and `close`

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,138 @@
...
+
+    }
+
+    function link($scope, $element, $attrs, ctrl) {
```
jelbourn: These arguments shouldn't start with a `$` because they are not injected. The `link` function is simply passed them.

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,138 @@
...
+      var i, trigger, child, children, focusable;
```
jelbourn: Move as much of this code from the `link` function to the controller as possible.

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,138 @@
...
+        }
+
+        // Return our link function
+        return {
+          post: link
+        }
+      }
+    };
+
+    function controller($scope, $element) {
```
jelbourn: The name of this function should be `MdFabSpeedDialCtrl` or `MdFabSpeedDialController`

src/components/fabSpeedDial/fabSpeedDial.js
jelbourn: Just declare the variables where they are used. The hoisting behavior only rarely ever affects anything.

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,138 @@
...
+      var i, trigger, child, children, focusable;
+
+      // Add the focus/blur events to the trigger element
+      trigger = $element.find('md-fab-trigger').children().eq(0);
+
+      trigger.on('focus', ctrl.focus);
+      trigger.on('blur', ctrl.blur);
+
+      // Add some focus/blur events to whatever (post-compiled) elements the user added in the speed dial
+      children = $element.find('md-fab-actions').children();
```
jelbourn: This seems like it could be its own directive.

marcysutton:
@topherfangio for keyboard accessibility of the speed dial, let's have it show on focus/hide on blur of the main FAB but only allow arrow keys to navigate between the child buttons...more like a menu. That way you don't force keyboard users through the whole list of child buttons. A good example is this dropdown menu: http://terrillthompson.com/tests/menus/accessible-mega-menu/test.html

src/components/fabSpeedDial/fabSpeedDial.scss
```diff
@@ -0,0 +1,167 @@
...
+    &.fling {
+      md-fab-actions {
+        .md-fab-action-item {
+          visibility: visible;
+          transform: translateY(0px) !important;
+        }
+      }
+    }
+
+    &.scale {
```
jelbourn: What is `.scale` ?

src/components/fabSpeedDial/fabSpeedDial.scss
```diff
@@ -0,0 +1,167 @@
...
+
+  $trigger-width: 56px;
+  $trigger-height: 52px;
+  $trigger-offset: 8px;
+
+  display: flex;
+  align-items: center;
+
+  &.md-opened {
+    &.fling {
```
jelbourn: All CSS classes must have the `md-` prefix.

src/components/fabSpeedDial/fabSpeedDial.scss
```diff
@@ -0,0 +1,167 @@
...
+      md-fab-actions {
+        .md-fab-action-item {
+          visibility: visible;
+          transform: translateY(0px) !important;
+        }
+      }
+    }
```
jelbourn: Can you try to restructure these rules to use less nesting? The nesting makes the rules raise the specificity and is a bit slower.

src/components/fabSpeedDial/fabSpeedDial.scss
```diff
@@ -0,0 +1,167 @@
...
+
+  display: flex;
+  align-items: center;
+
+  &.md-opened {
+    &.fling {
+      md-fab-actions {
+        .md-fab-action-item {
+          visibility: visible;
+          transform: translateY(0px) !important;
```
jelbourn: Any use of `!important` should be accompanies by a comment explaining what's going on and why it's necessary.

src/components/fabSpeedDial/fabSpeedDial.scss
```diff
@@ -0,0 +1,167 @@
...
+    .md-fab-action-item {
+      transform: scale(0);
+      transition: $custom-transition;
+    }
+  }
+
+  &.fling {
+    &.md-down {
+      .md-fab-action-item {
+        &:nth-child(1) { transform: translateY(-$trigger-height * 1 - $trigger-offset); z-index: 6; }
```
jelbourn: These could be done with a scss loop.

jelbourn: All `z-index` properties should be set from (or based on) something from the z-index master list in `variables.scss`

src/components/fabSpeedDial/fabSpeedDial.scss
jelbourn: My understanding is that these menus should be *extremely* limited, and 6 actually sounds about right. 

This is actually quite a lot of CSS, though, so I wonder if it would make sense to set the styles w/ javascript.

src/components/fabSpeedDial/fabSpeedDial.spec.js
```diff
@@ -0,0 +1,95 @@
...
+
+  beforeEach(module('material.components.fabSpeedDial'));
+
+  function setup(template) {
+    var el;
+
+    inject(function($compile, $rootScope) {
+      el = $compile(template)($rootScope.$new());
+      $rootScope.$apply();
+    });
```
jelbourn: To make testing easier, you can do something like:

```js
  pageScope = $rootScope.$new();
  element = $compile(template)(pageScope);
  scope = element.scope();
  controller = element.controller('mdFabSpeedDial');
```

This will let you make assertions against the directive's controller instance without having to make all direct DOM checks.

src/components/fabSpeedDial/fabSpeedDial.spec.js
```diff
@@ -0,0 +1,95 @@
+describe('<md-fab-speed-dial> directive', function() {
+
+  beforeEach(module('material.components.fabSpeedDial'));
+
+  function setup(template) {
```
jelbourn: `setup` > `compileAndLink`

src/components/fabSpeedDial/fabSpeedDial.spec.js
```diff
@@ -0,0 +1,95 @@
...
+      $rootScope.$apply();
+    });
+
+    return el;
+  }
+
+  it('applies a class for each direction', inject(function() {
+    var fab = setup([
+      '<md-fab-speed-dial md-direction="{{direction}}"></md-fab-speed-dial>'
+    ].join("\n"));
```
jelbourn: What is this `join` thing you're doing? Looks like a no-op.

src/components/fabSpeedDial/fabSpeedDial.spec.js
```diff
@@ -0,0 +1,95 @@
...
+    expect(fab.hasClass('md-opened')).toBe(true);
+  }));
+
+  it('closes when the speed dial elements are blurred', inject(function($timeout) {
+    var fab = setup([
+      '<md-fab-speed-dial><md-fab-actions><button></button></md-fab-actions></md-fab-speed-dial>'
+    ].join('\n'));
+
+    fab.find('button').triggerHandler('focus');
+    $timeout(function() {
```
jelbourn: The inside of a `$timeout` doesn't run in unit tests- `$timeout` is replaced with a mock service. 

Instead, call `$timeout.flush()` and then make your assertion.

src/components/fabSpeedDial/demoBasicUsage/index.html
```diff
@@ -0,0 +1,69 @@
+<div ng-controller="AppCtrl" layout="column">
+    <md-content class="md-padding" layout="column">
```
jelbourn: Use 2 spaces as indent (here and in other html files)

src/components/fabToolbar/fabToolbar.js
```diff
@@ -0,0 +1,134 @@
...
+  function MdFabToolbarDirective($timeout) {
+    return {
+      restrict: 'E',
+      transclude: true,
+      template: [
+        '<div class="md-fab-toolbar-wrapper">',
+        '  <div class="md-fab-toolbar-background"></div>',
+        '  <div ng-transclude></div>',
+        '</div>'
+      ].join('\n'),
```
jelbourn: Instead of using `.join()` for multi-line strings, can you follow the style given by the Google js style guide?
http://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml#Multiline_string_literals

src/components/fabToolbar/fabToolbar.js
```diff
@@ -0,0 +1,134 @@
...
+
+      scope: {
+        isOpen: '=?mdOpen',
+      },
+
+      controller: controller,
+
+      link: link
```
jelbourn: Same thing about `bindToController` and the distinction between `link` and the controller here.

src/components/fabSpeedDial/fabSpeedDial.scss
ThomasBurleson: :+1:  I missed this one.

src/components/fabToolbar/fabToolbar.js
ThomasBurleson: @jelbourn - can you elaborate. Not sure what you mean here.

src/components/fabToolbar/fabToolbar.js
```diff
@@ -0,0 +1,134 @@
...
+      $element.on('mouseleave', ctrl.blur);
+
+      // Don't allow focus on the trigger
+      $element.find('md-fab-trigger').find('button').attr('tabindex', '-1');
+
+      // Watch for changes to md-open
+      $scope.$watch('isOpen', function(isOpen) {
+        // Grab the elements
+        var backgroundElement = $element.children().eq(0).children().eq(0)[0];
+        var triggerElement = $element.children().eq(0).children().eq(1).children().eq(0)[0];
```
jelbourn: These lines are super cryptic and very difficult to understand.

src/components/fabToolbar/fabToolbar.js
```diff
@@ -0,0 +1,134 @@
...
+
+        // If we find both elements, use them to position the new backgroundt co
+        if (triggerElement && backgroundElement) {
+          // Get our variables
+          var width = $element[0].offsetWidth;
```
jelbourn: This section needs a high-level explanation of what's going on and should also probably be broken out into its own function.

src/components/fabToolbar/fabToolbar.scss
```diff
@@ -0,0 +1,95 @@
+md-fab-toolbar {
+  display: block;
+
+  $custom-duration: $swift-ease-in-duration / 1.5;
```
jelbourn: Add comment explaining the reasoning behind the custom duration.

ThomasBurleson: And per my earlier notes, the animation is too fast.

src/components/fabSpeedDial/fabSpeedDial.js
robertmesserle: @jelbourn I would disagree with this - it might be personal preference, but I much prefer having a set list up top so that I know the variables being used.

src/components/fabSpeedDial/fabSpeedDial.scss
robertmesserle: I would use an scss loop for this, using the syntax @jelbourn used above for the `nth-child` syntax.

ThomasBurleson: Agreed... but just prebuild the CSS output for 10 items. Right ?

jelbourn: I would lean towards having no CSS for this and instead applying the styles via JS.

jelbourn:
@topherfangio let me know when you've gone through the comments and are ready for another pass.

src/components/fabSpeedDial/fabSpeedDial.spec.js
gkalpak: Won't `pageScope` be the same as `scope` in the above snippet ?
Did you by any chance mean `scope = element.isolateScope()` ?
<sub>(I haven't looked at the actual code; I'm just making assumptions here (and they might be totally off).</sub>

jelbourn: Yes, I meant `isolatedScope`

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,130 @@
...
+
+  /**
+   * @ngdoc directive
+   * @name mdFabSpeedDial
+   * @module material.components.fabSpeedDial
+   *
+   * @restrict E
+   *
+   * @description
+   * The `<md-fab-speed-dial>` directive is used present a series of popup elements (usually `<md-button>`s) for
```
gkalpak: _is used **to** present_

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,130 @@
+(function() {
+  'use strict';
+
+  angular
+    .module('material.components.fabSpeedDial', [])
+    .directive('mdFabSpeedDial', MdFabSpeedDialDirective);
```
gkalpak: I am not sure what is the convention in the `ngMaterial` codebase (and feeling to lazy to go checkit out right now), but I prefer to have non-instantiable functions start with a lowercase.

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,130 @@
...
+      compile: function($element) {
+        // Wrap every child in a new div and add a class that we can scale independently
+        var children = $element.find('md-fab-actions').children(),
+          child, i;
+
+        for (i = 0; i < children.length; i++) {
+          child = children.eq(i);
+
+          child.wrap('<div class="md-fab-action-item">');
+        }
```
gkalpak: Lines 50 - 58 can be replaced with:

```js
$element.find('md-fab-actions').children().
  wrap('<div class="md-fab-action-item">');
```

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,130 @@
...
+          post: link
+        }
+      }
+    };
+
+    function FabSpeedDialController($scope, $element) {
+      var vm = this;
+
+      vm.scope = $scope;
+      vm.element = $element;
```
gkalpak: What is purpose of these 2 lines ? Seem totally redundant.

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,130 @@
...
+
+      // Set the default direction to 'down' if none is specified
+      if (typeof $scope.direction === 'undefined') {
+        $scope.$apply('direction = "down"');
```
gkalpak: Why use `$apply()` ? Why not just set the property ?

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,130 @@
...
+      }
+
+      $element.on('mouseenter', ctrl.focus);
+      $element.on('mouseleave', ctrl.blur);
+
+      // Watch for changes to the direction and update classes/attributes
+      $scope.$watch('direction', function(newDir, oldDir) {
+        // Add the appropriate classes so we can target the direction in the CSS
+        $element.removeClass('md-' + oldDir);
+        $element.addClass('md-' + newDir);
```
gkalpak: Maybe a little far-fetched, but using `$animate.add/removeClass()` would allow developers to set up nice transition animations. (Not that anyone is ever likely to do so...)

src/components/fabSpeedDial/fabSpeedDial.js
topherfangio: Just to confirm, you are saying that I should alter the directive/controller/template to use something like `vm.myScopeParam` instead of directly accessing `$scope`? I actually had this first and would be happy to revert to that.

src/components/fabSpeedDial/fabSpeedDial.scss
topherfangio: Ah, thanks for the code snippet. I haven't seen the variable interpolation within SCSS before. I'll update this.

src/components/fabSpeedDial/fabSpeedDial.scss
topherfangio: Using JS would also allow us to calculate the trigger height rather than specifying it in the SCSS. Could mean supporting more/different/custom FABs. I'll see what I can drum up.

src/components/fabSpeedDial/fabSpeedDial.js
topherfangio: I was basing this off of the existing directives I saw such as `MdButtonDirective` and `MdBottomSheetDirective`.

src/components/fabSpeedDial/fabSpeedDial.js
topherfangio: Here, you are correct, it is not necessary. Below, however, the isOpen value could legitimately be set to false and we don't want to override it.

src/components/fabSpeedDial/fabSpeedDial.js
topherfangio: Within the if statement, we should be able to. Within the `vm.focus`/`blur`, it wasn't updating the scope until something else triggered a digest. If there is a better way, I can totally change this, but @jelbourn suggested using the `scope.$apply` technique.

src/components/fabSpeedDial/fabSpeedDial.js
topherfangio: @ThomasBurleson said the same thing; I will give this a go when I have a slightly larger chunk of time to dedicate to it :-)

src/components/fabSpeedDial/fabSpeedDial.js
ThomasBurleson: We use **uppercamel-case** for Controller and Directive construction functions

gkalpak: My hopes are on @jelbourn to set you straight :stuck-out-tongue:

src/components/fabSpeedDial/fabSpeedDial.js
gkalpak: `$apply` is indeed necessary to notify Angular when something is invoked _from outside_.
Without getting into much detail, native/jqLite/jQuery event listener callbacks do need the `$apply`.
(Just make sure you are not calling the same functions from somewhere else while a digest cycle is already in progress.)

But they should be eliminated whenever possible (e.g. inside the ifs), because `$apply` does a great lot of stuff, which isn't necessary at that time.

src/components/fabSpeedDial/fabSpeedDial.js
topherfangio: @jelbourn or @ThomasBurleson - I am working on refactoring this into it's own directive. Should it also be in it's own module even though it is a "sub"-directive to the speed dial and toolbar? If so, does it need all of the proper documentation that usually comes with a directive or should I just point to the sd/toolbar docs?

src/components/fabSpeedDial/fabSpeedDial.js
jelbourn: Yes. Using the controller instance instead of the scope instance is much more in-line with how things iwll be in Angular 2 and is the recommendation of the angular team.

src/components/fabSpeedDial/fabSpeedDial.js
jelbourn: It should go in the same module. The documentation can all live with the top-level API surface.

src/components/fabSpeedDial/fabSpeedDial.js
jelbourn: The convention *should* be that only _constructors_ are PascalCase, where a constructor is is something that's called with the `new` keyword. 

A naming convention (PascalCase here) is the only way to communicate in JS that something is meant to be used as a constructor intended for use with the `new` keyword. This is the way of the Google js style guide and, as far as I know, is generally the accepted pattern among most of the wider JS community.

src/components/fabSpeedDial/fabSpeedDial.js
jelbourn: For that case you can just use
```js
this.isOpen = this.isOpen || false;
```

You shouldn't need that `$apply`

src/components/fabSpeedDial/fabSpeedDial.js
gkalpak: <sub>_All hope is lost... :(_</sub>

ajoslin:
At first sight, this looks way over-engineered to me...

Here's a version of the speed dial in pure css. A little bit of JS is needed for focus-management, but it's trivial: http://codepen.io/andrewtjoslin/pen/WvGaZv

ajoslin:
Now that I read the JS more carefully, I see that it's less than I thought. I just read all the comments and thought all this logic was being done in JS. But I see it's actually done well.

My mistake. *retreats back into the shadows*.

topherfangio:
@ThomasBurleson There are still a few pending tasks, but I have updated the code with almost all of the suggestions provided.

I could, however, use a bit more guidance on the toolbar animation changes that you suggested. I tried swapping the z-index of the trigger/background so that the trigger appears on top and then fades out, but it looked odd due to the trigger's drop shadow.

Based on the [animation in the spec](http://www.google.com/design/spec/components/buttons-floating-action-button.html#buttons-floating-action-button-transitions), it seems like the trigger itself expands into the toolbar, and only the icon fades. I'm not quite sure if we can achieve this with the current HTML structure, though, as we would have to modify the trigger element itself (in which case the whole toolbar would be clickable) or make the toolbar/background a child of the trigger (in which case it would get cropped to the size of the trigger).

I'm not really sure where to go from here, so I've left it pretty much the same as before until we can hammer this out a bit.

Also, I've modified the animation speed for both directives to use the defaults in `variables.scss`. I think this works well on the toolbar, but on the speed dial it feels a bit slow; particularly for the scale animation.

Thoughts? And thanks!

topherfangio:
Quick note: pending tasks are currently being stored in the commit message so that I can more easily track them.

ajoslin:
Tophier: set the scale and border radius of the toolbar so it becomes the
same size as the fab, then use a css transition back to default styles when
the toolbar is active. It's some simple math in JS

ajoslin:
http://codepen.io/andrewtjoslin/pen/ZGBBOW?editors=011

ThomasBurleson:
@topherfangio - want to setup and hangout and review this PR together ?

topherfangio:
@ThomasBurleson I think that would be a great idea; I should have taken your previous offer for help. I didn't realize there was quite so much left to do.

At the moment, I am fairly booked with a contract, but I'm sure I can find some time to get together. What works best for you? During the day or at night?

Also, you wouldn't happen to be in the San Francisco office would you? I am in town for a couple of weeks on business and would love to meet people for a beer or something :-)

gauravsaini03:
Any update when we will be having this merged ? Really eager to see floating fab buttons

topherfangio:
I sat down with @ThomasBurleson yesterday and made some modifications. He requested that I setup a new PR (https://github.com/angular/material/pull/3108) with the changes and close this one so that we can get it merged.

So, it should be done very soon ;-)

PR #2852, opened 2015-05-14
feat(build): snapshot docs site on successful build

matthewrfindley:
Pre-requisite for allowing users to select a version.

In order to serve older versions, we needed to update how the site is served.  We are following a similar pattern to how angularjs is being served using a seperate repo to store the snapshots and pre-built older versions (https://github.com/angular/code.material.angularjs.org).

This is only part of the larger puzzle, but I don't see any reason not to start capturing the snapshots now.  

The remaining pieces include, but not limited to, taking a snapshot of a release, serving the site using the new structure and implementing version selection.

### DIFF
--- a/scripts/snapshot-docs-site.sh
+++ b/scripts/snapshot-docs-site.sh
@@ -1,0 +1,41 @@
+#!/bin/bash
+
+ARG_DEFS=(
+  "--sha=(.*)"
+)
+
+function run {
+  cd ../
+
+  echo "-- Building docs site release..."
+  rm -rf dist
+  gulp docs --release
+
+  echo "-- Cloning code.material.angularjs.org..."
+  rm -rf code.material.angularjs.org
+  git clone https://github.com/angular/code.material.angularjs.org.git --depth=1
+
+  echo "-- Remove previous snapshot..."
+  rm -rf code.material.angularjs.org/snapshot
+
+  echo "-- Copying docs site to snapshot..."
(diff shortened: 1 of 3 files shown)


scripts/snapshot-docs-site.sh
```diff
@@ -0,0 +1,41 @@
...
+  echo "-- Building docs site release..."
+  rm -rf dist
+  gulp docs --release
+
+  echo "-- Cloning code.material.angularjs.org..."
+  rm -rf code.material.angularjs.org
+  git clone https://angular:$GH_TOKEN@github.com/angular/code.material.angularjs.org.git --depth=1
+
+  echo "-- Remove previous snapshot..."
+  rm -rf code.material.angularjs.org/snapshot
```
ThomasBurleson: Per discussion with @robertmesserle  - we need to rename `snapshot` to `HEAD`

scripts/snapshot-docs-site.sh
```diff
@@ -0,0 +1,41 @@
...
+
+  echo "-- Copying docs site to snapshot..."
+  cp -Rf dist/docs code.material.angularjs.org/snapshot
```
ThomasBurleson: See above

scripts/snapshot-docs-site.sh
matthewrfindley: I agree this makes more sense.  The change has been made.

ThomasBurleson:
@robertmesserle - this lgtm

Issue #2739, opened 2015-05-05
md-select interferes with the window scroll listener


PrashantAshok:
I created a [jsFiddle](http://jsfiddle.net/yLj8cc9w/1/). I have a sticky nav on scroll. If you click the select dropdown at the bottom of the page, since it removes the scroll bar, it messes (hides) the navigation bar at the top.  

On a quick glance, it looks like the `md-virtual-scroll-container` with `md-virtual-scroller` is wrapping the entire content inside `<body>`, hence hiding the scrollbar. I thought wrapping the md-select with md-content might help but it didn't make any difference. 

Any thoughts?

myagoo:
Same problem here, also, updating dom path seems dangerous, it can break css rules and other js behaviors. Why is this necessary ?

whoknows:
+1

PrashantAshok:
I think I found the problem in the code. In [util.js] (https://github.com/angular/material/blob/master/src/core/util/util.js#L57), I am not sure the need for checking `$document[0].body.scrollTop || $document[0].documentElement.scrollTop`. Just checking if the `lastParent` is defined or not should suffice. Something like 

```
if (!disableTarget) {
    if (!lastParent) {
        disableTarget = angular.element($document[0].body);
    } else {
        disableTarget = angular.element(lastParent);
    }
}
```
I did a quick test in Chrome/FF/IE - it seems to be working.

rschmukler:
@PrashantAshok can you check the latest source from master and see if your issue still persists?

PrashantAshok:
@rschmukler - thanks. Looks great now, that the body is not being wrapped. But the issue still persists. 

there is a minor issue, in [util.js](https://github.com/angular/material/blob/master/src/core/util/util.js#L137), shouldn't the check be `>=` rather than just `>`? Or am I missing something?

I had to make that change along with wrapping the md-select with `md-content`, Updated [jsFiddle](http://jsfiddle.net/k9yu5c6h/2/). Then works perfectly.

rschmukler:
An element's `scrollHeight` is always equal to its `offsetHeight` unless it is scrolling, in which case it is greater than it. So, in the case of `isScrolling()` we want to return true only if `>`.

The reason that your fiddle works with it that way is because it's not actually disabling the scrolling :) You can see that it's possible to scroll.

PrashantAshok:
Make sense. Oh, sorry didn't notice that. What's `md-scroll-disabled` class?

[...]

PR #2854, opened 2015-05-14
bug(select,mdUtil): fix disableScroll restructuing DOM

### LINKED ISSUE #2477: Sidenav auto-close on Nexus 5, Android5.1
I have created a sample application to understand sidenav functionality. To test that on nexus 5 it has a wired issue.
When i click on button to toggle the side nav. It opens the side nav and closes immediately.

Here is sample code.
HTML --->
`<body ng-app="PerformanceEvaluation">
        <div layout="row" layout-fill>
            <md-sidenav md-component-id="left" class="md-sidenav-left" >
                <md-content ng-controller="categoryController">
                    <md-list>
                        <md-subheader>Categories</md-subheader>
                        <md-list-item>
                            <div layout="row" layout-align="start center" style="padding:0 10px;" ng-repeat="category in categories">
                                
                                <p style="margin-left:10px;">{{category.title}}</p>
                            </div>
                        </md-list-item>
                    </md-list>
                </md-content>
            </md-sidenav>
            <div ng-controller="test">
                <md-button ng-click="toggleSideNav()">Toggle Side Nav</md-button>
            </div>
        </div>
    </body>`

js ---> 

```'use strict';
angular.module("PerformanceEvaluation", ['ngMaterial','PerformanceEvaluation.categories'])
.config(["$mdThemingProvider","$mdIconProvider",function ($mdThemingProvider, $mdIconProvider) {
    $mdThemingProvider.theme('default').primaryPalette('red').accentPalette('cyan').warnPalette('green');
}])
.controller("test",['$scope','$mdSidenav', function($scope, $mdSidenav){
    $scope.toggleSideNav = function(){
        if($mdSidenav("left").isOpen()){
            $mdSidenav("left").close();
        }
        else{
            $mdSidenav("left").open();
        }
    }
}]);```

to fill  the cotegories. --->

`'use strict';
angular.module("PerformanceEvaluation.categories",[])
.controller("categoryController",['$scope',function($scope){
    $scope.categories = [
        {
            "icon":'bower_components/material-design-icons/action/1x_web/ic_3d_rotation_black_18dp.png',
            "title":"tests"
        },
        {
            "icon":'bower_components/material-design-icons/action/1x_web/ic_3d_rotation_black_18dp.png',
            "title":"tests"
        },
        {
            "icon":'bower_components/material-design-icons/action/1x_web/ic_3d_rotation_black_18dp.png',
            "title":"tests"
        }
    ]
}])`

### LINKED ISSUE #2548: md-select: Causing stripe.js "CompleteDuplicationError" and "LoadDelayError" on open/close
So this is a werid one. Please take a look at the demo below.

It seems when you open or close an `md-select` it is causing `CompleteDuplicationError` and `LoadDelayError` in stripe js. I looked at the stripe js source and they both seem to be related to iframe dom issues and possibly socket errors. My gusse is that when you open an `md-select` there is an issue with the dom edit's that it does which is either removing stripes iframe or conflicting with it some other way.

I also just noticed that this only happens if the select is outside of `md-content` or in my case in the toolbar.

http://plnkr.co/edit/yP0gB9iTacIQb48thud8?p=preview

It also doesn't happen in `0.8.3` but it does in `0.9.0-rc2` and I'm suspecting it also happens in `0.9.0-rc1`. @rschmukler Maybe you can take a look at this. I asked the stripe people on irc and they didn't know much about it. But I think it is a dom issue with `md-select`.

The unobfuscated stripe js source can also be found below
https://js.stripe.com/v2/stripe-debug.js

### LINKED ISSUE #2564: md-select resets page scroll position.
I'm on: v0.9.0-rc2-master-fdcceb5
When the user clickes on the select element, the scroll position of its containing element is forced to the top of its scroll position. The panel of select options are then displayed at the bottom of the screen. Tested in firefox and chrome.

### LINKED ISSUE #2656: md-select ng-show/hide issue
md-select causes ng-hide/show on dom elements, resulting in re-rendering all ng-hide/show animation. 

plunker : http://plnkr.co/edit/o4VPvUUdfkFjzZrxwtRv?p=preview

### LINKED ISSUE #2692: md-select md-virtual-scroll-container breaks layout
The virtual scroll container that gets inserted into the DOM when the select box is clicked changes the DOM hierarchy, possibly breaking declared CSS rules.

When using SASS/Stylus etc. a lot of hierarchical CSS is used - like this:

``` css
.main {
  > div {
     > .content {
       ...
     }
   }
}
```

This breaks when DOM nodes are inserted in between.
I know it is possible to work around this by not using `>`, it might even be considered bad practice by some but I see a lot of people running into this problem.

Plunkr: http://plnkr.co/edit/XqY8NttC0euzMHkllULe?p=preview

### LINKED ISSUE #2783: md-select causes extra HTTP requests on page that uses an iframe
When `md-select` is opened, it takes all the elements in `body` and appends them to a virtual scroller. 
(it is ostensibly all element in the closest `md-content`, but the `md-select` menu is always appended directly to the body).

This is problematic because it makes the page fire HTTP requests if any of those elements contain an `iframe`. There are probably other, weirder issues related to wrapping the entire page in an extra div.

Reproduced: http://jsbin.com/mupipa/1/edit?html,js,output

This feature seems really invasive to the rest of the app just to prevent scrolling. 

As far as the behavior itself, some testing shows that the behavior of the native select actually depends on the underlying operating system:
* Linux (Ubuntu): closes the select when scrolling outside
* Mac: Freezes scrolling (actually all interactions outside of the select)
* Window (8): Same as Mac
* ChromeOS: TDB
* Android (Chrome): Opens a fully modal selection list, blocking all interaction w/ underlying page
* iOS: TBD

We need to either figure out a way to make to no-scroll behavior less invasive, or change the behavior to do something simpler (such as closing the select as Linux does).

### LINKED ISSUE #2828: md-select cuts off background to edge of content
This happened after updating to 0.9.0.

Here is a codepen illustrating the issue: (click the select)
http://codepen.io/anon/pen/yNOypa

rschmukler:
closes #2828, closes #2783, closes #2692, closes #2656, closes #2564,
closes #2548, closes #2477, references #2739

### DIFF
--- a/src/core/style/structure.scss
+++ b/src/core/style/structure.scss
@@ -146,16 +146,19 @@ input {
   margin: 8px;
 }
 
-.md-virtual-scroll-container {
-  display: block;
-  position: relative;
-  overflow: hidden;
-  width: 100%;
-}
-
-.md-virtual-scroller {
-  width: 100%;
-  min-height: 100%;
+.md-scroll-mask {
+  position: absolute;
+  background-color: transparent;
+  > .md-scroll-mask-bar {
+    display: block;
+    position: absolute;
+    background-color: #fafafa;
+    right: 0;
+    top: 0;
+    bottom: 0;
+    z-index: $z-index-scroll-mask;
+    box-shadow: inset 0px 0px 1px rgba(0, 0, 0, 0.3)
+  }
 }
 
 @media (min-width: $layout-breakpoint-sm) {
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -39,133 +39,104 @@ angular.module('material.core')
     offsetRect: function(element, offsetParent) {
       return Util.clientRect(element, offsetParent, true);
     },
+    // Disables scroll around the passed element. Goes up the DOM to find a
+    // disableTarget (a md-content that is scrolling, or the body as a fallback)
+    // and uses CSS/JS to prevent it from scrolling
     disableScrollAround: function(element) {
-      var parentContent = element instanceof angular.element ? element[0] : element;
-      var lastParent;
-      var disableTarget, scrollEl, useDocElement;
-      while (parentContent = this.getClosest(parentContent, 'MD-CONTENT', true)) {
-        if (isScrolling(parentContent)) {
-          disableTarget = angular.element(parentContent);
-        } else {
-          lastParent = parentContent;
+      element = element instanceof angular.element ? element[0] : element;
+      var parentEl = element;
+      var disableTarget;
+
+      // Find the highest level scrolling md-content
+      while (parentEl = this.getClosest(parentEl, 'MD-CONTENT', true)) {
+        if (isScrolling(parentEl)) {
+          disableTarget = angular.element(parentEl)[0];
         }
       }
+
+      // Default to the body if no scrolling md-content
       if (!disableTarget) {
-        if (!lastParent ||
-          $document[0].body.scrollTop || $document[0].documentElement.scrollTop ) {
-          disableTarget = angular.element($document[0].body);
-        } else {
-          disableTarget = angular.element(lastParent);
-        }
+        disableTarget = $document[0].body;
+        if (!isScrolling(disableTarget)) return angular.noop;
       }
 
-      if (disableTarget[0].nodeName == 'BODY' && $document[0].documentElement.scrollTop) {
-        scrollEl = $document[0].documentElement;
-        useDocElement = true;
+      if (disableTarget.nodeName == 'BODY') {
+        return disableBodyScroll();
       } else {
-        scrollEl = disableTarget[0];
+        return disableElementScroll();
       }
 
-      var heightOffset = scrollEl.scrollTop;
-      var originalWidth = scrollEl.clientWidth;
-
-      var restoreStyle = disableTarget.attr('style');
-      var disableStyle = $window.getComputedStyle(disableTarget[0]);
-      var wrapperEl = angular.element('<div class="md-virtual-scroll-container"><div class="md-virtual-scroller"></div></div>');
-      var virtualScroller = wrapperEl.children().eq(0);
-      virtualScroller.append(disableTarget[0].childNodes);
-      disableTarget.append(wrapperEl);
-      var originalScrollBarShow = originalWidth < scrollEl.clientWidth;
-
-      computeScrollbars(disableStyle);
-
-      virtualScroller.attr('layout-margin', disableTarget.attr('layout-margin'));
-      virtualScroller.css({
-        display: disableStyle.display,
-        '-webkit-flex-direction': disableStyle.webkitFlexDirection,
-        '-ms-flex-direction': disableStyle.msFlexDirection,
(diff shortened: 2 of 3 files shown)


src/core/style/structure.scss
```diff
@@ -146,16 +146,19 @@ input {
...
-.md-virtual-scroller {
-  width: 100%;
-  min-height: 100%;
+.md-scroll-mask {
+  position: absolute;
+  background-color: transparent;
+  > .md-scroll-mask-bar {
+    display: block;
+    position: absolute;
+    background-color: #fafafa;
```
jelbourn: Should this be a theme color?

src/core/style/structure.scss
```diff
@@ -146,16 +146,19 @@ input {
...
+    right: 0px;
+    top: 0px;
+    bottom: 0px;
```
jelbourn: The `px` part isn't necessary for `0`

src/core/style/structure.scss
```diff
@@ -146,16 +146,19 @@ input {
...
+    z-index: 5;
```
jelbourn: This should be in the z-index master list in `variables.scss`

src/core/util/util.js
```diff
@@ -40,132 +40,97 @@ angular.module('material.core')
       return Util.clientRect(element, offsetParent, true);
     },
     disableScrollAround: function(element) {
```
jelbourn: While this is fresh in your mind, can you add a function comment that explains what the function does and, at a high level, _how_ it does it?

src/core/util/util.js
```diff
@@ -40,132 +40,97 @@ angular.module('material.core')
...
+        $document.on('keydown', disableKeyNav);
+
+        return function restoreScroll() {
+          scrollMask.off('wheel');
+          scrollMask.off('touchmove');
+          scrollMask[0].parentNode.removeChild(scrollMask[0]);
+          $document.off('keydown', disableKeyNav);
+        };
+
+        function disableKeyNav(e) {
```
jelbourn: Could you comment on the intent of this function?

src/core/util/util.js
```diff
@@ -40,132 +40,97 @@ angular.module('material.core')
...
-        var scrollBarsShowing = !Util.floatingScrollbars()
-            && computedStyle.overflowY != 'hidden'
-            && (
-              virtualScroller[0].clientHeight > scrollEl.clientHeight
-              || originalScrollBarShow
-            );
-
-        var scrollerOffset = -1 * (heightOffset - parseFloat(disableStyle.paddingTop, 10));
-        disableTarget.css('padding-top', '0px');
+      // Converts the item to a position fixed div and translate it to the propper scroll position
```
jelbourn: What is "the item" ?

src/core/style/structure.scss
rschmukler: I don't think so because it's more of a browser emulation thing (most browsers don't allow styling the scroll bar color).

ThomasBurleson:
Closed with SHA 0cebafba47a6409cc41c

PR #2857, opened 2015-05-14
chore(util): simplify unique ID generation.

jelbourn:
The existing unique ID function was copied from an old version of angular and was originally written by Misko. It was then later replaced in Angular by Igor because it was a performance issue. 

What about overflow? `Number.MAX_SAFE_INTEGER` in js is over 9 quintillion. If you generated 100,000 IDs on every page, and loaded one page every second, it would take nearly 3,000 years to run out of IDs.

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -4,7 +4,7 @@
  * will create its own instance of this array and the app's IDs
  * will not be unique.
  */
-var nextUniqueId = ['0','0','0'];
+var nextUniqueId = 0;
 
 angular.module('material.core')
 .factory('$mdUtil', function($cacheFactory, $document, $timeout, $q, $window, $mdConstant) {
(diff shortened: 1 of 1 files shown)


src/core/util/util.js
```diff
@@ -254,34 +254,12 @@ angular.module('material.core')
...
-        if (digit == 90  /*'Z'*/) {
-          nextUniqueId[index] = '0';
-        } else {
-          nextUniqueId[index] = String.fromCharCode(digit + 1);
-          return nextUniqueId.join('');
-        }
-      }
-      nextUniqueId.unshift('0');
-      return nextUniqueId.join('');
+      return '' + nextUniqueId++;
```
gkalpak: I don't like `0` ids. I prefer `++nextUniqueId` (which is also closer to current Angular's implementation) :P

PR #2890, opened 2015-05-18
feat (mdTabs): Allow mdTab to not be direct child of mdTabs

FrigoEU:
Hey,

Right now the mdTabs/mdTab component assumes that if there is an mdTabs component above the mdTab, it will be it's direct parent. However, my use case is not like that. I have a component in between that shows/hides tabs based on the current user's rights. The calculation of the index fails in that case, making the tab order seemingly random.

This is a small patch that allows the mdTab components to not be direct children of the mdTabs component.

Simon

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2890)


robertmesserle:
LGTM

PR #2897, opened 2015-05-18
remove annotation as it fails closure compiler

tolysz:
This fails my closure.

    ERROR - @ngInject can only be used when defining a function or assigning a function expression.
        $get : function($$MdGestureHandler, $$rAF, $timeout) {
        ^


tolysz:
Ok, sorry for the noise, I need to update my build chain to use `ng-annotate -a in.js -o out.js`

Issue #2007, opened 2015-03-23
Bug: md-input-container textarea bug if inside of block with overflow: scroll


Nickproger:
Example here: http://codepen.io/anon/pen/emxYjW

Try input text with more then 10 lines.
Textarea jumps and i can't see text in the end.
Very strange behaviour.

RyanMullins:
+1

kikhtenko:
+1

andreykostenko:
+1

vovashevchenko:
+1

alexthewilde:
+1

chrisguerrero:
+1

ratscrew:
+1

Kourkis:
+1

sreevisakh:
+1

flaviolivolsi:
:+1:

slash13:
i love nantunes fix!!! also md-select wrapped in md-input-container makes absolute sense. 
+1

GustavSt:
This issue seems to be back

jalajc:
GustavSt +1,
Bug is even in 1.06
1. scroll-bar is not appearing
2. when in md-no-autogrow just one line of text appears (even if rows= 5, and one has 3 lines of text)
3. md-select-on-focus: just flashes the select and then select disappears.

PR #2899, opened 2015-05-19
Autogrow textarea fixes

nantunes:
Several changes I needed for a project.

One is a bug fix, the others are small new features, so it probably don't fulfill the pre-1.0 contribution guidelines. But I share it, maybe the patches are useful for someone else.

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2899)


src/components/input/input.js
```diff
@@ -243,7 +258,11 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
         onChangeTextarea();
       }
       element.on('keydown input', onChangeTextarea);
-      element.on('scroll', onScroll);
+
+      if(isNaN(min_rows)) {
+        element.on('scroll', onScroll);
```
nantunes: Don't really understand what onScroll does... Hence the if block.

src/components/input/input.scss
```diff
@@ -138,6 +138,10 @@ md-input-container {
       outline: none;
       box-shadow: none;
     }
+
+    &.md-no-flex {
+      flex: none !important;
```
ThomasBurleson: We try to avoid `!important` overrides. Is this the only solution?

ThomasBurleson:
@nantunes 
Please rebase with master and changes that include 52ca5e5ad3f56ba0555b62fad5669d98e1bbddeb.
Retest and determine if this PR is still valid and necessary.

nantunes:
(Did I screw up with the rebase?)

I think the PR is still valid, as the abilities to specify the textarea height with the "rows" attribute and to disable autogrow don't yet exist on master. I need these features in my project, but they're not critical so I leave it to your criterium.

Also the https://github.com/angular/material/commit/266520328315337731814a0cb00b1c0842837e81 still fixes #2007.

ThomasBurleson:
You will need to squash this PR or submit a new one with a single commit.

src/components/input/input.scss
nantunes: I don't really love this solution, as md-no-flex is an "utility" class. But I had no success defining temporarily the 'flex: none' on the element directly (where it would take precedence over all the other rules).

Do you have any suggestions?

nantunes:
Yep, sorry about that. Maybe it was that unintended merge commit.

I've opened #3544.

Issue #2313, opened 2015-04-14
mdBottomSheet: add md-bottomsheet-focus directive


dmackerman:
It would be wonderful if `$mdBottomSheet` had an open to set focus upon opening, just like `$mdDialog`. I'm aware that this breaks screen readers, but we would like to set focus manually versus not having the choice.

FWIW: the `focusOnOpen` docs from `mdDialog`:

> `focusOnOpen` - {boolean=}: An option to override focus behavior on open. Only disable if focusing some other way, as focus management is required for dialogs to be accessible. Defaults to true.

ThomasBurleson:
Recently we added a `md-sidenav-focus` directive to specify which element in the SideNav area should receive focus when the SideNav element is opened.

I think something similar for dialogs and bottomsheet would be very useful.

marcysutton:
That would be a good addition. It already does set focus on the first item in the bottom sheet, FWIW.

ThomasBurleson:
@marcysutton - true. With `md-dialog-focus` and `md-bottomsheet-focus` these would allow specific customized element focuses.

juanpujol:
+1

dmackerman:
Going to submit a pretty simple PR for this.

ThomasBurleson:
Fixed.

PR #2929, opened 2015-05-20
feat(bottomSheet): adds support for focusOnOpen

dmackerman:
This seems easy enough. I copied the pattern from `mdDialog`.

- [ ] write tests

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2929)

### DIFF
--- a/src/components/bottomSheet/bottomSheet.js
+++ b/src/components/bottomSheet/bottomSheet.js
@@ -71,8 +71,9 @@ function MdBottomSheetDirective() {
  *   - `controller` - `{string=}`: The controller to associate with this bottom sheet.
  *   - `locals` - `{string=}`: An object containing key/value pairs. The keys will
  *   be used as names of values to inject into the controller. For example,
- *   `locals: {three: 3}` would inject `three` into the controller with the value
- *   of 3.
+ *   `locals: {three: 3}` would inject `three` into the controller with the value of 3.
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
And why do you want this optional?

ThomasBurleson:
Refs https://github.com/angular/material/issues/2313

dmackerman:
Not sure I get what you're asking? I want a way to disable all focus events for bottomSheet, so users can implement their own focus methods. By default it should focus, for accessibility (as mentioned in `mdDialog`) and for the majority of bottom sheet usage focus seems appropriate. Sorry if I misunderstood!

ThomasBurleson:
@dmackerman - disregard my question, I found the original issue #2313 that discusses this. Thx

marcysutton:
Should we take this one step further and implement something similar to mdSidenav, which has the `md-sidenav-focus` directive?

dmackerman:
Up to you guys. This seems like a very lightweight option for now, but I can take a look at what it'd take to implement a similar `md-bottomsheet-focus` directive.

ThomasBurleson:
@dmackerman - while this PR is nice. I do not think it is an API extension we want. I think `md-bottomsheet-focus` is right way to go. @see https://github.com/angular/material/blob/master/src/components/sidenav/sidenav.js#L136-L144.

If you want to submit a PR for that feature, I would gladly review and merge. Thanks.

PR #2953, opened 2015-05-21
fix(textarea): Start at a height of one line

### LINKED ISSUE #2154: textarea initial height too tall
In Chrome, the initial height of a `<textarea>` in an `<md-input-container>` is set to the height of two lines of text, but according to the Material spec, it should start out at the height of a normal text field and expand as the user types more

http://www.google.com/design/spec/components/text-fields.html#text-fields-multi-line-text-field

This can be demonstrated on the angular material site demo page:

https://material.angularjs.org/#/demo/material.components.input

(just delete all the content from "Biography", for example).

calebegg:
Change min-height for textarea to one line height (and remove extra
padding space), and set rows=“1”, since it defaults to 2. Fixes #2154

[Review on Reviewable](https://reviewable.io/reviews/angular/material/2953)

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -235,6 +235,7 @@ function inputTextareaDirective($mdUtil, $window) {
       element.on('keydown input', onChangeTextarea);
       element.on('scroll', onScroll);
       angular.element($window).on('resize', onChangeTextarea);
+      element.attr('rows', '1');
 
       scope.$on('$destroy', function() {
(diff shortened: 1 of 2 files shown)


maelfrize:
:+1:

fabito:
:+1:

tehfreak:
:+1:

fludio:
:+1:

PR #3022, opened 2015-05-29
Make it clearer that ngMessages is needed for this

smehrbrodt:
The information that ngMessages is needed for errors is easily overlooked when it's part of the heading. It took me a while to figure this out.

So put this in an extra sentence below the example.


smehrbrodt:
signed the cla.

PR #3023, opened 2015-05-29
feat(autocomplete): add required attr to input

tsdragon0404:
The autocomplete don't validate required field.
I have some modify: 
- add ng-required="true" inner autocomplete tag
- add ng-required="isRequire" on function getInputElement()
- add attr.$observe('required', function (value) { scope.isRequire = value; }) on function link (scope, element, attr)
The attr work the same with ng-disabled.

[Review on Reviewable](https://reviewable.io/reviews/angular/material/3023)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,13 @@
+<a name="0.9.6"></a>
+### 0.9.6  (2015-05-28)
+
+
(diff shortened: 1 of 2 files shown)


DevVersion:
Closing this PR, because it's outdated and you just try to update the changelog.

Issue #2410, opened 2015-04-19
md-chip: md-chip-remove misaligned on Safari/Firefox 


websirnik:
Currently in 0.9.0-rc1 on Safari/Firefox. Test here: https://material.angularjs.org/#/demo/material.components.chips

![screen shot 2015-04-19 at 11 53 01](https://cloud.githubusercontent.com/assets/282871/7219175/17abe3e2-e68b-11e4-97a5-1193f790beb7.png)

Frank3K:
The first (black `x`) can be fixed by adding `vertical-align: bottom` [here](https://github.com/angular/material/blob/7d86788f607947774d247d5eef5e459fd8ecc616/src/components/chips/chips.scss#L82-L91). I did not test the latter (white `x`).

toddpress:
thanks @Frank3K! Tested code from a [pull-request](https://github.com/angular/material/pull/3043) and it also fixes this in all tested browsers. Tested in Firefox 35.0.1 and 38.0, Firefox Developer ed 40.0a2, and safari 8.0.6.

Issue #2920, opened 2015-05-20
md-chips : remove buttons not properly centered on Safari Mac OS X


kschwidder:
When running the demos under Safari on Mac OS X the remove button of a chip is not properly centered.

toddpress:
submitted a [pull request](https://github.com/angular/material/pull/3043) to offer a fix for this.

toddpress:
Also fixes issue [#2410] (https://github.com/angular/material/issues/2410). Confirmed to work in Firefox 35.0.1 and 38.0, Firefox Developer ed 40.0a2, and Safari 8.0.6.

ThomasBurleson:
Fixed with merge of #2410.

PR #3043, opened 2015-05-30
center align remove-chip button - issue #2920

toddpress:
Remove-chip button was improperly rendered in some browsers, such as
Safari 8 and FF Developer Edition. Added inline-level alignment, which
properly centers it.

[Review on Reviewable](https://reviewable.io/reviews/angular/material/3043)

### DIFF
--- a/src/components/chips/chips.scss
+++ b/src/components/chips/chips.scss
@@ -89,6 +89,8 @@ $contact-chip-name-width: rem(12) !default;
       box-shadow: none;
       margin: 0;
       position: relative;
+      display: inline-block;
+      vertical-align: middle;
       md-icon {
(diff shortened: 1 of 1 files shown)


toddpress:
I've signed it!

Frank3K:
+1. Would probably also fix #2410.

toddpress:
Confirmed to fix #2410 in Firefox 35.0.1 and 38.0, Firefox Developer ed. 40.0a2, and Safari 8.0.6. Tested on mac only.

EladBezalel:
@toddpress can you please provide a codepen of the problem?
before that, make sure it's still a valid pr on HEAD

Issue #3096, opened 2015-06-03
Navigating away from md-tabs triggers md-on-select


omida86:
In 0.9.7, every time I navigate away from a page that contains md-tabs using ng-click, it triggers "md-on-select".

Example at: http://plnkr.co/edit/zjuxgnB5z94s001Q63fQ?p=preview

omida86:
This does not happen if I change tab.

andrewburgess:
I'm experiencing this issue as well.

It seems to be caused by this line:
https://github.com/angular/material/blame/6c53d1127fc1cb7d2a86802e5831642a40ff1e41/src/components/tabs/js/tabsController.js#L184

Which was changed in relation to https://github.com/angular/material/issues/2837

That `removeTab` call is made, at least in my case, from here https://github.com/angular/material/blame/91576d0e7d7d1791e3688ba6ceac9a0750ea9e24/src/components/tabs/js/tabDirective.js#L127

If the tabs directive is being destroyed, we shouldn't care about selected tabs being changed I would assume.

shprink:
Having the same here!

nvanbenschoten:
Same issue here

Praggie:
facing same issue.

jaytula:
+1

CavemanNinja:
+1 0.9.8-rc1

robertmesserle:
This is resolved in 0.9.8 and higher.  Closing for now, please respond with browser details if you are experiencing it past 0.9.8.

@CavemanNinja Unfortunately, the fix was added just after RC1, but it should be good in 0.9.8.

Issue #2837, opened 2015-05-13
Few issues with tabs


gio82:
I'm dealing with few issues with tabs.  
Here below a brief description:

- Removing a tab do not call `md-on-select`
- Dynamically changing the tab label does not update the DOM
- When async loading some tabs the ink bar does not update his length 

Plus I think that you should add a method that allows to change tab via javascript.

epelc:
@gio82 on the last note if you set ` md-selected="selectedIndex"` on `md-tabs` you can change `selectedIndex` to switch between tabs.

I think the last bullet has to do with tabs not playing well with ng-if`s or dynamically adding/remove tabs see #2291

gio82:
Thanks for your help @epelc !

patriknyd:
+1

/ med vänlig hälsning
Patrik Nydensten

----- Reply message -----

PR #3105, opened 2015-06-03
fix(tabs): ensure $scope is not being destroyed when removing tab

andrewburgess:
Related to #3096  

Ensure that the $scope is not already destroyed before automatically selecting a new tab after removing one (which was implemented as part of a fix for #2837).

Alternative implementation could include passing a parameter to the `removeTab()` function indicating that the entire tab structure is being removed (not sure if `$scope.$$destroyed` is safe to use)


andrewburgess:
Signed the CLA

andrewburgess:
Looks like this was fixed by this commit https://github.com/angular/material/commit/1ba0686e7b42fef6072d011bf2f61fc3576a2017
Closing

PR #3108, opened 2015-06-03
Add the FAB Speed Dial and FAB Toolbar components.

### LINKED ISSUE #1416: mdSpeedDial: Implement floating-action-button menu

topherfangio:
Fixes #1416 - Adds support for two new components:
 1. fabSpeedDial - Creates a list of actions that expand from a single floating action button.
 2. fabToolbar - Creates a list of actions that transform from a floating action button into
    a toolbar.

Todo - The following are tasks we would like to implement in a future release.

 * Update keyboard support in-line with suggestions (i.e. tab to focus, then use arrow keys).
 * Add aria support.
 * Update animations to match new specs.
 * Use $animateCss to clean up code.

### DIFF
--- a/src/components/fabSpeedDial/fabSpeedDial.js
+++ b/src/components/fabSpeedDial/fabSpeedDial.js
@@ -1,0 +1,220 @@
+(function() {
+  'use strict';
+
+  angular
+    .module('material.components.fabSpeedDial', [
+      'material.core',
+      'material.components.fabTrigger',
+      'material.components.fabActions'
+    ])
+    .directive('mdFabSpeedDial', MdFabSpeedDialDirective)
+    .animation('.md-fling', MdFabSpeedDialFlingAnimation)
+    .animation('.md-scale', MdFabSpeedDialScaleAnimation);
+
+  /**
+   * @ngdoc directive
+   * @name mdFabSpeedDial
+   * @module material.components.fabSpeedDial
+   *
+   * @restrict E
+   *
+   * @description
+   * The `<md-fab-speed-dial>` directive is used to present a series of popup elements (usually
+   * `<md-button>`s) for quick access to common actions.
+   *
+   * There are currently two animations available by applying one of the following classes to
+   * the component:
+   *
+   *  - `md-fling` - The speed dial items appear from underneath the trigger and move into their
+   *    appropriate positions.
+   *  - `md-scale` - The speed dial items appear in their proper places by scaling from 0% to 100%.
+   *
+   * @usage
+   * <hljs lang="html">
+   * <md-fab-speed-dial direction="up" class="md-fling">
+   *   <md-fab-trigger>
+   *     <md-button aria-label="Add..."><md-icon icon="/img/icons/plus.svg"></md-icon></md-button>
+   *   </md-fab-trigger>
+   *
+   *   <md-fab-actions>
+   *     <md-button aria-label="Add User">
+   *       <md-icon icon="/img/icons/user.svg"></md-icon>
+   *     </md-button>
+   *
+   *     <md-button aria-label="Add Group">
+   *       <md-icon icon="/img/icons/group.svg"></md-icon>
+   *     </md-button>
+   *   </md-fab-actions>
+   * </md-fab-speed-dial>
+   * </hljs>
+   *
+   * @param {string=} md-direction From which direction you would like the speed dial to appear
+   * relative to the trigger element.
+   * @param {expression=} md-open Programmatically control whether or not the speed-dial is visible.
+   */
+  function MdFabSpeedDialDirective() {
+    return {
+      restrict: 'E',
+
+      scope: {
+        direction: '@?mdDirection',
+        isOpen: '=?mdOpen'
+      },
+
+      bindToController: true,
+      controller: FabSpeedDialController,
+      controllerAs: 'vm',
+
+      link: FabSpeedDialLink
+    };
+
+    function FabSpeedDialLink(scope, element) {
+      // Prepend an element to hold our CSS variables so we can use them in the animations below
+      element.prepend('<div class="md-css-variables"></div>');
+    }
+
+    function FabSpeedDialController($scope, $element, $animate) {
+      var vm = this;
+
+      // Define our open/close functions
+      // Note: Used by fabTrigger and fabActions directives
+      vm.open = function() {
+        $scope.$apply('vm.isOpen = true');
+      };
+
+      vm.close = function() {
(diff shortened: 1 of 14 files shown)


topherfangio:
Just to confirm: I signed the CLA under Profoundry LLC.

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -0,0 +1,215 @@
...
+   *
+   *  - `md-fling` - The speed dial items appear from underneath the trigger and move into their
+   *    appropriate positions.
+   *  - `md-scale` - The speed dial items appear in their proper places by scaling from 0% to 100%.
+   *
+   * @usage
+   * <hljs lang="html">
+   * <md-fab-speed-dial direction="up" class="md-fling">
+   *   <md-fab-trigger>
+   *     <md-button><md-icon icon="/img/icons/plus.svg"></md-icon></md-button>
```
marcysutton: Include `aria-label` on either `md-button` or `md-icon` in each usage.

ThomasBurleson:
@topherfangio -
SpeedDial 'left' or 'right' direction is not animating in the correct direction nor is it positioning properly if you change from 'up' or 'down'.
Also you will need to use:

```css
.lock-size {
  min-height: 300px;
  min-width: 300px;
  height: 300px;
  width: 300px;
}
```

topherfangio:
@ThomasBurleson 

1. **Need Input** - I see what you mean; this was actually the intended behavior as the items "open" toward the left or right (same with up/down). However, this is exactly opposite of the toolbar which is "positioned" to the left or right (even though it says direction). We should definitely fix one or the other to make them the same and update the documentation.

   *Do you have a preference?* If we go with position, I'd rather see `md-position=top,bottom,left,right`. I can refactor that super-quick if we want.

2. **Done** - I tried updating the `lock-size` as suggested, but I had to also add `margin-left/right: auto` to keep it centered.

ThomasBurleson:
1. I think the items should expand/scale left or right only when dir is the same. Similar to scale up/down with corresponding dir.

topherfangio:
@ThomasBurleson I added the fade effect yesterday to the speed dial as you suggested, and when doing so, I apparently introduced an opacity bug. If you switched the demo from fling to scale, and then back to fling again, the items never appeared because they were invisible.

I have fixed this, can you try again and see if it is working like you expect?

topherfangio:
@ThomasBurleson I think I may be confused :-)

> I think the items should expand/scale left or right only when dir is the same. Similar to scale up/down with corresponding dir.

In my mind, the speed dial is working exactly as you describe here, so I guess I am missing something.

Also, just saw your other comment:

> I also recommend using only 1 `md-content` container.

I have done this for the speed dial, however, if I also do it for the toolbar, the toolbar has whitespace around it that looks kinda weird since it doesn't reach the edge and thus has no boundary and "floats" so-to-speak.

topherfangio:
@ThomasBurleson @marcysutton Anything else I need to do to help get this merged? :-)

ThomasBurleson:
Topher,
You are good I think... I have been swamped with a mdIcon bug and have not
yet merged your PR.

Since I am OoO next week, I have asked @robertmesserle to merge your PR for
the m0.10.0-rc1 release. Talk with you during the week of June 16.

On Friday, June 5, 2015, Topher Fangio <notifications@github.com> wrote:

> @ThomasBurleson <https://github.com/ThomasBurleson> @marcysutton
> <https://github.com/marcysutton> Anything else I need to do to help get
> this merged? :-)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/angular/material/pull/3108#issuecomment-109449268>.
>


-- 
*Thomas Burleson*
Mindspace LLC

http://www.gridlinked.info
http://www.linkedin/in/ThomasBurleson
https://github.com/ThomasBurleson

miketran13:
Hi Topher, 
How can I make the tooltip of the button always visible ? I've already added this attribute:  md-visible="true" into the md-tooltip element. Unfortunately, it didn't work.

topherfangio:
@miketran13 In addition to the `md-visible="true"`, I think you also want the `md-autohide="false"` attribute set. The `md-visible` ensures that it is open on start, and the `md-autohide` ensures that it does not close when the button loses focus/hover.

Out of curiosity, is this a generic issue you are having, or is this issue specific to the speed dial/toolbar?

miketran13:
@topherfangio : thanks for your response. It's probably a generic issue when I'm using the md-autohide="false" & md-visible="true". It does not relate to the speed dial/toolbar.

tagazok:
Hi!
Is is the normal behaviour if the Fab Toolbar opens when I have the focus on the line and not only on the button?

topherfangio:
@robertmesserle Thanks for the merge! Quick question: it looks like two of the commits went un-merged. Was there a reason that they were left out? Should I have rebased the PR to be a single commit?

@tagazok Indeed, that is the normal behavior as I copied the though process from the Speed Dial component. However, now that you mention it, we should probably only open the toolbar if you hover the FAB. @ThomasBurleson thoughts? Should we create a separate issue for me to close?

robertmesserle:
@topherfangio We typically squash our feature branches down to a single commit before we merge in master - it makes it easier to track.  All of your commits should be in there, just squashed down to a single commit.

marcysutton:
@topherfangio I have some accessibility suggestions: I will open a new issue and ping you. 

People seem to have differing opinions about whether the components should open on hover (see #3214). I am more concerned with keyboard and screen reader support, but I think the hover effect could possibly still work even with modified keyboard functionality.

PR #3143, opened 2015-06-05
fix(tabs) resolves #3139

### LINKED ISSUE #3139: md-tabs programmatic selection of tab not updating pagination
When selectedIndex is changed programmatically, the tab is correctly selected and the correct tab content is shown.
But if the tabs are currently paginated and the newly selected tab is on a different page to the one currently shown, then the pagination is not updated, so you have no visibility of the selected tab header.

bug_before

Stexxen:
focusIndex was not being updated, so if pagination was being used, the current pagination view would not update to show the newly focused tab

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -82,7 +82,7 @@ function MdTabsController ($scope, $element, $window, $timeout, $mdConstant, $md
   function handleSelectedIndexChange (newValue, oldValue) {
     if (newValue === oldValue) return;
 
-    $scope.selectedIndex = getNearestSafeIndex(newValue);
+    $scope.selectedIndex = ctrl.focusIndex = getNearestSafeIndex(newValue);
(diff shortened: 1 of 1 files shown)


robertmesserle:
@Stexxen Thanks for the PR!  Unfortunately, this solution is similar to the original code being used here.  The reason that it was removed was because programmatically changing the selected index would hijack the browser's focus which wasn't always intended.

Stexxen:
@robertmesserle Thanks for the feedback. Can I clarify though, Programmatically changing selectIndex already works very well, and has use cases in wizards for example with the Next/Prev buttons.
This PR was just to fix a small edge case demonstrated in the first gif on the referenced issue. So I'm not trying to add programmatic tab selection, just make it consistent really.

Could you suggest how I could resolve the referenced issue in a different way that would be acceptable?

robertmesserle:
This should be fixed in 0.9.8 =)

Issue #2982, opened 2015-05-26
md-chips: md-require-match has no effect on md-chips behaviour


smashercosmo:
md-require-match attribute doesn't do anything because it is used in chips controller as "this.$scope.requireMatch", but should be used as "this.requireMatch".

Belicosus:
Also, md-require-match should be used with a true|false parameter:
`md-require-match="true"`

jahtalab:
+1
when this will be resolved?

houmark:
Has become somewhat related with #3833

smashercosmo:
Closed because of 86db4db9fd6e61314c1ce08cbffdffd52eae787f

PR #3246, opened 2015-06-12
fix(mdChipsController inputKeyDown): change use of property mdRequire…

Belicosus:
…Match on mdChips

the mdRequireMatch is not located on the $scope, due to a bug in the link or compile process.
The property is located on "this". This is a hotfix, and the proper solution is to fix the compile function of the directive

ISSUE: #2982


Belicosus:
Closed this one. Create new request with correct information for the CLA

PR #3262, opened 2015-06-15
Switched RawGit for GitCDN

schme16:
Due to the damage than can be caused to Ryan (owner of RawGit) if people abuse the dev link, I've switched it to GitCDN, it lets you get the latest master branch, without losing the CDN and without it's service getting harmed.

Full disclosure, I am the maintainer of GitCDN

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,15 +136,15 @@ when using online tools such as [CodePen](http://codepen.io/), [Plunkr](http://p
 
 > Note that the above sample references the 0.8.3 CDN release. Your version will change based on the latest stable release version.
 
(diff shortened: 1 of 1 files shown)


schme16:
I signed it!

gkalpak:
@ThomasBurleson, could also the "Edit in CodePen" thningy (at least for HEAD (master) version of the docs) benefit from GitCDN ?

schme16:
@gkalpak: I believe it will help just as much, good suggestion!

ThomasBurleson:
@gkalpak  - good catch/idea. I will make it so.

PR #3270, opened 2015-06-15
fix(mdChipsController inputKeyDown): Fix use of property mdRequireMatch

Belicosus:
the mdRequireMatch is not located on the $scope, due to a bug in the link or compile process.
The property is located on "this". This is a hotfix, and the proper solution is to fix the compile function of the directive

ISSUE: https://github.com/angular/material/issues/2982

[Review on Reviewable](https://reviewable.io/reviews/angular/material/3270)


EladBezalel:
Closed because of 86db4db

PR #3301, opened 2015-06-16
feat(Virtual repeater): Initial implementation

kseamon:
Implements a virtual repeater to use instead of ng-repeat for substantially
better performance in components.

Also tweaks the karma config to include component css.

### DIFF
--- a/src/components/virtualRepeater/virtualRepeater.js
+++ b/src/components/virtualRepeater/virtualRepeater.js
@@ -1,0 +1,530 @@
+/**
+ * @ngdoc module
+ * @name material.components.virtualRepeater
+ */
+angular.module('material.components.virtualRepeater', [
+  'material.core'
+])
+.directive('mdVirtualRepeatContainer', VirtualRepeatContainerDirective)
+.directive('mdVirtualRepeat', VirtualRepeatDirective);
+
+
+/**
+ * @ngdoc directive
+ * @name mdVirtualRepeatContainer
+ * @module material.components.virtualRepeat
+ * @restrict E
+ * @description
+ * `md-virtual-repeat-container` provides the scroll container for md-virtual-repeat.
+ *
+ * Virtual repeat is a limited substitute for ng-repeat that renders only
+ * enough dom nodes to fill the container and recycling them as the user scrolls.
+ *
+ * @usage
+ * <hljs lang="html">
+ * <md-virtual-repeat-container>
+ *   <div md-virtual-repeat="i in items" md-item-size="10">Hello {{i}}!</div>
+ * </md-virtual-repeat-container>
+ *
+ * <md-virtual-repeat-container md-orient-horizontal>
+ *   <div md-virtual-repeat="i in items" md-item-size="20">Hello {{i}}!</div>
+ * </md-virtual-repeat-container>
+ * </hljs>
+ *
+ * @param {boolean=} md-orient-horizontal Whether the container should scroll horizontally
+ *     (defaults to scrolling vertically).
+ */
+function VirtualRepeatContainerDirective() {
+  return {
+    controller: VirtualRepeatContainerController,
+    restrict: 'E',
+    template: virtualRepeatContainerTemplate,
+    compile: function virtualRepeatContainerCompile($element, $attrs) {
+      $element
+          .addClass('md-virtual-repeat-container')
+          .addClass($attrs.hasOwnProperty('mdOrientHorizontal')
+              ? 'md-orient-horizontal'
+              : 'md-orient-vertical');
+    }
+  };
+}
+
+
+function virtualRepeatContainerTemplate($element, $attrs) {
+  return '<div class="md-virtual-repeat-scroller">' +
+    '<div class="md-virtual-repeat-sizer"></div>' +
+    '<div class="md-virtual-repeat-offsetter">' +
+      $element[0].innerHTML +
+    '</div></div>';
+}
+
+
+/** @ngInject */
+function VirtualRepeatContainerController($$rAF, $scope, $element, $attrs) {
+  this.$scope = $scope;
+  this.$element = $element;
+  this.$attrs = $attrs;
+
+  /** @type {number} The width or height of the container */
+  this.size = 0;
+  /** @type {number} The scroll width or height of the scroller */
+  this.scrollSize = 0;
+  /** @type {number} The scrollLeft or scrollTop of the scroller */
+  this.scrollOffset = 0;
+  /** @type {boolean} Whether the scroller is oriented horizontally */
+  this.horizontal = this.$attrs.hasOwnProperty('mdOrientHorizontal');
+  /** @type {!VirtualRepeatController} The repeater inside of this container */
+  this.repeater = null;
+
+  this.scroller = $element[0].getElementsByClassName('md-virtual-repeat-scroller')[0];
+  this.sizer = this.scroller.getElementsByClassName('md-virtual-repeat-sizer')[0];
+  this.offsetter = this.scroller.getElementsByClassName('md-virtual-repeat-offsetter')[0];
+
+  $$rAF(angular.bind(this, this.updateSize));
+
+  // TODO: Come up with a more robust (But hopefully also quick!) way of
+  // detecting that we're not visible.
+  if ($attrs.ngHide) {
+    $scope.$watch($attrs.ngHide, angular.bind(this, function(hidden) {
+      if (!hidden) {
+        $$rAF(angular.bind(this, this.updateSize));
+      }
+    }));
+  }
+}
+
+
+/** Called by the md-virtual-repeat inside of the container at startup. */
+VirtualRepeatContainerController.prototype.register = function(repeaterCtrl) {
+  this.repeater = repeaterCtrl;
+  
+  angular.element(this.scroller)
+      .on('scroll wheel touchmove touchend', angular.bind(this, this.handleScroll_));
+};
+
+
+/** @return {boolean} Whether the container is configured for horizontal scrolling. */
+VirtualRepeatContainerController.prototype.isHorizontal = function() {
+  return this.horizontal;
+};
+
+
+/** @return {number} The size (width or height) of the container. */
+VirtualRepeatContainerController.prototype.getSize = function() {
+  return this.size;
+};
+
+
+/** Instructs the container to re-measure its size. */
+VirtualRepeatContainerController.prototype.updateSize = function() {
+  this.size = this.isHorizontal()
+      ? this.$element[0].clientWidth
+      : this.$element[0].clientHeight;
+  this.repeater && this.repeater.containerUpdated();
+};
+
+
+/** @return {number} The container's scrollHeight or scrollWidth. */
+VirtualRepeatContainerController.prototype.getScrollSize = function() {
+  return this.scrollSize;
+};
+
+
+/**
+ * Sets the scrollHeight or scrollWidth. Called by the repeater based on
+ * its item count and item size.
+ * @param {number} The new size.
+ */
+VirtualRepeatContainerController.prototype.setScrollSize = function(size) {
+  if (this.scrollSize !== size) {
+    this.sizer.style[this.isHorizontal() ? 'width' : 'height'] = size + 'px';
+  }
+
+  this.scrollSize = size;
+};
+
+
+/** @return {number} The container's current scroll offset. */
+VirtualRepeatContainerController.prototype.getScrollOffset = function() {
+  return this.scrollOffset;
+};
+
+
+VirtualRepeatContainerController.prototype.resetScroll = function() {
+  this.scroller[this.isHorizontal() ? 'scrollLeft' : 'scrollTop'] = 0;
+  this.handleScroll_();
+};
+
+
+VirtualRepeatContainerController.prototype.handleScroll_ = function() {
+  var transform, offset;
+  if (this.isHorizontal()) {
+    offset = this.scroller.scrollLeft;
+    transform = 'translateX(';
+  } else {
+    offset = this.scroller.scrollTop;
+    transform = 'translateY(';
+  }
+
+  if (offset === this.scrollOffset) return;
+  this.scrollOffset = offset;
+
+  var itemSize = this.repeater.getItemSize();
+  transform += Math.max(0, Math.floor(
+      this.scrollOffset / itemSize) - VirtualRepeatController.NUM_EXTRA) * itemSize + 'px)';
+
+  this.offsetter.style.webkitTransform = transform;
+  this.offsetter.style.transform = transform;
+
+  this.repeater.containerUpdated();
+};
+
+
+/**
+ * @ngdoc directive
+ * @name mdVirtualRepeat
+ * @module material.components.virtualRepeat
+ * @restrict A
+ * @priority 1000
+ * @description
+ * `md-virtual-repeat` specifies an element to repeat using virtual scrolling.
+ *
+ * Virtual repeat is a limited substitute for ng-repeat that renders only
+ * enough dom nodes to fill the container and recycling them as the user scrolls.
+ * Arrays, but not objects are supported for iteration.
+ * Track by, as alias, and (key, value) syntax are not supported.
+ *
+ * @usage
+ * <hljs lang="html">
+ * <md-virtual-repeat-container>
+ *   <div md-virtual-repeat="i in items" md-item-size="10">Hello {{i}}!</div>
+ * </md-virtual-repeat-container>
+ *
+ * <md-virtual-repeat-container md-orient-horizontal>
+ *   <div md-virtual-repeat="i in items" md-item-size="20">Hello {{i}}!</div>
+ * </md-virtual-repeat-container>
+ * </hljs>
+ *
+ * @param {number} md-item-size The height or width of the repeated elements (which
+ *     must be identical for each element). Required.
+ * @param {string=} md-extra-name Evaluates to an additional name to which
+ *     the current iterated item can be assigned on the repeated scope. (Needed
+ *     for use in md-autocomplete).
+ */
+function VirtualRepeatDirective($parse) {
+  return {
+    controller: VirtualRepeatController,
+    priority: 1000,
+    require: ['mdVirtualRepeat', '^^mdVirtualRepeatContainer'],
+    restrict: 'A',
+    terminal: true,
+    transclude: 'element',
+    compile: function VirtualRepeatCompile($element, $attrs) {
+      var expression = $attrs.mdVirtualRepeat;
+      var match = expression.match(/^\s*([\s\S]+?)\s+in\s+([\s\S]+?)\s*$/);
+      var repeatName = match[1];
+      var repeatListExpression = $parse(match[2]);
+      var extraName = $attrs.mdExtraName && $parse($attrs.mdExtraName);
+
+      return function VirtualRepeatLink($scope, $element, $attrs, ctrl, $transclude) {
+        ctrl[0].link_(ctrl[1], $transclude, repeatName, repeatListExpression, extraName);
+      };
+    }
+  };
+}
+
+
+/** @ngInject */
+function VirtualRepeatController($scope, $element, $attrs, $browser, $document) {
+  this.$scope = $scope;
+  this.$element = $element;
+  this.$attrs = $attrs;
+  this.$browser = $browser;
+  this.$document = $document;
+
+  /** @type {!Function} Backup reference to $browser.$$checkUrlChange */
+  this.browserCheckUrlChange = $browser.$$checkUrlChange;
+  /** @type {number} Most recent starting repeat index (based on scroll offset) */
+  this.newStartIndex = 0;
+  /** @type {number} Most recent ending repeat index (based on scroll offset) */
+  this.newEndIndex = 0;
+  /** @type {number} Previous starting repeat index (based on scroll offset) */
+  this.startIndex = 0;
+  /** @type {number} Previous ending repeat index (based on scroll offset) */
+  this.endIndex = 0;
+  // TODO: measure width/height of first element from dom if not provided.
+  // getComputedStyle?
+  /** @type {number} Height/width of repeated elements. */
+  this.itemSize = $scope.$eval($attrs.mdItemSize);
+  /**
+   * Presently rendered blocks by repeat index.
+   * @type {Object<number, !VirtualRepeatController.Block}
+   */
+  this.blocks = {};
+  /** @type {Array<!VirtualRepeatController.Block>} A pool of presently unused blocks. */
+  this.pooledBlocks = [];
+}
+
+
+/**
+ * An object representing a repeated item.
+ * @typedef {{element: !jqLite, new: boolean, scope: !angular.Scope}}
+ */
+VirtualRepeatController.Block;
+
+
+/**
+ * Number of additional elements to render above and below the visible area inside
+ * of the virtual repeat container. A higher number results in less flicker when scrolling
+ * very quickly in Safari, but comes with a higher rendering and dirty-checking cost.
+ * @const {number}
+ */
+VirtualRepeatController.NUM_EXTRA = 3;
+
+
+/**
+ * Called at startup by the md-virtual-repeat postLink function.
+ * @param {!VirtualRepeatContainerController} container The container's controller.
+ * @param {!Function} transclude The repeated element's bound transclude function.
+ * @param {string} repeatName The left hand side of the repeat expression, indicating
+ *     the name for each item in the array.
+ * @param {!Function} repeatListExpression A compiled expression based on the right hand side
+ *     of the repeat expression. Points to the array to repeat over.
+ * @param {string|undefined} extraName The optional extra repeatName.
+ */
+VirtualRepeatController.prototype.link_ =
+    function(container, transclude, repeatName, repeatListExpression, extraName) {
+  this.container = container;
+  this.transclude = transclude;
+  this.repeatName = repeatName;
+  this.repeatListExpression = repeatListExpression;
+  this.extraName = extraName;
+  this.sized = false;
+
+  this.container.register(this);
+};
+
+
+/**
+ * Called by the container. Informs us that the containers scroll or size has
+ * changed.
+ */
+VirtualRepeatController.prototype.containerUpdated = function() {
+  if (!this.sized) {
+    this.sized = true;
+    this.$scope.$watchCollection(this.repeatListExpression,
+        angular.bind(this, this.virtualRepeatUpdate_));
+    this.items = this.repeatListExpression(this.$scope);
+  }
+
+  this.updateIndexes_();
+
+  if (this.newStartIndex !== this.startIndex ||
+      this.newEndIndex !== this.endIndex ||
+      this.container.getScrollOffset() > this.container.getScrollSize()) {
+    this.virtualRepeatUpdate_(this.items, this.items);
+  }
+};
+
+
+/**
+ * Called by the container. Returns the size of a single repeated item.
+ * @return {number} Size of a repeated item.
+ */
+VirtualRepeatController.prototype.getItemSize = function() {
+  return this.itemSize;
+};
+
+
+/**
+ * Updates the order and visible offset of repeated blocks in response to scrolling
+ * or items updates.
+ * @private
+ */
+VirtualRepeatController.prototype.virtualRepeatUpdate_ = function(items, oldItems) {
+  var itemsLength = items ? items.length : 0;
+
+  // If the number of items shrank, scroll up to the top.
+  if (this.items && itemsLength < this.items.length && this.container.getScrollOffset() !== 0) {
+    this.items = items;
+    this.container.resetScroll();
+    return;
+  }
+
+  this.items = items;
+  if (items !== oldItems) {
+    this.updateIndexes_();
+  }
+
+  this.parentNode = this.$element[0].parentNode;
+  this.container.setScrollSize(itemsLength * this.itemSize);
+
+  // Detach and pool any blocks that are no longer in the viewport.
+  Object.keys(this.blocks).forEach(function(blockIndex) {
+    var index = parseInt(blockIndex, 10);
+    if (index < this.newStartIndex || index >= this.newEndIndex) {
+      this.poolBlock_(index);
+    }
+  }, this);
+
+  // Add needed blocks.
+  // For performance reasons, temporarily block browser url checks as we digest
+  // the restored block scopes ($$checkUrlChange reads window.location to
+  // check for changes and trigger route change, etc, which we don't need when
+  // trying to scroll at 60fps).
+  this.$browser.$$checkUrlChange = angular.noop;
+
+  var i, block,
+      newStartBlocks = [],
+      newEndBlocks = [];
+
+  // Collect blocks at the top.
+  for (i = this.newStartIndex; i < this.newEndIndex && this.blocks[i] == null; i++) {
+    block = this.getBlock_(i);
+    this.updateBlock_(block, i);
+    newStartBlocks.push(block);
+  }
+
+  // Update blocks that are already rendered.
+  for (; this.blocks[i] != null; i++) {
+    this.updateBlock_(this.blocks[i], i);
+  }
+  var maxIndex = i - 1;
+
+  // Collect blocks at the end.
+  for (; i < this.newEndIndex; i++) {
+    block = this.getBlock_(i);
+    this.updateBlock_(block, i);
+    newEndBlocks.push(block);
+  }
+
+  // Attach collected blocks to the document.
+  if (newStartBlocks.length) {
+    this.parentNode.insertBefore(
+        this.domFragmentFromBlocks_(newStartBlocks),
+        this.$element[0].nextSibling);
+  }
+  if (newEndBlocks.length) {
+    this.parentNode.insertBefore(
+        this.domFragmentFromBlocks_(newEndBlocks),
+        this.blocks[maxIndex] && this.blocks[maxIndex].element[0].nextSibling);
+  }
+
+  // Restore $$checkUrlChange.
+  this.$browser.$$checkUrlChange = this.browserCheckUrlChange;
+
+  this.startIndex = this.newStartIndex;
+  this.endIndex = this.newEndIndex;
+};
+
+
+/**
+ * @param {number} index Where the block is to be in the repeated list.
+ * @return {!VirtualRepeatController.Block} A new or pooled block to place at the specified index.
+ * @private
+ */
+VirtualRepeatController.prototype.getBlock_ = function(index) {
+  if (this.pooledBlocks.length) {
+    return this.pooledBlocks.pop();
+  }
+
+  var block;
+  this.transclude(angular.bind(this, function(clone, scope) {
+    block = {
+      element: clone,
+      new: true,
+      scope: scope
+    };
+
+    this.updateScope_(scope, index);
+    this.parentNode.appendChild(clone[0]);
+  }));
+
+  return block;
+};
+
+
+/**
+ * Updates and if not in a digest cycle, digests the specified block's scope to the data
+ * at the specified index.
+ * @param {!VirtualRepeatController.Block} block The block whose scope should be updated.
+ * @param {number} index The index to set.
+ * @private
+ */
+VirtualRepeatController.prototype.updateBlock_ = function(block, index) {
+  this.blocks[index] = block;
+
+  if (!block.new &&
+      (block.scope.$index === index && block.scope[this.repeatName] === this.items[index])) {
+    return;
+  }
+  block.new = false;
+
+  // Update and digest the block's scope.
+  this.updateScope_(block.scope, index);
+
+  // Perform digest before reattaching the block.
+  // Any resulting synchronous dom mutations should be much faster as a result.
+  // This might break some directives, but I'm going to try it for now.
+  if (!this.$scope.$root.$$phase) {
+    block.scope.$digest();
+  }
+};
+
+
+/**
+ * Updates scope to the data at the specified index.
+ * @param {!angular.Scope} scope The scope which should be updated.
+ * @param {number} index The index to set.
+ * @private
+ */
+VirtualRepeatController.prototype.updateScope_ = function(scope, index) {
+  scope.$index = index;
+  scope[this.repeatName] = this.items[index];
+  if (this.extraName) scope[this.extraName(this.$scope)] = this.items[index];
+};
+
+
+/**
+ * Pools the block at the specified index (Pulls its element out of the dom and stores it).
+ * @param {number} index The index at which the block to pool is stored.
+ * @private
+ */
+VirtualRepeatController.prototype.poolBlock_ = function(index) {
+  this.pooledBlocks.push(this.blocks[index]);
+  this.parentNode.removeChild(this.blocks[index].element[0]);
+  delete this.blocks[index];
+};
+
+
+/**
+ * Produces a dom fragment containing the elements from the list of blocks.
+ * @param {!Array<!VirtualRepeatController.Block>} blocks The blocks whose elements
+ *     should be added to the document fragment.
+ * @return {DocumentFragment}
+ * @private
+ */
+VirtualRepeatController.prototype.domFragmentFromBlocks_ = function(blocks) {
+  var fragment = this.$document[0].createDocumentFragment();
+  blocks.forEach(function(block) {
+    fragment.appendChild(block.element[0]);
+  });
+  return fragment;
+};
+
+
+/**
+ * Updates start and end indexes based on length of repeated items and container size.
+ * @private
+ */
+VirtualRepeatController.prototype.updateIndexes_ = function() {
+  var itemsLength = this.items ? this.items.length : 0;
+  var containerLength = Math.ceil(this.container.getSize() / this.itemSize);
+
+  this.newStartIndex = Math.max(0, Math.min(
+      itemsLength - containerLength,
+      Math.floor(this.container.getScrollOffset() / this.itemSize)));
+  this.newEndIndex = Math.min(itemsLength, this.newStartIndex + containerLength +
+      VirtualRepeatController.NUM_EXTRA);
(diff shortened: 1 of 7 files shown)


src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+    controller: VirtualRepeatContainerController,
+    replace: true,
+    require: 'virtualRepeatContainer',
+    restrict: 'E',
+    template: VirtualRepeatContainerTemplate
+  };
+}
+
+
+function VirtualRepeatContainerTemplate($element, $attrs) {
```
jelbourn: `virtualRepeatContainerTemplate` camelCase

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+  var innerHtml = $element[0].innerHTML;
+  $element[0].innerHTML = '';
+  // TODO: we probably don't need the container anymore.
```
jelbourn: What's up with this TODO?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+
+
+function VirtualRepeatContainerController($scope, $element, $attrs, $timeout, $window) {
+  this.$scope = $scope;
+  this.$element = $element;
+  this.$attrs = $attrs;
+  this.$timeout = $timeout;
+  this.$window = $window;
+
+  this.size = 0;
```
jelbourn: Could you add property descriptions for the properties that aren't just services?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+ * @param {string} lhs The left hand side of the repeat expression, indicating
+ *     the name for each item in the array.
+ * @param {!Function} rhs A compiled expression based on the right hand side
+ *     of the repeat expression. Points to the array to repeat over.
+ */
+VirtualRepeatController.prototype.link_ = function(container, transclude, lhs, rhs, extraName) {
+  this.container = container;
+  this.transclude = transclude;
+  this.lhs = lhs;
+  this.rhs = rhs;
```
jelbourn: What do you think of `repeatItemExpression` and `repeatListExpression` ?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+/**
+ * Called by the container. Returns the size of a single repeated item.
+ * @return {number} Size of a repeated item.
+ */
+VirtualRepeatController.prototype.getItemSize = function() {
+  return this.itemSize;
+};
+
+
+VirtualRepeatController.prototype.virtualRepeatUpdate_ = function(items, oldItems) {
```
jelbourn: + description?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+  var fragment = this.$document[0].createDocumentFragment();
+  blocks.forEach(function(block) {
+    fragment.appendChild(block.element[0]);
+  });
+  return fragment;
+};
+
+
+VirtualRepeatController.prototype.updateIndexes_ = function() {
+  var NUM_EXTRA = 3;
```
jelbourn: Extract `NUM_EXTRA` into a common constant?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+ *   <div md-virtual-repeat="i in items">Hello {{i}}!</div>
+ * </md-virtual-repeat-containr>
+ *
+ * <md-virtual-repeat-container md-horizonal>
+ *   <div md-virtual-repeat="i in items">Hello {{i}}!</div>
+ * </md-virtual-repeat-containr>
+ * </hljs>
+ *
+ * @param {number=} md-size The height or width of the repeated elements (which
+ *     must be identical for each element). TODO: If absent, the directive will attempt to
```
jelbourn: Should we remove the TODO and just say the attribute is required for now?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+
+  this.browserCheckUrlChange = $browser.$$checkUrlChange;
+  this.newStartIndex = 0;
+  this.newEndIndex = 0;
+  this.startIndex = 0;
+  this.endIndex = 0;
+  // TODO: measure width/height of first element from dom if not provided.
+  // getComputedStyle?
+  this.itemSize = $scope.$eval($attrs.mdSize);
+  this.blocks = {};
```
jelbourn: You use `blocks` a lot throughout the code, but it's not really explained what a `block` is in this context.

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+    var index = parseInt(blockIndex, 10);
+    if (index < this.newStartIndex || index >= this.newEndIndex) {
+      this.poolBlock_(index);
+    }
+  }, this);
+
+  // Add needed blocks.
+  // For performance reasons, temporarily block browser url checks as we digest
+  // the restored block scopes.
+  this.$browser.$$checkUrlChange = angular.noop;
```
jelbourn: Since `$$checkUrlChange` is not documented, perhaps it would be good to give a quick summary of what it does and why it's too slow for this.

src/components/virtualRepeater/virtualRepeater.scss
```diff
@@ -0,0 +1,57 @@
...
+  }
+
+  .md-virtual-repeat-offsetter {
+    box-sizing: border-box;
+    left: 0;
+    margin: 0;
+    padding: 0;
+    position: absolute;
+    // Leave room for the scorll bar.
+    // TODO: Will probably need to performa measurements at runtime.
```
jelbourn: TODO still relevant?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+ * <md-virtual-repeat-container md-horizonal>
+ *   <div md-virtual-repeat="i in items">Hello {{i}}!</div>
+ * </md-virtual-repeat-containr>
+ * </hljs>
+ *
+ * @param {number=} md-size The height or width of the repeated elements (which
+ *     must be identical for each element). TODO: If absent, the directive will attempt to
+ *     measure the computed style of the element at startup.
+ */
+function VirtualRepeatDirective($parse) {
```
jelbourn: Need `@ngInject` on your constructors (we have a tool that handles adding the `$inject` from that)

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
+/**
+ * @ngdoc module
+ * @name material.components.virtualRepeat
```
gkalpak: Why are files/directories called `virtualRepeater` while the module is `virtualRepeat` ?
(Is it a concious decision ?)

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+ * `md-virtual-repeat-container` provides the scroll container for md-virtual-repeat.
+ *
+ * Virtual repeat is a limited substitute for ng-repeat that allows renders only
+ * enough dom nodes to fill the container and recycling them as the user scrolls.
+ *
+ * @usage
+ * <hljs lang="html">
+ * <md-virtual-repeat-container md-horizonal>
+ *   <div md-virtual-repeat="i in items">Hello {{i}}!</div>
+ * </md-virtual-repeat-containr>
```
gkalpak: _containr_  -->  _container_

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+
+/**
+ * @ngdoc directive
+ * @name mdVirtualRepeatContainer
+ * @module material.components.virtualRepeat
+ * @restrict E
+ * @description
+ * `md-virtual-repeat-container` provides the scroll container for md-virtual-repeat.
+ *
+ * Virtual repeat is a limited substitute for ng-repeat that allows renders only
```
gkalpak: _allows renders_  -->  _renders_ (?)

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+ *
+ * @usage
+ * <hljs lang="html">
+ * <md-virtual-repeat-container md-horizonal>
+ *   <div md-virtual-repeat="i in items">Hello {{i}}!</div>
+ * </md-virtual-repeat-containr>
+ *
+ * <md-virtual-repeat-container md-horizonal>
+ *   <div md-virtual-repeat="i in items">Hello {{i}}!</div>
+ * </md-virtual-repeat-containr>
```
gkalpak: Isn't this identical with the above ? Is it intentional ?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+ * </hljs>
+ *
+ * @param {boolean=} md-horizontal Whether the container should scroll horizontally
+ *     (defaults to scrolling vertically).
+ */
+function VirtualRepeatContainerDirective() {
+  return {
+    controller: VirtualRepeatContainerController,
+    replace: true,
```
gkalpak: `replace` is deprecated. Although it is not very likely to be totally removed in the nearf future (not that it is improbable either), there are several corner-cases that don't work s expected and also reported bugs/issues are usually notdealt with (because of it being deprecated).

I would definitely try to stay away of it !

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+    require: 'virtualRepeatContainer',
```
gkalpak: This is redundant.

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+  return '<div class="md-virtual-repeat-container ' +
+      ($attrs.mdHorizontal ? 'md-horizontal' : 'md-vertical') + '">' +
+    '<div class="md-virtual-repeat-scroller">' +
+      '<div class="md-virtual-repeat-sizer"></div>' +
+      '<div class="md-virtual-repeat-offsetter">' +
+        innerHtml +
+      '</div>' +
+    '</div>' +
+  '</div>';
+}
```
gkalpak: Why not just use transclusion (probably along with an isolate scope) ?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+  this.scrollOffset = 0;
+  this.repeater = null;
+
+  this.scroller = $element[0].getElementsByClassName('md-virtual-repeat-scroller')[0];
+  this.sizer = this.scroller.getElementsByClassName('md-virtual-repeat-sizer')[0];
+  this.offsetter = this.scroller.getElementsByClassName('md-virtual-repeat-offsetter')[0];
+
+  this.handleScroll_ = angular.bind(this, this.handleScroll_, !!$attrs.mdHorizontal);
+
+  $window.requestAnimationFrame(angular.bind(this, this.updateSize));
```
gkalpak: Was there an internal service wrapper for raf ? (I might as well not recall correctly.)

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+  if ($attrs.ngHide) {
```
gkalpak: This seems too fragile ! What if someone uses `ngShow` ? Or `ngMaterial`'s `show/hide[-gt]-<size>` ?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+  this.size = 0;
+  this.scrollSize = 0;
+  this.scrollOffset = 0;
+  this.repeater = null;
+
+  this.scroller = $element[0].getElementsByClassName('md-virtual-repeat-scroller')[0];
+  this.sizer = this.scroller.getElementsByClassName('md-virtual-repeat-sizer')[0];
+  this.offsetter = this.scroller.getElementsByClassName('md-virtual-repeat-offsetter')[0];
+
+  this.handleScroll_ = angular.bind(this, this.handleScroll_, !!$attrs.mdHorizontal);
```
gkalpak: This shouldn't be necessary (since `$attrs` are available on `this`).

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+  this.repeater = repeaterCtrl;
+  
+  angular.element(this.scroller)
+      .on('scroll wheel touchmove touchend', this.handleScroll_);
+};
+
+
+/** @return {boolean} Whether the container is configured for horizontal scrolling. */
+VirtualRepeatContainerController.prototype.isHorizontal = function() {
+  return !!this.$attrs.mdHorizontal;
```
gkalpak: Since `mdHorizontal` is a boolean attribute, this should always evaluate to false.
Something like `this.$attrs.hasOwnProperty('mdHorizontal')` should be used instead (here and elsewhere).

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+
+/**
+ * @ngdoc directive
+ * @name mdVirtualRepeat
+ * @module material.components.virtualRepeat
+ * @restrict A
+ * @description
+ * `md-virtual-repeat` specifies an element to repeat using virtual scrolling.
+ *
+ * Virtual repeat is a limited substitute for ng-repeat that allows renders only
```
gkalpak: _allows renders_  -->  _renders_ (?)

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+ * enough dom nodes to fill the container and recycling them as the user scrolls.
+ *
+ * @usage
+ * <hljs lang="html">
+ * <md-virtual-repeat-container md-horizonal>
+ *   <div md-virtual-repeat="i in items">Hello {{i}}!</div>
+ * </md-virtual-repeat-containr>
```
gkalpak: _containr_  -->  _container_

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+ *
+ * <md-virtual-repeat-container md-horizonal>
+ *   <div md-virtual-repeat="i in items">Hello {{i}}!</div>
+ * </md-virtual-repeat-containr>
```
gkalpak: Identical example ?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+ *
+ * @param {number=} md-size The height or width of the repeated elements (which
+ *     must be identical for each element). TODO: If absent, the directive will attempt to
+ *     measure the computed style of the element at startup.
+ */
+function VirtualRepeatDirective($parse) {
+  return {
+    controller: VirtualRepeatController,
+    priority: 1000,
+    require: ['mdVirtualRepeat', '^mdVirtualRepeatContainer'],
```
gkalpak: `^^mdVirtualRepeatContainer` is better (it starts looking at the parent, not the element itself).

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+
+  this.repeater.containerUpdated();
+};
+
+
+/**
+ * @ngdoc directive
+ * @name mdVirtualRepeat
+ * @module material.components.virtualRepeat
+ * @restrict A
```
gkalpak: It is probably not used in the `ngMaterial` docs, but in Angular (core) there is also `@priority` (it might be a good idea to add it here as well).

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+ * @description
+ * `md-virtual-repeat` specifies an element to repeat using virtual scrolling.
+ *
+ * Virtual repeat is a limited substitute for ng-repeat that allows renders only
+ * enough dom nodes to fill the container and recycling them as the user scrolls.
+ *
```
gkalpak: The description should explain what subset of `ngRepeat`'s expression syntaxes are supported (e.g. no `track by`, no `(key, value)` etc).

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+    require: ['mdVirtualRepeat', '^mdVirtualRepeatContainer'],
+    restrict: 'A',
+    terminal: true,
+    transclude: 'element',
+    compile: function VirtualRepeatCompile($element, $attrs) {
+      var expression = $attrs.mdVirtualRepeat;
+      var match = expression.match(/^\s*([\s\S]+?)\s+in\s+([\s\S]+?)\s*$/);
+      var lhs = match[1];
+      var rhs = $parse(match[2]);
+      var extraName = $attrs.mdExtraName && $parse($attrs.mdExtraName);
```
gkalpak: Undocumented param ? What is this for ?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+    '</div>' +
+  '</div>';
+}
+
+
+function VirtualRepeatContainerController($scope, $element, $attrs, $timeout, $window) {
+  this.$scope = $scope;
+  this.$element = $element;
+  this.$attrs = $attrs;
+  this.$timeout = $timeout;
```
gkalpak: `$timeout` is unused. Remove it (ot use it :stuck_out_tongue:).

src/components/virtualRepeater/virtualRepeater.spec.js
```diff
@@ -0,0 +1,216 @@
+describe('<md-virtual-repeat>', function() {
+
+  beforeEach(module('material.components.virtualRepeat'));
+
+  var CONTAINER_HTML = '<md-virtual-repeat-container ' +
+      'style="height: 100px; width: 150px; box-sizing: border-box;">' +
+      '</md-virtual-repeat-cotainer>';
```
gkalpak: _cotainer_  -->  _container_

src/components/virtualRepeater/virtualRepeater.spec.js
```diff
@@ -0,0 +1,216 @@
+describe('<md-virtual-repeat>', function() {
+
+  beforeEach(module('material.components.virtualRepeat'));
+
+  var CONTAINER_HTML = '<md-virtual-repeat-container ' +
+      'style="height: 100px; width: 150px; box-sizing: border-box;">' +
```
gkalpak: Isn't the box-sizing set in the CSS file ?

src/components/virtualRepeater/virtualRepeater.spec.js
```diff
@@ -0,0 +1,216 @@
...
+
+  function mockAnimationFrame(fn) {
+    frameFunctions.push(fn);
+  }
+
+  mockAnimationFrame.flush = function() {
+    frameFunctions.forEach(function(fn) {
+      fn();
+    });
+  }
```
gkalpak: The functions should also be removed.

src/components/virtualRepeater/virtualRepeater.spec.js
```diff
@@ -0,0 +1,216 @@
...
+
+  function createRepeater() {
+    angular.element($document[0].body).append(container);
```
gkalpak: Is this necessary ?

src/components/virtualRepeater/virtualRepeater.spec.js
```diff
@@ -0,0 +1,216 @@
...
+    return component[0].querySelectorAll('[md-virtual-repeat]');
+  }
+
+  it('should render only enough items to fill the viewport + 3 (vertical)', function() {
+    createRepeater();
+    scope.items = createItems(100);
+    scope.$apply();
+    mockAnimationFrame.flush();
+
+    expect(getRepeated().length).toBe(100 / 10 + 3);
```
gkalpak: Lot's of magic numbers. A comment would be in place.

src/components/virtualRepeater/virtualRepeater.spec.js
```diff
@@ -0,0 +1,216 @@
...
+    var scopes = Array.prototype.map.call(getRepeated(), function(elem) {
+      var s = angular.element(elem).scope();
+      spyOn(s, '$digest').and.callThrough();
+      return s;
+    });
+
+    // Scroll up by one.
+    scroller[0].scrollTop = 90;
+    scroller.triggerHandler('scroll');
+    expect(scopes[15].$digest).toHaveBeenCalled();
```
gkalpak: Again, too much magic for me :(
(Why 14 and 15 :confused:)

src/components/virtualRepeater/virtualRepeater.scss
gkalpak: I think it is. Right now the scrollbar size is assumed to be 16px, but it's not the same across browsers/OSes/etc.

(I am sure there are battle-tested ways of doing that. I think Bootstrap has a function to measure scrollbar size, which I think they have taken from somewhere else (maybe jQueryUI).)

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+  this.itemSize = $scope.$eval($attrs.mdSize);
+  this.blocks = {};
+  this.pooledBlocks = [];
+}
+
+
+/**
+ * Called at startup by the md-virtual-repeat postLink function.
+ * @param {!VirtualRepeatContainerController} The container's controller.
+ * @param {!Function} The repeated element's bound transclude function.
```
gkalpak: Param name missing (here and above).

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+ * @param {string} lhs The left hand side of the repeat expression, indicating
+ *     the name for each item in the array.
+ * @param {!Function} rhs A compiled expression based on the right hand side
```
gkalpak: What does the `!` mean ?

src/components/virtualRepeater/virtualRepeater.js
gkalpak: Right now `lhs` is not an expression.

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+
+/**
+ * Called by the container. Informs us that the containers scroll or size has
+ * changed.
+ */
+VirtualRepeatController.prototype.containerUpdated = function() {
+  if (!this.sized) {
+    this.sized = true;
+    this.$scope.$watchCollection(this.rhs, angular.bind(this, this.virtualRepeatUpdate_));
+    this.items = this.rhs(this.$scope);
```
gkalpak: This is redundant (I think) - `$watchCollection()` will take care of that.

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,452 @@
...
+ */
+VirtualRepeatController.prototype.getItemSize = function() {
+  return this.itemSize;
+};
+
+
+VirtualRepeatController.prototype.virtualRepeatUpdate_ = function(items, oldItems) {
+  var itemsLength = items ? items.length : 0;
+
+  // If the number if items shrank, scroll up to the top.
```
gkalpak: _if_  -->  _of_ :smiley:

gkalpak:
I don't know how others feel, but big :+1: for ordering stuff (params, styles) alphabetically :smile:

src/components/virtualRepeater/virtualRepeater.js
ThomasBurleson: I think this is a doc bug. Should be `@name material.components.virtualRepeater`

gkalpak: OK, then `angular.module('...'` below, also needs fixing.

src/components/virtualRepeater/virtualRepeater.js
kseamon: Like ng-repeat, I want to avoid creating new scopes except for on the repeated items (which themselves should be non-isolate child scopes). I can probably get rid of 'replace', but I pretty much need this in order for a lot of use-cases to work properly.

gkalpak: This is different than `ngRepeat` (this is a wrapper around `virtualRepeat` - there is nothing similar with `ngRepeat`), so I don't get why you need it.
But it isn't too important either; I just thought it would be cleaner with transclusion.

src/components/virtualRepeater/virtualRepeater.js
kseamon: Now that I'm not using replace, it's moot.

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: went with repeatName and repeatListExpression

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.scss
kseamon: Leaving it as is. And yes, there are ways to do it that work, but I haven't gotten around to it yet.

src/components/virtualRepeater/virtualRepeater.js
kseamon: Added ngInject... does it need param annotations as well?

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: You don't want it to be a mid-2000s startup? (done)

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: updated

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

For what it's worth, Igor's official position is that it's never going to be removed, but I take your point that it's best to avoid when possible.

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: $$rAF apparently.

src/components/virtualRepeater/virtualRepeater.js
kseamon: Added a TODO. This is a trickier issue than I'd like to tackle right this moment and it shouldn't impact Jeremy's calendar component that I'm trying to unblock with this.

src/components/virtualRepeater/virtualRepeater.js
kseamon: done. Using isHorizontal everywhere now.

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.spec.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.spec.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.spec.js
kseamon: In order for it to have a real height or width, yes.

src/components/virtualRepeater/virtualRepeater.spec.js
kseamon: added in comments and cosntants

src/components/virtualRepeater/virtualRepeater.spec.js
kseamon: added comments

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
kseamon: non-nullable (reference types are nullable by default)

src/components/virtualRepeater/virtualRepeater.js
kseamon: It is needed. watchCollection doesn't seem to run the watcher synchronously.

src/components/virtualRepeater/virtualRepeater.js
kseamon: done

src/components/virtualRepeater/virtualRepeater.js
jelbourn: Just the @ngInject should be good

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,530 @@
...
+ */
+function VirtualRepeatContainerDirective() {
+  return {
+    controller: VirtualRepeatContainerController,
+    restrict: 'E',
+    template: virtualRepeatContainerTemplate,
+    compile: function virtualRepeatContainerCompile($element, $attrs) {
+      $element
+          .addClass('md-virtual-repeat-container')
+          .addClass($attrs.mdHorizontal ? 'md-horizontal' : 'md-vertical');
```
gkalpak: `$attrs.mdHorizontal` won't work as expected if `md-horizontal` is a boolean attribute.
`$attrs.hasOwnProperty('mdHorizontal')` ?

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,530 @@
...
+
+function virtualRepeatContainerTemplate($element, $attrs) {
+  var innerHtml = $element[0].innerHTML;
+  $element[0].innerHTML = '';
+
+  return '<div class="md-virtual-repeat-scroller">' +
+    '<div class="md-virtual-repeat-sizer"></div>' +
+    '<div class="md-virtual-repeat-offsetter">' +
+      innerHtml +
+    '</div></div>';
```
gkalpak: There is no need to empty the element, when you specify a template.
The function body could be replaced with:

```js
return
    '<div class="md-virtual-repeat-scroller">' +
      '<div class="md-virtual-repeat-sizer"></div>' +
      '<div class="md-virtual-repeat-offsetter">' + $element.html() + '</div>' +
    '</div>';
```

src/components/virtualRepeater/virtualRepeater.spec.js
```diff
@@ -0,0 +1,221 @@
+describe('<md-virtual-repeat>', function() {
+
+  beforeEach(module('ngMaterial-mock', 'material.components.virtualRepeater'));
+
+  var CONTAINER_HTML = '<md-virtual-repeat-container ' +
+      'style="height: 100px; width: 150px">' +
+      '</md-virtual-repeat-cotainer>';
```
gkalpak: _cotainer_  -->  _container_

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,530 @@
...
+  
+  angular.element(this.scroller)
+      .on('scroll wheel touchmove touchend', angular.bind(this, this.handleScroll_));
+};
+
+
+/** @return {boolean} Whether the container is configured for horizontal scrolling. */
+VirtualRepeatContainerController.prototype.isHorizontal = function() {
+  return this.$attrs.hasOwnProperty('mdHorizontal');
+};
```
gkalpak: Calling `hasOwnProperty()` too much might not be very efficient. Why not store the result in a variable instead ?
If the plan is to be able to dynamically adjust to direction switches, then the `md-virutal-repeat-container` class needs to also update accordingly.

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,530 @@
...
+ * Virtual repeat is a limited substitute for ng-repeat that renders only
+ * enough dom nodes to fill the container and recycling them as the user scrolls.
+ *
+ * @usage
+ * <hljs lang="html">
+ * <md-virtual-repeat-container>
+ *   <div md-virtual-repeat="i in items" md-size="10">Hello {{i}}!</div>
+ * </md-virtual-repeat-container>
+ *
+ * <md-virtual-repeat-container md-horizonal>
```
gkalpak: _md-horizonal_  -->  _md-horizontal_

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,530 @@
...
+ * Arrays, but not objects are supported for iteration.
+ * Track by and (key, value) syntax are not supported.
+ *
+ * @usage
+ * <hljs lang="html">
+ * <md-virtual-repeat-container>
+ *   <div md-virtual-repeat="i in items" md-size="10">Hello {{i}}!</div>
+ * </md-virtual-repeat-container>
+ *
+ * <md-virtual-repeat-container md-horizonal>
```
gkalpak: _md-horizonal_  -->  _md-horizontal_

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,530 @@
...
+ * @module material.components.virtualRepeat
+ * @restrict A
+ * @priority 1000
+ * @description
+ * `md-virtual-repeat` specifies an element to repeat using virtual scrolling.
+ *
+ * Virtual repeat is a limited substitute for ng-repeat that renders only
+ * enough dom nodes to fill the container and recycling them as the user scrolls.
+ * Arrays, but not objects are supported for iteration.
+ * Track by and (key, value) syntax are not supported.
```
gkalpak: The `as alias` is also not supported.

src/components/virtualRepeater/virtualRepeater.spec.js
```diff
@@ -0,0 +1,221 @@
...
+  var REPEATER_HTML = '<div md-virtual-repeat="i in items" md-size="10" ' +
+      'style="height: 10px; width: 10px; box-sizing: border-box;">' +
+      '{{i}} {{$index}}</div>';
+  var container, repeater, component, $$rAF, $compile, $document, scope,
+      frameFunctions, scroller, sizer, offsetter;
```
gkalpak: `frameFunctions` is not needed any more.

src/components/virtualRepeater/virtualRepeater.spec.js
```diff
@@ -0,0 +1,221 @@
...
+
+  beforeEach(inject(function(_$$rAF_, _$compile_, _$document_, $rootScope) {
+    repeater = angular.element(REPEATER_HTML);
+    container = angular.element(CONTAINER_HTML).append(repeater);
+    component = null;
+    $$rAF = _$$rAF_;
+    $compile = _$compile_;
+    $document = _$document_;
+    scope = $rootScope.$new();
+    frameFunctions = [];
```
gkalpak: `frameFunctions` is not needed any more.

src/components/virtualRepeater/virtualRepeater.spec.js
```diff
@@ -0,0 +1,221 @@
...
+    scope.$apply();
+    $$rAF.flush();
+
+    expect(getRepeated().length)
+        .toBe(VERTICAL_PX / ITEM_SIZE + VirtualRepeatController.NUM_EXTRA);
+    expect(sizer[0].offsetHeight).toBe(NUM_ITEMS * ITEM_SIZE);
+  });
+
+  it('should render only enough items to fill the viewport + 3 (horizontal)', function() {
+    container.attr('md-horizontal', 'true');
```
gkalpak: It would be more _realistic_ to replace `'true'` with `''` or something.

src/components/virtualRepeater/virtualRepeater.spec.js
```diff
@@ -0,0 +1,221 @@
...
+    scroller[0].scrollTop = 1000;
+    scroller.triggerHandler('scroll');
+    expect(offsetter.css('transform')).toBe('translateY(970px)');
+    repeated = getRepeated();
+    expect(repeated.length).toBe(VERTICAL_PX / ITEM_SIZE + VirtualRepeatController.NUM_EXTRA);
+    expect(repeated[0].textContent).toBe('s194s 97');
+  });
+
+  it('should reposition and swap items on scroll (horizontal)', function() {
+    container.attr('md-horizontal', 'true');
```
gkalpak: Same as above (no `'true'`).

src/components/virtualRepeater/virtualRepeater.spec.js
```diff
@@ -0,0 +1,221 @@
...
+    expect(scopes[15].$digest).toHaveBeenCalled();
+    expect(scopes[14].$digest).not.toHaveBeenCalled();
+
+    // Scroll down by two.
+    // Expect only the first scope to have $digested.
+    scroller[0].scrollTop = 110;
+    scroller.triggerHandler('scroll');
+    expect(scopes[0].$digest).toHaveBeenCalled();
+    expect(scopes[1].$digest).not.toHaveBeenCalled();
+  });
```
gkalpak: It's probably me, but I still can't figure out why 15/14 or 0/1 :disappointed:
(Could someone, please, explain ?)

src/components/virtualRepeater/virtualRepeater.js
```diff
@@ -0,0 +1,530 @@
...
+/**
+ * Called by the container. Informs us that the containers scroll or size has
+ * changed.
+ */
+VirtualRepeatController.prototype.containerUpdated = function() {
+  if (!this.sized) {
+    this.sized = true;
+    this.$scope.$watchCollection(this.repeatListExpression,
+        angular.bind(this, this.virtualRepeatUpdate_));
+    this.items = this.repeatListExpression(this.$scope);
```
gkalpak: Pretty minor (since it only happens once), but you might be able to replace this line with `return;`.
(Because, the `virtualRepeatUpdate_()` function will (eventually) get called and also call `updateIndexes_()`.)

<sub>It's propably not worth the trouble :)</sub>

src/components/virtualRepeater/virtualRepeater.spec.js
kseamon: As you scroll, it swaps items between the top and bottom and digests only the item it swapped as the data in the others will not have changed. I'm testing that only the first and last items were digested given that i've scrolled a net of one step in either direction.

src/components/virtualRepeater/virtualRepeater.js
kseamon: Even doing that breaks it. Probably not worth the effort of diving in any further.

src/components/virtualRepeater/virtualRepeater.spec.js
gkalpak: Is it because the blocks/elements are pooled ?
So, the element at index 15 is removed from bottom, pooled and then used again at the top ?

PR #3302, opened 2015-06-16
check tabs length before update tab order

pan-long:
When `tabs` is empty, updating tab order will cause `$scope.selectedIndex` to be `-1`, which could result in exception when we try to access  selected tab.


pan-long:
I signed it!

PR #3307, opened 2015-06-17
Make md-select work more like md-input-container/input

MagicIndustries:
Adds required md-input-container to md-select. 
Replaces md-select-label with md-select-value
Adjusts scss to make label transition the same way as with md-input-container/md-input
Removed md-select blacklist from placeholder

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -3,15 +3,15 @@
  * @name material.components.input
  */
 
-angular.module('material.components.input', [
+angular.module('material.components.input',[
   'material.core'
 ])
-  .directive('mdInputContainer', mdInputContainerDirective)
-  .directive('label', labelDirective)
-  .directive('input', inputTextareaDirective)
-  .directive('textarea', inputTextareaDirective)
-  .directive('mdMaxlength', mdMaxlengthDirective)
-  .directive('placeholder', placeholderDirective);
+  .directive('mdInputContainer',mdInputContainerDirective)
+  .directive('label',labelDirective)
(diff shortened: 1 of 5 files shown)


thvd:
I think we get a much nicer diff when you revert the 4 space indent back to the 2 space indent.

MagicIndustries:
Just noticed that, I'm not sure why it's gone in that way, it's 2 spaces in my IDE. I just cloned the repo, added my changes and re-committed. Shouldn't be anything adding extra spaces - I'm on OSX. 

I just checked, in my fork it's two spaces as well. Did I accidentally tick a box somewhere when making the pull request?

MagicIndustries:
Ah, I was checking select, it's the others that were messed up. I'll re-send the pull request.

MagicIndustries:
Never mind, just realised the commit was added to the request :)

nantunes:
You should also try to keep with the project's code style. The diff is polluted with whitespace surrounding the function declarations.

Meanwhile, a cleaner diff: https://github.com/angular/material/pull/3307/files?w=1

Looking forward to the enhancement. :+1:

MagicIndustries:
Thanks for the feedback @nantunes, I've updated the whitespace to remove as much of that noise as possible.

dmackerman:
I'm all about this. I've always been confused by the lack of `<md-input-container>` with selects.

dholcombe:
Thanks for doing this. I just implemented something similar in my site (the CSS portions anyhow). This more complete solution will be helpful when it ships.

PR #3316, opened 2015-06-17
fix($mdUtil): make "initOptionalProperties" compatible with angular >= 1.4.1

### LINKED ISSUE #3315: $mdUtil.initOptionalProperties breaks with angular 1.4.1
Since this commit in angularjs: https://github.com/angular/angular.js/commit/a6339d30d1379689da5eec9647a953f64821f8b0
$mdUtil.initOptionalProperties could not relay on presence of attr in $attrs to determinate "hasKey" variable value.

As for now, this causes invalid defaults, for example in mdTabs.

I'm working on a PR with updated initOptionalProperties.

Also I think it will be much safer (and more readable) to explicit provide defaults in all components.

mprokopowicz:
starting form angular 1.4.1 attrs defined in directive isolated scope are always present in $attrs

Closes #3315

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -334,15 +334,16 @@ angular.module('material.core')
     },
 
     /**
-     * Give optional properties with no value a boolean true by default
+     * Give optional properties with no value a boolean true if attr provided or false otherwise
      */
     initOptionalProperties: function (scope, attr, defaults ) {
        defaults = defaults || { };
        angular.forEach(scope.$$isolateBindings, function (binding, key) {
+         var normalizedAttrName = attr.$normalize(binding.attrName);
          if (binding.optional && angular.isUndefined(scope[key])) {
-           var hasKey = attr.hasOwnProperty(attr.$normalize(binding.attrName));
+           var attrIsDefined = normalizedAttrName in attr &&  !angular.isUndefined(attr[normalizedAttrName]);
 
-           scope[key] = angular.isDefined(defaults[key]) ? defaults[key] : hasKey;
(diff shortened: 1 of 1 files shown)


src/core/util/util.js
```diff
@@ -334,15 +334,16 @@ angular.module('material.core')
     },
 
     /**
-     * Give optional properties with no value a boolean true by default
+     * Give optional properties with no value a boolean true if attr provided or false otherwise
      */
     initOptionalProperties: function (scope, attr, defaults ) {
        defaults = defaults || { };
        angular.forEach(scope.$$isolateBindings, function (binding, key) {
+         var normalizedAttrName = attr.$normalize(binding.attrName);
```
gkalpak: This is redundant (I don't know why it was there in the first place).
`binding.attrName` is already normalized.

src/core/util/util.js
```diff
@@ -334,15 +334,16 @@ angular.module('material.core')
...
          if (binding.optional && angular.isUndefined(scope[key])) {
-           var hasKey = attr.hasOwnProperty(attr.$normalize(binding.attrName));
+           var attrIsDefined = normalizedAttrName in attr &&  !angular.isUndefined(attr[normalizedAttrName]);
```
gkalpak: Isn't `var attrIsDefined = angular.isDefined(attr[binding.attrName])` enough ?

mprokopowicz: Yep, that looks much better :)

mprokopowicz:
I've simplified this fix as gkalpak suggested.

PR #3319, opened 2015-06-17
README.md

mukilane:
Updated the CDN versions. (from 0.8.3 to 0.10.0)


ThomasBurleson:
@robertmesserle - Do you think the release script should modify/update the README also ?

PR #3329, opened 2015-06-17
fix(tabs): check for undefined tab

pan-long:
check for undefined tab before update tabs order. This is the same
  check as in updateInkBarStyles.

[Review on Reviewable](https://reviewable.io/reviews/angular/material/3329)


pan-long:
I did author these commits.

ThomasBurleson:
@pan-long  - Please rebase with master, confirm that this PR is still valid/needed, and confirm that Travis builds properly for this PR.

PR #3355, opened 2015-06-19
Autocomplete: Stop Event propagation for Key.ESC 

tirana:
1 line change to stop event propagation if ESC was pressed.
Reason: I use an autocomplete field in a dialog, so ESC closes the dialog too.


tirana:
I signed it!

Best regards,
Anatoli Kuzmin

tirana:
broken tests. withdraw

PR #3391, opened 2015-06-22
docs(virtualRepeat): fix example

gkalpak:
[Review on Reviewable](https://reviewable.io/reviews/angular/material/3391)

### DIFF
--- a/src/components/virtualRepeat/virtualRepeater.js
+++ b/src/components/virtualRepeat/virtualRepeater.js
@@ -22,7 +22,9 @@ angular.module('material.components.virtualRepeat', [
  *
  * @usage
  * <hljs lang="html">
- * <md-virtual-repeat-container></md-virtual-repeat-container>
+ * <md-virtual-repeat-container>
+ *   <div md-virtual-repeat="i in items" md-item-size="10">Hello {{i}}!</div>
+ * </md-virtual-repeat-container>
  *
  * <md-virtual-repeat-container md-orient-horizontal>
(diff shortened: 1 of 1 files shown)


src/components/virtualRepeat/virtualRepeater.js
```diff
@@ -22,7 +22,9 @@ angular.module('material.components.virtualRepeat', [
  *
  * @usage
  * <hljs lang="html">
- * <md-virtual-repeat-container></md-virtual-repeat-container>
+ * <md-virtual-repeat-container>
+ *   <div md-virtual-repeat="i in items" md-item-size="10">Hello {{i}}!</div>
+ * </md-virtual-repeat-container>
```
ThomasBurleson: @gkalpak - do we need to show two examples with essentially the same markup?

ThomasBurleson: @gkalpak - What are your thoughts here? It sees that md-item-size is misleading; 
*  the attribute is not the size but rather the md-num-visible-items, and  
*  the actual pixel size is dependent upon the full-size of the md-virtual-repeat-container viewport.

gkalpak: @ThomasBurleson, I think `mdItemSize` is the size (width/height) of each item (in pixels), not the number of visible items.

Regarding the 2 examples, I don't find it super-useful either (I just copied it from the `virtualRepeat` which has the same 2 examples. (TBH, I don't think horizontal scrolling of many elements is all that common, so leaving `mdOrientHorizontal` out of the example wouldn't bother me much :)).

Issue #3101, opened 2015-06-03
md-tabs causes many $digest executions


stevenmiles:
I've been suffering some performance problems with a fairly complex application in FF and IE. I found that my application was continually executing $digest. It seems that the digests are caused by md-tabs.

As the watch expressions are evaluated during a digest, the watch expression created by md-tabs that executes shouldPaginate() is executed continuously. In some profiling I did, shouldPaginate() accounted for the majority of the time in FF and IE. shouldPaginate() seems to be a fairly slow function which is amplified when it is continually being executed.

To measure digests I put a watch on $rootSope and logged out a message when the watch was evaluated. My assumption is the each evaluation of the watch is a $digest execution. Please correct me if my assumptions are wrong.

I measured by monitoring console output in the dev tools of various browsers. I found the following behaviours.

Chrome 43.0.2357.81 m
25ish $digests executions on initialization then 4-5 per tab change

Firefox 38.0.5
25ish $digests executions on initialization then continually executes $digest after first tab change

IE 11
Continually executes $digest after initialization.

OS: Windows 8.1
Angular Material: 0.9.7

http://codepen.io/stevenmiles/pen/yNMPzV

zwirn1981:
Same $digest error in IE 11 when using "md-stretch-tabs" and if there are more tabs then space left, so it should paginate but instead breaks with $digest execution error shouldPaginate().

stevenmiles:
I don't see a "$digest execution error". Also, what ever is causing the $digest executions isn't triggering the digest loop detection inside angularjs. I don't see any exceptions at all. What I'm seeing is just a lot of $digests, and in FF and IE, continuous $digests.

ThomasBurleson:
@robertmesserle - sounds like a priority issue.

stevenmiles:
This fix doesn't seem to have an effect on what I reported. I've' updated the code pen, and double checked that the changes from commit are present. I'm seeing the same behaviour in all browsers as in the first comment. 

http://codepen.io/stevenmiles/pen/LVWgJy

In the codepen, I'm not setting  md-stretch-tabs, so according to the docs this should default to auto.

rrajewski:
In IE 11 it seems that the updateInkBarStyles function is the culprit. Adding/Removing the classes (even if the end result is no change) is triggering the DOMSubtreeModified event and causing the infinite digest loop.

handler:
https://github.com/angular/material/blob/master/src/components/tabs/js/tabsController.js#L59

updateInkBarStyles:
[...]

PR #3405, opened 2015-06-23
Fix for issue #3101

hanthomas:
This change resolved the problem reported in issue #3101 with too many digest executions.


hanthomas:
There was an obvious bug in my code, so I cancelled this pull request.

PR #3443, opened 2015-06-24
Wip/css layouts

### LINKED ISSUE #3436: Nested layouts and flex
Flex seems to be broken on Chrome when used in a nested layout. Firefox renders it as expected.

I'm new to Angular Material so it might be my fault somehow. If this is expected bahaviour what is recommended solution? I have tried a lot of different combinations without any luck,  I would welcome any suggestions on solving this.

Pictures
===
Firefox
---
Firefox

Chrome
---
Chrome
In an inner layout only attribute "flex" without value works.

Code
===
http://codepen.io/monnef/pen/xGPXvB

Relevant snippet:
```html
  <body layout="row" ng-controller="AppCtrl">
    <div flex="30" layout="column" class=a>
      flex="30"
      <md-card flex="40">flex="40"</md-card>
      <md-card flex>flex</md-card>
    </div>
    <div flex layout="column" class=a>
      flex
      <md-card flex="70">flex="70"</md-card>
      <md-card flex="10">flex="10"</md-card>
    </div>
  </body>
```

### LINKED ISSUE #3439: md-menu Firefox 40, Icon and text are not aligned in the dropdown.
In the latest version of angular-material (v0.10), the icon are not aligned on Firefox.

Here is an example:
screenshot from 2015-06-24 14 51 13

ThomasBurleson:
Updates to support nested flex containers with `flex="<percentage value>"`.
Updates to md-list buttons to use inline-block mode.

Fixes #3436. Fixes #3439.

[Review on Reviewable](https://reviewable.io/reviews/angular/material/3443)

### DIFF
--- a/src/core/style/layout.scss
+++ b/src/core/style/layout.scss
@@ -11,11 +11,9 @@
 
 [layout] {
   box-sizing: border-box;
-  display: -webkit-box;
   display: -webkit-flex;
-  display: -moz-box;
   display: -moz-flex;
-  display: -ms-flexbox;
+  display: -ms-flex;
   display: flex;
 }
 
@@ -105,11 +103,10 @@
 @mixin layout-for-name($name) {
   [layout-#{$name}] {
     box-sizing: border-box;
-    display: -webkit-box;
     display: -webkit-flex;
-    display: -moz-box;
     display: -moz-flex;
-    display: -ms-flexbox;
+    display: -moz-flex;
+    display: -ms-flex;
     display: flex;
   }
   [layout-#{$name}=column] {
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
See [Nested Flex Containers](http://codepen.io/team/AngularMaterial/pen/OVOZMw/) demo.
Must run localhost:8080 server with PR code changes.

ThomasBurleson:
@robertmesserle - can you review and validate these changes?

src/core/style/layout.scss
```diff
@@ -153,37 +150,43 @@
...
     }
     [layout="column"] > [#{$flexName}="#{$i * 5}"] {
       max-height: #{$i * 5 + '%'};
+      min-height: #{$i * 5 + '%'};
     }
   }
 
   [#{$flexName}="33"], [#{$flexName}="34"] {
-    flex: 0 0 33.33%;
+    flex: 0.33;
```
gkalpak: This seems suspicious...
Why is it needed ?

ThomasBurleson: Nested layouts do **not** flex properly unless the shorthand, decimal version is used:

See [Nested Flex Containers](http://codepen.io/team/AngularMaterial/pen/OVOZMw/) demo.

gkalpak:
BTW, the _Nested Flex Containers_ demo does already work on Firefox.

From a quick look, it seems to me more like a Chrome bug.
It seems to have difficulties with the caclulation of a flex-items _flex-basis_ as a percentage of its flex-container's height (unless either height is explicitly set and is not a percentage).

--
<sub>It could also be a weirf Flexbox edge-case and Chrome might be doing the _right_ thing, but I have to take a closer look.</sub>

gkalpak:
Chrome might be doing the right thing (or it might be a gray area - or Firefox's being reasonably smart).

Relevant spec sections:

1. [flex-basis][1]
2. [definite/indefinite sizes][2]

--
**Summary:**

* `flex="XYZ" sets `flex: 0 0 XYZ%` => `flex-basis: XYZ%`.
* Because it is a percentage value and it is resolved against the flex container's indefinite block size, it becomes equivalent to `flex-basis: auto`.
* Since there is no `height` value (so it is by default `auto`), it falls back to `flex-basis: content`.
* The flex-basis is determined by the element's content (not our specified XYZ :disappointed:).

<sub>I am not 100% sure I am interpreting the spec correctly, but it kind of makes sense :)</sub>

--
That said, setting a different heght value on the flex-container (e.g. `height: 100%`), makes everything work as expected. ([Updated demo with master version][3])

[1]: http://www.w3.org/TR/css-flexbox-1/#flex-basis-property
[2]: http://www.w3.org/TR/css-flexbox-1/#indefinite
[3]: http://codepen.io/ExpertSystem/pen/WvXgZB?editors=110

src/core/style/layout.scss
gkalpak: It might work in this particular case, but flex 0.33 is quite arbitrary and won't work well with other values.
E.g. imagine a `[flex="33"]` next to a `[flex]`. One would expect that the former takes up 33% of the container and the latter takes up the rest. In practice, though, the former would take up 33/133 (~25%) of the container and the latter would take up 100/133 (~75%).

ThomasBurleson:
@gkalpak - that Demo [with the height correction for class "a" is great. And thx for the research/detailed info. I will close this PR with no change expected to the nest flex box settings; the md-list md-button issue has already been fixed in a separate commit.

gkalpak:
:+1:  (I have a reputation (as critical) to live up to :P)

Issue #3430, opened 2015-06-24
md-tooltip doesn't work on firefox 38.0.5


Nosthertus:
i am trying to make a button on toolbar to show a tooltip whenever the mouse hover on an icon button
```HTML
<md-button class="md-icon-button" ng-href="#/cotizacion">
    <md-tooltip>Cotizacion</md-tooltip>
    <md-icon md-font-icon="fa fa-lg fa-pencil-square-o" aria-label="Cotizacion"></md-icon>
</md-button>
```

but it doesn nothing, even on click, i am using the lastest version `0.10.0`

xforceco:
It works, please try to follow the demo page, and ensure you have properly included angular-material.min.css, angular material icons, and angular-material.min.js

Nosthertus:
the code is the same on the demo, and even the demo doesn't work for me, only if i check "insert dive" and i'm loading all libraries but yet it doesn't work, i am using firefox browser

xforceco:
I see you intend to use font-awesome.
Have you included it properly?

https://material.angularjs.org/latest/#/api/material.components.icon/service/$mdIconProvider

And please provide a plunker for details, as I am sure at least for:

  <md-button class="md-icon-button" ng-href="#/cotizacion">
      <md-tooltip>Cotizacion</md-tooltip>
      <md-icon md-font-icon="fa fa-lg fa-pencil-square-o" aria-label="Cotizacion"></md-icon>
  </md-button><md-button class="md-icon-button" ng-href="#/cotizacion">
      <md-tooltip>Cotizacion</md-tooltip>
      <md-icon></md-icon>
  </md-button>

the tooltip is working in my codepen

Nosthertus:
@xforceco but why it doesnt work even in the demo?

xforceco:
![pr3](https://cloud.githubusercontent.com/assets/12808889/8324203/89e0b798-1a80-11e5-9fd8-3dda342d2c00.png)

which demo page are you referring to?
tooltip works, but I am not sure whether your font-icon works

Nosthertus:
![firefox version](http://i.imgur.com/ApmBNvN.png)

i am still learning angular material and i have no idea how i would properly set the classes on $mdIconProvider

Nosthertus:
@xforceco on the demo, when you hover the button, does the tooltip work?

xforceco:
@Nosthertus Yes, it is a screen capture when I hover on the button, so I don't understand why you said demo not working.

[...]

Issue #3250, opened 2015-06-12
mdTooltip broken in Firefox since 0.9.8


gpothier:
Tooltips do not appear in Firefox (v38). This problem occurs even in the Tooltip demo on the Angular Material page:

https://material.angularjs.org/latest/#/demo/material.components.tooltip

I don't think it is related to #3047, as I checked the pointer-events style in the button's hierarchy in the demo page, and nothing turned up.

alirezamirian:
Actually it is related to #3047 and the issue will be resolved if you comment out the lines 132 and 134 which calls `setVisible(true)` only if `pointer-events` is 'none'

Issue #3047, opened 2015-05-30
mdTooltip is broken in Firefox when checking 'pointer-events' to not be 'none'


alirezamirian:
@ThomasBurleson :
In [this commit](https://github.com/angular/material/commit/667e4c244ccd9a1dfef6894a64b6df0c5e2f6305) you've added a an if statement which checks `pointer-events` of the `md-tooltip` itself to not be `none`! First of all I don't understand why, and secondly
it's not gonna work for obvious reason! In the styles `pointer-events` is set to `none` for `md-tooltip`!
However the computedStyle for `pointer-events` is not `none` in chrome for some reason!

jakubweber:
+1

bennet093:
+1

evertonrobertoauler:
+1

djfishe:
+1

ricardobfa:
+1

krunalsheth3:
This brings another problem. 
I added a css property to one of the parent div tags 
pointer-events : none
This shows up every tooltip present in its child elements 
Is there a work around for that ?

PR #3447, opened 2015-06-25
Fixing md-tooltip on Firefox issues #3047, #3250 and #3430

evertonrobertoauler:
var computedStyles = $window.getComputedStyle(element[0]);

console.log(computedStyles['pointer-events'] ); // Firefox returns 'none' as string
console.log(computedStyles['pointer-events'] ); // Chrome returns an empty string ''

The problem is that hasComputedStyleValue('pointer-events','none') always returns true on Firefox and setVisible(true); is not being called.

All the tests are still passing and now the Tooltip works fine on Chrome and Firefox again.

[Review on Reviewable](https://reviewable.io/reviews/angular/material/3447)

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -129,7 +129,7 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
     function bindEvents () {
       var mouseActive = false;
       var enterHandler = function() {
-        if (!hasComputedStyleValue('pointer-events','none')) {
+        if (!hasComputedStyleValue('pointer-events','block')) {
           setVisible(true);
         }
       };


ThomasBurleson:
@evertonrobertoauler - I cannot merge this without a signed CLA.

src/components/tooltip/tooltip.js
```diff
@@ -129,7 +129,7 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
     function bindEvents () {
       var mouseActive = false;
       var enterHandler = function() {
-        if (!hasComputedStyleValue('pointer-events','none')) {
```
ThomasBurleson: This change does not make sense. Valid values for the `pointer-events` style are:

*  auto
*  inherit
*  none

ThomasBurleson:
Closed with 6fc9212cb96cd3542015eea57fad20a1b6cf24af

PR #3457, opened 2015-06-25
fix($mdUtil): disableBodyScroll() no longer forces scrollbar

michaelbromley:
The disableScrollAround() as used by mdSelect and mdAutocomplete
would force a scrollbar by setting `overflow-y: scroll` on `<body>`.
This change sets it to `auto` so the browser can decide if a scrollbar
is needed.

[Review on Reviewable](https://reviewable.io/reviews/angular/material/3457)

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -116,7 +116,7 @@ angular.module('material.core')
           applyStyles(body, {
             position: 'fixed',
             width: '100%',
-            overflowY: 'scroll',
+            overflowY: 'auto',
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@rschmukler - does this change affect select and scroll features?

MarcosPortugal:
Please merge this!

rschmukler:
This was fixed in master (although there may be a new bug with IE 11 which we will check out). Closing PR.

Abhishek901:
how can i over come with this issu

PR #3461, opened 2015-06-25
feat(virtualRepeat): allow for scrolling areas larger than the browser's

jelbourn:
maximum element size.

@kseamon

### DIFF
--- a/src/components/virtualRepeat/virtualRepeater.js
+++ b/src/components/virtualRepeat/virtualRepeater.js
@@ -48,7 +48,7 @@ function VirtualRepeatContainerDirective() {
 }
 
 
-function virtualRepeatContainerTemplate($element, $attrs) {
+function virtualRepeatContainerTemplate($element) {
   return '<div class="md-virtual-repeat-scroller">' +
     '<div class="md-virtual-repeat-sizer"></div>' +
     '<div class="md-virtual-repeat-offsetter">' +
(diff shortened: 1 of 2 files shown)


src/components/virtualRepeat/virtualRepeater.js
```diff
@@ -56,6 +56,14 @@ function virtualRepeatContainerTemplate($element, $attrs) {
...
 }
 
+/**
+ * Maximum size, in pixels, that can be explicitly set to an element. The actual value varies
+ * between browsers, but IE11 has the very lowest size at a mere 1,533,917px. Ideally we could
+ * *compute* this value, but Firefox always reports an element to have a size of zero if it
+ * goes over the max, meaning that we'd have to binary search for the value.
+ * @const {number}
+ */
+var MAX_ELEMENT_SIZE = 1533917;
```
kseamon: Can we put this on a non-global variable? Maybe put it on one of the constructors.

src/components/virtualRepeat/virtualRepeater.js
```diff
@@ -131,11 +139,40 @@ VirtualRepeatContainerController.prototype.getScrollSize = function() {
...
+      // Divide the total size we have to render into N max-size pieces.
+      var numChildren = Math.floor(size / MAX_ELEMENT_SIZE);
+
+      // Element template to clone for each max-size piece.
+      var sizerChild = document.createElement('div');
+      sizerChild.style[dimension] = MAX_ELEMENT_SIZE + 'px';
+      sizerChild.style[crossDimension] = '1px';
+
+      for (var i = 0; i < numChildren; i++) {
+        this.sizer.appendChild(sizerChild.cloneNode(false));
```
kseamon: Probably doesn't matter in practice, but a dom fragment could be used here.

src/components/virtualRepeat/virtualRepeater.js
jelbourn: It's not global ; all of the files get wrapped by an immediately invoked function
(otherwise VirtualRepeatController etc. would all be global vars)

src/components/virtualRepeat/virtualRepeater.js
jelbourn: I considered it, but the modern browsers shouldn't care since we're not reading any layout properties while we're appending.

jelbourn: I considered it, but modern browsers shouldn't care since we're not reading any layout properties while we're appending.

ThomasBurleson:
Question: what is the driving use case, issue or missing feature for this enhancement?

jelbourn:
The datepicker's calendar needs this. Without this technique, we'd be more limited for how many months can be viewed (my initial setting of 1,000 years in either direction was too big for IE).

gkohen:
@ThomasBurleson , thanks for bringing virtual scroll to the project its much needed as sometimes we're dealing with large lists that, even with one-time-binding, are pretty slow since every object is rendered to a DOM element. It's be great if you can come up with a blog post about the feature and how it solve some of today's pain points.

PR #3467, opened 2015-06-26
fix(mdTooltip): fix regression that broke tooltips on Firefox

### LINKED ISSUE #3047: mdTooltip is broken in Firefox when checking 'pointer-events' to not be 'none'

Splaktar:
Remove check for pointer-events not equal to none to resolve non-visible tooltips on Firefox.
Add Chrome & Firefox as default test runners in karma task to avoid regressions like this in the future.

Closes #3047, #3250 and #3430

[Review on Reviewable](https://reviewable.io/reviews/angular/material/3467)

### DIFF
--- a/gulp/tasks/karma-fast.js
+++ b/gulp/tasks/karma-fast.js
@@ -4,7 +4,7 @@ var util = require('../util');
 var ROOT = require('../const').ROOT;
 var args = util.args;
 
-// NOTE: `karma-fas` does NOT pre-make a full build of JS and CSS
+// NOTE: `karma-fast` does NOT pre-make a full build of JS and CSS
 // exports.dependencies = ['build'];
 
 exports.task = function (done) {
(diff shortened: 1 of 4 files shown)


Splaktar:
I would be happy to hear if this code may have existed for a reason. If so, we'll want to add a test for whatever that reason is.

We probably don't want to be missing regressions like this on releases since they would be easily caught by the automated tests had they been run for Firefox instead of just PhantomJS. It only adds 3s to the task and there is a karma-fast task for faster development runs.

Splaktar:
This attempts to improve on the proposed PR #3447

PR #3521, opened 2015-06-30
update(tabs): cleans up tabs code to remove unnecessary watchers

robertmesserle:
Wanted to get feedback from @jelbourn and @gkalpak on this since they are typically the most critical in PR reviews. =p

[Review on Reviewable](https://reviewable.io/reviews/angular/material/3521)

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -6,7 +6,7 @@ angular
  * @ngInject
  */
 function MdTabsController ($scope, $element, $window, $timeout, $mdConstant, $mdTabInkRipple,
-                           $mdUtil, $animate, $attrs, $compile, $mdTheming) {
+                           $mdUtil, $animate, $attrs, $compile, $mdTheming, $parse) {
   //-- define private properties
   var ctrl       = this,
       locked     = false,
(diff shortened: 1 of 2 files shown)


src/components/tabs/js/tabsController.js
```diff
@@ -55,41 +66,70 @@ function MdTabsController ($scope, $element, $window, $timeout, $mdConstant, $md
...
-    });
-    $scope.$on('$destroy', cleanup);
+  /**
+   * Defines a two-way data-binding without requiring an isolated scope.
+   * @param key
+   * @param attr
+   * @param defaultValue
+   * @param handler
+   */
+  function defineBinding (key, attr, defaultValue, handler) {
```
ThomasBurleson: let's move these to $mdUtils.

src/components/tabs/js/tabsController.js
```diff
@@ -55,41 +66,70 @@ function MdTabsController ($scope, $element, $window, $timeout, $mdConstant, $md
...
+  }
+
+  /**
+   * Defines boolean attributes with default value set to true.  (ie. md-stretch-tabs with no value
+   * will be treated as being truthy)
+   * @param key
+   * @param handler
+   */
+  function defineBooleanAttribute (key, handler) {
+    var attr = 'md' + key.charAt(0).toUpperCase() + key.substr(1);
```
ThomasBurleson: why not use `$normalize()` ?

robertmesserle: This is actually generating the attribute with the `md` prefix, not just normalizing.

src/components/tabs/js/tabsController.js
robertmesserle: Agreed.

robertmesserle:
Review status: 0 of 2 files reviewed at latest revision, 2 unresolved discussions, some commit checks pending.

---

<sup>**[src/components/tabs/js/tabsController.js, line 130 \[r1\]](https://reviewable.io:443/reviews/angular/material/3521#-Jt4UHC1Oc5CfgL2R5Mh-r1-130)** ([raw file](https://github.com/angular/material/blob/5087b7b6699843dfc899cd61d9d7d0e2aa4f60f5/src/components/tabs/js/tabsController.js#L130)):</sup>
Done.

---


Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/angular/material/3521)

robertmesserle:
Reviewed 1 of 2 files at r1, 1 of 1 files at r2.
Review status: all files reviewed at latest revision, 2 unresolved discussions, all commit checks successful.

---



Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/angular/material/3521)

gkalpak:
Rushed it in to avoid my review; good call :stuck_out_tongue:

I am curious, why remove the isolate scope. Was it part of the problem ? It seemed quite suitable imo.

robertmesserle:
@gkalpak The original goal wasn't to remove the isolated scope.  While addressing other performance issues, I was cleaning up some of the code, including boolean attributes.  After that, I realized that there was only one property left on the isolated scope, so I wanted to remove that as a sort of experiment.

Earlier, we had run into issues because of the isolated scope, which we had worked through; however, we may be able to clean up some of those workarounds now that there is no isolated scope.

(If there proves to be no benefit, I'll most-likely just go back to the old setup.)

PR #3534, opened 2015-07-01
feat(virtualRepeat): Make md-item-size optional

kseamon:
When md-item-size is absent, the repeater will attempt to measure the size (width or height depending on md-orient-horizontal) of the repeated element. It still assumes that all such elements will have the same size.

### DIFF
--- a/src/components/virtualRepeat/virtualRepeater.js
+++ b/src/components/virtualRepeat/virtualRepeater.js
@@ -211,6 +211,8 @@ VirtualRepeatContainerController.prototype.handleScroll_ = function() {
   if (offset === this.scrollOffset) return;
 
   var itemSize = this.repeater.getItemSize();
+  if (!itemSize) return;
+
   var numItems = Math.max(0, Math.floor(offset / itemSize) - NUM_EXTRA);
 
   var transform = this.isHorizontal() ? 'translateX(' : 'translateY(';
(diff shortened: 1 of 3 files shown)


src/components/virtualRepeat/virtualRepeater.js
```diff
@@ -298,8 +303,8 @@ function VirtualRepeatController($scope, $element, $attrs, $browser, $document)
   this.endIndex = 0;
   // TODO: measure width/height of first element from dom if not provided.
   // getComputedStyle?
-  /** @type {number} Height/width of repeated elements. */
-  this.itemSize = $scope.$eval($attrs.mdItemSize);
+  /** @type {?number} Height/width of repeated elements. */
+  this.itemSize = $scope.$eval($attrs.mdItemSize) || null;
```
jelbourn: Maybe use zero as the no-value marker?

src/components/virtualRepeat/virtualRepeater.js
```diff
@@ -352,11 +357,33 @@ VirtualRepeatController.prototype.link_ =
  * changed.
  */
 VirtualRepeatController.prototype.containerUpdated = function() {
+  // If itemSize is unknown, attempt to measure it.
+  if (!this.itemSize) {
+    this.$$rAF(angular.bind(this, function() {
```
jelbourn: Break the inside of this `if` into a separate function?

jelbourn:
LGTM

src/components/virtualRepeat/virtualRepeater.js
kseamon: That would work too. Null feels more...not valid indicating to me.

src/components/virtualRepeat/virtualRepeater.js
kseamon: done

PR #3542, opened 2015-07-01
fix(tabs): add check tab before update its order

pan-long:


EladBezalel:
@pan-long please make sure it's still a valid pr on HEAD,
closing for now

PR #3544, opened 2015-07-01
Adds support for textarea's "rows" attribute and makes autogrow optional

### LINKED ISSUE #1209: md-input-container: growTextarea should be optional and should be possible to set max height
1. `md-input-container` needs parameter that disables autogrow of `<textarea>`.
2. There should be possible to set maximum height of `<textarea>` inside `md-input-container` if autogrow is enabled. If text height inside is bigger than `<textarea>` height, overflow scrollbar should appear.

### LINKED ISSUE #2006: Bug: md-input-container textarea max-height/rows limitation problem
Is there a way to limit a size of textarea with help "rows" attribute or similar?

```
<md-input-container>
  <label>Text</label>
  <textarea ng-model="feedback" rows="3"></textarea>
</md-input-container>
```

### LINKED ISSUE #2007: Bug: md-input-container textarea bug if inside of block with overflow: scroll

nantunes:
Replaces corrupted PR #2899.

Makes textarea autogrow optional, partly resolves #1209
Fixes textarea auto grow inside scrollable container (resolves #2007 and possibly others)
Adds support for textarea auto grow based on rows attribute (resolves #2006)
Avoid "rows" attribute unintended overwrite by default value

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -110,6 +110,7 @@ function labelDirective() {
  * The purpose of **`md-maxlength`** is exactly to show the max length counter text. If you don't want the counter text and only need "plain" validation, you can use the "simple" `ng-maxlength` or maxlength attributes.
  * @param {string=} aria-label Aria-label is required when no label is present.  A warning message will be logged in the console if not present.
  * @param {string=} placeholder An alternative approach to using aria-label when the label is not present.  The placeholder text is copied to the aria-label attribute.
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@nantunes - I will review and merge this. Can you add another PR that provides unit tests for these changes? Thx.

PR #3558, opened 2015-07-02
fix(toast): replace cancel with hide on hideDelay

cebor:
This replaces the `cancel()` with `hide()` in a toast with the option `hideDelay`.

What it changes is, that the promise get resolved instead of rejected what makes more sense to me.

### DIFF
--- a/src/core/services/interimElement/interimElement.js
+++ b/src/core/services/interimElement/interimElement.js
@@ -339,7 +339,7 @@ function InterimElementProvider() {
           show: function() {
             var compilePromise;
             if (options.skipCompile) {
-              compilePromise = $q(function(resolve) { 
+              compilePromise = $q(function(resolve) {
                 resolve({
                   locals: {},
                   link: function() { return options.element; }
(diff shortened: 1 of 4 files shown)


ThomasBurleson:
@cebor - plz update and dependent *.spec.js unit tests so all tests pass.

cebor:
@ThomasBurleson done!

src/core/services/interimElement/interimElement.js
```diff
@@ -386,7 +386,7 @@ function InterimElementProvider() {
 
               function startHideTimeout() {
                 if (options.hideDelay) {
-                  hideTimeout = $timeout(service.cancel, options.hideDelay) ;
```
ThomasBurleson: @rschmukler - this changes seems reasonable: hide === close. 
What is the difference (wrt `interimElement` between `cancel()` and `hide()`).

btw, a promise wrapper is also needed in hide() change is also needed:

```js
function hide(response) {
  var interimElement = stack.shift();
  return $q.when(interimElement && interimElement.remove().then(function() {
    interimElement.deferred.resolve(response);
  }));
}
```

Issue #3233, opened 2015-06-11
md-select converts numbers to strings


daniel-nagy:
The `md-select` directive will convert a `Number` to a `String`. This may result in unexpected `$watch` events triggering. For example,

```javascript
$scope.number = 10;

$scope.$watch('number', function (newValue, oldeValue) {
    // newValue = "10", oldValue = 10    
    if(newValue === oldeValue) {
        return;
    }
    ...
}
```

[codepen](http://codepen.io/anon/pen/xGrWOE)

I suggest adding a `type` attribute to the `md-select` directive.

Schweigi:
A proper support for 'ng-options' would probably solve this issue . There is an open ticket: https://github.com/angular/material/issues/1837

chinnichaitanya:
Yes even I've faced this issue. Took some time to figure it out :smile:

uschmann:
I've faved the same issue. I get douplicated $watch triggers when ever the Selektion changes.

skosno:
@daniel-nagy looking at your example - can you just use ng-value instead of value? Please notice that ng-change is not triggered at first load as expected (no change happened yet).

http://codepen.io/anon/pen/gavMoe

@ThomasBurleson initially created fix for this particular issue targeted for numbers with https://github.com/angular/material/commit/c9f2b9f9a96010c667bbe0e2937d54ed99e3057f but it was reverted with recommendation to use "ng-value" with https://github.com/angular/material/commit/b983c0d4575f6701cb327f29c75354dc31447c74 by @rschmukler 

I believe unless "ng-value" is not an option for you guys, this issue can be closed?

daniel-nagy:
@skosno I've been using `ng-value` without any issues. I believe this issue can be closed.

PR #3560, opened 2015-07-02
mdSelect: Attributes are always strings.

daniel-nagy:
Attributes are always strings, this results in unexpected `ng-change` or `$scope.watch` events firing when using numbers.

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -644,7 +644,7 @@ function OptionDirective($mdButtonInkRipple, $mdUtil) {
     if (angular.isDefined(attr.ngValue)) {
       scope.$watch(attr.ngValue, setOptionValue);
     } else if (angular.isDefined(attr.value)) {
-      setOptionValue(attr.value);
+      setOptionValue(isNaN(attr.value) ? attr.value : Number(attr.value));
     } else {
       scope.$watch(function() { return element.text(); }, setOptionValue);
     }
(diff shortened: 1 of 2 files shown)


daniel-nagy:
I signed the CLA! #3233

ThomasBurleson:
@daniel-nagy - please provide a unit test that validates these changes.

daniel-nagy:
I've put more though into this and the same issue will be experienced for any value that is not a string. For example, true will be converted to "true". The following line,

```javascript
setOptionValue(attr.value);
```

should of been 

```javascript
setOptionValue(scope.value);
```

however this will break some unit tests. Further more, there probably should of been an isolated scope on `<md-option>` directive with `scope: { value: '=' }` but this breaks aaalllllllot of unit tests.

I see two solutions, either enforce the use of `ng-value` because the watch listener will take care of type conversion, or optionally specify a `type` attribute.

One more thing to note is, while using `scope.value` breaks 5 or 6 unit test, using `scope.$eval(attr.value)` only breaks the following unit test,

```javascript
it('errors for duplicate md-options, non-dynamic value', inject(function($rootScope) {
    expect(function() {
      setup('ng-model="$root.model"', '<md-option value="a">Hello</md-option>' +
            '<md-option value="a">Goodbye</md-option>');
    }).toThrow();
  }));
```

which can be passed by explicitly using a string, `value="\'a\'"` or defining `$rootScope.a` and then `value="{{a}}"`. In conclusion, I believe more thought needs to go into this.

PR #3590, opened 2015-07-05
Check for possible step calculation

fabiocarneiro:
Currently, if you set md-discrete and step 0 in sliders, it will still try to calculate the numSteps, causing the entire application to fail.


fabiocarneiro:
@ThomasBurleson Thanks for merging it.

Just to clarify, I was using early return and yoda conditions, which usually leads to a cleaner code and avoids "else" statements. You inverted the logic.

I understand you're using log instead of error, but it still could be done by keeping the same (better) code structure.

```js
if(0 >= step) {
    $log.error('Slider step value must be greater than zero when in discrete mode');
    return;
}
```

PR #3612, opened 2015-07-06
md-tabs: Param md-no-tabs-if-single to hide md-tabs-wrapper when only one tab exists

nantunes:


EladBezalel:
What is this trying to solve? please open and issue, and rebase this. until then it seems irrelevant, closing.

nantunes:
Hi @EladBezalel,

Yes, I'm sorry for the empty description.

But what it does is on the title: adds a parameter to md-tabs that hides the tab bar when only one tab exists. It's a common pattern, present for instance in the major browsers.

The commit contains tests and demo updates. It is more than 8 months old and was labeled as targeting some milestones.

I'm currently not working in any project using angular/material. I can rebase it if it is minimally interesting to anyone. If not, let it go.

PR #3648, opened 2015-07-09
bug(datepicker): Fix datepicker disabled state

mhchen:
Disable hover state and functionality of datepicker opener button when
the element is disabled.


mhchen:
@jelbourn please review at your convenience

jelbourn:
I actually took care of this already in my local copy, sorry! I just made the down-triangle button hidden when the picker is disabled.

PR #3662, opened 2015-07-09
feat(core) create mdResolve service

ivnivnch:
The $mdResolve service helps to resolve acyclic dependencies (ui.router like resolve).
Used internally in $mdCompiler service.

Now you can do this:

```js
$mdCompiler.compile({
  locals: {
    value: 'value'
  },
  resolve: {
    asyncService: function(value, $q) {
      return $q.when(value);
    },
    service: function(value, asyncService ) {
     return value == asyncService;
    }
  },
  ...
});
```

### DIFF
--- a/src/core/services/compiler/compiler.js
+++ b/src/core/services/compiler/compiler.js
@@ -69,7 +69,7 @@ function mdCompilerService($q, $http, $compile, $controller, $templateCache, $md
     var controller = options.controller;
     var controllerAs = options.controllerAs;
     var resolve = angular.copy(options.resolve || {});
-    var locals = angular.copy(options.locals || {});
+    var locals = options.locals;
     var transformTemplate = options.transformTemplate || angular.identity;
     var bindToController = options.bindToController;
 
--- a/src/core/services/compiler/compiler.js
+++ b/src/core/services/compiler/compiler.js
@@ -1,7 +1,7 @@
 angular.module('material.core')
   .service('$mdCompiler', mdCompilerService);
 
-function mdCompilerService($q, $http, $injector, $compile, $controller, $templateCache) {
+function mdCompilerService($q, $http, $compile, $controller, $templateCache, $mdResolve) {
   /* jshint validthis: true */
 
   /*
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
@jelbourn - want to review this with me ?

ThomasBurleson:
References https://github.com/angular/material/commit/bee05239014e5eb8a0ae1b9e842f32c9caa9474b#commitcomment-12119240

ivnivnch:
Changed the behavior of $mdResolve:
* Promises can be passed to the controller via locals.
* Locals passed by reference.
```javascript
var value = 'value';
var promise = $q.when('promise');
$mdCompiler.compile({
  locals: {
    localValue: value,
    localPromise: promise
  },
  resolve: {
    first: $q.when('first'),
    second: function($q, localValue, localPromise) {
      return $q.when(localValue == value &&
        localPromise == promise && isPromise(localPromise) &&
        'second');
    },
    third: function(first, second) {
      return !isPromise(first) && first == 'first' &&
        !isPromise(second) && second == 'second' &&
        'third';
    }
  },
  controller: function(localValue, localPromise, first, second, third) {
    assert(localValue == value);
    assert(localPromise == promise && isPromise(localPromise));
    assert(!isPromise(first) && first == 'first');
    assert(!isPromise(second) && second == 'second');
    assert(third == 'third');
  }
});
```

ThomasBurleson:
@ivnivnch - what specific problem do these changes solve for you? Can you provide a demo/plunkr ?

ivnivnch:
[$mdResolve example](http://plnkr.co/edit/SgDdaT8mqMjxhc74iBHN?p=preview). Pay attention to the MainController.search method. Without $mdResolve, I get an [error](http://plnkr.co/edit/wZyepho0e6RHw5vASlaI?p=preview), or there is a a [bad solution](http://plnkr.co/edit/SmNDwaxe8n0mYWZC6iEK?p=preview).

ThomasBurleson:
@ivnivnch - Your two (2) demos were very helpful. Thank you.

I would not consider promise chaining and *temp* closure variable a **bad** solution; in fact, I think it is a **classical** solution. Nevertheless, it is not ideal **AND** it would be nice to support both the **resolve** and the **classical** approaches.

gkalpak:
I, too, don't get what is so bad about the _"bad solution"_. (It can get a little cleaner, but no big deal: [Updated plnkr][1]).

In any case, having resolvable objects that depend on other resolvable objects is quite of an edge-case imo and although nice-to-have, I am not sure if it's worth the compexity (more complex code, more edge-cases to get right, less maintainable, less easy to reason about).

I am not saying I am against it - I'm just not convinced (yet) :smile: 

[1]: http://plnkr.co/edit/3WM14gy24rkpfNk4AP1h?p=preview

ivnivnch:
@ThomasBurleson Yes, perhaps it would be better to call the good, the bad and the ugly :smile:. This pull request not break backwards compatibility and supports classical approaches, as You can see in the tests. Also the same approach is used in the ui.router, and developers using ui.router will not be surprised. On the contrary, I was surprised that this feature is not working.

ivnivnch:
@gkalpak it works, but it looks too confusing.

EladBezalel:
@ivnivnch can you please rebase?

ivnivnch:
I already deleted my fork.

PR #3666, opened 2015-07-09
Early preview datepicker PR - DO NOT MERGE

jelbourn:
This is a PR to give people a change to review the code for `md-datepicker` as it nears a releasable point.

There is still some work remaining, as can be seen in the `TODO`s in the code.


src/components/calendar/calendar.js
```diff
@@ -0,0 +1,513 @@
...
+     * Handles a click event on a date cell.
+     * Created here so that every cell can use the same function instance.
+     * @this {HTMLTableCellElement} The cell that was clicked.
+     */
+    this.cellClickHandler = function() {
+      if (this.hasAttribute('data-timestamp')) {
+        $scope.$apply(function() {
+          var timestamp = Number(this.getAttribute('data-timestamp'));
+          self.setNgModelValue(self.dateUtil.createDateAtMidnight(timestamp));
+        }.bind(this)); // The `this` here is the cell element.
```
robertmesserle: Since you have the `self` reference up above, do you need to bind this?

As an alternative, would it make sense to define this on the prototype, then overwrite it here with `this.cellClickHandler = this.cellClickHandler.bind(this)`?

ThomasBurleson:
@Jeremy - PhantomJS will fail tests and code that uses `Function.bind()`.

robertmesserle:
@ThomasBurleson We use `Function.prototype.bind` quite a few times in Angular Material.

gkalpak:
@robertmesserle, I don't think you do. I could only find one instance of `Function.prototype.bind` (in [checkbox.js][1]. But it could be that that particular path is not covered in tests (didn't actually looked into it). But in all other cases, `angular.bind()` is used.

<sub>(PhantomJ will eventually support `bind` (or so I've been hearing for so long) :disappointed:)</sub>

[1]: https://github.com/angular/material/blob/249631748be39566e3eeebb5dac24c853fd4188e/src/components/checkbox/checkbox.js#L82

jelbourn:
@ThomasBurleson I know. I was hoping PhantomJS 2.0 would be available by the time I got this far. But it looks like they're not making any progress on their Mac / Linux binaries. I've already changed the `bind` calls after I sent this PR.

robertmesserle:
@gkalpak Ah, you're right - most of them are using `angular.bind`.  I saw the `checkbox` example at the top of the list and didn't look at the rest of the results too closely.

jelbourn:
Updated with latest version.

If the only comment here is about that one `.bind` call, I must be doing pretty well.

src/components/datepicker/calendar.js
```diff
@@ -0,0 +1,485 @@
...
+      }
+
+      // Remaining key events fall into two categories: selection and navigation.
+      // Start by checking if this is a selection event.
+      if (event.which === self.keyCode.ENTER) {
+        self.setNgModelValue(self.displayDate);
+        event.preventDefault();
+        return;
+      }
+
```
robertmesserle: Why not:
```
switch (event.which) {
  case self.keyCode.ESCAPE:
  case self.keyCode.TAB:
    //...
  case self.keyCode.ENTER:
    //...
```

src/components/datepicker/calendar.js
```diff
@@ -0,0 +1,485 @@
...
+  CalendarCtrl.prototype.changeSelectedDate = function(date) {
+    var self = this;
+    var previousSelectedDate = this.selectedDate;
+    this.selectedDate = date;
+    this.changeDisplayDate(date).then(function() {
+
+      // Remove the selected class from the previously selected date, if any.
+      if (previousSelectedDate) {
+        var prevDateCell =
+            self.calendarElement.querySelector('#' + self.getDateId(previousSelectedDate));
```
robertmesserle: Why not `getElementById`?

src/components/datepicker/calendar.js
```diff
@@ -0,0 +1,485 @@
...
+        if (prevDateCell) {
```
robertmesserle: Would `document.getElementById()` make sense here?  Or is it possible for the ID to live outside of calendarElement?

src/components/datepicker/calendarMonth.js
```diff
@@ -0,0 +1,239 @@
...
+        var monthCtrl = controllers[1];
+
+        monthCtrl.calendarCtrl = calendarCtrl;
+        monthCtrl.generateContent();
+
+        // The virtual-repeat re-uses the same DOM elements, so there are only a limited number
+        // of repeated items that are linked, and then those elements have their bindings updataed.
+        // Since the months are not generated by bindings, we simply regenerate the entire thing
+        // when the binding (offset) changes.
+        scope.$watch(function() { return monthCtrl.offset }, function(offset, oldOffset) {
```
robertmesserle: Not sure if it's really better, but you could do: `scope.$watch('mdMonthCtrl.offset', ...)`

src/components/datepicker/datePicker.js
```diff
@@ -0,0 +1,361 @@
...
+  /** Attach event listeners for user interaction. */
+  DatePickerCtrl.prototype.attachInterationListeners = function() {
+    var self = this;
+    var $scope = this.$scope;
+    var keyCodes = this.$mdConstant.KEY_CODE;
+
+    // Add event listener through angular so that we can triggerHandler in unit tests.
+    angular.element(self.inputElement).on('keydown', function(event) {
+      $scope.$apply(function() {
+        if (event.altKey && event.keyCode == keyCodes.DOWN_ARROW) {
```
robertmesserle: Should `$apply` live inside the if statement?  Or do we want to run it regardless?

src/components/datepicker/datePicker.js
```diff
@@ -0,0 +1,361 @@
...
+   */
+  DatePickerCtrl.prototype.installPropertyInterceptors = function() {
+    var self = this;
+
+    // Intercept disabled on the date-picker element to disable the internal input.
+    // This avoids two bindings (outer scope to ctrl, ctrl to input).
+    Object.defineProperty(this.$element[0], 'disabled', {
+      get: function() { return self.isDisabled; },
+      set: function(value) { self.setDisabled(value) }
+    });
```
robertmesserle: Nice.

src/components/datepicker/datePicker.js
```diff
@@ -0,0 +1,361 @@
...
+    return angular.element(this.calendarPane.querySelector('md-calendar')).controller('mdCalendar');
+  };
+
+  /** Focus the calendar in the floating pane. */
+  DatePickerCtrl.prototype.focusCalendar = function() {
+    // Use a timeout in order to allow the calendar to be rendered, as it is gated behind an ng-if.
+    var self = this;
+    this.$timeout(function() {
+      self.getCalendarCtrl().focus();
+    }, 0, false);
```
robertmesserle: `$mdUtil.nextTick(function () { /* ... */ }, false);`

src/components/datepicker/calendar.js
```diff
@@ -0,0 +1,485 @@
...
+
+  /**
+   * @ngdoc module
+   * @name material.components.datepicker
+   * @description Datepicker
+   */
+  angular.module('material.components.datepicker', [
+    'material.core',
+    'material.components.virtualRepeat',
+    'material.components.icon'
```
kseamon: Alphabetize?

src/components/datepicker/calendar.js
```diff
@@ -0,0 +1,485 @@
...
+    return {
+      template:
+          '<table aria-hidden="true" class="md-calendar-day-header"><thead></thead></table>' +
+          '<div class="md-calendar-scroll-mask">' +
+            '<md-virtual-repeat-container class="md-calendar-scroll-container">' +
+              '<table role="grid" tabindex="0" class="md-calendar" aria-readonly="true">' +
+                '<tbody role="rowgroup" md-virtual-repeat="i in ctrl.items" md-calendar-month ' +
+                    'md-month-offset="$index" class="md-calendar-month" ' +
+                    'md-start-index="ctrl.getSelectedMonthIndex()" ' +
+                    'md-item-size="' + TBODY_HEIGHT + '"></tbody>' +
```
kseamon: You can optionally omit md-item-size now.

src/components/datepicker/calendar.js
```diff
@@ -0,0 +1,485 @@
...
+   * Hides the vertical scrollbar on the calendar scroller by setting the width on the
+   * calendar scroller and the `overflow: hidden` wrapper around the scroller, and then setting
+   * a padding-right on the scroller equal to the width of the browser's scrollbar.
+   *
+   * This will cause a reflow.
+   */
+  CalendarCtrl.prototype.hideVerticalScrollbar = function() {
+    var element = this.$element[0];
+
+    var scrollMask = element.querySelector('.md-calendar-scroll-mask');
```
kseamon: Minor, but getElementsByClassName is probably a bit faster.

src/components/datepicker/calendar.js
```diff
@@ -0,0 +1,485 @@
...
+      case keyCode.RIGHT_ARROW: return dateUtil.incrementDays(this.displayDate, 1);
+      case keyCode.LEFT_ARROW: return dateUtil.incrementDays(this.displayDate, -1);
+      case keyCode.DOWN_ARROW:
+        return event.metaKey ?
+          dateUtil.incrementMonths(this.displayDate, 1) :
+          dateUtil.incrementDays(this.displayDate, 7);
+      case keyCode.UP_ARROW:
+        return event.metaKey ?
+          dateUtil.incrementMonths(this.displayDate, -1) :
+          dateUtil.incrementDays(this.displayDate, -7);
```
kseamon: Any thoughts on cmd+left and right for mac users (They generally have the same effect as home/end on windows)

src/components/datepicker/calendar.js
```diff
@@ -0,0 +1,485 @@
...
+  CalendarCtrl.prototype.focus = function(opt_date) {
+    var date = opt_date || this.selectedDate;
+
+    var previousFocus = this.calendarElement.querySelector('.md-focus');
+    if (previousFocus) {
+      previousFocus.classList.remove('md-focus');
+    }
+
+    var cellId = this.getDateId(date);
+    var cell = this.calendarElement.querySelector('#' + cellId);
```
kseamon: similarly, getElementById

src/components/datepicker/calendar.js
```diff
@@ -0,0 +1,485 @@
...
+
+    // If trying to show an invalid date, do nothing.
+    if (!this.dateUtil.isValidDate(date)) {
+      return this.$q.when();
+    }
+
+    // Do nothing if animation is in progress.
+    if (this.isMonthTransitionInProgress) {
+      return this.$q.when();
+    }
```
kseamon: Use an || with the above block to be a little less repetitive?

src/components/datepicker/calendar.scss
```diff
@@ -0,0 +1,141 @@
...
+@mixin md-calendar-cell() {
+  height: $md-calendar-cell-size;
+  width: $md-calendar-cell-size;
+
+  text-align: center;
+
+  // Remove all padding and borders so we can completely
+  // control the size of the table cells.
+  padding: 0;
+  border: none;
```
kseamon: Alphabetize?

src/components/datepicker/calendar.scss
```diff
@@ -0,0 +1,141 @@
...
+  &:last-child {
+    padding-right: $md-calendar-side-padding;
+  }
+}
+
+// Styles for tables used in mdCalendar (the day-of-the-week header and the table of dates itself).
+@mixin md-calendar-table() {
+  // Fixed table layout makes IE faster.
+  // https://msdn.microsoft.com/en-us/library/ms533020(VS.85).aspx
+  table-layout: fixed;
```
kseamon: Alphabetize

src/components/datepicker/calendar.scss
```diff
@@ -0,0 +1,141 @@
...
+  display: inline-block;
+  overflow: hidden;
+  height: $md-calendar-weeks-to-show * $md-calendar-cell-size;
+
+  // The actual scrolling element.
+  .md-virtual-repeat-scroller {
+    // These two properties are needed to get touch momentum to work.
+    // See https://css-tricks.com/snippets/css/momentum-scrolling-on-ios-overflow-elements
+    overflow-y: scroll;
+    -webkit-overflow-scrolling: touch;
```
kseamon: We should have a broader discussion of whether this should be the default for the repeater. It would mean an always-visible scrollbar (which I know isn't relevant here).

src/components/datepicker/calendarMonth.js
```diff
@@ -0,0 +1,239 @@
...
+   * Creates a single cell to contain a date in the calendar with all appropriate
+   * attributes and classes added. If a date is given, the cell content will be set
+   * based on the date.
+   * @param {Date=} opt_date
+   * @returns {HTMLElement}
+   */
+  CalendarMonthCtrl.prototype.buildDateCell = function(opt_date) {
+    var calendarCtrl = this.calendarCtrl;
+
+    // TODO(jelbourn): cloneNode is likely a faster way of doing this.
```
kseamon: +1

src/components/datepicker/datePicker.js
```diff
@@ -0,0 +1,361 @@
...
+    this.$$rAF = $$rAF;
+
+    /** @type {!angular.NgModelController} */
+    this.ngModelCtrl = null;
+
+    /** @type {HTMLInputElement} */
+    this.inputElement = $element[0].querySelector('input');
+
+    /** @type {HTMLElement} */
+    this.inputContainer = $element[0].querySelector('.md-datepicker-input-container');
```
kseamon: getElementsByClassName?

src/components/datepicker/datePicker.js
kseamon: Looks like a scope-local $digest suffice?

src/components/datepicker/datePicker.js
```diff
@@ -0,0 +1,361 @@
...
+   * This is done to avoid setting up additional $watches.
+   */
+  DatePickerCtrl.prototype.installPropertyInterceptors = function() {
+    var self = this;
+
+    // Intercept disabled on the date-picker element to disable the internal input.
+    // This avoids two bindings (outer scope to ctrl, ctrl to input).
+    Object.defineProperty(this.$element[0], 'disabled', {
+      get: function() { return self.isDisabled; },
+      set: function(value) { self.setDisabled(value) }
```
kseamon: angular.bind(self, self.setDisabled)

src/components/datepicker/datePicker.js
```diff
@@ -0,0 +1,361 @@
...
+  };
+
+  /** Detach the floating calendar pane from the document. */
+  DatePickerCtrl.prototype.detachCalendarPane = function() {
+    this.$element.removeClass('md-datepicker-open');
+    this.calendarPane.classList.remove('md-pane-open');
+
+    // Use native DOM removal because we do not want any of the angular state of this element
+    // to be disposed.
+    this.calendarPane.parentNode.removeChild(this.calendarPane);
```
kseamon: FWIW, detach() would be fine.

src/components/datepicker/datePicker.scss
```diff
@@ -0,0 +1,147 @@
+/** Styles for mdDatepicker. */
+$md-datepicker-button-gap: 12px;  // Space between the text input and the calendar-icon button.
+$md-datepicker-border-bottom-gap: 5px;  // Space between input and the grey underline.
+$md-datepicker-open-animation-duration: 0.2s;
+
+
+// The calendar icon button used to open the calendar pane.
+// Need absurd specificty to override md-button.md-icon-button.
+.md-datepicker-button {
+  display: inline-block;
```
kseamon: alphabetize (here and below)

src/components/virtualRepeat/virtualRepeater.scss
```diff
@@ -14,7 +14,8 @@ $virtual-repeat-scrollbar-width: 16px;
     left: 0;
     margin: 0;
     overflow-x: hidden;
-    overflow-y: auto;
+    overflow-y: scroll;
+    -webkit-overflow-scrolling: touch;
```
kseamon: If we're making this change, you don't need the override in the date picker css.

src/components/datepicker/calendar.js
jelbourn: I think it's cleaner to use `querySelector` when there's only one element.

src/components/datepicker/calendar.js
jelbourn: If there were one more case then I'd `switch`

robertmesserle: Still saves some typing =p

src/components/datepicker/calendar.js
jelbourn: Done. These were originally two different cases that did different things.

src/components/datepicker/calendar.scss
jelbourn: We don't alphabetize css. It's more ordered by how the properties fit together (which _is_ rather subjective)

src/components/datepicker/calendar.scss
jelbourn: I think it's fine for the repeater to not do this by default.

src/components/datepicker/calendarMonth.js
jelbourn: I did it this way because using a function is a tiny bit faster (angular doesn't have to parse the expression).

src/components/datepicker/calendarMonth.js
jelbourn: Added to the TODO list for further enhancement.

src/components/datepicker/datePicker.js
jelbourn: Done and done.

src/components/virtualRepeat/virtualRepeater.scss
jelbourn: This was an experimental change I didn't mean to leave in.

src/core/services/theming/theming.js
```diff
@@ -404,13 +405,13 @@ function parseRules(theme, colorType, rules) {
   var themeNameRegex = new RegExp('.md-' + theme.name + '-theme', 'g');
   // Matches '{{ primary-color }}', etc
   var hueRegex = new RegExp('(\'|")?{{\\s*(' + colorType + ')-(color|contrast)-?(\\d\\.?\\d*)?\\s*}}(\"|\')?','g');
-  var simpleVariableRegex = /'?"?\{\{\s*([a-zA-Z]+)-(A?\d+|hue\-[0-3]|shadow)-?(\d\.?\d*)?\s*\}\}'?"?/g;
+  var simpleVariableRegex = /'?"?\{\{\s*([a-zA-Z]+)-(A?\d+|hue\-[0-3]|shadow)-?(\d\.?\d*)?(contrast)?\s*\}\}'?"?/g;
```
jelbourn: I didn't make the changes in this file. Not sure where they came from. I'm going to revert them.

src/components/datepicker/calendar.js
```diff
@@ -0,0 +1,481 @@
...
+    this.dateLocale = $mdDateLocale;
+
+    /** @final {!angular.JQLite} */
+    this.$element = $element;
+
+    /** @final {!angular.Scope} */
+    this.$scope = $scope;
+
+    /** @final {HTMLElement} */
+    this.calendarElement = $element[0].querySelector('.md-calendar');
```
rschmukler: Depending on how dynamic you anticipate the DOM structure of this component being in the future (potential refactors) it could be beneficial to hard-code the relations here rather than use `querySelector` - eg. `$element[0].fistElementChild.children[0]`

src/components/datepicker/calendarMonth.js
```diff
@@ -0,0 +1,239 @@
...
+        });
+      }
+    };
+  }
+
+  /** Class applied to the cell for today. */
+  var TODAY_CLASS = 'md-calendar-date-today';
+
+  /** Class applied to the selected date cell/. */
+  var SELECTED_DATE_CLASS = 'md-calendar-selected-date';
```
rschmukler: Maybe a constant for `md-focus` as well?

src/components/datepicker/calendar.js
```diff
@@ -0,0 +1,481 @@
...
+  };
+
+  /**
+   * Animates the transition from the calendar's current month to the given month.
+   * @param {Date} date
+   * @returns {angular.$q.Promise} The animation promise.
+   */
+  CalendarCtrl.prototype.animateDateChange = function(date) {
+    this.scrollToMonth(date);
+    return this.$q.when();
```
rschmukler: This doesn't actually return the animation promise. Could probably compute the scrolltime and use `$timeout` to mock an animate promise.

src/components/datepicker/calendar.js
jelbourn: I see that as leading to really hard-to-understand code for a potential minuscule performance improvement.

src/components/datepicker/calendar.js
jelbourn: Added a TODO

src/components/datepicker/calendar.js
jelbourn: Originally, this code used `$animate` to actually perform some animation, and the code that consumed this function, of course, treated it as if it were async. When I changed it to not actually do any animation upon date change, I decided to leave the API the same so that an animation could be added back in later without changing the call sites. As far as the caller is concerned, this function _is_ performing some animation that may or may not be async (which is one of the great things about promises).

src/components/datepicker/calendarMonth.js
jelbourn: Done.

PR #3688, opened 2015-07-11
placeholder directive ng-click problem fix

dbwhddn10:


iver56:
Does this pull request fix the issue that is shown in the gif below?
double placeholder

I.e. there are two labels (that overlap), while it should be one. This happens when I put an angular expression such as `{{ 'EMAIL' | translate }}` in my placeholder.

EladBezalel:
Irrelevant, please check on HEAD if your code is still valid please say and i will reopen

PR #3711, opened 2015-07-13
docs(dialog): demo will use it's cancel function

Aaron-Hartwig:
Change the `ng-click` event of the dialog's close button to use the cancel function.

Otherwise the `cancel` function was dead code. I think calling the `answer` function with an arbitrary parameter is a poor example of how to close a dialog.


ThomasBurleson:
@Aaron-Hartwig - agreed. I tried to merge your PR, but experienced rebase issues.

PR #3718, opened 2015-07-14
fix(icon): Log network errors without throwing exceptions.

### LINKED ISSUE #2530: md-icon: Gracefully handle network errors
It would be nice if `md-icon` gracefully handled network errors. Right now it isn't safe to use `md-icon` if your endpoint is down. It will throw exceptions because it doesn't check certain objects/properties before it uses them.

For example the following line isn't safe https://github.com/angular/material/blob/977a495e7d3de27bebabdebaaf5611c9d06a1e22/src/components/icon/iconService.js#L357
```
"TypeError: id.lastIndexOf is not a function
    at loadFromIconSet (http://dev.*******.com/app/vendor.js:25561:40)
    at https://ajax.googleapis.com/ajax/libs/angularjs/1.3.15/angular.min.js:112:113
    at n.$get.n.$eval (https://ajax.googleapis.com/ajax/libs/angularjs/1.3.15/angular.min.js:126:15)
    at n.$get.n.$digest (https://ajax.googleapis.com/ajax/libs/angularjs/1.3.15/angular.min.js:123:106)
    at n.$get.n.$apply (https://ajax.googleapis.com/ajax/libs/angularjs/1.3.15/angular.min.js:126:293)
    at l (https://ajax.googleapis.com/ajax/libs/angularjs/1.3.15/angular.min.js:81:240)
    at M (https://ajax.googleapis.com/ajax/libs/angularjs/1.3.15/angular.min.js:85:342)
    at XMLHttpRequest.e (https://ajax.googleapis.com/ajax/libs/angularjs/1.3.15/angular.min.js:86:418)"
```

You can test this by just shutting your server off after you load the first page and go to a second page which has a new icon. It might also be a good idea when this is fixed to implement a missing icon which could be hard coded into the js. This could be used instead if there are any errors loading icons.

Also this hopefully isn't too important on desktop as servers don't go down often(we hope) and you shouldn't be losing your network. But it might be a much larger problem on mobile where your signal constantly goes in and out if your not in a large city.

programmist:
fixes #2530

### DIFF
--- a/src/components/icon/iconService.js
+++ b/src/components/icon/iconService.js
@@ -389,10 +389,8 @@
      if ( urlRegex.test(id)     ) return loadByURL(id).then( cacheIcon(id) );
      if ( id.indexOf(':') == -1 ) id = '$default:' + id;
 
-     return loadByID(id)
-         .catch(loadFromIconSet)
-         .catch(announceIdNotFound)
-         .catch(announceNotFound)
+     var load = config[id] ? loadByID : loadFromIconSet;
+     return load(id)
          .then( cacheIcon(id) );
    }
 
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@programmist - needs some unit tests for these, plz.

programmist:
@ThomasBurleson Test added.

jelbourn:
LGTM

Issue #1259, opened 2015-01-21
Trying to do a md-toolbar with md-scroll-shrink and a fixed Tab/Search Box below of md-toolbar.


ekaitzht:
Hi, I am trying to do [this](http://material-design.storage.googleapis.com/publish/v_2/material_ext_publish/0B0NGgBg38lWWU0RhQy1ab3FWOWM/patterns-scrollingtech-scrolling-070801_TabBar_xhdpi_003.mp4), which  is not the same like [this](http://material-design.storage.googleapis.com/publish/v_2/material_ext_publish/0B0NGgBg38lWWcFhaV1hiSlB4aFU/patterns-scrollingtech-scrolling-070801_Flexible_Space_xhdpi_003.webm) because the content is not scrollable, about this feature people are discusssing here https://github.com/angular/material/issues/1053.

Should be this open for a new enhancement? or is possible to do with the current state of angular-material.

Thanks!

EladBezalel:
Can you explain why its not the same? 
i opened an issue about it #1054

ekaitzht:
If you see the Navigation Bar disappears and the Tab Bar goes to the top but in the second case the Navigation Bar is fixed. [here](http://www.google.es/design/spec/patterns/scrolling-techniques.html#scrolling-techniques-scrolling) are both examples and I think that behaviours are different.

EladBezalel:
This is pretty much the same.. I will edit my issue just to be clear with it. thanks

infacq:
hi has it been implemented?

marminthibaut:
:+1:

ismarslomic:
See this example: http://codepen.io/infacq/pen/ZGabYq

Zephyr07:
Thanks ismarslomic it's work for me

Issue #1054, opened 2014-12-26
mdToolbar: Implement Flexible space with drag  and animations


EladBezalel:
The app bar is flexible and can be extended to accommodate larger typography or images. To extend the app bar, add a flexible space block.

The [Flexible space](http://www.google.com/design/spec/patterns/scrolling-techniques.html#scrolling-techniques-scrolling) shrinks until only the navigation bar is left. The title should also shrink into place to become a 20sp title in the navigation bar. When scrolling all the way back, the flexible space and the title grow into place again.
The whole app bar scrolls off. When the user reverse-scrolls, the navigation bar returns anchored to the top. When scrolling all the way back, the flexible space and the title grow into place again.

[fexlible space video](http://material-design.storage.googleapis.com/publish/v_2/material_ext_publish/0B0NGgBg38lWWcFhaV1hiSlB4aFU/patterns-scrollingtech-scrolling-070801_Flexible_Space_xhdpi_003.webm)
![flexible space](http://material-design.storage.googleapis.com/publish/v_2/material_ext_publish/0Bx4BSt6jniD7TS1teXNiZ1ZLVUU/patterns_scrolling_scroll4.png)

ajoslin:
Flexible space needs to be implemented in some other components as well. This will be an interesting one...

EladBezalel:
As mentioned in #1259 - Flexible space should support the existence of tabs

![scrolling techniques - patterns - google design guidelines](https://cloud.githubusercontent.com/assets/6004537/5873577/a8fe2f5e-a300-11e4-80e7-631cba4d18a2.png)

cskiwi:
Any update on when this would be implemented? the milestone is just being updated.

infacq:
Any update on this? It's interesting as I couldn't get to make header just like http://www.materialup.com/

brianfeister:
+1 surprised not to see more votes on this one.

adarshmuthappa:
+1

TiS:
+1

guiporto:
+1

EladBezalel:
as @guiporto mentioned in a linked issue, this is a nice implementation:
http://codepen.io/jbltx/pen/KwQxBX

ThomasBurleson:
@EladBezalel - that is an excellent demo; very impressive.

patouche7519:
[...]

Issue #1053, opened 2014-12-26
md-toolbar: allow fixed (sticky)


DmitryEfimenko:
Hello,

This is more of a feature request.

I know there is a way to make toolbar appear and disappear as user scrolls via the use of `md-scroll-shrink`, but I'm wondering if there is a way to make toolbar fixed at the top so that it's visible at all times.

To be exact, I'm actually looking for a way to make the effect similar to the effect that can be observed on [TV-Series app](https://play.google.com/store/apps/details?id=fema.serietv2&hl=en), where originally toolbar is large and has a background image, but as you scroll down the image disappears and the toolbar shrinks a bit.

I understand that the behavior described above is quite custom, and I'm ok writing custom code to achieve that, but for starters it'd be nice to have an easy way to make toolbar fixed at the top.

Perhaps it would make sense to create a new directive `md-header` (similar to how [ionic is doing it](http://ionicframework.com/docs/components/#header)) that would have fixed position by default and could potentially have an option to behave like described above.

Any help is appreciated,
Thanks.

Maxxoo:
@DmitryEfimenko, wrapping the toolbar in a `<header>...</header>` makes it sticky even though other elements (absolutely positioned) exist at the root of the `<body>` (for instance, an `<md-sidenav>`)

AMoQu:
Any update on this missing feature?

wlingke:
@AMoQu My current fix for this is just to add a class to md-toolbar that implements `position: fixed;`

infacq:
@wlingke do you have the example?

wlingke:
@infacq it's pretty easy to do it yourself... `style="position: fixed;"`

infacq:
do you have example with the shrink one. I cannot get my head around it

wlingke:
No this does not do a shrinking one. You'll have to wait until it comes out.
ᐧ

Best,
Lingke

--
Lingke Wang
Stanford GSB '16

noamamit92:
Will be useful to have this this with the shrinking effect built in

guiporto:
I believe this is a good reference for this issue:
http://codepen.io/jbltx/pen/KwQxBX

shinrox:
+1

DmitryEfimenko:
I believe the use-case when you cannot put the `<md-toolbar>` as the first child element of the `<body>` is quite common and is the main source for the need for `<md-toolbar>` to be sticky inside other elements.
```html
<body>
    <ui-view></ui-view>
[...]

PR #3735, opened 2015-07-15
Wip/scroll fade toolbar

NickBolles:
Implement scroll-fade behavior on toolbars as discussed in #1054

modified from
http://codepen.io/jbltx/pen/KwQxBX

Related to #1053 #1259

I would love feedback on this and any suggestions!

Demos:
http://nickbolles.github.io/material/dist/demos/toolbar/

The new demos are:
demoScrollFade
demoScrollAway
demoScrollTabs (Not working yet I dont think, just threw this in at the end)


NickBolles:
Also there are no tests for this yet. I'm assuming that will need to be done for sure before merging. I will work on this, although my experience with testing is next to none, it will be good practice :)

ThomasBurleson:
Please rebase (from origin master) and squash your commits; after you includes tests. Thx.

NickBolles:
Updated to include tests. There a few things that are not quite ready for prime time.
-The content is longer then the container sometimes, causing the toolbar to scroll off the screen. I am not quite sure how to fix this. I thought that using flex would fix it, but it didnt. I will be looking into this more.
- Displaying the picture in toolbar on desktop. How should this be handled? right now it just goes by the toolbar header and only shows the top amount of the picture.
- I started adding a `content-speed-factor` that could keep the top of the content visible untill the toolbar was completely condensed (by adding padding to the top of the content as you scroll) but it caused very very poor performance. So I commented it out for the commit. Ideas would be awesome for this. Possibly calculating the padding we need at the beginning and adding it then translating the content down as you scroll up
-tabs/search bar in the toolbar (this is more of a demo then an implementation, it should be doable with the current version

Let me know questions, comments, concerns and suggestions!

NickBolles:
Here are the links to the demos:
- Simple Fade Demo: http://nickbolles.github.io/material/dist/demos/toolbar/demoScrollFade
- Fade Away Demo: http://nickbolles.github.io/material/dist/demos/toolbar/demoScrollAway
- Overlap Demo: http://nickbolles.github.io/material/dist/demos/toolbar/demoScrollOverlap
- Nested Demo: http://nickbolles.github.io/material/dist/demos/toolbar/demoNested

NickBolles:
Also mobile performance is pretty poor. How can we improve this? It's not terrible but it definitely stutters when scrolled fast.

ThomasBurleson:
@NickBolles  :+1: This looks impressive. I will test next week and merge if appropriate.

NickBolles:
@ThomasBurleson have you had a chance to take a look at this? If so any questions, comments, concerns? 

My concern is that this isn't the best way to implement this. And that more of a "toolbar panel" approach would be better. Then the toolbar panel can control the toolbar flexible space and content inside it. Let me know your thoughts

ThomasBurleson:
@NickBolles - Let me look at this tomorrow.

NickBolles:
@ThomasBurleson sounds good!

ThomasBurleson:
@NickBolles - As I mentioned I am very impressed with the effort and attention to detail you have invested in ScrollFade Toolbars. My concern is that these types of effects should not be hardcoded into Toolbar (with lots of control directives). 

Instead almost all of these effects (resizing, repositioning, alpha changes, color changes, etc) are typically seen in animation UX(es). Consider a animation timeline that animates from expanded to shrunk modes. The scrolling is effectively a manual scrubbing of the timeline between those two modes. 

I contend that the appropriate way to enhance the toolbar with transform effects (like scroll fade) is ngAnimate Timeline features; but those are not yet available.

> This weekend I will prepare a CodePen to show you exactly what I mean. And we can continue this discussion

ThomasBurleson:
Once we identify an architectural approach that makes sense we can merge in your demos, etc. So this PR remains valuable and useful.

NickBolles:
@ThomasBurleson Sounds good. As you can tell from my comments I was starting to see that my approach may not have been the best as I was completing it. I am interested to see your codePen. I have a feeling it will be something like whats in your repo here: https://github.com/ThomasBurleson/angularjs-gsTimelines

Which I had seen previously and am very interested in.

NickBolles:
@ThomasBurleson I never heard back for a code pen on what you meant. (I'm sure your busy it's no problem) Just wondering why this was closed. If this isn't the way to go with the toolbar what would be the best way to go? Or is the material team going to add this feature soon?

JunedS:
+1 

Hope to see this make soon to 1.0-rc1. Badly in need for this feature. I had done a temp work around but its not as polished as this. 

+1
+1
+1

Common let this happen! Btw good job NickBolles

ThomasBurleson:
@NickBolles - the animation features with element anchoring will be added after the Angular Material v1.0 release.

NickBolles:
Awesome. I will look at it and reevaluate this PR, unless this is going to be developed by the material team?

PR #3753, opened 2015-07-15
fix(tabs): Use correct variable for md-center-tabs

### LINKED ISSUE #2450: md-tabs not center corret
md-tabs not center correct if  md-center-tabs="true" and change browser windows size.

epelc:
It was using `shouldCenterTabs` instead of `centerTabs` which `shouldCenterTabs()` to always return false as `shouldCenterTabs` was always undefined.

Fixes #2450

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -41,7 +41,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
   ctrl.lastSelectedIndex = null;
   ctrl.hasFocus = false;
   ctrl.lastClick = true;
-  ctrl.shouldCenterTabs = shouldCenterTabs();
+  ctrl.centerTabs = shouldCenterTabs();
 
   //-- define public methods
   ctrl.updatePagination = $mdUtil.debounce(updatePagination, 100);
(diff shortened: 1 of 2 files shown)


epelc:
@robertmesserle Can you take a look at this? It's a simple fix.

ThomasBurleson:
@robertmesserle - sgty ?

robertmesserle:
@ThomasBurleson @epelc I'm not sure what the benefit here is, but I have no complaints - it's just a variable change.  I think I originally called it `centerTabs

epelc:
@robertmesserle @ThomasBurleson It was assigning it to the wrong variable as it wasn't syncing with the `defineBooleanAttribute()` so it was always undefined and `shouldCenterTabs()` always returned false because of it.

robertmesserle:
Let me look over the code a bit more before merging, this may be against
the intent of the code. (On my phone now so I can't really check just yet.)

robertmesserle:
@epelc This is due to a timing bug where `$attrs.$observe` was not firing until after `ctrl.shouldCenterTabs` was defined, thus it was always undefined.  I submitted a fix for this here: https://github.com/angular/material/commit/d52e9c29a43e496ad822278ae3b5e1758b7b9577

epelc:
@robertmesserle Is that the right commit? It's adding `preventDefault` to autocomplete's key events which seems really unrelated.

It's also not working in the latest master but I don't see any other commit's related to this.

robertmesserle:
Sorry about that - no, no it was not the right commit, lol.  Apparently my push failed and I just grabbed the latest SHA.

Here is the correct commit, pushed just now: https://github.com/angular/material/commit/aa1e47d37e692cf2bd0a0d4e3646defd4f89fc7b

epelc:
Thanks it's working now.

PR #3755, opened 2015-07-16
Contradictory input validation

rrgarciach:
Evaluating a required input to be dirty and then conditioning the error message to be displayed only when dirty seems very contradictory.


rrgarciach:
I signed it!

EladBezalel:
Input is designed and behaves as the spec defines, 
messages shouldn't be visible unless the input is dirty.

ThomasBurleson:
Out-dated, invalid, or not core focus for Material 1.

PR #3756, opened 2015-07-16
Fix documentation for tabs disabled attribute.

meatherly:
The docs say use the disabled attribute but it's actually ng-disabled.


meatherly:
I signed it!

PR #3794, opened 2015-07-18
fix(select): interimElement will resolve its promise even when

### LINKED ISSUE #3704: md-select in latest master locks the screen
when something is clicked in an md-select, ice the md-select closes you can no longer click any other element on the screen

robertmesserle:
transition end event is not fired

Closes #3704

### DIFF
--- a/src/core/services/interimElement/interimElement.js
+++ b/src/core/services/interimElement/interimElement.js
@@ -412,7 +412,7 @@ function InterimElementProvider() {
           },
           remove: function() {
             self.cancelTimeout();
-            return removeDone = $q.when(showDone).then(function() {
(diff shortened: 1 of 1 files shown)


robertmesserle:
@ThomasBurleson Wanted you to look over this before merging in.

robertmesserle:
@ThomasBurleson I may spend some time over the weekend looking into a more proper fix for this, so we should probably hold off until Monday to merge it - but it should be fine to merge either way.

ThomasBurleson:
This looks good (and in hindsight so obvious). Now:
*  `show()` and `remove()` use `.finally()`. 
*  dialogPopOut. Select, and menu-interim-element use `.finally`

Issue #2770, opened 2015-05-07
secondary and primary placeholders switched in md-chips


typotter:
In the [Chips Basic Usage Demo](https://material.angularjs.org/#/demo/material.components.chips), under 'Use Placeholders and override hint texts'. The placeholder text should be +tag when the collection is non empty, and the longer, "Enter a tag" when the collection is empty.

ericgundrum:
This is caused by a logic error on line 186 of src/components/chips/js/chipsController.js
`return useSecondary ? this.placeholder : this.secondaryPlaceholder;`
should be
`return useSecondary ? this.secondaryPlaceholder : this.placeholder;`

Note that this error appears in production code, not just in the demo. The error is present on branch `master` as well as tag `1.0.0-rc7`.

ericgundrum:
The pull request @dbwhddn10 associated with #3804 will fix this issue.

PR #3804, opened 2015-07-18
fix secondary placeholder problem in chips

dbwhddn10:


topherfangio:
I need to dig into this a bit further before we can review it.

@dbwhddn10 Can you provide any additional info as to what this fixes? Ideally a description, test case and an issue # would be fantastic!

ericgundrum:
@topherfangio, I have unit tests demonstrating issues #2770 and #4476. (Currently there are no unit tests for the _placeholder_ feature of `mdChips` input.) If you like, I can offer PRs with those tests and the related fix. (The fix is a single line.) Or, I could just post the tests here or in one of the issue threads. Just let me know what your prefer.

I wrote a test to match the issue PR a5dba2a27e44ee04ad47f1cf0ef4a9e4212eef79 is trying to fix. However, there is much other `mdChips` code that fails to check if the `ngModel` is defined. There seems to be an inherent assumption that `ngModel` is defined or that catastrophic failure is preferred when it isn't. It may be better to separate this issue and PR from issues #2770 and #4476.

EladBezalel:
@ericgundrum thanks.
@dbwhddn10 can you please rebase?
@topherfangio can you please investigate? :)

topherfangio:
I'm going to go ahead and close this in favor of #6175 since it does the NPE check only and #6535 has already been merged.

PR #3805, opened 2015-07-18
remove not necessary code in placeholderDirective

dbwhddn10:


EladBezalel:
This class is necessary

PR #3816, opened 2015-07-19
don't allow undefined objects in md-chips items list

bengoldsh:
There is no easy way to eliminate invalid chips before insertion. By adding a simple check for undefined before items are pushed to the items' array we get a simple way to filter chips that we don't want to be shown in the chips list.
This feature gives the option to filter unwanted items on md-on-append by returning undefined.


bengoldsh:
I signed it!

PR #3825, opened 2015-07-20
feat(autocomplete): adds a new attribute option `md-select-on-match`

### LINKED ISSUE #3324: md-autocomplete compels you to click on dropdown item even typing the full word
When typing on autocomplete, lets say "Alabama", the dropdown shows "Alabama", but if you don't click the dropdown item and just leave the focus, no object is selected.

I think if if you fully write an item, the directive should match with the object without require you to click it.

It could be bypassed in many ways, for example, when you blur the field, it automatically selects the first object of the list. Or if you type "Alaba.." and hit Enter, also selects the first item.

I tested this behavior with 11 people and 9 of them typed the full word without selecting the dropdown item. =/

robertmesserle:
When enabled, an exact match in the `searchText` will automatically
select a matching item if (a) there is exactly one match found and (b)
the `searchText` matches the display value exactly

Closes #3324

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -529,6 +529,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       if (searchText !== $scope.searchText) return; //-- just cache the results if old request
       ctrl.matches = matches;
       ctrl.hidden  = shouldHide();
+      if ($scope.selectOnMatch) selectItemOnMatch();
       updateMessages();
       positionDropdown();
     }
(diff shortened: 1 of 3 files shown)


robertmesserle:
This PR also cleans up the `spec` code for older tests to remove unnecessary steps based on the current implementation.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -593,4 +594,13 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     if (hasFocus) ctrl.hidden = shouldHide();
   }
 
+  function selectItemOnMatch () {
```
ThomasBurleson: Need a code comment

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -128,6 +128,7 @@ function MdAutocomplete () {
       itemText:       '&mdItemText',
       placeholder:    '@placeholder',
       noCache:        '=?mdNoCache',
+      selectOnMatch:  '=?mdSelectOnMatch',
```
ThomasBurleson: Like the name

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -529,6 +529,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       if (searchText !== $scope.searchText) return; //-- just cache the results if old request
       ctrl.matches = matches;
       ctrl.hidden  = shouldHide();
+      if ($scope.selectOnMatch) selectItemOnMatch();
```
ThomasBurleson: I think we should `angular.noop` this by default

ThomasBurleson:
@robertmesserle - my comments aside, looks good to me.

src/components/autocomplete/js/autocompleteController.js
robertmesserle: The user only provides a flag, not the function itself.  No need for `angular.noop`, unless I am misunderstanding your suggestion.

PR #3840, opened 2015-07-21
fix(select): text-align:end is not supported in IE

### LINKED ISSUE #2213: md-select ▼ displaying underneath text in IE10
For IE10, the arrow, ▼, that appears to the right in browsers other than IE10 for md-select appears under the text in IE10.  This can be seen at https://material.angularjs.org/#/demo/material.components.select.

prestonvanloon:
`text-align: end` is not supported by some browsers. This can cause improper positioning of the md-select-icon

https://developer.mozilla.org/en-US/docs/Web/CSS/text-align

Fixes #2213

### DIFF
--- a/src/components/select/select.scss
+++ b/src/components/select/select.scss
@@ -104,7 +104,7 @@ md-select {
 
   .md-select-icon {
     align-items: flex-end;
-    text-align: right;
+    @include rtl(text-align, right, left);
     width: 3 * $baseline-grid;
     margin: 0 .5 * $baseline-grid;
     transform: translate3d(0, 1px, 0);
--- a/src/components/select/select.scss
+++ b/src/components/select/select.scss
@@ -103,7 +103,7 @@ md-select {
   }
 
   .md-select-icon {
-    align-items: flex-end;
+    @include rtl(align-items, flex-end, flex-start);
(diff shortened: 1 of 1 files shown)


clshortfuse:
Using right instead of end would conflict with right-to-left languages.

prestonvanloon:
Good point, I will update

ThomasBurleson:
@prestonvanloon - you should probably change this also;

`@include rtl(align-items, flex-end, flex-start);`

prestonvanloon:
@ThomasBurleson - agreed

prestonvanloon:
I think this has been resolved elsewhere

Splaktar:
This was tweaked in https://github.com/angular/material/commit/70c0c6e80391026c09785c4703becb34a690ba7b, but that only made the carrot not show up under the text. It didn't fix the left-to-right alignment.

This is still a problem as noted in https://github.com/angular/material/issues/10714.

clshortfuse:
It's a bit confusing, but it seems what happened was that this PR was closed by the creator and never merged. It would have solved the issue properly. The second commit that came later in August was a partial fix, but that was a direct commit without a PR and wasn't reviewed.

Issue #3828, opened 2015-07-20
mdDialog - IE11/Win8.1 targetEvent no longer causing dialog to open/close from button


Londovir:
Confirmed on Demo site using #master. Under previous commits, including most of 0.10.1-rc1, using IE11/Win8.1 showed the expected opening of an md-dialog as "growing" from the targetEvent button originator, and "shrinking" down to that area as well.

With recent code change around 0.10.1-rc2, the targetEvent parameter is not getting used by the code to perform this behavior. Instead, IE11/Win8.1 seems to lag after clicking a button, and after about a half second delay the backdrop and dialog appear almost simultaneously on the screen with no/little animation. When dismissing, the dialog dismisses downwards towards the center bottom of the viewport, which seems to be the behavior when no targetEvent is indicated.

I do not see this behavior in Chrome or Firefox, however.

Londovir:
@ThomasBurleson Although this was closed on July 21, I'm still (as of 9/14/15) seeing the behavior I described above failing in IE11. This is running on Windows 10 IE11:

![ie11dialog](https://cloud.githubusercontent.com/assets/11982575/9858637/c620b4b0-5aef-11e5-8c43-a3073ec07587.gif)

IE11 has a number of differences to note:

1) The backdrop does not become semi-transparent, but looks to be completely opaque.
2) The backdrop completes its animation into place, and **then** the dialog animates into place.
3) The dialog does not expand from the clicked button, nor dismiss away to the button either.

Now, for contrast, here's the same in Chrome:

![chromedialog](https://cloud.githubusercontent.com/assets/11982575/9858647/df887f3c-5aef-11e5-8b6c-35f883da8e72.gif)

This is as it is expected. (Performed from most updated IE11 in Windows 10 and Chrome in Windows 10 from MD Demo page.)

If you believe this is an unrelated issue to the original issue here, then let me know and I can create this as a new issue rather than reopen this one, but I don't believe it has been addressed for IE11.

BTW, Edge performs identical to Chrome for me. It is only IE11 that is exhibiting this issue.

abrahamtorres10:
http://plnkr.co/edit/6URLKYpatz3EeJ7bkpVA?p=preview i can also reproduce this issue, not fixed yet

sikha:
Same problems. Does anyone have any solutions?

PR #3841, opened 2015-07-21
fix(backdrop, dialog, css): improve position and animations of backdrop

### LINKED ISSUE #3826: md-dialog: Timing change on backdrop removal in newer commits to master
Is it just me, or is there a definite change in the newer commits that changes the way the backdrop is removed when an md-dialog is dismissed? In previous versions 0.10.0 for certain, perhaps even 0.10.1-rc1, dismissing a dialog would transition the backdrop simultaneously with the removal of the dialog itself. Now, however, as of the recent commits, this behavior has been replaced with a two stage process where the dialog first shrinks away (stage 1), and *then* the backdrop fades out (stage 2).

This change has been somewhat jarring, and although I can not find a specific example or page in the Material Design guidebook related to UI interactions for the display and removal of dialogs, by grabbing my Android phone it seems like this is not the behavior cooked into material design in the Google apps I tested. It seems like the backdrop fade and dialog dismissal are simultaneous, and not delayed. I'd guess the backdrop is now tied to a timeout or promise in the code which explains why it is waiting to happen after the dialog has disappeared.

Will this be corrected or is this the planned UI behavior as of this time for the future?

ThomasBurleson:
improve backdrop animations and css to account for scroll and parenting:
-  backdrop
  - css uses transition and ng-enter/ng-leav
  - css transitions opacity instead of background-color
  - css no longer uses key frames
  - postLink uses $animate.pin if available
  - postLink logs a warning if the backdrop parent has a style 'position:static'
- util
  - disableElementScroll uses specified element or body; used by dialog to disable dialog parent scrolling when parent is explicitly specified 
- dialog
  - css for 'md-dialog-container' is now 'position:absolute'; 'fixed' is deprecated
  - hide backdrop now runs in parallel with hide dialog
  - basic demo #1 now uses element '#popupContainer' as parent; instead of document.body
  - basic demo #1 has #popupContainer 'position:relative' styling

Fixes #3826, #3828, #1967, #1106

### DIFF
--- a/config/build.config.js
+++ b/config/build.config.js
@@ -3,7 +3,7 @@ var fs = require('fs');
 var versionFile = __dirname + '/../dist/commit';
 
 module.exports = {
-  ngVersion: '1.3.15',
+  ngVersion: '1.4.3',
   version: pkg.version,
   repository: pkg.repository.url
     .replace(/^git/,'https')
--- a/src/components/backdrop/backdrop.js
+++ b/src/components/backdrop/backdrop.js
@@ -17,11 +17,35 @@
  *
  */
 
-angular.module('material.components.backdrop', [
-  'material.core'
-])
-  .directive('mdBackdrop', BackdropDirective);
+angular
+  .module('material.components.backdrop', ['material.core'])
+  .directive('mdBackdrop', function BackdropDirective($mdTheming, $animate, $rootElement, $window, $log, $$rAF) {
 
-function BackdropDirective($mdTheming) {
-  return $mdTheming;
-}
+    return {
+        restrict: 'E',
+        link: postLink
+      };
+
+    function postLink(scope, element, attrs) {
+      // backdrop may be outside the $rootElement, tell ngAnimate to animate regardless
+      if( $animate.pin ) $animate.pin(element,$rootElement);
+
+      $$rAF(function(){
+        // Often $animate.enter() is used to append the backDrop element
+        // so let's wait until $animate is done...
+
+        var parent = element.parent()[0];
+        if ( parent ) {
+          var position = $window.getComputedStyle(parent).getPropertyValue('position');
+          if (position == 'static') {
+            // backdrop uses position:absolute and will not work properly with parent position:static (default)
+            var positionError = "<md-backdrop> may not work properly in a scrolled, static-positioned parent container.";
+            $log.warn( positionError );
+          }
+        }
+
+        $mdTheming.inherit(element, element.parent());
+      });
+
+    };
+  });
--- a/src/components/backdrop/backdrop.scss
+++ b/src/components/backdrop/backdrop.scss
@@ -16,6 +16,9 @@ md-backdrop {
     z-index: $z-index-sidenav - 1;
   }
 
+  opacity: 1;
+  transition: opacity 600ms $swift-ease-in-timing-function;
+
   background-color: rgba(0,0,0,0);
 
   position: absolute;
@@ -25,15 +28,23 @@ md-backdrop {
   right: 0;
 
   &.md-click-catcher {
-    top: 0;
     position: fixed;
   }
 
   &.ng-enter {
-    animation: $swift-ease-out-timing-function mdBackdropFadeIn 0.5s both;
+    opacity: 0;
+    &.ng-enter-active {
+      opacity: 1;
+    }
   }
+
   &.ng-leave {
-    animation: $swift-ease-in-timing-function mdBackdropFadeOut 0.4s both;
+    opacity: 1;
+    transition: opacity 400ms linear;
+
+    &.ng-leave-active {
+      opacity: 0;
+    }
   }
 
   &.md-opaque {
@@ -41,11 +52,3 @@ md-backdrop {
   }
 }
 
-@keyframes mdBackdropFadeIn {
-  from { opacity: 0; }
-  to { opacity: 1; }
-}
-@keyframes mdBackdropFadeOut {
-  from { opacity: 1; }
-  to { opacity: 0; }
-}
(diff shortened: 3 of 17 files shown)


ThomasBurleson:
@robertmesserle - would you mind reviewing for inclusion in RC3

src/components/backdrop/backdrop.js
```diff
@@ -17,11 +17,35 @@
...
+        // Often $animate.enter() is used to append the backDrop element
+        // so let's wait until $animate is done...
+
+        var parent = element.parent()[0];
+        if ( parent ) {
+          var position = $window.getComputedStyle(parent).getPropertyValue('position');
+          if (position == 'static') {
+            // backdrop uses position:absolute and will not work properly with parent position:static (default)
+            var positionError = "<md-backdrop> may not work properly in a scrolled, static-positioned parent container.";
+            $log.warn( positionError );
```
robertmesserle: :+1:

src/components/backdrop/backdrop.scss
```diff
@@ -16,6 +16,9 @@ md-backdrop {
     z-index: $z-index-sidenav - 1;
   }
 
+  opacity: 1;
+  transition: opacity 600ms $swift-ease-in-timing-function;
```
robertmesserle: A bit nit-picky, but this probably belongs inside the `.ng-enter` selector.

src/components/dialog/dialog.js
```diff
@@ -605,9 +605,7 @@ function MdDialogProvider($$interimElementProvider) {
 
         element.css('top', parentOffset + 'px');
 
-        options.backdrop = angular.element('<md-backdrop class="md-dialog-backdrop md-opaque">');
-        options.backdrop.css('top', parentOffset + 'px');
-        $mdTheming.inherit(options.backdrop, options.parent);
+        options.backdrop = $compile('<md-backdrop class="md-dialog-backdrop md-opaque">')(scope);
```
robertmesserle: Any reason for compiling this rather than just using `angular.element`?  It doesn't look to have anything dynamic in it.

src/components/dialog/dialog.scss
```diff
@@ -7,7 +7,7 @@ $dialog-padding: $baseline-grid * 3;
   display: flex;
   justify-content: center;
   align-items: center;
-  position: fixed;
+  position: absolute;
```
robertmesserle: I'd like to hear from @rschmukler on this one, just to confirm that this is safe and we aren't missing some edge case.

src/components/menu/menu-interim-element.js
```diff
@@ -83,7 +83,7 @@ function MenuProvider($$interimElementProvider) {
           target: angular.element(opts.target), //make sure it's not a naked dom node
           parent: angular.element(opts.parent),
           menuContentEl: angular.element(element[0].querySelector('md-menu-content')),
-          backdrop: opts.hasBackdrop && angular.element('<md-backdrop class="md-menu-backdrop md-click-catcher">')
+          backdrop: opts.hasBackdrop && $compile('<md-backdrop class="md-menu-backdrop md-click-catcher">')(scope)
```
robertmesserle: Same question here - do we have to compile it for any reason?

src/components/select/select.js
```diff
@@ -774,7 +774,7 @@ function SelectProvider($$interimElementProvider) {
         parent: angular.element(opts.parent),
         selectEl: element.find('md-select-menu'),
         contentEl: element.find('md-content'),
-        backdrop: opts.hasBackdrop && angular.element('<md-backdrop class="md-select-backdrop md-click-catcher">')
+        backdrop: opts.hasBackdrop && $compile('<md-backdrop class="md-select-backdrop md-click-catcher">')(scope)
```
robertmesserle: Same question - also, after seeing this string 3 times, I wonder if this should be shared somewhere rather than duplicated 3 times in the code.

src/components/sidenav/sidenav.js
```diff
@@ -240,9 +240,7 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
         $mdMedia: $mdMedia
       });
     };
-    var backdrop = $compile(
-      '<md-backdrop class="md-sidenav-backdrop md-opaque ng-enter">'
-    )(scope);
+    var backdrop = $compile('<md-backdrop class="md-sidenav-backdrop md-opaque ng-enter">')(scope);
```
robertmesserle: 4 times.

config/build.config.js
```diff
@@ -3,7 +3,7 @@ var fs = require('fs');
 var versionFile = __dirname + '/../dist/commit';
 
 module.exports = {
-  ngVersion: '1.3.15',
+  ngVersion: '1.4.3',
```
ThomasBurleson: Note: this sets the local testing and online docs to use 1.4.3.

src/components/backdrop/backdrop.js
```diff
@@ -17,11 +17,35 @@
...
+      $$rAF(function(){
+        // Often $animate.enter() is used to append the backDrop element
+        // so let's wait until $animate is done...
+
+        var parent = element.parent()[0];
+        if ( parent ) {
+          var position = $window.getComputedStyle(parent).getPropertyValue('position');
+          if (position == 'static') {
+            // backdrop uses position:absolute and will not work properly with parent position:static (default)
+            var positionError = "<md-backdrop> may not work properly in a scrolled, static-positioned parent container.";
```
ThomasBurleson: perhaps this should be `var ERROR_CSS_POSITION` ?

config/build.config.js
robertmesserle: Yep, I think that's good - as long as it's stable, we should be running against the latest version of Angular.

PR #3845, opened 2015-07-21
Add stopPropagation on keyEvents in md-menu

clshortfuse:
I also added a preventDefault on ESCAPE.

### DIFF
--- a/src/components/menu/menu-interim-element.js
+++ b/src/components/menu/menu-interim-element.js
@@ -141,9 +141,9 @@ function MenuProvider($$interimElementProvider) {
         opts.menuContentEl.on('keydown', function(ev) {
           scope.$apply(function() {
             switch (ev.keyCode) {
-              case $mdConstant.KEY_CODE.ESCAPE: opts.mdMenuCtrl.close(); break;
-              case $mdConstant.KEY_CODE.UP_ARROW: focusMenuItem(ev, opts.menuContentEl, opts, -1); break;
-              case $mdConstant.KEY_CODE.DOWN_ARROW: focusMenuItem(ev, opts.menuContentEl, opts, 1); break;
+              case $mdConstant.KEY_CODE.ESCAPE: opts.mdMenuCtrl.close(); ev.stopPropagation(); ev.preventDefault(); break;
+              case $mdConstant.KEY_CODE.UP_ARROW: focusMenuItem(ev, opts.menuContentEl, opts, -1); ev.stopPropagation(); break;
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@clshortfuse - what was the original issue that prompted this PR ?

clshortfuse:
I have a table with rows that allows keyboard input for navigation. When the menu pop ups, the menu passes the up and down keys to its parent element (in this case the table). This isn't ideal. 

I can create a test if the explanation isn't clear.

ThomasBurleson:
A unit test would be awesome. And how does `preventDefault()` help with ESC keys presses?

clshortfuse:
It seems as though the issue is related to using a controller on the the body element, which I would have to do to support global commands. preventDefault() on would escape would stop the browser from doing it's regular action (stopped the page from loading).

Here's the unit test.
http://codepen.io/anon/pen/gpdxpq

Use up and down to navigate the table and press space bar to activate the menu. When you try to navigate the menu with up and down, those keys are passed back to the body. I'm also noticing that pressing spacebar on the menu selects the item, but also passes that through as well.

ThomasBurleson:
@rschmukler - can you review. I will add unit tests if this is approved.

src/components/menu/menu-interim-element.js
```diff
@@ -141,9 +141,9 @@ function MenuProvider($$interimElementProvider) {
         opts.menuContentEl.on('keydown', function(ev) {
           scope.$apply(function() {
             switch (ev.keyCode) {
-              case $mdConstant.KEY_CODE.ESCAPE: opts.mdMenuCtrl.close(); break;
-              case $mdConstant.KEY_CODE.UP_ARROW: focusMenuItem(ev, opts.menuContentEl, opts, -1); break;
-              case $mdConstant.KEY_CODE.DOWN_ARROW: focusMenuItem(ev, opts.menuContentEl, opts, 1); break;
+              case $mdConstant.KEY_CODE.ESCAPE: opts.mdMenuCtrl.close(); ev.stopPropagation(); ev.preventDefault(); break;
+              case $mdConstant.KEY_CODE.UP_ARROW: focusMenuItem(ev, opts.menuContentEl, opts, -1); ev.stopPropagation(); break;
+              case $mdConstant.KEY_CODE.DOWN_ARROW: focusMenuItem(ev, opts.menuContentEl, opts, 1); ev.stopPropagation(); break;
```
jelbourn: Do up and down also not need preventDefault() ?

clshortfuse:
I think I misunderstood preventDefault. I incorrectly assumed preventDefault was related to browser actions but this isn't true. You don't need preventDefault() if you do stopPropagation().

PR #3861, opened 2015-07-22
Update input.js

### LINKED ISSUE #3730: textarea growTextArea() uses non-existent function
The following code generates an javascript exception:

https://github.com/angular/material/blob/master/src/components/input/input.js#L294

```
lineHeight = element.height();
```

height() function does not exist in jqLite and thus throws an exception.

This should probably be updated to 

```
lineHeight = element[0].offsetHeight;
```

dcgolden:
Fixes #3730 

 growTextArea for environments without jQuery.


dcgolden:
Signed it

ThomasBurleson:
Already fixed.

PR #3867, opened 2015-07-22
fix(gridList, toolbar): remove overly broad box-sizing in grid-list and toolbar

### LINKED ISSUE #3516: Overly broad box-sizing in grid-list and toolbar.
Both md-toolbar and md-gridlist have a rule like
```css
md-grid-list * {
  box-sizing: border-box;
}
```

This is a problem because these components are containers for arbitrary content. As such, any other components or existing content can be messed up by this overly broad style. The rules should only target elements that ng-material itself creates.

kennethcachia:
Fixes #3516


ThomasBurleson:
:+1:

PR #3868, opened 2015-07-23
fix(calendar): Update ngModel instead of ctrl.date on input change

mhchen:
Current behavior doesn't propagate user input to ng-model when a valid
date is entered.


mhchen:
@jelbourn not sure if this was the original intent but I think this minor change might make sense? It seems like it shouldn't affect much since the purpose of this is to influence the calendar date that pops up but if the datepicker is sharing an ng-model then it matters.

jelbourn:
LGTM. Yeah, this was the original intent of my TODO. Will merge shortly.

jelbourn:
Closed via 852cc1c834897f91955dbad0398aff945131b1a0

Issue #3885, opened 2015-07-24
md-autocomplete: unable to select item from dropdown if minLength is 0


stephen-james:
**Applies to:** v0.10.1-rc1, v0.10.1-rc3, others?...

**Problem:**
Given I have an autocomplete configured with a min length of 0 `md-min-length="0"`
When I focus on the input
And the dropdown list appears with all items
When I click on a suggestion item
Then the autocomplete does not select the item I just clicked.

Note: Keyboard selection works.

**Expected:**
Given I have an autocomplete configured with a min length of 0 `md-min-length="0"`
When I focus on the input
And the dropdown list appears with all items
When I click on a suggestion item
Then the autocomplete selects the item I just clicked.

**Notes:**
Investigation showed that MdAutocompleteCtrl.select was not firing

clshortfuse:
If you need a workaround, this is what I use:

searchTextChange: function (text) {
            if (!text) {
                //force query
                $scope.mySearchElement.searchText = '';
                return;
            }
            //do work
}

on the html: 

md-search-text-change="mySearchElement.searchTextChange(mySearchElement.searchText)"
md-search-text="mySearchElement.searchText"

stephen-james:
hey folks, please let me know if I can help by providing better example cases or more information

topherfangio:
I'm not seeing this issue on HEAD anymore. Can you please provide a Codepen showing the problem?

You can use the [Basic Usage Demo](https://material.angularjs.org/HEAD/#/demo/material.components.autocomplete) as a template; just click the "Edit on Codepen" button in the top right.

robertmesserle:
Closing for now - please respond with a CodePen demo if this issue still exists.

PR #3886, opened 2015-07-24
autocomplete: bugfix - ensures item selection doesn't trigger handleQuery() with autocomplete having minLength=0

stephen-james:
ensures item selection doesn't trigger handleQuery() with autocomplete having minLength=0

resolves issue #3885

The onMouseup handler is fired after the user selects an item from the drop down before the click event is fired to actually select the item.

Focus performs a query if the min length is zero, in doing so it potentially replaces all the items and destroys their event handlers so the event is never fired.

by placing the call to focus from onMouseUp on the nextTick, we eliminate this bug.

Normal focus remains untouched and will still perform initial search.


EladBezalel:
Closing this as you didn't replied on the issue, if it's still happening i will reopen

PR #3893, opened 2015-07-24
fix(sidenav): Fix width on mobile devices

NickBolles:
Remove the min-width on devices smaller then the default width so that the sidenav will be 84% of the screen instead of locked at $sidenav-default-width

If this isn't a bug Feel free to close this. But it sure seems like one to me.


EladBezalel:
Was fixed in one of my commits

Issue #3892, opened 2015-07-24
mdMenu sometimes blocks the entire screen


vprenner:
I managed to reproduce this on the Angular Material demo site as well (see screenshot). 

When clicking on the menu sometimes it creates a md-backdrop element upon opening. This element covers everything and you cannot click anything anymore.

I saw this bug was reported a few times ago as well. Seems it's still there.

I manage to reproduce it in the rc3 version as well as the head version.

Steps to reproduce: just keep clicking an icon (i was clicking on the width="4" one on the bottom of the demo page). Once the menu opens, close it by clicking outside the menu. Do this a few times and the UI will be blocked by the md-menu-backdrop.

On the Angular Material demo site it happens rarely. On the site I'm working on it happens quite often (in 50% of cases cca).

Testing was done on Chrome v44.

![md-menu-bug](https://cloud.githubusercontent.com/assets/2089751/8882214/8fc92e6a-3245-11e5-9b21-4b5718ee2ad3.png)

luclanet:
+1 Same problem for me

clshortfuse:
Does it have focus? If it doesn't, then it's the same issue as this one I posted. I thought it was Android specific, but it might just occur often on Android?

https://github.com/angular/material/issues/3888

There were recent changes with dealing with the open animation. Maybe it's getting stuck.

lcpereira:
Hello,

I'm having the same problem, the fix is already in the master?

Thank you.

![bug](https://cloud.githubusercontent.com/assets/10385917/8904820/112d5a0e-3439-11e5-89af-c8e2c8a5f3d0.gif)

stevenmiles:
Having the same issue with 0.10.1-rc3 with angular 1.4.3 in my application. Just tried it in the in the documentation app and manage to reproduce it. I find much it easier to reproduce it with IE11.

stefan-schweiger:
Relevant #3837

KReden:
Relevant - See my update to my closed issue https://github.com/angular/material/issues/3773

jozsi:
Still happening in 0.10.1-rc4

ThomasBurleson:
This will be fixed within the today and committed to master.

stefan-schweiger:
@ThomasBurleson but now everything with a ng-click on it (including md-menu) does not work in IE anymore #3943

vprenner:
@stefan-schweiger I don't think that's it because it is broken in RC4 as well.
[...]

PR #3899, opened 2015-07-25
Fix md-menu animation timeout issues

clshortfuse:
Fixes https://github.com/angular/material/issues/3892

### DIFF
--- a/src/components/menu/menu-interim-element.js
+++ b/src/components/menu/menu-interim-element.js
@@ -65,7 +65,7 @@ function MenuProvider($$interimElementProvider) {
       // Return the promise for when our menu is done animating in
       return animator
           .waitTransitionEnd(element, {timeout: 370})
-          .then( function(response) {
+          .finally( function(response) {
             opts.cleanupInteraction = activateInteraction();
             return response;
           });


ThomasBurleson:
@clshortfuse - The issue with the animation timeout is a conflict between ngAnimate and ngMaterial's animation solution. We are preparing a full solution all ngMaterial animation.

clshortfuse:
Are you sure this won't fix? I understand you're restructuring the animation system, but right now now md-menu does not work at all on devices that take longer than 370ms to animate (pretty much all mobile devices). 

I don't understand how this doesn't fix the issue.

When the timeout expires, it doesn't run the next piece of code because what equates to a single typo. It should call .finally instead of .then as it does with md-select.

stevenmiles:
@clshortfuse Interestingly enough, I has been suffering blocking with md-select in 0.10.1-rc2 until it was fixed in 0.10.rc3

https://github.com/angular/material/commit/171b7ed874b3a29749628159afebcfb14d38cad9

I've just noticed this happening for mdMenu.

@ThomasBurleson 

Seems you fixed the md-select locking, thank you. I'm guessing the animation solution won't be released in between a release candidate and a release. This would a good short term fix until the animation solution appears, as was the case for md-select locking issue.

KReden:
@ThomasBurleson Is there a time frame for this animation solution? If it is a ways off it might make sense to push a hotfix for this in the meantime, otherwise md-menu is pretty much broken on mobile and slower machines.

ThomasBurleson:
@clshortfuse, @KReden, @stevenmiles - here is an interim fix to transitionEnd event timeouts [due to ngAnimate 1.4.x conflicts ]:  https://github.com/angular/material/commit/ce46a9a4ca49d88f0b58de5ebeada09971034682

> Btw, In our code base, using `finally()` is not a great solution for two (2) reasons:

*  The `finally()` handler function is not passed the result or error from the promise reject/resolve.
*  The `finally()` will not allow you to modify the value propagating. So rejections will continue to propagate.

clshortfuse:
Thanks for the other workaround! It makes much more sense to never reject. I was aware of the fact finally would have its issues only after, when I learned about the then(successCallback, errorCallback) option.

ThomasBurleson:
@clshortfuse - :+1: , `finally()` is a tricky one.

PR #3914, opened 2015-07-27
fix(autocomplete): prevent scroll lock on navigation

### LINKED ISSUE #3912: md-autocomplete locks page scroll on navigation
When utilizing the ```md-selected-item-change``` attribute in ```md-autocomplete``` to trigger page navigation (a user clicks an item in the autocomplete dropdown and is navigated to another page), the scroll mask and body styles (position:fixed, etc) are never removed.

4storia:
Check only for ctrl.hidden being set to false, rather than all falsey values

Closes #3912

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -199,7 +199,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     if (!hidden && oldHidden) {
       positionDropdown();
       if (elements) $mdUtil.nextTick(function () { $mdUtil.disableScrollAround(elements.ul); }, false);
-    } else if (hidden && !oldHidden) {
+    } else if (hidden !== false && !oldHidden) {
       $mdUtil.nextTick(function () { $mdUtil.enableScrolling(); }, false);
     }
   }


4storia:
I signed it!

ThomasBurleson:
@robertmesserle - can you review?

robertmesserle:
This should be a falsey check and not an explicit check - it seems like something else is the root cause of the issue.  (ie. this fixes the symptom, but not the problem)

I'm going to trace this bug to see the root cause, but I suspect it has to do with needing to unlock scrolling in the cleanup method.

robertmesserle:
@4storia Instead of this change, can you try adding the following inside of `cleanup()` in `autocompleteController`:

```
$mdUtil.nextTick(function () { $mdUtil.enableScrolling(); }, false);
```

And let me know if that fixes your problem.  It would probably be easier for you to confirm this since there is no codepen for me to confirm against.

4storia:
Adding that line doesn't seem to fix the problem for me.
Let me add some context:
In the function I provide to ```md-selected-item-change``` looks something like this:
```
function itemChange(item) {
    $state.go('someRoute', {param: item.id});
}
```

I suspect that this does not trigger the $destroy event, so ```cleanup``` is never actually called.
I changed the ```md-selected-item-change``` function to something like this:
```
function itemChange(item) {
    $timeout(function() {
        $state.go('someRoute', {param: item.id});
    });
}
```

This solves the issue, as (I'm assuming) mdUtil ticks before the timeout and closes the dropdown (which in turn removes the scroll lock). I'd prefer not to have to wrap my $state.go in a timeout though :D 

I do agree that the code change to check specifically for false makes things unnecessarily complicated. I'll explore some alternatives that make more sense, and try to get a codepen up to help demonstrate the problem.

robertmesserle:
Hm, so it may make sense to simple wrap our call to selectedItemChange in a
`nextTick` wrapper changing location doesn't cause the destroy event.

4storia:
I just saw this now (I guess github doesn't send new emails when comments are edited) but the solution you initially proposed (```$mdUtil.enableScrolling();```) actually works just fine (I missed a build step in my project and got a false negative initially).

Going to close this since you've got the better solution :D

PR #3937, opened 2015-07-28
fix(select): fix incorrect styling of md-select

robteix:
Change style rule to correctly colour a md-select when its value
is invalid. A rule had been left over with md-select-label instead of the
new md-select-value. This trivial patch fixes it.


ThomasBurleson:
:+1:

PR #3946, opened 2015-07-29
fix(dialog, menu, select, interimElement): use $animateCss

### LINKED ISSUE #3527: Angular 1.4.* & 1.3.16 breaks md-menu enter animation
Since angular 1.4 and 1.3.16 md-menu grow animation does not work. I've checked the change log and found that commit:
https://github.com/angular/angular.js/commit/0adc0364265b06c567ccc8e90a7f09cc46f235b2

Hope that someone can look at it.

### LINKED ISSUE #3640: md-toast no longer animating with angular-animate 1.4.x
When using angular-animate 1.4.x, the animation of the toast moving is no longer happening.

Here's an example of it work with 1.3.15:
http://codepen.io/ryanespin/pen/mJxZNJ

Here's one of it not working with 1.4.2:
http://codepen.io/ryanespin/pen/domBaE

### LINKED ISSUE #3653: <md-tabs>  no longer animating with angular-animate 1.4.x
Here's an example of it work with 1.3.15:
http://codepen.io/anon/pen/wajGgE

Here's one of it not working with 1.4.2:
http://codepen.io/anon/pen/bdMpgm

Something like with **md-toast** animatig https://github.com/angular/material/issues/3640

### LINKED ISSUE #3773: mdMenu: menu not closing on firefox 39 and chrome
Reference Issue: https://github.com/angular/material/issues/3438

This was reported as fixed in master several days ago but as of right now on FF and chrome it is broken again. 

When the menu opens it seems to kill all input for the page. A refresh is needed to recover

### LINKED ISSUE #3837: md-menu: completly broken in IE (current HEAD + 0.10.1-rc2)
After you open the menu in IE (11 + win8) you can't do anything else, the whole site gets blocked by a div and you can't select anything.

Also the alignment is completely messed up.

(You can see it on your [demo page](https://material.angularjs.org/HEAD/#/demo/material.components.menu))

menu

EDIT: Hit enter a bit prematurely

### LINKED ISSUE #3919: md-dialog no longer works with $q.all()
I have created a Dialog service which basically pops-up a "processing" md-dialog while a data service (or other promise or promises) runs.  As soon as the promise (or promises) returns, the processing dialog is supposed to hide automatically.  The Dialog service works fine with Material v 0.10.  It breaks with Material v 0.10.1-rc3 and beyond.

A Codepen of my Dialog service with methods to invoke the main functions is located here: http://codepen.io/JDGoldman/pen/ZGqjdY

The Codepen uses Material v0.10. If I try to change this to v0.10.1-rc4, there is a problem with the dialog controller when using a $q.all() syntax (you can tell because the {{ }} show up instead of text).  If you have only one promise/function, the dialog service appears to work normally.  If I pass my DialogService in to the $q.all() call as one of the parameters, the dialog service finds the controller, but the dialog never closes (since, of course, it is now waiting for itself to close).  

I assume that the newer versions of Material were not supposed to break my Dialog service.  I'm not sure exactly where this problem cropped up, but it definitely occurred after v0.10.  If I had to guess, perhaps it is caused by eliminating the dialogTransitionEnd function and $mdUtil.transitionEndPromise(dialogEl) calls, all of which returned a promise.   

Thanks for reading.  Great work on this project.  Angular Material is incredible.

ThomasBurleson:
replace programmatic use of element.css for style changes and use of transitionEnd event listeners with use of ngAnimate's $animateCss; use polyfill for Angular <1.4.

$mdUtil.dom.animator.translate3d() uses $animateCss() instead of waitTransitionEnd()
use animateCss.js polyfill for 'material.animate' module
- add mock `createMockStyleSheet` for animateCss tests
refactors to menu-interim-element.js and select.js
- refactor logic and patterns used
- use $animateCss in place of waitTransitionEnd() 
debounce Select and Menu window resize handlers
Dialog uses same showBackdrop/hideBackdrop pattern as Menu and Select
select async demo no longer clears users list when reloading
select demos use `md-input-container { margin-right: 10px;}`
hide Select Backdrop with zero duration
enable full click detection in select-value area by using background color (with zero alpa).

BREAKING-CHANGES:  select and backdrop styles added

*  select list text is not selectable, 
*  select backdrop hide duration is 0ms
*  select text value background has zero alpha

```scss
.md-text {  @include not-selectable(); }

.md-select-value {
  background-color: rgba(0,0,0,0);
}

md-backdrop {
  &.md-select-backdrop {
      transition-duration: 0ms;
  }  
}
```

Fixes #3919. Fixes #3837. Fixes #3773, Fixes #3640. Fixes #3527. Fixes #3653.


ThomasBurleson:
@jelbourn, @robertmesserle, @rschmukler - this is ready for your reviews and testing. Use of `$animateCss()` works for Angular v1.3.x and v1.4.x (using a polyfill). Code refactors present logic onShow/onRemove logic in clear, concise fashion(s).

> Once ready/approved, this will be merged into master for the v0.11 builds

ThomasBurleson:
@robertmesserle - this PR actually has 2 commits... one that is already in master SHA b71fdfb609a9b8c5bf6e2f4b41638020e265acd2. Can you help me remove this commit and rebase properly?

src/components/backdrop/backdrop.js
```diff
@@ -28,17 +28,28 @@ angular
       };
 
     function postLink(scope, element, attrs) {
+
+        // If body scrolling has been disabled using mdUtil.disableBodyScroll(),
+        // adjust the 'backdrop' height to account for the fixed 'body' top offset
+        var body = $window.getComputedStyle($document[0].body);
```
jelbourn: `body` => `bodyStyles` ?

src/components/backdrop/backdrop.js
```diff
@@ -28,17 +28,28 @@ angular
...
+        if ( body.position == 'fixed') {
+          var hViewport = parseInt(body.height,10) + Math.abs(parseInt(body.top,10));
```
jelbourn: Space after commas

src/components/dialog/dialog.js
```diff
@@ -619,9 +620,10 @@ function MdDialogProvider($$interimElementProvider) {
         }
         if (options.disableParentScroll) {
           options.restoreScroll();
+          delete options.restoreScroll;
         }
 
-        options.hideBackdrop = null;
+        delete options.hideBackdrop;
```
jelbourn: Can you add a comment saying why you are using `delete` instead of setting to null? (since the latter is generally much faster)

src/components/dialog/dialog.js
```diff
@@ -702,18 +704,38 @@ function MdDialogProvider($$interimElementProvider) {
     }
 
     /**
+     * Ensure the dialog container fill-stretches to the viewport
+     */
+    function adjustDialogContainer(container, options) {
```
jelbourn: Just now realizing that this PR looks like it needs to be rebased.

PR #4002, opened 2015-08-03
feat(virtualRepeat): Infinite scroll and deferred data loading

kseamon:
Includes demos for both infinite scroll and deferred loading.
Could be enhanced further in the future with a loading spinner of some sort.

### DIFF
--- a/src/components/virtualRepeat/demoDeferredLoading/index.html
+++ b/src/components/virtualRepeat/demoDeferredLoading/index.html
@@ -1,0 +1,22 @@
+<div ng-controller="AppCtrl as ctrl">
+  <md-content layout="column">
+    <p>
+       Display a list of 50,000 items that load on demand in a viewport of only 7 rows (height=40px).
+       <br/><br/>
+       This demo shows scroll and rendering performance gains when using <code>md-virtual-repeat</code>;
+       achieved with the dynamic reuse of rows visible in the viewport area. Developers are required to
+       explicitly use <code>md-virtual-repeat-container</code> as a wrapping parent container.
+       <br/><br/>
+       To enable load-on-demand behavior, developers must pass in a custom instance of
+       mdVirtualRepeatModel (see the example's source for more info).
+    </p>
+
+    <md-virtual-repeat-container id="vertical-container">
+      <div md-virtual-repeat="item in ctrl.dynamicItems" md-on-demand
+          class="repeated-item" flex>
+        {{item}}
+      </div>
+    </md-virtual-repeat-container>
+  </md-content>
+
+</div>
--- a/src/components/virtualRepeat/demoDeferredLoading/script.js
+++ b/src/components/virtualRepeat/demoDeferredLoading/script.js
@@ -1,0 +1,70 @@
+(function () {
+  'use strict';
+
+    angular
+      .module('virtualRepeatDeferredLoadingDemo', ['ngMaterial'])
+      .controller('AppCtrl', function($timeout) {
+
+        // In this example, we set up our model using a class.
+        // Using a plain object works too. All that matters
+        // is that we implement getItemAtIndex and getLength.
+        var DynamicItems = function() {
+          /**
+           * @type {!Object<?Array>} Data pages, keyed by page number (0-index).
+           */
+          this.loadedPages = {};
+
+          /** @type {number} Total number of items. */
+          this.numItems = 0;
+
+          /** @const {number} Number of items to fetch per request. */
+          this.PAGE_SIZE = 50;
+
+          this.fetchNumItems_();
+        };
+
+        // Required.
+        DynamicItems.prototype.getItemAtIndex = function(index) {
+          var pageNumber = Math.floor(index / this.PAGE_SIZE);
+          var page = this.loadedPages[pageNumber];
+
+          if (page) {
+            return page[index % this.PAGE_SIZE];
+          } else if (page !== null) {
+            this.fetchPage_(pageNumber);
+          }
+        };
+
+        // Required.
+        DynamicItems.prototype.getLength = function() {
+          return this.numItems;
+        };
+
+        DynamicItems.prototype.fetchPage_ = function(pageNumber) {
+          // Set the page to null so we know it is already being fetched.
+          this.loadedPages[pageNumber] = null;
+
+          // For demo purposes, we simulate loading more items with a timed
+          // promise. In real code, this function would likely contain an
+          // $http request.
+          $timeout(angular.noop, 300).then(angular.bind(this, function() {
+            this.loadedPages[pageNumber] = [];
+            var pageOffset = pageNumber * this.PAGE_SIZE;
+            for (var i = pageOffset; i < pageOffset + this.PAGE_SIZE; i++) {
+              this.loadedPages[pageNumber].push(i);
+            }
+          }));
+        };
+
+        DynamicItems.prototype.fetchNumItems_ = function() {
+          // For demo purposes, we simulate loading the item count with a timed
+          // promise. In real code, this function would likely contain an
+          // $http request.
+          $timeout(angular.noop, 300).then(angular.bind(this, function() {
+            this.numItems = 50000;
+          }));
+        };
+        
+        this.dynamicItems = new DynamicItems();
+      });
+})();
--- a/src/components/virtualRepeat/virtualRepeater.js
+++ b/src/components/virtualRepeat/virtualRepeater.js
@@ -311,6 +311,15 @@ VirtualRepeatContainerController.prototype.handleScroll_ = function() {
  * @param {string=} md-extra-name Evaluates to an additional name to which
  *     the current iterated item can be assigned on the repeated scope. (Needed
  *     for use in md-autocomplete).
+ * @param {boolean=} md-on-demand When present, treats the md-virtual-repeat argument
+ *     as an object that can fetch rows rather than an array. This object must implement
+ *     the following two methods:
+ *     getItemAtIndex: function(index) -> item at that index or null if it is not yet
+ *         loaded (It should start downloading the item in that case).
+ *     getLength: function() -> number The data legnth to which the repeater container
+ *         should be sized. Ideally, when the count is known, this method should return it.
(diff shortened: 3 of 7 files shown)


src/components/virtualRepeat/demoDeferredLoading/script.js
```diff
@@ -0,0 +1,74 @@
...
+
+    angular
+      .module('virtualRepeatDeferredLoadingDemo', ['ngMaterial'])
+      .controller('AppCtrl', function($timeout, mdVirtualRepeatModel) {
+
+        // In this example, we set up our model using prototypal inheritance.
+        // Using a mixin works too. All that matters
+        // is that we inherit from mdVirtualRepeatModel and implement
+        // getItemAtIndex and getLength.
+        this.infiniteItems = Object.create(mdVirtualRepeatModel.prototype);
```
gkalpak: Is this needed here ?

src/components/virtualRepeat/demoDeferredLoading/index.html
```diff
@@ -0,0 +1,22 @@
...
+  <md-content layout="column">
+    <p>
+       Display a list of 50,000 items that load on demand in a viewport of only 7 rows (height=40px).
+       <br/><br/>
+       This demo shows scroll and rendering performance gains when using <code>md-virtual-repeat</code>;
+       achieved with the dynamic reuse of rows visible in the viewport area. Developers are required to
+       explicitly use <code>md-virtual-repeat-container</code> as a wrapping parent container.
+       <br/><br/>
+       To enable load-on-demand behavior, developers must pass in a custom instance of
+       mdVirtualRepeatModel (see the example's source for more info).
```
jelbourn: I'm thinking it would be easier for the user to instead explicitly set some attribute when they want infinite scrolling and then check that the model provided has the appropriate methods. Something like:
```html
<div md-virtual-repeat="item in ctrl.dynamicItems" md-infinite>
  {{item}}
</div>
```

I think most angular users would rather specify an attribute than extend a base class.

src/components/virtualRepeat/virtualRepeater.js
```diff
@@ -423,6 +428,19 @@ VirtualRepeatController.prototype.readItemSize_ = function() {
 };
 
 
+VirtualRepeatController.prototype.repeatListExpression_ = function(scope) {
```
jelbourn: Function description?

src/components/virtualRepeat/demoDeferredLoading/index.html
kseamon: @ThomasBurleson Thoughts? That would be less Google JSish, but might also result in less useful errors if they don't implement the interface methods (unless I manually check and throw exceptions).

jelbourn: I'd say to manually check and throw.

kseamon: Also, md-infinite isn't quite right... infinite scroll is one of two behaviors unlocked by this feature (hence the two demos).

md-deferred? md-deferred-data? md-deferred-load?

jelbourn: `md-on-demand` ?

kseamon: Done

src/components/virtualRepeat/virtualRepeater.js
```diff
@@ -665,6 +702,28 @@ VirtualRepeatController.prototype.updateIndexes_ = function() {
...
-  this.newEndIndex = Math.min(itemsLength, this.newStartIndex + containerLength + NUM_EXTRA);
+  this.newVisibleEnd = this.newStartIndex + containerLength + NUM_EXTRA;
+  this.newEndIndex = Math.min(itemsLength, this.newVisibleEnd);
   this.newStartIndex = Math.max(0, this.newStartIndex - NUM_EXTRA);
 };
+
+
+function VirtualRepeatModelArrayLike(model) {
+  this.model = model;
+  this.length = model.getLength();
```
jelbourn: Check here whether `model` has the required functions?

kseamon: Hmm... I'd rather not have to do these checks on each digest cycle, but for full correctness I'd have to.

jelbourn:
LGTM for the code itself, would be good to hear Thomas' thoughts on the attribute name.

src/components/virtualRepeat/virtualRepeater.js
jelbourn: I think a one-time check is sufficient for now.

src/components/virtualRepeat/demoDeferredLoading/index.html
ThomasBurleson: @jelbourn, @kseamon - I like the `md-on-demand` attribute and the use of the attribute instead of the model subclassing. Currently issues:

*  I do not think we are clearly articulating the [expected model interface methods](https://github.com/angular/material/pull/4002/files#diff-30883b316af6f89bca301e70a4243fbfR10) when `onDemand` is enabled; the comment is very easy to miss.
  * 
  *  Should this information also be in the source docs?
*  If the interface is not correct, I think we should throw an Error **or** at least warn the developer in the console.

src/components/virtualRepeat/demoDeferredLoading/script.js
```diff
@@ -0,0 +1,70 @@
...
+          // For demo purposes, we simulate loading more items with a timed
+          // promise. In real code, this function would likely contain an
+          // $http request.
+          $timeout(angular.noop, 300).then(angular.bind(this, function() {
+            this.loadedPages[pageNumber] = [];
+            var pageOffset = pageNumber * this.PAGE_SIZE;
+            for (var i = pageOffset; i < pageOffset + this.PAGE_SIZE; i++) {
+              this.loadedPages[pageNumber].push(i);
+            }
+          }));
```
ThomasBurleson: In my experience the following is always a better approach:

```js
var buffer = [], pageOffset = pageNumber * this.PAGE_SIZE;

for (var i = pageOffset; i < pageOffset + this.PAGE_SIZE; i++) {
    buffer.push(i);
}

this.loadedPages[pageNumber] = buffer;
```

ThomasBurleson:
@kseamon - really cool feature here.

src/components/virtualRepeat/demoDeferredLoading/script.js
kseamon: Better how? Remember this is just a demo script.

src/components/virtualRepeat/demoDeferredLoading/index.html
kseamon: The interface is enforced by exceptions and is documented in the virtual repeat source code.

kseamon:
What's blocking merging of this PR at this point?

jelbourn:
@ThomasBurleson added the "needs work" label, but I don't know what he had in mind.

ThomasBurleson:
@kseamon, @jelbourn - On the local source (with this PR merged),  I just updated the source with additional comments to clearly articulate the use of the required interface:

```js
/**
 * This VirtualRepeatModelArrayLike class enforces the interface requirements
 * for infinite scrolling within a mdVirtualRepeatContainer. An object with this
 * interface must implement the following interface with two (2) methods:
 *
 * getItemAtIndex: function(index) -> item at that index or null if it is not yet
 *     loaded (It should start downloading the item in that case).
 *
 * getLength: function() -> number The data legnth to which the repeater container
 *     should be sized. Ideally, when the count is known, this method should return it.
 *     Otherwise, return a higher number than the currently loaded items to produce an
 *     infinite-scroll behavior.
 *
 * @usage
 * <hljs lang="html">
 *  <md-virtual-repeat-container md-orient-horizontal>
 *    <div md-virtual-repeat="i in items" md-on-demand>
 *      Hello {{i}}!
 *    </div>
 *  </md-virtual-repeat-container>
 * </hljs>
 *
 */
function VirtualRepeatModelArrayLike(model) {
  if (!angular.isFunction(model.getItemAtIndex) ||
      !angular.isFunction(model.getLength)) {
    throw Error('Object passed to md-virtual-repeat must implement functions getItemAtIndex() ' +
        'and getLength() when md-on-demand is enabled');
  }

  this.model = model;
}
```

ThomasBurleson:
I am merging now.

RonB:
@ThomasBurleson or @kseamon 

Maybe not the best place to ask, but I'm exited about this functionality, just having diffculty to understand the api. I'm confused on how this VirtualRepeatModel should work. I have defined a class using a factory like so:

    angular.module('app' )
            .factory(
                    'Store',
                    function ($http) {

                        function Store(options) {
                            // params for the api-call are passed thru the options parameter
                            this.params = options || {};
                            // limit page size
                            this.params.limit = this.params.limit || 100;
                            // start at offset
                            this.offset = 0;
                            // data array ???
                            this.data = [];
                            // keymap
                            this.keys = {};
                            // init first page
                            this.fetch_(0);
                        }

                        // Required.
                        Store.prototype.getItemAtIndex = function (index) {
                            if (this.data[index]) {
                                return this.data.index;
                            } else {
                                this.fetch_(index);
                            }
                        };
                        
                        // Required.
                        Store.prototype.getLength = function () {
                            return this.data.length;
                        };
                        // fetch a page with data
                        Store.prototype.fetch_ = function (offset) {
                            this.params.offset = offset || 0;
                            var url = '/api/' + this.params.schema + '/search';
                            $http({
                                url: url,
                                method: 'get',
                                params: this.params
                            }).then(angular.bind(this, function (response) {
                                  this.offset += response.data.rows.length;
                                  angular.forEach(response.data.rows, function (row) {
                                      this.data.push(row);
                                      this.keys[row.docunid] = this.data[this.data.length-1];
                                  }, this);
                            }));
                        };
                        return Store;
                    });


In my controller:
    
     employees = new Store({schema: 'employee'});

I want to iterate over the data array, but in your example you seem to have the data stored in the loadedPages object, which is not used as such in the markup. I cannot use emplyee in employee because the employees are stored in the "data" array of the store. What am I missing here?

PR #4010, opened 2015-08-04
feat(datepicker): Debounce datepicker input element event

mhchen:


mhchen:
@jelbourn 

I'm still not 100% sure on the debounce interval. 100 normally seems right but dates are pretty hard to type so I made it 150. 200 feels a bit laggy. Also I didn't debounce the input resize event as I don't think that should be debounced. Let me know if you have any thoughts please!

src/components/datepicker/datePicker.js
```diff
@@ -274,6 +263,29 @@
...
   };
 
+  /**
+   * Resizes the input element based on the size of its content.
+   */
+  DatePickerCtrl.prototype.resizeInputElement = function() {
+    this.inputElement.size = this.inputElement.value.length + EXTRA_INPUT_SIZE;
+  };
+
+  DatePickerCtrl.prototype.handleInputEvent = function() {
```
jelbourn: Can you add a function description here that says that the handling entails?

src/components/datepicker/datePicker.js
```diff
@@ -207,22 +210,8 @@
...
-        self.ngModelCtrl.$setViewValue(parsedDate);
-        self.inputContainer.classList.remove(INVALID_CLASS);
-        self.$scope.$apply();
-      } else {
-        // If there's an input string, it's an invalid date.
-        self.inputContainer.classList.toggle(INVALID_CLASS, inputString);
-      }
-    });
+    self.inputElement.addEventListener('input', angular.bind(self, self.resizeInputElement));
+    self.inputElement.addEventListener('input', self.$mdUtil.debounce(self.handleInputEvent, INPUT_DEBOUNCE_INTERVAL, self));
```
jelbourn: Can you break this line to stay under 100 columns?

jelbourn:
LGTM aside from those two comments. Good thinking on the resize.

mhchen:
@jelbourn rebased and pushed

jelbourn:
I pulled in the change and Rob and I messed around with it. At first the 150ms felt too fast for when you're typing a valid date. We tried out a few different times before realizing that the focus state of the input is missing (should be a blue underline, like md-input) and that should override the invalid state, which makes this much more straightforward. What do you think?

mhchen:
Agreed, I was thinking about this as I implemented the invalid state. The focus has a bit of a wider border in addition to being the primary color too. Do you want me to do this in this PR or create a separate one?

jelbourn:
Whatever is easier for you

mhchen:
@jelbourn added the theming for the focus state. Not sure what the guidance is on using ng-focus/blur vs. element.on to attach focus/blur events. Let me know if you'd prefer the other way.

jelbourn:
When the datepicker is focused, the focus style should just always override the invalid style, so that the latter only shows up when you focus another element. Other than that LGTM.

jelbourn:
Closed via fae0495590d5cdfce002d3da56231dd4cf99d538, 84db12bced787c660a5624adaca0c2d2a7718879, 1bc79cc6f9d75a5db1c635f2d25a3394de1db47b

PR #4013, opened 2015-08-04
fix(dialog): md-actions left padding has to be same as dialog

dciccale:
screen shot 2015-08-04 at 14 37 01

the left padding of the actions container has to be 24, the same as the padding in the dialog content


EladBezalel:
I think i fixed it in one of my commits, can you check in HEAD if it's still necessary? 
closing for now

PR #4045, opened 2015-08-06
fix(dialog): use foreground-1 text color

remcohaszing:
This makes the dialog content better readable when using a dark theme.


EladBezalel:
Please check against HEAD if it's still an issue,
closing for now, if this pr is needed, please rebase

PR #4054, opened 2015-08-06
fix(fabSpeedDial/Toolbar): Refactor default click/hover behavior.

### LINKED ISSUE #3214: Speed Dial & FAB Toolbar should appear only on clicking the FAB
According to the Material Design spec by Google, the FAB can transform into a toolbar upon press and it can fling out the speed dial also upon press. 
Hovering is not apt for a button. A button is meant to be pressed, not hovered.  And also when we hover away from it, the toolbar/speed dial disappears. This is not desirable. 

It is also given that, the FAB should transform into a 'most common task' or a 'close/back' button, in case of speed dial.

And, the FAB toolbar appears if I simply hover the area where it would appear, making it impossible to access other background contents in that area.

Reference: http://www.google.com/design/spec/components/buttons-floating-action-button.html

### LINKED ISSUE #3925: FAB Speed Dial iOS issue
I'm trying to use the FAB Speed Dial (here the references: https://material.angularjs.org/0.10.1-rc4/#/demo/material.components.fabSpeedDial) but i found a issue using an ipad. In every Android device works well and in every web browser too. When i try with iOS the FAB open but not close. I can't close it until tap on an action of one of the mini-fab. Is it possible solve it?

topherfangio:
Multiple users have requested a configurable way to open the speed dial
and toolbar on hover instead of it being automatic. This PR greatly simplifies
the existing code to be more standard.

BREAKING CHANGE: The FAB Speed Dial/Toolbar no longer automatically
adds the hover effect. The demo has been updated with example code
which utilizes `ng-mouseenter` and `ng-mouseleave` instead.

BREAKING CHANGE: The FAB Speed Dial/Toolbar no longer automatically
opens when receiving keyboard focus via the tab key. You must now
activate the trigger to open the component. This allows you to
skip all of the children if you are attempting to tab to another
element.

This also provides better support for iOS devices (and presumably
other mobile devices too).

fixes #3925, fixes #3214

### DIFF
--- a/src/components/fabActions/fabActions.js
+++ b/src/components/fabActions/fabActions.js
@@ -47,14 +47,13 @@
           // Grab whichever parent controller is used
           var controller = controllers[0] || controllers[1];
 
-          // Make the children open/close their parent directive
+          // Make the children close their parent directive
           if (controller) {
             angular.forEach(element.children(), function(child) {
-              // Attach listeners to the `md-fab-action-item`
+              // Attach listeners to the child of each `md-fab-action-item`
               child = angular.element(child).children()[0];
 
-              angular.element(child).on('focus', controller.open);
-              angular.element(child).on('blur', controller.close);
+              angular.element(child).on('click', controller.close);
             });
           }
         }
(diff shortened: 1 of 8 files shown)


topherfangio:
Closing this in favor of a more complete branch which adds ARIA and Keyboard Support. Will do a PR tomorrow after talking with the team.

PR #4064, opened 2015-08-08
fix(font): use Roboto font instead of RobotoDraft

### LINKED ISSUE #4063: fix(RobotoDraft): drop the Draft. Just Roboto. It's cleaner.
From what I understand, RobotoDraft was the temporary font while the Roboto 2.0 was being prepared for release, which has now happened (see the update here: https://www.google.com/fonts/specimen/Roboto). I think it's safe and recommended to remove "RobotoDraft" from css rules that have it as the primary choice. There have been internal UI issues (b/20920031) reported with using the deprecated RobotoDraft font specifically with Angular Material.

jeffbcross:
RobotoDraft was temporary until the new version of Roboto was
released to Google Web Fonts. Now that Roboto has been updated
as of January 2015, RobotoDraft is considered deprecated.

Closes #4063

### DIFF
--- a/docs/config/template/index.template.html
+++ b/docs/config/template/index.template.html
@@ -7,7 +7,7 @@
 </title>
 <link rel="icon" type="image/x-icon" href="favicon.ico" />
 <meta name="viewport" content="initial-scale=1" />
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
Thx @jeffbcross

PR #4090, opened 2015-08-10
fix(chips): hide remove button on custom template when readonly

### LINKED ISSUE #3697: Readonly chips allows deletion with custom template
https://material.angularjs.org/HEAD/#/demo/material.components.chips

Check the Readonly checkbox under Basic Usage. The custom template example allows the items to be deleted.

typotter:
Add ng-if to the remove button, conditional on not readonly. Also tweak the custom input styling in the demo, and add the 'readonly' class to md-chip when it is readonly.

fixes #3697

### DIFF
--- a/src/components/chips/demoBasicUsage/style.scss
+++ b/src/components/chips/demoBasicUsage/style.scss
@@ -1,7 +1,6 @@
 .custom-chips {
   .md-chip {
     position: relative;
-    padding-right: 35px;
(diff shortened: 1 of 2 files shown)


typotter:
Will need to merge this after PR#4114 goes in, but @jelbourn, care to take a look?

jelbourn:
LGTM

Issue #3426, opened 2015-06-23
Palette contrast ratios don't match Material spec


calebegg:
https://github.com/angular/material/blob/master/src/core/services/theming/theme.palette.js

For instance, 'light-blue' has 500 under `contrastLightColor`, but it should have black text according to the spec:

http://www.google.com/design/spec/style/color.html#color-color-palette

jelbourn:
Looks like the spec has been updated since those swatches were pulled. I have a separate list of the color / contrast color that I pulled a while ago with a script and it has `white` for light-blue-500.

jelbourn:
I think this should be fixed now. Chime in if you notice any new mismatches.

PR #4103, opened 2015-08-11
Fix(theming): Update contrast colors to match Material Spec

justingrayston:
Contrast colors seemed out of date with
http://www.google.com/design/spec/style/color.html#color-color-palette

Relates to #3426


justingrayston:
I signed it!

justingrayston:
This came up on an internal accessibility review on a project I am working on, and I noted they were out of date.

I have updated the contrast colors, however I note contrastStrongLightColors which isn't in the docs, and I am not sure what it refers to or where it is referenced in the spec. I was working from https://www.google.co.uk/design/spec/style/color.html#color-ui-color-application

If someone can let me know I will update this PR.

PR #4108, opened 2015-08-11
fix(chips): interaction with Autocomplete

### LINKED ISSUE #3475: Chips have blank template when search text not found in items collection
When I check Chips in the demode, Chips work quite well when search text not found in collection. And I don't have any idea why I am having this problem. I am using Angularjs 1.4.0 and other dependencies. 

screen shot 2015-06-26 at 3 48 57 pm

typotter:
Properly set require-match=true in demos so that the value is correct in the components.
Check for requiredMatch on the controller instead of on the scope.

fixes #3475

### DIFF
--- a/src/components/chips/demoContactChips/index.html
+++ b/src/components/chips/demoContactChips/index.html
@@ -7,7 +7,7 @@
         md-contact-name="name"
         md-contact-image="image"
         md-contact-email="email"
-        md-require-match
+        md-require-match="true"
         filter-selected="ctrl.filterSelected"
         placeholder="To">
     </md-contact-chips>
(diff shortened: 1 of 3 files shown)


houmark:
@robertmesserle Now that you are on a PR streak, maybe this one could make it as well? :)

houmark:
Thanks a bunch. I tested and it all works well, **as long** as you make sure to get `md-require-match="true"` set.

robertmesserle:
Yea, this is a known issue that I plan to look into (hopefully tomorrow).

houmark:
It might be my lack of understanding all the potential options and combinations here, but it's still weird behavior for me that not setting md-require-match makes it choose a more or less "random" auto complete option when I hit ENTER, while TAB always works. But this is at least fixing the very worst bug for me, the rest seems more like cleanup and refactoring to get more consistency.

PR #4113, opened 2015-08-11
fix(mdChips): Run duplicate check after mdOnAppend.

### LINKED ISSUE #2748: md-chips checks for duplicates before md-on-append is called
The appendChip function in chipsController.js does some duplicate-checking to prevent adding the same chip multiple times.  Unfortunately, it checks for matches before md-on-append is called.  If md-on-append converts a string input into an object, for instance, the duplicate checking will always fail.  

Presumably it should either check for duplicates after md-on-append has been called, or keep a separate collection of unmodified chips to check against.

topherfangio:
The duplicate check was firing before the user's onAppend
method had a chance to run, so the check would always
compare a string against an object and fail.

fixes #2748

### DIFF
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -135,6 +135,30 @@ describe('<md-chips>', function() {
         expect(scope.items[3].name).toBe('Grape');
         expect(scope.items[3].uppername).toBe('GRAPE');
       });
+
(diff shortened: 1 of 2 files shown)


topherfangio:
@typotter ready for review :-)

typotter:
LGTM

PR #4114, opened 2015-08-11
fix(chips): cursor style on static chips

### LINKED ISSUE #3253: Static chips should not have cursor `text`
Since they are static, they can't be edited or even selected, so they should have cursor `default`, not `text`.
This can be seen here:
When hovering between chips, a `text` cursor appears.

typotter:
Do not set the cursor style on the md-chips element if there are only static chips.
Also remove some TODO docs that have already been done.

fixes #3253

### DIFF
--- a/src/components/chips/chips.scss
+++ b/src/components/chips/chips.scss
@@ -55,7 +55,10 @@ $contact-chip-name-width: rem(12) !default;
   font-size: $chip-font-size;
   padding: $chip-wrap-padding;
   vertical-align: middle;
-  cursor: text;
+
+  &:not('.md-readonly') {
+    cursor: text;
+  }
   .md-chip {
     cursor: default;
     border-radius: $chip-height / 2;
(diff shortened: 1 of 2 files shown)


typotter:
@jelbourn care to take a look?

src/components/chips/chips.scss
```diff
@@ -55,7 +55,10 @@ $contact-chip-name-width: rem(12) !default;
   font-size: $chip-font-size;
   padding: $chip-wrap-padding;
   vertical-align: middle;
-  cursor: text;
+
+  &:not('.static-only') {
```
jelbourn: Should be `md-static`only`

Also, what do you think of instead naming it `md-readonly` ?

typotter: actually readonly is a good idea. I have another PR that should use this style as well.

jelbourn:
LGTM

PR #4116, opened 2015-08-12
fix(tabs): update inkBar to not show when tab does not exist

oste:
When setting the slectedIndex to -1 the inkBar will still be visible below the last selected tab. This commit fixes that by setting the inkBar css left property to 'auto' thus hiding it.


oste:
I signed it!

ThomasBurleson:
@robertmesserle - Can you review this?

robertmesserle:
@ThomasBurleson LGTM!

PR #4117, opened 2015-08-12
fix(chips): adjust chips css.

### LINKED ISSUE #3402: md-chips uses wrong font-size and padding
All input components should look the same. so md-chips uses another font and left padding
I think the correct font-size is 16px, and left padding is 3px
image

typotter:
Change font size and padding to match spec.

fixes #3402


typotter:
@jelbourn for a look.

jelbourn:
LGTM

PR #4136, opened 2015-08-12
fix(mdChips): Allow custom $interpolate start/end symbols.

### LINKED ISSUE #2985: md-select do not handle multiple keypress like native select element
Let's say we have a simple `md-select` element.

    <md-select ng-model="test">
        <md-option ng-value="A1">A1</md-option>
        <md-option ng-value="A2">A2</md-option>
        <md-option ng-value="A3">A3</md-option>
        <md-option ng-value="A4">A4</md-option>
    </md/select>

In native `select` element, if I press "A" two times, it will select "A1" then "A2" and after if I press "A4" fast, it will select "A4". Angular Material does not do it that way, if I press "A" multiple times it will select the first option with the "A" letter same thing for "A4".

It's easy to test it with the demo here : https://material.angularjs.org/latest/#/demo/material.components.select

topherfangio:
Refactor to pull template processing out of $interimElement
and into $mdUtil.

Fix chips to enable custom start/end symbols.

fixes #2985

### DIFF
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -1,140 +1,145 @@
 describe('<md-chips>', function() {
   var scope;
+
   var BASIC_CHIP_TEMPLATE =
-      '<md-chips ng-model="items"></md-chips>';
+    '<md-chips ng-model="items"></md-chips>';
   var CHIP_APPEND_TEMPLATE =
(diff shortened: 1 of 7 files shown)


topherfangio:
@typotter ready for review :-)

PR #4185, opened 2015-08-15
fix(fabToolbar): Fix md-direction attribute to be string instead of expression.

### LINKED ISSUE #3390: md-fab-toolbar - strange margin
Hi!
I've put md-fab-toolbar in my ui-router template and this happens:
bez tytulu
Template is in:
md-content
    div layout="column" ui-view
    /div
/md-content
In template md-fab-toolbar is next to table in div.
This happens also when I put this element outside ui-view.

topherfangio:
The fabSpeedDial treats the `md-direction` attribute as a string as does the shared fabController. However, the fabToolbar treated it as an expression.

Update the code to make it a string instead.

fixes #3390

### DIFF
--- a/src/components/fabToolbar/fabToolbar.js
+++ b/src/components/fabToolbar/fabToolbar.js
@@ -79,7 +79,7 @@
       '</div>',
 
       scope: {
-        direction: '=?mdDirection',
+        direction: '@?mdDirection',
         isOpen: '=?mdOpen'
       },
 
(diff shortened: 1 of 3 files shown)


topherfangio:
@jelbourn Can you review real fast?

ThomasBurleson:
@topherfangio - can you provide a test to accompany these changes?

src/components/fabToolbar/fabToolbar.js
```diff
@@ -65,7 +65,7 @@
       '</div>',
 
       scope: {
-        direction: '=?mdDirection',
+        direction: '@?mdDirection',
```
jelbourn: Need unit test for this.

PR #4191, opened 2015-08-16
fix(tooltip): Fix md-tooltip showing when window receives focus

### LINKED ISSUE #3035: md-tooltip with md-autohide displays on window re-focus
REPRO STEPS
* For easy repro, there is an example at http://codepen.io/anon/pen/doNWNV, where all of the tooltips have had md-autohide enabled.
1. Add an md-tooltip to a focusable element, and set md-autohide on it.  Ex:
   <md-tooltip md-autohide>Click here to learn more!</md-tooltip>
2. Click on the focusable element, and move the mouse away from the element.  As expected, the tooltip disappears.
3. Switch tabs in your browser, or windows in your O/S.
4. Switch back to the tab/window with your focusable element.

CURRENT BEHAVIOR
The tooltip will appear for the focused element, and remain until moused over or another element is focused.

EXPECTED BEHAVIOR
The tooltip should not appear until mouseover.

mhchen:
When the parent of an md-tooltip was clicked, it received focus. When
the window was blurred and re-focused (e.g. on tab change), the tooltip
shows again due to a new focus event firing. This fix prevents this
first focus event from having any effect. Note that this bug only
affected browsers that focus buttons on click.

Fixes #3035

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -134,7 +134,25 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
 
     function bindEvents () {
       var mouseActive = false;
-      var enterHandler = function() {
+
+      var ngWindow = angular.element($window);
+
+      // Store whether the element was focused when the window loses focus.
+      var windowBlurHandler = function() {
+        elementFocusedOnWindowBlur = document.activeElement === parent[0];
+      };
+      var elementFocusedOnWindowBlur = false;
+      ngWindow.on('blur', windowBlurHandler);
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@mhchen - can you provide a test to accompany these improvements?

mhchen:
@ThomasBurleson Rebased and re-pushed. The md-button module had to be pulled in in order to do actual simulation of the button.focus()/document.activeElement behavior. Because I was able to simulate this behavior (instead of relying on jasmine.mockElementFocus), it seemed like a good time to fix the fragility of the test above it as well.

ThomasBurleson:
@mhchen - :+1:

PR #4202, opened 2015-08-17
Fixed CSS issue with dialog and z-index

m0g:
md-backdrop had a z-index value superior md-dialog-container preventing the buttons inside a confirm dialog to be clickable


m0g:
bundle exec sidekiq -q tweets, -q accounts, -q sessions, -q disqus

ThomasBurleson:
This PR will not be accepted. See the **[variables.js](https://github.com/angular/material/blob/master/src/core/style/variables.scss#L65-L74)** where we clearly establish z-indices for our components.

PR #4221, opened 2015-08-18
fix(input): fix bad label initialization when using static value. Fixes #3599

### LINKED ISSUE #3599: readonly md-input without a ng-model
If you set a value on an input field without a ng-model, the value only appears after you click on the input field:

http://plnkr.co/edit/b90ENPhrAagfejN4AuMN?p=preview

I understand that having a md-input without a ng-model is useless, but I can see cases where having a readonly md-input with a (static) value would be convenient.

jelbourn:
@robertmesserle quick review?


robertmesserle:
LGTM

PR #4222, opened 2015-08-18
fix(css): remove styles that could apply to user-controlled elements.

jelbourn:
BREAKING CHANGE

@ThomasBurleson @robertmesserle 

Related to #3516

I'm removing style rules that would affect elements that are not created/owned by ng-material. As far as I can tell form looking at the demos after this, everything is still the same on our end. Users that were inadvertently depending on these styles will be affected.


robertmesserle:
@jelbourn LGTM - might want to include a BREAKING CHANGE warning in the commit message.

ThomasBurleson:
@jelbourn - what about mdGridList scss ?

PR #4223, opened 2015-08-19
fix(gesture): fix overzealous preventDefault on label+input clicks.

### LINKED ISSUE #4110: Click hijacking on mobile disrupting label behavior
Mobile users are not able to click on radio / checkbox buttons within a label. Specifically, when the user clicks on the label, the browser automatically fires a click on the input; but [this code](https://github.com/angular/material/blob/249631748be39566e3eeebb5dac24c853fd4188e/src/core/services/gesture/gesture.js#L451) hijacks it and prevents it altogether. 

I made a simple reproduction of the problem [here](http://codepen.io/anon/pen/waOVrr). Visit with Chrome's device mode spoofing as Nexus 5 (remember to refresh after choosing device from the dropdown), and you'll find that clicking on the label does not toggle the radio button. Clicking on the input itself works, but not the label.

jelbourn:
Fixes #4110 

This change is missing a unit test. After spending quite a bit of time trying to write one, I've been unable to find a way to imitate this browser behavior manually. 

Using `label.click()` will correctly make the associated input also fire a click event, but the coordinates for both events will be (0, 0), which $mdGesture interprets as a keyboard-triggered click.

Manually creating an event and using `label.dispatchEvent` lets you set the coordinates, but dispatching it in this way does not also automatically trigger a click on the input.

@robertmesserle @ThomasBurleson

### DIFF
--- a/src/core/services/gesture/gesture.js
+++ b/src/core/services/gesture/gesture.js
@@ -6,6 +6,12 @@
    */
   var pointer, lastPointer, forceSkipClickHijack = false;
 
+  /**
+   * The position of the most recent click if that click was on a label element.
+   * @type {{x: number, y: number}?}
+   */
+  var lastLabelClickPos = null;
+
   // Used to attach event listeners once when multiple ng-apps are running.
   var isInitialized = false;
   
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@jelbourn - I have encountered similar situations where a custom event was needed, but it would not trigger the handler. While a test would be ideal, I will merge for now...

Issue #4078, opened 2015-08-10
md-select with scrollbars auto closing


bijoux:
current head, the issue is present in both latest chrome and firefox.
To reproduce, go to the select demo page, open an md-select with a scrollbar, and try to scroll down.

viztastic:
For me the issue is with both version 10.0 and 10.1

I have an md-select. Appears fine on the page. And it also behaves fine when in large or medium screen mode:

http://screencast.com/t/k0rrG0Iu3EO

When I srhink the screen to a mobile width, and then click on md-select. It hides the page scroll bars and I can't seem to click out of the md-select or scroll up/down the page (even with the mouse scroll wheel)

http://screencast.com/t/RWhz8rGcKTvH

viztastic:
Also worth noting, that when I resize the screen to desktop.. everything is still frozen (no scroll bars and select is still open).

AleksueiR:
+1

rschmukler:
Closed by #4229

PR #4229, opened 2015-08-19
fix(select): Don't close on mouseup on scrollbar

FrankMeyer:
md-select does not auto-close anymore when the scrollbar is used to scroll the options.
Checked the event position in checkCloseMenu()
Testet on Windows: FF 40, Chrome 44, IE 11. Should be
See issue #4078 'md-select with scrollbars auto closing'


rrajewski:
@FrankMeyer 
I've seen more than a few people have issues with the menu closing while dragging the scroll bar and mousing up over the options. Could this be changed to not close if you are over any part of the dropdown? The native select stays open until you click outside of it.

ThomasBurleson:
This seems reasonable: 

> mouseUp events over the menu container (not just the scrollbar) should be ignored and not cause an auto-close.

FrankMeyer:
Oh yes, I also discovered the bug some days ago. I have provided a fix for that in #4767
I just took the mouse position of the mousedown event instead of the mouseup event to check for scrolling.

PR #4241, opened 2015-08-19
fix(slider): Update CSS to better conform to spec.

### LINKED ISSUE #2427: md-slider: Ripple positioned slightly off and Outer circle background color not set
It seems that `md-slider`'s ripple is positioned slightly off to the top left.

You can see this on the slider [demo](https://material.angularjs.org/#/demo/material.components.slider) if you click on one of the slider lines. It's very fast but I managed to get a screenshot. I tried clicking on the slider in multiple places as I thought the ripple origin had to do with my mouse position but it didn't affect it. I think it should either ripple from the center or from the position of the mouse relative to the sliders dot. 
image
Note that right now the ripple never reaches the bottom right corner and it appears to be created slightly off center.

It also seems the spec doesn't mention the outer grey circle seen below when you click and drag. 
edit - I think it is supposed to be colored the same as the ripple so it can look like the ripple doesnt go away until you release it. But the background color was never set.
image
Slider spec - http://www.google.com/design/spec/components/sliders.html#

topherfangio:
Remove grey focus ring in favor of transparent background ring.

fixes #2427

### DIFF
--- a/src/components/slider/slider-theme.scss
+++ b/src/components/slider/slider-theme.scss
@@ -10,7 +10,7 @@ md-slider.md-THEME_NAME-theme {
     background-color: '{{foreground-2}}';
   }
   .md-focus-ring {
-    border-color: '{{foreground-4}}';
+    background-color: '{{accent-color}}';
   }
   .md-disabled-thumb {
     border-color: '{{background-color}}';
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
Closed with SHA 7fd8b0985f385cc791

PR #4262, opened 2015-08-20
fix(datepicker): update input size and invalid class on calendar click

mhchen:
@jelbourn Please review

### DIFF
--- a/src/components/datepicker/datePicker.js
+++ b/src/components/datepicker/datePicker.js
@@ -212,8 +212,11 @@
 
     self.$scope.$on('md-calendar-change', function(event, date) {
       self.ngModelCtrl.$setViewValue(date);
+      self.date = date;
       self.inputElement.value = self.dateLocale.formatDate(date);
       self.closeCalendarPane();
+      self.resizeInputElement();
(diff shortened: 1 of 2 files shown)


mhchen:
Also updated a few unit tests involving the invalid class to use the .toHaveClass Jasmine matcher that I should've seen the first time around.

src/components/datepicker/datePicker.js
```diff
@@ -212,8 +212,11 @@
 
     self.$scope.$on('md-calendar-change', function(event, date) {
       self.ngModelCtrl.$setViewValue(date);
+      self.date = date;
       self.inputElement.value = self.dateLocale.formatDate(date);
       self.closeCalendarPane();
+      self.resizeInputElement();
+      self.inputContainer.classList.remove(INVALID_CLASS);
```
jelbourn: Would it work if you just called ngModelCtrl.$render here?

mhchen: Hm no, it doesn't look like calling $render actually triggers the input handlers

jelbourn:
LGTM

PR #4276, opened 2015-08-21
fix(subheader/sticky): Stick subheader to content instead of parent.

### LINKED ISSUE #3535: md-subheader / sticky discussion
I don't like how md-subheader wraps whatever I put into it an h2. There also doesn't seem to be any other option to make something sticky.
I wanted to make a card sticky (customer summary header), but no way to do it.

If you have 2 md-contents side by side with a md-subheader, this happens:

http://codepen.io/robertbaker/pen/bdLgmr

topherfangio:
The `$mdSticky` service used to stick it's contents to the parent of the `<md-content>` container. This resulted in some odd styling issues and the inability to have two lists with sticky subheaders side-by-side.

This fix appends the subheaders directly to the `<md-content>` to fix this issue.

Also, updated subheaders to use `<div>` tags instead of `<h2>`.

fixes #3535

### DIFF
--- a/src/components/sticky/sticky.js
+++ b/src/components/sticky/sticky.js
@@ -28,7 +28,7 @@ angular.module('material.components.sticky', [
  *     If not provided, it will use the result of `element.clone()`.
  */
 
-function MdSticky($document, $mdConstant, $compile, $$rAF, $mdUtil) {
+function MdSticky($document, $mdConstant, $$rAF, $mdUtil) {
 
   var browserStickySupport = checkStickySupport();
 
@@ -95,7 +95,9 @@ function MdSticky($document, $mdConstant, $compile, $$rAF, $mdUtil) {
       };
       self.items.push(item);
 
-      contentEl.parent().prepend(item.clone);
+      $mdUtil.nextTick(function() {
+        contentEl.prepend(item.clone);
+      });
 
       debouncedRefreshElements();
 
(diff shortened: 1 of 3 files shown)


topherfangio:
@jelbourn @ThomasBurleson @rschmukler  Can you guys please review this and let me know if you see any issues with the updates? This is related to the discussion we had in Monday's meeting (#3535), but I found a way to do it without the visual glitch I mentioned.

src/components/sticky/sticky.js
```diff
@@ -169,22 +171,22 @@ function MdSticky($document, $mdConstant, $compile, $$rAF, $mdUtil) {
         setCurrentItem(null);
 
       // Going to next item?
-      } else if (isScrollingDown && self.next) {
-        if (self.next.top - scrollTop <= 0) {
+      } else if (isScrollingDown) {
+        if (self.next && self.next.top - scrollTop <= 0) {
```
jelbourn: I know it was like this before, but could you change this? 
`self.next.top - scrollTop <= 0` is a less readable way of saying `self.next.top <= scrollTop`

src/components/sticky/sticky.js
```diff
@@ -169,22 +171,22 @@ function MdSticky($document, $mdConstant, $compile, $$rAF, $mdUtil) {
...
           } else {
-            translate(self.current, null);
+            translate(self.current, scrollTop);
           }
         }
         
       // Scrolling up with a current sticky item?
-      } else if (!isScrollingDown && self.current) {
-        if (scrollTop < self.current.top) {
+      } else if (!isScrollingDown) {
```
jelbourn: This can just be an `else` now.

src/components/sticky/sticky.js
```diff
@@ -169,22 +171,22 @@ function MdSticky($document, $mdConstant, $compile, $$rAF, $mdUtil) {
         setCurrentItem(null);
 
       // Going to next item?
-      } else if (isScrollingDown && self.next) {
-        if (self.next.top - scrollTop <= 0) {
+      } else if (isScrollingDown) {
```
jelbourn: It feels like this whole block could be a better high-level comment explaining what's going on.

topherfangio:
@jelbourn @ThomasBurleson As requested, I have refactored the `onScroll()` function and it's comments quite a bit to increase readability/understanding. In particular, I replaced the `if/else if` blocks with more simple `if/return` statements to more clearly separate the different cases and remove some duplicate code.

Can you two please take a look and provide feedback? It may be easier if you look at the [whole file](https://github.com/angular/material/blob/fix-subheader-sticky-3535/src/components/sticky/sticky.js#L164) rather than the diff so you can see the overall structure.

PR #4278, opened 2015-08-21
feat(contactChips): Add md-highlight-flags support.

### LINKED ISSUE #3182: md-contact-chips add md-highlight-flags
When the search function bound to the `md-contacts` attribute is case insensitive (like the example in the documentation), the `md-highlight-text` in `MD_CONTACT_CHIPS_TEMPLATE` is still case sensitive. It would be great if it was possible to define the `md-highlight-flags`.

topherfangio:
Contact Chips currently allows the use of `md-highlight-text` to match the search text. Users have requested we add `md-highlight-flags` as well so that the highlighted text can match the search field parameters.

Mainly, if a match is found with case-insensitive search, it would be nice if the highlight text was also case-insensitive.

Also, slightly refactor tests.

fixes #3182

### DIFF
--- a/src/components/chips/contact-chips.spec.js
+++ b/src/components/chips/contact-chips.spec.js
@@ -7,6 +7,7 @@ describe('<md-contact-chips>', function() {
           md-contact-name="name"\
           md-contact-image="image"\
           md-contact-email="email"\
+          md-highlight-flags="i"\
           placeholder="To">\
       </md-contact-chips>';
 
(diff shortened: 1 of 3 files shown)


topherfangio:
@typotter Can you review this super-fast and make sure this is the solution we discussed at the summit?

typotter:
lgtm, but please add docs for the new parameter to chips.

PR #4280, opened 2015-08-21
fix(subheader): Set styles to allow click/hover when stickied.

### LINKED ISSUE #3932: md-subheader don't allow clicks or hover events for md-button inside
Hello,
I'm using 0.9.8 because 0.10.0 still has some little bugs.
When an md-button is inside an md-subheader there is no HOVER event for css rules neither click event.
I've tried this css solution

```
.md-subheader .md-subheader-content > *{
    position: relative;
    z-index: 10;
}
```
this rule works but it give me this on the page when scrolling

schermata 2015-07-28 alle 17 49 52

I've no hover event or click event for any of element inside md-subheader.
How can I fix it?

Thanks in advance,
\m

topherfangio:
Previously, the inner DIVs of the subheader component were not set to use `display: block`, so you could potentially have contents that were inaccessible.

fixes #3932

### DIFF
--- a/src/components/subheader/subheader.scss
+++ b/src/components/subheader/subheader.scss
@@ -66,10 +66,12 @@ $subheader-sticky-shadow: 0px 2px 4px 0 rgba(0,0,0,0.16) !default;
   position: relative;
 
   .md-subheader-inner {
+    display: block;
     padding: $subheader-padding;
   }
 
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
Closed with SHA 27ea0c93fba5f5ffdff0a9

PR #4310, opened 2015-08-25
chore(layout): style changes for layout code

jelbourn:
@ThomasBurleson Looked through the layout code and figured this was the easiest way to give feedback.

* Removed underscores from functions names to follow Google JS style guide.
* Fixed JsDoc syntax
* Remove `try-catch` blocks. `removeAttr` is a no-op if the attribute isn't present, and using a `try-catch` block forces V8 to go into unoptimized mode, potentially having a very negative performance impact.
* Changed inconsistent indentation.

### DIFF
--- a/src/core/services/layout/layout.js
+++ b/src/core/services/layout/layout.js
@@ -45,69 +45,69 @@
 
       // Attribute directives with optional value(s)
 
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@jelbourn - lgtm. Will pull locally and test.

PR #4341, opened 2015-08-26
fix(mdTabs): Fix 0-height animation on iOS devices.

### LINKED ISSUE #4339: mdTabs: iOS dynamic height jumps incorrectly
On iOS, the Dynamic Height demo shows the following issue when changing tabs:

The tab immediately becomes 0 height and then animates to the proper height over a period of about one second.

Tested against latest master branch.

topherfangio:
On iOS devices, when switching between tabs, the height would jump
to 0, then animate to the proper place causing a bad glitch.

Fixes #4339

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -6,7 +6,7 @@ angular
  * @ngInject
  */
 function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipple,
-                           $mdUtil, $animate, $attrs, $compile, $mdTheming) {
+                           $mdUtil, $animateCss, $attrs, $compile, $mdTheming) {
   // define private properties
   var ctrl      = this,
       locked    = false,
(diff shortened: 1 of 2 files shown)


src/components/tabs/js/tabsController.js
```diff
@@ -635,17 +635,36 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+
+    var oldCssHeight = { height: currentHeight + 'px'},
+        newCssHeight = { height: newHeight + 'px' };
+
+    // Set the height to the current, specific pixel height to fix a bug on iOS where the height
+    // first animates to 0, then back to the proper height causing a visual glitch
+    $element.css(oldCssHeight);
+
+    // Animate the height from the old to the new
+    $animate.animate($element, oldCssHeight, newCssHeight).then(function () {
```
ThomasBurleson: why not use `$animateCss(element, { from, to, duration } )`?

src/components/tabs/js/tabsController.js
```diff
@@ -635,17 +635,36 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+      // Then (to fix the same iOS issue as above), disable transitions
+      $element.css('transition', 'none');
+
+      // And remove the specific pixel height so the height can size with browser width/content
+      // changes, etc.
+      $element.css('height', '');
```
ThomasBurleson: combine these into:

```js
$element.css({
   transitions : 'none',
   height : ''
});
```

src/components/tabs/js/tabsController.js
topherfangio: It was using `$animate` so I didn't want to alter the existing code too much, but I'll give this a try and run all of the tests.

PR #4359, opened 2015-08-27
fix(mdChips): Backspace key and custom input focus/blur.

### LINKED ISSUE #2607: md-chips: Focus state is triggered incorrectly
I just updated to v0.9.0rc3 and noticed that the `.md-focused` class is applied to the wrong element.

The focus state is applied properly if I press ctrl or cmd while focussed on the element.

### LINKED ISSUE #3562: Back button in FireFox inside md-chips and md-autocomplete causes weird issues
When I type something and add one chip in the md-autocomplete and hit enter and press backspace/delete, FireFox triggers the back button of the actual browser. Sometimes, when pressing backspace too doesn't delete the actual chip. It looks like the escape is being reroute to the back button of the browser.

### LINKED ISSUE #3960: chips: backspace to select broken in Edge
In Edge browser, backspace navigates back rather than selecting the previous chip.

Not sure if you're in the business of supporting Edge yet, so maybe this can be thrown in the backlog.

topherfangio:
* Fix backspace key in Firefox to not change browser history/url

   In Firefox, the backspace key would also modify the browser's
   history/url.

 * Fix focus/blur on chips with custom input:

   Chips with a custom input did not properly update it's class
   list when the inner input was focused/blurred.

Fixes #3562. Fixes #3960. Fixes #2607.

### DIFF
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -217,6 +217,58 @@ describe('<md-chips>', function() {
         expect(scope.appendChip).toHaveBeenCalled();
         expect(scope.appendChip.calls.mostRecent().args[0]).toBe('Apple');
       });
+
+      it('should prevent the default when backspace is pressed', inject(function($mdConstant) {
+        var element = buildChips(BASIC_CHIP_TEMPLATE);
+        var ctrl = element.controller('mdChips');
+
+        var backspaceEvent = {
+          type: 'keydown',
+          keyCode: $mdConstant.KEY_CODE.BACKSPACE,
+          which: $mdConstant.KEY_CODE.BACKSPACE,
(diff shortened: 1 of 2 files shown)


ngraef:
This fixes the issue in Edge as well (#3960). However, I got "$apply already in progress" errors during my testing (both Chrome and Edge), likely from the digest loop you added.

topherfangio:
@ngraef Thanks for the confirmation, and for finding the issue! I'm looking into it now. Looks like it only happens on chips with custom inputs...

topherfangio:
@ngraef Thanks again for catching the issue! I have changed the code to use `$evalAsync` instead of `$apply` and it looks to be fixed.

PR #4360, opened 2015-08-27
fix(autoComplete):  dynamic changing  ng-required not working

ljcp:
Setting the <autocomplete ng-required="var" to false doesn't update the input field ng-required="$mdAutocompleteCtrl.isRequired"  $mdAutocompleteCtrl.isRequired remains true


ljcp:
I signed it!

siulung4073:
+1

loic:
This breaks plain `required`. #4724 should do the trick.

EladBezalel:
@ljcp, @loic solution seems more right,
thanks!

PR #4366, opened 2015-08-28
Fix scrolling behavior of sidenav when popped out

isaaclyman:
Add `position: fixed;` to sidenav and backdrop when open but not locked, so that it is not possible to scroll below the backdrop (previously possible when body content is longer than the current screen height, which is ugly).

### DIFF
--- a/src/components/sidenav/sidenav.scss
+++ b/src/components/sidenav/sidenav.scss
@@ -73,11 +73,25 @@ md-sidenav {
     min-width: $sidenav-default-width;
     transform: translate3d(0%, 0, 0);
   }
+  
+  &:not(.md-closed):not(.md-locked-open) {
+    position: fixed;
+    top: 0;
+    left: 0;
+  }
 
   @extend .md-sidenav-left;
 }
-.md-sidenav-backdrop.md-locked-open {
-  display: none;
+.md-sidenav-backdrop {
+  &.md-locked-open {
+    display: none;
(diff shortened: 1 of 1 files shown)


isaaclyman:
I still think this could be a problem, though I've discovered that it's potentially fixable by giving the sidenav and all of its parents/ancestors the style `height: 100%;`.

src/components/sidenav/sidenav.scss
```diff
@@ -73,11 +73,25 @@ md-sidenav {
...
+.md-sidenav-backdrop {
+  &.md-locked-open {
+    display: none;
+  }
+  
+  &:not(.md-locked-open) {
+    position: fixed;
+    top: 0;
+    left: 0;
+  }
```
ThomasBurleson: why not also have `bottom:0`?

ThomasBurleson:
I want to review this some more... so I will reopen. Thx

isaaclyman:
Only thing is that for sidenavs that aren't full-screen (as in the Angular-Material demo), it won't stay inside of the container if it's `position: fixed;`. Something to consider.

ThomasBurleson:
@topherfangio - can you review this? Change to "needs: merging" when ready.

EladBezalel:
@isaaclyman can you please check it against HEAD? 
Closing for now, if it's still valid please rebase and i will reopen

Issue #3901, opened 2015-07-26
md-content no longer scrolls correctly in Chrome 44


hodeyp:
The latest version of Chrome (44.0.2403.107) on the mac no longer supports `md-content` overflow scrolling.

In previous versions, (as well as Safari) the `md-content` element would expand to fill the available space and overflow would scroll. In the latest version of Chrome, the `md-content` section grows to fit the child elements instead.

hodeyp:
take a look at https://code.google.com/p/chromium/issues/detail?id=506893

This explains the new default behaviour in Chrome, it suggests adding `* {min-height:0}' to css

dg1202:
I woke up to this same behavoir. I eventually found it was due to my use with nested ui-views with ui-router, so exactly what hodeyp links to as nested flex columns.

goncaloneves:
@hodeyp Thanks for this !

I have changed only the flex attribute on my custom css file:
```css
[flex] {
	min-height: 0;
    min-width: 0;
}
```

nkoterba:
@goncaloneves Thanks for the fix...working for now.

Seems related to: #4022

topherfangio:
Do any of you have a specific example/codepen you can provide that shows exactly how it's failing in material? I'd like to make sure we get a good/performant fix in.

ThomasBurleson:
Fixed with SHA 303ab0dc0e695f

PR #4372, opened 2015-08-29
fix(layout): Set the initial height to inherit the parent.

### LINKED ISSUE #4022: md-content doesn't work correctly with flex.
Maybe it works as expected but if it does, I'm not sure I fully understand the purpose of md-content and the documentation resumes up to one single line. I made a shorter example of my issue.

I want two divs with overflowing content each covering half the page, 

This example works as I would expect (the two divs are immediate children of the body):

http://codepen.io/langley/pen/Nqorzv

However if they are not direct children of the body, but a div instead, or even another md-card, md-whiteframe or any other, it stops working:

http://codepen.io/langley/pen/waNWYe

In order to make it work I have to apply height: inherit or height: 100% to the parent:

http://codepen.io/langley/pen/KpJMbR

It's not a big deal in this example but if the object I want to be scrollable is a nested child of 10 elements, I have to apply this to every single parent, not just the closest. 

I see that in the demo a fixed px value is used, but I thought the whole idea of being responsive is to use flex-xx style and % instead of fixed px values ?

### LINKED ISSUE #4130: Buttons not aligned in official site
The `Icon Buttons` are not aligned properly in the official site [here](https://material.angularjs.org/latest/#/demo/material.components.button) in Chrome with Angular-Material version - 0.10.1

Screenshot below:

screenshot 80

topherfangio:
In some cases, browsers have issues with rendering flex layouts
when the height of an element is not set. This simply forces each
layout element (which is the most common case when using flex) to
the height it would normally be.

Also fixes random icon padding in demo causing the icon to be
mis-aligned.

Fixes #4022. References #3901. Fixes #4130.

### DIFF
--- a/src/components/button/demoBasicUsage/style.css
+++ b/src/components/button/demoBasicUsage/style.css
@@ -20,7 +20,3 @@ section .md-button {
   font-size: 14px;
   opacity: 0.54;
 }
-
-.launch {
-  padding-top:12px;
-}
--- a/src/core/services/layout/layout.scss
+++ b/src/core/services/layout/layout.scss
@@ -17,6 +17,7 @@
   display: -moz-flex;
   display: -ms-flexbox;
   display: flex;
+  height: inherit;
 }
 
 .layout-column {


topherfangio:
@ThomasBurleson Per your request, I tested the demo site in all three browsers and did not see any difference in adding this additional property.

topherfangio:
@ThomasBurleson Everything was passing locally for me, but when it failed on Travis, I removed my `node_modules` directory and reran `npm install`. Now, I'm getting about 125-140 failures on master, and it's not consistent (but it appears to be consistent with Travis).

ThomasBurleson:
The issue is something in the latest Angular 1.4.5 release. @petebacondarwin, @matsko - can you guys help use investigate? Angular 1.4.4 works fine with our tests and Karma.... Angular 1.4.5 causes many failures.

ThomasBurleson:
@topherfangio - I temporarily locked out Angular version to 1.4.4; see package.json.

topherfangio:
@ThomasBurleson I also found one other issue this sometimes causes on the docs site: there is some way in the demo site that we can get the following:

```xml
<md-toolbar>Stuff Here</md-toolbar>
<md-content flex>Demos Here</md-content>
<md-content flex><!-- empty container with no conent //--></md-content>
```

Because of the CSS change, the empty `<md-content>` is still taking up 50% of the screen.

I'm trying to track it down to find a fix, but I'm having trouble finding where this HTML markup actually lives. I'll keep digging into it later today.

topherfangio:
@ThomasBurleson Finally found the random `<md-content>` issue and I think it's a non-issue (or at least not relevant to this). I have opened a separate issue for it: #4388.

I believe this is ready to merge now.

PR #4384, opened 2015-08-31
fix(mdSelect): Close menu on hitting Enter key

### LINKED ISSUE #4377: Selection of an item with keyboard recently broken in md-select
In current HEAD, if you try to select an item by hitting Enter, the menu won't close and you will get "Uncaught TypeError: Cannot read property 'currentTarget' of undefined":

https://material.angularjs.org/HEAD/#/demo/material.components.select

quietshu:
`checkCloseMenu` is being called without an event in `onMenuKeyDown`, which causes a `Uncaught TypeError: Cannot read property 'currentTarget' of undefined` error.

Fixes #4377.

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -1058,7 +1058,7 @@ function SelectProvider($$interimElementProvider) {
                 });
                 ev.preventDefault();
               }
-              checkCloseMenu();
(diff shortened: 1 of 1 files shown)


quietshu:
I signed it (CLA)! :ok:

PR #4391, opened 2015-08-31
fix(mdAutocomplete): Compile autocomplete template against proper scope.

### LINKED ISSUE #4390: mdChips: Contact Chips autocomplete dropdown is empty.
On master branch, on the Contact Chips demo, when you type in some content to search for a user, the autocomplete pops up with suggestions and filters properly, however all of the suggestions are empty.

screen shot 2015-08-31 at 11 15 51 am

### LINKED ISSUE #4495: md-item-template for md-autocomplete cannot access parent scope
This seems to be an issue that was introduced at some point in 0.11, as my upgrade to 0.11 from 0.10.1 seems to break things.

In short, I have a directive that essentially wraps `md-autocomplete`, and provides some app-specific logic that helps with the display of autocomplete items. It is still intended to be just as generic as `md-autocomplete`, but part of wrapping `md-autocomplete` means I would prefer to be able to abstract some of the templating logic for each item. 
I've added a codepen below to give some insight into this problem, but essentially what I'm doing is:
```
<div class="extended-autocomplete-directive">
    <md-autocomplete md-items="item in ...">
        <md-item-template>
            <div  class='template-text' >{{item[someDisplayAttributeAttachedToDirectiveScope]}}</div>
        </md-item-template>
    </md-autocomplete>
    <div class="some-other-stuff">...</div>
</div>
```
This worked in 10.1, and allowed me to attach an arbitrary attribute to the scope of my directive that I could use in my `md-item-template` to help display the information I wanted. This seems to no longer work in 11.0.

I've also tried using `ng-transclude` to pull in the `md-item-template` from further up. The markup would look like:
```
<extended-md-autocomplete>
    <md-item-template>
        ...
    </md-item-template>
</extended-md-autocomplete>
```
With my directive template looking something like:
```
<div class="extended-md-autocomplete">
    <md-autocomplete ng-transclude></md-transclude>
</div>
```
This throws a transclusion error (duh), as md-autocomplete does not transclude.

I'm thinking this is a bug, but maybe its intended behavior. If so, advice on how to proceed would be much appreciated.

Code pen: http://codepen.io/Flare/pen/KdpoeZ?editors=101

topherfangio:
After the performance modifications to make autocomplete use the
virtual repeat directive, the autocomplete no longer compiled the
contents of the dropdown against the proper parent scope

This PR fixes it by manually compiling against the proper parent scope
and copying over the necessary `item` and `$index` properties.

Fixes #4390. Fixes #4495.


topherfangio:
@jelbourn @robertmesserle Can you guys take a quick look at this PR and see if there is a better way to accomplish it? I am really hesitant to say that the `scope.$parent.$parent.$parent` is a "good" fix...

src/components/chips/js/contactChipsAutocompleteScope.js
```diff
@@ -0,0 +1,29 @@
+angular
+    .module('material.components.chips')
+    .directive('mdContactChipsAutocompleteScope', MdContactChipsAutocompleteScope);
```
robertmesserle: Can we rename this file to `contactChipsAutocompleteScopeDirective.js`?  It's verbose, but this is the convention we are using for the other files in here.

src/components/chips/js/contactChipsAutocompleteScope.js
```diff
@@ -0,0 +1,29 @@
...
+    restrict: 'A',
+    link:     postLink,
+    scope:    false
+  };
+
+  function postLink (scope, element, attr) {
+    // Since the autocomplete popup is appended to the body, we cannot traverse up the DOM looking
+    // for the proper controller and grabbing it's scope, so we must go WAY up to get the proper
+    // scope; create a new child scope based off of the parent, and copy the item over on changes
+    var newScope = scope.$parent.$parent.$parent.$new();
```
robertmesserle: If we require the chips controller with `require: '^mdChips'`, I think we could get the root controller and grab the proper scope from there.  That should be cleaner than the parent chain.

topherfangio: Unfortunately, I already tried that as I too thought it would be cleaner. However, the `mdChips` controller isn't available because the element is no longer in the same spot in the DOM (it gets `detach()`ed and moved) so it can't find it.

If there is another way to get that controller, I'll all for it.

robertmesserle: Hm.  If we're using `controllerAs` for `mdChips`, and there are no isolated scopes between this one and the variable the controller is defined as, we might be able to access it directly from this scope.

robertmesserle: In either case, I say this is good to merge - just maybe include a TODO for finding a cleaner way to get the root scope.  (You could even pass it in as an attribute if you have to.)

topherfangio: This was all based off of the autocomplete's [`parentScope.js`](https://github.com/angular/material/blob/master/src/components/autocomplete/js/parentScope.js) directive which grabs the autocomplete controller which stores a reference to the parent, and I am able to use that, but I still had to do `ctrl.parent.$parent.$new()` to get it to work, and that turned out to be more code.

Can you clarify exactly what you mean by the following? I'm super-happy to test out an alternative.

> we might be able to access it directly from this scope

robertmesserle: See if the scope has `$mdChipsCtrl` available on it.  If there are no isolated scopes between this element and the root chips element, it may inherit that variable.

robertmesserle: If not, I'd maybe pass the scope into this directive.  In either case, what you have now is fine, I just try to avoid the $parent chain whenever possible.

topherfangio:
@robertmesserle I was able to modify it to grab the autocomplete scope's parent and it seems to be working. Slightly nicer code:

```js
var newScope = scope.$mdAutocompleteCtrl.parent.$new();
```

Look good to you?

robertmesserle:
Yep, LGTM!

topherfangio:
@ThomasBurleson @robertmesserle I have updated this PR to be more generic than just the contact chips and fix the problem at the autocomplete level due to #4495 as well as some other comments on the initial autocomplete virtual repeat PR.

Can you guys do another review and make sure this looks reasonable/correct?

mackelito:
Any idea when this will make it out to the bower build?
Would love to see this fix so that we don't need to do a temp workaround :)

robertmesserle:
@topherfangio LGTM.  I may want to collaborate with you on this at some point in the future to look for ways to simplify this (which may or may not be possible).

topherfangio:
@robertmesserle Thanks! We should definitely pair! This does seem like a strange workaround, but it appears to follow with the original code in the `parentScope.js` file.

robertmesserle:
@topherfangio Right, I am not sure that code was entirely necessary either.  It may be, but if we could simplify, that would make our lives easier in the future.

houmark:
I have also tested this one and while it does not solve both of our non-working autocomplete use cases, it solves the most common one and the issue we have with the other one that is not working might be due to something else.

:+1: for merging this one ASAP :)

houmark:
@topherfangio @robertmesserle After some more debugging and refactoring we were able to make our custom chips work in combo with this PR so I'm all for LGTM! 

Nice work!

topherfangio:
@houmark Thanks so much for the updates and feedback!

PR #4394, opened 2015-08-31
fix(mdToolbar): Allow md-scroll-shrink usage with ng-if.

### LINKED ISSUE #2751: Toolbar in directive, scroll-shrink not working
Hi,

I've created a directive for a global nav-bar for my app, which creates a toolbar with some default stuff, and then transcludes page-specific content (actions etc).

Having the `md-toolbar` with `md-scroll-shrink=""` does not work, though.
As far as I can see, it never catches the `$mdContentLoaded` event, because the event is broadcast before my directive has loaded.

Is this a fault in my design, or is this an error in the way `md-toolbar` works?

topherfangio:
Previously, the toolbar would fail to enable scroll shrinking
if the developer used an `ng-if` statement on the `md-toolbar`.

This commit allows usage of `ng-if` as well as watching the
`md-scroll-shrink` for changes so they can enable/disable
scroll shrinking.

Fixes #2751.

### DIFF
--- a/src/components/toolbar/toolbar.js
+++ b/src/components/toolbar/toolbar.js
@@ -56,19 +56,22 @@ angular.module('material.components.toolbar', [
  * at one fourth the rate at which the user scrolls down. Default 0.5.
  */
 
-function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate ) {
-  var translateY = angular.bind(null, $mdUtil.supplant, 'translate3d(0,{0}px,0)' );
+function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
+  var translateY = angular.bind(null, $mdUtil.supplant, 'translate3d(0,{0}px,0)');
 
   return {
     restrict: 'E',
+    template: '<div ng-transclude></div>',
+    transclude: true,
     controller: angular.noop,
+    scope: {
+      scrollShrink: '=?mdScrollShrink'
+    },
     link: function(scope, element, attr) {
 
       $mdTheming(element);
 
-      if (angular.isDefined(attr.mdScrollShrink)) {
-        setupScrollShrink();
-      }
+      setupScrollShrink();
 
       function setupScrollShrink() {
         // Current "y" position of scroll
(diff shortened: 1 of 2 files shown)


topherfangio:
@ThomasBurleson When you review this PR, could you please check to ensure this **doesn't** constitute a breaking change?

Particularly, I've added an isolate scope, but it should be okay since I added the transclusion. I just wanted a second pair of eyes.

src/components/toolbar/toolbar.js
```diff
@@ -113,7 +157,7 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate )
           //
           // As the user scrolls down, the content will be transformed up slowly
           // to put the content underneath where the toolbar was.
-          var margin =  (-toolbarHeight * shrinkSpeedFactor) + 'px';
+          var margin = (-toolbarHeight * shrinkSpeedFactor) + 'px';
           contentElement.css('margin-top', margin);
           contentElement.css('margin-bottom', margin);
```
ThomasBurleson: I think this is better:

```js
contentElement.css({
 'margin-top' : margin,
'margin-bottom' : margin
});
```

src/components/toolbar/toolbar.js
```diff
@@ -89,6 +94,48 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate )
...
+          }
+        });
+
+        // If the scope is destroyed (which could happen with ng-if), make sure
+        // to disable scroll shrinking again
+        scope.$on('$destroy', function() {
+          disableScrollShrink();
+        });
+
+        function enableScrollShrink() {
```
ThomasBurleson: `enableScrollShrink( )` should return the disable function:

```js
var disableScrollShrink;

function enableScrollShrink() {
  contentElement.on('scroll', debouncedContentScroll);
  contentElement.attr('scroll-shrink', 'true');
  
  $$rAF(updateToolbarHeight);

  // Publish method to reverse the `enable`
  return function disableScrollShink() {
    contentElement.off('scroll', debouncedContentScroll);
    contentElement.attr('scroll-shrink', 'false');
  
    $$rAF(updateToolbarHeight);
  };
}


// If the scope is destroyed (which could happen with ng-if), make sure
// to disable scroll shrinking again

scope.$on('$destroy', function() {
  disableScrollShrink();
});
```

src/components/toolbar/toolbar.js
```diff
@@ -89,6 +94,48 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate )
...
+        function disableScrollShrink() {
+          contentElement.off('scroll', debouncedContentScroll);
+          contentElement.attr('scroll-shrink', 'false');
+
+          $$rAF(updateToolbarHeight);
+        }
+
         function onMdContentLoad($event, newContentEl) {
           // Toolbar and content must be siblings
           if (element.parent()[0] === newContentEl.parent()[0]) {
```
ThomasBurleson: FYI - `scope.$on(...)` will call `onMdContentLoad( )` **without** the 2nd argument.

src/components/toolbar/toolbar.js
```diff
@@ -56,12 +56,17 @@ angular.module('material.components.toolbar', [
...
  */
 
-function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate ) {
-  var translateY = angular.bind(null, $mdUtil.supplant, 'translate3d(0,{0}px,0)' );
+function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
+  var translateY = angular.bind(null, $mdUtil.supplant, 'translate3d(0,{0}px,0)');
 
   return {
     restrict: 'E',
+    template: '<div ng-transclude></div>',
```
ThomasBurleson: lgtm

topherfangio: Will move the handler.

Re: `isDefined`, my understanding is that if the attribute is present, we want the functionality. So, if they just do
`<md-toolbar md-scroll-shrink>`, we should treat that as `<md-toolbar md-scroll-shrink="true">`.

If we test the scope, though, the first case will be undefined, but we still want to run it.

I will add some comments to make this more clear.

topherfangio: Heh...actually, it looks like we can just remove the `if` altogether. The setup function already checks whether it should enable/disable, so we just have to call it.

src/components/toolbar/toolbar.js
topherfangio: In what case will it do so? All of my testing showed that it always passed the `newContentEl`.

PR #4395, opened 2015-08-31
docs(datepicker): add missing return

martvaha:


EladBezalel:
@martvaha seems to be redundant comment that left in the code.

PR #4407, opened 2015-09-01
ProgressLinear - improve docs regarding modes.

EladBezalel:
Please notice that 'query' mode is not working


ThomasBurleson:
@EladBezalel - plz file a separate issue if `query` mode is not working for ProgressLinear.

ThomasBurleson:
We should also update ProgressLinear to have the same performance improvements and API as the recently enhanced ProgressCircular. 

> I will reopen this to synchronize features between the two Progress indicator components.

PR #4416, opened 2015-09-01
docs(menuBar): Fix menu bar demo after recent toolbar changes.

topherfangio:
After updating the `<md-toolbar>` to allow `ng-if`, the toolbar no longer allows the `ng-controller` directive directly on it.

Update demo to move `ng-controller` to outer `<div>` and fix the layout.

_Note: In the GitHub compare, it seems like a massive change, but all I did was wrap the toolbar contents in an additional `<div>`._


ThomasBurleson:
@topherfangio - thx for the subnote explanation. That helped.

PR #4422, opened 2015-09-01
fix disableScrollShrink()

clshortfuse:
Seems like a syntax typo


topherfangio:
This is apparently a common practice in Angular Material: https://github.com/angular/material/pull/4394#discussion_r38381706

Also, after discussion with the team, we may be removing this functionality and finding a different way to do it; so this is likely to disappear anyway :-)

clshortfuse:
I'm having a lot of issues with the toolbar. I've noticed that I'm getting disabledscrollshrink isn't defined in my console and realized it wasn't because this. There shouldn't be a call to it if it's null. 

But there are a lot of little bugs so I rather wait for you guys to polish it more before doing a PR.

Issue #2825, opened 2015-05-12
Dangling md-subheader and md-toolbar


nobarany:
When a md-toolbar with md-scroll-shrink shrinks, the following subheader or md-toolbar doesn't move up to the top of the page, and it just dangles in the middle.
This codepen shows the problem for subheader, which appears only in Chrome (related to issues #1647 and #1726):
http://codepen.io/anon/pen/rVeaaL

But the problem with a second toolbar happens in Firefox as well:
http://codepen.io/anon/pen/QbNwQj

Interestingly, by adding md-scroll-shrink to the second toolbar you can achieve the behavior that you'd expect from the above code (i.e., the second one doesn't shrink). 
http://codepen.io/anon/pen/MwyYvj

topherfangio:
@nobarany I just pushed some code that ***might*** fix this problem and will be available in 0.11. When it lands, could you try it out and see if it solved your problem?

topherfangio:
Can you please provide a demo using the latest master code about exactly what you are attempting to accomplish?

I believe your first issue has been addresses, but I'm not exactly sure what you're wanting your second and third demos to do.

nobarany:
In the second demo, there are two toolbars: the first one is supposed to shrink as user scrolls down and the second one is supposed to stick to the top of the page. It still doesn't work (with the latest master code from gitcdn.xyz):
http://codepen.io/anon/pen/EVeEgb

topherfangio:
Okay, so I think you have three options here:

1. Toolbar 1 Stays Visible and Toolbar 2 Shrinks - I don't think you need `md-scroll-shrink`. Just put the second toolbar inside of the `md-content` like this: http://codepen.io/topherfangio/pen/epLKmq

2. Toolbar 2 Stays Visible and Toolbar 1 Shrinks - You can put the toolbars side-by-side and put the `md-scroll-shrink` on the Toolbar 2: http://codepen.io/topherfangio/pen/EVeRjj

3. A variation of #2 - You put `md-scroll-shrink` on both and it still does the same as #2, but it gives a slightly different "push" effect rather than the "cover" effect: http://codepen.io/topherfangio/pen/zvJaBG

Do any of those meet your needs, or are you trying to do something else?

In general, the `md-content` has really only been designed to work with a single `md-toolbar`, so anything outside of that is probably going to require some customization.

nobarany:
[...]

PR #4437, opened 2015-09-02
fix(mdSubheader): Non top-level md-content causes incorrect position.

### LINKED ISSUE #4420: md-subheader only works if at top of page
Here's the test/proof http://codepen.io/anon/pen/zvYJNB

It seems to keep traveling up through every parent

Edit: more analysis brings me to this [line](https://github.com/angular/material/blob/5e6aa6893ec77582268d4e74e828f34dbab8b954/src/components/sticky/sticky.js#L73)

Edit2: Since moving to v0.11, it also breaks toolbars that aren't at the top. Seems to be related to $mdSticky

topherfangio:
If the subheader's parent `<md-content>` was not the top-level item,
the stickied subheader would be incorrectly positioned further down
the page.

Also added more tests to toolbar and fixed issue to ensure subheader
works with a toolbar using `md-scroll-shrink`.

Fixes #4420. References #2825.

### DIFF
--- a/src/components/sticky/sticky.js
+++ b/src/components/sticky/sticky.js
@@ -70,7 +70,6 @@ function MdSticky($document, $mdConstant, $$rAF, $mdUtil) {
     contentEl.on('$scroll', onScroll);
 
     var self;
-    var stickyBaseoffset = contentEl.prop('offsetTop');
     return self = {
       prev: null,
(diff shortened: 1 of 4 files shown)


ThomasBurleson:
Merged with SHA 391479b044cd9dc39

Issue #4033, opened 2015-08-05
(perf)md-tooltip uses angular watcher per item


clshortfuse:
This isn't a major issue at all, but md-tooltip is using 1 (or maybe 2) watchers per tooltip. While many won't see this as major, I'm currently using a large table that's completely structured to be bind once.

If, for example, I have 200 rows, and each row has a md-tooltip, it would be 200/400 watchers which are called on each digest. It has terrible performance. I am aware of the ability to use virtual-repeat, but for UX reasons, I cannot use it. I have opted to remove the md-tooltip for this reason. It seems a better performance fix would be to use DOM's hover/mouse-over/mouse-enter and then use an $apply when necessary instead of having each tooltip being watched on every digest.

I can construct a codepen to show the performance issues.

nestorneto:
+1

mcapu:
Is there a workaround for having performance problems due to having many md-tooltips right now? My situation is similar, although it's caused by several tooltips in each row.

ThomasBurleson:
What about using a virtual repeater with 10-20 rows visible ?

clshortfuse:
@ThomasBurleson I would if I could, but I've designed a full height single page app, so I can't mix a virtual repeater. At least I couldn't when I tried before.

I have 2 tables in one vertical element. Both tables are of dynamic height, but you scroll through the one parent element. I've solved the performance issue there with bindonce but I had to remove every tooltip.

If virtual repeater had something like ng-repeat-start and ng-repeat-end, then I could probably do it.

SanderElias:
The watch is in there to make it possible to show/hide the tooltip manually. I found no easy(read performant) way to lose the $watch and keep this functionality. If anyone has a good idea how to solve this catch 22?

no-more:
Hi,

I'm not an expert, but in order to keep track can't we use a service (or an optional parent directive) and store the status outside of the component itself ?
I'm not sure if this won't generate other issues or if it will work as well as the current version. But it could be a way to limit the number of watchers. One important point is that only a single tooltip can be visible at time (or at least I think so).

skosno:
Could you consider to not create watcher if one is not passing "md-visible"? In most of my use cases I do not need any manual control on tooltips and as indicated in directive bindings, it is optional (=?). 

Same with direction maybe - if not passed, just skip the watcher?

clshortfuse:
[...]

PR #4441, opened 2015-09-02
fix(md-tooltip): removed watcher for performance

clshortfuse:
md-tooltip uses a watcher per tooltip. The watcher is just to watch if scope.visible changes. This variable only changes internally so there's no reason to need a watcher when the changes could be tracked manually.

fixes https://github.com/angular/material/issues/4033

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -82,10 +82,12 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
 
     function configureWatchers () {
       scope.$on('$destroy', function() {
-        scope.visible = false;
+        if (scope.visible !== false) {
+          scope.visible = false;
+          $mdUtil.nextTick(checkVisibility);
+        }
         element.remove();
         angular.element($window).off('resize', debouncedOnResize);
-        checkVisibility();
       });
     }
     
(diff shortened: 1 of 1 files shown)


clshortfuse:
Okay. I see the issue. md-visible can change scope.visible from the outside. I was wrong to assume that .visible as just an internal value. That's really the only issue left.

src/components/tooltip/tooltip.js
```diff
@@ -81,14 +82,20 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
 
     function configureWatchers () {
       scope.$on('$destroy', function() {
-        scope.visible = false;
+        if (scope.visible !== false) {
+          scope.visible = false;
+          $mdUtil.nextTick(checkVisibility);
```
ThomasBurleson: Why are you waiting a frame and then calling hideTooltip when the scope has been destroyed?

ThomasBurleson:
@clshortfuse - please provide 2-n `it(...)` tests that validate these changes.

clshortfuse:
That was me trying to exactly replicate the $watch behavior trying to figure out why the tests failed. Then I realized it was because md-visible.

ThomasBurleson:
@clshortfuse - should I close this PR without trying merge? If no, then please squash and rebase.

clshortfuse:
It's not right and buggy. I'll have to revisit it another day. For now, I'll close it.

Issue #980, opened 2014-12-18
Export Theming logic


Sandreu:
Hello,

I used to be angular, but I'm starting playing with react. I like it and I would like to use angular-material markup to build a [react version of angular-material](https://github.com/Sandreu/react-angular-material).

I just met the new ThemeProvider... This is much nicer than the previous theme handling. 

But I ask myself how to extract the theme building logic from the angular module to be able to do a gulp task and/or a react mixin using the same code... I can do pull request, but I just want to be agree on how to do that.

What do you think about it ?

sscovil:
I would love to see the theming logic in a separate package as well, as a node module that can be run in a grunt/gulp task. I'll take a swing at it and share my progress here.

sscovil:
I've been giving this some thought, and wanted to share my observations and ideas in case anyone else is interested...

## Problem

Right now, theme-specific CSS is generated at runtime and added to the `<head>` section of the application in several `<style>` tags. This is fine for themes that are created on the fly by the end user, but is not ideal for themes that ship with the application and could be served as static CSS files.

## Observations

Looking at the underlying code, here is how the theme generation currently works:

1. During the Angular Material build process, the [themeBuildStream](https://github.com/angular/material/blob/f60b5e5a7713af9058f9a29966f6e324cae901b9/gulp/util.js#L187) function compiles all `*-theme.scss` files from components into a single CSS string that is injected into the `material.core` module as a constant called `$MD_THEME_CSS`.

1. The module `angular.core` appears to only consist of the `$MD_THEME_CSS` constant definition. (See the end of the [unminified angular-material.js](https://cdnjs.cloudflare.com/ajax/libs/angular-material/0.10.1/angular-material.js) file.)

1. In [theming.js](https://github.com/angular/material/blob/master/src/core/services/theming/theming.js#L450) there is a function called `generateThemes` that utilizes `$MD_THEME_CSS` to generate the `<style>` tags that get added to the `<head>` section at runtime.

1. For each of the four (4) palettes in each registered theme (including the default theme), there are four (4) `<style>` tags added to the `<head>` section (e.g. `.md-primary`, `.md-primary.md-hue-1`, `.md-primary.md-hue-2`, `.md-primary.md-hue-3`). That means each registered theme results in 16 `<style>` tags being generated and added to the `<head>` section at runtime.

[...]

PR #4448, opened 2015-09-03
docs(theming): add Under the Hood section

sscovil:
document the inner workings of theme generation, per discussion in #980

### DIFF
--- a/docs/content/Theming/03_configuring_a_theme.md
+++ b/docs/content/Theming/03_configuring_a_theme.md
@@ -132,3 +132,28 @@ angular.module('myApp', ['ngMaterial'])
 
 });
 </hljs>
+
+### Under the Hood
+
+Angular Material dynamically generates CSS for registered themes by injecting several
+`<script>` tags into the `<head>` section of the application at runtime. Here is how
+that process currently works:
+
+1. During the Angular Material build process, a function in `gulp/util.js` called
+[themeBuildStream](https://github.com/angular/material/blob/master/gulp/util.js#L187)
+compiles all `*-theme.scss` files from components into a single CSS string that is
+injected into the `material.core` module as a constant called `$MD_THEME_CSS`.
+
+1. At runtime, a function in `src/core/services/theming/theming.js` called
+[generateThemes](https://github.com/angular/material/blob/master/src/core/services/theming/theming.js#L450)
+utilizes `$MD_THEME_CSS` to generate the `<style>` tags that are added to the `<head>`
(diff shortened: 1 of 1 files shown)


docs/content/Theming/03_configuring_a_theme.md
```diff
@@ -132,3 +132,28 @@ angular.module('myApp', ['ngMaterial'])
...
 </hljs>
+
+### Under the Hood
+
+Angular Material dynamically generates CSS for registered themes by injecting several
+`<script>` tags into the `<head>` section of the application at runtime. Here is how
+that process currently works:
+
+1. During the Angular Material build process, a function in `gulp/util.js` called
+[themeBuildStream](https://github.com/angular/material/blob/master/gulp/util.js#L187)
```
ThomasBurleson: I do not believe `default-theme.scss` exists anymore.

docs/content/Theming/03_configuring_a_theme.md
```diff
@@ -132,3 +132,28 @@ angular.module('myApp', ['ngMaterial'])
...
+compiles all `*-theme.scss` files from components into a single CSS string that is
+injected into the `material.core` module as a constant called `$MD_THEME_CSS`.
+
+1. At runtime, a function in `src/core/services/theming/theming.js` called
```
ThomasBurleson: I need to confirm... but I believe we changed our rules so **default** components styles do not have `.md-default-theme` selectors in the CSS. This allows the components to be more easily styled. 

@rschmukler - can you confirm?

docs/content/Theming/03_configuring_a_theme.md
sscovil: The files I'm referring to are in each component directory (e.g. `autocomplete-theme.scss`, `backdrop-theme.scss`, `bottom-sheet-theme.scss`).

docs/content/Theming/03_configuring_a_theme.md
sscovil: Not sure if that is the case, but the default theme definitely produces 16 style tags in the head section. Here is a plunk with a build of the latest master commit: http://plnkr.co/edit/5W0q58xmvg76TSWmf32W?p=preview

ThomasBurleson: @rschmukler - did we not make some perf changes to Theme injection regarding the default themes? Or was that only to not namespacing the default CSS styles ?

PR #4451, opened 2015-09-03
fix(mdChips): prevent undefined onAppend return values from adding to items

mxab:
if the mdOnAppend does not return any value, undefined is added to the items


EladBezalel:
@mxab can you please rebase?

mxab:
@EladBezalel I looked into it, but during the rebase I noticed this:

`// Check for a null (but not undefined), or existing chip and cancel appending`
https://github.com/angular/material/blob/master/src/components/chips/js/chipsController.js#L280

wich makes me if you now want to add undefined on purpose now?

EladBezalel:
I guess that this is not necessary anymore, thanks!

Issue #4388, opened 2015-08-31
mdBottomSheet: TypeError: Cannot read property 'runner' of undefined


topherfangio:
There is an issue on the docs site which can throw the above error and break the site's formatting.

Steps to reproduce:

1. Open the [bottom sheet demo](https://material.angularjs.org/HEAD/#/demo/material.components.bottomSheet)
1. Click the "Show as Grid" button
1. Click the "Twitter" icon/button
1. VERY QUICKLY click another demo on the left before the toast starts to disappears

The error will be thrown and there will be two `<md-content>` tags on the page, the bottom one will be empty and cause 50% of the page to be empty (due to PR #4372).

<img width="1440" alt="screen shot 2015-08-31 at 11 03 41 am" src="https://cloud.githubusercontent.com/assets/54370/9583275/0a034cae-4fd0-11e5-8c26-b273377ea7bb.png">

ThomasBurleson:
@matsko - how do we cancel an animation runner in `$destory()` ?

DevVersion:
This will be easy to fix after we have #6684, because then the interim elements won't get removed directly if removing. They will be removed from stack, after their animation is done. 

Then we can use $mdBottomsheet.destroy() because after that we will find an interim element (see [here](https://github.com/DevVersion/material/blob/6a564508567cf1e352495bdf803f6db94334c601/src/core/services/interimElement/interimElement.js#L381-L381)).

ThomasBurleson:
This issue is closed as part of our ‘Surge Focus on Material 2' efforts.
      For details, see our forum posting @ http://bit.ly/1UhZyWs.

PR #4454, opened 2015-09-03
fix(progressLinear): sync logic, perf upgrades, fix animations

### LINKED ISSUE #1926: progress-circular generates too much CSS
[This SASS while-loop](https://github.com/angular/material/blob/master/src/components/progressCircular/progressCircular.scss#L87) in the progress-circular styles expands to be about 2 kilobytes after gzip. 

The component should be refactored to do this same thing with $animate from javascript. It might be tricky, though, to make it as smooth.

### LINKED ISSUE #2364: md-progress-linear: IE 11 Flickering, Constant resizing of md-dialog container width
Relevant CodePen: http://codepen.io/londovir/pen/PwMMZJ

Under Chrome, an md-progress-linear that is embedded within an md-dialog component will show a slightly resizing (likely due to the transformX/scale CSS transformation done for a "determinate" style md-progress-linear) every time the attached value attribute is changed. However, the overall outer md-dialog will not resize accordingly. In addition, there is no apparently "flicker" of the darker colored progress bar portion as it is redrawn/resized by the CSS transformation - it smoothly glides smaller or larger.

Under IE 11, the exact same CodePen is causing multiple issues. The dark color "value bar" flickers out of existence before coming back, making for a jarring visual experience. In addition, the outer md-dialog component visually resizes in the horizontal (and often vertical) directions, causing a noticeable "shiver" of the md-dialog component it's embedded within. We have tried various combinations of wrapping different elements in outer DIVs with various attributes to try and prevent this layout resizing from taking place, but nothing has yet worked that doesn't mandate a fixed size for the outer md-dialog, which we are trying to avoid if possible.

### LINKED ISSUE #2540: md-progress-linear doesn't show progress bar in content area
md-progress-linear doesn't show the progress bar in the content area but shows it right below it instead. You can see this in the demo: https://material.angularjs.org/#/demo/material.components.progressLinear. The bar is drawn in the margin area of the md-progress-linear element and the content area is empty. This is because there is a top declaration in the bar container that is equal to the height declarations of both the container and the larger element.

I am creating a UI element that has a progress bar flush against the top of another element and noticed that the bar wasn't showing because of this.

Example: http://codepen.io/anon/pen/NqWeda
You would expect the progress bar to be touching the top lime outline. It is instead in the top of the black box. If the black box were drawn on top of the progress bar, the progress bar would not be seen at all.

From material/src/components/progressLinear/progressLinear.scss
```SCSS
$progress-linear-bar-height:5px !default;

md-progress-linear {
  display: block;
  width: 100%;
  height: $progress-linear-bar-height;

  .md-container {
    overflow: hidden;
    position: relative;
    height: $progress-linear-bar-height;
    top: $progress-linear-bar-height;  /* <--- Why is this here? */
    transform: translate(0, 5px) scale(1, 0);
    transition: all .3s linear;
  }
}
```

### LINKED ISSUE #3802: md-autocomplete 'searching' progress bar not displaying
I am using an`md-autocomplete` to search for data that is returned from a promise. Looking at the demo a progress bar is displayed during the search but this is not appearing on master.

### LINKED ISSUE #4409: mdProgressLinear: query mode is not working
Seems like it's not working quite a while (since [0.9.0](https://material.angularjs.org/0.9.0/#/demo/material.components.progressLinear))

### LINKED ISSUE #4421: ProgressLinear: sync API with enhanced ProgressCircular
We should update ProgressLinear to have the same performance improvements and API as the recently enhanced ProgressCircular.

ThomasBurleson:
synchronize progressLinear with similar logic used in progressCircular.

* improve animation performances
* watch md-mode for changes
* refactor animation SCSS
* enable hiding and no-animations with undefined/empty md-mode attributes
* for both indicators, use `display:inline-block;`
* update demos with enable switch
* fix query mode
* update Select to use enhanced progressCircular component
* fix autocomplete styling of progress-linear.md-mode-indeterminate

BREAKING-CHANGES

Before:

```css
md-progress-linear {
  display: block;
}
md-progress-circular {
   // display not set
   // position not set
}
```

```css
md-progress-linear {
  display: inline-block;
  position: relative;
}
md-progress-circular {
  display: inline-block;
  position: relative;
}
```

Fixes #4421. Fixes #4409. Fixes #2540. Fixes #2364. Fixes #1926. Fixes #3802.

### DIFF
--- a/src/components/progressCircular/progress-circular.js
+++ b/src/components/progressCircular/progress-circular.js
@@ -26,9 +26,13 @@ angular.module('material.components.progressCircular', [
  * not necessary to expose what's happening behind the scenes and how long it will take, use an
  * indeterminate indicator.
  *
- * @param {string} md-mode Select from one of two modes: **'determinate'** and **'indeterminate'**.<br/>
- * Note: if the `md-mode` value is undefined or not 1 of the two (2) valid modes, then `.ng-hide`
+ * @param {string} md-mode Select from one of two modes: **'determinate'** and **'indeterminate'**.
+ *
+ * Note: if the `md-mode` value is set as undefined or specified as not 1 of the two (2) valid modes, then `.ng-hide`
  * will be auto-applied as a style to the component.
+ *
+ * Note: if not configured, the `md-mode="indeterminate"` will be auto injected as an attribute.
+ * If `value=""` is also specified, however, then `md-mode="determinate"` would be auto-injected instead.
  * @param {number=} value In determinate mode, this number represents the percentage of the
  *     circular progress. Default: 0
  * @param {number=} md-diameter This specifies the diamter of the circular progress. The value
@@ -46,7 +50,7 @@ angular.module('material.components.progressCircular', [
  * <md-progress-circular md-mode="indeterminate"></md-progress-circular>
  * </hljs>
  */
-function MdProgressCircularDirective($mdConstant, $mdTheming, $mdUtil) {
+function MdProgressCircularDirective($mdTheming, $mdUtil, $log) {
   var DEFAULT_PROGRESS_SIZE = 100;
   var DEFAULT_SCALING = 0.5;
 
@@ -87,43 +91,68 @@ function MdProgressCircularDirective($mdConstant, $mdTheming, $mdUtil) {
   function postLink(scope, element, attr) {
     $mdTheming(element);
 
-    var circle = element[0];
+    var circle = element;
     var spinnerWrapper =  angular.element(element.children()[0]);
-
     var lastMode, toVendorCSS = $mdUtil.dom.animator.toCss;
 
-    // Update size/scaling of the progress indicator
-    // Watch the "value" and "md-mode" attributes
+    updateScale();
+    validateMode();
+    watchAttributes();
 
-    circle.style[$mdConstant.CSS.TRANSFORM] = 'scale(' + getDiameterRatio() + ')';
+    /**
+     * Watch the value and md-mode attributes
+     */
+    function watchAttributes() {
+     attr.$observe('value', function(value) {
+           var percentValue = clamp(value);
+           element.attr('aria-valuenow', percentValue);
 
-    attr.$observe('value', function(value) {
-      var percentValue = clamp(value);
-      element.attr('aria-valuenow', percentValue);
+           if (mode() == MODE_DETERMINATE) {
+             animateIndicator(percentValue);
+           }
+         });
+     attr.$observe('mdMode',function(mode){
+       switch( mode ) {
+         case MODE_DETERMINATE:
+         case MODE_INDETERMINATE:
+           spinnerWrapper.removeClass('ng-hide');
+           spinnerWrapper.removeClass( lastMode );
+           spinnerWrapper.addClass( lastMode = "md-mode-" + mode );
+           break;
+         default:
+           spinnerWrapper.removeClass( lastMode );
+           spinnerWrapper.addClass('ng-hide');
+           lastMode = undefined;
+           break;
+       }
+     });
+    }
 
-      if (attr.mdMode == "determinate") {
-        animateIndicator(percentValue);
+    /**
+     * Update size/scaling of the progress indicator
+     * Watch the "value" and "md-mode" attributes
+     */
+    function updateScale() {
+      circle.css(toVendorCSS({
+        transform : $mdUtil.supplant('scale( {0} )',[getDiameterRatio()])
+      }));
+    }
+
+    /**
+     * Auto-defaults the mode to either `determinate` or `indeterminate` mode; if not specified
+     */
+    function validateMode() {
+      if ( angular.isUndefined(attr.mdMode) ) {
+        var hasValue = angular.isDefined(attr.value);
+        var mode = hasValue ? MODE_DETERMINATE : MODE_INDETERMINATE;
+        var info = "Auto-adding the missing md-mode='{0}' to the ProgressCircular element";
+
+        $log.debug( $mdUtil.supplant(info, [mode]) );
+
+        element.attr("md-mode",mode);
+        attr['mdMode'] = mode;
       }
-    });
-
-    attr.$observe('mdMode',function(mode){
-      switch( mode ) {
-        case MODE_DETERMINATE:
-        case MODE_INDETERMINATE:
-          spinnerWrapper.removeClass('ng-hide');
-
-          // Inject class selector instead of attribute selector
-          // (@see layout.js changes for IE performance issues)
-
-          if ( lastMode ) spinnerWrapper.removeClass( lastMode );
-               lastMode = "md-mode-" + mode;
-          if ( lastMode ) spinnerWrapper.addClass( lastMode );
-
-          break;
-        default:
-          spinnerWrapper.addClass('ng-hide');
-      }
-    });
+    }
 
     var leftC, rightC, gap;
 
@@ -136,6 +165,8 @@ function MdProgressCircularDirective($mdConstant, $mdTheming, $mdUtil) {
      * - use attribute selectors which had poor performances in IE
      */
     function animateIndicator(value) {
+      if ( !mode() ) return;
+
       leftC  = leftC  || angular.element(element[0].querySelector('.md-left > .md-half-circle'));
       rightC = rightC || angular.element(element[0].querySelector('.md-right > .md-half-circle'));
       gap    = gap    || angular.element(element[0].querySelector('.md-gap'));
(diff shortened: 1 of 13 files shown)


ThomasBurleson:
@robertmesserle, @topherfangio - ready for review.

topherfangio:
Will dig into the code changes shortly, but when running locally, I noticed a small issue:

1. Open either of the Progress Linear or Progress Circular demos
1. Turn off the animations using the On/Off toggle switch at the bottom
1. Switch to the other demo
1. Sometimes, the `TypeError: Cannot read property 'runner' of undefined` error is thrown

It appears identical to issue #4388.

Originally, it seemed to happen every time I switched the demos, but since your (now next to latest) commit, it seems to happen less.

src/components/progressCircular/progress-circular.js
```diff
@@ -87,43 +89,76 @@ function MdProgressCircularDirective($mdConstant, $mdTheming, $mdUtil) {
...
+     */
+    function watchAttributes() {
+     attr.$observe('value', function(value) {
+           var percentValue = clamp(value);
+           element.attr('aria-valuenow', percentValue);
+
+           if (mode() == MODE_DETERMINATE) {
+             animateIndicator(percentValue);
+           }
+         });
```
topherfangio: Is there a particular reason why the formatting has extra indentation here?

Also, should it be `mode() === MODE_DETERMINATE` (i.e. with the triple `=`)?

src/components/progressCircular/progress-circular.js
```diff
@@ -87,43 +89,76 @@ function MdProgressCircularDirective($mdConstant, $mdTheming, $mdUtil) {
...
+       switch( mode ) {
+         case MODE_DETERMINATE:
+         case MODE_INDETERMINATE:
+           spinnerWrapper.removeClass('ng-hide');
+
+           // Inject class selector instead of attribute selector
+           // (@see layout.js changes for IE performance issues)
+
+           if ( lastMode ) spinnerWrapper.removeClass( lastMode );
+                lastMode = "md-mode-" + mode;
```
topherfangio: If you're not paying attention, the indentation makes this look like it is part of the above `if` conditional.

Also, doesn't it make the next line `if (lastMode)` irrelevant since we're specifically setting it here? I mean, even if `mode` is undefined, `last-mode` would still be `"md-mode-"` right?

src/components/progressCircular/progress-circular.js
```diff
@@ -87,43 +89,76 @@ function MdProgressCircularDirective($mdConstant, $mdTheming, $mdUtil) {
...
-          // Inject class selector instead of attribute selector
-          // (@see layout.js changes for IE performance issues)
+    /**
+     * Auto-defaults the mode to either `determinate` or `indeterminate` mode; if not specified
+     */
+    function validateMode() {
+      if ( angular.isUndefined(attr.mdMode) ) {
+        var hasValue = angular.isDefined(attr.value);
+        var mode = hasValue ? MODE_DETERMINATE : MODE_INDETERMINATE;
+        var info = "Auto-adding the missing md-mode='{0}' to the ProgressLinear element";
```
topherfangio: Possibly typo: `ProgressCircular` instead of `ProgressLinear`?

src/components/progressLinear/progress-linear.js
```diff
@@ -59,57 +79,112 @@ function MdProgressLinearDirective($$rAF, $mdConstant, $mdTheming) {
...
+          case MODE_BUFFER:
+          case MODE_DETERMINATE:
+          case MODE_INDETERMINATE:
+            container.removeClass('ng-hide');
+
+            // Inject class selector instead of attribute selector
+            // (@see layout.js changes for IE performance issues)
+
+            if ( lastMode ) container.removeClass( lastMode );
+                 lastMode = "md-mode-" + mode;
```
topherfangio: Same comments as for ProgressCircular.

src/components/progressLinear/progress-linear.js
```diff
@@ -59,57 +79,112 @@ function MdProgressLinearDirective($$rAF, $mdConstant, $mdTheming) {
...
+          case MODE_DETERMINATE:
+          case MODE_INDETERMINATE:
+          case MODE_BUFFER:
+          case MODE_QUERY:
+            break;
+          default:
+            value = undefined;
+            break;
+        }
+      }
```
topherfangio: Just curious, is this wrapping `if( value )` necessary? Shouldn't the `default` case catch that and set it to `undefined`?

Or is this just a way to optimize the case of it being undefined so you don't have to run the switch?

src/components/progressLinear/progress-linear.scss
```diff
@@ -1,92 +1,110 @@
...
 
-  &[md-mode=query] {
-    .md-bar2 {
-      animation: query .8s infinite cubic-bezier(0.390, 0.575, 0.565, 1.000);
+    .md-bar1, .md-bar2 {
+
+      // Just set the transition information here.
+      // Note: the actual transform values are calculated in JS
+
+      transition: transform 0.2s linear;
```
topherfangio: I believe using `transform` here instead of `all` disabled the spiffy height transition that makes the component "scale" into view instead of just immediately appearing.

src/components/progressCircular/progress-circular.js
robertmesserle: Actually, contrary to what most of the JS world says, the Google JS styleguide uses `==` instead of `===`.  I struggle to break the `===` habit on this project. =p

src/components/progressCircular/progress-circular.js
```diff
@@ -27,8 +27,10 @@ angular.module('material.components.progressCircular', [
  * indeterminate indicator.
  *
  * @param {string} md-mode Select from one of two modes: **'determinate'** and **'indeterminate'**.<br/>
- * Note: if the `md-mode` value is undefined or not 1 of the two (2) valid modes, then `.ng-hide`
- * will be auto-applied as a style to the component.
+ * <br/>Note: if the `md-mode` value is set as undefined or specified as not 1 of the two (2) valid modes, then `.ng-hide`
+ * will be auto-applied as a style to the component.<br/>
+ * <br/>Note: if not configured, the `md-mode="indeterminate"` will be auto injected as an attribute.
+ * If `value=""` is also specified, however, then `md-mode="determinate"` would be auto-injected instead.
  * @param {number=} value In determinate mode, this number represents the percentage of the
```
robertmesserle: I thought you could just add newlines in here rather than manually adding `<br/>`.  I could be wrong, though.

src/components/progressCircular/progress-circular.js
```diff
@@ -87,43 +89,76 @@ function MdProgressCircularDirective($mdConstant, $mdTheming, $mdUtil) {
...
+         case MODE_DETERMINATE:
+         case MODE_INDETERMINATE:
+           spinnerWrapper.removeClass('ng-hide');
+
+           // Inject class selector instead of attribute selector
+           // (@see layout.js changes for IE performance issues)
+
+           if ( lastMode ) spinnerWrapper.removeClass( lastMode );
+                lastMode = "md-mode-" + mode;
+           if ( lastMode ) spinnerWrapper.addClass( lastMode );
```
robertmesserle: As @topherfangio mentioned, I'm pretty sure this `if()` is not necessary.

src/components/progressCircular/progress-circular.js
```diff
@@ -87,43 +89,76 @@ function MdProgressCircularDirective($mdConstant, $mdTheming, $mdUtil) {
...
+
+           break;
+         default:
+           spinnerWrapper.addClass('ng-hide');
+           if ( lastMode ) {
+             spinnerWrapper.removeClass( lastMode );
```
robertmesserle: If you use my suggested logic above, this will also need to be changed:

```
spinnerWrapper.removeClass('md-mode-' + lastMode)
```

src/components/progressCircular/progress-circular.js
ThomasBurleson: I will fix the indentation. And it is in a switch, so the mode is always defined.

src/components/progressCircular/progress-circular.js
ThomasBurleson: oops. yep

src/components/progressLinear/progress-linear.scss
ThomasBurleson: That slide up is superfluous to the component.

src/components/progressCircular/progress-circular.js
ThomasBurleson: I will try

src/components/progressCircular/progress-circular.js
ThomasBurleson: We need to cache the lastMode for use in the `default` case:

```js
switch( mode ) {
          case MODE_QUERY:
          case MODE_BUFFER:
          case MODE_DETERMINATE:
          case MODE_INDETERMINATE:
            container.removeClass('ng-hide');

            // Inject class selector instead of attribute selector
            // (@see layout.js changes for IE performance issues)

            container.removeClass( lastMode );
            lastMode = "md-mode-" + mode;
            container.addClass( lastMode );

            break;
          default:
            container.addClass('ng-hide');
            container.removeClass( lastMode );
            lastMode = undefined;
            break;
        }
```

robertmesserle: Right, I included that in the last line.

PR #4463, opened 2015-09-03
performance(menuController): remove unnecessary $watch

clshortfuse:
We don't need a watcher for an value that only changes internally.

This will help speed up pages with many buttons (for example in a table).

It's useful while md-virtual-repeat continues to mature.

### DIFF
--- a/src/components/menu/js/menuController.js
+++ b/src/components/menu/js/menuController.js
@@ -31,7 +31,9 @@ function MenuController($mdMenu, $attrs, $element, $scope, $mdUtil, $timeout) {
 
     menuContainer.on('$mdInterimElementRemove', function() {
       self.isOpen = false;
+      $mdUtil.nextTick(function(){ $scope.onIsOpenChanged(self.isOpen);});
     });
+    $mdUtil.nextTick(function(){ $scope.onIsOpenChanged(self.isOpen);});
   };
 
   this.enableHoverListener = function() {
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@clshortfuse - how does a *watcher* improvement with **menuController** speed up performance for large table of mdButton(s)?

clshortfuse:
I was too specific. I have an HTML table with varying row height and uses a verticle ellipsis button. It works something like this: http://screenshots.en.sftcdn.net/blog/en/2013/05/google-music-player-1024x536.png

If I have a table with 100 items, and each row has a md-menu button, that's 100 extra watchers that are inspected per digest/apply cycle. (md-virtual-repeat doesn't play nice with tables yet).

Eliminating the watcher is a big performance gain.

ThomasBurleson:
@clshortfuse - How do you know the changes create a performance gain? Did you measure this with the profiler? 
> Please understand that I am not being rude; rather I want to validate and corroborate all changes before accepting PRs and code refactors.

clshortfuse:
I use ng-stats available here: https://github.com/kentcdodds/ng-stats

I also use Chrome's profiler, but ng-stat is a bit better because it's tuned for angular.  

If needed, I can create a codepen to show it's usage, but essentially, it's just be a big list of md-menu objects.

If the digest loop get very large, the performance will suffer. With the watcher, on every iteration, it will synchronously check and compare the return value of the function. The digest loop can get called rather frequently on a complex angular-material app. If we don't need the watcher, it's better not to use it.

topherfangio:
@clshortfuse Can you double-check and see if this is still valid? It seems to have some merge conflicts now.

Also, is there a reason that `onIsOpenChanged` is defined on `$scope` instead of the controller (i.e. `this`)?

EladBezalel:
closed due to inactivity

clshortfuse:
I'm sorry, I've never rebased anything before. I'm trying to do it now.

Should I start a new pull request or can this be reopened?

Issue #4466, opened 2015-09-04
Md-chips allows duplicate objects


gpopovic:
Md-chips currently doesn't allow duplicate strings, but will happily allow duplicate objects.

MdChipsCtrl.prototype.appendChip checks for duplicates with indexOf and because of that only duplicate strings and duplicate items referencing the same object are removed.

Because of that I often find myself writing md-on-append function to check for myself if objects are the same. 

Solution would be to check if $chip is an object and use angular.Equals()? or to allow user to specify expression used to check for duplicates.

raunakpsingh:
Is there anyworkaround for this??

topherfangio:
See PR #4479 which should hopefully be merged soon; we're trying to get 0.11 pushed out the door first.

Dwiga:
how to make a count on it?
for example, 
Banana (1), Apple (1), Tomato (1)
----------------------------------
after that, i write on it tomato again
it'll become like this
Banana (1), Apple (1), Tomato (2)...
thank for your help.
dwi.ahmad@protonmail.com

Issue #4193, opened 2015-08-16
mdChips: feature: md-on-append as a simple callback


emps:
im using md-on-append with md-chips and it works ok except of 'undefined' pushed in ng-model array
so there is no string chip and the chips i add via md append

topherfangio:
I'm not sure I fully understand. Why is your `md-on-append` returning an undefined chip?

smashercosmo:
@topherfangio because I think that md-on-append is used here as a simple callback, like newly added md-on-remove. It is not used to transform chip string to object. May be we need simple md-on-add, that won't force us to return an object value?

e-oz:
Please explain how to use md-on-append. How handler function should decide what to return, if it doesn't receive any arguments?

topherfangio:
@e-oz `md-on-append` is passed the text that the user typed so that you can translate it into an actual object (or just return the string if that's what you want).

Please see the [Basic Usage Demo](https://material.angularjs.org/HEAD/#/demo/material.components.chips) for more information.

e-oz:
@topherfangio please try it with autocomplete and objects instead of text.

e-oz:
@topherfangio thanks for advice - on second I try I found usage in code and I see I should pass ctrl.method($chip) in template to expect it in controller. Unusual but usable :)
Sorry for taking your time and thanks for help.

topherfangio:
@e-oz Not a problem!

facuxt:
For example...
 <md-chips ng-model="selectedVegetables" md-autocomplete-snap="" md-require-match="true" md-on-append="wtf($chip)">

and in controller...

    $scope.wtf = function(chip){
    	//chip.wtf = $scope.contador++;
    	var newChip = {};
    	angular.copy(chip, newChip)
    	return newChip;
    	//return chip;
    }

topherfangio:
@facuxt We have an outstanding PR (#4479) to automatically disallow identical chips.

Do you have a use-case for identical chips? Could you easily add an `id` field or something to make them unique?

facuxt:
Hello topherfangio. im doing a mathematical formula builder... i show you here some screenshots... maybe i can add a unique id or something in order to avoid that new feature... but... i think that a more apropiate approach to this should be to leave a configuration-thingy like md-allow-repeated to solve this issue for all of us.
![screen](https://cloud.githubusercontent.com/assets/7550822/10377320/fbad5b96-6dd8-11e5-891f-1f1fc03617e0.png)

This works like a charm...

PR #4479, opened 2015-09-05
fix(md-chips): appendChip disallows identical objects

JordanMajd:
md-chips' appendChip disallows appending an identical string or an exact object (duplicate object) more than once. It now disallows identical objects from being appended.

fixes angular/material#4466

### DIFF
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -192,13 +192,19 @@ MdChipsCtrl.prototype.getAdjacentChipIndex = function(index) {
  * call out to the md-on-append method, if provided
  * @param newChip
  */
-MdChipsCtrl.prototype.appendChip = function(newChip) {
-  if (this.useOnAppend && this.onAppend) {
-    newChip = this.onAppend({'$chip': newChip});
-  }
-  if (this.items.indexOf(newChip) + 1) return;
-  this.items.push(newChip);
-};
+ MdChipsCtrl.prototype.appendChip = function(newChip) {
+   if (this.useOnAppend && this.onAppend) {
+     newChip = this.onAppend({'$chip': newChip});
+   }
+   if(typeof newChip === 'object'){
+     var identical = this.items.filter(function(item){
+       return angular.equals(newChip, item);
+     });
+     if(identical.length > 0) return;
+   }
+   if (this.items.indexOf(newChip) + 1) return;
+   this.items.push(newChip);
+ };
 
 /**
  * Sets whether to use the md-on-append expression. This expression is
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@typotter, @topherfangio - can you review plz.

src/components/chips/js/chipsController.js
```diff
@@ -192,13 +192,19 @@ MdChipsCtrl.prototype.getAdjacentChipIndex = function(index) {
...
-    newChip = this.onAppend({'$chip': newChip});
-  }
-  if (this.items.indexOf(newChip) + 1) return;
-  this.items.push(newChip);
-};
+ MdChipsCtrl.prototype.appendChip = function(newChip) {
+   if (this.useOnAppend && this.onAppend) {
+     newChip = this.onAppend({'$chip': newChip});
+   }
+   if(typeof newChip === 'object'){
```
topherfangio: I might use `angular.isObject()` here.

Also, if you could add some whitespace and a few comments describing what each of the `if`s check for, I think it would help readability.

Otherwise, LGTM :-)

JordanMajd: Good idea,

I'm on it.

topherfangio:
@ThomasBurleson This looks good to me; should be ready for merging in 0.12 pending your review.

src/components/chips/js/chipsController.js
```diff
@@ -192,13 +192,27 @@ MdChipsCtrl.prototype.getAdjacentChipIndex = function(index) {
...
-};
+ MdChipsCtrl.prototype.appendChip = function(newChip) {
+
+   // If useOnAppend and onAppend function is provided call it.
+   if (this.useOnAppend && this.onAppend) {
+     newChip = this.onAppend({'$chip': newChip});
+   }
+
+   // If items contains identical object to newChip do not append
+   if(angular.isObject(newChip)){
```
ThomasBurleson: Too many returns... can we refactor to make it more clear and concise ?

JordanMajd: Hey Thomas, that sounds like a good idea. I'll work on it.

topherfangio:
@ThomasBurleson Just got an interesting [comment](https://github.com/angular/material/issues/4193#issuecomment-146662822) about someone using chips with repeated items (building a formula editor).

Is it worth it to make this a configuration option like `md-allow-repeats="true"` (defaults to false) for people who want to use it like this?

JordanMajd:
@topherfangio that sounds like a good idea.

topherfangio:
@JordanMajd I just spoke with Thomas, he said that we should just merge this code and provide a codepen showing how to make items unique instead of adding an additional attribute. He should be merging soon.

PR #4530, opened 2015-09-10
fix(datepicker): handle DST incocnsistency encountered in some timezones. Fixes #4215

### LINKED ISSUE #4215: datepicker: Past dates are not correct
Well, i played a little with the new [date picker](https://material.angularjs.org/HEAD/#/demo/material.components.datepicker) and inserted my birthday: ``` 10/20/1992 ``` 
image
than i opened the date picker and clicked again on 20:
image2
than it became ``` 10/19/1992 ```
image3

Don't know why but it seems that the line is between ``` 4/2/1992 ``` to ``` 4/3/1992 ```

jelbourn:
I'm well aware that this is a pretty hacky solution, but I haven't been able to think of something cleaner. 

It's also not something that I can see how to write a unit test for, since it depends on a specific timezone being set for the _operating system_.

All of this only comes into play when users use the default parsing/formatting instead of something like moment (like they're supposed to).

@ThomasBurleson @robertmesserle for review

### DIFF
--- a/src/components/datepicker/dateLocaleProvider.js
+++ b/src/components/datepicker/dateLocaleProvider.js
@@ -154,7 +154,23 @@
        * @returns {string}
        */
       function defaultFormatDate(date) {
-        return date ? date.toLocaleDateString() : '';
+        if (!date) {
+          return '';
+        }
+
+        // All of the dates created through ng-material *should* be set to midnight.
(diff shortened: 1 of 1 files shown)


robertmesserle:
@jelbourn LGTM

PR #4547, opened 2015-09-10
feat(datepicker): add shadow and animation cue for scrolling. Fixes #4313

### LINKED ISSUE #4313: md-date-picker: Usability scrolling
It is not obvious that you can scroll through the dates. 
In addition keyboard interactions are not enabled too.

jelbourn:
@robertmesserle for review

### DIFF
--- a/src/components/datepicker/calendar.scss
+++ b/src/components/datepicker/calendar.scss
@@ -8,6 +8,8 @@ $md-calendar-weeks-to-show: 7 !default;
 $md-calendar-month-label-padding: 8px !default;
 $md-calendar-month-label-font-size: 13px !default;
 
+$md-calendar-scroll-cue-shadow-radius: 6px;
+
 $md-calendar-width: (7 * $md-calendar-cell-size) + (2 * $md-calendar-side-padding);
 $md-calendar-height:
     ($md-calendar-weeks-to-show * $md-calendar-cell-size) + $md-calendar-header-height;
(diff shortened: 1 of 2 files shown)


src/components/datepicker/calendar.scss
```diff
@@ -79,9 +81,16 @@ md-calendar {
 
 // Contains the scrolling element (this is the md-virtual-repeat-container).
 .md-calendar-scroll-container {
+  // Add an inset shadow to help cue users that the calendar is scrollable. Use a negative x
+  // offset to push the vertical edge shadow off to the right so that it's cut off by the edge
+  // of the calendar container.
+  box-shadow: inset -3px 3px $md-calendar-scroll-cue-shadow-radius rgba(black, 0.2);
```
ThomasBurleson: Why not use a `white-frame` instead of a custom `box-shadow` variation?

src/components/datepicker/datePicker.scss
```diff
@@ -163,3 +163,17 @@ md-datepicker[disabled] {
     display: none;
   }
 }
+
+// Animate the calendar inside of the floating calendar pane such that it appears to "scroll" into
+// view while the pane is opening. This is done as a cue to users that the calendar is scrollable.
+.md-datepicker-calendar-pane {
+  .md-calendar {
+    transform: translateY(150px);
+    transition: transform 0.4s $swift-ease-out-timing-function;
```
ThomasBurleson: Should `transition: ....` be in here instead:

```css
&.md-pane-open .md-calendar {
    transition: transform 0.4s $swift-ease-out-timing-function;
    transform: translateY(0);
}
```

to set the duration only when opening... but `0` duration when **not** opening.

src/components/datepicker/calendar.scss
jelbourn: All of the whiteframes were more complicated than what I wanted to do here, which is add a shadow to *only* the top of element (and to be `inset` on top of that). The whiteframes, as I understand them, are for raising an entire box to a certain level.

src/components/datepicker/datePicker.scss
jelbourn: The calendar pane disappears immediately when closed without an animation, so it doesn't make a difference.

PR #4552, opened 2015-09-10
perf($mdSticky): Fix performance issue when sticky is emulated.

### LINKED ISSUE #4459: md-subheader low performance
Hello, I went into a possible performance issue when adding md-subheader (i.e: as is in the official demo in the Angular Material website).

It seems to behave properly according to documentation and demo page, but when scrollbar is available to interact, both scrolling and window scaling interactions would result in a remarcable loss of fps.

Also, the current demo seems to have such fps loss > https://material.angularjs.org/latest/#/demo/material.components.subheader

Can someone confirm this?

Chrome Version 44.0.2403.157 m

### LINKED ISSUE #4523: Serious performance issues on some demos and doc pages
I'm experiencing some serious performance issues when trying to scroll the webpage when accessing the [Subheader Demo](https://material.angularjs.org/latest/#/demo/material.components.subheader) and the docs for [Child Alignment](https://material.angularjs.org/latest/#/layout/alignment).

I'm using Chrome 45 on MacOS.

Can anyone confirm this issue?

topherfangio:
When the browser is not naitively capable of handling sticky elements, we fake it by watching for scroll changes and manually triggering certain events. These use `$$rAF`, but were not throttling the call.

Fix: use `$$rAF.throttle()` when calling our loop function.

Fixes #4523. Fixes #4459.

### DIFF
--- a/src/components/sticky/sticky.js
+++ b/src/components/sticky/sticky.js
@@ -296,7 +296,7 @@ function MdSticky($document, $mdConstant, $$rAF, $mdUtil) {
     element.on('scroll touchmove', function() {
       if (!isScrolling) {
         isScrolling = true;
-        $$rAF(loopScrollEvent);
+        $$rAF.throttle(loopScrollEvent);
         element.triggerHandler('$scrollstart');
       }
       element.triggerHandler('$scroll');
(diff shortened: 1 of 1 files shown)


topherfangio:
@ThomasBurleson This should fix the performance issues we noticed with subheader and the child alignment pages (it was specifically with the `$mdSticky` service which both of those utilize).

Needs quick review/merging.

ThomasBurleson:
This is awesome. Thank you for fixing this glaring performance issue.

PR #4556, opened 2015-09-10
Update tabsDirective.js not to use multiline string literal

lmanners:
Multiline string literals aren't considered a best practice.
References:
http://stackoverflow.com/a/6247331
http://google-styleguide.googlecode.com/svn/trunk/javascriptguide.xml?showone=Multiline_string_literals#Multiline_string_literals


ThomasBurleson:
I reconsidered since there is some practice defined in Google StyleGuide. Thx

PR #4596, opened 2015-09-11
feat(datepicker): prevent calendar from going off-screen. Fixes #4333.

### LINKED ISSUE #4333: Datepicker has issues with resizing and repositioning with varying window sizes
First of all the it won't fit into small screen.
screen shot 2015-08-26 at 3 46 53 pm
Secondly, if you resize the window back, it stays where it is.
screen shot 2015-08-26 at 3 47 40 pm

jelbourn:
@robertmesserle @rschmukler for review

Approach is pretty basic, it's possible I'm not thinking of something.

### DIFF
--- a/src/components/datepicker/datePicker.js
+++ b/src/components/datepicker/datePicker.js
@@ -100,6 +100,26 @@
   var DEFAULT_DEBOUNCE_INTERVAL = 500;
 
   /**
+   * Height of the calendar pane used to check if the pane is going outside the boundary of
+   * the viewport. See calendar.scss for how $md-calendar-height is computed; an extra 20px is
+   * also added to space the pane away from the exact edge of the screen.
+   *
+   *  This is computed statically now, but can be changed to be measured if the circumstances
+   *  of calendar sizing are changed.
+   */
+  var CALENDAR_PANE_HEIGHT = 368;
+
+  /**
+   * Width of the calendar pane used to check if the pane is going outside the boundary of
+   * the viewport. See calendar.scss for how $md-calendar-width is computed; an extra 20px is
+   * also added to space the pane away from the exact edge of the screen.
+   *
+   *  This is computed statically now, but can be changed to be measured if the circumstances
+   *  of calendar sizing are changed.
+   */
+  var CALENDAR_PANE_WIDTH = 360;
+
+  /**
    * Controller for md-datepicker.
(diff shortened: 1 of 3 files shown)


src/components/datepicker/datePicker.js
```diff
@@ -328,8 +351,29 @@
...
+    // If the bottom edge of the pane would be off the screen and shifting it up by the
+    // difference would not go passed the top edge of the screen.
+    if (paneTop + CALENDAR_PANE_HEIGHT > bodyRect.bottom &&
+        bodyRect.bottom - CALENDAR_PANE_HEIGHT > 0) {
+      paneTop = bodyRect.bottom - CALENDAR_PANE_HEIGHT;
+      calendarPane.classList.add('md-datepicker-pos-adjusted');
+    }
+
+    calendarPane.style.left = paneLeft + 'px';
+    calendarPane.style.top = paneTop + 'px';
```
ThomasBurleson: I think this code should be refactored to `function updatePosition( ){ .... }`.

jelbourn: I thought about this, but refactoring the positioning out of this function would leave just two lines (the first `addClass` and the last `addClass`) or need a function signature that involves passing around ClientRect objects, which I rather dislike.

This is because the inputMask is also updated using the elementRect, and you don't want to re-invoke getBoundingClientRect after the append since it would cause another reflow.

src/components/datepicker/datePicker.js
```diff
@@ -328,8 +351,29 @@
...
 
-    calendarPane.style.left = (elementRect.left - bodyRect.left) + 'px';
-    calendarPane.style.top = (elementRect.top - bodyRect.top) + 'px';
+    // Check to see if the calendar pane would go off the screen. If so, adjust position
+    // accordingly to keep it within the viewport.
+    var paneTop = elementRect.top - bodyRect.top;
+    var paneLeft = elementRect.left - bodyRect.left;
+
+    // If the right edge of the pane would be off the screen and shifting it left by the
+    // difference would not go passed the left edge of the screen.
```
robertmesserle: *past

jelbourn: http://blog.oxforddictionaries.com/2015/01/passed-past/

It's "passed".

Issue #4105, opened 2015-08-11
mdDialog: focus trap broken, incorrectly allowing keyboard navigation behind the modal


marcysutton:
Somehow in the latest rewrite of the dialog, it no longer traps keyboard focus inside the modal. You can shift+tab and get to the Alert, Confirm and Custom demo dialog buttons behind the modal (and beyond), or tab to cycle through the whole document. This is incorrect–it should trap focus inside of the dialog in addition to disabling everything in the background with `aria-hidden="true"` (which it still does).

The alert & custom dialog demos were also changed to only cover the embedded demo part of the screen, while the custom one still covers the whole screen...which is confusing for a keyboard user. They should really cover the entire viewport to visually indicate that all other links behind the dialog are disabled.

ThomasBurleson:
@marcysutton - the demos covering the embedded demo only (instead of the whole screen) is intended to visually show how different parents can be used for the dialogs.

erinkidd01:
Do we have a timeline for when this change is expected? As of right now, I'm forcing focus to the dialog when it's opened, capturing all of the TAB keypresses, and controlling the focus manually. It's quite tedious as we have many dialogs on the page.

skovalyov:
Can it be optional to add these focus traps before and after? I've been using the old behavior by purpose for quite a while. To be more specific, I'm using `md-dialog` for data tabel small edit dialogs, described at https://www.google.com/design/spec/components/data-tables.html#data-tables-interaction. And if the user presses Tab key I want to navigate to the next row, i.e. open small edit dialog for it. This worked perfectly with some tweaks before this change.

Issue #2443, opened 2015-04-21
md-select arrow keys (Up/Down) do not work when used in md-dialog


red010b37:
Using the arrows keys on an open select inside a md-dialog does not work at all - the up down arrows do not work even after you manually click on an element to force focus on the select.

Using the list in a non modal section the select functions as expected

mateusleon:
This appears to be with the `trapFocus()` function used on `md-dialog`.

https://github.com/angular/material/blob/master/src/components/dialog/dialog.js#L441-L448

```javascript
    function trapFocus(ev) {
      var dialog = document.querySelector('md-dialog');

      if (dialog && !dialog.contains(ev.target)) {
        ev.stopImmediatePropagation();
        dialog.focus();
      }
    }
```

I think it's kind of tricky to deal with, because of its numerous approachs, coming from defining exceptions on the `trapFocus` handler...

But, what do you think to place an element, like a sandbox of dinamically created instances, like `mdDialog`, or the `mdSelect` options, where all elements placed there belogs to the current manipulation:

```html
    [...]
    <md-sandbox>
        <md-sandbox-instance [attrs-for-identification-or-configuration]>
            <!-- Here goes a mdDialog, for example -->
        </md-sandbox-instance>

        <md-sandbox-instance [attrs-for-identification-or-configuration]>
            <!-- Here goes a mdSelect options, for example -->
        </md-sandbox-instance>
    </md-sandbox>
</body>
```

From the sandbox, events are contextualized, avoiding that `trapFocus` to be a shoot on the toe...

redyyu:
+1

joshribakoff:
+1

I imagine lots of people will get distracted by the eye candy & invest lots of time in porting from ui.bootstrap or some other framework, only to find out 8hrs in that they destroyed usability, etc.

gitawego:
I have the same problem because of ``trapFocus``, I can't focus on any input elements in my dropdown menu (menu is appended to body.) if a dialog is open.

PR #4609, opened 2015-09-13
fix(dialog): disallow tabbing into a aira-hidden regions created when…

onumossn:
aria-hidden="true" does not prevent elements inside the hidden region from being tabbed to. Because the element is still within an aria-hidden="true" region, screen readers will not read the content of the tabbed element. I think this can cause a great deal of confusion. Therefore, I set the tabindex of all tabbable elements that I am aware of within the aria-hidden="true" regions that are set with the walkDOM function to -1, and set them back when walkDOM is used to show set aria-hidden to false.

Also, http://www.w3.org/TR/wai-aria-practices/#modal_dialog suggests trapping focus in the dialog entirely, but I thought that might be going a bit too far as you may want to tab out of the page/application, and into the address bar or something else.

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -751,8 +751,8 @@ function MdDialogProvider($$interimElementProvider) {
       };
 
       /**
-       * Walk DOM to apply or remove aria-hidden on sibling nodes
-       * and parent sibling nodes
+       * Walk DOM to apply or remove aria-hidden and set tabindex appropriately
+       * based on aria-hidden value on sibling nodes and parent sibling nodes
        *
        */
       function walkDOM(element) {
@@ -760,12 +760,37 @@ function MdDialogProvider($$interimElementProvider) {
           if (element === document.body) {
             return;
           }
-          var children = element.parentNode.children;
+          var children = element.parentNode.children,
+            attrToSet = 'md-saved-tabindex',
+            attrToDefault = 'tabindex',
+            tabModifierSelector = '[md-saved-tabindex]:not([md-saved-tabindex="-1"])';
+
+          if (isHidden) {
+            attrToSet = 'tabindex';
+            attrToDefault = 'md-saved-tabindex';
+            tabModifierSelector = 'button:not([tabindex="-1"]),\
+              a[href]:not([tabindex="-1"]),\
+              input:not([tabindex="-1"]),\
+              textarea:not([tabindex="-1"]),\
+              select:not([tabindex="-1"]),\
+              [tabindex]:not([tabindex="-1"])';
+          }
+
           for (var i = 0; i < children.length; i++) {
             // skip over child if it is an ascendant of the dialog
             // or a script or style tag
             if (element !== children[i] && !isNodeOneOf(children[i], ['SCRIPT', 'STYLE'])) {
+              var tabbables = children[i].querySelectorAll(tabModifierSelector);
+
               children[i].setAttribute('aria-hidden', isHidden);
+
+              for (var j = 0; j < tabbables.length; j++) {
+                var tabbable = tabbables[j],
+                  attrToSetValue = tabbable.getAttribute(attrToSet) || '0',
+                  attrToDefaultValue = tabbable.getAttribute(attrToDefault) || '-1';
+                tabbable.setAttribute(attrToDefault, attrToSetValue);
+                tabbable.setAttribute(attrToSet, attrToDefaultValue);
+              }
             }
           }
 
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -469,6 +469,26 @@ function MdDialogProvider($$interimElementProvider) {
       }
     };
 
+
+    function ariaFocus(ev) {
+      var dialog = document.querySelector('md-dialog');
+
+      if (isHidden(ev.target)) {
+        ev.stopImmediatePropagation();
+        dialog.focus();
+      }
+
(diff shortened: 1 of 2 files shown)


onumossn:
I signed it!

ThomasBurleson:
@marcysutton - can you review this when you have a chance? Thx.

src/components/dialog/dialog.js
```diff
@@ -734,12 +734,37 @@ function MdDialogProvider($$interimElementProvider) {
...
+
+          if (isHidden) {
+            attrToSet = 'tabindex';
+            attrToDefault = 'md-saved-tabindex';
+            tabModifierSelector = 'button:not([tabindex="-1"]),\
+              a[href]:not([tabindex="-1"]),\
+              input:not([tabindex="-1"]),\
+              textarea:not([tabindex="-1"]),\
+              select:not([tabindex="-1"]),\
+              [tabindex]:not([tabindex="-1"])';
```
marcysutton: Nicely done here, this ensures custom directives will get the same behavior as natively focusable elements.

src/components/dialog/dialog.spec.js
```diff
@@ -885,6 +885,68 @@ describe('$mdDialog', function() {
...
+
+      var template = '<md-dialog aria-label="Some Other Thing">Hello</md-dialog>';
+      var parent = angular.element('<div>');
+      var expectedIndex = [ -1, 0, 1, 0, 3, 0 ];
+      parent.append('<div class="sibling">\
+          <button tabindex="-1"></button>\
+          <a href="b"></a>\
+          <span tabindex="1"></span>\
+          <input/>\
+          <textarea tabindex="3"></textarea>\
```
marcysutton: :+1: for checking positive numbers, as terrible as they are most of the time.

src/components/dialog/dialog.spec.js
```diff
@@ -885,6 +885,68 @@ describe('$mdDialog', function() {
       var sibling = angular.element(parent[0].querySelector('.sibling'));
       expect(sibling.attr('aria-hidden')).toBe('true');
     }));
+
+    it('should set tabindex of to sibling\'s tabbable nodes to -1', inject(function($mdDialog, $rootScope, $timeout) {
+
+      var template = '<md-dialog aria-label="Some Other Thing">Hello</md-dialog>';
+      var parent = angular.element('<a href="a">');
```
marcysutton: Why `<a>` as the parent?

marcysutton:
Love this idea! It is especially helpful since the dialog's focus trap feature was erroneously disabled/removed, which I opened as a bug in https://github.com/angular/material/issues/4105.

Thanks for including test coverage, it looks good. I wonder if there's a way to track performance implications by way of benchmarks. @ThomasBurleson, any idea how we could determine impact of this DOM walking and manipulation?

onumossn:
It seems that the bug #4105 was introduced by the fix of #2443 removing `trapFocus`. If `trapFocus` was adequate from the perspective of this issue, we could just bring that back. Instead of checking for whether or not the focus target is inside the modal, which is what I think was causing issue #2443, we can check if it or its ancestors have aria-hidden="true." That cost should be significantly lower in the common case, and significantly less noticeable by a user of the application. If that would be better, should I create a new pull request with that, or should I continue on this one?

Original `trapFocus`:
```js
function trapFocus(ev) {
  var dialog = document.querySelector('md-dialog');
  if (dialog && !dialog.contains(ev.target)) {
    ev.stopImmediatePropagation();
    dialog.focus();
  }
}

document.addEventListener('focus', trapFocus, true);
```

src/components/dialog/dialog.spec.js
onumossn: @marcysutton I wanted to test that it didn't set tabindex of the parent to -1, so I just went with something tabbable. It could be pulled out into another test, or the test may not even be necessary.

marcysutton:
I like your idea about checking for presence of `aria-hidden` up the DOM tree. The fact that directives like `mdSelect` and `mdMenu` append options in a separate part of the DOM (and most likely outside of the dialog) means the focus trap solution wasn't going to work anyway.

marcysutton:
That actually made me think of a potential bug with the existing DOM walker: in its current state, it is probably going to put `aria-hidden` onto the child options for menus and selects, since it thinks they aren't relevant to the dialog. Hello complexity :(

onumossn:
I think the DOM walker is only run on the open and close of the dialog, so I don't think it would be an issue since the select options are only added to the DOM after you open it and removed once closed.

ThomasBurleson:
@onumossn re > "we can check if it or its ancestors have aria-hidden="true."
Go ahead and add that functionality to this PR and then we can merge. Thx so much for this contribution.

marcysutton:
@onumossn ah yes, that occurred to me after I wrote it. We'll still want to check...maybe a good candidate for a test?

Also kudos for the contribution, it's very much appreciated!

onumossn:
This PR is for the dialog, but I think this problem can be more generalized. As far as I understand, tabbing/focus should never be allowed in aria-hidden regions, so this could be an option in angular-aria. In this case, there is a known element (the dialog) for it to focus on when it attempts to enter an aria-hidden region. In the general case, it should find the next appropriate focusable element and set the focus there. However, I am having trouble thinking of an efficient way of doing that. Any suggestions?

EladBezalel:
Was suggested in another commit, if you do think it's still relevant please rebase and i will reopen

PR #4611, opened 2015-09-13
Add rtl support

hosseinkhaledi:
We add full rtl support to angular material except material.components.slider
All components tested( except material.components.slider) and work correctly in rtl and ltr mode

### DIFF
--- a/src/components/chips/chips.scss
+++ b/src/components/chips/chips.scss
@@ -14,9 +14,9 @@ $contact-chip-name-width: rem(12) !default;
 .md-contact-chips {
   .md-chips {
     .md-chip {
-      padding: $contact-chip-padding;
+      @include rtl(padding, $contact-chip-padding, reverse($contact-chip-padding));
       .md-contact-avatar {
-        float: left;
+        @include rtl(float, left, right);
         img {
           height: $chip-height;
           border-radius: $chip-height / 2;
@@ -25,7 +25,8 @@ $contact-chip-name-width: rem(12) !default;
       .md-contact-name {
         display: inline-block;
         height: $chip-height;
-        margin-left: rem(0.8);
+        @include rtl(margin-left, rem(0.8), null);
+        @include rtl(margin-right, null, rem(0.8));
       }
     }
   }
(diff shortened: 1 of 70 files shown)


ThomasBurleson:
Please:

*  squash your commits, and 
*  make sure you have done a `pull --rebase origin master` (where origin is **https://github.com/angular/material/**).

src/core/style/mixins.scss
```diff
@@ -105,7 +105,15 @@
...
-  #{$prop}: $value;
+  html[dir=ltr] & {
+    #{$prop}: $value;
+  }
+  body[dir=ltr] & {
+    #{$prop}: $value;
+  }
+  html:not([dir]) body:not([dir]) & {
+    #{$prop}: $value;
+  }
```
EladBezalel: Why all of this is necessary if all of them does the same thing?

src/components/chips/chips.scss
```diff
@@ -25,7 +25,8 @@ $contact-chip-name-width: rem(12) !default;
       .md-contact-name {
         display: inline-block;
         height: $chip-height;
-        margin-left: rem(0.8);
+        @include rtl(margin-left, rem(0.8), null);
+        @include rtl(margin-right, null, rem(0.8));
```
EladBezalel: Maybe there's a way to make it smarter?

src/core/style/mixins.scss
hosseinkhaledi: line 108 is to handle html[dir=ltr]
line 111 is to handle body[dir=ltr]
line 114  is to handle case <html><body>... (html and body has not dir) so must be ltr
if do not this we can not reverse ltr to rtl simply for example
margin-left: rem(0.8);
must be 
 @include rtl(margin-left, rem(0.8), null);
@include rtl(margin-right, default_value, rem(0.8));
that default_value depends on other css rule and must be calculated

src/components/chips/chips.scss
hosseinkhaledi: if define rtl with 4 parameter like rlt2(margin-left, rem(0.8), margin-right, rem(0.8))

src/core/style/mixins.scss
EladBezalel: so maybe 
``` css
html[dir=ltr] &, body[dir=ltr] &, html:not([dir]) body:not([dir]) & {
  #{$prop}: $value;
}
```

ThomasBurleson:
@hosseinkhaledi - this is looking very nice. Can you rebase with latest code from master and then make sure the tests pass. Thx.

hosseinkhaledi:
@ThomasBurleson I applied the rebase action

EladBezalel:
Closed with SHA 08bf4f9 #7245

PR #4621, opened 2015-09-14
fix(mdChips): Autocomplete styling is incorrect.

### LINKED ISSUE #4600: md-chips + autocomplete does not look right
Compare the md-chips autocomplete here:

autocomplete-bad

To the regular autcomplete here (last example):
https://material.angularjs.org/latest/#/demo/material.components.autocomplete

autocomplete-good

I may be off, but the text (and label/placeholder when either is empty) looks distractingly displaced in the chips example, and the font size is different too. Should they both not match a regular input in order to give a consistent look and feel?

Tested on Chrome 45 under Fedora 22.

topherfangio:
Remove unnecessary padding from the chips autocomplete and add the `md-input` class to the default autocomplete directive so it matches other inputs.

Also fixes issue with md-chip readonly padding styles.

Fixes #4600.

### DIFF
--- a/src/components/chips/chips.scss
+++ b/src/components/chips/chips.scss
@@ -1,6 +1,7 @@
 $chip-font-size: rem(1.6) !default;
 $chip-height: rem(3.2) !default;
 $chip-padding: 0 rem(1.2) 0 rem(1.2) !default;
+$chip-input-padding: 0 !default;
 $chip-remove-padding-right: rem(2.2) !default;
 $chip-remove-line-height: rem(2.2) !default;
 $chip-margin: rem(0.8) rem(0.8) 0 0 !default;
@@ -62,7 +63,7 @@ $contact-chip-name-width: rem(12) !default;
   &:not(.md-readonly) {
     cursor: text;
 
-    .md-chip {
+    .md-chip:not(.md-readonly) {
       padding-right: $chip-remove-padding-right;
     }
   }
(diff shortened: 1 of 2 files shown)


topherfangio:
@ThomasBurleson @robertmesserle Can you guys think of a reason why we should not automatically add the `md-input` class to the autocomplete's child input element?

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -222,7 +222,8 @@ function MdAutocomplete () {
                   aria-autocomplete="list"\
                   aria-haspopup="true"\
                   aria-activedescendant=""\
-                  aria-expanded="{{!$mdAutocompleteCtrl.hidden}}"/>\
```
ThomasBurleson: String construction using `\` is denigrated in Google Style docs. 
We should be using string concatenation; e.g. [tabsDirective.js](https://github.com/angular/material/blob/master/src/components/tabs/js/tabsDirective.js#L98-L119)

jsenecal:
+1

topherfangio:
@ThomasBurleson I made the requested updates, so this one should also be good. May want to give it a quick once-over though.

Also, you can ignore my comment about the `md-input` class above. I changed the code to not do that.

ThomasBurleson:
@topherfangio - can you rebase plz.

topherfangio:
@ThomasBurleson This should be good to merge now :smile:

PR #4623, opened 2015-09-14
feat(mdInput): Add support for both labels and placeholders.

### LINKED ISSUE #4258: mdInput + mdInputContainer: Floating labels "zooming" on page load
Having an issue where labels are zooming when the form input components are initialized.

The form is not being rendered until the object containing the model is available, so the ng-model values are available immediately.

For eg:
```
<md-input-container flex>
  <label>Length (months)</label>
  <input
  type="number"
  name="length"
  ng-model="vm.plan.length"
  >
</md-input-container>
```

Is resulting in the label "zooming" out to it's floating position when the component is created.

I can switch between routes in an app to watch this happening to the same view over and over again.

### LINKED ISSUE #4462: Cannot have a label and placeholder in ng-input-container
Labels clobber placeholders when `md-input-container` has the `md-no-float` attribute set. Instead of the label should not be clobbering the `placeholder`, the label should remain in label position. 

angular-material-placeholder-label-bug

**See demo:** http://codepen.io/anon/pen/RWNVJz

HTML:
```
<div ng-controller="DemoCtrl" layout="column" class="inputdemoIcons" ng-app="MyApp">
  <br>
  <md-content layout-padding="" class="autoScroll">
    <md-input-container md-no-float>
      <label>Name</label>
      <input ng-model="user.name" type="text" placeholder="Johnny Appleseed">
    </md-input-container>
  </md-content>
</div>
```

JavaScript:
```
angular
  .module('MyApp')
  .controller('DemoCtrl', function($scope) {
    $scope.user = {
      name: '',

    };
  });
```

topherfangio:
Previously, if a user supplied both a placeholder and a label,
the label would float on top of the placeholder when the input
did not have a value.

Fix by adding styles/code to support both at the same time.

**Note:** If the users provides both a label and a placeholder,
the label will no longer animate.

Also fix input styles so transition does not happen if input
already has a value to avoid unneccessary and eratic-looking
animations.

Fixes #4462. Fixes #4258.

### DIFF
--- a/src/components/input/demoBasicUsage/index.html
+++ b/src/components/input/demoBasicUsage/index.html
@@ -53,7 +53,7 @@
         </md-input-container>
         <md-input-container flex>
           <label>Postal Code</label>
-          <input ng-model="user.postalCode">
+          <input ng-model="user.postalCode" placeholder="12345">
         </md-input-container>
       </div>
 
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -27,8 +27,10 @@ angular.module('material.components.input', [
  * Input and textarea elements will not behave properly unless the md-input-container
  * parent is provided.
  *
- * @param md-is-error {expression=} When the given expression evaluates to true, the input container will go into error state. Defaults to erroring if the input has been touched and is invalid.
- * @param md-no-float {boolean=} When present, placeholders will not be converted to floating labels
+ * @param md-is-error {expression=} When the given expression evaluates to true, the input container
+ *   will go into error state. Defaults to erroring if the input has been touched and is invalid.
+ * @param md-no-float {boolean=} When present, placeholders will not be converted to floating
+ *   labels.
  *
  * @usage
  * <hljs lang="html">
@@ -55,6 +57,7 @@ function mdInputContainerDirective($mdTheming, $parse) {
   function postLink(scope, element, attr) {
     $mdTheming(element);
   }
+
   function ContainerCtrl($scope, $element, $attrs) {
     var self = this;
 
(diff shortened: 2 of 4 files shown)


topherfangio:
@ThomasBurleson Not sure who else besides you needs to review this.

jelbourn:
LGTM

ThomasBurleson:
@topherfangio - can you update this PR with latest changes from @jelbourn ( now in master ). Specifically we should use `md-placeholder` and `md-active` styles.

src/components/input/demoBasicUsage/index.html
```diff
@@ -53,7 +53,7 @@
         </md-input-container>
         <md-input-container flex>
           <label>Postal Code</label>
-          <input ng-model="user.postalCode">
+          <input ng-model="user.postalCode" placeholder="12345">
```
ThomasBurleson: This uses the `placeholder` directive. But what uses/inserts the `md-placeholder` style?

topherfangio: The styling is added dynamically based on whether they include both the `label` and the `placeholder` which are standard attributes. I think the idea was to make the inputs as close to a standard HTML input as possible.

It's all done/decided in the bottom of `input.js` in the `function placeholderDirective()` and documented in the input docs.

Is this what you were asking?

ThomasBurleson:
@topherfangio - can you rebase and update this PR.

topherfangio:
@ThomasBurleson Rebased. And sorry for the confusion, were you suggesting we change the class from `md-input-has-placeholder` to just `md-placeholder`? I think the former is a bit more clear and helps ensure that we don't have a conflict later.

PR #4629, opened 2015-09-14
fix typo line 78

Kielan:


DevVersion:
Already fixed in `master`

Issue #4622, opened 2015-09-14
layout attribute no longer works with ng-attr in 0.11.0 or using a variable


langan:
When you use the following markup

```HTML
<div ng-attr-layout="{{vm.widgetLayout}}"></div>
```

The div gets assigned the class

```
layout-{{vm.widgetLayout}}
```

I've added some karma tests to look for this, would be happy to fix it if someone can give me some pointers on the best way to evaluate the attribute in layout.js

Pretty sure its line 165 of layout.js that needs to change

```JS
element.addClass(className + "-" + attr[directive].replace(/\s+/g, "-"));
```

attr[directive] needs to be evaluated I think

carlosafonso:
Could this somehow be related to #4520?

langdonx:
All the layout/flex attributes are now directives that just populate classes on the element for you.  This was done to fix performance in IE11 (where it suffered using so many attribute selectors).  Is there some reason you're needing to use ng-attr?  If so, I would switch to using the material layout classes for now.

langan:
We can drop the ng-attr I suppose but shouldn't the layout directive work with a variable.

I made a quick test using a variable to change the layout from row to column and the classes aren't updated

```HTML
<div ng-init="layout = 'column'">
    <div layout="{{layout}}">
        <span>Top</span>
        <span>Bottom</span>
    </div>

    <md-button ng-click="layout='row'">Row</md-button>
    <md-button ng-click="layout='column'">Column</md-button>

    <code>{{ layout }}</code>
</div>
```

jelbourn:
@ThomasBurleson was it ever the intent to support dynamic layouts via data binding?

ThomasBurleson:
@jelbourn - Originally, no... The pending PR, however, has support for dynamic attribute values.

langan:
@jelbourn The comment in layout.js suggests that this was planned to work

```JS
          /**
           * After link-phase, do NOT remove deprecated layout attribute selector.
           * Instead watch the attribute so interpolated data-bindings to layout
           * selectors will continue to be supported.
           *
           * $observe the className and update with new class (after removing the last one)
           *
           * e.g. `layout="{{layoutDemo.direction}}"` will update...
           */
```

ThomasBurleson:
This has been fixed and will be merged into master soon.

ThomasBurleson:
See SHA 7264cab40349

PR #4638, opened 2015-09-15
Fix attribute directives (layout, flex, etc) observe attribute value

langan:
I think this fixes the issue #4622, the attribute value wasn't being observed so the $observe code wasn't being executed when the attribute value changed.

### DIFF
--- a/src/core/services/layout/layout.js
+++ b/src/core/services/layout/layout.js
@@ -161,7 +161,7 @@
         }
 
         if (attr[directive]) {
-          element.addClass(className + "-" + attr[directive].replace(/\s+/g, "-"));
+          element.addClass(className + '-' + attr[directive].replace(/\s+/g, '-'));
         }
 
         if ( scope ) {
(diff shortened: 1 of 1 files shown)


langan:
@ThomasBurleson will this be added in the 0.11.1 release?

ThomasBurleson:
@langan - in the future, please do not replace the string quotes. That creates huge `diff clutter`

ThomasBurleson:
This feature will be in the next release: v0.11.1-rc1 and greater.

langan:
@ThomasBurleson whoops sorry, just thought it looked odd using double quotes on those particular lines when the majority of code uses single.

Will make sure it doesn't happen again.

There is nothing in the [coding styles](https://github.com/angular/material/blob/master/docs/guides/CODING.md) about quotes.

Do you need me to add any unit tests for observing a variable?  Looks like layout.spec.js doesn't cover that case at the moment.

ThomasBurleson:
See pending PR https://github.com/angular/material/pull/4698/files#diff-08c7a16dd7bb0b15845b6b6b6ae637c5R191 for the process to watch interpolated attribute values.

ThomasBurleson:
Merged and closed with SHA 7264cab40349

PR #4472, opened 2015-09-04
Support for ng-message-exp directive.

bopm:
Support for ng-message-exp directive. Fixes #1793.

PR #4008, opened 2015-08-03
support ng-message-exp in css selectors

iksose:
https://docs.angularjs.org/api/ngMessages/directive/ngMessageExp

PR #4647, opened 2015-09-15
fix(mdInput): Support multiple ng-messages simultaneously.

### LINKED ISSUE #1793: Material design on ng-message-exp
Hi

Since I think v1.4 of angular js ngMessage as been improve with the new directive "ng-message-exp"
It works great and does what is expected
Problem, this new directive is not included in angular material and will not add any style.
Would be good if it can be corrected in the new version

Thanks

### LINKED ISSUE #1957: md-input-container(firefox): input type="number" doesn't fill width
The discount rate input has `type="number"` which doesnt fill it's parents width in firefox as you can see below.
image

Note input `type="text"` or not specifiying a type works as expected which you can see for the `name` field. I did a dif between the input with `type="number"` and without a type and the only added style was `-moz-apperance: none;` on the one with `type="number". I also tried adding `flex` to the input but that didn't work. Also the `md-input-container` still takes up the full width so it's a problem with the `input` styles.

### LINKED ISSUE #2648: ng-messages with multiple flag not working correctly inside md-input-container
Multiple message will show in one line, message will mass with each other.
test on Chrome.

screen shot 2015-05-01 at 6 07 53 am
screen shot 2015-05-01 at 6 08 52 am

topherfangio:
Previously, multiple `ng-message`s would render on top of each other. Fix by altering CSS position and altering transition to support multiple messages (i.e. potentially varying height).

Also some other small fixes to inputs/errors:

 * Fix number input widths in Firefox
 * Update errors demo messages to be more dynamic and show multiple errors
 * Update SCSS to allow `ng-message-exp` and associated `data-` and `x-` attributes
 * Add demo using `ng-message-exp` to show new SCSS styles being applied

Fixes #2648. Fixes #1957. Fixes #1793.

Should also close PRs #4472 and #4008. Thanks to @bopm and @iksose for the initial PRs!

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -11,7 +11,8 @@ angular.module('material.components.input', [
   .directive('input', inputTextareaDirective)
   .directive('textarea', inputTextareaDirective)
   .directive('mdMaxlength', mdMaxlengthDirective)
-  .directive('placeholder', placeholderDirective);
+  .directive('placeholder', placeholderDirective)
+  .directive('ngMessages', ngMessagesDirective);
 
 /**
  * @ngdoc directive
@@ -68,6 +69,9 @@ function mdInputContainerDirective($mdTheming, $parse) {
     };
     self.setHasValue = function(hasValue) {
       $element.toggleClass('md-input-has-value', !!hasValue);
+    };
+    self.setHasMessages = function(hasMessages) {
+      $element.toggleClass('md-input-has-messages', !!hasMessages);
     };
     self.setInvalid = function(isInvalid) {
       $element.toggleClass('md-input-invalid', !!isInvalid);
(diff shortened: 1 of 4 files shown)


topherfangio:
@ThomasBurleson @jelbourn Would love for you both to review; in order to accommodate multiple lines of errors, I had to alter the transition. It looks different than before, but I think it works and looks pretty good. Definitely open to suggestions though.

src/components/input/input.js
```diff
@@ -408,3 +412,24 @@ function placeholderDirective($log) {
 
   }
 }
+
+function ngMessagesDirective() {
+  return {
+    restrict: 'EA',
+    require: '^^?mdInputContainer',
```
jelbourn: Add a comment like
```
This is optional because we don't want target *all* ngMessage instances, just those inside of mdInputContainer.
```

jelbourn:
LGTM

topherfangio:
@jelbourn Thanks! I'll update with your other comment suggestion after lunch :-)

topherfangio:
@ThomasBurleson I just updated this PR with a few more additions which should resolve some existing/outstanding PRs where we were waiting on demos to be updated.

Issue #4631, opened 2015-09-14
md-datepicker placeholder color doesn't match other placeholders


calebegg:
See screenshot.
![download 5](https://cloud.githubusercontent.com/assets/782920/9860183/82df7d5e-5af8-11e5-92fd-6de76c0b8657.png)

jelbourn:
Closed via 416dc4c023da00a78c9779ea9b958c052d44d353

PR #4648, opened 2015-09-15
fix(datepicker): ngModel updates for empty input. Fixes #4643

### LINKED ISSUE #4643: md-datepicker doesn't change model value when date is empty
sometimes we need to remove dates from model. but we can't do it with removing date in datepicker
ng-change doesn't trigger.
https://github.com/angular/material/blob/master/src/components/datepicker/datePicker.js#L334
i guess we should allow to pass empty values in this expression

jelbourn:
Also sneak in a tiny fix for #4631(placeholder text color), fix the wrong placeholder attributes in the demo, and tweak the datepicker open animation.

@topherfangio or @ThomasBurleson to review

### DIFF
--- a/src/components/datepicker/datePicker.js
+++ b/src/components/datepicker/datePicker.js
@@ -331,7 +331,11 @@
     var parsedDate = this.dateLocale.parseDate(inputString);
     this.dateUtil.setDateTimeToMidnight(parsedDate);
 
-    if (this.dateUtil.isValidDate(parsedDate) &&
+    if (inputString === '') {
+      this.ngModelCtrl.$setViewValue(null);
+      this.date = null;
+      this.inputContainer.classList.remove(INVALID_CLASS);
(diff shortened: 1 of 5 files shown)


src/components/datepicker/datePicker.js
```diff
@@ -331,7 +331,11 @@
     var parsedDate = this.dateLocale.parseDate(inputString);
     this.dateUtil.setDateTimeToMidnight(parsedDate);
 
-    if (this.dateUtil.isValidDate(parsedDate) &&
+    if (inputString === '') {
```
ThomasBurleson: This should have a comment regarding the intent (as it is not clear).

jelbourn: I see this as being very clear from just the code; "if the input is empty, the model is set to null, which is a valid value."

PR #4681, opened 2015-09-17
bug: Keyboard handlers not removed if speed-dial is removed from DOM, breaks spacebar

hannahhoward:
Our project uses a speed dial where clicking on the actions actually causes the containing block to be removed from the dom (due to an ng-if). The result is the document still has a global keydown handler that causes all spacebar presses to be ignored! Seems like the keyboard handlers should be removed if the element is removed from the DOM.

As an aside, a global keydown handler that disables spacebar seems like a dangerous choice. What if I have a speed dial open but want to enter text somewhere else on the page.... I can't use a space bar anymore? That doesn't seem like a good idea.

### DIFF
--- a/src/components/fabSpeedDial/fabController.js
+++ b/src/components/fabSpeedDial/fabController.js
@@ -59,6 +59,9 @@
         angular.forEach(eventTypes, function(eventType) {
           $element.off(eventType, parseEvents);
         });
+        // remove any attached keyboard handlers in case element is removed while
+        // speed dial is open
+        disableKeyboard();
       });
     }
 


ThomasBurleson:

topherfangio:
LGTM

@hannahhoward The speed dial should be treated like any other menu/select/interim element where it gains focus when open so that you can interact with it via the keyboard. If you select an input, this should blur the component closing it and removing the keyboard capture.

Do you see other elements working in a different way?

hannahhoward:
Ok that makes sense. I can understand why it works that way. Still think it makes sense to make sure any keyboard handlers are removed if the element is destroyed from the DOM.

topherfangio:
@hannahhoward re handlers destroyed: Absolutely! This was definitely an oversight; many thanks for the patch! :-)

PR #4592, opened 2015-09-11
feat(layout): allow flex-order to be negative

flowersinthesand:
Fixes #4482.

PR #4698, opened 2015-09-17
fix(layout): updates to flexbox styles, directives

### LINKED ISSUE #4482: Allow flex-order to be negative
According to [CSS Flexible Box Layout Module Level 1](http://www.w3.org/TR/css3-flexbox/#order-property) and [CSS Values and Units Module Level 3](http://www.w3.org/TR/css-values/#integers), the `order` property can be negative, -9 to -1, as well as positive, 1 to 9, and zero.

With this feature, developers can use positive and negative flex-orders for general (search results) and special (ads) flex items respectively.

Here's a demo based on v0.11.0-rc2 with a simple patch for this feature request:
http://codepen.io/flowersinthesand/pen/ojgrmj

ThomasBurleson:
* allow flex-order to be negative
* add support to observe interpolated layout values.
*  restore max-width for layouts
  *  restore `layout > flex` specifiers of max-width or max-height.
  * publish `$$mdLayout` service to allow Attribute selectors to be removed when the translation inject to Class selector finishes.
*  add `md-css-only` feature , 
  *  support use of `<body class="md-css-only">`  to use Layout features **without** any JS requirements or `ngMaterial` module dependencies.
    *  May use standard Attribute selectors
    *  May use class selectors
    *  Disables Layout directive postlinks for optimization

According to CSS Flexible Box Layout Module Level 1 and CSS Values and Units Module Level 3, the order property can be negative, -9 to -1, as well as positive, 1 to 9, and zero.

Fixes #4482. Closes #4592.

> Thx @flowersinthesand for original PR

### DIFF
--- a/src/core/services/layout/layout.js
+++ b/src/core/services/layout/layout.js
@@ -1,6 +1,8 @@
 (function () {
   'use strict';
 
+  var _$$mdLayout, _$parse;
+
     /**
      *
      *   The original ngMaterial Layout solution used attribute selectors and CSS.
@@ -41,6 +43,26 @@
      *  ```
      */
     angular.module('material.core.layout', [ 'ng' ])
+
+      /**
+       * Model of flags used by the Layout directives
+       * Allows changes while running tests or runtime app changes
+       */
+      .factory("$$mdLayout", function() {
+        return {
+          /**
+           * Should we remove the original layout Attribute selectors
+           * after translation injection
+           */
+          removeAttributes : true,
+
+          /**
+           * Special internal flag used to optimize
+           * noop(s) for the directive postLinks below
+           */
+          disablePostLinks : undefined
+        };
+      })
 
       // Attribute directives with optional value(s)
 
@@ -136,17 +158,24 @@
      * @param {boolean=} addDirectiveAsClass
      */
     function attributeWithObserve(className, addDirectiveAsClass) {
-      return function() {
+      return ['$$mdLayout', '$document', '$parse', function($$mdLayout, $document, $parse) {
+        _$$mdLayout = $$mdLayout;
+        _$parse = $parse;
+
         return {
             restrict : 'A',
             compile: function(element, attr) {
+              if ( postLinkIsDisabled($document[0]) ) return angular.noop;
+
               attributeValueToClass(null, element, attr);
 
               // Use for postLink to account for transforms after ng-transclude.
               return attributeValueToClass;
             }
         };
-      };
+      }];
+
+
 
       /**
        * Add as transformed class selector(s), then
@@ -154,14 +183,19 @@
        */
       function attributeValueToClass(scope, element, attr) {
         var directive = attr.$normalize(className);
+        var attrExpression = _$parse( directive[directive] );
 
         // Add transformed class selector(s)
         if (addDirectiveAsClass) {
           element.addClass(className);
         }
 
-        if (attr[directive]) {
+        if ( attr[directive] ) {
           element.addClass(className + "-" + attr[directive].replace(/\s+/g, "-"));
+
+          if ( _$$mdLayout.removeAttributes ) {
+            element.removeAttr(directive);
+          }
         }
 
         if ( scope ) {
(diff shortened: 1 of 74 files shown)


ThomasBurleson:

src/core/services/layout/layout.js
```diff
@@ -42,25 +44,41 @@
      */
     angular.module('material.core.layout', [ 'ng' ])
 
+      /**
+       * Model of flags used by the Layout directives
+       * Allows changes while running tests or runtime app changes
+       */
+      .factory("$$mdLayout", function() {
```
ThomasBurleson: Should probably be a provider to allow config changes but not runtime changes.

src/core/services/layout/layout.js
```diff
@@ -106,21 +118,36 @@
...
       .directive('showGtSm'            , attributeWithoutValue('show-gt-sm')         )
-      .directive('showLtMd'            , warnAttrNotSupported ('show-lt-md')         )
       .directive('showMd'              , attributeWithoutValue('show-md')            )
       .directive('showGtMd'            , attributeWithoutValue('show-gt-md')         )
-      .directive('showLtLg'            , warnAttrNotSupported ('show-lt-lg')         )
       .directive('showLg'              , attributeWithoutValue('show-lg')            )
-      .directive('showGtLg'            , attributeWithoutValue('show-gt-lg')         );
+      .directive('showGtLg'            , attributeWithoutValue('show-gt-lg')         )
+
+      // !! Deprecated attributes: use the `-lt` (aka less-than) notations
```
ThomasBurleson: Provides `$log.warn( )` details useful for migration

src/core/services/layout/layout.js
```diff
@@ -239,4 +290,35 @@
 
     }
 
+    /**
```
ThomasBurleson: API addition:

*  Allows use of Attribute selectors for browsers that do not suffer performance degradations
*  Allows use of ngMaterial Layout/Flexbox CSS **without** loading the `angular-material.js` source.

src/core/services/layout/layout.spec.js
```diff
@@ -111,4 +123,32 @@ describe('layout directives', function() {
...
+      });
+
+      expect(element.hasClass('flex-gt-md-32')).toBe(true);
+
+      scope.$apply(function() {
+        scope.size = "fishCheeks";
+      });
+
+      expect(element.hasClass('flex-gt-md-32')).toBe(false);
+      expect(element.hasClass('flex-gt-md-fishCheeks')).toBe(true);
```
ThomasBurleson: Maybe we should check the validity of the value to see if the value (for the attribute) is an acceptable, known value; warn if not?

src/core/services/layout/layout.scss
```diff
@@ -19,71 +19,90 @@
...
 }
 @-moz-document url-prefix() {
-  .layout-fill {
+  .layout-fill, .md-css-only [layout-fill] {
     margin: 0;
     width: 100%;
-    min-height: auto;
-    height: inherit;
+    min-height: 100%;
+    height: 100%;
```
topherfangio: I don't think it needs a comment, but can you briefly explain why these were changed from auto/inherit to 100%?

src/core/services/layout/layout.scss
```diff
@@ -154,28 +178,42 @@
...
+  .layout-row, .md-css-only [layout="row"] {
+    > .#{$flexName}-33, > .#{$flexName}-34 { max-width: 33%; }
+    > .#{$flexName}-66, > .#{$flexName}-67 { max-width: 67%;  }
   }
-  .layout-column {
-    > .#{$flexName}-33, > .#{$flexName}-34 {  }
-    > .#{$flexName}-66, > .#{$flexName}-67 {  }
+  .layout-column, .md-css-only [layout="column"] {
+    > .#{$flexName}-33, > .#{$flexName}-34 { max-height: 33%;  }
+    > .#{$flexName}-66, > .#{$flexName}-67 { max-height: 33%; }
```
topherfangio: Shouldn't this be `max-height: 67%`? Copy/paste issue?

src/core/services/layout/layout.spec.js
topherfangio: I'm not so sure we should check. If someone wants to add to "extend" our CSS with their own custom sizes, it seems like that would generate a lot of warnings. Also, I think if it doesn't match, then they will have a visual warning by the fact that their site's layout is broken :-)

topherfangio:
LGTM aside from my comments.

ThomasBurleson:
Closed. See new PR #4760

Issue #4586, opened 2015-09-11
Whiteframe


ThomasBurleson:
We no longer talk about depth in terms of z1-z5. we should talk about depth in dp values from the background following these shadow levels: http://codepen.io/shyndman/pen/MaYoaV

![screen shot 2015-09-11 at 9 06 54 am](https://cloud.githubusercontent.com/assets/210413/9816750/75e934d2-5864-11e5-92e7-264500673a5a.png)

PR #4706, opened 2015-09-18
style(whiteframe): working with dp insted of z-index

EladBezalel:
Changed z-index to dp as described in #4586


ThomasBurleson:
:+1:  Nice.

EladBezalel:
Notice it's supporting past versions, so when upgrading it won't break anything.

ThomasBurleson:
Yes:

```css
.md-whiteframe-7dp, .md-whiteframe-z3 {
  box-shadow: $whiteframe-shadow-7dp;
}
```
That is a very nice **attention to details***. :+1:

PR #4724, opened 2015-09-19
fix(autocomplete): support ng-required

### LINKED ISSUE #2949: mdAutocomplete should support ng-required
mdAutocomplete does not pass ng-required to the underlying input. This causes problems performing form validation.

loic:
Closes #2949

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -1154,4 +1154,66 @@ describe('<md-autocomplete>', function() {
     }));
   });
 
+  describe('required', function () {
+    it('should respect the required attribute', inject(function () {
+      var scope    = createScope();
+      var template = '\
+          <form name="searchForm">\
+          <md-autocomplete\
+              required\
+              md-input-name="autocompleteField"\
+              md-selected-item="selectedItem"\
+              md-search-text="searchText"\
+              md-items="item in match(searchText)"\
+              md-item-text="item.display"\
+              placeholder="placeholder">\
+            <span md-highlight-text="searchText">{{item.display}}</span>\
+          </md-autocomplete>\
+          </form>';
+      var element  = compile(template, scope);
+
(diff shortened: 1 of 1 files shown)


topherfangio:
Can you please provide a test case that shows exactly what this fixes?

loic:
Added tests, unless I missed it, required wasn't tested for. I couldn't find any tests for the rest of the validation stack either, e.g. minlength, maxlength, ng-messages, ng-disabled, etc.

EladBezalel:
@loic please rebase

loic:
It looks like 32ba92f44e5ffcb18f0663074d7c9c1913e72836 addressed the issue. I rebased the tests in case they are of interest, otherwise feel free to close the PR.

DevVersion:
@loic Thanks for the updates. I like those tests, they're additionally testing the behavior of required autocomplete's  inside of the form validation.

src/components/autocomplete/autocomplete.spec.js
```diff
@@ -1154,4 +1154,66 @@ describe('<md-autocomplete>', function() {
     }));
   });
 
+  describe('required', function () {
```
DevVersion: Please notice, that we already have a section for `required`.

https://github.com/angular/material/blob/master/src/components/autocomplete/autocomplete.spec.js#L1035

loic: Good catch, it didn't exist when I made the original PR, let me move it around.

DevVersion: Understood, thanks :)

loic:
Actually both seem already tested https://github.com/angular/material/blob/master/src/components/autocomplete/autocomplete.spec.js#L1066-L1112. I think we can close this PR.

DevVersion:
@loic I agree, thanks for your contribution. I'm sorry, we went ahead with another PR.

PR #4726, opened 2015-09-19
fix(input): Sizes textareas properly when the container is shown

### LINKED ISSUE #1202: md-input-container: Textarea problems with model initialization when element not visible on view load
Overview of the Issue:
There are some issues with updating the model for textareas contained in md-input-containers. The view model is probably not getting updated since the view isn't visible upon initialization of the page. 

We've seen this same thing a lot with different libraries since updating to AngularJS 1.3.x. Here's a similar issue with angular-bootstrap-switch (https://github.com/frapontillo/angular-bootstrap-switch/issues/52). We've also seen it with ui-select2.

We were able to solve both issues using $intervals that just sit and wait for the view to become visible and then initialize the model properly. But this seems like a horrible hack, so we're hoping to hear about a better fix for this.

Angular Material Version: master, 0.7.0-rc2, 0.7.0-rc3
Browsers and Operating System: Chrome 39 and 42
Reproduce the Error: http://plnkr.co/edit/cV2xejsabJIhM70IC6Zg?p=preview
Related Issues: #559 #565

mhchen:
Added an optional md-detect-hidden attribute for textareas inside of
md-input-containers. This will check on every digest cycle whether the
element was hidden and is now shown, and will auto-size the textarea
properly.

Fixes #1202

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -126,6 +126,7 @@ function labelDirective() {
  * @param {string=} placeholder An alternative approach to using aria-label when the label is not
  *   PRESENT. The placeholder text is copied to the aria-label attribute.
  * @param md-no-autogrow {boolean=} When present, textareas will not grow automatically.
+ * @param md-detect-hidden {boolean=} When present, textareas will be sized properly when they are revealed after being hidden. This is off by default for performance reasons because it guarantees a reflow every digest cycle.
  *
  * @usage
  * <hljs lang="html">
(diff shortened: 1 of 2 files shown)


mhchen:
@ThomasBurleson Please review

src/components/input/input.js
```diff
@@ -341,6 +342,31 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
         var height = node.offsetHeight + line;
         node.style.height = height + 'px';
       }
+
+      // Attach a watcher to detect when the textarea gets shown.
+      if (angular.isDefined(element.attr('md-detect-hidden'))) {
```
ThomasBurleson: Let's refactor this to a separate function.

ThomasBurleson:
@mhchen - Should we document the `md-detect-hidden` feature of the textArea?

mhchen:
@ThomasBurleson I meant to ask you about documentation. I definitely think we should do it, but should it go in the same place as the rest of the attributes? It still seems to me like a fairly uncommon use case and I like that our attributes documentation is nice and concise right now. If you think it should be documented next to the rest of the attributes, happy to put it there.

I can open a new PR with the refactor and the documentation.

PR #4740, opened 2015-09-21
Fixed incorrect and duplicate layout tests

arush:
There were a number of typos and incorrect combinations of values in the alignmentValues array that didn't correspond to the layout.scss causing incorrect tests to be run. There were also duplicates, there were 31 values, now 19.


arush:
I signed it!

ThomasBurleson:
:+1:

PR #4741, opened 2015-09-21
chore(npm): add standard package.json fields

### LINKED ISSUE #3552: Add license field to package.json
I noticed a build warning that there's no license field in package.json.

See [this](https://travis-ci.org/angular/material/builds/69292268) or any other build on travis under `npm install`.

`npm WARN package.json angular-material-source@0.10.0 No license field.`

Splaktar:
Fix license warnings from npm
Add a useful `npm start` script. It opens the docs in your default browser
Add other standard fields as detailed in https://docs.npmjs.com/files/package.json

Closes #3552.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,7 +1,13 @@
 {
   "name": "angular-material-source",
   "version": "0.11.0",
+  "description": "The Angular Material project is an implementation of Material Design in Angular.js. This project provides a set of reusable, well-tested, and accessible UI components based on the Material Design system. Similar to the Polymer project's Paper elements collection, Angular Material is supported internally at Google by the Angular.js, Material Design UX and other product teams.",
+  "keywords": "client-side, browser, material, material-design, design, angular, css, components, google",
+  "homepage": "https://material.angularjs.org",
+  "bugs": "https://github.com/angular/material/issues",
+  "license": "MIT",
   "repository": {
+    "type" : "git",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -59,6 +65,7 @@
     "through2": "^0.6.3"
   },
   "scripts": {
+    "start": "gulp server -o http://localhost:8080/dist/docs/index.html",
```
topherfangio: Would it make more sense to use `gulp site watch --dev` here?

topherfangio:
@ThomasBurleson LGTM; would love your thoughts on the what the `start` command should be.

package.json
ThomasBurleson: changed to `"watch": "gulp watch site --dev",`

Splaktar: The name of the command was `start` because that is a [standard command](https://docs.npmjs.com/cli/start) to start projects in Node, i.e. `npm start`. I was hoping that this would make the initial process of getting started with the project a little easier for some people. Either way, someone familiar with Node will still find this watch script and be able to run it via `npm runScript watch`.

Splaktar: It would also be nice if this command would auto-open the docs site in the browser. `gulp-connect` doesn't have that feature without using the `opn` package, but I'm not sure why we're using both `gulp-connect` and `gulp-webserver`. We should probably pick one.

PR #4754, opened 2015-09-21
check for undefined before trying to compute animation style values, …

### LINKED ISSUE #4752: Menu TypeError: undefined is not an object (evaluating 'str.split')
On Safari (Mac), and Chrome on Linux when opening a menu with dynamic items the menu does not complete it's animation and throws:

        [Error] TypeError: undefined is not an object (evaluating 'str.split')
	    parseMaxTime (angular-material.js, line 5000)
	    computeTimings (angular-material.js, line 4975)
	    (anonymous function) (angular-material.js, line 4878)
	    (anonymous function) (angular-material.js, line 5030)

When running the debugger, the problem starts in computeTimings(element) function:
https://github.com/angular/material/blob/a5d84c372f87e96469f6c7d72d42d0aea83fe8fd/src/core/util/animation/animateCss.js#L293

    // angular-material.js L5097
    var adr = parseMaxTime(cs[prop('animationDuration')]);

no animationDuration property is available, so when parseMaxTime runs it bombs out:
https://github.com/angular/material/blob/a5d84c372f87e96469f6c7d72d42d0aea83fe8fd/src/core/util/animation/animateCss.js#L318

    // angular-material.js L5122
    var values = str.split(/\s*,\s*/); //str = undefined

This results in the entire screen blocked requiring a full page refresh. I tried adding ng-init with values as outlined here: https://github.com/angular/material/issues/4720, however it did not help. My menu is in a directive with a dynamic ng-repeat of elements. Everything works fine in Chrome and Firefox on Mac/Windows, only occurs on Safari/Mac and Chrome/Linux.

### FIX
Wrapping an if (str) { ... } around 
https://github.com/angular/material/blob/a5d84c372f87e96469f6c7d72d42d0aea83fe8fd/src/core/util/animation/animateCss.js#L318-L327

Fixed the problem for me.

mlabieniec:
Check for str before trying to compute cs values. Some values are not present in other browsers/OS's.
…Fixes #4752

### DIFF
--- a/src/core/util/animation/animateCss.js
+++ b/src/core/util/animation/animateCss.js
@@ -315,16 +315,18 @@ if (angular.version.minor >= 4) {
 
       function parseMaxTime(str) {
         var maxValue = 0;
-        var values = (str || "").split(/\s*,\s*/);
-        forEach(values, function(value) {
-          // it's always safe to consider only second values and omit `ms` values since
-          // getComputedStyle will always handle the conversion for us
-          if (value.charAt(value.length - 1) == 's') {
-            value = value.substring(0, value.length - 1);
-          }
-          value = parseFloat(value) || 0;
-          maxValue = maxValue ? Math.max(value, maxValue) : value;
-        });
+        if (str) {
+          var values = (str || "").split(/\s*,\s*/);
+          forEach(values, function(value) {
+            // it's always safe to consider only second values and omit `ms` values since
+            // getComputedStyle will always handle the conversion for us
+            if (value.charAt(value.length - 1) == 's') {
+              value = value.substring(0, value.length - 1);
+            }
+            value = parseFloat(value) || 0;
+            maxValue = maxValue ? Math.max(value, maxValue) : value;
+          });
+        }
         return maxValue;
       }
 


src/core/util/animation/animateCss.js
```diff
@@ -315,16 +315,18 @@ if (angular.version.minor >= 4) {
...
-        forEach(values, function(value) {
-          // it's always safe to consider only second values and omit `ms` values since
-          // getComputedStyle will always handle the conversion for us
-          if (value.charAt(value.length - 1) == 's') {
-            value = value.substring(0, value.length - 1);
-          }
-          value = parseFloat(value) || 0;
-          maxValue = maxValue ? Math.max(value, maxValue) : value;
-        });
+        if (str) {
```
topherfangio: Can you explain how this `if` statement actually fixes the issue (ideally with a test)?

The following `(str || "")` should catch the case of str being undefined/null and make the code work. What is the value of `str` that is causing this to fail?

mlabieniec:
Ah sorry about that, I completely scanned over the fact that this was fixed in master, it is not in the 11.0.0 branch
https://github.com/angular/material/blob/v0.11.0/src/core/util/animation/animateCss.js#L318

Issue #4707, opened 2015-09-18
md-select closes when scrolling and mousing up over the menu container


rrajewski:
MouseUp events over the menu container (not just the scrollbar) should be ignored and not cause an auto-close. The menu is closing while dragging the scroll bar and mousing up over the options.

![](https://i.imgflip.com/r7oxe.gif)

See #4229 and #4078

rschmukler:
This has been fixed as part of the gesture enhancements

PR #4767, opened 2015-09-22
fix(mdSelect): Do not close on scroll

FrankMeyer:
Do not close the md-select when scrolled and mouseup event is above the md-options.
Remember the mouse x-position of the mousedown event.
Take the remembered x-position instead of the mouseup-x-position to check if the user has scrolled or not.
Do not close the select menu when scrolling is detected.

See issue #4707 and #4078


src/components/select/select.js
```diff
@@ -1136,11 +1142,11 @@ function SelectProvider($$interimElementProvider) {
            */
           function mouseOnScrollbar() {
             var clickOnScrollbar = false;
-            if (ev && (ev.currentTarget.children.length > 0)) {
+            if (mouseDownXPosition != null && ev && (ev.currentTarget.children.length > 0)) {
```
ThomasBurleson: Should we not also reset `mouseDownXPosition = null` on 'mouseup' ?

FrankMeyer: I htought it would not be neccessary because there is always the mousedown before the mouseup, but maybe that is not the case when the mousedown was outside the mdSelect. I'll add that.

ThomasBurleson: Agreed that such scenarios are edge cases. I like, however, to always *balance* my event logic: if one (1) action sets a state then a *paired* action should clear that state.

FrankMeyer: Completely understandable. Another question btw: Would it be a better way, or more angular way, to make a container element with a content child to have a clear distinction between mouseevents? Container mouseevent would then always originate from the scrollbar and content mouseevent always from a mdOption.

ThomasBurleson: That sounds like a good idea. Can you add that to this PR?

ThomasBurleson: @FrankMeyer - While that feature may be more complex and require significant regression issues, my intuition suggests that it may be a better long term solution.

FrankMeyer: Yes you are right, so should this solution be merged now and revised later on? Or should I revise it first?

ThomasBurleson: Let me merge this one. Then, if you finish the **container** enhancement, you can submit another PR. Does that sound reasonable?

> BTW, thanks for all your great comments, issues, research, and PR(s).

FrankMeyer: Yes, I can do that in the beginning of next week.

ThomasBurleson: I will merge this now. With your next PR, plz provide a paired test for each change (and perhaps one for this fix here). Thx again.

FrankMeyer: Hello ThomasBurleson,
I ve not made the check with another container Element. It was much much easier.
I skipped all the mouse position calculations and added a one-liner.
The mouse event already provided the originating element, so this was perfekt to check whether an event belongs to scrolling or not. I don't know why I ve not seen that before.

FrankMeyer:
@ThomasBurleson ,
I have not made the check with another container Element. It was much much easier.
I skipped all the mouse position calculations and added a one-liner.
The mouse event already provided the originating element, so this was perfekt to check whether an event belongs to scrolling or not. I don't know why I ve not seen that before.
The changes are in this PR.

EladBezalel:
Please check this against HEAD, closing for now, if it's still valid please rebase and squash your commits and i will reopen

FrankMeyer:
The problem does not occur anymore, this can stay closed.

Issue #4558, opened 2015-09-11
md-datepicker usability on mobile devices


matthiasg:
when opening a popup such as the new md-datepicker the popup should be positioned centered and fullscreen and not aligned with the original field which doesnt really work on mobile. 

this is also needed for md-dialog of course but there is a related issue for that https://github.com/angular/material/issues/2148 

generally a popup directive/service which manages this might be helpful

without full screen (such as the native picker / which might be a sensible fallback) the picker is cut off, requiring to pan around...

![image](https://cloud.githubusercontent.com/assets/231889/9809644/66a316cc-586a-11e5-84b6-93a0f50c97dd.png)

tdotcspot:
+1

josh256:
+1

jelbourn:
The tentative idea is that, for small screen sizes, a more dialog-y picker will pop-up instead of what's currently there.

arlowhite:
The popup looks fine with my app on my Galaxy Nexus and does not run off the right side. (No horizontal scrolling) However, I'm using Ionic. So ion-view > ion-content might be doing something to the viewport.
Also, index.html has this:
```html
    <meta name="viewport" content="initial-scale=1, maximum-scale=1, user-scalable=no, width=device-width">
```

JoseGCheca:
Im having the same issue when displaying md-datepicker inside a md-dialog. 

![screenshot_2015-09-21-17-27-36](https://cloud.githubusercontent.com/assets/6599858/9996297/adfd79bc-6086-11e5-8a6c-b3017717a462.png)

jelbourn:
The datepicker's calendar pop-up should now remain fully visible on the screen even on small devices with 64fb8037f8cdccbbc90b5f6ffa0f65d4cab94d4f. [Check it out here](https://material.angularjs.org/HEAD/#/demo/material.components.datepicker). 

Chime in if the result is still _not usable_ on a mobile device.

JoseGCheca:
Thank you very much. Works fine now.

jelbourn:
Closing this issue since nobody has chimed in that they're still having problems.

PR #4782, opened 2015-09-22
fix(datepicker): prevent calendar clipping on small screens. For #4558

jelbourn:
@robertmesserle for review


robertmesserle:
LGTM

PR #4784, opened 2015-09-22
fix(dateLocale): guard for setting midnight to null. Fixes #4780

### LINKED ISSUE #4780: md-date-locale-provider: Error returning falsy value from parseDate
Returning eg. null from parseDate function on md-date-locale-provider causes a crash when manually setting a date value of an md-datepicker:
````
TypeError: Cannot read property 'setHours' of null
    at Object.g [as setDateTimeToMidnight] (angular-material.min.js:9)
    at o.handleInputEvent (angular-material.min.js:9)
    at angular-material.min.js:7
    at angular.min.js:145
    at e (angular.min.js:42)
    at angular.min.js:45
````

Expected behaviour:
Setting null marks the md-datepicker input as invalid.

Workaround:
Return new Date(NaN)

jelbourn:
@mhchen for review

### DIFF
--- a/src/components/datepicker/dateUtil.js
+++ b/src/components/datepicker/dateUtil.js
@@ -195,7 +195,9 @@
      * @param {Date} date
      */
     function setDateTimeToMidnight(date) {
-      date.setHours(0, 0, 0, 0);
+      if (isValidDate(date)) {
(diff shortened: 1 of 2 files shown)


mhchen:
LGTM, sorry this was an issue

PR #4786, opened 2015-09-22
fix(mdInput): Fix md-maxlength when used with ng-messages.

### LINKED ISSUE #4783: Latest commit broke md-maxlength
Check demo site with latest commit. Chrome displays correctly; however Firefox and IE display text length above input elements now.

topherfangio:
The recently added support for input elements to have multiple
messages caused a rendering issue for the `md-maxlength`
character counter.

Update code/CSS to properly position the character counter when
used with and without ng-messages.

Fixes #4783.

POSSIBLE BREAKING CHANGE - The `<div class="md-char-counter">`
that is automatically added to the input is now added after
the input if no messages are found, but inside if they are.

This may cause some styling issues if users provide custom
CSS. Users may need to update their CSS to take the extra
HTML into account.

Example rendered HTML:

**Without `ng-messages`:**

```html
<label>Label is here</label>
<input />
<div class="md-char-counter">10/20</div>
```

**With `ng-messages`:**

```html
<label>Label is here</label>
<input />
<div ng-messages="...">
  <div class="md-char-counter">10/20</div>
  <div ng-message="requried">This is required</div>
</div>
```

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -362,7 +362,25 @@ function mdMaxlengthDirective($animate) {
     // Stop model from trimming. This makes it so whitespace
     // over the maxlength still counts as invalid.
     attr.$set('ngTrim', 'false');
-    input.after(charCountEl);
+
+    var ngMessagesSelectors = [
+      'ng-messages',
+      'data-ng-messages',
+      'x-ng-messages',
+      '[ng-messages]',
+      '[data-ng-messages]',
+      '[x-ng-messages]'
+    ];
+
+    var ngMessages = containerCtrl.element[0].querySelector(ngMessagesSelectors.join(','));
+
+    // If we have an ngMessages container, put the counter at the top; otherwise, put it after the
+    // input so it will be positioned properly in the SCSS
+    if (ngMessages) {
+      angular.element(ngMessages).prepend(charCountEl);
+    } else {
+      input.after(charCountEl);
+    }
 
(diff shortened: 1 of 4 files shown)


topherfangio:
@ThomasBurleson @jelbourn Can you guys review this real fast and let me know if you see any problems? Would be nice to have in 11.1 if possible.

src/components/input/input.js
```diff
@@ -362,7 +362,25 @@ function mdMaxlengthDirective($animate) {
...
-    input.after(charCountEl);
+
+    var ngMessagesSelectors = [
+      'ng-messages',
+      'data-ng-messages',
+      'x-ng-messages',
+      '[ng-messages]',
+      '[data-ng-messages]',
+      '[x-ng-messages]'
+    ];
```
jelbourn: Would it be easier to make a directive that registers itself with this than duplicating all these syntaxes from Angular?

ThomasBurleson: you mean a `.value( )` service to manage this array of constants ?

topherfangio: @jelbourn You mean an `ng-messages` directive that requires the `md-input-container` and registers itself with the container; then this directive just asks the container if there are any messages.

That definitely sounds a bit cleaner; I'll take a look. I think we already have one I can reuse...

jelbourn: Yep, that's what I meant.

PR #4800, opened 2015-09-23
fix(mdTooltip): Tooltip inside subheader causes Firefox hang.

### LINKED ISSUE #4777: Using an mdTooltip inside an mdSubheader causes Firefox to hang
I'm trying to use an mdTooltip to show extra information when a user hovers on the text contained in an mdSubheader, and it works fine on both Chrome and IE but, on Firefox for some reason it stalls until the browser prompts me giving the option to interrupt the script. I've tried to let the script run to completion several times, but it never ends.

The following error is shown on Firefox console when it stalls:
```
"Error: [$rootScope:inprog] $digest already in progress
http://errors.angularjs.org/1.4.5/$rootScope/inprog?p0=%24digest
minErr/<@http://localhost:8080/sgo/js/lib/angular.js:68:12
beginPhase@http://localhost:8080/sgo/js/lib/angular.js:16235:1
$RootScopeProvider/this.$get</Scope.prototype.$apply@http://localhost:8080/sgo/js/lib/angular.js:15976:11
timeout/timeoutId<@http://localhost:8080/sgo/js/lib/angular.js:17753:25
completeOutstandingRequest@http://localhost:8080/sgo/js/lib/angular.js:5490:7
Browser/self.defer/timeoutId<@http://localhost:8080/sgo/js/lib/angular.js:5762:7
```

The following code reproduces the error on my environment (Firefox version 41.0) consistently:
```html
<md-subheader class="md-warn" layout-align="center center">
   Dummy text
   <md-tooltip>* Extra info.</md-tooltip>
</md-subheader>
```

topherfangio:
When a `<md-tooltip>` was placed inside the `<md-subheader>`, an infinite loop was being hit (only in Firefox) which was causing the browser to hang.

The infinite loop was caused by the tooltip's `getParentWithPointerEvents()` method which travels up the
DOM looking for a parent who has pointer events. In Firefox, the `element.parent()` can apparently return an empty jqLite object that is not null, but whose length is 0. In this case, calling `parent.parent()` again will return the same object.

Add a check to the tooltip's loop to ensure that `parent.length` is never 0.

Fixes #4777.

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -106,9 +106,12 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
      */
     function getParentWithPointerEvents () {
       var parent = element.parent();
-      while (parent && hasComputedStyleValue('pointer-events','none', parent[0])) {
+
+      // jqLite might return a non-null, but still empty, parent; so check for parent and length
+      while (parent && parent.length && hasComputedStyleValue('pointer-events','none', parent[0])) {
         parent = parent.parent();
       }
+
       return parent;
     }
 


ThomasBurleson:
Where is the unit test for this fix?

src/components/tooltip/tooltip.js
```diff
@@ -106,9 +106,12 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
      */
     function getParentWithPointerEvents () {
       var parent = element.parent();
-      while (parent && hasComputedStyleValue('pointer-events','none', parent[0])) {
+
+      // jqLite might return a non-null, but still empty, parent; so check for parent and length
+      while (parent && parent.length && hasComputedStyleValue('pointer-events','none', parent[0])) {
```
ThomasBurleson: @topherfangio - I think we should move the **hasComputedStyleValue** to `$mdUtil.hasComputedStyleValue( )`; which would simplify testing of this feature.

PR #4820, opened 2015-09-24
feat(interimElement): add onShowing event

Patrik-Lundqvist:
Currently hideElement has an onRemoving event, this is the showElement equivalent. Useful when dealing with loading spinners that should disappear before the dialog shows.

### DIFF
--- a/src/core/services/interimElement/interimElement.spec.js
+++ b/src/core/services/interimElement/interimElement.spec.js
@@ -362,6 +362,28 @@ describe('$$interimElement service', function() {
         expect(autoClosed).toBe(true);
       }));
 
+      it('calls onShowing before onShow', inject(function() {
+        var onShowingCalled = false;
+
+        Service.show({
+          template: '<some-element />',
+          passingOptions: true,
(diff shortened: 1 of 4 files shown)


src/core/services/interimElement/interimElement.spec.js
```diff
@@ -362,6 +362,21 @@ describe('$$interimElement service', function() {
...
+          passingOptions: true,
+          onShowing: onShowing
+        });
+
+        expect(onShowingStarted).toBe(true);
+
+        function onShowing(scope, el) {
+          onShowingStarted = true;
+        }
+      }));
```
jelbourn: This test asserts that onShowing is called, but not that it's called in the right order.

Patrik-Lundqvist:
- interimElement test now checks that onShowing is called before onShow
- added dialog test that makes sure onShowing is called before the dialog shows

jelbourn:
Great tests, thanks!

@ThomasBurleson This looks good to me for merging in 0.11.3

PR #4828, opened 2015-09-24
fix(datepicker): stop calendar going off-screen if body is scrollable.  Fixes #4781

### LINKED ISSUE #4781: datepicker repositions based on body size, not visible viewport (w/ scrolling)
screen shot 2015-09-22 at 1 15 30 pm
Whenever I use the md-datepicker and if it is near the page margins it doesn't open properly. By not opening properly I mean it is not scrollable and it should by itself know where it needs to show the popup based on its space requirement. Just like the image showing it should have open on top and not keep opening in space where the user has no option to select the date

jelbourn:
@robertmesserle for review

### DIFF
--- a/src/components/datepicker/datePicker.spec.js
+++ b/src/components/datepicker/datePicker.spec.js
@@ -8,13 +8,14 @@ describe('md-date-picker', function() {
   var initialDate = new Date(2015, FEB, 15);
 
   var ngElement, element, scope, pageScope, controller;
-  var $timeout, $$rAF, $animate, keyCodes, dateUtil, dateLocale;
+  var $timeout, $$rAF, $animate, $window, keyCodes, dateUtil, dateLocale;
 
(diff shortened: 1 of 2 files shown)


src/components/datepicker/datePicker.spec.js
```diff
@@ -194,9 +195,35 @@ describe('md-date-picker', function() {
...
+
+      // Absolutely position the picker near (say ~30px) the edge of the viewport.
+      element.style.position = 'absolute';
+      element.style.top = (window.innerHeight - 30) + 'px';
+      element.style.left = '0';
+      document.body.appendChild(element);
+
+      // Open the pane.
+      element.querySelector('md-button').click();
+      $timeout.flush();
```
robertmesserle: Would the material flush work for this?

robertmesserle:
LGTM

PR #4847, opened 2015-09-25
fix(fabSpeedDial/Toolbar): Make hovering an option via CSS.

### LINKED ISSUE #4259: FAB speed dial mouse hit area includes invisible elements.
For the FAB speed dial, when in hover mode, it also responds to mouse movements over the area defined for the invisible child elements to appear in, meaning that if you hover the mouse over inactive spot where one of the children are set to appear, then the fab activates and shows its children.

What is the expected behavior.

When you move a mouse over the trigger area, it should activate, not when you hover the mouse over the area that is designated for its popup buttons.

This is very important if you want to have fab spinners act as a style of toolbox for a card or something, so that when its closed up all the buttons are hidden and where they were you can still click on the underlying elements, and when its active the buttons appear over other content, right now anything in the area where buttons would appear is inaccessible.

This can be demonstrated on the second demo on the fab spinner demo page, hover the mouse in the white area under the fab.

topherfangio:
Initially, the speed dial was designed to always open when the user hovered over any portion of the speed dial (including the area where the actions would eventually appear). However, this made the speed dial unusable on mobile (and sometimes unusable on desktop) because it disallowed the user from pressing anything underneath the actions.

* Add the `md-hover-full` CSS class to allow developer configuration of this behavior.
* Ensure fabToolbar also works this way (it did already), and fix jumpy animation.
* Also updated the docs/demos and made the demos easier to use on mobile.

BREAKING CHANGE - The fabSpeedDial no longer automatically opens when hovering over the invisible actions. Add the `md-hover-full` class to enable this interaction.

Fixes #4259.

### DIFF
--- a/src/components/fabSpeedDial/demoMoreOptions/style.scss
+++ b/src/components/fabSpeedDial/demoMoreOptions/style.scss
@@ -1,10 +1,13 @@
-.lock-size {
-  min-width: 300px;
-  min-height: 300px;
-  width: 300px;
-  height: 300px;
-  margin-left: auto;
-  margin-right: auto;
+// Line the fab up properly on different screen sizes/layouts
+.md-fab-top-right {
+  top: 16px;
+}
+
+@media only screen and (max-device-width: 600px) {
+  .md-fab-top-right {
+    top: 9px;
+    right: 9px;
+  }
 }
 
 .md-fab.demo-fab.trigger-fab, .md-fab.demo-fab.action-fab {
@@ -21,5 +24,26 @@
 md-content div {
   &[flex="50"] {
     padding: 15px;
+  }
+}
+
+
+md-fab-speed-dial {
+  // Add a simple scale transition to the trigger when hiding/showing the speed dial
+  md-fab-trigger {
+    transition: all 0.3s ease-in-out;
+    transform: scale(1);
+  }
+
+  // Note: you MUST use an existing CSS class for the animation to fire properly
+  &.md-scale, &.md-fling {
+    &.ng-hide {
+      // Use !important to override ng-hide's `display: none !important`
+      display: flex !important;
+
+      md-fab-trigger {
+        transform: scale(0);
+      }
+    }
   }
 }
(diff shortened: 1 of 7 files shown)


src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -182,6 +187,12 @@
       var ctrl = element.controller('mdFabSpeedDial');
       var items = el.querySelectorAll('.md-fab-action-item');
 
+      // Grab our element which stores CSS variables
```
jelbourn: "CSS variables" -> "CSS modifiers" ?

src/components/fabSpeedDial/demoMoreOptions/style.scss
```diff
@@ -23,3 +26,23 @@ md-content div {
...
+  // Add a simple scale transition to the trigger when hiding/showing the speed dial
+  md-fab-trigger {
+    transition: all 0.3s ease-in-out;
+    transform: scale(1);
+  }
+
+  // Note: you MUST use an existing CSS class for the animation to fire properly
+  &.md-scale, &.md-fling {
+    &.ng-hide {
+      display: flex !important;
```
jelbourn: Add comment explaining use of `!important` here

jelbourn:
LGTM aside from comments

src/components/fabSpeedDial/fabSpeedDial.js
topherfangio: Funny thing, I actually meant variables there. The SCSS defines the z-index as a variable and I wanted to pass that along to the JS code in a clean way. This ensured that I had a known element to look for those variables.

I'm happy to rename/refactor it though if you think something else makes more sense :-)

topherfangio:
@jelbourn Responded/fixed comments. Can you please re-review at your convenience?

src/components/fabSpeedDial/fabSpeedDial.js
jelbourn: Ah, I see. I misread it the first time around. Carry along.

Issue #4779, opened 2015-09-22
mdContactChips: add option to specify default image


Frank3K:
In the current implementation of `md-contact-chips` one can specify the field name of the image corresponding to a contact using the `md-contact-image` attribute.

However, when a given contact does not have set that field (i.e. the contact does not have an image), an empty `<img />` tag [is rendered](https://github.com/angular/material/blob/00474c39fb2c305ab43874809db3032a65cf40d6/src/components/chips/js/contactChipsDirective.js#L70-L72). Since an `alt`-attribute is added to the image some browsers (at least Chromium on Linux) render the image as wide as the contents of the `alt`-attribute. This messes with the layout of the contact chips.

It would be nice if there was a way to specify a default image, which is used in case there is no corresponding image. Another (slightly inferior) option would be to not render the `<img />` at all.

topherfangio:
I think this is a good idea, but as a workaround, you can handle this at the "model" layer by making sure the data returns something useful when null.

I typically handle this at the JS model or even server layer (usually Ruby on Rails) so that I can use existing model data to dynamically change the URL based on that particular user (i.e. different pictures for male/female, etc).

Frank3K:
I also thought of solving the issue in the model layer, but I thought a solution in the contact chips itself would be easier for the developer.

Do you think it is a good idea to hide the image if not image has been provided at all? I think I can work out a PR for that.

topherfangio:
If you can find a solution that replaces the image with something of the same size (and maybe even an icon). I think that might work well.

PR #4851, opened 2015-09-26
fix(chips): do not display broken image when no image is provided

Frank3K:
When no image property is defined on the user model, do not render an `img`-element.

References #4779.

### DIFF
--- a/src/components/chips/contact-chips.spec.js
+++ b/src/components/chips/contact-chips.spec.js
@@ -60,6 +60,28 @@ describe('<md-contact-chips>', function() {
       expect(ctrl.highlightFlags).toEqual('i');
     });
 
+    it('renders an image element for contacts with an image property', function() {
+        scope.contacts.push(scope.allContacts[2]);
+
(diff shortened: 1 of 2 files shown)


topherfangio:
Any chance you can provide some pictures on how this looks with some items having a valid image and some not? I'd like to make sure that it looks consistent both ways.

Frank3K:
Current display state when there is no image for the contacts [codepen](https://codepen.io/anon/pen/gaLLYb). Chrome 45, Firefox 40 and Internet Explorer 11 show the complete `alt` text here, which messes with the layout.

Demo with new code follows asap.

topherfangio:
Okay; I think that looks pretty good. Can you add a quick set of unit tests for this change? Basically, just check that if the image is provided, the element has the `` tag, and if not, it doesn't?

Frank3K:
@topherfangio: done.

topherfangio:
@ThomasBurleson LGTM :-)

PR #4884, opened 2015-09-28
fix(dialog) use keydown instead of keyup

### LINKED ISSUE #4041: md-dialog is receiving duplicated ESCAPE key events
When you press ESCAPE in a dialog, the dialog will get a duplicated key event (I checked by checking event.timestamp). preventDefault/stopPropogation don't work with the dialog because the dialog is getting a different one.

Here's the codepen.

http://codepen.io/anon/pen/pJGdap

At first I thought this was something to do with md-autocomplete, but I added md-select and the issue appears there too. To test, press escape with the autocomplete or select open.

I was searching for the reason why and I couldn't exactly figure it out. It might be related to async events.

Edit: I think I figured it out. md-dialog is watching for keyup while md-autocomplete is preventing keydown

clshortfuse:
The rest of angular material watches for keydown events, not keyup. 

fixes #4041

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -630,16 +630,16 @@ function MdDialogProvider($$interimElementProvider) {
           }
         };
 
-        // Add keyup listeners
-        element.on('keyup', keyHandlerFn);
-        target.on('keyup', keyHandlerFn);
+        // Add keydown listeners
+        element.on('keydown', keyHandlerFn);
+        target.on('keydown', keyHandlerFn);
         window.on('resize', onWindowResize);
 
         // Queue remove listeners function
         removeListeners.push(function() {
 
-          element.off('keyup', keyHandlerFn);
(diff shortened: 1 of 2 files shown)


jelbourn:
@clshortfuse Can you squash your two commits into one?

Code LGTM; someone on the team just needs to run this and make sure everything is still good.

clshortfuse:
@jelbourn I'm honestly not sure how to do that.

jelbourn:
Don't worry about it, then.

ThomasBurleson:
@EladBezalel - should this work be merged into #4972 ?

EladBezalel:
@ThomasBurleson this code fixes #4041, when autocomplete or select is open inside a dialog and you use ESC key to close it, it closes the dialog also. 

On #4972 I switched the click event to mousedown mouseup because the click event closed the dialog if the click happened inside and dragged outside, so keeping the original element with mousedown solved it.

PR #4888, opened 2015-09-29
Added an internal function to handle `debounce` in demo.

dlochrie:
Please see this [Google Closure compiler PR](https://github.com/google/closure-compiler/pull/1170) for more information/reference.

Basically, the demo code is using `$mdUtil`, which is not part of the public API, and therefore not suitable for the public (docs) reference.


jelbourn:
LGTM, thanks for fixing this!

dlochrie:
Sorry, just now seeing this! Should have merged it - thanks @ThomasBurleson for pulling it in.

PR #4891, opened 2015-09-29
fix(card): position md-card-footer at bottom of card

### LINKED ISSUE #3144: md-card-footer doesn't put content flush against the bottom edge of the card
As the title says, the directive docs say that md-card-footer puts the content against the bottom of the card, but doesn't actually do it. [Plunker.](http://plnkr.co/edit/keO0rG79Y53OcYL6e3fB?p=preview)

I am trying to create a news section that works similarly to google play newsstand. The cards have to stay the same height, and the footer containing the author information should always be at the bottom.
This image shows how google play newsstand does it. Unfortunately the footer isn't actually at the bottom, so I can't achieve this.

The only styling applied to the cards in my Plunker is a width of 375px if the screen size is larger than small, to fit for desktop screens or landscape phones.

**Google Play Newstand**
screenshot_2015-06-05-17-40-26

**My Plunker**
image

mhchen:
Old behavior did not cause the footer to stick to the bottom of the card
if the card had an assigned height taller than its content.

Fixes #3144

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -28,6 +28,7 @@ md-card {
     }
   }
   md-card-footer {
+    margin-top: auto;
     padding: $card-padding;
   }
 }


mhchen:
Thanks @rschmukler for the excellent and elegant fix

mhchen:
@ThomasBurleson Please review

jelbourn:
LGTM

PR #4915, opened 2015-09-29
fix(inputs): Fix floating label and char counter positions.

### LINKED ISSUE #4872: md-char-counter breaks visual alignment
After upgrading to 0.11.1, applying the css-class 'md-char-counter' breaks visual alignment.
skaermbillede 2015-09-28 kl 11 34 02

[Plunkr](http://plnkr.co/edit/bEaRUfa1dRNpSZL2AX37?p=preview)

topherfangio:
Recent updates to the `md-char-counter` class caused alignment issues when the input was side-by-side with another input. The underlying cause was that the other input uses `maring-top: auto` instead of `margin-top: 0` which made the inputs "bottom aligned" as the height of it's neighbors changed.

This fixes the CSS on the char counter, inputs and select to better align and updates the inputs demo to show the select next to an input with the char counter.

Also, I noticed a bug with the floating labels when using only the placeholder, so this adds some additional tests and fixes the bug.

Fixes #4872.

### DIFF
--- a/src/components/input/demoBasicUsage/index.html
+++ b/src/components/input/demoBasicUsage/index.html
@@ -16,9 +16,9 @@
 
       <div layout layout-sm="column">
         <md-input-container style="width:70%">
-            <label>Company (Disabled)</label>
-            <input ng-model="user.company" disabled>
-          </md-input-container>
+          <label>Company (Disabled)</label>
+          <input ng-model="user.company" disabled>
+        </md-input-container>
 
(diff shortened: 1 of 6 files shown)


topherfangio:
@rschmukler Can you do a quick review? I updated the select CSS to use `margin: 0` instead of `margin: auto 0 0 0` so that it now matches my updates to the other inputs. I can't see any issues with this, but wanted to run it by you.

topherfangio:
Spoke with Ryan in Thursday's meeting and he said this looked good.

@ThomasBurleson Should be good to merge when ready.

PR #4922, opened 2015-09-29
docs: fix demos GitCDN link to use proper URL for versioning

topherfangio:
The GitCDN link used in the current demos points to a URL which is redirected to a new one. When this redirect occurs, the version info supplied in the URL is lost and the file always points to HEAD making it difficult to track down.

Fix snapshot script to use the new link which retains the version information.

_**NOTE:** I can't really test this without actually pushing a snapshot, but I copy/pasted the URL to make sure it worked._

### DIFF
--- a/scripts/snapshot-docs-site.sh
+++ b/scripts/snapshot-docs-site.sh
@@ -19,7 +19,7 @@ function run {
   rm -rf code.material.angularjs.org/HEAD
 
   echo "-- Copying docs site to snapshot..."
-  sed -i "s,http://localhost:8080/angular-material,https://gitcdn.xyz/repo/angular/bower-material/v$VERSION/angular-material,g" dist/docs/docs.js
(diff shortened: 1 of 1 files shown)


topherfangio:
@ThomasBurleson @robertmesserle Can one of you review this real fast to make sure it's sound?

robertmesserle:
@topherfangio This is just for versioned sites and will not address the SHA issue with the HEAD, correct?

robertmesserle:
If so, LGTM!

topherfangio:
@robertmesserle Indeed; that's why I was asking about ngMaterial bot. Wherever/however that runs, I think we need to change something there too to use the specific SHA.

robertmesserle:
Yea, I believe that was setup and working at some point in the past, but
I'm not entirely sure what state it's in now.

topherfangio:
@robertmesserle Is it worth re-running the script to update the site so the 0.11.1 codepens are correct? Or should we just tell people that it will come in 0.11.2 (or 0.12.0 or whatever)?

robertmesserle:
It might be pretty trivial to fix manually, let me look.

PR #4947, opened 2015-09-30
fix(docs): fix $mdDateLocaleProvider sample code

brunoqc:
Please disregard if it was intentionally like this for any reason.


jelbourn:
I did do that on purpose since I wanted the line to just get the point across without being too long. Thanks for the contribution, though!

PR #4953, opened 2015-09-30
fix(input): fix bad char counter when value is a number. Fixes #4635.

### LINKED ISSUE #4635: Input with maxlength and value as a number, show undefined/maxlength
When the value of a text input is given as number the character tracker on the right show undefined/<maxlength>

See attached:
selection_004

jelbourn:
Also did some cleanup on the unit tests.

@robertmesserle for review

### DIFF
--- a/src/components/input/input.spec.js
+++ b/src/components/input/input.spec.js
@@ -1,47 +1,64 @@
 describe('md-input-container directive', function() {
+  var $compile, pageScope;
 
-  beforeEach(module('material.components.input'));
-  beforeEach(module('ngAria'));
+  beforeEach(module('ngAria', 'material.components.input'));
+
+  beforeEach(inject(function($injector) {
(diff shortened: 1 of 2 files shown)


src/components/input/input.spec.js
```diff
@@ -1,47 +1,65 @@
...
 
-  beforeEach(module('material.components.input'));
-  beforeEach(module('ngAria'));
+  beforeEach(module('ngAria', 'material.components.input'));
+
+  beforeEach(inject(function($injector) {
+    $compile = $injector.get('$compile');
+
+    var $rootScope = $injector.get('$rootScope');
+    pageScope = $rootScope.$new();
```
robertmesserle: Should these be merged into a single line?
`pageScope = $injector.get('$rootScope').$new()`

Issue #3287, opened 2015-06-16
md-autocomplete: unable to click/focus on anything after a click on the "clear" button


jgoux:
Steps to reproduce the bug : 
- Go on https://material.angularjs.org/HEAD/#/demo/material.components.autocomplete
- On the first autocomplete, select something
- Click the "clear" button
- Now I can't click/focus on anything on the page

I'm on the master and my browser is Chrome v43

DrDanRyan:
This issue also occurs if you select an item, and then backspace to delete the input to an empty state and then click away

kschwidder:
+1

siderova:
+1

tolemac:
+1

ssiderova:
+1

GautierThiriez:
+1 freezing the chrome tab.

SimeonC:
To add to this - this seems to occur on the third consecutive click, if you use arrow keys + enter to select this doesn't occur.

If you click the autocomplete, click a selection, then click clear it locks up. I've recorded a video demonstrating here - http://screencast.com/t/NYkj14IYG9F

When we say "locks up" what is actually happening is that the `<div class="md-scroll-mask">` is not being removed at some point and when you click the element it doesn't remove either.

charliewynn:
in 0.10.0 I commented out line 12469 if angular-material
    `defineProperty('hidden', handleHiddenChange, true);`

and the function at 12642  `handleHiddenChange` (probably redundant to do the 2nd thing)

And it seems to be working for me.. Not sure what horrible thing I'm missing out on without the element 'hiding', but at least my app doesn't lock up :smile:

vlad-lifliand:
+1

gitawego:
same here.

sricc:
+1

mwymann:
+1

Patzick:
+1

jozsi:
+1

GuruM:
+1

piyushchauhan2011:
:+1: serious issue mate, It really affected our website. Need a quick fix as soon as possible!

robertmesserle:
This is resolved in master and will be fixed in 0.10.1

GuruM:
Thanks @robertmesserle!

tolemac:
Work fine here, thats ok!

bdteo:
+1

prerna12:
[...]

Issue #4905, opened 2015-09-29
Layout manager broken in 11.1 release


britztopher:
For some reason the layout flexbox grid system seems to break with the upgrade to 0.11 from 0.10. It appears that the changes to the layout css properties are the culprit and there is no documentation over these changes or something broke.

This plunker here http://plnkr.co/edit/cXgLWewAPzr7B2hxH9gF?p=preview shows the broken flexbox properties and if you change the css sheet that is added to 0.10.1 from 0.11.1 then you will see that the flexbox grid will work properly.

For Example: 
Change line number 5 from: 
 `<link rel="stylesheet" href="https://ajax.googleapis.com/ajax/libs/angular_material/0.11.1/angular-material.min.css" />
`
To: 
` <link rel="stylesheet" href="https://ajax.googleapis.com/ajax/libs/angular_material/0.10.1/angular-material.min.css" />
`

ThomasBurleson:
Please test with the lastest 0.11.2 released just today: See Layout fixes in the [Change Log](https://github.com/angular/material/blob/master/CHANGELOG.md)

ThomasBurleson:
Here is an updated Plunkr that works: http://plnkr.co/edit/vr1ZVnkDkbsL19E0Wo3M?p=preview

!! Important: You **MUST** configure a module dependency for ngMaterial:
```js
angular.module('MyApp', ['ngMaterial']);
```

britztopher:
Looks good. Thanks

Issue #4678, opened 2015-09-16
Suggestion list template appears before md-min-length is fulfilled for md-autocomplete


lscordilis:
I'm sure I'm not the only one who's noticed this since the version bump. The list template appears blank before the min length is reached and then starts getting populated with suggestions.

It's a minor annoyance, as they cover the ng-messages below the input. In general, maybe there should be an option to have the suggestions list extend above the input as opposed to always below, as validation for ng-messages may be more sophisticated than just ```required```

topherfangio:
Possibly related to #4665.

topherfangio:
This should be fixed in master. Can you please let us know if you are still experiencing the issue, and if so, provide a Codepen demonstrating the problem? If so, I will reopen.

Also, regarding the position of the suggestions list, can you please open a separate issue so that we can mark it as a feature request?

Issue #4673, opened 2015-09-16
mdChips: with autocomplete minLength=0 and noCache=true, menu doesn't show if all choices are selected and removed


antoinebrault:
Using angular 1.4.5 and material 0.11, when all choices are selected, and deselected, menu with choices doesn't show until a caracter is entered and removed.

https://jsfiddle.net/fcngt79d/1/

1. Select both items
2. Remove one of the items
3. ** try to click in the autocomplete field. Menu doesn't appear.
4. Write something in the autocomplete field and remove it.
5. Menu appears.

lscordilis:
Yeah, md-autocomplete directives such as md-min-length and md-no-cache aren't working like they used to. +1

topherfangio:
I just updated your fiddle to use master, and this unfortunately still looks to be an issue. We will investigate.

enderdickerson:
Why was this marked as Depricated? It's still present in 1.1.0-rc4. Is there a known workaround for this then?

topherfangio:
@enderdickerson Please see the following forum post for more information: https://groups.google.com/d/msg/ngmaterial/3Ip9nd68bfk/X448kGtvAQAJ

enderdickerson:
Thanks. I guess I'll open a new issue and cite this one following the spring cleaning effort. I'm assuming this fits the depricated category for one of these two reasons:
- Issue is real, but only affects a very small use case
- Issue is real, but would require significant work to fix, and is less urgent than other needs.

Unfortunately it's an issue I could really use a fix for. Thanks for the quick reply @topherfangio.

PR #4958, opened 2015-10-01
fix(autocomplete): Fix many issues with showing/hiding.

### LINKED ISSUE #4025: [master#5722449] md-autocomplete - presseing (x) focuses the field but the list drop-down list doesn't appear
After selecting a value from the autocomplete list, press (X). 

The autocomplete field is focused but the list doesn't appear.

This was found on master at commit 5722449b372604fc2231f73140a44094c3a94cf8

### LINKED ISSUE #4469: md-autocomplete does not receive focus events after clear
Re-producable in this forked Codepen: http://codepen.io/hannupekka/pen/LpEegj

Steps:
1. Click md-autocomplete field
2. Select any suggestion
3. Click reset from the end of the field
4. Click md-autocomplete field

Expected outcome:
I get list of suggestions like in the beginning.

Actual outcome:
Suggestions not showing up. I have to click outside md-autocomplete field and then click back inside the field to get suggestions.

### LINKED ISSUE #4618: md-autocomplete open empty drop-down when default searchText exist.
When default value of `selectedItem` exist (not null), every deletion a character of the default value keeps the empty drop-down open.

CODEPEN : http://codepen.io/anon/pen/YywqBv

image

### LINKED ISSUE #4665: autocomplete: virtual repeat is shown even with no matches; sizing incorrect with md-not-found
When using the autocomplete, if the user does not supply a `<md-not-found>` template, the virtual repeat popup is still shown and is blank.

If you add the `<md-not-found>` template, it appears but the sizing is incorrect.

http://codepen.io/topherfangio/pen/rOeBdV?editors=100

### LINKED ISSUE #4788: md-autocomplete: md-not-found flashes up briefly when md-items bound to a promise
Repro:
http://codepen.io/anon/pen/PPGoba

Press a key and you should see the md-not-found template show briefly and disappear before the promise is resolved.

Update: It seems something still flashes up even when there's no md-not-found template specified: http://codepen.io/anon/pen/bVwGqO

### LINKED ISSUE #4855: md-authcomplete make the not-found item clickable
There is a way to make the md-not-found item clickable?
I tried to put an ng-click on the md-not-found, or to wrap the content with div with ng-click (which worked on 0.10.0) but I doesn't work on 0.11.1
```html
<md-autocomplete
	md-search-text="searchText"
	md-items="channel in search(searchText)"
	md-delay="100"
	md-min-length="1"
	md-selected-item-change="goToChannel(channel)"
	placeholder="Search for an artist">

	<md-item-template>
		<span md-highlight-text="searchText" md-highlight-flags="^i">{{ channel.name }}</span>
		<span style="float: right">{{ channel.subscriptions }} Subscribers</span>
	</md-item-template>

	<md-not-found>
		<div ng-click="goToCreateChannel(searchText)">
			No matches found for "{{ searchText }}".
			Find {{ searchText }} on the web!
		</div>
	</md-not-found>
</md-autocomplete>
```

### LINKED ISSUE #4906: md-autocomplete show loading indicator on initial load/focus
1. [Basic usage] (https://material.angularjs.org/0.11.1/#/demo/material.components.autocomplete)
2. Check "Simulate query for results?"
3. Focus auto-complete field

Expected that initial data fetching(loading either by ajax call or async promise) will show progress indicator until the query is done.

Possible scenarios/dependancies to check
1. [Basic usage with query enabled] (https://material.angularjs.org/0.11.1/#/demo/material.components.autocomplete)
2. auto-complete with [md-floating-label] attr
3. Maybe but not necessary to test... should also work within md-chips with 1,2 cases

topherfangio:
The logic behind showing and hiding the autocomplete's list of suggestions was difficult to understand and was causing some issues with displaying at the proper times.

Refactor code to use a "show when criteria is met" approach instead of a "hide when criteria is met" approach and fix some issues with the `md-not-found` and suggestions list appearing when they should not.

Fixes #4665. Fixes #4788. Fixes #4906. Fixes #4855. Fixes #4618. Fixes #4469. Fixes #4025.

**References:** #4309, #4678, #4673, #4518, #4503, #4358, #4905.

### DIFF
--- a/src/components/autocomplete/demoBasicUsage/script.js
+++ b/src/components/autocomplete/demoBasicUsage/script.js
@@ -16,6 +16,12 @@
     self.selectedItemChange = selectedItemChange;
     self.searchTextChange   = searchTextChange;
 
+    self.newState = newState;
+
+    function newState(state) {
+      alert("Sorry! You'll need to create a Constituion for " + state + " first!");
+    }
+
     // ******************************
     // Internal methods
     // ******************************
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -34,6 +34,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   ctrl.id         = $mdUtil.nextUid();
   ctrl.isDisabled = null;
   ctrl.isRequired = null;
+  ctrl.hasNotFound = false;
 
   //-- public methods
   ctrl.keydown                       = keydown;
@@ -48,6 +49,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   ctrl.registerSelectedItemWatcher   = registerSelectedItemWatcher;
   ctrl.unregisterSelectedItemWatcher = unregisterSelectedItemWatcher;
   ctrl.notFoundVisible               = notFoundVisible;
+  ctrl.loadingIsVisible              = loadingIsVisible;
 
   return init();
 
@@ -209,19 +211,16 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     if (!hidden && oldHidden) {
       positionDropdown();
 
-      if (elements)
+      if (elements) {
         $mdUtil.nextTick(function () {
-
           $mdUtil.disableScrollAround(elements.ul);
-
-        }, false, $scope);
-      } else if (hidden && !oldHidden) {
-        $mdUtil.nextTick(function () {
-
-          $mdUtil.enableScrolling();
-
         }, false, $scope);
       }
+    } else if (hidden && !oldHidden) {
+      $mdUtil.nextTick(function () {
+        $mdUtil.enableScrolling();
+      }, false, $scope);
+    }
   }
 
   /**
@@ -236,7 +235,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    */
   function onListLeave () {
     noBlur = false;
-    if (!hasFocus) ctrl.hidden = true;
+    ctrl.hidden = shouldHide();
   }
 
   /**
@@ -327,9 +326,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
 
         // cancel results if search text is not long enough
         if (!isMinLengthMet()) {
-          ctrl.loading = false;
           ctrl.matches = [];
-          ctrl.hidden  = shouldHide();
+          setLoading(false);
           updateMessages();
         } else {
           handleQuery();
@@ -343,8 +341,18 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    * Handles input blur event, determines if the dropdown should hide.
    */
   function blur () {
-    hasFocus = false;
-    if (!noBlur) ctrl.hidden = true;
+    if (!noBlur) {
+      hasFocus = false;
+      ctrl.hidden = shouldHide();
+    }
+  }
+
+  function doBlur(forceBlur) {
+    if (forceBlur) {
+      noBlur = false;
+    }
+
+    elements.input.blur();
   }
 
   /**
@@ -354,7 +362,6 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     hasFocus = true;
     //-- if searchText is null, let's force it to be a string
     if (!angular.isString($scope.searchText)) $scope.searchText = '';
-    if ($scope.minLength > 0) return;
     ctrl.hidden = shouldHide();
     if (!ctrl.hidden) handleQuery();
   }
@@ -392,9 +399,10 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         event.stopPropagation();
         event.preventDefault();
         clearValue();
-        ctrl.matches = [];
-        ctrl.hidden  = true;
-        ctrl.index   = getDefaultIndex();
+
+        // Force the component to blur if they hit escape
+        doBlur(true);
+
         break;
       default:
     }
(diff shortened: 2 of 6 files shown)


topherfangio:
I have tested this on Chrome and Firefox; still need to test on IE/Safari to make sure it looks/works correctly. Will do this tomorrow.

topherfangio:
Just came across another one which I think this will fix: #3287.

jelbourn:
@robertmesserle for review

src/components/autocomplete/demoBasicUsage/script.js
```diff
@@ -16,6 +16,12 @@
     self.selectedItemChange = selectedItemChange;
     self.searchTextChange   = searchTextChange;
 
+    self.newState = newState;
+
+    function newState(state) {
+      alert("Sorry! You'll need to create a Constituion for " + state + " first!");
```
robertmesserle: I was worried we might break the US by allowing people to create states.  Thankfully, there's an error. =p

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -48,6 +49,11 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   ctrl.registerSelectedItemWatcher   = registerSelectedItemWatcher;
   ctrl.unregisterSelectedItemWatcher = unregisterSelectedItemWatcher;
   ctrl.notFoundVisible               = notFoundVisible;
+  ctrl.loadingIsVisible              = loadingIsVisible;
+
+  //-- "private" methods
+  ctrl.shouldHide = shouldHide;
+  ctrl.shouldShow = shouldShow;
```
robertmesserle: Are these exposed for any reason?

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -236,7 +243,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    */
   function onListLeave () {
     noBlur = false;
-    if (!hasFocus) ctrl.hidden = true;
+    ctrl.hidden = shouldHide();
```
robertmesserle: I just want to confirm that this works properly.  I believe at the time this logic was added, we had to explicitly set it to true.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -392,9 +408,13 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
-        ctrl.matches = [];
-        ctrl.hidden  = true;
-        ctrl.index   = getDefaultIndex();
+
+        // Force the component to blur if they hit escape
+        doBlur(true);
+
+        //ctrl.matches = [];
+        //ctrl.hidden  = true;
+        //ctrl.index   = getDefaultIndex();
```
robertmesserle: Please remove commented code.  If we need it, we can pull it up in git, but we typically don't want to keep commented code in our documents.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -327,9 +334,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
 
         // cancel results if search text is not long enough
         if (!isMinLengthMet()) {
-          ctrl.loading = false;
           ctrl.matches = [];
-          ctrl.hidden  = shouldHide();
+          setLoading(false);
```
robertmesserle: The convention for this I've been using is to use `defineProperty` to create a watcher when the value is changed.  See line 23.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -209,19 +215,20 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
         $mdUtil.nextTick(function () {
 
           $mdUtil.disableScrollAround(elements.ul);
 
         }, false, $scope);
-      } else if (hidden && !oldHidden) {
-        $mdUtil.nextTick(function () {
+      }
+    } else if (hidden && !oldHidden) {
+      $mdUtil.nextTick(function () {
```
robertmesserle: While we're editing this, can we clean up some of these empty lines?

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -354,7 +370,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     hasFocus = true;
     //-- if searchText is null, let's force it to be a string
     if (!angular.isString($scope.searchText)) $scope.searchText = '';
-    if ($scope.minLength > 0) return;
+    //if ($scope.minLength > 0) return;
```
robertmesserle: Please removed commented code.

src/components/autocomplete/js/autocompleteController.js
topherfangio: The `shouldHide()` method now takes into account the `hasFocus` variable, so yes, this should work exactly as before.

src/components/autocomplete/js/autocompleteController.js
topherfangio: Nope; I was using them for debugging in the console; will remove!

src/components/autocomplete/js/autocompleteController.js
topherfangio: So, the reason I put it in a method was because the existing way caused it to (very often) wait until the next digest cycle before checking whether or not to be hidden. This isn't a huge deal, but it means that we sometimes show the suggestion list too early/too late. This is particularly apparent on IE which has slower performance due to the number of digest cycles.

Additionally, I just updated it to use you're suggested approach and it breaks the ability to have an `ng-click`able element inside the `md-not-found`. I'm sure I can dig in and figure it out if you think it's important to change this, but I feel like it's a good performance optimization.

topherfangio:
@robertmesserle Made some updates/responses. Can you re-review when you get a chance?

src/components/autocomplete/js/autocompleteController.js
robertmesserle: The approach I suggested should function exactly the same as the setter method, but you'd remove the part of that method when it actually sets the value.  This was a fairly minor complaint, so I'm fine leaving this.

It also shouldn't wait for a digest cycle using the proposed method.

topherfangio:
Heard from Rob offline that this looks good to him.

@ThomasBurleson Should be ready for merge into 0.12 whenever is convenient :smile:

Issue #2895, opened 2015-05-18
mdBottomSheet: add ability to disable clickOutsideToClose option


yccteam:
The bottom sheet gives the site a quick and beautiful "bottom popup". Is it possible to add a ```clickOutsideToClose``` feature to it? Much like the dialog box?

ThomasBurleson:
The bottomSheet already supports a clickOutside (on the mask area) to close:

*  https://github.com/angular/material-start
*  https://material.angularjs.org/#/demo/material.components.bottomSheet

yccteam:
Sorry - I was not clear. What I meant is - I want to set it to ```false```, but there is no such option.

marcysutton:
The naming should probably be updated to `tapOutsideToClose` because it's intended to be a mobile-only component.

dmackerman:
I agree with @yccteam in that this is not "necessarily" a mobile only component, because we're allowing custom templates.  On desktop, there are some very interesting options for bottom sheets (or bottom dialogs) as alternatives to full screen modal dialogs.

marcysutton:
That was according to the MD UX team. But so far there has been no discussion about what a bottom sheet should morph into in a responsive app, so that's why it's important that it have focus management (sending focus into the bottom sheet on open, and back to triggering button on close).

ThomasBurleson:
I think the `md-is-locked-open="<expression>"` should work for SideNav, BottomSheet, and Dialogs.

PR #4969, opened 2015-10-01
enhance(BottomSheet):Removed targetEvent and added clickOutsideToClose

### LINKED ISSUE #2201: [docs] md-bottom-sheet - targetEvent
Obviously documentation for `md-bottom-sheet` was copy-pasted from `md-dialog` since this is what it says for the option `targetEvent`:
> A click's event object. When passed in as an option, the location of the click will be used as the starting point for the opening animation of the the dialog.

Perhaps this needs to be fixed.
Also looking at how `md-bottom-sheet` is animated, it just comes from the bottom of the screen instead of how it's described in the docs. Perhaps option `targetEvent` could be removed all  together from `md-bottom-sheet`?

EladBezalel:
closes #2201 and #2895

### DIFF
--- a/src/components/bottomSheet/bottom-sheet.spec.js
+++ b/src/components/bottomSheet/bottom-sheet.spec.js
@@ -2,6 +2,52 @@ describe('$mdBottomSheet service', function() {
   beforeEach(module('material.components.bottomSheet'));
 
   describe('#build()', function() {
+    it('should not close when `clickOutsideToClose == true`', inject(function($mdBottomSheet, $rootElement, $material) {
+      var parent = angular.element('<div>');
+      $mdBottomSheet.show({
+        template: '<md-bottom-sheet>',
+        parent: parent,
+        clickOutsideToClose: true
+      });
+
+      $material.flushOutstandingAnimations();
+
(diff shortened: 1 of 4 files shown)


src/components/bottomSheet/bottom-sheet.spec.js
```diff
@@ -2,6 +2,52 @@ describe('$mdBottomSheet service', function() {
   beforeEach(module('material.components.bottomSheet'));
 
   describe('#build()', function() {
+    it('should not close when `clickOutsideToClose == true`', inject(function($mdBottomSheet, $rootElement, $material) {
```
jelbourn: Keep lines under 100 columns

jelbourn:
LGTM other than that one comment

src/components/bottomSheet/bottom-sheet.spec.js
EladBezalel: all of this page is more than 100 columns, should i fix it?

jelbourn: I generally wouldn't ask someone to change existing lines, but you can if you want.

PR #4972, opened 2015-10-01
fix(Dialog): change event to mousedown and mouseup

### LINKED ISSUE #3873: mdDialog: clickOutsideToClose responds to mouseUp even if mouseDown was inside the dialog
Steps to reproduce:

1) Open a dialog with clickOutsideToClose set to true.
2) Hold mouse button down anywhere inside the dialog.
3) Move the mouse outside the dialog while still holding the mouse button down.
4) Release the mouse button.
5) The dialog closes.

This can be annoying with slider controls inside the dialog - if you click and drag the slider outside the boundary of the dialog, the dialog will close. It seems like the dialog should only close if mouseDown and mouseUp are both outside the dialog.

http://codepen.io/anon/pen/yNRBBo

EladBezalel:
so we can check the origin element and close on mouseup if the target is the backdrop

closes #3873

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -648,9 +648,21 @@ function MdDialogProvider($$interimElementProvider) {
       }
       if (options.clickOutsideToClose) {
         var target = element;
-        var clickHandler = function(ev) {
-          // Only close if we click the flex container outside on the backdrop
-          if (ev.target === target[0]) {
+        var sourceElem;
+
+        // Keep track of the element on which the mouse originally went down
+        // so that we can only close the backdrop when the 'click' started on it.
+        // A simple 'click' handler does not work,
+        // it sets the target object as the element the mouse went down on.
+        var mousedownHandler = function(ev) {
+          sourceElem = ev.target;
+        };
+
+        // We check if our original element and the target is the backdrop
+        // because if the original was the backdrop and the target was inside the dialog
+        // we don't want to dialog to close.
+        var mouseupHandler = function(ev) {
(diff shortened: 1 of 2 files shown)


src/components/dialog/dialog.js
```diff
@@ -646,9 +646,14 @@ function MdDialogProvider($$interimElementProvider) {
       }
       if (options.clickOutsideToClose) {
         var target = element;
-        var clickHandler = function(ev) {
+        var originalElem;
+        var mousedownHandler = function(ev) {
+          originalElem = ev.target;
```
jelbourn: This needs a comment explaining why you keep track of this original element.

src/components/dialog/dialog.js
```diff
@@ -646,9 +646,16 @@ function MdDialogProvider($$interimElementProvider) {
       }
       if (options.clickOutsideToClose) {
         var target = element;
-        var clickHandler = function(ev) {
-          // Only close if we click the flex container outside on the backdrop
-          if (ev.target === target[0]) {
+        var sourceElem;
+
+        var mousedownHandler = function(ev) {
+          // Saving the source element for later checking if it was the backdrop
```
jelbourn: This is more along the lines of the explanation I'm looking for:
```
Keep track of the element on which the mouse originally went down so that we can only close the backdrop when the "click" started on it. A simple 'click' handler does not work because...
```

(filling in why `click` doesn't work)

ThomasBurleson:
@EladBezalel - 3 tasks:

PR #4980, opened 2015-10-02
only in menubar should the element be preserved

tflori:
enables nested menues. example:
```html
<md-menu>
  <md-button ng-click="$mdOpenMenu($event)">Open menu</md-button>
  <md-menu-content>
    <md-menu-item>
      <md-menu>
        <md-button ng-click="$mdOpenMenu($event)">Submenu</md-button>
        <md-menu-content>
          <md-menu-item>Item 1</md-menu-item>
          <md-menu-item>item 2</md-menu-item>
        </md-menu-content>
      </md-menu>
    </md-menu-item>
  </md-menu-content>
</md-menu>
```

### DIFF
--- a/src/components/menu/js/menuController.js
+++ b/src/components/menu/js/menuController.js
@@ -97,7 +97,7 @@ function MenuController($mdMenu, $attrs, $element, $scope, $mdUtil, $timeout) {
       nestLevel: self.nestLevel,
       element: menuContainer,
       target: triggerElement,
-      preserveElement: self.isInMenuBar || self.nestedMenus.length > 0,
+      preserveElement: self.isInMenuBar,
       parent: self.isInMenuBar ? $element : 'body'
     });
   };


tflori:
I signed it!

jelbourn:
@tflori Can you elaborate on what this PR does? Also, it would also need a unit test.

cc @rschmukler

tflori:
without that change it is not possible to open a menu that has an md-menu inside. i've just debuged what happens: without that change the md-menu-backdrop and the md-scroll-mask got added but the md-open-menu-container not. the reason it is not added is because it should preserve the element. 

i simply do not understand why. maybe a nested menu should preserve the element but than it have to be detected:
```javascript
this.isNested = menuContainer.hasClass('md-nested-menu')
```
and then it should be:
```javascript
    $mdMenu.show({
      scope: $scope,
      mdMenuCtrl: self,
      nestLevel: self.nestLevel,
      element: menuContainer,
      target: triggerElement,
      preserveElement: self.isInMenuBar || self.isNested,
      parent: self.isInMenuBar ? $element : 'body'
    });
```

ThomasBurleson:
@rschmukler - please review.

rschmukler:
Escalating to the UX team to figure out what the intended behavior should be

ThomasBurleson:
@joeyperson

rschmukler:
@tflori we are going to close this PR while we decide what the supported use case should be. Thanks for submitting, but we need to get clarification from the UX team on whether this interaction is supported and if it is, how it should behave in terms of placement, animations, etc.

Issue #4150, opened 2015-08-14
Calling $mdDialog.hide with an empty string resolves to the string 'true'


ddimitrop:
See codepen

http://codepen.io/anon/pen/oXOyjz

Problem at line:

https://github.com/angular/material/blob/6fd31c9a9fef9a6bda099883fd104a186d9b36dd/src/core/services/interimElement/interimElement.js#L301

If 'reason' (which is the value returned by the dialog) is empty, the expression remove(reason || SHOW_CLOSED, false) passes SHOW_CLOSED (='true') to the caller.

jelbourn:
@ThomasBurleson Looks like this came from your batch of dialog changes.

jaymanned:
It also resolves to boolean true if you call `hide()` with a boolean false as in `$mdDialog.hide(false)`. My current workaround is passing boolean values as strings, e.g. `$mdDialog.hide('false')`.

PR #4982, opened 2015-10-02
fix(interimElement): Removed unnecessary fallback

### LINKED ISSUE #4094: Impossible to resolve a dialog with a boolean (always returns true)
This is due to this line in interimElement: https://github.com/angular/material/blob/d9ba0e13a5abca05d6afa3b3b781b0dc7677183f/src/core/services/interimElement/interimElement.js#L314 (and L339):

```js
.remove(reason || SHOW_CLOSED, false)
```

Instead of using `||` this should check for `undefined`.

EladBezalel:
closes #4094 and #4150


jelbourn:
This change needs unit tests to cover those filed issues.

ThomasBurleson:
@EladBezalel - 3 tasks:

Issue #3963, opened 2015-07-31
md-dialog: forced padding


michaelkrog:
I'm a bit unsure about the specs regarding using tabs in dialogs, but I do have an issue with it.

The md-dialog-content tag has a forced padding of 24px. Putting an md-tabs inside that makes it look like this:
![skaermbillede 2015-07-31 kl 08 38 09](https://cloud.githubusercontent.com/assets/681472/9002115/4b491f5a-3760-11e5-8972-c6bb68d80f28.png)

If I set `style:"padding:0px"` on md-dialog-content, then it looks much better:
![skaermbillede 2015-07-31 kl 08 44 21](https://cloud.githubusercontent.com/assets/681472/9002124/600d819c-3760-11e5-8304-efc99dabd7e6.png)

EladBezalel:
@michaelkrog It's still happening?

michaelkrog:
@EladBezalel Yes, it seems so.

http://plnkr.co/edit/2QzUUmtVRvTwjZIX1WxY?p=preview

ThomasBurleson:
*  For the Tabs demo's use `class="md-padding"` which adds `padding:16px;`; I have removed the padding for the first demo (to show how the tabs *snaps* to the parent container).
*  Here is one of our Dialog demos with and without padding:
![screen shot 2015-10-02 at 4 35 47 am](https://cloud.githubusercontent.com/assets/210413/10243548/140063c4-68bf-11e5-98ef-0096df1c555d.png)
![screen shot 2015-10-02 at 4 35 57 am](https://cloud.githubusercontent.com/assets/210413/10243550/18473b60-68bf-11e5-80e0-c1688e820de3.png)

> You can see why we **offer** padding as a default. But it is easy enough to override.

For the Dialog, if you do not want padding just do this in your custom CSS:
```css
md-dialog-content {
  padding : 0;
}
```

EladBezalel:
@ThomasBurleson maybe we should make `.md-dialog-content` class that users can put on their content and not force padding on the element

ThomasBurleson:
@EladBezalel - great idea. Make it so!

michaelkrog:
I like the fact that padding is not forced, but the implementation is a bit confusing.

__HTML example__
```
 <md-dialog-content>
    <div class="md-dialog-content">
     ....
   </div>
 </md-dialog-content>
````

We already have directives that tell whether padding should be added or not. Why introduce another css class? And then a class that carries the same name as a directive?

[...]

PR #4983, opened 2015-10-02
update(demo, dialog):moved padding to class and removed padding from element

EladBezalel:
closing #3963

### DIFF
--- a/src/components/dialog/demoBasicUsage/dialog1.tmpl.html
+++ b/src/components/dialog/demoBasicUsage/dialog1.tmpl.html
@@ -10,7 +10,8 @@
     </div>
   </md-toolbar>
   <md-dialog-content style="max-width:800px;max-height:810px; ">
-    <div>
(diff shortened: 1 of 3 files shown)


jelbourn:
The documentation should mention this class if it's meant to be consumed by users.

ThomasBurleson:
@EladBezalel - 3 tasks:

* can you include some doc changes to mention this new `.md-dialog-content` class.

EladBezalel:
@ThomasBurleson where should I doc it?

ThomasBurleson:
Ideally it should go in a CSS doc similar to https://material.angularjs.org/0.11.2/#/CSS/button.  Something like `#/CSS/dialog` with a sample override and snapshot of its effects.

But perhaps [for time sake] just a paragraph to show here: https://material.angularjs.org/0.11.2/#/api/material.components.dialog/service/$mdDialog

PR #4985, opened 2015-10-02
added badges

fassetar:


jelbourn:
We appreciate the contribution, but we're going to stick with our current set of links. Thanks!

PR #5006, opened 2015-10-04
fix(mdDialog): fixed alert and confim `md-transition-in` class attachment

### LINKED ISSUE #4862: dialog: defaults to opacity:0 and hides
When configuring a dialog as below it happens that the opacity:0 is set.
The dialog is shown behind the scroll-mask and therefore not visible.
Sometimes the opacity is set to 1 but when using confirm it is set to 0.

```js
function dialog(event) {
        .title($translate.instant('DIALOG_TITLE_SKILL_DELETE'))
        .content('DELETE')
        .ariaLabel($translate.instant('STANDARD_BTN_DELETE'))
        .ok($translate.instant('STANDARD_BTN_DELETE'))
        .cancel($translate.instant('STANDARD_BTN_CANCEL'))
        .targetEvent(event);

    $mdDialog.show(confirm).then(function() {
        deleteSkill();
    }, function() {
    });
}

element.bind('click', function (event) {
    if (hasAccessRight) {
        dialog(event);
    }
});
```

Angular 1.4.6, Material 0.11.1

EladBezalel:
Alert and confirm dialogs had in their template class attribute that was binded to `dialog.css`, in case the css property was undefined it caused the `md-transition-in` class to be applied but then deleted.

The solution was to change the usage to ng-class and let angular set the class.

fixes #4862

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -414,7 +414,7 @@ function MdDialogProvider($$interimElementProvider) {
   function advancedDialogOptions($mdDialog, $mdTheming) {
     return {
       template: [
-        '<md-dialog md-theme="{{ dialog.theme }}" aria-label="{{ dialog.ariaLabel }}" class="{{ dialog.css }}">',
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@rschmukler - can you review this also?

PR #5009, opened 2015-10-05
fix(virtual-repeat): Prevent nested calls to virtualRepeatUpdate_

### LINKED ISSUE #4950: md-datepicker min-max demo doesn't work on Safari
I open the DatePicker demo on Safari 8.08 / OS X.

`Date-picker with min date and max date`

If I scroll up to July it is fine. Scrolling down, Oct & Nov 2015 repeat infinitely. It seems to work fine on Chrome.

mhchen:
Due to some browser issues, the $watchCollection callback that calls
virtualRepeatUpdate_ sometimes fires in the middle of a
virtualRepeatUpdate_ call. This can result in duplicate items showing up
in the virtual repeat list.

Fixes #4950

### DIFF
--- a/src/components/virtualRepeat/virtual-repeater.js
+++ b/src/components/virtualRepeat/virtual-repeater.js
@@ -422,6 +422,12 @@ function VirtualRepeatController($scope, $element, $attrs, $browser, $document, 
   /** @type {boolean} Whether this is the first time that items are rendered. */
   this.isFirstRender = true;
 
+  /**
+   * @private {boolean} Whether the items in the list are already being updated. Used to prevent
+   *     nested calls to virtualRepeatUpdate_.
+   */
+  this.isVirtualRepeatUpdating_ = false;
+
   /** @type {number} Most recently seen length of items. */
   this.itemsLength = 0;
 
@@ -543,7 +549,11 @@ VirtualRepeatController.prototype.containerUpdated = function() {
     this.unwatchItemSize_();
     this.sized = true;
     this.$scope.$watchCollection(this.repeatListExpression,
-        angular.bind(this, this.virtualRepeatUpdate_));
+        angular.bind(this, function(items, oldItems) {
+          if (!this.isVirtualRepeatUpdating_) {
+            this.virtualRepeatUpdate_(items, oldItems);
+          }
+        }));
   }
 
   this.updateIndexes_();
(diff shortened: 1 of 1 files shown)


mhchen:
Hi @kseamon,

Could you please take a look at this? The issue that is linked in the datepicker is actually one in which one of the months is being rendered twice, which breaks the height of the scroller used in the min-max logic:

Duplicate datepicker month

I've confirmed that this happens on both Mac Safari and FF on Mac. It's pretty consistently reproducible on these browsers, although not perfectly consistent, and doesn't seem to happen if you click on the calendar too quickly on load. It also only ever happens on the first time you render the calendar, so it's a pretty annoying issue to track down.

I'm not really at all confident of the fix I'm proposing, but what is happening at least in this particular issue is that [this $watchCollection](https://github.com/angular/material/blob/master/src/components/virtualRepeat/virtual-repeater.js#L503) is firing inappropriately in the middle of a virtualRepeatUpdate_(within [this loop](https://github.com/angular/material/blob/master/src/components/virtualRepeat/virtual-repeater.js#L588)), rendering the items improperly. The proposed fix prevents that $watchCollection from firing in the middle of virtualRepeatUpdate_ callback, which seems like a fairly reasonable thing to do, but it unfortunately doesn't address why it was fired at all and why only in these particular browsers. I was hoping to tap your insight about this problem and my approach. Please let me know what you think!

ThomasBurleson:
@kseamon - can you review ? Mark as "needs: merge: if appropriate.

src/components/virtualRepeat/virtual-repeater.js
```diff
@@ -531,6 +541,7 @@ VirtualRepeatController.prototype.getItemSize = function() {
  * @private
  */
 VirtualRepeatController.prototype.virtualRepeatUpdate_ = function(items, oldItems) {
+  this.isVirtualRepeatUpdating_ = true;
```
kseamon: Please add a blank line after this and one before the related line at the bottom of this method.

mhchen:
Done with fix proposed by @kseamon

ThomasBurleson:
@mhchen - please rebase with latest master. Thx.

mhchen:
@ThomasBurleson all set

jelbourn:
@ThomasBurleson can you merge this?

ThomasBurleson:
Thx @mhchen for these changes and fixes.

Issue #1960, opened 2015-03-19
md-maxlength gives error for valid number inputs


EugeneKostrikov:
See this fiddle for details http://jsfiddle.net/7fj2ad9n/3/ 
Try changing number input to see the error. Works fine for text though.

For sure maxlength doesn't make much sense for numbers :) But it would be better to skip validation if value type is not a string maybe? In my use case for instance there's a wrapper on top of md-input-container that works for both numbers and text.

gkalpak:
Indeed, `mdMaxlength` behaves strangely at times.
IMO, it should follow `ngMaxlength`'s approach and check the `$viewValue`' length (instead of the `$modelValue`'s).

brunofin:
Sometimes it makes sense in number fields, or at least it's just easier to use a length validator instead of a max value validator.

For example, I'm writing a form which one of the fields accepts any number between `0` and `999999`. That's a 6-digits length number. At first I used `min="0" md-maxlength="6"`, but not only it didn't work but it also caused the attached model's value to become `undefined`. When I realized `md-maxlength` was causing this problem, I simply switched to `min="0" max="999999"`. But I had to write down 6 times the number `9`.

In @EugeneKostrikov 's example, the input have `md-maxlength="123"`. Can you imagine to write down the number `9` 123 times in the `max` property, just to make this field work as required?

So there are two problems here: `md-maxlength` shouldn't make the value `undefined` in a number field, if it doesn't work, at least it should gracefully not work =P
And it would be good if it worked in number fields as well, so we have just one way to validate fields' length, even for number fields, after all, number also do have a length.

PR #5011, opened 2015-10-05
fix(mdInput): Fix md-maxlength validation

edbentinck:
`mdMaxlength` currently doesn't validate properly on inputs with `type="number"`. When the `$modelValue` returns a number, its `length` property is `undefined`. This change parses the `$modelValue` to a string when it's appropriate to do so.

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -432,7 +432,7 @@ function mdMaxlengthDirective($animate) {
       if (!angular.isNumber(maxlength) || maxlength < 0) {
         return true;
       }
-      return ( modelValue || element.val() || viewValue || '' ).length <= maxlength;
+      return ( element.val() || viewValue || '' ).length <= maxlength;
     };
 
     function renderCharCount(value) {
--- a/src/components/input/input.spec.js
+++ b/src/components/input/input.spec.js
@@ -191,6 +191,25 @@ describe('md-input-container directive', function() {
       expect(pageScope.form.foo.$error['md-maxlength']).toBeFalsy();
       expect(getCharCounter(el).length).toBe(0);
     });
+
+    it('should not show an error on inputs with type="number" and a valid number value', function() {
+      var el = $compile('<form name="form">' +
+        ' <md-input-container>' +
+        '   <input type="number" md-maxlength="5" ng-model="foo" name="foo">' +
+        ' </md-input-container>' +
+        '</form>')(pageScope);
+
+      pageScope.$apply();
+      el.find('input').val('123').triggerHandler('input');
+      expect(pageScope.form.foo.$error['md-maxlength']).toBeFalsy();
+      expect(getCharCounter(el).length).toBe(1);
+      expect(getCharCounter(el).text()).toBe('3/5');
+
+      el.find('input').val('123456').triggerHandler('input');
+      expect(pageScope.form.foo.$error['md-maxlength']).toBeTruthy();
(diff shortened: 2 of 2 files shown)


edbentinck:
I signed the Google Contributor License Agreement (CLA).

ThomasBurleson:
@topherfangio  - can you review ? Mark as "needs: merge: if appropriate.

src/components/input/input.js
```diff
@@ -404,6 +404,9 @@ function mdMaxlengthDirective($animate) {
       if (!angular.isNumber(maxlength) || maxlength < 0) {
         return true;
       }
+      if (angular.isNumber(modelValue)) {
+        modelValue = modelValue.toString();
```
topherfangio: This will fail for very large numbers:

```js
var num = 2000000929340720349810293480192834091823409172394712934;
num.toString() // => "2.0000009293407202e+54"
num.length // => 22; should be 55
```

Instead of checking the model value, perhaps we should check the element value in this case? Does that provide what we need?

edbentinck: That would be ideal. [It has been mentioned before](https://github.com/angular/material/issues/1960#issuecomment-83494622) as a possible solution, so perhaps line `410` can just be changed to `return ( element.val() || viewValue || '' ).length <= maxlength;`

edbentinck:
@topherfangio Please review my latest commit for the corrections :)

topherfangio:
@edbentinck Fix looks good!

Can you please add a quick test that shows what this fixes? There are a lot of others in `input.spec.js` so you should be able to copy one of them to get started. Let me know if you have any trouble or need some guidance!

edbentinck:
@topherfangio I've added a test for the fix now, see what you think. Let me know if I ought to make any changes.

topherfangio:
@ThomasBurleson Assuming this passes, it looks good to me. Would recommend waiting until after 0.12 to merge, though.

EladBezalel:
i assume @ThomasBurleson would ask this to be squashed into one commit :)

edbentinck:
@EladBezalel Cool, thanks. Merged the 3 commits into one :) @ThomasBurleson

ThomasBurleson:
@edbentinck - Thanks for your PR; the test addition is also very helpful.

ThomasBurleson:
@edbentinck - plz rebase with master. I will merge afterwards.

edbentinck:
@ThomasBurleson - rebased now, hope that works for you!

edbentinck:
Anything else you need from me, @ThomasBurleson? Happy to help if so

EladBezalel:
@edbentinck sorry for this being delayed, please check against HEAD and see if the issue is still happening, closing for now, if you think it's valid please rebase and i will re open

edbentinck:
It's [still an issue](http://codepen.io/edbentinck/pen/YqNGyG). I will rebase later when I have time.

edbentinck:
I've rebased now, @EladBezalel. Could you re-open?

ThomasBurleson:
@edbentinck - We are unable to reopen this closed PR. Would you mind submitting a new PR and we will merge your fix into master. Thanks for your help.

edbentinck:
[Done](https://github.com/angular/material/pull/7622). Thanks

Splaktar:
@AnirudhaGohokar please open a new issue with a CodePen that reproduces the issue.

Issue #4873, opened 2015-09-28
md-tabs inside md-dialog: width issue


lmignot:
I've created a codepen: https://codepen.io/anon/pen/XmNqOz

If you have a custom dialog with md-tabs inside and md-tabs are set to md-dynamic-height and you have a container inside the dialog with a width or min-width set, neither Chrome nor Firefox calculate the width of the tab labels correctly and truncates the label

Browsers: Chrome 45.0.2454.85, Firefox 40.0.3, IE 11.0.9600

IE11 works fine... below is a screenshot of how it renders in Chrome

![mdtabs in mddialog](https://cloud.githubusercontent.com/assets/372523/10132501/9535098e-65ce-11e5-8039-a3279d59477f.PNG)

lmignot:
I've narrowed it down to `getMaxTabWidth` in tabsController (line 540) 
Seems like when the tabs controller is initialising the element hasn't been rendered yet so `clientWidth` is 0. Hence max-width on `md-tab-item` is set to 0

kskoropada:
Any workaround?

websirnik:
Similar https://github.com/angular/material/issues/4834 & https://github.com/angular/material/issues/4744

hannahhoward:
maybe set a $scope.$watch on $element.clientWidth and call getMaxTabWidth when it changes? Seems like an expensive solution. Or there's the always famous use $timeout to delay all that checking (not recommending this one :) in the meantime since if I understand maxTabWidth is a public property of the controller I'm going to put a directive underneath that does my $scope.$watch solution. will gist it if anyone needs. wouldn't the proper thing for the md-dialog to somehow let the tab controller to recheck width once it's visible?

osuritz:
It was working fine for us (at least on Chrome) in 0.10... not sure what has changed yet.

hannahhoward:
0.11 is the breaking point I think.

also, if anyone needs a workaround while the team fixes it, this little directive seemed to do the trick for me:
https://gist.github.com/hannahhoward/35b87c2b5d0154a7edd4

hannahhoward:
I don't recommend that as a permaneant solution -- checking the clientWidth constantly will probably (not sure) trigger browser repaints and slow things down.

osuritz:
Thanks for the temporary fix, @hannahhoward!

blmcquillan:
+1

lmignot:
[...]

Issue #4834, opened 2015-09-25
mdDialog with mdTabs tabs are not rendered correctly


jeroenvheel:
When placing mdTabs in a mdDialog the mdTab is not rendered correctly in the master branch

![image](https://cloud.githubusercontent.com/assets/228655/10096105/480e5928-636f-11e5-89cc-627e87fbe783.png)

```html
<md-dialog class="admin-dialog">
    <md-toolbar>
        <div class="md-toolbar-tools">
            <h2>Create a new User</h2>
        </div>
    </md-toolbar>
    <md-dialog-content>
        <form name="createUser" id="createUser" role="form">
            <md-tabs md-dynamic-height md-stretch-tabs="always" md-center-tabs>

                <md-tab label="Gebruiker">

                    <md-input-container>
                        <label>Name</label>
                        <input id="name" name="name" type="text" md-maxlength="50" required
                               ng-model="user.Name" />
                    </md-input-container>

                    <md-input-container>
                        <label>Description</label>
                        <input id="description" name="description" type="text" md-maxlength="250"
                               ng-model="user.Description" />
                    </md-input-container>

                    <md-input-container>
                        <label>E-mail</label>
                        <input id="email" name="email" type="email" md-maxlength="255" required
                               ng-model="user.Email" />
                    </md-input-container>

                    <md-input-container>
                        <label>Password</label>
                        <input id="password" name="password" type="password" md-maxlength="255" required
                               ng-model="user.Password" />
                    </md-input-container>

                    <md-input-container>
                        <label>Hoofd Bedrijf</label>
                        <md-select name="maincompany" ng-model="user.MainCompany" required>
                            <md-option ng-repeat="company in companyList" value="{{company.Name}}">{{company.Name}}</md-option>
                        </md-select>
                    </md-input-container>

                </md-tab>
                <md-tab label="Rechten">
                    <md-switch ng-model="user.ManageCompanies" aria-label="ManageCompanies?" ng-disabled="!currentUser.ManageCompanies">
                        Company Beheer
                    </md-switch>
                    <md-switch ng-model="user.ManageProjects" aria-label="ManageProjects?">
                        Projecten Beheer
                    </md-switch>
                    <md-switch ng-model="user.ManageUsers" aria-label="ManageUsers?">
                        Gebruiker Beheer
                    </md-switch>

                </md-tab>

            </md-tabs>
        </form>
    </md-dialog-content>
    <div class="md-actions" layout="row">

[...]

Issue #4744, opened 2015-09-21
md-tabs doesn't render itself correctly if is used inside md-sidenav. in IE


kikhtenko:
Hi, I made an example here:
http://codepen.io/anon/pen/xwOrRq

if you use md-tabs in md-sidebar it works fine in Chrome but UI is broken in IE 11.

Regards,

ivorobkalo:
+1

kikhtenko:
This is more about md-sidebar in IE. Any UI will be broken in the right sidenav.

kikhtenko:
http://codepen.io/anon/pen/QjEgYb

asadovnikov:
+

medcomcr:
+1

Nickproger:
+1

websirnik:
The demo above is also broken for me in Chrome. This is related to https://github.com/angular/material/issues/4834
![screen shot 2015-09-26 at 13 20 00](https://cloud.githubusercontent.com/assets/282871/10117512/7af62108-6451-11e5-9b6d-18b8b40b1250.png)

jeroenvheel:
+1

kskoropada:
+1

robertmesserle:
I am not able to recreate this in IE11.  Are there any specific steps needed?  Does screen/browser size matter?

robertmesserle:
Closing for now.  If this issue still exists, please provide steps to recreate against the latest code.

austinthedeveloper:
I did not see this issue when I created my issue: #6420 . My solution will fix that issue.

Issue #4743, opened 2015-09-21
TabsController handleWindowResize $apply


MarijnMensinga:
Hi,

While testing I detected that upon a window resize event the tabsController triggers an `$digest already in progress` error. This happens because [here] (https://github.com/angular/material/blob/9ac7496ed59afc58a38ec3b1793001f3cbbcfbd5/src/components/tabs/js/tabsController.js#L338) an `$apply()` is called. 

Normally this does not occur, but since it is possible to have custom listeners that trigger on the same event an `$digets` might already be in progress.

So I suggest the `$scope.$apply` gets replaced by `$scope.$applyAsync` to fix this. Since the body of the function aleady implements this behaviour by calling `$mdUtil.nextTick` I think this is not a problem:

```javascript
function handleWindowResize () {
  $scope.$apply(function () {
    ctrl.lastSelectedIndex = ctrl.selectedIndex;
    ctrl.offsetLeft        = fixOffset(ctrl.offsetLeft);
    $mdUtil.nextTick(ctrl.updateInkBarStyles, false);
    $mdUtil.nextTick(updatePagination);
  });
}
```
[Source from here] (https://github.com/angular/material/blob/9ac7496ed59afc58a38ec3b1793001f3cbbcfbd5/src/components/tabs/js/tabsController.js#L337) 

Stacktrace:
```
Error: [$rootScope:inprog] $digest already in progress
http://errors.angularjs.org/1.4.5/$rootScope/inprog?p0=%24digest
minErr/<@project/library/Angular/1.4.5/angular.js:68:14
beginPhase@project/library/Angular/1.4.5/angular.js:16235:1
$RootScopeProvider/this.$get</Scope.prototype.$digest@project/library/Angular/1.4.5/angular.js:15669:13
$RootScopeProvider/this.$get</Scope.prototype.$apply@project/library/Angular/1.4.5/angular.js:15986:17
handleWindowResize@project/library/AngularMaterial/0.10.1/angular-material.js:16531:9
createEventHandler/eventHandler@project/library/Angular/1.4.5/angular.js:3293:11
createScrollableDirective/directive/<.link/<@project/angular-scrollable.js:126:15
$RootScopeProvider/this.$get</Scope.prototype.$digest@project/library/Angular/1.4.5/angular.js:15707:40
$RootScopeProvider/this.$get</Scope.prototype.$apply@project/library/Angular/1.4.5/angular.js:15986:17
tick@project/library/Angular/1.4.5/angular.js:11231:29
```

[...]

PR #5017, opened 2015-10-05
fix(mdTabs): Tab label width correction in mdDialog

FrankMeyer:
The Tab label width is set correctly when used in a dialog.
Changed one line to a previous version (1 commit prior to 4a16038be56b285fdc2c7cbdfcd8853d65408b21)
That commit addressed issue #4743. This issue is not reproduced by reverting this line of code.

Addressing issues #4834 #4873 #4744

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -486,7 +486,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
    */
   function shouldPaginate () {
     if (ctrl.noPagination || !loaded) return false;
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@robertmesserle - can you review ?

robertmesserle:
@ThomasBurleson This looks like it would break other use-cases.  Still looking though.

robertmesserle:
@ThomasBurleson Actually, this looks okay.  It's a pretty brittle fix, but it seems to work.

@FrankMeyer Can you add a comment explaining the line change so that it isn't confusing in the future?

robertmesserle:
I believe I have a proper fix for this.  Closing.

PR #5031, opened 2015-10-06
fix(repeater): resize scroller correctly

### LINKED ISSUE #5027: Virtual Repeat Scroller doesn't shrink when items are removed
I start out with the vertical repeat demo with 1000 items. Then I remove the last 900 items. Now, after I scroll past 100 it repeats the last page infinitely until it reaches the bottom of the original thousand.

Demonstration: https://codepen.io/colinskow/pen/PPmNbL

colinskow:
There was some recursion inside `VirtualRepeatController.prototype.virtualRepeatUpdate_` which was causing the scroller not to shrink properly when items were removed. I simply moved the changes check below the recursion so that it detects changes properly.

In addition, in `onDemand` mode an undefined length would stop the container from sizing correctly. An undefined length will now be changed to zero.

It could also possibly fix https://github.com/angular/material/issues/4950 and may be a better solution than https://github.com/angular/material/pull/5009, but I haven't yet tested it against that issue.

 fixes #5027

### DIFF
--- a/src/components/virtualRepeat/virtual-repeater.js
+++ b/src/components/virtualRepeat/virtual-repeater.js
@@ -574,19 +574,19 @@ VirtualRepeatController.prototype.getItemSize = function() {
  * @private
  */
 VirtualRepeatController.prototype.virtualRepeatUpdate_ = function(items, oldItems) {
-  var itemsLength = items ? items.length : 0;
+  var itemsLength = items && items.length || 0;
   var lengthChanged = false;
-
-  if (itemsLength !== this.itemsLength) {
-    lengthChanged = true;
-    this.itemsLength = itemsLength;
-  }
 
   // If the number of items shrank, scroll up to the top.
   if (this.items && itemsLength < this.items.length && this.container.getScrollOffset() !== 0) {
     this.items = items;
     this.container.resetScroll();
     return;
+  }
+
+  if (itemsLength !== this.itemsLength) {
+    lengthChanged = true;
+    this.itemsLength = itemsLength;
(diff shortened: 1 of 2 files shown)


colinskow:
@mhchen Could you test the calendar against this PR and see if it impacts the issue? I'm having trouble getting the demos to run at all in Safari.

colinskow:
@mhchen Never mind. The datepicker bug seems to happen less often, but still happens with these changes. I did confirm that #5009 fixes that issue though. By combining our two PRs everything should work as expected.

ThomasBurleson:
@kseamon - can you review ? Mark as "needs: merge: if appropriate.

src/components/virtualRepeat/virtual-repeater.js
```diff
@@ -531,14 +531,9 @@ VirtualRepeatController.prototype.getItemSize = function() {
  * @private
  */
 VirtualRepeatController.prototype.virtualRepeatUpdate_ = function(items, oldItems) {
-  var itemsLength = items ? items.length : 0;
+  var itemsLength = items && items.length ? items.length : 0;
```
kseamon: somewhat more concisely, could be

items && items.length || 0

kseamon:
Looks reasonable, overall.

ThomasBurleson:
@colinskow - please squash and rebase with latest from master.

colinskow:
Oops... looks like a squashed too much. Let me revise.

colinskow:
There we go, all in order.

Issue #4686, opened 2015-09-17
Update gulp-sass to v2 to support Node.js 4.0


oliverzy:
Since Node.js 4.0 has been released, gulp-sass has to be updated to the latest to support it

JordanMajd:
#5045 will resolve this issue.

PR #5045, opened 2015-10-07
Upgrade build system to support Node 4.1

colinskow:
The current build system uses a version of Node-Sass which does not work with recent versions of Node. However, the newest version of Node-Sass (3.3.3) writes quotes slightly different, which causes `cssToNgConstant` to fail. This is fixed by adding a global flag to the escape regex.

I am unable to build Angular Material on Node 4.x without this patch.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
     "gulp-ng-html2js": "^0.1.8",
     "gulp-plumber": "^1.0.0",
     "gulp-rename": "^1.2.0",
-    "gulp-sass": "^1.3.3",
+    "gulp-sass": "^2.0.4",
     "gulp-uglify": "^1.2.0",
     "gulp-util": "^3.0.1",
     "gulp-webserver": "^0.9.1",
--- a/scripts/gulp-utils.js
+++ b/scripts/gulp-utils.js
@@ -290,7 +290,7 @@ exports.cssToNgConstant = function(ngModule, factoryName) {
   return through2.obj(function(file, enc, next) {
 
     var template = '(function(){ \nangular.module("%1").constant("%2", "%3"); \n})();\n\n';
-    var output = file.contents.toString().replace(/\n/g, '').replace(/\"/,'\\"');
+    var output = file.contents.toString().replace(/\n/g, '').replace(/\"/g,'\\"');
 
     var jsFile = new gutil.File({
       base: file.base,


topherfangio:
Thanks for the PR! I actually ran into this issue a while back, but didn't have time to investigate.

Did you test this against node 0.10.x? I believe that is what we are currently using, and I'd like to make sure it doesn't have any negative side-effects.

colinskow:
@topherfangio according to Travis the build is set for Node 0.12 and it passes. If I remember correctly Node-Sass 3.x does not support Node 0.10.

colinskow:
Since it is just the theming engine effected, later today I will build the docs and visually inspect the demos carefully.

colinskow:
@topherfangio I checked the Node-Sass docs, and 3.3.x DOES work with Node 0.10.
https://github.com/sass/node-sass/blob/master/package.json

colinskow:
I also built the docs using this patch and all the demos look good on both Chrome and Safari. All tests pass.

JordanMajd:
+1 for merge.

Went to go work on an issue, but I can't build right now because I'm using the latest Node for another project. In the meantime I guess I will install NVM so I can have multiple versions of node for different projects.

Looks like this will fix #4686

EladBezalel:
@colinskow can you check for node 5.0.0 please?

colinskow:
@EladBezalel I managed to install Node 5.0.0 and all tests were successful.

However, I ran into a number of Node/NPM issues such as https://github.com/nodejs/node/issues/3606 (NPM must be manually removed before upgrading). I also had trouble getting the latest Gulp to install. Node 4.2.2 LTS is much more dependable for now.

Todd-Werelius:
### Environment 
windows 10
node 5 
npm 3.4.0
gulp 3.9.0 

### additional fixes 
`package.json update`
```
"gulp-sass": "^2.1.0",
```

`update to gulp-util.js as per @colinskow` ( you rock man! ) , line 293 ( as of this post ) 
```json
 var output = file.contents.toString().replace(/\n/g, '').replace(/\"/g,'\\"');
```

### build pass
### docs pass
### tests pass

Can't see any issues at all now

JordanMajd:
Nice guys, thanks for working on this!

ThomasBurleson:
@jelbourn - If I merge this, will this create issues on Travis?

jelbourn:
@ThomasBurleson If everything works cleanly from a fresh install Travis should be fine.

PR #5057, opened 2015-10-07
fix(toast.scss): Change height to defined for IE11

### LINKED ISSUE #4946: IE 11 md-toast needs height defined for auto centering
(From the code pen example)
```javascript
$mdToast.show(
                                $mdToast.simple()
                                    .content(translation)
                                    .position('top center')
                                    .parent(container)
                                    .hideDelay(6000)
                            );
```
Center works fine in FF and Chrome. Left aligns in IE 11.
CSS fix for working in IE 11.
```css
md-toast {
    width: 100%;
}
```

jsr6720:
Height updated to value below min-height. IE11 does not layout correctly
without. Tried 'inherit' but caniuse.com says not supported.

Fixes #4946

### DIFF
--- a/src/components/toast/toast.scss
+++ b/src/components/toast/toast.scss
@@ -18,7 +18,7 @@ md-toast {
   font-size: 14px;
   cursor: default;
 
-  height:auto;
+  height: 0px;
   max-height: 7*$toast-height;
   max-width: 100%;
 


jsr6720:
Ugh. My github CLA is under my personal account, but the commit is under my corporate account.
I've added both emails to both google settings and github settings. Should I just re-submit another pull request with the correct author? Or will the check re-run?

PR #5058, opened 2015-10-07
fix(divider): Fix stlying for md-divider inside md-list-item.

### LINKED ISSUE #3021: md-divider not render properlly
Updated to v0.9.6 and my list view item using the divider is not render properly

<md-divider></md-divider>

All my dividers are render to a fixed position

topherfangio:
When an `<md-divider>` was used inside of a `<md-list-item>` with an `<md-button>`, the divider would improperly show up at the bottom of the list, instead of the bottom of each list item.

Fix by making the list item's position relative.

Also, we automatically apply special styling when we detect an `<md-button>` inside of the list item because we assume that it wraps all of the elements. If you simply wanted to use a secondary `<md-button>` or other clickable button inside a list, we would incorrectly apply this styling.

Fix by automatically excluding `md-secondary` buttons and providing an additional `md-exclude` class that can be added to buttons.

Fixes #3021.

### DIFF
--- a/src/components/list/list.js
+++ b/src/components/list/list.js
@@ -63,6 +63,10 @@ function mdListDirective($mdTheming) {
  *  </md-list>
  * </hljs>
  *
+ * _**Note:** We automatically apply special styling when the inner contents are wrapped inside
+ * of a `<md-button>` tag. This styling is automatically ignored for `class="md-secondary"` buttons
+ * and you can include a class of `class="md-exclude"` if you need to use a non-secondary button
+ * that is inside the list, but does not wrap the contents._
  */
 function mdListItemDirective($mdAria, $mdConstant, $mdUtil, $timeout) {
   var proxiedTypes = ['md-checkbox', 'md-switch'];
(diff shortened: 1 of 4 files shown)


src/components/list/list.js
```diff
@@ -86,7 +90,7 @@ function mdListItemDirective($mdAria, $mdConstant, $mdUtil, $timeout) {
         }
         if (hasProxiedElement) {
           wrapIn('div');
-        } else if (!tEl[0].querySelector('md-button')) {
+        } else if (!tEl[0].querySelector('md-button:not(.md-secondary):not(.md-exclude)')) {
```
ThomasBurleson: Where is `.md-exclude` documented?   At least let's provide some information in the API section.

topherfangio: It's documented on lines 66-69 above which appear as a note in the directive API docs. Does it need to go somewhere else too?

> and you can include a class of `class="md-exclude"` if you need to use a non-secondary button

PR #5074, opened 2015-10-08
add the reload function inside the themingProvider

fmarcelo:
We were trying to register theme on the fly and the only way to do it is assign $mdThemingProvider as $provide.value('themeService', $mdThemingProvider) inside the angular config.

In this case the themeService will be accessible to angular controllers and we can registerTheme with $http resolve values.

To apply the theme we call themeService.reload($injector);

### DIFF
--- a/src/core/services/theming/theming.js
+++ b/src/core/services/theming/theming.js
@@ -145,6 +145,7 @@ function ThemingProvider($mdColorPalette) {
     alwaysWatchTheme: function(alwaysWatch) {
       alwaysWatchTheme = alwaysWatch;
     },
+    reload: registerTheme,
     $get: ThemingService,
     _LIGHT_DEFAULT_HUES: LIGHT_DEFAULT_HUES,
     _DARK_DEFAULT_HUES: DARK_DEFAULT_HUES,


breity:
This will be very useful for us too!  One question:

Shouldn't the reload function be `reload: generateThemes` instead, so that the new themes actually get generated and output as css in the head?  `registerTheme` doesn't actually output the css, right?  Am I missing something here?

Thanks!

Frank3K:
+1

ThomasBurleson:
@rschmukler, @jelbourn - can you review plz.

jelbourn:
This needs documentation and at least one unit test making sure that calling `reload` does what you want it to do.

JeJan:
in 1.0-rc6 it has to be "reload: generateAllThemes" to work, but this still throws some warning logs

chenav:
+1 (a real important feature for multi-sites angular-based CMS initiative, where the theming is dynamic)

EladBezalel:
@jelbourn i remember that registering themes on the fly was an idea,
If it's not valid please close it.

jelbourn:
The change seems okay, but it still needs unit testing.

EladBezalel:
@fmarcelo please add tests

CodyMorris:
Would this also allow dynamic registration of palettes ? In my use case I need to get hex values from a backend service and use that to create palettes.  No luck with any of these changes

ThomasBurleson:

PR #5075, opened 2015-10-08
feat(mdDialog): added openFrom and closeTo properties

### LINKED ISSUE #4228: md-dialog close animation incorrect when opened from md-menu item
When closing a dialog that's opened from a md-menu it animates back to the top left of the page rather than from the location where it appeared from.

It would also be nice to explicitly set the element to animate out from and back to, as in my case I'd like the animation to emanate and return to the component that had the menu, and not the menu item itself.

EladBezalel:
showFrom specifies the origin of the transition animation
closeTo specifies the target of the transition animation

both except string (query selector), element and rect

closes #4228

### DIFF
--- a/src/components/dialog/demoOpenFromCloseTo/index.html
+++ b/src/components/dialog/demoOpenFromCloseTo/index.html
@@ -1,0 +1,30 @@
+<div ng-controller="AppCtrl" layout="row" ng-cloak style="height: 300px">
+  <div layout="column" layout-align="center center"
+       style="background-color: #f2f2f2" class="md-padding">
+    <span id="left">left</span>
+  </div>
+  <div layout="column" layout-align="center">
+    <p class="inset">
+      A dialog can specify its origin and target with <code>openFrom</code> and
+      <code>closeTo</code> properties.
+
+      The options showFrom and closeTo can be specified as a string [where internally
+      querySelector( ) is used to perform the DOM element lookup],
+      element or an Rect object [with top, left, width, height fields].
+    </p>
+
+    <div class="dialog-demo-content" layout="row" layout-wrap>
+      <md-button class="md-primary md-raised" ng-click="openFromLeft()" flex flex-md="100">
+        Open From Left Close To Right
+      </md-button>
+      <md-button class="md-primary md-raised" ng-click="openOffscreen()" flex flex-md="100">
+        From Offscreen
+      </md-button>
+    </div>
+  </div>
+
+  <div layout="column" layout-align="center center"
+       style="background-color: #f2f2f2" class="md-padding">
+    <span id="right">right</span>
+  </div>
+</div>
--- a/src/components/dialog/demoOpenFromCloseTo/script.js
+++ b/src/components/dialog/demoOpenFromCloseTo/script.js
@@ -1,0 +1,38 @@
+angular.module('dialogDemo2', ['ngMaterial'])
+
+.controller('AppCtrl', function($scope, $mdDialog) {
+  $scope.openFromLeft = function() {
+    $mdDialog.show(
+      $mdDialog.alert()
+        .clickOutsideToClose(true)
+        .title('Opening from the left')
+        .content('Closing to the right!')
+        .ariaLabel('Left to right demo')
+        .ok('Nice!')
+        // You can specify either sting with query selector
+        .openFrom('#left')
+        // or an element
+        .closeTo(angular.element(document.querySelector('#right')))
+    );
+  };
+
+  $scope.openOffscreen = function() {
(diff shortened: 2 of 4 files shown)


src/core/util/animation/animate.js
```diff
@@ -87,12 +87,13 @@ function AnimateDomUtils($mdUtil, $q, $timeout, $mdConstant, $animateCss) {
...
     calculateZoomToOrigin: function (element, originator) {
       var origin = originator.element;
+      var bounds = originator.bounds;
       var zoomTemplate = "translate3d( {centerX}px, {centerY}px, 0 ) scale( {scaleX}, {scaleY} )";
       var buildZoom = angular.bind(null, $mdUtil.supplant, zoomTemplate);
       var zoomStyle = buildZoom({centerX: 0, centerY: 0, scaleX: 0.5, scaleY: 0.5});
 
-      if (origin) {
-        var originBnds = self.clientRect(origin) || self.copyRect(originator.bounds);
+      if (origin || bounds) {
```
ThomasBurleson: :+1:

src/components/dialog/dialog.js
```diff
@@ -572,39 +578,67 @@ function MdDialogProvider($$interimElementProvider) {
...
+
+      if(options.targetEvent) {
+        options.origin = saveBoundingClientRect(options.targetEvent.target, options.origin);
       }
-      // If we have a reference to a raw dom element, always wrap it in jqLite
-      options.parent = angular.element(options.parent || $rootElement);
 
+      options.parent = getDomElement(options.parent, $rootElement);
+
+      options.closeTo = getDomElement(options.closeTo);
```
ThomasBurleson: This is very confusing to read... I think a refactoring is needed.

src/components/dialog/demoOpenFromCloseTo/script.js
```diff
@@ -0,0 +1,38 @@
...
+      $mdDialog.alert()
+        .clickOutsideToClose(true)
+        .title('Opening from the left')
+        .content('Closing to the right!')
+        .ariaLabel('Left to right demo')
+        .ok('Nice!')
+        // You can specify either sting with query selector
+        .openFrom('#left')
+        // or an element
+        .closeTo(angular.element(document.querySelector('#right')))
```
ThomasBurleson: we should simply support .closeTo('#right'); where internal we use `querySelector()`

> also perhaps we provide a second argument `.closeTo(  '#right',  '#nonBodyParent' )`

src/components/dialog/demoOpenFromCloseTo/index.html
```diff
@@ -0,0 +1,26 @@
+<div ng-controller="AppCtrl" layout="row" ng-cloak style="height: 300px">
+  <div layout="column" layout-align="center center"
+       style="background-color: #f2f2f2" class="md-padding">
+    <span id="left">left</span>
+  </div>
+  <div layout="column" layout-align="center">
+    <p class="inset">
+      A dialog can specify its origin and target with <code>showFrom</code> and <code>closeTo</code> properties,
```
ThomasBurleson: **openFrom** instead of **showFrom**

src/components/dialog/demoOpenFromCloseTo/index.html
```diff
@@ -0,0 +1,26 @@
...
+      <code>showFrom</code> and <code>closeTo</code> both can get string(query selector), element or a rect.
```
ThomasBurleson: The options  <code>showFrom</code> and <code>closeTo</code> can be specified as a string [where internally `querySelector( )` is used to perform the DOM element lookup], element or an Rect object [with top, left, width, height fields].

src/components/dialog/dialog.js
```diff
@@ -336,6 +336,10 @@ function MdDialogDirective($$rAF, $mdTheming, $mdDialog) {
  *   - `targetEvent` - `{DOMClickEvent=}`: A click's event object. When passed in as an option,
  *     the location of the click will be used as the starting point for the opening animation
  *     of the the dialog.
+ *   - `openFrom` - `{string|DOMElement|object}`: The query selector, DOM element or the rect
+ *     that the dialog will open from
```
ThomasBurleson: or the Rect object that is used to determine the bounds (top, left, height, width) from which the Dialog will originate.

src/components/dialog/dialog.js
EladBezalel: what do you suggest?

ThomasBurleson: I will fix.

EladBezalel:
Done

PR #5077, opened 2015-10-09
feat(slider): add md-max class when at max value

### LINKED ISSUE #3513: md-slider md-min but no md-max
Sliding an `md-slider` to the minimum value adds a class `md-min` which effectively whites out the thumb control (making it a circular outline). Why is there no `md-max` class when the `md-slider` is at the maximum value? I'm not sure what the style *should* be but it seems odd to have that inconsistency.

The code in question is:
```javascript
/**
 * @param percent 0-1
 */
function setSliderPercent(percent) {
  activeTrack.css('width', (percent * 100) + '%');
  thumbContainer.css(
    'left',
    (percent * 100) + '%'
  );
  element.toggleClass('md-min', percent === 0);
}
```
While in there, how about:
```javascript
var percentStr = (percent * 100) + '%';
activeTrack.css('width', percentStr);
thumbContainer.css('left', percentStr);
```
to DRY it out a little?

CarmenPopoviciu:
add md-max class when at max value. This is to keep
things consistent with the fact that md-min is added when
at min value

fixes #3513

### DIFF
--- a/src/components/slider/slider.js
+++ b/src/components/slider/slider.js
@@ -291,12 +291,13 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
      * @param percent 0-1
      */
     function setSliderPercent(percent) {
-      activeTrack.css('width', (percent * 100) + '%');
-      thumbContainer.css(
-        'left',
-        (percent * 100) + '%'
-      );
-      element.toggleClass('md-min', percent === 0);
+        var percentStr = (percent * 100) + '%';
+
+        activeTrack.css('width', percentStr);
+        thumbContainer.css('left',percentStr);
+
+        element.toggleClass('md-min', percent === 0);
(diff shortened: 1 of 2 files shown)


CarmenPopoviciu:
Signed CLA

src/components/slider/slider.js
```diff
@@ -291,12 +291,13 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
...
-        'left',
-        (percent * 100) + '%'
-      );
-      element.toggleClass('md-min', percent === 0);
+        var percentStr = (percent * 100) + '%';
+
+        activeTrack.css('width', percentStr);
+        thumbContainer.css('left',percentStr);
+
+        element.toggleClass('md-min', percent === 0);
```
ThomasBurleson: :+1:

PR #5089, opened 2015-10-09
fix(layout): Fix IE align-items: center issue.

topherfangio:
In Internet Explorer 10/11, flex items inside of a flex column container that had `align-items: center;` set would not properly calculate their width, resulting in 0-width items.

Fix by setting `max-width: 100%` and `box-sizing: border-box` on the child elements, and fixing the demos to properly use `layout-fill` where appropriate.

Also fix issue checkbox demo issue with flex fieldsets not properly rendering in IE.


ThomasBurleson:
@topherfangio :+1:  Awesome work.

topherfangio:
@ThomasBurleson Thanks! FYI - I merged your branch and tested; as far as I can tell, it had identical results :smile:

PR #5090, opened 2015-10-09
fix(layouts): interpolated values, validations, test & CSS fixes

### LINKED ISSUE #2954: flex=number within layout-fill does not work
In the codepen below, you should be seeing a rectangle block of 30%, but the div ends up taking on no value.

http://codepen.io/anon/pen/yNaKbK

### LINKED ISSUE #4902: offset attribute conflict (since 0.11.0)
ng-attr-offset is ignored when using md-material builds 0.11.1 and current master
Here is rendered html example with svg gradient, where other (here: stop-color) ng-attr- are evaluated but ng-attr-offset is ignored:
```
<linearGradient x1="100%" y1="100%" x2="100%" y2="0%" id="hpp-gradient">
            <stop ng-attr-offset="0%" ng-attr-stop-color="{{fill}}" stop-color="#93CE8C"></stop>
            <stop ng-attr-offset="{{(complete)?complete:0}}" ng-attr-stop-color="{{fill2}}" stop-color="white></stop>
</linearGradient>
```
This was working ok in 10.0.1.

### LINKED ISSUE #4959: how use expressions for flex value
i would like to use an expression in my table div flex values within a nested ng-repeat

```
<div class="table">
  <div class="row" ng-repeat="row in ::data">
      <div class="cell ng-repeat="field in ::fields" flex="field.width">
         {{row[field.name].value}}
      </div>
```

but this seems to simply add a class name of "flex-" + expression : ie. "flex-field.width"

i tried wrapping in single and double braces, nothin works. when I use ng-class to emulate i get mixed results.

what is the proper way to use expressions with flex?

### LINKED ISSUE #4994: layout: `flex="flex"` should be replaced to just `flex`
Please compare:
http://codepen.io/evgeniyoz/pen/ojZoJb?editors=100
http://codepen.io/evgeniyoz/pen/qOrVgO?editors=100
Was not an issue in 0.11.0, but issue in 0.11.2 (easy to fix, but still should be noted in changelog, to save time).

### LINKED ISSUE #5014: layout-margin / layout-padding and flex="?"
In release 0.10.x layout-margin and layout-padding works as expected, but in 0.11.2 it does not.

The issue seems to be that the new class based approach does not add "flex" to the class list and is thus not hit by the following:
```css
.layout-margin, .layout-margin > .flex, .layout-margin > .flex-gt-sm, .layout-margin > .flex-md, .layout-margin > .flex-lt-lg {
  margin: 8px; }
```

In 0.10.1:
```html
<div layout="row" layout-margin>
```
becomes
```html
<div layout="row" class="layout layout-row layout-margin">
```
whereas in 0.11.2 it becomes:
```html
<div class="md-layout-row layout-margin">
```

I tried to setup a codepen to demonstrate the issue, but for some reason I cannot get it to work at all with 0.11.2
[0.10.1](http://codepen.io/tinusn/pen/rOyEOK)
[0.11.2](http://codepen.io/tinusn/pen/ZbedQx) (not, this one does not work at all... dunno why)

Please note, that if you just use "flex" instead of "flex="?"" then it works.

### LINKED ISSUE #5076: Problem with flex's value from variable
Hi all,

When setting the flex value from a variable:

```
<div flex="{{ pane.flex }}"></div>
```

The resulting HTML is (assuming pane.flex=50):

```
<div flex="50" class="flex flex-50"></div>
```

As you can see the "flex" class is the problem. Things get even more strange if there is an ng-repeat:

```
<div ng-repeat="a in [1]"> 
  <div flex="{{ flexValue }}">TEST</div>
</div>
```

the resulting html is

```
<div ng-repeat="a in [1]" class="ng-scope"> 
 <div class="flex-{{-flexValue-}} flex-50" flex="50">TEST</div>
</div>
```

It's really not a big problem but I figured you may want to know. 

Thanks

ThomasBurleson:
fix(layouts): interpolated values, validations, test & CSS fixes

* improved consistent use of `attrs.$observe( )` to watch for interpolated attribute values (for Layout attributes which support value specifiers).
* add validation of Layout attribute values with fallbacks to default value approprate to each attribute type
* significant code cleanup for use of `$observe( )`
* removed possible classname generation with raw, uninterpolated strings
* removed multiple classname generation
* fixed invalid classname generation when interpolation values are invalid
* remove `md-` prefix from layout class names
* remove all attribute selectors from layout.scss
* deprecated use of `offset` attribute; now uses `layout-offset`
* add box-sizing to all flex variants
* add max-width/height to flex variants
* fixed max-width/height for flex 33, 34, 66, and 67 variants.
* complete refactor for spec testing of Layouts

Fixes #5076. Fixes #5054. Refs #5014. Fixes #4994. Fixes #4959. Fixes #4902. Fixes #2954. Fixes #5014.

### DIFF
--- a/src/core/services/layout/layout.js
+++ b/src/core/services/layout/layout.js
@@ -1,319 +1,414 @@
-(function () {
+(function() {
   'use strict';
 
-  var $mdUtil, $$mdLayout, $parse, $interpolate;
+  var $mdUtil, $$mdLayout, $interpolate;
+
+  var SUFFICES = /(-sm|-gt-sm|-md|-gt-md|-lg|-gt-lg)/g;
+  var WHITESPACE = /\s+/g;
+
+  var FLEX_OPTIONS = ['grow', 'initial', 'auto', 'none'];
+  var LAYOUT_OPTIONS = ['row', 'column'];
+  var ALIGNMENT_OPTIONS = [
+        "start start", "start center", "start end",
+        "center", "center center", "center start", "center end",
+        "end", "end center", "end start", "end end",
+        "space-around", "space-around center", "space-around start", "space-around end",
+        "space-between", "space-between center", "space-between start", "space-between end"
+      ];
+
+
+  var config = {
+    /**
+     * Enable directive attribute-to-class conversions
+     */
+    enabled: true,
 
     /**
+     * List of mediaQuery breakpoints and associated suffixes
      *
-     *   The original ngMaterial Layout solution used attribute selectors and CSS.
-     *
-     *  ```html
-     *  <div layout="column"> My Content </div>
-     *  ```
-     *
-     *  ```css
-     *  [layout] {
+     *   [
+     *    { suffix: "sm", mediaQuery: "screen and (max-width: 599px)" },
+     *    { suffix: "md", mediaQuery: "screen and (min-width: 600px) and (max-width: 959px)" }
+     *   ]
+     */
+    breakpoints: []
+  };
+
+  /**
+   *   The original ngMaterial Layout solution used attribute selectors and CSS.
+   *
+   *  ```html
+   *  <div layout="column"> My Content </div>
+   *  ```
+   *
+   *  ```css
+   *  [layout] {
      *    box-sizing: border-box;
      *    display:flex;
      *  }
-     *  [layout=column] {
+   *  [layout=column] {
      *    flex-direction : column
      *  }
-     *  ```
-     *
-     *  Use of attribute selectors creates significant performance impacts in some
-     *  browsers... mainly IE.
-     *
-     *  This module registers directives that allow the same layout attributes to be
-     *  interpreted and converted to class selectors. The directive will add equivalent classes to each element that
-     *  contains a Layout directive.
-     *
-     * ```html
-     *   <div layout="column" class="layout layout-column"> My Content </div>
-     *```
-     *
-     *  ```css
-     *  .layout {
+   *  ```
+   *
+   *  Use of attribute selectors creates significant performance impacts in some
+   *  browsers... mainly IE.
+   *
+   *  This module registers directives that allow the same layout attributes to be
+   *  interpreted and converted to class selectors. The directive will add equivalent classes to each element that
+   *  contains a Layout directive.
+   *
+   * ```html
+   *   <div layout="column" class="layout layout-column"> My Content </div>
+   *```
+   *
+   *  ```css
+   *  .layout {
      *    box-sizing: border-box;
      *    display:flex;
      *  }
-     *  .layout-column {
+   *  .layout-column {
      *    flex-direction : column
      *  }
-     *  ```
-     */
-    angular.module('material.core.layout', [ 'ng' ])
+   *  ```
+   */
+  angular.module('material.core.layout', ['ng'])
 
-      /**
-       * Model of flags used by the Layout directives
-       * Allows changes while running tests or runtime app changes
-       */
-      .factory("$$mdLayout", function() {
-        return {
-          removeAttributes : true
-
-        };
-      })
-
-      // Attribute directives with optional value(s)
-
-      .directive('layout'              , attributeWithObserve('layout'      )  )
-      .directive('layoutSm'            , attributeWithObserve('layout-sm'   )  )
-      .directive('layoutGtSm'          , attributeWithObserve('layout-gt-sm')  )
-      .directive('layoutMd'            , attributeWithObserve('layout-md'   )  )
-      .directive('layoutGtMd'          , attributeWithObserve('layout-gt-md')  )
-      .directive('layoutLg'            , attributeWithObserve('layout-lg'   )  )
-      .directive('layoutGtLg'          , attributeWithObserve('layout-gt-lg')  )
-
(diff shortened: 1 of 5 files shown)


ThomasBurleson:
@jelbourn, @robertmesserle, @topherfangio, @rschmukler - here is the new PR. Which has improvements/changes based on reviews and comments from [older, deprecated] PR.

src/core/services/layout/layout.js
```diff
@@ -1,319 +1,414 @@
-(function () {
+(function() {
   'use strict';
 
-  var $mdUtil, $$mdLayout, $parse, $interpolate;
+  var $mdUtil, $$mdLayout, $interpolate;
 
+  var SUFFICES = /(-sm|-gt-sm|-md|-gt-md|-lg|-gt-lg)/g;
```
jelbourn: Suffixes (spelling)

jelbourn: Could do `/((-gt)?(-sm|-md|-lg))/g`

src/core/services/layout/layout.js
```diff
@@ -1,319 +1,414 @@
...
+
+  var FLEX_OPTIONS = ['grow', 'initial', 'auto', 'none'];
+  var LAYOUT_OPTIONS = ['row', 'column'];
+  var ALIGNMENT_OPTIONS = [
+        "start start", "start center", "start end",
+        "center", "center center", "center start", "center end",
+        "end", "end center", "end start", "end end",
+        "space-around", "space-around center", "space-around start", "space-around end",
+        "space-between", "space-between center", "space-between start", "space-between end"
+      ];
```
jelbourn: Indent should just be +2 for items and +0 for end `]`

src/core/services/layout/layout.js
robertmesserle: While we're at it: `/(-gt)?-(sm|md|lg)/g`

src/core/services/layout/layout.js
```diff
@@ -1,319 +1,414 @@
...
-     */
-    function attributeWithoutValue(className) {
-      return ['$$mdLayout', '$document', function(_$$mdLayout_, $document) {
-        $$mdLayout = _$$mdLayout_;
-        return {
-          restrict : 'A',
-          compile: function(element, attrs) {
+        case 'flex'          :
+          if (!findIn(value, FLEX_OPTIONS)) {
+            if (isNaN(+value)) {
```
jelbourn: Are you using unary plus to cast to a number? I generally prefer doing `Number(value)` since it's much more explicit

src/core/services/layout/layout.js
robertmesserle: Jeremy edited his.  I swear I didn't post the same suggestion as Jeremy!

ThomasBurleson: LOL

src/core/services/layout/layout.js
ThomasBurleson: Instead of formatting feedback, I was hoping more for logic and refactoring feedback and manual testing.... ;-)

src/core/services/layout/layout.js
```diff
@@ -1,319 +1,414 @@
...
-    function attributeWithObserve(className) {
-
-      return ['$mdUtil', '$$mdLayout', '$document', '$parse', '$interpolate', function(_$mdUtil_, _$$mdLayout_, $document, _$parse_, _$interpolate_) {
-        $mdUtil = _$mdUtil_;
-        $$mdLayout = _$$mdLayout_;
-        $parse = _$parse_;
-        $interpolate = _$interpolate_;
+    function translateWithValueToCssClass(scope, element, attrs) {
+      var updateFn = updateClassWithValue(element, className, attrs);
+      var unwatch = attrs.$observe(attrs.$normalize(className), updateFn);
```
jelbourn: IS this $observe gated on the expression having interpolation symbols in it?

src/core/services/layout/layout.spec.js
```diff
@@ -132,13 +95,235 @@ describe('layout directives', function() {
...
+    testAllSufficesWithValues("flex", allowedValues );
+
+  });
+  describe('using [flex-order] attributes', function() {
+    var flexOrderValues = [
+      -9, -8, -7, -6, -5, -4, -3, -2, -1,
+      0, 1, 2, 3, 4, 5, 6, 7, 8, 9
+    ];
 
+    it('should support attribute without value "<div flex-order>"', inject(function($compile, $rootScope) {
```
jelbourn: Since you're adding a bunch of new tests I'll mention this:
You can do 
```js
beforeEach(inject(function($injector) {
  $compile = $injector.get('$compile');
  $rootScope = $injector.get('$rootScope);
});
```

and avoid the boilerplate of having the same `inject` on every test.

jelbourn: Could also use this to coalece calls to `$rootScope.$new()` in one place by doing something like
```js
var $rootScope = $injector.get('$rootScope');
pageScope = $rootScope.$new();
// Then use pageScope throughout the tests
```

src/core/services/layout/layout.js
ThomasBurleson: My understanding is that $observe will NOT establish a watcher if the value does not have interpolations symbols.... that there is no performance hit to use it this way.

PR #5096, opened 2015-10-11
fix(tabs): tab paging and sizing on IE11 when tabs are in a dialog

### LINKED ISSUE #3953: md-tabs inside of md-dialog on IE11 throws exceptions
When use md-tab inner md-dialog and run on IE 11, I receive a bug when re-size window.
- step one: https://material.angularjs.org/latest/#/demo/material.components.dialog open in code pen
- change md-dialog-content with tab https://material.angularjs.org/latest/#/demo/material.components.tabs

Splaktar:
`elements.dummies` is not always valid on IE11. It often throws `Invalid calling object`.
The fix is to call `getElements()` to do a fresh look up of the selectors each time.
Only doing the `getElements()` call when the exception is throw was tried.
It resulted in improperly loading/sizing of tabs on small screen sizes.

Fixes #3953.

### DIFF
--- a/src/components/dialog/demoBasicUsage/index.html
+++ b/src/components/dialog/demoBasicUsage/index.html
@@ -4,15 +4,18 @@
     send focus back to the triggering button.
   </p>
 
-  <div class="dialog-demo-content" layout="row" layout-wrap >
-    <md-button class="md-primary md-raised" ng-click="showAlert($event)" flex flex-md="100">
+  <div class="dialog-demo-content" layout="row" layout-wrap layout-margin>
+    <md-button class="md-primary md-raised" ng-click="showAlert($event)" flex-sm="100" flex-md="100" flex-gt-md="auto">
       Alert Dialog
     </md-button>
-    <md-button class="md-primary md-raised" ng-click="showConfirm($event)" flex flex-md="100">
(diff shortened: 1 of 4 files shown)


Splaktar:
I used the demo with a tab in the dialog to debug this since I was unable to get `md-tabs` to work inside of an `md-dialog` in the Karma tests. Let me know if you'd rather not have this demo included in the PR and I can remove it.

If you have some pointers for getting karma/jasmine/injectors/mocks/modules setup properly for creating a unit test that includes both tabs and dialogs, please let me know.

Splaktar:
@robertmesserle can you please take a look at this and let me know if you see any issues with the changes to `tabsController`?

ThomasBurleson:
@robertmesserle - can you review for 0.11.3 inclusion?

ThomasBurleson:
@robertmesserle  - ok to merge this ?

robertmesserle:
@ThomasBurleson We discussed this issue the other day, but never had a followup I don't think.  I haven't had time to do so yet, but I wanted to look into ways to optimize this so that it doesn't need to do that lookup on every reference to dummies; however, I think this is fine to merge as it should be a pretty minimal performance hit.

PR #5102, opened 2015-10-12
fix(chips): remove right padding on disabled chips

mjhasbach:
An existing style for preventing right padding on disabled chips uses the wrong selector.


mjhasbach:
There's also [padding-right: rem(0.4) on .md-chip-content](https://github.com/angular/material/blob/53f25889a251620e3663037ea7c06f1043d9af68/src/components/chips/chips.scss#L85). Why?

mjhasbach:
This issue has been resolved by other commits.

PR #5108, opened 2015-10-12
fix(alert): restore scroll to proper place

yarl:
It looks like `scrollTop` is not cross-browser consistent. See https://bugs.webkit.org/show_bug.cgi?id=106133 for more information.

Fix tested on Windows 7 IE10.


EladBezalel:
Please check it against HEAD,
There's no place that this is documented?
or used? please explain... closing for now, if you think it's a valid PR i will reopen

PR #5157, opened 2015-10-14
Fixed issue with dialog double opening and inability to close second dialog.

erikogenvik:
There seems to be a bug which prevents a dialog from being closed
if it's opened while another dialog is shown.

This adds a test for this scenario (which currently fails), but doesn't fix the issue.

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -27,7 +27,7 @@ function MdDialogDirective($$rAF, $mdTheming, $mdDialog) {
         }
 
         scope.$on('$destroy', function() {
-          $mdDialog.destroy();
+          $mdDialog.destroyInstance(element[0].parentNode);
         });
 
         /**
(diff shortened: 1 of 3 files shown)


erikogenvik:
I signed it!

erikogenvik:
Updated with a fix.
The error was caused by commit 77a34bd8f122814527159d08536d5776b314f527
This fixed an issue with interimElements (of which dialog is one sort) not being destroyed when the scope was destroyed from outside causes (such as $location change).
However, it also introduced this issue, since the deferred nature of calls to .destroy() made it possible for one dialog to get destroyed after another had been created. With the result that the first dialog effectively removed the second dialog from the stack, making any subsequent calls to $mdDialog.hide() not work.

This fixed the issue for dialogs.
It's not totally clear to me whether other interimElements such as menus or select element suffer from the same issues.

PR #5160, opened 2015-10-14
fix(autocomplete): force $scope.searchText to string

justmike:
Discovered when I had an autocomplete that was hidden until a previous autocomplete had data.
Failing example: http://codepen.io/justmike/pen/wKrzdg

This only started happening when I upgraded from 0.11.1 to 0.11.4.


justmike:
I signed it!

justmike:
Fixed by https://github.com/angular/material/commit/fbf45fdf9e07c161c04db02f890cd132c87ed46d

PR #5180, opened 2015-10-15
Variables are not included in the standalone flexgrid

MartinMuzatko:
These Variables are missing:

```sass
$layout-gutter-width
$layout-breakpoint-sm
$layout-breakpoint-md
$layout-breakpoint-lg
```

### DIFF
--- a/src/core/services/layout/standalone.scss
+++ b/src/core/services/layout/standalone.scss
@@ -15,6 +15,11 @@
 *
 */
 
+$layout-gutter-width: 1em;
+$layout-breakpoint-sm: 30em;
+$layout-breakpoint-md: 50em;
+$layout-breakpoint-lg: 80em;
+
 @-moz-document url-prefix() {
   [layout-fill] {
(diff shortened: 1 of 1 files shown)


MartinMuzatko:
Btw: I signed the Google CLA

ThomasBurleson:
@MartinMuzatko - excellent catch. We will fix in master; but NOT with your `..em;` changes.

MartinMuzatko:
Feel free to change them to px. I prefer em as they serve a more comprehensible format for me. Of course, only if the root size is 16px.  10em steps match almost every breakpoint of modern mobile and desktop devices. Which is why I prefer ems.

em - px
10	160
20	320
30	480
40	640
50	800
60	960
70	1120
80	1280
90	1440
100	1600
110	1760
120	1920

It makes me able to think more in terms of "percentage"  rather than a fixed number.
This makes the relation of breakpoints more clear to me than px can do.

PR #5202, opened 2015-10-16
feat(core): add ngMaterial global with version info.

jelbourn:
@ThomasBurleson @robertmesserle I just added the `full` version for now since I don't think we need to bother independently listing `major`, `minor`, and `dot` at this point. And we don't have release code-names like angular does.

Adding this as part of the build process for now so that it gets the version directly from package.json.


ThomasBurleson:
@jelbourn - this may have some issues when building locally and testing. I think ngMaterial will be undefined in local builds.

PR #5214, opened 2015-10-16
WIP: Gesture Fix

rschmukler:
This PR greatly extends the click-hijacking functionality of
`$mdGestures`.

**Background:** iOS emulates click events on a `350ms` delay in order to
avoid sending clicks when users pan, perform gestures, etc. This leads
to pages often feeling laggy in mobile browsers. This appears to be less
of an issue in Safari, but is a major issue in embedded webview
applications such as Chrome on iOS and PhoneGap built apps.

To get around this `$mdGesture` currently listens to `touchstart` and `touchend`
events to manually dispatch `click` events. This allows it to
trigger events much quicker than the 350ms.  It then configures the
page to ignore clicks that were not originated by `$mdGesture`.

**Problem:** While `$mdGesture` did successfully hijack the `click`
event, it did not hijack other emulated events from iOS, including `mouseup`,
`mousedown` and the subsequent `focus` event that could originate from a
delayed `click`.

**Example of problem:** Issue #4850 was caused by the `md-select`
opening quickly via our simulated `click` event, followed by receiving a
`350ms` OS delayed `mousedown` event resulting in an option immediately
being selected.

Some other example problems that impact the framework:

- Ink ripples are 350ms delayed (use `mousedown` and `mouseup`)
- Some components / items randomly ripple or trigger
after being focused / click.  Usually this isn't an issue, but can be
(eg. #4850)
- `focus` events are 350ms delayed (Affects input)

**Soltuion:** This PR makes `$mdGesture` perform hijacking of
`mousedown`, `mouseup` and subsequent `focus` events in order to more
quickly dispatch them.

**Notes:** A few of our internal components needed some
refactoring as they were properly discarding the emulated `mousedown`
(which is now ignored DOM wide) and relying on those events for state,
etc.

Ripples will need a fix as they now properly ripple on `mousedown` but
can hit a situation where they receive no later `mouseup` event. Perhaps
listening to the document-wide `mouseup` event should be registered on
`mousedown`. An example is scrolling in the Docs' Sidenav via touch.

closes #4850, closes #4757

Issue #5195, opened 2015-10-15
md-dialog not closing anymore


kschwidder:
With the latest HEAD the md-dialog does not close anymore.
it works in 0.11.3

md-background is hidden but dialog still stays.

ThomasBurleson:
I am unable to reproduce this in our docs; please provide more details.

kschwidder:
@ThomasBurleson: I have nailed it down more precisely:

1) animation is disabled.
2) dialog opacity is set to 0 and the dialog is not visible anymore (as reported once by myself)

In our app we are using menus which are triggering dialogs via directives as follows:

a)  <md-menu access-control resource="Project" operation="UPDATE,DELETE" owner="{{oid}}">
                            <div aria-label="Menu"
                                 class="md-icon-button"
                                 ng-click="$mdOpenMenu()">
                                <img class="pointer"
                                     ng-src="template/icons/view/ic_arrow_drop_down_grey600_24dp.png">
                            </div>
                            <md-menu-content>
                                <md-menu-item access-control resource="Project" operation="UPDATE">
                                    <md-button resource="Project" operation="UPDATE" project-member-workflow="member"
                                               aria-label="User prozess" ng-click="" member="user">
                                        <img src="template/icons/view/ic_create_grey600_24dp.png">
                                        {{'STANDARD_BTN_PROCESS'| translate}}
                                    </md-button>
                                </md-menu-item>
                                <md-divider access-control resource="Project" operation="DELETE"></md-divider>
                                <md-menu-item access-control resource="Project" operation="DELETE">
                                    <md-button resource="Project" operation="DELETE" delete-match-member="member" ng-click="" aria-label="undo user"
                                               member="user">
                                        <img src="template/icons/view/ic_delete_grey600_24dp.png">{{'STANDARD_BTN_DELETE'
                                        | translate}}
                                    </md-button>
                                </md-menu-item>
                            </md-menu-content>
                        </md-menu>

b) the directive triggers the dialog as follows:

                           $mdDialog.show({
                                controller: myController,
                                templateUrl: 'app/module/mod_project/views/dialog.project.customer.edit.tpl.html',
                                targetEvent: event,
                                clickOutsideToClose: false,
                                escapeToClose: false,
                                bindToController: true,
                                locals: {object: localCopy}
                            }).then(onUpdate, onAbort);

Problem is the following:

1) When disabling animation the dialog is not shown since opacity is 0
2) When animation is enabled dialog it is shown but weired behaviour...

[...]

PR #5234, opened 2015-10-18
Fix issues with gestures PR (#5214).

topherfangio:
This PR provides fixes to Ryan's gestures PR (#5214) for the fab speed dial.

* Fix the click/touch/keyboard focus support for FAB to work with new gestures system.

* Temporarily fix dialog issue by making sure the FAB actions only scale to `0.1` instead of `0` as the new animation for the dialog breaks if the scale is 0 - tracked in https://github.com/angular/material/issues/5195#issuecomment-148974822 (this works because the opacity is still 0 so it's still invisible to the user).

*  Temporarily `xit` a failing tooltip unit test as I believe it is functioning properly, I think test just needs to be updated, but I'm not 100% sure why the gesture changes are breaking it.

### DIFF
--- a/docs/app/js/preload.js
+++ b/docs/app/js/preload.js
@@ -1,0 +1,2 @@
+// Used only to optionally test
+// jQuery w/ docs (see docs/gulpfile#docs-js)
--- a/docs/gulpfile.js
+++ b/docs/gulpfile.js
@@ -15,6 +15,7 @@ var utils = require('../scripts/gulp-utils.js');
 var karma = require('karma').server;
 var argv = require('minimist')(process.argv.slice(2));
 var gutil = require('gulp-util');
+var series = require('stream-series');
 
 var config = {
   demoFolder: 'demo-partials'
@@ -116,12 +117,22 @@ gulp.task('docs-js-dependencies', ['build'], function() {
 });
 
 gulp.task('docs-js', ['docs-app', 'docs-html2js', 'demos', 'build', 'docs-js-dependencies'], function() {
-  return gulp.src([
-    'node_modules/angularytics/dist/angularytics.js',
-    'dist/docs/js/**/*.js'
-  ])
-  .pipe(concat('docs.js'))
-  .pipe(gulpif(!argv.dev, uglify()))
+  var preLoadJs = ['docs/app/js/preload.js'];
+  if (process.argv.indexOf('--jquery') != -1) {
+    preLoadJs.push('node_modules/jquery/dist/jquery.js');
+  }
+
+  return series(
+    gulp.src([
+      'node_modules/angularytics/dist/angularytics.js',
+      'dist/docs/js/**/*.js'
+    ])
+      .pipe(concat('docs.js'))
+      .pipe(gulpif(!argv.dev, uglify())),
+    gulp.src(preLoadJs)
+      .pipe(concat('preload.js'))
+      .pipe(gulpif(!argv.dev, uglify()))
+  )
   .pipe(gulp.dest('dist/docs'));
 });
 
--- a/src/components/fabSpeedDial/fabController.js
+++ b/src/components/fabSpeedDial/fabController.js
@@ -46,7 +46,7 @@
 
     function setupListeners() {
       var eventTypes = [
-        'mousedown', 'mouseup', 'click', 'touchstart', 'touchend', 'focusin', 'focusout'
+        'click', 'focusin', 'focusout'
       ];
 
       // Add our listeners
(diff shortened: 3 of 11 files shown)


topherfangio:
@ThomasBurleson @rschmukler @jelbourn @robertmesserle Fixes ready for review.

docs/app/js/preload.js
```diff
@@ -0,0 +1,2 @@
+// Used only to optionally test
+// jQuery w/ docs (see docs/gulpfile#docs-js)
```
robertmesserle: Would this fit on a single line?  The line break feels odd.

docs/gulpfile.js
```diff
@@ -116,12 +117,22 @@ gulp.task('docs-js-dependencies', ['build'], function() {
...
+    gulp.src([
+      'node_modules/angularytics/dist/angularytics.js',
+      'dist/docs/js/**/*.js'
+    ])
+      .pipe(concat('docs.js'))
+      .pipe(gulpif(!argv.dev, uglify())),
+    gulp.src(preLoadJs)
+      .pipe(concat('preload.js'))
+      .pipe(gulpif(!argv.dev, uglify()))
+  )
```
robertmesserle: Is this going to include `preload.js` for all docs builds?  Don't we not want that in production?

src/components/fabSpeedDial/fabController.js
```diff
@@ -42,11 +42,13 @@
       resetActionIndex();
     }
 
-    var events = [];
-
     function setupListeners() {
       var eventTypes = [
-        'mousedown', 'mouseup', 'click', 'touchstart', 'touchend', 'focusin', 'focusout'
+        '$md.pressdown',
+
```
robertmesserle: Can we remove these empty lines between events?

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -199,7 +199,7 @@
           offsetDelay = index * delay;
 
         styles.opacity = ctrl.isOpen ? 1 : 0;
-        styles.transform = styles.webkitTransform = ctrl.isOpen ? 'scale(1)' : 'scale(0)';
+        styles.transform = styles.webkitTransform = ctrl.isOpen ? 'scale(1)' : 'scale(0.1)';
```
robertmesserle: Just curious - why `0.1` instead of `0`.  Maybe include a comment explaining this.

src/components/fabSpeedDial/fabSpeedDial.scss
```diff
@@ -114,7 +114,7 @@ md-fab-speed-dial {
   &.md-scale {
     .md-fab-action-item {
       opacity: 0;
-      transform: scale(0);
+      transform: scale(0.1);
```
robertmesserle: Same question as above.

src/components/tooltip/tooltip.spec.js
```diff
@@ -141,7 +141,7 @@ describe('<md-tooltip> directive', function() {
       expect($rootScope.testModel.isVisible).toBe(false);
     });
 
-    it('should set visible on touchstart and touchend', function() {
+    xit('should set visible on touchstart and touchend', function() {
```
robertmesserle: Curious why this test was removed.

src/components/tooltip/tooltip.js
```diff
@@ -168,13 +168,13 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
...
-        parent.on('blur mouseleave touchend touchcancel', leaveHandler );
+        parent.on('blur mouseleave touchcancel', leaveHandler );
         setVisible(true);
       };
-      var leaveHandler = function () {
+      var leaveHandler = function (e) {
         var autohide = scope.hasOwnProperty('autohide') ? scope.autohide : attr.hasOwnProperty('mdAutohide');
         if (autohide || mouseActive || ($document[0].activeElement !== parent[0]) ) {
-          parent.off('blur mouseleave touchend touchcancel', leaveHandler );
+          parent.off('blur mouseleave touchcancel', leaveHandler );
```
robertmesserle: Why was `touchend` removed?

robertmesserle:
Added some feedback.  Mostly LGTM, just had some questions about a few parts.

docs/gulpfile.js
ThomasBurleson: **preload.js** will be empty for production and normal dev builds; only builds with **--jquery**  will preload contain valid jquery code.

src/components/fabSpeedDial/fabSpeedDial.js
ThomasBurleson: @topherfangio - should this be `0.01` to be completely visually transparent but still support mouse clicks ?

src/components/tooltip/tooltip.spec.js
jelbourn: Yeah- if this needs to be omitted, it needs a comment explanation for why.

docs/app/js/preload.js
topherfangio: Not sure; @rschmukler can you respond?

src/components/fabSpeedDial/fabController.js
topherfangio: Will do.

src/components/fabSpeedDial/fabSpeedDial.js
topherfangio: It doesn't really matter as we also set opacity to 0 and set CSS `pointer-events: none`. As I noted in the description, this should be changed back when #5195 is fixed. I will add a TODO about it.

src/components/fabSpeedDial/fabSpeedDial.scss
topherfangio: Same response as above :wink:

src/components/tooltip/tooltip.spec.js
topherfangio: I'll add a comment. Would love to get @rschmukler to take a look too though.

src/components/tooltip/tooltip.js
topherfangio: Not 100% sure; @rschmukler can you respond?

I think it was unnecessary and causing issues after the new gesture support, but this is exactly why the unit test above is failing.

PR #5259, opened 2015-10-19
cssToNgConstant should generate valid json literal using double quote

loganfreeman:
when I run 'gulp build-js', the gulp-uglify throws an error due to the fact that cssToNgConstant contains invalid string literal.


loganfreeman:
I have signed the Google Individual CLA!

Soulcyber:
a PR for the same problem was added here:
https://github.com/angular/material/pull/5045

Issue #5279, opened 2015-10-20
md-chips: Add support for comma separated entry


Ignigena:
I have a use case for the Chips component in which a user is entering a list of tags in the field. For most of my users it is not very intuitive to hit the enter key after each input. Rather, they enter a comma and expect the field to update accordingly.

While I can create custom handling for this in my controller, it seems to me this is a rather common use case which could use an opt-in setting on the directive. I'd be more than happy to submit a PR for this feature if it's something that makes sense in the framework.

philippeluickx:
+1

topherfangio:
Just to confirm, you basically want comma to work identically to the enter key correct?

Ignigena:
Correct, so entering a comma would append the buffer to the list and create a new chip. @Frank3K suggested in my PR that semi-colons might also be a good use case for contact chips when typing an e-mail address. I have since modified the original PR to include a more generic approach (although understandable if this approach is not ideal.)

satpaltanan:
The issue is closed for the desktop browser, but in case of mobile devices it is not working

topherfangio:
@satpaltanan If you're having an issue on mobile, can you please open a new issue with a relevant Codepen and ping me on it? Thanks!

satpaltanan:
Please check the codepen link
http://codepen.io/bsrinivasguptha/pen/YwamLm/
It has custom separator keys as Enter and Comma, but only Enter is working in the mobile device, Comma is not working.

satpaltanan:
HI @topherfangio 

Any update on this issue.

topherfangio:
@satpaltanan This seems to work ok on my iPhone. What mobile device are you using? Also, can you verify your CodePen email so that we can look at the example in fullscreen so that we can test on mobile a bit easier?

![screen shot 2017-04-07 at 4 28 54 pm](https://cloud.githubusercontent.com/assets/54370/24820560/52f4dff2-1baf-11e7-9ab5-e93db6aa8758.png)

rcasurion:
Do we have option to disable this feature? Because in my case, we have options that includes ',' then the problem is we ca not write ',' anymore.

PR #5281, opened 2015-10-20
feat(chips): add support for custom separator keys

Ignigena:
Implements #5279 

Add the optional ability for chips to be created on keydown of any key
in the `md-separator-keys` attribute. Custom key codes are supported
in addition to common ones defined in `$mdConstant.KEY_CODE`.

### DIFF
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -344,6 +344,7 @@ describe('<md-chips>', function() {
 
           expect(enterEvent.preventDefault).toHaveBeenCalled();
         }));
+
       });
 
       it('focuses/blurs the component when focusing/blurring the input', inject(function() {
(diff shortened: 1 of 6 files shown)


Ignigena:
I signed it!

Frank3K:
There's also a valid use case for a semicolon-separator, e.g. when entering multiple email-addresses. I think it would be more flexible if one could specify a list of separators instead of having an attribute specifically for the comma-separated option.

What do you think?

Ignigena:
@Frank3K Good point. And I'm definitely all for a more generic option to support for other use cases. Let me see what I can do...

topherfangio:
This looks really interesting. Can you add/edit a demo to show how this would work?

Ignigena:
Here's a quick and dirty demo: http://codepen.io/anon/pen/rOKexW

Top demo is comma separated and bottom demo adds the ability to separate on semi-colons by injecting and modifying `$mdConstant.KEY_CODES`

topherfangio:
@Ignigena That demo looks great, but I was hoping you could incorporate it into an existing demo (or create a new one) so that end-users would have something to look at as an example :smile:

Ignigena:
Ahhh I see now. Of course! Sorry, it's been a long day :smile:

Ignigena:
Basic demo for both comma and custom separators pushed. I incorporated it into a new demo so as not to complicate the existing "basic usage" one further.

Frank3K:
Is it good practice to add keys to the $mdConstant object by changing the object after creation? Is does not completely feel right to me.

topherfangio:
@Ignigena - @Frank3K makes a good point. Rather than using strings for the keys, I think it would be better if you passed the actual key codes so that you could pass constants defined in `$mdConstant` or some service provided by your own application.

Forcing users to inject the key codes into `$mdConstant` seems like it could cause trouble later on.

Ignigena:
@topherfangio Makes complete sense. Refactoring now.

Ignigena:
Refactored and updated demo and spec. I left the addition of the `COMMA` definition on the constant since it seems there would be a broad enough use case for this, but let me know if this should be removed in favor of a strictly generic approach.

ThomasBurleson:
@topherfangio - is the ready for merging ?

topherfangio:
@ThomasBurleson LGTM.

DevVersion:
I love that feature. Great job @Ignigena

topherfangio:
@Ignigena Can you rebase against the latest master branch and squash your commits into one so that @ThomasBurleson can merge when he is ready?

Thanks!

Ignigena:
@topherfangio Done and done!

topherfangio:
Awesome, thanks so much!

ThomasBurleson:
@Ignigena - thank you for your PR and your responsiveness to our change requests.

PR #5303, opened 2015-10-21
fix(dialog, menu, animate): added origin element fallback to bounds

### LINKED ISSUE #5195: md-dialog not closing anymore

EladBezalel:
The dialog was opened from the menu item, when the dialog tries to close to the element that opened it, the element is long gone.

Added fallback of the origin clientRect to the Rect of the bounds.

fixes #5195


ThomasBurleson:
Fixed.

PR #5350, opened 2015-10-25
theming.js - fix missing bracket

BoLaMN:
Add missing bracket

### DIFF
--- a/src/core/services/theming/theming.js
+++ b/src/core/services/theming/theming.js
@@ -75,7 +75,7 @@ var DARK_SHADOW = '1px 1px 0px rgba(0,0,0,0.4), -1px -1px 0px rgba(0,0,0,0.4)';
 var LIGHT_SHADOW = '';
 
 var DARK_CONTRAST_COLOR = colorToRgbaArray('rgba(0,0,0,0.87)');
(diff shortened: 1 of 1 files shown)


BoLaMN:
I signed it!

src/core/services/theming/theming.js
```diff
@@ -75,7 +75,7 @@ var DARK_SHADOW = '1px 1px 0px rgba(0,0,0,0.4), -1px -1px 0px rgba(0,0,0,0.4)';
 var LIGHT_SHADOW = '';
 
 var DARK_CONTRAST_COLOR = colorToRgbaArray('rgba(0,0,0,0.87)');
```
BoLaMN: is there any reason why these arent just defined as arrays but instead we convert it each run?

eg.
```javascript
var DARK_CONTRAST_COLOR = [ 0, 0, 0, 0.87 ]
var LIGHT_CONTRAST_COLOR = [ 255, 255, 255, 0.87 ]
var STRONG_LIGHT_CONTRAST_COLOR = [ 255, 255, 255 ]
```

ThomasBurleson: More than likely `rgba(...)` was perceived as more clear.

PR #5366, opened 2015-10-26
Fix position of toast element

idoo:
If page has too long  content — user can miss toast element. Just fixed it. :facepunch:


EladBezalel:
Fixed is not always the solution, if i'm not wrong a pr about it was merged not long ago, please check your issue against HEAD and if it's still happening i will reopen

PR #5373, opened 2015-10-26
fix(mdTooltip): observing direction changes

### LINKED ISSUE #5362: toolTip: does not support dynamic md-direction with data binding
When you use `md-tooltip` and `md-direction` with data binding -- as in the following code snippet --, `md-tooltip` continuously uses the initial value of `md-direction`:

```javascript
<md-radio-group ng-model="direction"
  ng-init="direction='right'">
  <md-radio-button value="left">Left</md-radio-button>
  <md-radio-button value="top">Top</md-radio-button>
  <md-radio-button value="right">Right</md-radio-button>
  <md-radio-button value="bottom">Bottom</md-radio-button>
</md-radio-group>
<div layout="row">
  <md-button class="md-primary md-raised">
    Point me
    <md-tooltip
        md-direction="{{direction}}">
      Nice tooltip
    </md-tooltip>
  </md-button>
</div>
```
When you hover the mouse over the button, the tooltip is displayed always to the right, independently of the `direction` value.

See this CodePen: http://codepen.io/Dotneteer/pen/epMjpQ

EladBezalel:
Observing tooltip direction changes and using positionTooltip for live tooltip positioning if the tooltip is visible

fixes #5362

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -63,6 +63,15 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
         tooltipParent = angular.element(current || document.body),
         debouncedOnResize = $$rAF.throttle(function () { if (scope.visible) positionTooltip(); });
 
+    // Observing for tooltip direction changes
+    attr.$observe('mdDirection', function (val) {
+      direction = val;
+
+      if (scope.visible) {
+        positionTooltip();
+      }
+    });
+
     // Initialize element
 
     setDefaults();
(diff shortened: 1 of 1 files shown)


src/components/tooltip/tooltip.js
```diff
@@ -63,6 +63,15 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
         tooltipParent = angular.element(current || document.body),
         debouncedOnResize = $$rAF.throttle(function () { if (scope.visible) positionTooltip(); });
 
+    // Observing for tooltip direction changes
+    attr.$observe('mdDirection', function (val) {
```
ThomasBurleson: For consistency, I think we should `$watch( )` md-direction just like we watch `md-visible`.

src/components/tooltip/tooltip.js
```diff
@@ -86,6 +93,10 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
       }
     }
 
+    function setTransformOrigin (direction) {
```
ThomasBurleson: What is the intent of this function? 
Why is it not just contained (as a closure) in `configureWatchers( ... ) `?

PR #5379, opened 2015-10-27
add non-zero priority to mdButton

amitport:
Since mdButton replaces the element it should have a non-zero priority so that other directives could be reliably linked after this element is replaced.

Specifically, this issue was encoutered when trying to add [ngclipboard](https://github.com/sachinchoolur/ngclipboard/blob/master/src/ngclipboard.js) directive.

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -67,6 +67,7 @@ function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
     restrict: 'EA',
     replace: true,
     transclude: true,
+    priority: 100,
     template: getTemplate,
     link: postLink
   };


EladBezalel:
@amitport why 100?

amitport:
peeking at the source it seemed to be the default priority for directives that don't have obvious priority requirements (that is, `ThemingDirective` added by @rschmukler and `ngMessageDirective` added @topherfangio)

EladBezalel:
@ThomasBurleson please review

topherfangio:
This seems like a pretty simple change, but I'm honestly a bit worried about the repercussions this may have in larger apps with complex interactions.

@amitport What was the issue you ran into with the ngClipboard directive?

amitport:
ngclipboard used to work unreliably on md-button (used as element or attribute)
(there was a race since md-button changes the element and they relied on a custom attribute added to the element)

In any case, I opened this issue almost half a year ago. The ngcliboard issue may have been resolved with https://github.com/sachinchoolur/ngclipboard/commit/636e1c5a5b1c3d6cd2df930eee103feaecb741b6
(I haven't tested it)

topherfangio:
Sine this is so old, we have released 1.0, and it may have already been fixed in ngClipboard, I am going to go ahead and close this PR.

@amitport If you are still needing this, please let us know and we will reopen.

Issue #4886, opened 2015-09-28
Disabled button cursor doesn't match disabled inputs


joeyperson:
Disabled states should use the default pointer to match other components.

After asking if this special cursor should be added, the word from UX was "I think no cursor change. A visual disabled state to the component is enough of an affordance."

chriscalo:
This is fixed in the [Buttons Demo page](https://material.angularjs.org/latest/demo/button). Seems good to close to me.

topherfangio:
Thanks for the update; I do remember this getting fixed a while back. We'll reopen if it comes up again.

PR #5389, opened 2015-10-27
#4886 set disabled input cursor to 'not-allowed'

adepatie:


EladBezalel:
Disabled buttons changed to be `default` cursor and not `not-allowed`

Issue #4835, opened 2015-09-25
Firefox Subheader/Menu z-index


pulse14:
The subheader is over the menu in firefox 41.0. It works with z-index 1, strange. 
![n2pyjlyg](https://cloud.githubusercontent.com/assets/8348807/10096419/a079cf8c-6371-11e5-85d6-1995e7dd9527.png)

topherfangio:
All of our zIndexes are based off of variables; we should look at subheader/menu to make sure they are using the proper ones.

PR #5390, opened 2015-10-27
#4835 Created subheader zindex variable [closed]

adepatie:


EladBezalel:
Can't see why is this necessary if it was only used once

PR #5396, opened 2015-10-28
fix(speedDial): Don't intercept spaces and fix animations.

### LINKED ISSUE #4750: fabSpeedDial: Close animation does not fire on Safari.
On Safari, the close animation of the speed dial does not fire, so the speed dial's actions immediately disappear with no animation and it is jarring.

### LINKED ISSUE #5065: md-fab-speed-dial close with outside click in firefox not working
If you click on another DOM-Element than md-fab-speed-dial it closes the speed-dial in chrome but not in Firefox. I assume it should close the speed-dial in Firefox, too. This bug is reproducible in your demo.

If you click into an input field with open speed-dial, it prevents typing spaces.
fabSpeedDial/fabController.js line 230:
case $mdConstant.KEY_CODE.SPACE: event.preventDefault(); return false;

### LINKED ISSUE #5085: Can't enter space into input when FAB speed-dial is open by default.
http://codepen.io/katolikau/pen/gaRBLj

If there is an `<md-fab-speed-dial>` with  `md-open="true"` on the page, then you can't enter a space character into any `<input>` on the page. This means that there's some overzealous event listener doing `preventDefault` somewhere on the page.

cc @topherfangio

topherfangio:
* The speed dial watched the document for spacebar presses
  and intercepted them by preventing the default. There was
  really no reason for this, so the code has been removed.
  Additionally, all keypress events are now bound to the
  element itself instead of document.

* Fix the close animations to properly fire on Safari/iOS.

* Fix some issues after recent gesture updates.

Fixes #5085. Fixes #4750. Fixes #5065.

### DIFF
--- a/src/components/fabSpeedDial/fabSpeedDial.js
+++ b/src/components/fabSpeedDial/fabSpeedDial.js
@@ -2,6 +2,13 @@
   'use strict';
 
   /**
+   * The duration of the CSS animation in milliseconds.
+   *
+   * @type {number}
+   */
+  var cssAnimationDuration = 300;
+
+  /**
    * @ngdoc module
    * @name material.components.fabSpeedDial
    */
@@ -103,7 +110,9 @@
     }
   }
 
-  function MdFabSpeedDialFlingAnimation() {
+  function MdFabSpeedDialFlingAnimation($timeout) {
+    function delayDone(done) { $timeout(done, cssAnimationDuration, false); }
+
     function runAnimation(element) {
       var el = element[0];
       var ctrl = element.controller('mdFabSpeedDial');
@@ -169,17 +178,19 @@
       addClass: function(element, className, done) {
         if (element.hasClass('md-fling')) {
           runAnimation(element);
-          done();
         }
+        delayDone(done);
       },
       removeClass: function(element, className, done) {
         runAnimation(element);
-        done();
+        delayDone(done);
       }
     }
   }
 
-  function MdFabSpeedDialScaleAnimation() {
+  function MdFabSpeedDialScaleAnimation($timeout) {
+    function delayDone(done) { $timeout(done, cssAnimationDuration, false); }
+
     var delay = 65;
 
     function runAnimation(element) {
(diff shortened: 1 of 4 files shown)


src/components/fabSpeedDial/fabSpeedDial.scss
```diff
@@ -109,6 +107,15 @@ md-fab-speed-dial {
   }
 
   /*
+   * Hide some graphics glitches if switching animation types
+   */
+  &.md-fling-remove, &.md-scale-remove {
+    .md-fab-action-item > * {
+      visibility: hidden;
```
ThomasBurleson: Why not use `display: none;` ?

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -103,7 +110,9 @@
     }
   }
 
-  function MdFabSpeedDialFlingAnimation() {
+  function MdFabSpeedDialFlingAnimation($timeout) {
+    function delayDone(done) { $timeout(done, cssAnimationDuration, false); }
+
     function runAnimation(element) {
```
ThomasBurleson: Why are you manually setting the style animation properties?
Why not use `$animateCSS( )`

src/components/fabSpeedDial/fabSpeedDial.scss
topherfangio: The speed dial currently uses flex to order/align both the trigger and the actions. When I used `display: none` with the speed dial in direction `up`, the actions height changed and the trigger jumped into the incorrect position :/

I think ideally we'll rewrite this in material2 to use absolute positioning that is calculated in the JS so we don't have to worry about issues like this.

src/components/fabSpeedDial/fabSpeedDial.js
topherfangio: I tried switching to `$animateCSS()` during my last bout fixing bugs, but I seem to remember running into some issues and after about an hour of playing with it I reverted back to the current method.

If you think it's worth the time, I am sure I can update it, but this PR was more concerned with getting it functional again after the gestures update :-) Perhaps we should create a separate issue to track updating this?

ThomasBurleson: No need... I was simply curious. 

It is my goal that the ngMaterial2 implementatioins should require consistent uses of gesture listeners and  uses of `$animateCss( )`, or `element.style({ ... });`.

PR #5399, opened 2015-10-28
perf(theming): add on-demand style generation.

jelbourn:
This change adds a new method to the `$mdTheming` service: `generateTheme(themeName)`. 

This is meant to super large apps that have such a large number of themes that the style generation is a substantial start-up cost. Being able to load themes on-demand as their needed should address this.

For now, I'm purposefully leaving this *out* of the documentation in order to get feedback from the people who are running into this issue in Google.

I tried to make my changes to theming as absolutely minimal as possible as to avoid causing any regressions. 

cc @rschmukler and @ThomasBurleson for review.
@ThomasBurleson don't merge this until I have a chance to run the change by a couple of people with the problem.

### DIFF
--- a/src/core/services/theming/theming.js
+++ b/src/core/services/theming/theming.js
@@ -2,7 +2,7 @@ angular.module('material.core.theming', ['material.core.theming.palette'])
   .directive('mdTheme', ThemingDirective)
   .directive('mdThemable', ThemableDirective)
   .provider('$mdTheming', ThemingProvider)
-  .run(generateThemes);
+  .run(generateAllThemes);
 
 /**
  * @ngdoc service
@@ -122,6 +122,9 @@ var VALID_HUE_VALUES = [
   '700', '800', '900', 'A100', 'A200', 'A400', 'A700'
 ];
 
+// Whether or not themes are to be generated on-demand (vs. eagerly).
+var generateOnDemand = false;
+
 function ThemingProvider($mdColorPalette) {
   PALETTES = { };
   THEMES = { };
@@ -145,6 +148,9 @@ function ThemingProvider($mdColorPalette) {
     },
     alwaysWatchTheme: function(alwaysWatch) {
       alwaysWatchTheme = alwaysWatch;
+    },
+    generateThemesOnDemand: function(onDemand) {
+      generateOnDemand = onDemand;
     },
     $get: ThemingService,
     _LIGHT_DEFAULT_HUES: LIGHT_DEFAULT_HUES,
@@ -349,6 +355,7 @@ function ThemingProvider($mdColorPalette) {
     applyTheme.THEMES = angular.extend({}, THEMES);
     applyTheme.defaultTheme = function() { return defaultTheme; };
     applyTheme.registered = registered;
+    applyTheme.generateTheme = generateTheme;
 
     return applyTheme;
 
@@ -450,10 +457,11 @@ function parseRules(theme, colorType, rules) {
   return generatedRules;
 }
 
-// Generate our themes at run time given the state of THEMES and PALETTES
-function generateThemes($injector) {
+var rulesByType = {};
 
-  var head = document.getElementsByTagName('head')[0];
+// Generate our themes at run time given the state of THEMES and PALETTES
+function generateAllThemes($injector) {
+  var head = document.head;
   var firstChild = head ? head.firstElementChild : null;
   var themeCss = $injector.has('$MD_THEME_CSS') ? $injector.get('$MD_THEME_CSS') : '';
 
@@ -467,7 +475,6 @@ function generateThemes($injector) {
   // components as templates
 
   // Break the CSS into individual rules
-  var rulesByType = {};
   var rules = themeCss
                   .split(/\}(?!(\}|'|"|;))/)
                   .filter(function(rule) { return rule && rule.length; })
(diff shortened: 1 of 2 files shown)


src/core/services/theming/theming.js
```diff
@@ -146,6 +149,9 @@ function ThemingProvider($mdColorPalette) {
     alwaysWatchTheme: function(alwaysWatch) {
       alwaysWatchTheme = alwaysWatch;
     },
+    generateThemesOnDemand: function(onDemand) {
```
ThomasBurleson: perhaps `enableOnDemand` is a better argument name ?

src/core/services/theming/theming.js
```diff
@@ -503,32 +510,15 @@ function generateThemes($injector) {
...
-
-        if (theme.colors.primary.name == theme.colors.accent.name) {
-          console.warn("$mdThemingProvider: Using the same palette for primary and" +
-                       " accent. This violates the material design spec.");
-        }
-
-        GENERATED[theme.name] = true;
-      }
-    });
+  angular.forEach(THEMES, function(theme) {
```
ThomasBurleson: :+1:

src/core/services/theming/theming.spec.js
```diff
@@ -311,6 +311,64 @@ describe('$mdThemingProvider', function() {
...
+    return document.head.querySelectorAll('style[md-theme-style]');
+  }
+
+  function cleanThemeStyleElements() {
+    angular.forEach(getThemeStyleElements(), function(style) {
+      document.head.removeChild(style);
+    });
+  }
+
+  beforeEach(module('material.core', function($provide, $mdThemingProvider) {
```
ThomasBurleson: This is very nice; really understandable. :+1:

src/core/services/theming/theming.js
```diff
@@ -503,32 +510,15 @@ function generateThemes($injector) {
...
-          while (styleStrings.length) {
-            var style = document.createElement('style');
-                style.setAttribute('type', 'text/css');
-            style.appendChild(document.createTextNode(styleStrings.shift()));
-            head.insertBefore(style, firstChild);
-          }
-        });
-
-
-        if (theme.colors.primary.name == theme.colors.accent.name) {
```
ThomasBurleson: What about these console warning messages? 
Did we lose these in the refactor ?

src/core/services/theming/theming.js
```diff
@@ -589,10 +579,41 @@ function generateThemes($injector) {
...
+  if (!GENERATED[name]) {
+    // For each theme, use the color palettes specified for
+    // `primary`, `warn` and `accent` to generate CSS rules.
+    THEME_COLOR_TYPES.forEach(function(colorType) {
+      var styleStrings = parseRules(theme, colorType, rulesByType[colorType]);
+      while (styleStrings.length) {
+        var styleContent = styleStrings.shift();
+        if (styleContent) {
+          var style = document.createElement('style');
+          style.setAttribute('md-theme-style', '');
```
ThomasBurleson: What about the  `style.setAttribute('type', 'text/css');` do we not still need that ?

jelbourn: From [MDN](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/style):
 > **type**
> This attribute defines the styling language as a MIME type (charset should not be specified). This attribute is optional and default to text/css if it's missing.

src/core/services/theming/theming.js
jelbourn: All this code was just moved to a different function without substantive changes.

src/core/services/theming/theming.js
jelbourn: `generateThemesOnDemand` is less ambiguous. `enableOnDemand` doesn't say which action is actually happening on-demand.

src/core/services/theming/theming.js
ThomasBurleson: kk.

src/core/services/theming/theming.spec.js
```diff
@@ -311,6 +311,64 @@ describe('$mdThemingProvider', function() {
 
 });
 
+describe('$mdThemeProvider with on-demand generation', function() {
+  var $mdTheming;
+
+  function getThemeStyleElements() {
+    return document.head.querySelectorAll('style[md-theme-style]');
```
ThomasBurleson: :+1:  to using `md-theme-style` to easily find generated styles.

src/core/services/theming/theming.js
```diff
@@ -589,10 +579,41 @@ function generateThemes($injector) {
...
+  var firstChild = head ? head.firstElementChild : null;
+
+  if (!GENERATED[name]) {
+    // For each theme, use the color palettes specified for
+    // `primary`, `warn` and `accent` to generate CSS rules.
+    THEME_COLOR_TYPES.forEach(function(colorType) {
+      var styleStrings = parseRules(theme, colorType, rulesByType[colorType]);
+      while (styleStrings.length) {
+        var styleContent = styleStrings.shift();
+        if (styleContent) {
```
jelbourn: FYI: I added this `if (styleContent)` check because the code was adding a whole bunch of empty `<style>` elements for some reason. I didn't take the time to go figure out why.

jelbourn:
Currently have patch out to user, waiting to hear back on whether this works for them or not.

PR #5416, opened 2015-10-28
fix(datepicker): properly set ngModel validity. Fixes #4926

### LINKED ISSUE #4926: md-datepicker inside form should throw error if non-valid date format
Hi,

The md-datepicker seems to enable ```ng-messages```which is great. Is it possible to enable ```ngFormController```somehow, that the whole form is getting correct ```$error``` /```$invalid```states? Currently the error message is directive only and doesn't affect the parent ```form```. 

Thanks!

jelbourn:
@mhchen for review

### DIFF
--- a/src/components/datepicker/datePicker.spec.js
+++ b/src/components/datepicker/datePicker.spec.js
@@ -8,11 +8,21 @@ describe('md-date-picker', function() {
   var initialDate = new Date(2015, FEB, 15);
 
   var ngElement, element, scope, pageScope, controller;
-  var $timeout, $$rAF, $animate, $window, keyCodes, dateUtil, dateLocale;
+  var $compile, $timeout, $$rAF, $animate, $window, keyCodes, dateUtil, dateLocale;
+
+  var DATEPICKER_TEMPLATE =
+    '<md-datepicker name="birthday" ' +
+         'md-max-date="maxDate" ' +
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
@mhchen - please label as "needs: merge" when ready.

src/components/datepicker/datePicker.spec.js
```diff
@@ -25,29 +35,28 @@ describe('md-date-picker', function() {
...
-          'ng-required="isRequired" ' +
-          'ng-disabled="isDisabled">' +
-      '</md-datepicker>';
+    createDatepickerInstance(DATEPICKER_TEMPLATE);
+    controller.closeCalendarPane();
+  }));
+
+  /** Compile and link the given template and store values for element, scope, and controller. */
+  function createDatepickerInstance(template) {
     ngElement = $compile(template)(pageScope);
```
mhchen: Optional: Can we instead always set ngElement to the datepicker element and return this element from this function instead if the caller needs to use it? I think it might be a little misleading to have ngElement potentially refer to something other than the datepicker element.

mhchen:
LGTM other than optional comment

src/components/datepicker/datePicker.spec.js
jelbourn: Done.

PR #5419, opened 2015-10-28
refactor(tooltip, util): moved generic functions from tooltip to util

EladBezalel:
* Refactored the tooltip a bit, made the code a lot approachable to read with adding updatePosition function that now is managing the whole tooltip positioning update

* Moved useful generic functions from the tooltip to the util

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -56,11 +56,11 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
 
     $mdTheming(element);
 
-    var parent        = getParentWithPointerEvents(),
+    var parent        = $mdUtil.getParentWithPointerEvents(element),
         content       = angular.element(element[0].getElementsByClassName('md-content')[0]),
-        current       = getNearestContentElement(),
+        current       = $mdUtil.getNearestContentElement(element),
         tooltipParent = angular.element(current || document.body),
-        debouncedOnResize = $$rAF.throttle(function () { if (scope.visible) positionTooltip(); });
+        debouncedOnResize = $$rAF.throttle(function () { updatePosition(); });
 
     // Initialize element
 
@@ -103,7 +103,7 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
         else hideTooltip();
       });
 
-      scope.$watch('direction', positionTooltip );
+      scope.$watch('direction', updatePosition );
     }
 
     function addAriaLabel () {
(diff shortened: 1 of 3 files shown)


src/core/util/util.js
```diff
@@ -633,6 +633,42 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
+        current = current.parentNode;
+      }
+      return current;
+    },
+
+    hasComputedStyleValue: function (key, value, target) {
+      var hasValue = false;
+
+      if ( target && target.length  ) {
+        key    = attr.$normalize(key);
```
ThomasBurleson: Why do this here. Should we not expect the key to be **already** normalized? Or is this an extra safety check ?

src/core/util/util.js
```diff
@@ -633,6 +633,42 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
+    },
+
+    /**
+     * Scan up dom hierarchy for enabled parent;
+     */
+    getParentWithPointerEvents: function (element) {
+      var parent = element.parent();
+
+      // jqLite might return a non-null, but still empty, parent; so check for parent and length
+      while (this.hasComputedStyleValue('pointer-events','none', parent[0] || element[0])) {
```
ThomasBurleson: Do no use `this.` here. The `$mdUtil.getParentWithPointerEvents( )` is a generic util that cannot require the `this` scope to be set.

src/components/tooltip/tooltip.js
```diff
@@ -219,13 +181,13 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
 
       // Check if we should display it or not.
       // This handles hide-* and show-* along with any user defined css
-      if ( hasComputedStyleValue('display','none') ) {
+      if ( $mdUtil.hasComputedStyleValue('display','none', element[0]) ) {
```
ThomasBurleson: Invert the method signature:  `$mdUtil.hasComputedStyleValue( target, key, defaultVal )`

src/core/util/util.js
EladBezalel: I didn't got to push the fixed code.. This is not there anymore

src/core/util/util.js
EladBezalel: I did it because I saw it in other functions in mdUtil, will be fixed

src/components/tooltip/tooltip.js
EladBezalel: :+1:

PR #5454, opened 2015-10-30
fix(layout) remove attribute only if it exists

alissoncunha:
captura de tela 2015-10-30 as 12 18 14

Using the ng-messages-include causing error when you try to remove the element attribute.

### DIFF
--- a/src/core/services/layout/layout.js
+++ b/src/core/services/layout/layout.js
@@ -264,7 +264,7 @@
       updateFn(getNormalizedAttrValue(className, attrs, ""));
       scope.$on("$destroy", function() { unwatch() });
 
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@alissoncunha - can you elaborate why this issue happens with `ng-messages` ?

alissoncunha:
http://codepen.io/alissoncunha/pen/XmYeGV

ThomasBurleson:
The issue is that ng-messages (and perhaps other ng directives) inserts a **comment** node; which does not support `.removeAttribute()`.

screen shot 2015-11-04 at 5 49 43 am

The fix is a different solution. Nevertheless thank you for the alert.

Issue #5243, opened 2015-10-18
md-fab-dial: click event not recognized correctly


kschwidder:
With latest DEMO-HEAD a double click of a md-fab-dial does not recognize correctly a later click event.
In another case the fab-dial immediately opens and closes on a first click.

The current change of the gesture seems to break the behaviour.

Tested on CHROME

dohomi:
+1 I see the same problem in current 0.12.0-rc1. Sometimes double click sometimes single click is needed.

HBO2:
+1 also a problem in 1.0.0-rc1 After first click it opens and closes and from that point the behaviour is like expected. A click outside the FAB area will it's start all over again.

pzmosquito:
+1 the behavior is very inconsistnent

sfaizanh:
+1 double click to open for the first time, when i clicked on action button same behavior occur, have to double click on fab icon again to open. [1.0.0-RC1] Tested on Chrome Version 46.0.2490.80 (64-bit)

sebakerckhof:
The problem seems to be that sometimes the click event fires twice. So it quickly toggles open state and goes back to the original state.
I've modified the plunkr from the docs so it shows the open state in the console: http://codepen.io/anon/pen/qOYKLQ

Sometimes one click triggers 2 state changes...

If you play around with it a lot, you'll notice that it seems to happen completely random

topherfangio:
@all Can you please verify if this is still an issue in master?

We pushed a fix last night (https://github.com/angular/material/commit/32c0fe1891eb658bc57082de0c836c98e99c22cf) and the codepen that @sebakerckhof appears to be working properly (the codepen unfortunately uses a bad URL that always points to the HEAD version of our files; this issue should be fixed in the next release).

sebakerckhof:
Ah, that's why I couldn't reproduce it on my other PC this morning, I've been clicking like crazy for 10 minutes :p. Anyway, seems fixed.

kschwidder:
@topherfangio Has something changed to open/close the fab-dial ? Now the fab-dial is always open and does not close when clicked.

        <md-fab-speed-dial md-open="uiDial.isOpen" md-direction="right"
                           ng-class="uiDial.selectedMode">
            <md-fab-trigger>
                <md-button md-theme="light-green" class="md-fab md-raised cs-button-new md-mini" aria-label="More">
                    <img src="template/icons/view/ic_more_vert_black_24dp.png">
[...]

PR #5458, opened 2015-10-30
docs(fabToolbar): Replace radio buttons with checkbox

sebakerckhof:
Make behavior consistent with fab speed dial
As per https://github.com/angular/material/issues/5243#issuecomment-152566675


sebakerckhof:
@ThomasBurleson Not that not changing this, means the toolbar will close when you click `open` if it's already open. While this is not a bug, it is very weird behavior to close if you click `open`.

Issue #4450, opened 2015-09-03
Autocomplete dropdown is too narrow when inside of md-chips


mxab:
Hi,

if I use a autocomplete in a chips the dropdown items are too short, is there any way to influence the dropdown's width?
![screen shot 2015-09-03 at 11 22 55](https://cloud.githubusercontent.com/assets/1607547/9654929/38abf122-522e-11e5-8fe4-993a2ef6b668.png)

http://codepen.io/mxab/pen/xwbZBL?editors=101

nestorneto:
same here. 0.10.1 and 0.11.0-rc1

topherfangio:
I was looking into this the other day but didn't have time to figure out why the CSS wasn't working like it should, as the inner items are the correct size. If either of you have a moment to dig in, some more information about exactly why the width is off would be very much appreciated.

nestorneto:
the width is set in directive programmatically.

topherfangio:
@nestorneto From what I was seeing, the `min-width`/`max-width` was set programmatically, but not the actual width. Are you seeing the actual `width` property being set?

nestorneto:
it didnt set width. maybe its the problem. the component assumes the min-width

topherfangio:
@nestorneto That's the part I couldn't figure out. Just setting min/max shouldn't actually affect it at all, it should still be sizing to the items contents as long as it's between min/max. I think there is an odd CSS interaction going on here. We'll try to track it down.

paulflo150:
This is what's causing the items to be 16px not as wide (if you set it to 0 it will fill the whole area):
````
.md-virtual-repeat-container .md-virtual-repeat-offsetter {
 .....
right: 16px;
... }
````

There is also the a sizer which is causing the scrollbars, if you take it out the scrollbars go away, but I am sure it has a use :)

````
<div class="md-virtual-repeat-sizer" style="height: 48px;"></div>
````

It would be great if you could also fix the height of the dropdown when there are no matches - it defaults to 226 px because the resizing does not execute if there are no items:
````
VirtualRepeatController.prototype.containerUpdated = function() { 
  // If itemSize is unknown, attempt to measure it.
    if (!this.itemSize) {
    this.unwatchItemSize_ = this.$scope.$watchCollection(
        this.repeatListExpression,
        angular.bind(this, function(items) {
          if (items && items.length) {
            this.$$rAF(angular.bind(this, this.readItemSize_));
          }
````

[...]

PR #5497, opened 2015-11-02
refactor(chips): Deprecate md-on-append in favor of new methods.

### LINKED ISSUE #4193: mdChips: feature: md-on-append as a simple callback

### LINKED ISSUE #4412: chips with mdOnAppend and autocomplete yields append function two times on enter
Hi,

if I use the chips component with mdOnAppend and the autocomplete.
When start typing and then hit enter, the mdOnAppend gets called two times, once with the string value and once with the selected chip.

See this example
http://codepen.io/mxab/pen/PPoRVQ?editors=101

### LINKED ISSUE #4666: chips/contactChips: md-on-append and md-require-match need better docs
Currently, the docs for `md-on-append` and `md-require-match` do not explain very well how they work alone or when mixed together.

Need to provide some more info/usage snippets and potentially a demo to clarify.

Should wait until PRs #3816 and #4479 are finished/merged so functionality works better (particularly, you should be ale to use both and have `md-on-append` return null to disallow adding a chip, but this doesn't work yet).

### LINKED ISSUE #4863: Contact chips bugs (elements cut off)
Chrome 47: content is cut off, there's an ugly scrollbar (on purpose?!) and there's a white gap between entries and the scrollbar:
ng-material-contact-chips-chrome

Firefox 40: same as in Chrome, but additionally a problem with overly long entries in the contacts overview (no truncation)
ng-material-contact-chips-firefox

topherfangio:
The usage of `md-on-append` was not well documented and it's behavior was inconsistent and confusing; many users assumed it was a simple notification of chip additions which caused issues.

After much discussion, we have renamed `md-on-append` to `md-transform-chip` and provided a new `md-on-add` method that is strictly a notification.

Additionally, we have updated the docs and functionality of `md-transform-chip` to show expected return values and their associated behavior.

This new behavior also adds support for simlultaneously using an autocomplete to select an existing value along with the ability to create new chips. The most common case for this is a tag system which shows existing tags, but also allows you to create new ones.

Demos have been updated to show new functionality as well as to workaround a few display issues with the contact chips demo (#4450).

_**Note 1:** This work supercedes PR #3816 which can be closed when this is merged._

_**Note 2:** This also adds an autocomplete demo inside of a dialog to ensure I did not break that functionality._

BREAKING CHANGE:
`md-on-append` has been renamed/deprecated in favor of `md-transform-chip` or the simple notifier `md-on-add`.

We expect to remove this completely in 1.0, so please update your code to use one of the new methods.

Fixes #4666. Fixes #4193. Fixes #4412. Fixes #4863.

### DIFF
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -73,8 +73,30 @@ function MdChipsCtrl ($scope, $mdConstant, $log, $element, $timeout) {
    * Whether to use the onAppend expression to transform the chip buffer
    * before appending it to the list.
    * @type {boolean}
+   *
+   *
+   * @deprecated Will remove in 1.0.
    */
   this.useOnAppend = false;
+
+  /**
+   * Whether to use the transformChip expression to transform the chip buffer
+   * before appending it to the list.
+   * @type {boolean}
+   */
+  this.useTransformChip = false;
+
+  /**
+   * Whether to use the onAdd expression to notify of chip additions.
+   * @type {boolean}
+   */
+  this.useOnAdd = false;
+
+  /**
+   * Whether to use the onRemove expression to notify of chip removals.
+   * @type {boolean}
+   */
+  this.useOnRemove = false;
 
   /**
    * Whether to use the onSelect expression to notify the component's user
@@ -92,6 +114,11 @@ function MdChipsCtrl ($scope, $mdConstant, $log, $element, $timeout) {
  */
 MdChipsCtrl.prototype.inputKeydown = function(event) {
   var chipBuffer = this.getChipBuffer();
+
+  // If we have an autocomplete, and it handled the event, we have nothing to do
+  if (this.hasAutocomplete && event.isDefaultPrevented && event.isDefaultPrevented()) {
+    return;
+  }
 
   switch (event.keyCode) {
     case this.$mdConstant.KEY_CODE.ENTER:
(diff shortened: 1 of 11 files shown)


topherfangio:
@typotter - @ThomasBurleson asked if you could please review this chips PR for me when you have a moment :smile:

src/components/chips/js/chipsDirective.js
```diff
@@ -67,11 +67,16 @@
    *    displayed when there is at least on item in the list
    * @param {boolean=} readonly Disables list manipulation (deleting or adding list items), hiding
    *    the input and delete buttons
-   * @param {expression} md-on-append An expression that when called expects you to return an object
-   *    representation of the chip input string.
+   * @param {expression} md-on-append An expression of form `myFunction($chip)` that when called
```
typotter: I think this is not a good API... Partly my fault for the bad naming.
The original intent of md-on-append was to provide a means for the Chips user to build a new object from the input string. It should really have been named something like "chip-builder".
Using the md-on-append as a notification mechanism is troublesome because the String published is often not the final value to be added (esp. since it's expected that on-append will change it), and would be further broken if any other processing is added.

Is md-on-select not sufficient for notifying of new chips?

topherfangio: The `md-on-select` could indeed be used, but that forces the user to watch the list for changes to see whether the item was added or just selected by the user.

Additionally (and more importantly), there is currently no way to use the autocomplete functionality along with the "add chip" functionality, so if users attempt to do this, they wind up getting the chip from the autocomplete AND a blank/empty chip (#4412).

typotter: Would it make more sense then to add a properly named event method for when something is added? like md-on-add? We could mark md-on-append as deprecated, giving it a better name like md-build-chip to avoid confusion.

topherfangio: Okay, talked briefly with Thomas, and we came up with splitting it into `md-transform-chip` and `md-on-append`.

Sound good to you? If so, I'll refactor it soon.

typotter: Sounds Great. Thanks Topher (And Thomas)!

topherfangio:
@typotter Okay, I have renamed `md-on-append` as `md-transform-chip`, added the new `md-on-add` method and added a deprecation warning urging users to use one of the new methods.

Can you take one more look and make sure it looks good?

src/components/chips/js/chipsController.js
```diff
@@ -73,10 +73,32 @@ function MdChipsCtrl ($scope, $mdConstant, $log, $element, $timeout) {
    * Whether to use the onAppend expression to transform the chip buffer
    * before appending it to the list.
    * @type {boolean}
+   *
+   *
+   * DEPRECATED: Will remove in 1.0.
```
typotter: @deprecated ?

src/components/chips/js/chipsController.js
```diff
@@ -189,30 +216,40 @@ MdChipsCtrl.prototype.getAdjacentChipIndex = function(index) {
...
+      newChip = transformedChip;
+    }
+  }
 
-   // If useOnAppend and onAppend function is provided call it.
-   if (this.useOnAppend && this.onAppend) {
-     newChip = this.onAppend({'$chip': newChip});
-   }
+  // If items contains an identical object to newChip, do not append
+  if(angular.isObject(newChip)){
```
typotter: nit: space after if

src/components/chips/js/chipsController.js
```diff
@@ -189,30 +216,40 @@ MdChipsCtrl.prototype.getAdjacentChipIndex = function(index) {
...
+    var transformedChip = this.transformChip({'$chip': newChip});
+
+    // Check to make sure the chip is defined before assigning it, otherwise, we'll just assume
+    // they want the string version.
+    if (angular.isDefined(transformedChip)) {
+      newChip = transformedChip;
+    }
+  }
 
-   // If useOnAppend and onAppend function is provided call it.
```
typotter: shouldn't you still be using onAppend, since it hasn't been removed yet?

src/components/chips/js/chipsController.js
```diff
@@ -220,9 +257,35 @@ MdChipsCtrl.prototype.getAdjacentChipIndex = function(index) {
  * {@code onAppend}. Due to the nature of directive scope bindings, the
  * controller cannot know on its own/from the scope whether an expression was
  * actually provided.
+ *
+ // TODO: Remove deprecated md-on-append functionality in 1.0
  */
 MdChipsCtrl.prototype.useOnAppendExpression = function() {
-  this.useOnAppend = true;
+  this.$log.warn("md-on-append is deprecated; please use md-transform-chip or md-on-add instead");
+  this.useTransformChip = true;
```
typotter: maybe only set these if md-transform-chip is not already set.

typotter:
Left some comments. Overall LGTM. nice work!

topherfangio:
@typotter Thanks! I've made your suggested changes :-)

@ThomasBurleson Should be ready for merge now!

PR #5498, opened 2015-11-02
BUG: defaultFormatDate

paulorwd:
did not support date as string

### DIFF
--- a/src/components/datepicker/dateLocaleProvider.js
+++ b/src/components/datepicker/dateLocaleProvider.js
@@ -158,6 +158,10 @@
         if (!date) {
           return '';
         }
+				
+				if(typeof date === 'string') {
+					date = new Date(date);
+				}
 
         // All of the dates created through ng-material *should* be set to midnight.
         // If we encounter a date where the localeTime shows at 11pm instead of midnight,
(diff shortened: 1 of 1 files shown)


paulorwd:
I signed it!

ThomasBurleson:
@jelbourn - can you review ?

src/components/datepicker/dateLocaleProvider.js
```diff
@@ -158,6 +158,10 @@
         if (!date) {
           return '';
         }
+				
+        if (typeof date === 'string') {
```
jelbourn: The entire point of this function is that it turns a Date object into a string. Under what circumstance would it be passed a string?

paulorwd: When the date comes from the database, comes in string format. I also realized that require further adjustment, because it ends up disabling the calendar so we need to convert the $ scope before reaching the DatePicker.

ThomasBurleson: How else would you do you propose to the the date value/string to the DatePicker instance (except my $scope or `controllerAs` references?

jelbourn:
The API for the datepicker requires that the model value be an instance of `Date` rather than a string. You should do the conversion in your own controller code before binding to the control.

paulorwd:
It would be interesting component already streamline this part, but for now I am doing this way, treating before reaching variable $scope.

Issue #5266, opened 2015-10-20
md-Datepicker always throws an error


Raidri:
Hi,

i use the md-datepicker with moment.js and the $mdDateLocaleProvider. I get dates from a Server with values like 2100-12-31 00:00:00 or 2015-10-11 or something else. Now i want to use the datepicker to show the date, but i got this error:

![error](https://cloud.githubusercontent.com/assets/148783/10599648/473137fc-7704-11e5-92a1-c5481c932ce0.png)

And here a simple example for it:

http://codepen.io/Raidri/pen/avyWEB

jelbourn:
The problem here is that you're trying to bind a string as the model to the `md-datepicker` when it is expecting a `Date` instance. #5503 makes this less ambiguous.

stunaz:
I having the same issue, and yes also trying to bind a string. But I thought that the ```$mdDateLocaleProvider.parseDate``` was there for that ie to parse a given string to a date, and ```$mdDateLocaleProvider.formatDate``` to take care of the way it is displayed. @jelbourn  If I am wrong, can you please explain the purpose of these two functions, the doc doesn't mention when and how it is used.

jelbourn:
`parseDate` converts a string that the user has typed into a Date instance.
`formatDate` converts a Date instance from the model into a string that's set to the text input.

stunaz:
Thanks for the clarification

PR #5503, opened 2015-11-02
fix(datepicker): throw error if model is not a Date. For #5266

jelbourn:
@mhchen for review


ThomasBurleson:
@jelbourn - can you rebase plz.

PR #5510, opened 2015-11-03
fix(datepicker): apply ngMessages errors when using text input. Fixes #5353

### LINKED ISSUE #5353: md-datepicker with key input won't fire a change event
Hi,

in the actual version md-datepicker fires a ng-change event only, if you select a date with the mouse. If you enterred it with the keyboard it won't be updated. That have a side effect on ng-messages, because they will be only updated with a ng-change event. A correct enterred date with keyboard will not be accepted.

jelbourn:
Problem was that error flags only got called on `$render`, which didn't occur because the ngModel isn't updated when the value is invalid.

@mhchen for review


ThomasBurleson:
@jelbourn - needs rebased.

jelbourn:
Just waiting on LGTM from @mhchen

mhchen:
So sorry, thought I LGTMed already

PR #5528, opened 2015-11-03
fix(sidenav): Notify other close events like datepicker-close too

### LINKED ISSUE #5522: md-datepicker does not close when inside md-sidenav
md-datepicker does not close when inside an md-sidenav and the user clicks off of the sidenav. It appears that the sidenav is capturing the click event

Codepen adding a datepicker to the official md-sidenav demo: http://codepen.io/anon/pen/KdBmPL Click "Toggle Right". The datepicker is in the right sidenav.

datepicker-sidenav-interaction

DevVersion:
md-datepicker won't close if sidebar is open and the sidebar event stops propagation

Codepen to checkout: http://codepen.io/anon/pen/zvLeRN

Fixes #5522

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -375,7 +375,6 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
      */
     function close(ev) {
       ev.preventDefault();
-      ev.stopPropagation();
 
       return sidenavCtrl.close();
     }


ThomasBurleson:
@jelbourn - see PR https://github.com/angular/material/pull/5528

ThomasBurleson:
@jelbourn - I think this makes sense and has no side-affects. Your thoughts ?

jelbourn:
Seems reasonable to me, but I have no idea why the sidenav was stopping propagation to being with.

DevVersion:
Perhaps they want to enable layer dismissing. So the user can only close per click the ```active``` layer.

PR #5532, opened 2015-11-04
fix(ripple): removing ripples on touchmove

### LINKED ISSUE #5261: md-list-item ripple + Iphone/android user agent
http://codepen.io/0101adm/full/LpOwBq/

in chrome's device mode emulating iphone 5s, open dialog. 

dragging up/down (to scroll) causes the ripple to permanently shade the md-list-item.

EladBezalel:
When touching a button in order to scroll, the button creates a ripple as its being pressed, now when the touch moved we assume that this gesture is a scroll so we remove all the ripple effects

fixes #5261


EladBezalel:
@ThomasBurleson @robertmesserle

PR #5533, opened 2015-11-04
fix(lists): Fix alignment of secondary icons/controls.

### LINKED ISSUE #3699: md-list control alignment issue
I am using `md-list` to display a list that has a different control depending on context. The control may be a checkbox or icon.

The icons and checkboxes are not aligned...

screen shot 2015-07-13 at 10 31 34

topherfangio:
Icons and controls (like checkbox and switch) were incorrectly positioned within a list. Fix by ensuring that they are all right-aligned.

Fixes #3699.

### DIFF
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -135,9 +135,16 @@ md-list-item, md-list-item .md-list-item-inner {
 
   md-checkbox.md-secondary,
   md-switch.md-secondary {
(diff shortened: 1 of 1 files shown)


topherfangio:
@ThomasBurleson Ready for review/merging. Definitely want a second pair of eyes to make sure I'm not missing something in the way they were previously aligned.

PR #5538, opened 2015-11-04
fix(all): fix issue with removeClass removing previous classes

mallowigi:
In certain parts of the project, a "removeClass(className)" is used to remove the class from the element. While this works fine in latest versions of jQuery, in jQuery < 1.9 this results in the element classes being removed when className is undefined, thus breaking the whole layout.

Adding a null check for each of those statements to prevent this behavior.

### DIFF
--- a/src/components/checkbox/checkbox.js
+++ b/src/components/checkbox/checkbox.js
@@ -167,7 +167,7 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
         if(ngModelCtrl.$viewValue) {
           element.addClass(CHECKED_CSS);
         } else {
-          element.removeClass(CHECKED_CSS);
+          if (CHECKED_CSS) element.removeClass(CHECKED_CSS);
         }
       }
     };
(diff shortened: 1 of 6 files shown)


src/components/checkbox/checkbox.js
```diff
@@ -167,7 +167,7 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
         if(ngModelCtrl.$viewValue) {
           element.addClass(CHECKED_CSS);
         } else {
-          element.removeClass(CHECKED_CSS);
+          if (CHECKED_CSS) element.removeClass(CHECKED_CSS);
```
ThomasBurleson: Why do this `if( ...)` when CHECKED_CSS is defined?

src/components/progressCircular/progress-circular.js
```diff
@@ -118,11 +118,11 @@ function MdProgressCircularDirective($mdTheming, $mdUtil, $log) {
          case MODE_DETERMINATE:
          case MODE_INDETERMINATE:
            spinnerWrapper.removeClass('ng-hide');
-           spinnerWrapper.removeClass( lastMode );
+           if (lastMode) spinnerWrapper.removeClass(lastMode);
```
ThomasBurleson: Would this not work `spinnerWrapper.removeClass(lastMode || "")`

mallowigi: Yes it could work too. Why? Is there a problem with the null check?

src/components/checkbox/checkbox.js
mallowigi: I see, this is a constant. I just did a full search for removeClass with variables, I didn't pay attention to the details.

ThomasBurleson:
Please update PR - use of constants should not be modified.

PR #5556, opened 2015-11-04
feat(button): add download flag (boolean) to allow anchor download

### LINKED ISSUE #5542: md-button not respecting "download" attribute
I have an md-button with ng-href. I would like to use html5 attribute "download", but when I add this attribute to md-button, it's ignored. I do not see this attribute in the <a> created.

DevVersion:
Flag allows to download the specified ```ng-href```.

Closes #5542

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -25,6 +25,7 @@ angular
  * the `md-primary` class.
  *
  * @param {boolean=} md-no-ink If present, disable ripple ink effects.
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
This is not a full solution... and needs a team discusson; post v1.0

BAByrne:
If the team decides to add this, should similar functionality to be added to md-list-item as well?

PR #5564, opened 2015-11-04
fix(autocomplete): Allow clicks inside md-not-found.

### LINKED ISSUE #5424: Error on Autocomplete Demo
Hi, 
I don't understand how use mdAutocomplete and mdNotFound with clickable item into mdNotFound.
On Demo(1.0.0 RC1) the link is clickable and working but on codepen the link is not clickable. If i use the exemple on my project ng-click don't work as codepen. 

Any Issue for this problem? 
Regards

topherfangio:
Fix bug causing the autocomplete to blur, lose focus, and hide the not found message if the user attempted to click the not found message (or a link inside of it).

Fixes #5424.

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -359,8 +359,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    * Handles input blur event, determines if the dropdown should hide.
    */
   function blur () {
-    hasFocus = false;
     if (!noBlur) {
+      hasFocus = false;
       ctrl.hidden = shouldHide();
     }
   }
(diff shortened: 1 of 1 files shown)


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -359,8 +359,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    * Handles input blur event, determines if the dropdown should hide.
    */
   function blur () {
-    hasFocus = false;
```
topherfangio: @robertmesserle Can you think of any reason why the `hasFocus` was outside of the `if` statement? All tests pass after moving it and my manual testing didn't show any differences except that it fixed my bug.

robertmesserle: Yes, this is so that we know when the element lost focus so that we can refocus when they leave the dropdown. This was related to ie 11.

ThomasBurleson:
@topherfangio - what is the status of this PR ?

src/components/autocomplete/js/autocompleteController.js
ThomasBurleson: I restored it:

```js
  function blur () {
    hasFocus = false;
    if (!noBlur) {
      ctrl.hidden = shouldHide();
    }
  }
```

ThomasBurleson:
Fixed with SHA 5ae3d4cd69a

Issue #4336, opened 2015-08-26
md-virtual-repeat does not render items if it begins in hidden state


uriva:
http://codepen.io/uriva/pen/PqMMqV

Anaphase:
+1 I just ran across this issue too

Anaphase:
This could possibly be related to a similar bug documented in #4499.

Anaphase:
[This bug is still present in 0.11.1](http://codepen.io/Anaphase/pen/EVZqrX)

kseamon:
Fixed as of 1.0-rc4

apils:
Probably related, virtual-repeat does not render when it is in a sidnav that is initially hidden.  I'm using 1.0.5 and it's still happening.  Works fine in sidenav that is initially open.

impressto:
I am using 1.1.0-rc2 and it is still an issue. I was setting the scope for my repeater items before setting visible the md-virtual-repeat component. When setting the component visible, I then used this hack to force a redraw - angular.element(window).triggerHandler('resize');

demsey2:
I am on 1.1.0-rc5 and the problem still exists.

jbernard1:
Also on 1.1.0-rc5 and the problem persist, ill test it on rc4

vucalur:
I came across this thread when looking for a cause of my virtualRepeat staying hidden.
Since it was in nested UI-Router states, I though bug described here was the case.

Turned out the case was trivial:
I haven't added CSS `height` property to `md-virtual-repeat-container` element. (`width` for horizontal mode)
Analogous `ng-repeat` works without it,
so it should be mentioned in the docs…

Unrelated to the bug, but I hope someone will find this helpful.

CharlesBean:
My quick hack was just wrapping the md-virtual-repeat-container with an ng-if:

`<div ng-if="someCtrl.isSidenavReady">
  <md-virtual-repeat-container>
    <md-virtual-repeat md-on-demand md-virtual-repeat="thing in someCtrl.things">
    </md-virtual-repeat>
  </md-virtual-repeat-container>
</div>`

I set the flag to true when the button that opens the sidenav is clicked. Causes the repeater to re-draw and works for me temporarily!

Anaphase:
@CharlesBean this is what I ended up doing as well

aeslinger0:
[...]

Issue #4314, opened 2015-08-25
md-virtual-repeat does not work without predefined height


uriva:
codepen: http://codepen.io/uriva/pen/doxdVP

Pusnow:
I think md-virtual-repeat-container should have fixed height

uriva:
Yup. That was the issue. Perhaps a doc clarification on that point?

Pusnow:
I referred to official doc's sample codes

robertbaker:
+1 Would be nice to figure out a way to allow flexable heights.

kseamon:
It's on my to-do list. It's tricky because any solution comes with performance trade-offs.

alias1:
+1 This tripped me up for a while, a note on the docs would be cool.

colinskow:
Here's my workaround in the mean time:

```js
this.getListHeight = function() {
  return {height: '' + ($window.innerHeight - 72) + 'px'};
};
$window.addEventListener('resize', onResize);
function onResize() {
  $scope.$digest();
}
$scope.$on('$destroy', function() {
  $window.removeEventListener('resize', onResize);
});
```

Then in the HTML

```html
<md-virtual-repeat-container id="vertical-container" ng-style="ctrl.getListHeight()">
...
</md-virtual-repeat-container>
```

Madd0g:
@colinskow - Thanks, it's close to working, it works for me on refresh, but not on regular state-changes
(Note that I'm measuring the parent node of the virtual list, not the body like in your example).

A huge +1 for making virtual-repeat flexable.

colinskow:
@Madd0g - I actually refactored it a little bit and it seems to work better. At least on Safari and Chrome it always appears to be the right size.

```js
this.listStyle = {
      height: ($window.innerHeight - 112) + 'px'
    };
    $window.addEventListener('resize', onResize);
    function onResize() {
      self.listStyle.height = ($window.innerHeight - 112) + 'px';
      if(!$scope.$root.$$phase) $scope.$digest();
    }
```

```html
<md-virtual-repeat-container id="vertical-container" ng-style="ctrl.listStyle">
...
</md-virtual-repeat-container>
```

Madd0g:
@colinskow - that does work better, but it seems like the internal `md-list` doesn't get resized to the necessary height, I don't even understand what causes it to be smaller in size, I've combed through the styles tab and don't see anything that can cause it :(

colinskow:
[...]

PR #5579, opened 2015-11-05
feat(showHide): Introduce directives to notify child directives on hi…

kseamon:
…de/show.

Features extreme tail-hooking technology.
Step 2 of addressing #4314 and #4336

### DIFF
--- a/src/components/showHide/showHide.js
+++ b/src/components/showHide/showHide.js
@@ -1,0 +1,35 @@
+/**
+ * @ngdoc module
+ * @name material.components.showHide
+ */
+
+// Add additional handlers to ng-show and ng-hide that notify directives
+// contained within that they should recompute their size.
+// These run in addition to Angular's built-in ng-hide and ng-show directives.
+angular.module('material.components.showHide', [
+  'material.core'
+])
+  .directive('ngShow', createDirective('ngShow', true))
+  .directive('ngHide', createDirective('ngHide', false));
+
+
+function createDirective(name, targetValue) {
+  return ['$mdUtil', function($mdUtil) {
+    return {
+      restrict: 'A',
+      multiElement: true,
+      link: function($scope, $element, $attr) {
+        $scope.$watch($attr[name], function(value) {
+          if (!!value === targetValue) {
+            $mdUtil.nextTick(function() {
+              $scope.$broadcast('$md-resize');
+            });
+            $mdUtil.dom.animator.waitTransitionEnd($element).then(function() {
+              $scope.$broadcast('$md-resize');
+            });
+          }
+        });
+      }
+    };
+  }];
+}
--- a/src/components/showHide/showHide.spec.js
+++ b/src/components/showHide/showHide.spec.js
@@ -1,0 +1,86 @@
+describe('showHide', function() {
+  var $compile, $timeout, defered, scope, spy;
+
+  beforeEach(module('material.components.showHide'));
+
+  beforeEach(inject(function(_$compile_, $mdUtil, $q, $rootScope, _$timeout_) {
+    $compile = _$compile_;
+    $timeout = _$timeout_;
+    defered = $q.defer();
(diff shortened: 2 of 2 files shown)


kseamon:
@jelbourn

src/components/showHide/showHide.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * @ngdoc module
+ * @name material.components.showHide
+ */
+
+angular.module('material.components.showHide', [
+  'material.core'
+])
+  .directive('ngShow', createDirective('ngShow', true))
```
jelbourn: I would move part of the later comment here, saying something like
```js
// Create our own directives 'ngShow' and 'ngHide' that are run in 
// _addition_ to the normal ngShow/ngHide. We use these to...
```

jelbourn:
LGTM

src/components/showHide/showHide.js
```diff
@@ -0,0 +1,32 @@
...
+function createDirective(name, targetValue) {
+  return ['$mdUtil', function($mdUtil) {
+    return {
+      restrict: 'A',
+      multiElement: true,
+      link: function($scope, $element, $attr) {
+        $scope.$watch($attr[name], function(value) {
+          if (!!value === targetValue) {
+            $scope.$broadcast('$md-resize');
+            $mdUtil.dom.animator.waitTransitionEnd($element).then(function() {
```
ThomasBurleson: @kseamon - why are you broadcasting 2x?

src/components/showHide/showHide.spec.js
```diff
@@ -0,0 +1,77 @@
...
+      var element = $compile('<div ng-hide="hide"></div>')(scope);
+      scope.$apply();
+      expect(spy).not.toHaveBeenCalled();
+
+      scope.hide = false;
+      scope.$apply();
+      expect(spy).toHaveBeenCalled();
+
+      spy.calls.reset();
+      defered.resolve();
```
ThomasBurleson: What are you doing here... it is not clear.

src/components/showHide/showHide.spec.js
```diff
@@ -0,0 +1,77 @@
...
+      scope.show = false;
+      var element = $compile('<div ng-show="show"></div>')(scope);
+      scope.$apply();
+      expect(spy).not.toHaveBeenCalled();
+
+      scope.show = true;
+      scope.$apply();
+      expect(spy).toHaveBeenCalled();
+
+      spy.calls.reset();
```
ThomasBurleson: Can you break this group out into a understand function call ?

src/components/showHide/showHide.js
```diff
@@ -0,0 +1,32 @@
...
+// Add additional handlers to ng-show and ng-hide that notify directives
+// contained within that they should recompute their size.
+function createDirective(name, targetValue) {
+  return ['$mdUtil', function($mdUtil) {
+    return {
+      restrict: 'A',
+      multiElement: true,
+      link: function($scope, $element, $attr) {
+        $scope.$watch($attr[name], function(value) {
+          if (!!value === targetValue) {
```
ThomasBurleson: :+1:

src/components/showHide/showHide.js
kseamon: Yeah, my concern was that not every element has transitions, and 3s is longer than I'd like to wait in those cases.

src/components/showHide/showHide.js
kseamon: Done

src/components/showHide/showHide.spec.js
kseamon: Added comments

src/components/showHide/showHide.spec.js
kseamon: I think the comments added are sufficient. Have a look.

PR #5583, opened 2015-11-05
fix(sidenav): Only restore focus if user interaction was keyboard

DevVersion:
Normally the Sidenav restores every time the last active element, so the focus won't disappear from the triggered button after closing the sidenav

Fixes #5563

PR #5589, opened 2015-11-06
feat(interaction): added service to detect last interaction

### LINKED ISSUE #5434: Ripple 'sticks' if using UI Router in the app under certain conditions
This is an odd one, but since so many people use UI-Router it might be worth looking at 

Possibly related to or interacts with https://github.com/angular/material/issues/5294 https://github.com/angular/material/issues/5046

To Reproduce ( Using Chrome + Windows 10 or [Codpen Example](http://codepen.io/Mopholo/pen/LprEdW?editors=001) )

1. Click on a component that has a route attached using ui-sref 
2. Click outside the app ( another tab in browser or whatever)
3. Click back inside the app, notice the ripple is now 'stuck' until you click something inside the app that can get focus

This gif loop probably explains it better ;-) 

focus-problem

### LINKED ISSUE #5563: md-toolbar and sidenav focus problems
For testing purposes I tried to create a ```sidenav```, which should be triggered by the ```toolbar```
But as you can see in the bad quality GIF, the css class ```is-focused``` will be added after dismissing the sidebar. So I already checked the code and I found this snippet
````
return $q(function(resolve){
  // Toggle value to force an async `updateIsOpen()` to run
  scope.isOpen = isOpen;

  $mdUtil.nextTick(function() {
    // When the current `updateIsOpen()` animation finishes
    promise.then(function(result) {

      if ( !scope.isOpen ) {
        // reset focus to originating element (if available) upon close
        triggeringElement && triggeringElement.focus();
        triggeringElement = null;
      }

      resolve(result);
    });
  });

});
```
![gif](https://i.gyazo.com/42f9eaa131791d1431aeaf0d64f696cf.gif)

I already fixed that in my version, but I want to ask the development team first about their thoughts.
So I created a merge for that, please take a look at it and tell me more about potential side-affects

@jelbourn @ThomasBurleson

DevVersion:
Sidenav want's to restore focus after the pane gets closed, this should only happen for keyboard interactions.
The service got created to use that feature for other components too

Fixes #5563 Fixes #5434 Closes #5583

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -203,7 +203,7 @@ function SidenavFocusDirective() {
  *   - `<md-sidenav md-is-locked-open="$mdMedia('min-width: 1000px')"></md-sidenav>`
  *   - `<md-sidenav md-is-locked-open="$mdMedia('sm')"></md-sidenav>` (locks open on small screens)
  */
-function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, $compile, $parse, $log, $q, $document) {
+function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $mdInteraction, $animate, $compile, $parse, $log, $q, $document) {
   return {
     restrict: 'E',
     scope: {
@@ -223,6 +223,7 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
   function postLink(scope, element, attr, sidenavCtrl) {
     var lastParentOverFlow;
     var backdrop;
+    var triggeringInteractionType;
     var triggeringElement = null;
     var promise = $q.when(true);
     var isLockedOpenParsed = $parse(attr.mdIsLockedOpen);
@@ -295,6 +296,7 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
       if ( isOpen ) {
         // Capture upon opening..
         triggeringElement = $document[0].activeElement;
+        triggeringInteractionType = $mdInteraction.getLastInteractionType();
       }
 
       disableParentScroll(isOpen);
@@ -351,9 +353,9 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
             // When the current `updateIsOpen()` animation finishes
             promise.then(function(result) {
 
-              if ( !scope.isOpen ) {
+              if ( !scope.isOpen && triggeringElement && triggeringInteractionType === 'keyboard') {
                 // reset focus to originating element (if available) upon close
-                triggeringElement && triggeringElement.focus();
+                triggeringElement.focus();
                 triggeringElement = null;
               }
 
--- a/src/components/sidenav/sidenav.spec.js
+++ b/src/components/sidenav/sidenav.spec.js
@@ -199,6 +199,84 @@ describe('mdSidenav', function() {
 
   });
 
+  describe("focus", function() {
+
+    var $material, $mdInteraction, $mdConstant;
+
+    beforeEach( inject(function(_$material_, _$mdInteraction_, _$mdConstant_) {
+      $material = _$material_;
+      $mdInteraction = _$mdInteraction_;
+      $mdConstant = _$mdConstant_
+    }));
+
+    function flush() {
+      $material.flushInterimElement();
+    }
+
+    function setupTrigger() {
+      var el;
+      inject(function($compile, $rootScope) {
+        var parent = angular.element(document.body);
+        el = angular.element('<button>Toggle</button>');
+        parent.append(el);
+        $compile(parent)($rootScope);
+        $rootScope.$apply();
+      });
+      return el;
+    }
+
+    it("should restore after sidenav triggered by keyboard", function(done) {
+      var sidenavElement = setup('');
+      var triggerElement = setupTrigger();
+      var controller = sidenavElement.controller('mdSidenav');
+
+      triggerElement.focus();
+
+      var keyboardEvent = document.createEvent("KeyboardEvent");
+      keyboardEvent.initEvent("keydown", true, true, window, 0, 0, 0, 0, $mdConstant.KEY_CODE.ENTER,  $mdConstant.KEY_CODE.ENTER);
+      triggerElement[0].dispatchEvent(keyboardEvent);
+
+      controller.$toggleOpen(true);
+      flush();
+
+      triggerElement.blur();
+
+      controller.$toggleOpen(false);
+      flush();
+
+      expect($mdInteraction.getLastInteractionType()).toBe("keyboard");
+      expect(document.activeElement).toBe(triggerElement[0]);
+      done();
+    });
+
+    it("should not restore after sidenav triggered by mouse", function(done) {
+      var sidenavElement = setup('');
+      var triggerElement = setupTrigger();
+      var controller = sidenavElement.controller('mdSidenav');
+
+      triggerElement.focus();
+
+      var mouseEvent = document.createEvent("MouseEvent");
+      mouseEvent.initMouseEvent("mousedown", true, true, window, null, 0, 0, 0, 0, false, false, false, false, 0, null);
+      triggerElement[0].dispatchEvent(mouseEvent);
+
+      controller.$toggleOpen(true);
+      flush();
+
+      expect(document.activeElement).toBe(triggerElement[0]);
+
+      triggerElement.blur();
+
+      controller.$toggleOpen(false);
+      flush();
+
+      expect($mdInteraction.getLastInteractionType()).toBe("mouse");
+      expect(document.activeElement).not.toBe(triggerElement[0]);
+      done();
+    });
+
+  });
+
   describe("controller Promise API", function() {
     var $material, $rootScope;
 
--- a/src/core/services/interaction/interaction.js
+++ b/src/core/services/interaction/interaction.js
@@ -1,0 +1,95 @@
+angular
+  .module('material.core.interaction', [])
+  .service('$mdInteraction', MdInteractionService);
+
+/*
+ * @ngdoc service
+ * @name $mdInteraction
+ * @module material.core.interaction
+ *
+ * @description
+ *
+ * Service which keeps track of the last interaction type and validates them for several browsers.
+ * The service hooks into the document's body and listens for touch, mouse and keyboard events.
+ *
+ * The last interaction type can be retrieved by using the `getLastInteractionType` method, which returns
+ * the following possible values:
+ * - `touch`
+ * - `mouse`
+ * - `keyboard`
+ *
+ * Here is an example markup for using the interaction service.
+ * ```
+ *   var lastType = $mdInteraction.getLastInteractionType();
+ *   if (lastType === 'keyboard') {
+ *     restoreFocus();
+ *   }}
+ * ```
+ *
+ */
+function MdInteractionService($timeout) {
+  var body = angular.element(document.body);
+  var mouseEvent = window.MSPointerEvent ? 'MSPointerDown' : window.PointerEvent ? 'pointerdown' : 'mousedown';
+  var buffer = false;
+  var timer;
+  var lastInteractionType;
+
+  // Type Mappings for the different events
+  // There will be three three interaction types
+  // `keyboard`, `mouse` and `touch`
+  // type `pointer` will be evaluated in `pointerMap` for IE Browser events
+  var inputMap = {
+    'keydown': 'keyboard',
+    'mousedown': 'mouse',
+    'mouseenter': 'mouse',
+    'touchstart': 'touch',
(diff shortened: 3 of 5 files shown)


DevVersion:
I added a few more tests, more test will come as soon as possible

src/components/sidenav/sidenav.js
```diff
@@ -227,7 +227,7 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
    */
   function postLink(scope, element, attr, sidenavCtrl) {
     var lastParentOverFlow;
-    var triggeringElement = null;
+    var triggeringElement = null, triggeringInteractionType;
```
EladBezalel: Please separate to another decleration

src/core/services/interaction/interaction.js
```diff
@@ -0,0 +1,53 @@
+angular
+  .module('material.core.interaction', [])
+  .service('$mdInteraction', MdInteractionService);
+
+function MdInteractionService($timeout) {
+  var body = angular.element(document.body);
+  var _mouseEvent = window.MSPointerEvent ? 'MSPointerDown' : window.PointerEvent ? 'pointerdown' : 'mousedown';
+  var timer, lastInteractionType, buffer = false;
```
EladBezalel: Separate declerations

src/core/services/interaction/interaction.js
```diff
@@ -0,0 +1,53 @@
...
+  var inputMap = {
+    'keydown': 'keyboard',
+    'mousedown': 'mouse',
+    'mouseenter': 'mouse',
+    'touchstart': 'touch',
+    'pointerdown': 'pointer',
+    'MSPointerDown': 'pointer'
+  }, pointerMap = {
```
EladBezalel: Separate declarations,
What's the meaning of the keys?

src/core/services/interaction/interaction.js
```diff
@@ -0,0 +1,53 @@
...
+    2: 'touch',
+    3: 'touch',
+    4: 'mouse'
+  };
+
+  function onInput(event) {
+    if (buffer) return;
+    var type = inputMap[event.type];
+    if (type === 'pointer') type = (typeof event.pointerType === 'number') ? pointerMap[event.pointerType] : event.pointerType;
```
EladBezalel: as it's a long line, please wrap with brackets

EladBezalel:
Please fix and squash your commits to 1 commit

DevVersion:
Thanks for your review @EladBezalel, I will fix this today :+1:

src/core/services/interaction/interaction.js
DevVersion: Thats the mapping of the different browser events.

EladBezalel: I was talking about `2: 'touch'`.. etc

src/components/sidenav/sidenav.js
marcysutton: Agreed...the trickiness here hinders code readability.

DevVersion: Won't happen again, didn't know about this style policy ;)

src/core/services/interaction/interaction.js
DevVersion: That provides the type of interaction, as marcy told me. This should be right.

DevVersion:
Commits successfully squashed, thanks for reviewing @EladBezalel @marcysutton

src/core/services/interaction/interaction.js
marcysutton: Also limit line length to 100 chars. https://github.com/angular/material/blob/master/docs/guides/CODING.md#coding

src/core/services/interaction/interaction.spec.js
```diff
@@ -0,0 +1,26 @@
+describe("$mdInteraction", function() {
```
marcysutton: I'd love to see some additional tests here. Since the focus style hinges off of `.md-focused` being added, we could test that it gets added on keyboard interaction and not mouse.

DevVersion: I'm already working on more tests, should I provide tests for the specific components or only for the interaction service (for example mouse -> I already tried to add a mouse test, but phantomjs doesn't allows that) -> But I will look for it.

src/core/services/interaction/interaction.js
DevVersion: I think after my rework of the commits: 105 Chars should be okay ;)

src/core/services/interaction/interaction.js
```diff
@@ -0,0 +1,58 @@
...
+  var inputMap = {
+    'keydown': 'keyboard',
+    'mousedown': 'mouse',
+    'mouseenter': 'mouse',
+    'touchstart': 'touch',
+    'pointerdown': 'pointer',
+    'MSPointerDown': 'pointer'
+  };
+  var pointerMap = {
+    2: 'touch',
```
EladBezalel: i'd really like to know what are those, please add an explanation

DevVersion: # Input Map
MSPointerdown < IE11: https://msdn.microsoft.com/de-de/library/windows/apps/hh465891.aspx
Pointerdown >= IE11: https://msdn.microsoft.com/de-de/library/windows/apps/hh465891.aspx
**Example** -> touchstart: Is associated with a touch interaction (so return the api value: "touch")
The rest you definitely  know :+1: 

#  Pointer Map:
https://msdn.microsoft.com/de-de/library/windows/apps/hh466130.aspx
MSPOINTER_TYPE_TOUCH: 0x00000002
MSPOINTER_TYPE_PEN: 0x00000003
MSPOINTER_TYPE_MOUSE: 0x00000004

Then I will validate the right pointer event type :)

I hope, know its clear :)
I'm up for more questions

EladBezalel: :+1: please add this in a comment :)

DevVersion: @EladBezalel Can I contact you on something else, I mean (live) thats too slow :smile:

EladBezalel: @DevVersion i pinged you on gitter

DevVersion:
Added two tests into the ````sidenav.specs```.
@EladBezalel @marcysutton

src/components/sidenav/sidenav.spec.js
```diff
@@ -173,6 +173,85 @@ describe('mdSidenav', function() {
...
+
+    function flush() {
+      $material.flushInterimElement();
+    }
+
+    function setupTrigger() {
+      var el;
+      inject(function($compile, $rootScope) {
+        var parent = angular.element(document.body);
+        el = angular.element('<button aria-label="Toggle Menu">Toggle</button>');
```
marcysutton: So that the tests show correct requirements, you don't need `aria-label` here since the button has text in it.

src/components/sidenav/sidenav.spec.js
```diff
@@ -173,6 +173,85 @@ describe('mdSidenav', function() {
...
+
+    it("should restore after sidenav triggered by keyboard", function(done) {
+      var sidenavElement = setup('');
+      var triggerElement = setupTrigger();
+      var controller = sidenavElement.controller('mdSidenav');
+
+      triggerElement.focus();
+
+      var evt = document.createEvent("KeyboardEvent");
+      evt.initEvent("keydown", true, true, window, 0, 0, 0, 0, 13, 13);
```
marcysutton: This method is deprecated, is there a reason for using it here?

src/components/sidenav/sidenav.spec.js
```diff
@@ -173,6 +173,85 @@ describe('mdSidenav', function() {
...
+      triggerElement[0].dispatchEvent(evt);
+
+      controller.$toggleOpen(true);
+      flush();
+
+      triggerElement.blur();
+
+      expect(document.activeElement).not.toBe(triggerElement[0]);
```
marcysutton: Not sure what this is doing, other than asserting `element.blur()` works. Can you explain what this test is trying to do?

src/components/sidenav/sidenav.spec.js
```diff
@@ -173,6 +173,85 @@ describe('mdSidenav', function() {
...
+
+    it("should not restore after sidenav triggered by mouse", function(done) {
+      var sidenavElement = setup('');
+      var triggerElement = setupTrigger();
+      var controller = sidenavElement.controller('mdSidenav');
+
+      triggerElement.focus();
+
+      var event = document.createEvent("MouseEvent");
+      event.initMouseEvent("mousedown", true, true, window, null, 0, 0, 0, 0, false, false, false, false, 0, null);
```
marcysutton: This method is also deprecated: https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/initMouseEvent

marcysutton: Here is an example of testing mouse and keyboard. Or is there some nuance with Phantom.js being headless making testing difficult? https://github.com/angular/material/blob/master/src/components/checkbox/checkbox.spec.js#L139

src/components/sidenav/sidenav.spec.js
DevVersion: fixed

src/components/sidenav/sidenav.spec.js
DevVersion: This test is focusing the triggering element, then it executes a click to notify the ```ìnteraction``` service, then I open the Sidenav (all things are setup, like a real keydown). Then I manually open the sidenav with ```$toggleOpen````. After its opened I set the focus to the body (because PhantomJS isn't setting the focus as preferred to the sidenav). Then I validate the activeElement not to be the triggeringElement (because the Sidenav / Body is focused). After that I close that sidenav manually, and the sidenav thoughts I triggered the sidenav with keyboard, so its restoring my old activeElement.

src/components/sidenav/sidenav.spec.js
DevVersion: Yes it was difficult to get this runnning on Phantom.js. It took a lot of time in getting the test working. I already tried that ````triggerHandler``` but then the interaction service, won't get triggered because its not notifying the parent nodes. (this imitates a ```real``` click).

DevVersion:
Squashed will be later.. :+1:

DevVersion:
please checkout @ThomasBurleson

DevVersion:
@EladBezalel @marcysutton - Reworked the PR, rebased to master, improved unit tests, added comments

src/components/sidenav/sidenav.js
```diff
@@ -344,9 +346,9 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
             // When the current `updateIsOpen()` animation finishes
             promise.then(function(result) {
 
-              if ( !scope.isOpen ) {
+              if ( !scope.isOpen && triggeringElement && triggeringInteractionType && triggeringInteractionType === 'keyboard') {
```
```js
triggeringInteractionType && triggeringInteractionType === 'keyboard'
```
can be simplified to
```js
triggeringInteractionType === 'keyboard'
```

src/core/services/interaction/interaction.js
```diff
@@ -0,0 +1,68 @@
+angular
+  .module('material.core.interaction', [])
+  .service('$mdInteraction', MdInteractionService);
+
+function MdInteractionService($timeout) {
```
EladBezalel: What about documentation?

src/core/services/interaction/interaction.js
```diff
@@ -0,0 +1,68 @@
...
+
+  function onBufferInput(event) {
+    $timeout.cancel(timer);
+
+    onInput(event);
+    buffer = true;
+
+    timer = $timeout(function() {
+      buffer = false;
+    }, 1000);
```
EladBezalel: Have we talked about it yet?
Why do we need a timeout?

src/components/sidenav/sidenav.js
DevVersion: Yeah, good idea.

src/core/services/interaction/interaction.js
DevVersion: I don't think this service should be documented, it's not really generic and there aren't much use cases for a user to use that service. It's mostly a service which helps us with the accessibility for the components. But if you wan't me to document it, it won't be a problem.

src/core/services/interaction/interaction.js
DevVersion: This is needed to delay the `touchstart` event. I've added a comment for this and also changed the delay to `650`ms, because that's a appropriate value for IOS and desktop mouse.

src/core/services/interaction/interaction.js
EladBezalel: @ThomasBurleson @topherfangio any thoughts?

topherfangio: Regardless of whether we make the docs public, it is always nice to have them for ourselves :smile:

DevVersion: Agree, will do that tomorrow :+1:

DevVersion:
#7965

PR #5590, opened 2015-11-06
fix(all): fix issue with removeClass removing previous classes

mallowigi:
In certain parts of the project, a "removeClass(className)" is used to remove the class from the element. While this works fine in latest versions of jQuery, in jQuery < 1.9 this results in the element classes being removed when className is undefined, thus breaking the whole layout.

Adding a null check for each of those statements to prevent this behavior.


ThomasBurleson:
Already fixed with SHA 6dd783f3c

PR #5612, opened 2015-11-07
style(disabled components): changed cursor to default when disabled

### LINKED ISSUE #4831: Disabled radio seems clickable with pointer cursor (hand)
There is also a ripple that happens on click.

EladBezalel:
Replaced all `no-drop` and `not-allowed` cursors to `default`

fixes #4831

### DIFF
--- a/src/components/autocomplete/autocomplete.scss
+++ b/src/components/autocomplete/autocomplete.scss
@@ -42,7 +42,7 @@ md-autocomplete {
   min-width: 190px;
   &[disabled] {
     input {
-      cursor: not-allowed;
+      cursor: default;
     }
   }
   &[md-floating-label] {
--- a/src/components/button/button-theme.scss
+++ b/src/components/button/button-theme.scss
@@ -142,7 +142,7 @@ a.md-button.md-THEME_NAME-theme,
   &.md-accent[disabled],
   &.md-warn[disabled] {
     color: '{{foreground-3}}' !important;
-    cursor: not-allowed;
+    cursor: default;
 
     md-icon {
       color: '{{foreground-3}}';
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -121,7 +121,7 @@ md-checkbox {
 
   // disabled
   &[disabled] {
-    cursor: no-drop;
+    cursor: default;
   }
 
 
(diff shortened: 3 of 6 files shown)


EladBezalel:
@ThomasBurleson before merging this, we should consult the UX team about replacing the `no-drop` cursor with `default`

gkalpak:
Why change it to `default` (and how is it related to #4831 ?
I prefer `not-allowed` (but #4831 still needs to be fixed).

EladBezalel:
@gkalpak this is fixing #4831 by adding a cursor style 

Take a look [here](http://codepen.io/EladBezalel/pen/epbXbZ), this is disabled native components codepen, as you can see the cursor is default

gkalpak:
Yes, adding a cursor style for disabled radios (or other disabled controls, e.g. select, switch) lgtm. But it doesn't have to be `default`, right ?

`not-allowed` conveys the "disabled-ness" (not a real word, I know) better imo, so why not use that ?

EladBezalel:
@gkalpak this is why it's labeled as team discussion :smile: 

IMO our components should behave as the native components. 

@jelbourn @ThomasBurleson @robertmesserle any thoughts?

gkalpak:
@EladBezalel, that's why we're discussing :smiley:
 What I wasn't sure about (and that's why I adked), is what's the rational behind changing `not-allowed` to `default` (and how/if it relates to fixing #4831).

So, if I have understood correctly, there are 2 different things in this PR:

1. Add some missing cursor styles for disabled radios, selects and switches (which fixes #4831).

2. Change the cursor style for disabled controls to `default` (instead of `not-allowed`).

--
The former totally lgtm. The latter (which could go in a different PR imo) is what needs discussion, I guess :smile:

Being consistent with the native controls makes sense. On the other hand, we are putting some considerable effort into looking _much better_ than the native controls :grin:
Since I wouldn't expect `not-allowed` to cause any confusion to users, I slightly favor it over `default`. (I admit it can appear as too "bold" in some occasions, but according to my experience they are the minority and the styles can be overwritten on a case-by-case basis.)

All in all, I am fine with either alternative, as long as the styles are consistent across controls.

So, let's hear what others have to say :ear:

SanderElias:
I'm in favor for the default cursor option, for 2 reasons.
1. it's in line with native controls
2. when a user is just moving his mouse around, it is way less distracting.

jelbourn:
@EladBezalel Yes- the cursor should absolutely, 100% be `default`.  LGTM.

EladBezalel:
@ThomasBurleson rebased and ready to be merged

PR #5619, opened 2015-11-08
fix(select): disabled option no longer reacting to hover and closing on click

### LINKED ISSUE #4967: Disabled md-options still ripple and close md-select on click
I'd expect `<md-option>`s that have been disabled (with a `disabled` or  `ng-disabled` attribute) not to ripple or close the select dropdown on click.  However, sometimes disabled options still animate on click, still show a pointer cursor on hover, and still close on click.

It seems that some browsers result in animation/close and some do not.  Though even in browsers that do not animate, the background of disabled options changes.

See: http://codepen.io/anon/pen/bVwOWG

EladBezalel:
Added a not disabled check before closing menu,
Now when clicking the menu anywhere but a valid option the menu closing will be prevented

fixes #4967

### DIFF
--- a/src/components/select/select-theme.scss
+++ b/src/components/select/select-theme.scss
@@ -64,7 +64,7 @@ md-select-menu.md-THEME_NAME-theme {
       }
     }
   }
-  md-option:focus:not([selected]) {
+  md-option:focus:not([disabled]):not([selected]) {
     background: '{{background-200}}';
   }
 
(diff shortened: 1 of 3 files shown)


DevVersion:
What happened to the test? It failed

EladBezalel:
@DevVersion thanks saw it, just fixed

DevVersion:
Perfect :+1: Test looks good now :smile:

PR #5620, opened 2015-11-08
style(listItem): removed unnecessary button background color and hover background color

### LINKED ISSUE #2922: flat md-button no hover effect inside md-list
md-button flat style has hover effect in the official demo
https://material.angularjs.org/#/demo/material.components.button

but is not show up while inside md-list (background is while)

EladBezalel:
fixes #2922


EladBezalel:
@marcysutton please review

PR #5636, opened 2015-11-09
fix(textarea): text area auto grow fix

### LINKED ISSUE #5627: Rows attribute breaks textarea resizing
I use the rows attribute to show my users that it is a textarea and not a single line text input and as of v. 1.0.0rc3 this no longer works.

[Here](http://codepen.io/chr22/pen/dYgabY) is a codepen showing the issue. It is a fork of your own example, where I added a rows="5" to the "biography" textarea in the bottom of the page.

DevVersion:
Textarea should grow automatically if no `rows` attribute is specified.
Textarea should stay fixed with given `rows` attribute (no growing or shrinking).

If there are other thinks of the textarea, not wished as I fixed it, please tell me.
I'm always up for improving my PR through suggestions or wishes.

Fixes #5627

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -322,7 +322,9 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
             node.style.minHeight = null;
           }
 
-          var rows = Math.max(min_rows, Math.round(node.scrollHeight / lineHeight));
+          var rows = Math.min(min_rows, Math.round(node.scrollHeight / lineHeight));
+          node.style.height = lineHeight * rows + "px";
+
           node.setAttribute("rows", rows);
         }
 


src/components/input/input.js
```diff
@@ -322,7 +322,9 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
             node.style.minHeight = null;
           }
 
-          var rows = Math.max(min_rows, Math.round(node.scrollHeight / lineHeight));
+          var rows = Math.min(min_rows, Math.round(node.scrollHeight / lineHeight));
+          node.style.height = lineHeight * rows + "px";
+
           node.setAttribute("rows", rows);
```
ThomasBurleson: I am confused by your conversation comments. I thought you said the "rows" attribute should be **removed** to fix auto-grow ?

DevVersion: No, if the attribute ```rows``` is not specified, the rows should get a dynamic value of rows, which auto grows and auto shrinks. If there is the ````rows``` attribute specified the textarea should not grow and shrink, it should stay fixed at the specified amount of rows

IGx89: Why did you change Math.max to Math.min? I have my textarea set to rows="3" and want it to stay at a minimum of 3 rows, growing if necessary. With your Math.min, the "min_rows" variable (and hence the "rows" attribute) basically becomes "max_rows", which seems unintended.

DevVersion: Thanks for having a look at this things :smile:

PR #5654, opened 2015-11-10
fix(tooltip): set initial position of tooltip

### LINKED ISSUE #4345: tooltip location slightly off on first hover
md-tooltip is slightly not centered when first hovered, but succeeding hover action does correct itself.

After app load, hover over to button with tooltip shows ...

not-centered

After hovering 2nd time and succeeding times, tooltip centered by itself ...

centered

lelanophilus:
Set initial position of the tooltip so that its inclusion at the bottom
of the page does not cause the scroll bar to appear and offset the
calculation of the tooltip's final position.

Fixes #4345

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -175,8 +175,9 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
     }
 
     function showTooltip() {
-      // Insert the element before positioning it, so we can get the position
+      // Insert the element and position at top left, so we can get the position
       // and check if we should display it
+      element.css({top: 0, left: 0});
       tooltipParent.append(element);
 
       // Check if we should display it or not.


lelanophilus:
I singed it!

jelbourn:
@robertmesserle mentioned that he thought this change might not work in all cases. Rob, what was the problem?

ThomasBurleson:
@SanderElias - can you review this plz?

SanderElias:
@ThomasBurleson, If @robertmesserle has more input this, I would love to hear it. I din't find issues with this.

robertmesserle:
This looks good to me, I think I was thinking of a different issue when talking to Jeremy.

jelbourn:
LGTM based comments from Rob and Sander.

lelanophilus:
What is the status on merging? When is 1.0-rc4 supposed to be cut?

src/components/tooltip/tooltip.js
```diff
@@ -175,8 +175,9 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
     }
 
     function showTooltip() {
-      // Insert the element before positioning it, so we can get the position
+      // Insert the element and position at top left, so we can get the position
       // and check if we should display it
+      element.css({top: 0, left: 0});
```
ThomasBurleson: *  Why is this necessary to check computed style 'display' ?
*  Would you not want to init the position **off-screen**?
  *  eg  `element.css({top:-100, left:0})`

lelanophilus: It fixes a bug where the tooltip being displayed at its default location causes the scrollbar to appear and then the positioning of the tooltip is off when the tooltip moves and the scrollbar disappears. The bug: https://github.com/angular/material/issues/4345 

element.css({top:-100, left:0}) also works. I don't have a preference since either way works just as well.

SanderElias: As @lelanophilus noted, this prevents toolbars.

lelanophilus: What is the action item right now? I'd like to fix this bug but it isn't clear to me what steps I should be taking. Thanks!

EladBezalel:
LGTM

PR #5655, opened 2015-11-10
fix(menu): use light menu item texts in a dark theme

remcohaszing:
The menu item texts were badly readable, because they were black on a
dark background.


EladBezalel:
this is not a complete solution. please take a look at #5744

PR #5676, opened 2015-11-11
fix(mdUtil): disableBodyScroll no longer scrolls page to top in IE

### LINKED ISSUE #3627: md-select page jump in IE10/11
http://codepen.io/anon/pen/EjERWg

I've wrapped the form area in between two long pieces of text so it's easier to see the issue. Here are the two behaviors I've observed:

1) If you click on the select then click outside to exit, it jumps to top of page
2) if you click on the select and select an item, pages to focus on bottom of select

### LINKED ISSUE #4640: md-select causes page to jump when an option is clicked, only in IE
md-select causes page to jump when an option is clicked, only in IE. Works fine in chrome and FF. But trying in IE11, it does this.
Any workarounds?

### LINKED ISSUE #5334: $mddialog close() methods scrolls page to top on IE11
Hello all:

when calling to the $mdDialog.cancel() or $mdDialog.hide() methods, my page scrolls to the top.

This works well in Chrome though as it closes the dialog and stays where it should. 

However in ie11, you close it and the page scrolls to the top; I have tried to hide the scrollbar and return a promise but seems that is capable of scrolling even with the scrollbar hidden

Patrik-Lundqvist:
IE needs to set scrollTop on html element in order to not scroll the page to the top

Fixes #4640
Fixes #5334
Fixes #3627

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -252,6 +252,7 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
           body.setAttribute('style', restoreBodyStyle);
           htmlNode.setAttribute('style', restoreHtmlStyle);
           body.scrollTop = scrollOffset;
+          htmlNode.scrollTop = scrollOffset;
         };
       }
 


rschmukler:
Merged, thank you!

PR #5677, opened 2015-11-11
fix(mdUtil): fix IE rendering bug when body scroll is disabled

### LINKED ISSUE #5650: $mdUtil: disableBodyScroll causes IE rendering bug
When using components that hides the scrollbar, IE will not render anything under the body (which  is fixed to the top). Open [this plunker](http://embed.plnkr.co/L5pZtk) in IE and scroll all the way down, try to select an option a few times and you will see a rendering bug at the bottom of the page. IE either displays the html element totally black or a weird duplicate rendering of the body.

Recording of the bug (the black background should not be there)

Patrik-Lundqvist:
Applying auto height to the body when disabling the scrollbar fixes rendering issue in IE.

Fixes #5650

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -238,6 +238,7 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
           applyStyles(body, {
             position: 'fixed',
             width: '100%',
+            height: 'auto',
             top: -scrollOffset + 'px'
           });
 


Magador:
It should also fix #5690

ThomasBurleson:
This change interferes with the following demo layouts:

*  Menu,
*  ProgressLinear
*  Checkbox
*  SubHeader
*  Toast
*  Toolbar

Can you include fixes for those demos with this PR ?

ThomasBurleson:
@topherfangio - can you test this on IE? And on Mac browsers. It seems to work without side affects: on Mac Chrome and Safari.

EladBezalel:
@topherfangio please take a look

topherfangio:
In IE10, if you open the Autocomplete demos page and click on any autocomplete, it causes the height of the body to stop at the autocomplete and chop off everything below it. See image below.

height-auto-issue

topherfangio:
Sorry; I meant IE11, not IE10 :smile:

Magador:
I do not have any problem on my own.

no-bug-autocomplete-angular

topherfangio:
Apologies, I must not have done a hard refresh. That said, after a hard refresh, I am seeing a different issue which appears in your image above as well. There is a double scrollbar on the right (it appears only when the autocomplete is open).

double-scrollbar-issue

Magador:
It is because [`disableBodyScroll`](https://github.com/angular/material/blob/master/src/core/util/util.js#L235) set an `overflow-x: scroll` on `<html>`.

topherfangio:
@Magador Can you try changing that to `overflowY: auto` and see if it makes a difference?

Magador:
Changing it to `auto` seems to make disappear the disabled scrollbar in IE 11.

no-bug-autocomplete-angular-2

topherfangio:
@Magador I'm so sorry; I didn't realize that you weren't the original poster! Thanks for checking that out :smile:

@Patrik-Lundqvist Can you update this PR with the change mentioned above (`overflowY: auto`) and see if it causes any issues.

ThomasBurleson:
Closing due to lack of response and high-risk with these changes.

PR #5680, opened 2015-11-11
fix(autocomplete): adjusts dropdown position for standard input style

### LINKED ISSUE #5558: Wrong position of autocomplete loading bar
Loading bar position is wrong (large space between input and loading bar)

(Just write 2 letters, and you'll see that loading bar position is wrong)
http://codepen.io/anon/pen/bVxpzP

robertmesserle:
Closes #5558 

@ThomasBurleson @jelbourn

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -2,9 +2,10 @@ angular
     .module('material.components.autocomplete')
     .controller('MdAutocompleteCtrl', MdAutocompleteCtrl);
 
-var ITEM_HEIGHT  = 41,
-    MAX_HEIGHT   = 5.5 * ITEM_HEIGHT,
-    MENU_PADDING = 8;
+var ITEM_HEIGHT   = 41,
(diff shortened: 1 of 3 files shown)


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -84,11 +84,18 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
-          minWidth: width + 'px',
-          maxWidth: Math.max(hrect.right - root.left, root.right - hrect.left) - MENU_PADDING + 'px'
-        };
+        styles;
+    // If using `md-input-container` layout, push dropdown up to cover `ng-messages` area
+    if ($attrs.mdFloatingLabel) {
+      left += 2;
+      top -= 18;
+      width -= 4;
+    }
```
jelbourn: So much magic number.

robertmesserle: I agree, but I'm not sure how else to handle this.  This is reliant on CSS variables that are not available to JS.  Should I add a comment explaining further where these numbers came from?

PR #5687, opened 2015-11-12
make sure resize event gets detached on $destroy

### LINKED ISSUE #5688: tooltip window.event handler leaked.
the tooltip binds an `resize` event handler to `window`. It never takes it off, and this will pile up during runtime, causing lag and/or memory issues in the long run.

fixed in pr #5687

SanderElias:
While working on tooltip I saw it was leaking event bindings. this PR fixes that.

Fixes #5688

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -127,9 +127,12 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
         elementFocusedOnWindowBlur = document.activeElement === parent[0];
       };
       var elementFocusedOnWindowBlur = false;
+
       ngWindow.on('blur', windowBlurHandler);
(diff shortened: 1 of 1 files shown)


gkalpak:
You should add `Fixes #5688` at the bottom of the commit message, so that the related issue will be automatically closed when/if this one gets merged.
(Then you also don't have to mention this PR in the issue, as GitHub will do it for you :smiley:)

gkalpak:
Other than that it LGTM :)

SanderElias:
@gkalpak, ok, It must be me, but how can I change the commit message? 
(also I prepared the PR before the issue.. next time I'll better do it the other way around... )

gkalpak:
> next time I'll better do it the other way around

Yeah, that helps :stuck_out_tongue: 

For changing the commit message, you can amend the commit and then force-push to your branch. GitHub will automatically update the PR.

PR #5699, opened 2015-11-12
fix(tabs): rename `disabled` to the right `ng-disabled`

### LINKED ISSUE #5691: Tab is not disabled
Tab is not disabled when "disable" attribute is used as mention in the documentation. Here is the codepen where tab2 and tab3 are set disabled="true"

http://codepen.io/sadiqkhoja/pen/MaZYMV

DevVersion:
In the documentation the wrong attribute ```disabled``` is present, it should be ```ng-disabled```

Fixes #5691

### DIFF
--- a/src/components/tabs/js/tabDirective.js
+++ b/src/components/tabs/js/tabDirective.js
@@ -20,7 +20,7 @@
  * be initiated via data binding changes, programmatic invocation, or user gestures.
  *
  * @param {string=} label Optional attribute to specify a simple string as the tab label
- * @param {boolean=} disabled If present, disabled tab selection.
+ * @param {boolean=} ng-disabled If present, disabled tab selection.
  * @param {expression=} md-on-deselect Expression to be evaluated after the tab has been de-selected.
  * @param {expression=} md-on-select Expression to be evaluated after the tab has been selected.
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@robertmesserle - is this correct? I think the docs should (a) reference the `ng-disabled` attribute **and** (b) the expression must evaluate to **true** ?

robertmesserle:
@ThomasBurleson The proper attribute is `disabled`.  `ng-disabled` is just a way to set `disabled` through a JS value.

DevVersion:
@robertmesserle Please take a look at this issue #5691.

robertmesserle:
@DevVersion @ThomasBurleson Hm, yea, this may have changed at some point without being updated in the code.  Let me look into the history to try and figure out why it was changed, but it's currently looking for `ngDisabled` and not `disabled`.

robertmesserle:
@DevVersion @ThomasBurleson Okay, not finding any reason in the history, but I'm fine with this being merged in.  The way it's written now, it should be `ng-disabled`

DevVersion:
@robertmesserle Thanks for rechecking :)

DevVersion:
Thanks for reviewing. Can some please change the labels @ThomasBurleson @robertmesserle @EladBezalel

PR #5703, opened 2015-11-12
fix(datepicker): Icon jumping upon open/close.

### LINKED ISSUE #4570: Input - demo tweaks
* animated input fields should never move down. the label is the only thing that should move, up. noticed in the Icons section at the bottom. inputs should always be contained within a set row height, when they animate, they should stay within that row height so that nothing else on the page moves.  
* weird how titles animate when you arrive on this page.
* date picker never fits on screen, icon on date picker shouldn't move.

screen shot 2015-09-11 at 8 43 09 am

topherfangio:
When the user opened/closed the datepicker in the inputs demo, the icon would jump to the left. This was caused by two separate issues:

1. The demo CSS overrode only the closed styles.
2. The browser would scroll the parent datepicker to the left upon
   focusing the down-triangle button so that it could be visible.
   You can easily reproduce this by tabbing to the down-triangle
   then checking the `<md-datepicker>`s `scrollLeft` style.

Fix by removing unnecessary demo CSS and adding proper padding/margin to the datepicker so that the down-triangle has room to appear.

Fixes #4570.

### DIFF
--- a/src/components/datepicker/datePicker.scss
+++ b/src/components/datepicker/datePicker.scss
@@ -2,11 +2,16 @@
 $md-datepicker-button-gap: 12px;  // Space between the text input and the calendar-icon button.
 $md-datepicker-border-bottom-gap: 5px;  // Space between input and the grey underline.
 $md-datepicker-open-animation-duration: 0.2s;
+$md-datepicker-triangle-button-width: 36px;
 
 md-datepicker {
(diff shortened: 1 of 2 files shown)


topherfangio:
@jelbourn Can you please take a quick look and review?

jelbourn:
LGTM

PR #5711, opened 2015-11-13
fix(tabs): fix last tab wrapping under tab bar

### LINKED ISSUE #5439: md-pagination-wrapper width is calculated incorrectly when parent container uses flex
I've encountered a problem whereby the `md-pagination-wrapper` is calculated incorrectly when the `md-content` parent has a `flex` assigned to it.

I've forked your example to show this defect:

http://codepen.io/anon/pen/RWJryx

I've also attached a screenshot showing how the wrapper forces a tab element to wrap to the next line.

screen shot 2015-10-29 at 4 12 00 pm

xiphiaz:
Fixes #5439

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -540,7 +540,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
   }
 
   function updatePagingWidth() {
-    var width = 1;
+    var width = 2;
     angular.forEach(getElements().dummies, function (element) { width += element.offsetWidth; });
     angular.element(elements.paging).css('width', width + 'px');
   }


xiphiaz:
^ I signed it!

ThomasBurleson:
@robertmesserle - can you review please ?

robertmesserle:
@ThomasBurleson Yea, this should be fine.

alexandrebrasil:
Simply adding 1px to the width won't solve it for every situation. For example:

1. Go here: http://codepen.io/anon/pen/RWJryx
2. Add 4 tabs with label "LK" (no quotes).
3. The 4th tab is sent to a new line. If you use Chrome, inspect element, go to md-pagination-element container and add 1px to the width inline style. This resolves the issue and the 4th LK tab is shown again.
4. Add 3 more "LK" tabs and manually increase the width by 1px again. The 7th LK tab is still hiddeng because it was kept on the next (paging.width < sum(label.width)). If you add 2px instead of 1px, the tab is shown.
5. If you keep iterating, you'll need to add 3px, 4px and so on...

As far as I can tell, this is happening because the CSS widths of the labels have a fraction, while the element.offsetWidth used to calculate the container width is an integer (so they differ because of rounding/truncation). 

See #5794 and PR #5801 for more details.

xiphiaz:
Closing PR in favour of #5801 as demonstrated by @alexandrebrasil

PR #5716, opened 2015-11-13
fix(list): Copy md-icon.md-secondary attributes to button.

### LINKED ISSUE #3356: ng-disabled on md-icon in md-list-item does not work
When ng-disabled is applied on md-icon acting as a secondary on md-list-item, it does not have any effect on the dynamically generated button which still fires.

Here is a codepen: http://codepen.io/anon/pen/bdoqLb

topherfangio:
When an secondary md-icon was used in combination with a checkbox, the md-icon was only copying the `ng-click` attribute instead of all attributes including things like `ng-disabled`.

Fix by ensuring we copy all relevant attributes to the wrapper.

Fixes #3356.


topherfangio:
@ThomasBurleson This is has been rebased and is ready for review/merge.

PR #5717, opened 2015-11-13
fix(dialog): break `content` into `textContent` and `htmlContent` to help keep users from accidentally introducing an XSS vector.

jelbourn:
**BREAKING CHANGE**

For $mdDialog's `alert` and `confirm` presets, splits `content` into `textContent` and `htmlContent`. Using `htmlContent` requires `ngSanitize` to be loaded and uses `ng-bind-html`, which sanitizes the html. HTML content is no longer run through `$compile` (and thus isn't treated as an angular template, just static html).


robertmesserle:
LGTM

src/components/dialog/demoBasicUsage/script.js
```diff
@@ -23,7 +23,7 @@ angular.module('dialogDemo1', ['ngMaterial'])
     // Appending dialog to document.body to cover sidenav in docs app
     var confirm = $mdDialog.confirm()
           .title('Would you like to delete your debt?')
-          .content('All of the banks have agreed to <span class="debt-be-gone">forgive</span> you your debts.')
+          .textContent('All of the banks have agreed to forgive you your debts.')
```
ThomasBurleson: @jelbourn - This needs to remain as `.htmlContent()` to demonstrate html content usages to users.

src/components/dialog/dialog.js
```diff
@@ -495,7 +505,18 @@ function MdDialogProvider($$interimElementProvider) {
     function onShow(scope, element, options, controller) {
       angular.element($document[0].body).addClass('md-dialog-is-showing');
 
-      wrapSimpleContent();
+      if (controller) {
+        controller.mdHtmlContent = controller.htmlContent || options.htmlContent || '';
+        controller.mdTextContent = controller.textContent || options.textContent || '';
+
+        if (controller.mdHtmlContent && !$injector.has('$sanitize')) {
+          throw Error('The ngSanitize module must be loaded in order to use htmlContent.');
```
ThomasBurleson: @jelbourn - Since dialogs are instantiated via interimElement, I believe `throw Error` never surfaces as expected. I think we should:

* make sure the exception is captured and logged in a $log 
* manually log the error also.

src/components/dialog/dialog.spec.js
```diff
@@ -243,7 +244,7 @@ describe('$mdDialog', function() {
           ok: 'Next',
           cancel: 'Back',
           title: 'Which Way ',
-          content: '<div class="mine">Choose</div>'
+          template: '<div class="mine">Choose</div>'
```
ThomasBurleson: @jelbourn - this checks the `template` option... but what about the `htmlContent` option

src/components/dialog/dialog.spec.js
```diff
@@ -255,7 +256,7 @@ describe('$mdDialog', function() {
       expect(content.text()).toBe('Choose');
     }));
 
-    it('shows a basic confirm dialog with HTML content using custom types', inject(function($rootScope, $mdDialog, $animate) {
+    it('should NOT allow custom elements in confirm htmlContent', inject(function($mdDialog) {
```
ThomasBurleson: What does this mean?

src/components/dialog/dialog.spec.js
```diff
@@ -1256,14 +1278,39 @@ describe('$mdDialog with custom interpolation symbols', function() {
...
+    var dialog = $mdDialog.
+      alert({parent: parent}).
+      title('Title').
+      htmlContent('Hello, world !').
+      ok('OK');
+
+    expect(function() {
+      $mdDialog.show(dialog);
+      $rootScope.$digest();
+    }).toThrowError(/unsafe/);
```
ThomasBurleson: Who throws the `/unsafe/` ?

src/components/dialog/demoBasicUsage/script.js
jelbourn: I did this to discourage use of this API. People should almost always use textContent for Ashley and confirm.

src/components/dialog/dialog.spec.js
```diff
@@ -264,16 +265,37 @@ describe('$mdDialog', function() {
           ok: 'Next',
           cancel: 'Back',
           title: 'Which Way ',
-          content: '<my-content class="mine">Choose</my-content>'
+          htmlContent: '<my-content class="mine">Choose</my-content> breakfast'
```
gkalpak: `class="mine"` seems unnecessary now and `breakfast` doesn't go well with the title...

src/components/dialog/dialog.spec.js
```diff
@@ -264,16 +265,37 @@ describe('$mdDialog', function() {
...
 
       var container = angular.element(parent[0].querySelector('.md-dialog-container'));
-      var content = angular.element(container[0].querySelector('.mine'));
+      var contentBody = container[0].querySelector('.md-dialog-content-body');
 
-      expect(content.text()).toBe('Choose');
+      expect(contentBody.textContent).toBe('Choose breakfast');
+    }));
+
+    it('should NOT evaluate angular templates in confirm htmlContent', inject(function($mdDialog) {
```
gkalpak: `confirm` ??

src/components/dialog/dialog.js
```diff
@@ -431,7 +437,11 @@ function MdDialogProvider($$interimElementProvider) {
         '<md-dialog md-theme="{{ dialog.theme }}" aria-label="{{ dialog.ariaLabel }}" ng-class="dialog.css">',
         '  <md-dialog-content class="md-dialog-content" role="document" tabIndex="-1">',
         '    <h2 class="md-title">{{ dialog.title }}</h2>',
-        '    <div class="md-dialog-content-body" md-template="::dialog.mdContent"></div>',
+        '    <div ng-if="dialog.mdHtmlContent" class="md-dialog-content-body" ',
+        '        ng-bind-html="dialog.mdHtmlContent"></div>',
+        '    <div ng-if="!dialog.mdHtmlContent" class="md-dialog-content-body">',
+        '      {{dialog.mdTextContent}}',
+        '    </div>',
```
gkalpak: Not a big deal (since usually there is max one dialog at any given moment), but using `ngSwitch` and one-time-bindings should be (unnoticably) better for performance.

src/components/dialog/dialog.js
```diff
@@ -294,10 +294,12 @@ function MdDialogDirective($$rAF, $mdTheming, $mdDialog) {
  *
  * @returns {obj} an `$mdDialogPreset` with the chainable configuration methods:
  *
- * - $mdDialogPreset#title(string) - sets title to string
- * - $mdDialogPreset#content(string) - sets content / message to string
- * - $mdDialogPreset#ok(string) - sets okay button text to string
- * - $mdDialogPreset#theme(string) - sets the theme of the dialog
+ * - $mdDialogPreset#title(string) - Sets the alert title.
+ * - $mdDialogPreset#textContent(string) - Sets the alert message.
+ * - $mdDialogPreset#htmlContent(string) - Sets the alert message as HTML. Requires ngSanitize
```
topherfangio: Is there a specific reason we aren't running it through the compiler? I can see a use-case for allowing custom directives; especially things like `<md-icon>`.

src/components/dialog/dialog.js
```diff
@@ -548,27 +569,6 @@ function MdDialogProvider($$interimElementProvider) {
           return angular.element(closeButton);
         }
       }
-
```
topherfangio: Did we replace this with some similar functionality so that the text would be wrapped with `<p>` tags? It looks like the paragraph tags are styled slightly differently than regular `<div>` contents, so this may cause some styling issues.

src/components/dialog/dialog.js
jelbourn: You're right that the errors here are caught and do nothing. This is too severe for `$log`, though. I've changed the code so that the errors really get thrown now.

src/components/dialog/dialog.spec.js
jelbourn: Yes, it should have been `htmlContent`

src/components/dialog/dialog.spec.js
jelbourn: It means custom elements are not allowed in the confirm's htmlContent?

src/components/dialog/dialog.spec.js
jelbourn: I will always choose breakfast as my way.

src/components/dialog/dialog.spec.js
jelbourn: It's for the `confirm` preset.

src/components/dialog/dialog.js
jelbourn: I meant to make them all one-time bindings before I pushed but forgot. Fixed.

src/components/dialog/dialog.js
jelbourn: `$sanitize` does not remove stuff like interpolation symbols. This means an unwary user could do:
```js
var e = document.createElement('p')
...
e.textContent(userName);
var htmlContent = e.innerHTML;
```

Then, a user could set their username to `{{ runEvilThing() }}` and it would get compiled and run.

src/components/dialog/dialog.js
jelbourn: Good point. Added the `<p>` back around the textContent.

ThomasBurleson: @topherfangio  - @jelbourn asked me about the `<p>` wrappers and I could not recall why I did that originally. LOL

PR #5728, opened 2015-11-14
fix(tooltip): change default delay from zero to 300 as referenced in …

### LINKED ISSUE #1847: md-tooltip does not default to 400ms delay
The docs [here](https://material.angularjs.org/#/api/material.components.tooltip/directive/mdTooltip) indicate that the default tooltip delay is 400ms: "How many milliseconds to wait to show the tooltip after the user focuses, hovers, or touches the parent. Defaults to 400ms."

The problem is that there does not appear to be any default delay. The tooltips are displayed instantly.

Notice the difference in the code pen. The 'working' tooltips have manually set `md-delay="400"` but the non working tooltips are just using the default. They should be showing the same delay, but you can see that the non working tooltips are much faster (instant).

http://codepen.io/Splaktar/pen/emLbQN

Codepen example is using 0.8.3 with Angular 1.3.14.

DevVersion:
…the docs

https://material.angularjs.org/latest/api/directive/mdTooltip says: defaults to 300ms

Fixes #1847

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -35,7 +35,7 @@ angular
 function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdTheming, $rootElement,
                             $animate, $q) {
 
-  var TOOLTIP_SHOW_DELAY = 0;
+  var TOOLTIP_SHOW_DELAY = 300;
   var TOOLTIP_WINDOW_EDGE_SPACE = 8;
 
   return {


EladBezalel:
Actually, i removed the delay, although i didn't changed the docs, I didn't found any delay described on the Material Spec

DevVersion:
I created a second PR, which changes the value of the documentation, so you only need to decide :+1:

Splaktar:
Any explanation why this was closed?

EladBezalel:
@Splaktar we're going with this #5748

Splaktar:
OK, good to know. Thanks for linking the issues.

PR #5607, opened 2015-11-07
feat(card): improved to behave as spec

EladBezalel:
Added all functionality as seen in:
http://www.google.com/design/spec/components/cards.html#cards-content-blocks

fixes #1900

PR #5729, opened 2015-11-14
docs(card): improve image quality to the computed width of the card

DevVersion:
The currently displaying card image isn't very sharp, if it become as sharp as now the component would be better represented.

Current: https://gyazo.com/daea22f3c151297df8e9ece0380998f5
After: https://gyazo.com/ac9d8fc2c6943598bc605ff0747ac325

If that was done by intention, excuse me ;)


EladBezalel:
@DevVersion It's actually doesn't change a lot as my PR (#5607) is about to be merged, it changes the Cards components and now the preview will be much more relevant and the image will fit :)

DevVersion:
Okay then we should close this commit right? :smile:

PR #5739, opened 2015-11-15
Use scroll position of parent element for the calculation of the client rect to fix tooltip positioning in a scrolled md-content container.

### LINKED ISSUE #2791: md-tooltip does not position correctly with vertical scroll
Hi,

in my app my md-content has vertical scroll. When I hover over button to show tooltip and scroll has position 0, then tooltip appears in correct place. But when I scroll down, tooltip does not follow the scroll. Is appears over the button. Is it a bug or do I have something wrong with my layout? (I use flexbox...). Or maybe you could help me with telling how tooltip positions itself?

Thanks,
Paweł

r0b-:
Should fix #2791

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -43,17 +43,21 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
 
     clientRect: function(element, offsetParent, isOffsetRect) {
       var node = getNode(element);
-      offsetParent = getNode(offsetParent || node.offsetParent || document.body);
       var nodeRect = node.getBoundingClientRect();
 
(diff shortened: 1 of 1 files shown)


r0b-:
Signed the CLA!

SanderElias:
This will not work if the parent is of type `flex: 1; overflow: auto`. And in a couple of another scenarios. There are also still some flex-box differences between browsers that interfere with this line of solution,

I have a PR in the work that will solve this in all situations.

r0b-:
Sounds good - can't wait to get the fix as tooltips are not really usable right know...
If your PR is not ready for the next release, maybe this code can be merged to have (at least) the md-content use case working?!

PR #5747, opened 2015-11-16
fix(tooltip): make sure resize event detaches from window.

SanderElias:
fix(tooltip): make sure resize event detaches from window.

fixes: #5688


ThomasBurleson:
@SanderElias  - please rebase.

PR #5757, opened 2015-11-16
fix(textarea): fix textarea grow

### LINKED ISSUE #5751: RC4: Rows attribute doesn't resize textarea when no content
Textareas are always rendered with only one row if there isn't any content in it.

See this [CodePen](http://codepen.io/chr22/pen/dYgabY)

It also limits the height of the textarea to the number of rows set.

DevVersion:
Fixes Textarea Input

Fixes #5751

### DIFF
--- a/src/components/input/demoTextarea/index.html
+++ b/src/components/input/demoTextarea/index.html
@@ -1,0 +1,28 @@
+<div layout="column" ng-cloak class="md-inline-form" ng-init="textBind = 'Input is binded to a Model'">
+
+  <md-content layout-padding>
+
+    <div class="demoTextarea">
+      <md-input-container class="md-block">
+        <label>Fixed Rows</label>
+        <textarea rows="5" ng-model="textBind">This is an example Content</textarea>
+      </md-input-container>
+    </div>
+
+    <div class="demoTextarea">
+      <md-input-container class="md-block">
+        <label>Minimum Rows</label>
+        <textarea min-rows="5" ng-model="textBind"></textarea>
+      </md-input-container>
+     </div>
+
+    <div class="demoTextarea">
+      <md-input-container class="md-block">
+        <label>Dynamic Rows</label>
+        <textarea>This is an example Content</textarea>
+      </md-input-container>
+    </div>
+
+  </md-content>
+
+</div>
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -323,22 +323,27 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
       var node = element[0];
       var container = containerCtrl.element[0];
 
-      var min_rows = NaN;
+      var rows = NaN;
+      var minRows = NaN;
       var lineHeight = null;
       // can't check if height was or not explicity set,
       // so rows attribute will take precedence if present
       if (node.hasAttribute('rows')) {
-        min_rows = parseInt(node.getAttribute('rows'));
+        rows = parseInt(node.getAttribute('rows'));
+      } else if (node.hasAttribute('min-rows')) {
+        minRows = parseInt(node.getAttribute('min-rows'));
       }
 
-      var onChangeTextarea = $mdUtil.debounce(growTextarea, 1);
+      var onChangeTextarea = $mdUtil.debounce(function() {
+        $mdUtil.nextTick(growTextarea);
+      }, 1);
 
       function pipelineListener(value) {
         onChangeTextarea();
         return value;
       }
 
-      if (ngModelCtrl) {
+      if (hasNgModel) {
         ngModelCtrl.$formatters.push(pipelineListener);
         ngModelCtrl.$viewChangeListeners.push(pipelineListener);
       } else {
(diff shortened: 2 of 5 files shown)


DevVersion:
@ThomasBurleson @EladBezalel 
I created another commit (not squashed yet - by intention - ability to revert), which contains `min-rows` (Minimum Rows), `rows` (Fixed Rows) and unspecified (starting at `1`).

Can you please review that commit and give me a suggestion about adding the `min-rows` attribute. Should we use another attribute name?

morriq:
Working well on my website. :+1:

DevVersion:
**Update**
- Added complicated tests for `min-rows`, `rows` and unspecified rows attribute
- Fixed Minimum Rows if text content is exactly at minimum

src/components/input/demoTextarea/index.html
```diff
@@ -0,0 +1,22 @@
+<div layout="column" ng-cloak class="md-inline-form">
+
+  <md-content layout-padding>
```
EladBezalel: layout-padding is making the text areas look weird, wrap them with divs or make a class

src/components/input/demoTextarea/index.html
```diff
@@ -0,0 +1,22 @@
...
+
+    <md-input-container class="md-block">
+      <label>Fixed Rows</label>
+      <textarea columns="1" rows="5"></textarea>
```
EladBezalel: columns is not necessary, i couldn't find any place defining it.

src/components/input/input.js
```diff
@@ -264,12 +264,15 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
...
 
-      var min_rows = NaN;
+      var rows = NaN;
+      var minRows = NaN;
       var lineHeight = null;
       // can't check if height was or not explicity set,
       // so rows attribute will take precedence if present
       if (node.hasAttribute('rows')) {
-        min_rows = parseInt(node.getAttribute('rows'));
+        rows = parseInt(node.getAttribute('rows'));
```
EladBezalel: this is a breaking change isn't it?

DevVersion: Yes it is.

src/components/input/demoTextarea/index.html
DevVersion: I just copied that from the Basic Demo

src/components/input/demoTextarea/index.html
DevVersion: Yes, you are right, that unnecessary

DevVersion:
The most suggestions are now ready. @EladBezalel

DevVersion:
This should be fixed as soon as possible, because the current `textarea` is hopping around. @ThomasBurleson. Please take a look at that :smile:

src/components/input/demoTextarea/index.html
topherfangio: The CSS for `layout-padding` changed recently. We are aware of some issues with it, but the quick workaround is to wrap the content in a `<div>`.

morriq:
Seems there is problem with this PR.

sorry for no codepen or something. I don't know how to include changed library to sourcecode.

It's removing default value if exists. 

<md-input-container ng-init="test = '123';">
<textarea ng-model="test"></textarea>
</md-input-container>

src/components/input/input.js
```diff
@@ -287,12 +290,15 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
...
-      if (isNaN(min_rows)) {
-        element.attr('rows', '1');
+      if (isNaN(rows)) {
+        element.attr('rows', isNaN(minRows) ? '1' : minRows);
 
         element.on('scroll', onScroll);
       }
 
+      // Trigger the input event to compute style
+      element.triggerHandler('input');
```
morriq: this line causing problems

DevVersion:
All problems are solved @morriq 
Thanks for keeping an eye on that :)

src/components/input/input.js
```diff
@@ -381,8 +387,22 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
...
+          if (!isNaN(minRows) && contentHeight) {
+            var calcHeight;
+            var minHeight = lineHeight * minRows;
+
+            if (contentHeight <= minHeight) calcHeight = minHeight;
+            else calcHeight = minHeight + (contentHeight - minHeight);
+
+            node.style.height = calcHeight + "px";
+          } else if (contentHeight) {
+            node.style.height = contentHeight + 'px';
```
EladBezalel: i'd expect a variable to save the height into and applying `node.style.height =` only once for more readability

EladBezalel:
Hey @DevVersion, because we're past 1.0 we try our best not to introduce breaking changes.
We should make this PR without any breaking API, if you will need my help ping me.

DevVersion:
@EladBezalel Okay then I try to remove the breaking change, and I will take a look in #5919.

But with the attribute rows we cant define fixed row, how was that managed before?

DevVersion:
@EladBezalel Just thought a bit about the breaking change. I don't think it such a huge breaking change, it's just adding the attribute `min-rows` to allow growing rows. 

Little Explanation: 
- `rows="5"` - Fixed 5 Rows
- `min-rows="5"` - Minimum 5 Rows (can grow)

I think it's not a big change, and I think it also needed to provide this new attribute to allow the users to decide how the `<textarea>` should behave. Let me know what you think? And what are you thinking @ThomasBurleson?

EladBezalel:
@ThomasBurleson assigned to you not because it's ready, please review.

alexthewilde:
+1

EladBezalel:
As this is not the complete and desired behavior of textarea, im closing this one,
@crisbeto can you please open an issue about our new documented textarea and note there our concerns and wanted behavior?

PR #5774, opened 2015-11-16
fix(dialog): removed no dialog actions warning

### LINKED ISSUE #5767: $mdDialog shouldn't emit a warning when there are no md-dialog-actions
It is a valid use-case to have a dialog that you only close programmatically based on some user action. As such, there shouldn't be a warning when there are no `md-dialog-actions` present.

@EladBezalel is this something you could look at?

EladBezalel:
It is a valid use-case to have a dialog that you only close programmatically based on some user action. As such, there shouldn't be a warning when there are no md-dialog-actions present.

fixes #5767


jelbourn:
LGTM

PR #5777, opened 2015-11-17
fix(interimElement): added isRemoved indication to support remove before compile finish

### LINKED ISSUE #5635: mdDialog cannot be closed before onComplete is called
Normally, you wouldn't want to close a dialog immediately after you open it, but I am using a loading dialog, so sometimes I have all the data and it should close pretty fast, but it does not always close.

This is due to the fact that I am using templateUrl, so it takes some time to load the template, and during this time $mdDialog.hide call is ignored

http://codepen.io/anon/pen/avRevK

EladBezalel:
Sometimes `compileElement` didn't finished before `close` has been called so the element was `transitionIn` but never out cause technically it was removed.

Now when remove is call we turn on the `isRemoved` flag and when `compileElement` will finish it won't show the element.

fixes #5635

### DIFF
--- a/src/core/services/interimElement/interimElement.js
+++ b/src/core/services/interimElement/interimElement.js
@@ -404,7 +404,7 @@ function InterimElementProvider() {
        * Used internally to manage the DOM element and related data
        */
       function InterimElement(options) {
-        var self, element, showAction = $q.when(true);
+        var self, element, isRemoved, showAction = $q.when(true);
 
         options = configureScopeAndTransitions(options);
 
@@ -421,9 +421,12 @@ function InterimElementProvider() {
          */
         function createAndTransitionIn() {
           return $q(function(resolve, reject){
-
             compileElement(options)
               .then(function( compiledData ) {
+                if (isRemoved) {
+                  return resolve();
+                }
+
                 element = linkElement( compiledData, options );
 
                 showAction = showElement(element, options, compiledData.controller)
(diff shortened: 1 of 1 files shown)


src/core/services/interimElement/interimElement.js
```diff
@@ -421,9 +421,12 @@ function InterimElementProvider() {
          */
         function createAndTransitionIn() {
           return $q(function(resolve, reject){
-
             compileElement(options)
               .then(function( compiledData ) {
+                if (isRemoved) {
```
ThomasBurleson: What about checking `scope.$$destroyed` instead?

EladBezalel: what if `options.$destroy` is [set to false](https://github.com/angular/material/blob/fix/interim-remove-before-compile-finish/src/core/services/interimElement/interimElement.js#L463)?

EladBezalel: @ThomasBurleson ping :bell:

src/core/services/interimElement/interimElement.js
```diff
@@ -404,7 +404,7 @@ function InterimElementProvider() {
        * Used internally to manage the DOM element and related data
        */
       function InterimElement(options) {
-        var self, element, showAction = $q.when(true);
+        var self, element, isRemoved, showAction = $q.when(true);
```
ThomasBurleson: @EladBezalel - should this not be initialized to true ?

ThomasBurleson:

Issue #2228, opened 2015-04-09
md-divider(enhancement): add `vertical` option


jossef:
The `md-divider` directive works great for horizontal use cases ( :+1:) . However, i would like to use it as a vertical divider as well.

This can be added as an option for `md-divider` directive:

```html
<md-divider vertical></md-divider>
```

ThomasBurleson:
The Material Design does not provide a specification for Vertical Dividers. Can you provide a mockup of your needs [and we will escalate to the UX Design team]?

Jossef:
@ThomasBurleson Thanks :smile: 

![image](https://cloud.githubusercontent.com/assets/1287098/7086908/bdeb5fca-df8e-11e4-9fa5-105e3f4c6f77.png)

ThomasBurleson:
@Jossef - What is the URL for that image ?

ThomasBurleson:
Specification at http://www.google.com/design/spec/layout/structure.html#structure-ui-regions

cdonohue:
Would it make sense to base this off of the container that the `<md-divider></md-divider>` is located in?
```
[layout=row] > md-divider {
  border-top: 0;
  border-left: 1px solid;  
  border-left-color: rgba(0, 0, 0, 0.12);
}
```

JobaDiniz:
+1

jossef:
תודה אלעד, יא זלמה (Thanks Elad, my hero)
:+1:

EladBezalel:
:heart: באהבה אחי (No problem)

Tallyb:
Was the vertical divider added? How is it used?

EladBezalel:
@Tallyb when the divider is in layout-row it will automatically be vertical

Tallyb:
Seems to be a problem with a selector. It does not work on a toolbar.

EladBezalel:
Can you provide a codepen please?

Tallyb:
http://codepen.io/Tallyb/pen/adpObe

Your divider is between the arrow icons. mine to the right of it. It is the same css as yours, just explicitly added. As noted - it is some selector issue.

EladBezalel:
@Tallyb thanks, toolbar defines `align-items: center` which make the divider shrink, do you mind making a PR adding `height: 100%` to the vertical divider?

arron21:
Hey don't forget to add support for
 `layout-gt-sm="row"`

etc etc

smurugavel:
[...]

Issue #930, opened 2014-12-12
mdGestures: swipe up and down


MathiasLM:
Currently md-swipe-left/right exists by implementing hammerjs' gestures. I see no reason not to include up and down as well. :)

jeffthompson1971:
Has swipe up and down been implemented?  Sorry I'm not following what the resolution was here.  I don't see it in the docs.

gpopovic:
+1

mackelito:
+1

PR #5781, opened 2015-11-17
feat(toast, gesture, divider): improved toast, added up/down gestures and vertical divider

### LINKED ISSUE #4583: Toast: fab positions
FAB position should only change when toast would otherwise overlay FAB. 

- swiping away a toast, it should not rotate
- worth noting in the text that on mobile it's full width and always at the bottom. 
- should be able to swipe in the direction the toast came from. So if the toast came from the bottom, you should be able to swipe down to dismiss

> doublecheck toast specks: https://spec.googleplex.com/quantum/components/snackbars-toasts.html#

screen shot 2015-09-11 at 9 03 09 am

EladBezalel:
* Added up and down gestures
* Added support in vertical divider
* Toast -
  * Fade away is not rotating anymore
  * When on mobile, toast is always full-width and on the bottom
  *  Now able to swipe in the direction the toast came from. So if the toast came from the bottom, you can swipe down to dismiss.

fixes #4583, #930 and #2228

### DIFF
--- a/src/components/divider/divider.scss
+++ b/src/components/divider/divider.scss
@@ -8,3 +8,11 @@ md-divider {
     margin-left: $baseline-grid * 10;
   }
 }
+
+.layout-row {
+  & > md-divider {
+    border-top-width: 0;
+    border-right-width: 1px;
+    border-right-style: solid;
+  }
+}
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -125,8 +125,8 @@ button.md-button::-moz-focus-inner {
     background-clip: padding-box;
     overflow: hidden;
 
-    transition: 0.2s linear;
-    transition-property: background-color, box-shadow;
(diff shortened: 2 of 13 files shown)


DevVersion:
I just want to notify you about the continuous integration error, perhaps you didn't see it :+1:

EladBezalel:
This is because i run tests on Chrome locally and not on Phantom, will fix, thanks @DevVersion!

ThomasBurleson:
@EladBezalel - please rebase.

src/components/divider/divider.scss
```diff
@@ -8,3 +8,11 @@ md-divider {
     margin-left: $baseline-grid * 10;
   }
 }
+
+.layout-row {
+  & > md-divider {
```
ThomasBurleson: Where is the demo associated with this?

PR #5793, opened 2015-11-17
feat(dialog): added fullscreen option to dialog

### LINKED ISSUE #2148: adding fullscreen option for mobile to $mdDialog
Hello,
I would like to know if you plan to include a fullscreen option to $mdDialog with some more responsive functionalities? Would be very great for using dialogs on mobile devices:

screen shot 2015-04-03 at 8 37 49 am

http://www.google.com/design/spec/components/dialogs.html#dialogs-full-screen-dialogs

Thanks for your great work!

EladBezalel:
Added `fullscreen` option that if it's set to true `.md-dialog-fullscreen` class is added to `<md-dialog>` that maximizing the dialog to the whole screen.

fixes #2148

### DIFF
--- a/src/components/dialog/demoBasicUsage/dialog1.tmpl.html
+++ b/src/components/dialog/demoBasicUsage/dialog1.tmpl.html
@@ -10,7 +10,7 @@
       </div>
     </md-toolbar>
 
-    <md-dialog-content style="max-width:800px;max-height:810px; ">
+    <md-dialog-content>
       <div class="md-dialog-content">
         <h2>Using .md-dialog-content class that sets the padding as the spec</h2>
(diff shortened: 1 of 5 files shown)


ThomasBurleson:
I will change this to be consistent with Layouts and md- attributes:  `md-fullscreen-sm`

EladBezalel:
fullscreen is only dialog feature, are you sure it should be in layouts?

ThomasBurleson:
I am sure that if you are using a breakpoint syntax then we must be consistent.

EladBezalel:
how are you going to implement it? 
`md-fullscreen-sm` is too specific.

PR #5795, opened 2015-11-17
fix(chips): remove style of chip if it blurs.

### LINKED ISSUE #5662: md-chips not leave focus when click on other input.
md-chips-blur

http://codepen.io/anon/pen/rOQpgq?editors=100

DevVersion:
After the user clicks everywhere else to deselect the chip, the selected chip variable will change in the controller but this needs to be applied to the scope

Fixes #5662

### DIFF
--- a/src/components/chips/js/chipDirective.js
+++ b/src/components/chips/js/chipDirective.js
@@ -52,7 +52,8 @@ function MdChip($mdTheming, $mdUtil) {
 
       if (ctrl) angular.element(element[0].querySelector('.md-chip-content'))
           .on('blur', function () {
-            ctrl.selectedChip = -1;
+            ctrl.resetSelectedChip();
+            ctrl.$scope.$apply();
           });
     };
   }


EladBezalel:
@DevVersion please check CI

DevVersion:
Fixed and Squashed, thanks for alerting me @EladBezalel :smile: We both do that often :D

DevVersion:
That's an easy fix, should be easy to review. Would be great if you do :smile:  @EladBezalel

topherfangio:
This definitely needs a test, and some manual testing. I have run into multiple cases in the FAB Speed Dial where calling `$scope.$apply()` during a `blur()` caused issues. This mainly happened if the user was tabbing to another item which meant that the `blur()` happened inside of an existing digest loop.

This is why I switched to `$scope.evalAsync()`: https://github.com/angular/material/blob/master/src/components/fabSpeedDial/fabController.js#L13

DevVersion:
I wan't to combine this PR with #5941, because there I already created a test section for the focus functionality.

PR #5799, opened 2015-11-18
feat(chips): add functionality to disable removing.

### LINKED ISSUE #3820: Maybe add md-chips read-and-delete-only ?
Since I have this use case right now I was wondering if it is a good idea to add this as a feature in the future. 

Basically the user may delete chips but at the same time the user should not be able to add chips by just typing stuff and pressing enter as I have a custom form for that.

### LINKED ISSUE #5796: allow disable text input md-chips
Add option to only disabled the keyboard / input aspect of `md-chips` while allowing mouse clicks on the delete button.  This would essentially become a list of filters which have been selected elsewhere, perhaps in conjunction with a very large multi select.  Clicking on `X` would de-select the filter and remove it from the model list.

DevVersion:
- By default - chips are writable **and** deletable
- `readonly="true"`- chips are readonly and not deletable
- `readonly="true" md-removable="true"` - chips are readonly, and deletable
- `readonly="true" md-removable="false"` - chips are readonly, not deletable
- `md-removable="false"`- chips are writable, not deletable

Closes #5796. Fixes #3820.

### DIFF
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -18,6 +18,8 @@ describe('<md-chips>', function() {
     '<md-chips ng-model="items" readonly="true">' +
     '  <md-autocomplete md-items="item in [\'hi\', \'ho\', \'he\']"></md-autocomplete>' +
     '</md-chips>';
+  var CHIP_NOT_REMOVABLE_TEMPLATE =
+    '<md-chips ng-model="items" readonly="true" md-removable="false"></md-chips>';
 
   afterEach(function() {
     attachedElements.forEach(function(element) {
@@ -200,6 +202,106 @@ describe('<md-chips>', function() {
         expect(scope.selectChip).toHaveBeenCalled();
         expect(scope.selectChip.calls.mostRecent().args[0]).toBe('Grape');
       });
+      
+      describe('when removable', function() {
+
+        it('should not append the input div when not removable and readonly is enabled', function() {
+          var element = buildChips(CHIP_NOT_REMOVABLE_TEMPLATE);
+          var wrap = element.children();
+          var controller = element.controller("mdChips");
+
+          expect(wrap.hasClass("md-removable")).toBe(false);
+          expect(controller.removable).toBe(false);
+
+          var containers = wrap[0].querySelectorAll(".md-chip-input-container");
+
+          expect(containers.length).toBe(0);
+
+          var removeContainer = wrap[0].querySelector('._md-chip-remove-container');
+          expect(removeContainer).not.toBeTruthy();
+        });
+
+        it('should not remove chip through the backspace/delete key when removable is set to false', inject(function($mdConstant) {
+          var element = buildChips(CHIP_NOT_REMOVABLE_TEMPLATE);
+          var wrap = element.find('md-chips-wrap');
+          var controller = element.controller("mdChips");
+          var chips = getChipElements(element);
+
+          expect(wrap.hasClass("md-removable")).toBe(false);
+          expect(controller.removable).toBe(false);
+
+          controller.selectChip(0);
+
+          wrap.triggerHandler({
+            type: 'keydown',
+            keyCode: $mdConstant.KEY_CODE.BACKSPACE
+          });
+
+          var updatedChips = getChipElements(element);
+
+          expect(chips.length).toBe(updatedChips.length);
+        }));
+
+        it('should remove a chip by default through the backspace/delete key', inject(function($mdConstant) {
+          var element = buildChips(BASIC_CHIP_TEMPLATE);
+          var wrap = element.find('md-chips-wrap');
+          var controller = element.controller("mdChips");
+          var chips = getChipElements(element);
+
+          controller.selectChip(0);
+
+          wrap.triggerHandler({
+            type: 'keydown',
+            keyCode: $mdConstant.KEY_CODE.BACKSPACE
+          });
+
+          var updatedChips = getChipElements(element);
+
+          expect(chips.length).not.toBe(updatedChips.length);
+        }));
+
+        it('should set removable to true by default', function() {
+          var element = buildChips(BASIC_CHIP_TEMPLATE);
+          var wrap = element.children();
+          var controller = element.controller('mdChips');
+
+          expect(wrap.hasClass('md-removable')).toBe(true);
+          // The controller variable is kept undefined by default, to allow us to difference between the default value
+          // and a user-provided value.
+          expect(controller.removable).toBe(undefined);
+
+          var containers = wrap[0].querySelectorAll("._md-chip-input-container");
+          expect(containers.length).not.toBe(0);
+
+          var removeContainer = wrap[0].querySelector('._md-chip-remove-container');
+          expect(removeContainer).toBeTruthy();
+        });
+
+        it('should append dynamically the remove button', function() {
+          var template = '<md-chips ng-model="items" readonly="true" md-removable="removable"></md-chips>';
+
+          scope.removable = false;
+
+          var element = buildChips(template);
+          var wrap = element.children();
+          var controller = element.controller("mdChips");
+
+          expect(wrap.hasClass("md-removable")).toBe(false);
+          expect(controller.removable).toBe(false);
+
+          var containers = wrap[0].querySelectorAll("._md-chip-remove-container");
+          expect(containers.length).toBe(0);
+
+          scope.$apply('removable = true');
+
+          expect(wrap.hasClass("md-removable")).toBe(true);
+          expect(controller.removable).toBe(true);
+
+          containers = wrap[0].querySelector("._md-chip-remove-container");
+          expect(containers).toBeTruthy();
+        });
+
+      });
 
       describe('when readonly', function() {
         var element, ctrl;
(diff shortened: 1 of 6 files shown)


DevVersion:
**Updates**
- Added Test for `Removeonly`
- Added Attribute Expression to NGDocs

topherfangio:
I think I would rather see this as an `md-removable` option that works in tandem with the `readonly` option.

For instance:

 - By default - chips are writable **and** deletable _(current functionality)_
 - `readonly="true"` - chips are readonly **and** deletable _(current functionality)_
 - `readonly="true" md-removable="true"` - chips are readonly, **and** deletable _(new syntax, but same functionality as above)_
 - `readonly="true" md-removable="false"` - chips are readonly, **not** deletable _(new)_
 - `md-removable="false"` - chips are writable, **not** deletable _(new)_

I think this gives the user a bit more customizability and follows better conventions.

_Note: In this case, we should begin the attribute with `md-`. The `readonly` attribute doesn't contain an `md-` because it is an existing/standard HTML attribute._

Thoughts?

DevVersion:
Yes, sorry for the late answer - Seems to be better, I will try to implement that ;)

DevVersion:
@topherfangio - All done - Added rich Tests and added a demo checkbox to toggle the chips ;)

src/components/chips/demoBasicUsage/style.scss
```diff
@@ -40,10 +40,9 @@
         }
       }
     }
-    &:not(.md-readonly) {
-      md-chip-template {
-        padding-right: 5px;
-      }
+
```
topherfangio: I don't think you can *always* add this; I think it should have the padding only if `md-removable="true"`. Right?

src/components/chips/js/chipsDirective.js
```diff
@@ -65,6 +65,8 @@
    * @param {string=} placeholder Placeholder text that will be forwarded to the input.
    * @param {string=} secondary-placeholder Placeholder text that will be forwarded to the input,
    *    displayed when there is at least on item in the list
+   * @param {boolean=} md-removeable Allows only to remove a chip from the list through the delete buttons
```
topherfangio: I think I would change this to the following:

> Enables or disables the deletion of chips through the removal icon or the Delete/Backspace key. Defaults to true.

Right now, it sounds like it somehow changes what controls are used for deletion; not the deletion capability itself :smile:

src/components/chips/js/chipsDirective.js
```diff
@@ -121,7 +123,7 @@
       <md-chips-wrap\
           ng-if="!$mdChipsCtrl.readonly || $mdChipsCtrl.items.length > 0"\
           ng-keydown="$mdChipsCtrl.chipKeydown($event)"\
-          ng-class="{ \'md-focused\': $mdChipsCtrl.hasFocus(), \'md-readonly\': !$mdChipsCtrl.ngModelCtrl }"\
+          ng-class="{ \'md-focused\': $mdChipsCtrl.hasFocus(), \'md-readonly\': !$mdChipsCtrl.ngModelCtrl, \'md-not-removeable\': !$mdChipsCtrl.removeable}"\
```
topherfangio: Why not just `md-removable` here instead of `md-not-removable`? In the CSS, you should be able to use `:not(.md-removable)` which I think makes a bit more sense.

src/components/chips/demoBasicUsage/style.scss
DevVersion: Yes, thanks

src/components/chips/js/chipsDirective.js
DevVersion: Haha, thanks - used your description.

src/components/chips/js/chipsDirective.js
DevVersion: Definitely better, also privatized the variable :)

topherfangio:
Quick question re:

> readonly="true"- chips are readonly and deletable (current functionality)

Is that actually true? What I mean is: right now, are readonly chips actually deleteable? I couldn't see that in any of the existing demos, but I seem to remember that it might have been a bug.

DevVersion:
@topherfangio - Updated a few things
- Made the `md-removeable` class public.
- Disabled the Backspace/Delete Key when `md-removeable="true"`
- Added accompanying tests.

About your question - I just implemented everything as you wrote in your previous comment. Let's discuss that per Chat.

DevVersion:
Note to myself: 
- Needs to be discussed with Team
- Needs to be Rebased
- Needs to fix attribute names (`removeable` => `removable`) - Thank you very much @ilovett

topherfangio:
So, I think the behavior may have changed at some point which means we may need to update this PR accordingly.

#### Current Functionality (as just tested in 1.1.0-RC1)
- Default (no modifiers) - chips are writable **and** deletable
- `readonly="false"` - chips are writable **and** deletable (same as above)
- `readonly="true"` - chips are readonly; **not** deletable

#### Proposed Functionality
- Default (no modifiers) - chips are writable **and** deletable
- `readonly="false"` - chips are writable **and** deletable (same as above)
- `readonly="true"`- chips are readonly; **not** deletable

So this should be identical to the current functionality, but then we add the following:

- `readonly="true" md-removable="false"` - chips are readonly, **not** deletable *(new syntax, but same functionality as just `readonly="true"`)*
- `readonly="true" md-removable="true"` - chips are readonly, **and** deletable *(new syntax/functionality)*
- `md-removable="false"`- chips are writable, **not** deletable*(new/syntax functionality)*

#### Conclusion
So in conclusion, I think all we're saying is that `readonly` should work as it always has (i.e. also modifying removability), but the `md-removable` attribute should allow overriding how `readonly` currently functions.

Thoughts?

DevVersion:
@topherfangio Yes, I think that is a very good solution. I will implement that soon and let you know if ready.

DevVersion:
@topherfangio Finally got the time to update this old PR - and it's finished - please take a look again.

src/components/chips/chips.spec.js
```diff
@@ -256,6 +258,107 @@ describe('<md-chips>', function() {
...
+          var wrap = element.children();
+          var controller = element.controller("mdChips");
+
+          expect(wrap.hasClass("md-removable")).toBe(false);
+          expect(controller.removable).toBe(false);
+
+          var containers = wrap[0].querySelectorAll("._md-chip-remove-container");
+          expect(containers.length).toBe(0);
+
+          scope.removable = true;
```
topherfangio: You should be able to refactor this to just `scope.$apply('removable = true')` to save space 😄

src/components/chips/chips.spec.js
```diff
@@ -303,6 +406,32 @@ describe('<md-chips>', function() {
...
+          expect(element.find('md-chips-wrap')).not.toHaveClass('md-readonly');
+
+          scope.$apply('isReadonly = true');
+
+          expect(element.find('md-chips-wrap')).toHaveClass('md-readonly');
+
+          expect(ctrl.removable).toBeUndefined();
+
+          var removeContainer = wrap[0].querySelector('._md-chip-remove-container');
+          expect(removeContainer).not.toBeTruthy();
```
topherfangio: Can you use `toBeFalsy()` here instead of `.not.toBeTruthy()`?

src/components/chips/js/chipsController.js
```diff
@@ -167,11 +166,17 @@ MdChipsCtrl.prototype.updateChipContents = function(chipIndex, chipContents){
...
  * @return {boolean}
  */
-MdChipsCtrl.prototype.isEditingChip = function(){
+MdChipsCtrl.prototype.isEditingChip = function() {
   return !!this.$element[0].getElementsByClassName('_md-chip-editing').length;
 };
 
 
+MdChipsCtrl.prototype.isRemovable = function() {
+  return this.readonly ?
```
topherfangio: Although very succinct (:smile:), this is a bit difficult to read with a double-ternary operator. Can we refactor this into a few `if`s?

src/components/chips/demoBasicUsage/index.html
```diff
@@ -58,6 +61,13 @@ <h2 class="md-title">Display an ordered set of objects as chips (with custom tem
 
     <br/>
     <md-checkbox ng-model="ctrl.readonly">Readonly</md-checkbox>
-
+    <md-checkbox ng-model="ctrl.removable">
+      Removable
+      <span ng-if="ctrl.removable === undefined">(Currently is <code>undefined</code>)</span>
+    </md-checkbox>
+    <p class="md-caption">
+      <b>Note</b>: When removable is not defined, readonly will overwrite the remove behavior
```
topherfangio: Can we change this to:

> `When <code>md-removable</code> is <code>undefined</code>, <code>readonly</code> automatically sets <code>md-removable</code> to <code>false</code>.

topherfangio:
Added a few comments, but overall looks great! Thanks so much for all of your work on this one!

src/components/chips/chips.spec.js
DevVersion: Very nitpicky :smile: but that's delivers perfect code!

src/components/chips/chips.spec.js
DevVersion: Yes, that should work.

src/components/chips/demoBasicUsage/index.html
DevVersion: Yes, this seems to be more descriptive.

src/components/chips/js/chipsController.js
DevVersion: I had that before, and it looked really unclear. I definitely prefer a ternary operator here.  I think it should be also fine with the ternary :smile:

DevVersion:
@topherfangio Haha, yeah this PR is really old and I needed to rework it completely again. Overall I think once it's finished, it will provide a very cool feature.

src/components/chips/js/chipsController.js
ThomasBurleson: The standard format for ternary is this:

```js
MdChipsCtrl.prototype.isRemovable = function() {
  return this.readonly ?  this.removable :       
            angular.isDefined(this.removable) ? this.removable : true;
};
```

Please fix.

ThomasBurleson:
@DevVersion  - CI is failing with this PR.

DevVersion:
@ThomasBurleson As already said on Slack, it seems like the latest changes in master are causing that issue..

src/components/chips/js/chipsController.js
DevVersion: @ThomasBurleson Done, but the Google Code Conventions mention IMO a bit more clear format.

- https://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting

joaovieira:
Waiting for this 😭

topherfangio:
@DevVersion I know this says "needs: review" but is it actually ready? The comment above leads me to believe that there is more work that needs to be done.

alice-kroch:
Really need this feature

topherfangio:
@ThomasBurleson LGTM!

SneakyTurtle20:
Any update on this feature? Thanks.

Splaktar:
@SneakyTurtle20 as the PR says, this was released in 1.1.0. You can see a demo of it here: https://material.angularjs.org/latest/demo/chips#basic-usage. There is a checkbox for "Removable".

Issue #5794, opened 2015-11-17
md-tabs: scroll not working depending on md-tab-item width


alexandrebrasil:
When creating an md-tab with enough tabs to make scrolling necessary, the scrolling button gets disabled sometimes. From what I could check, the button gets enabled/disabled depending on the position of the last tab (function canPageForward() of tabsController). The problem seems to be caused by the last md-tab-item being pushed to a new line (see image below), making its offset to be smaller than the end of the controller.

![tabs](https://cloud.githubusercontent.com/assets/8227203/11224192/e655ab80-8d5a-11e5-8fd1-eadf419ea715.png)

This problem seems to be created by the fixed "width" style being set in the md-tabs-canvas block  being smaller than the sum of the tabs' width (by updatePagingWidth() function in tabsController). From what I could see, this is caused by element.offsetWidth being an integer, while the tabs' width being a floating point (see image below). When the container width become the sum of the element.offsetWidth of the tabs, the rounding error (when the tab width fraction is < 0.5) makes the container width becomes smaller than the tabs.

![tabs2](https://cloud.githubusercontent.com/assets/8227203/11224446/26847d2a-8d5c-11e5-8394-f696c72d4a5a.png)

I was able to replicate this bug consistently using the following snippet rendered with Roboto font:
```html
		<div style="width: 500px; margin: 0 auto;">
			<md-tabs>
				<md-tab label="Em #{{i}}" ng-repeat="i in [1,2,3,4,5,6,7,8,9,10]">
					<span>Tab content #{{i}}</span>
				</md-tab>
			</md-tabs>
		</div>			
```

alexandrebrasil:
Forgot to mention: I was able to circumvent this bug by changing the updatePagingWidth() function with the following code:
```javascript
  function updatePagingWidth() {
    var width = 1;
    angular.forEach(getElements().dummies, function (element) { width += element.offsetWidth + 1; });
    angular.element(elements.paging).css('width', width + 'px');
  }
```

Unfortunatelly I don't have access to commit in github from behind my company's firewall, so I can't make a repository clone to run the tests and send a pull request.

alexandrebrasil:
I didn't find it before, but this is a duplicate for #5439.

alexandrebrasil:
@ThomasBurleson 

Commit a242a5f has brought this problem back. Using Math.ceil won't solve the problem because `element.offsetWidth` is already an integer.

Issue #5770, opened 2015-11-16
Tabs being eaten. Unknown.


jaunkst:
Very strange behavior in  tabs  v1.0.0-rc4-master-9d52697

below does not work
One of the tabs seems to not show/go behind other tabs or something.
````
openTabs = [{title: 'LEADS'}, {title: 'LEADS'}, {title: 'LEADS'}]
````

below works find.
````
openTabs = [{title: 'LEAD1'}, {title: 'LEAD2'}, {title: 'LEAD3'}]
````

When testing with various random tab titles sometimes one will get eaten up and not show up until one of the tabs are removed dynamically.

This also happens when hard-coded
````

md-tabs(style='width: 500px;')
  md-tab
    md-tab-label
      | Leads
  md-tab
    md-tab-label
      | Leads
  md-tab
    md-tab-label
      | Leads


````

jaunkst:
I see that the md-pagination-wrapper can have the wrong width causing the tabs to wrap.
md-pagination-wrapper

jaunkst:
![cursor_and_arxia_and_2__electron____electron_helper_](https://cloud.githubusercontent.com/assets/982066/11195110/b6f99530-8c75-11e5-9c50-23a7006383fd.png)

jaunkst:
starting the width at 2 instead of 1 corrects the issue for me. 
````
  function updatePagingWidth() {
    var width = 2;
    angular.forEach(getElements().dummies, function (element) { width += element.offsetWidth; });
    angular.element(elements.paging).css('width', width + 'px');
  }
````

gpopovic:
+1

scottreisdorf:
This is still not fixed in  v1.0.0-rc7, See my code Pen:  http://codepen.io/scottreisdorf/pen/EPjXzN
and related Issue I just made: https://github.com/angular/material/issues/6212

Issue #5692, opened 2015-11-12
md-tabs not pagination full width


chacbumbum:
I have a problem with md-tabs

[http://codepen.io/anon/pen/NGeqNa](http://codepen.io/anon/pen/NGeqNa)

paging did not seem to work properly.

Hope someone can help me!

alexandrebrasil:
Seems like the same problem is reported in #5770 and #5794.

PR #5801, opened 2015-11-18
fix(tabs): labels with fraction CSS width disabling pagination

alexandrebrasil:
When the md-tab-label CSS width is calculated to a fraction numbem, its value and element.offsetWidth don't match. When the paging container's width is calculated, depending on how the offsetWidth of each label is truncated/rounded, the resulting width is smaller then the actual sum of
all labels. The smaller width causes a line break and forces the last label to go to next line with its offsetLeft to be zero, and that results in the next page button to be disabled. The new width calculation adds 1px to the offsetWidth to account for rounding/truncating errors, resulting in a slightly bigger paging container.

Fixes angular/material#5794, Fixes angular/material#5770, Fixes angular/material#5692.

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -541,7 +541,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
 
   function updatePagingWidth() {
     var width = 1;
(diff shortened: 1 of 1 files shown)


stewones:
+1 for this PR please!

Issue #5430, opened 2015-10-29
tooltip: position with scrolled content is wrong


nestorbritez:
md-tooltip is not working properly when scrolling or is close to the edge of the page.

![tooltip2](https://cloud.githubusercontent.com/assets/2065058/10820619/e91b9ef2-7e2c-11e5-972d-07f416a38235.gif)
![tooltip1](https://cloud.githubusercontent.com/assets/2065058/10820620/e91ced20-7e2c-11e5-9234-391f151c5209.gif)

topherfangio:
Can you please tell us what version you are using and ideally provide a Codepen which demonstrates the issue so that we can debug it?

nestorbritez:
@topherfangio You can use the official site to test it

ThomasBurleson:
In Tooltip Demo page, scroll the page so the content moves up 15px. Hovers over the **View Source** or the **Edit on CodePen** buttons does NOT set the Tooltip position properly.

dohomi:
This might be a duplicate of https://github.com/angular/material/issues/5175 

In all scrolling containers the tooltip are misplaced way much to the top since ```1.0.0-rc1```

SanderElias:
Duplicate/varioation of #2791. Closed to concentrate discussion in there

Issue #4954, opened 2015-10-01
mdTooltip inside mdDialog wrong position


weisk:
mdTooltip position gets messed up when it is inside an $mdDialog, and the body is scrolled.

plunkr provided:
http://plnkr.co/edit/iEmViRI0wYke8onEsaok?p=preview

related to https://github.com/angular/material/issues/1883

jelbourn:
I can confirm that the tooltip is off.

SanderElias:
Duplicate of #2791. Closed to concentrate discussion in there.

Issue #2844, opened 2015-05-14
md-tooltip won't scroll down with buttons


butaixianran:
Issue with 0.9.0, works fine with 0.8.3

When using ng-repeat on the container of md-button,  the position of md-tooltip won't work fine.

check this codepen:
[http://codepen.io/butaixianran/pen/pJyWMK](http://codepen.io/butaixianran/pen/pJyWMK)

here is a screenshot:

![2015-05-14 7-54-15](https://cloud.githubusercontent.com/assets/5558722/7623731/ddfb8a6e-fa10-11e4-9cee-b371a6c6e046.jpg)

If ng-repeat is only used on md-button, not it's container, then tooltip will be fine. But for my site, I really need to use ng-repeat on the container.

So, how should I fix this? thanks!!

shak18:
i can confirm this issue happening with 0.9, event scrolling a little bit makes it noticiable :S

marcysutton:
This sounds like a duplicate of #2791 and possibly #2406.

butaixianran:
@marcysutton

I've seen these 2 issues before, look like the same issue, but they don't give a online codepen for testing.

And they seem don't need a solution for now, which I really need...

r0b-:
+1 (can't give up on scrolling nor tooltips)

SanderElias:
Duplicate of #2791, closed to concentrate discussion in there

PR #5809, opened 2015-11-18
Fix(tooltip): positional issues.

SanderElias:
Fix(tooltip): positional issues.

The tootltip is changed from absolute to fixed positioning. This is done to support the different available layout modes. Also the tooltip will get removed on scroll events to prevent wrong offsets

fixes: #5430 #4954 #2844 #2791 and #2037

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -117,6 +117,12 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
       element.attr('role', 'tooltip');
     }
 
+    function windowScrollHandler (e) {
+      // in case of scroll, remove the fixed positioned tooltip.
+      setVisible(false);
+    };
+
+
     function bindEvents () {
       var mouseActive = false;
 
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@topherfangio - can you review and test plz.

topherfangio:
Overall I think this looks pretty good and fixes a lot of important issues. A couple of issues that I noticed:

1. **Minor** - The "Refresh" button's tooltip is positioned very far left; not exactly sure why this is happening.

1. **Medium** - The tooltips on the FAB Speed Dial demo (the second one) are not correctly positioned (although, I think they have been broken for a while since other tooltip changes were made)

1. **Medium** - If you check the "Insert Drive" checkbox and then scroll, the tooltip does not disappear (I think you may have mentioned this as a known issue in our meeting)

1.  **Big** - If you scroll the button off the screen and check the "Insert Drive" checkbox, the tooltip will still appear, and it shows above the toolbar at the top. I say this is a big issue because I can imagine many scenarios where people make the tooltips visible/invisible via code and it would be bad if they all magically appeared in odd places.

   Perhaps we can check if the element's `offsetTop` is above the parent container's `offsetTop` to see if it's off the screen? Same thing with `offsetLeft` and height/width, etc?

SanderElias:
@topherfangio 
1. I'll check this out.
2. this is a separate issue(#4445), and had to do with animation work needs to be done here
3. I'll recheck this. probably a scroll-event bubble issue.
4. hmm. I'll check the offsetTop's. Problem here is that the 'layout mode' has a lot of influence here. I'll report back on this.

Issue #5697, opened 2015-11-12
md-tabs doesn't scroll when width of tab wider than width of md-tabs-canvas tag 


ildarnm:
For replay, set the window's width less than wide of the second tab.
http://codepen.io/anon/pen/Yydwmz

PDSBGreg:
We can across this the other day as well. You can still swipe or use the mouse wheel to scroll the tabs, which isn't terribly intuitive. We solved it by putting a max width on the tabs to something that would be smaller than our smallest screen size. It does the ellipsis on the overflowing text. That works Ok for our case, but may not for yours.

stewones:
+1 here. thought this had been fixed in rc5, @ThomasBurleson ?

stewones:
my workaround has been adding +1 in updatePagingWidth()
```js
function updatePagingWidth() {
    var width = 1;
    angular.forEach(getElements().dummies, function (element) {
      width += Math.ceil(element.offsetWidth)+1;
    });
    angular.element(elements.paging).css('width', width + 'px');
  }
```

danielgranat:
I encountered the same problem and used the solution proposed by @PDSBGreg of max width. This help in allowing navigation, but visually it doesn't look so good.
Isn't the inability to click the arrow a critical bug for the tabs functionality?

EladBezalel:
dup #4247

PR #4273, opened 2015-08-21
fix(mdtabs): Fix scrolling for tabs with a long last label

j-meyer:
Having the last tab longer than the canvas width will allow you to scroll to the end of the tab
rather than stop at the start of the tab.

Closes #4247

PR #5821, opened 2015-11-19
fix(md-tabs): doesn't scroll when width of tab wider than width of md-tabs-canvas tag

alexandrebrasil:
The pagination tries to move the viewport to the start of the tab when going forward, and to its end when going backwards. This strategy won't work when the tab width is bigger than that of the canvas, because it'll get stuck in the same tab. Behaviour changed to:
- when paging forward, search for next tab after current offset, so it discards the current showing tab when it's bigger than the viewport;
- when paging backwards, if the tab is bigger than the current viewport, move the canvas to the start of the tab. Otherwise, keep sliding the canvas to its end (current behaviour/RC4).

Fixes angular/material#5697.

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -313,7 +313,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
         i, tab;
     for (i = 0; i < elements.tabs.length; i++) {
       tab = elements.tabs[ i ];
-      if (tab.offsetLeft + tab.offsetWidth > totalWidth) break;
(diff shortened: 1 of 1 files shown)


EladBezalel:
might be the same as #4273 
@alexandrebrasil @j-meyer can you please validate it's the same issue?

alexandrebrasil:
@EladBezalel Going over both PR, it seems to be the same fix to the same bug.

j-meyer:
Great minds think alike.

EladBezalel:
@alexandrebrasil with that said, closing this, as @j-meyer was first :) thanks

PR #5830, opened 2015-11-19
fix(slider): allow tabindex overwrite

### LINKED ISSUE #5829: md-slider :: cannot set tabindex
The md-slider component sets `tabindex` to 0 and it cannot be overridden.

I have a component created that combines the use of a numeric input and a range input. Similar to the one in the [demo](https://material.angularjs.org/latest/demo/slider) but the components are swapped (numeric input on the left, range on the right). When the user is tabbing through the form and they hit the numeric input and change the value there is no need to tab to the slider field.

https://github.com/angular/material/blob/master/src/components/slider/slider.js#L78

DevVersion:
This commit allows the user to specify the ```tabindex``` of the slider element.

If there are anymore suggestions or other wishes, please tell me :+1: 

Fixes #5829

### DIFF
--- a/src/components/slider/slider.js
+++ b/src/components/slider/slider.js
@@ -74,10 +74,8 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
   // **********************************************************
 
   function compile (tElement, tAttrs) {
-    tElement.attr({
-      tabIndex: 0,
-      role: 'slider'
-    });
+    if (!tAttrs.tabindex) tElement.attr('tabindex', 0);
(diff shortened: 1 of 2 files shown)


danjarvis:
Looks good. I would consider adding a small test for the fix:

    it('should preserve tabindex', function() {
        var slider = setup('tabindex="2", ng-model="model"');
        expect(slider.attr('tabindex')).toBe(2);
    }));

DevVersion:
Done! Thanks @danjarvis for the suggestion.

EladBezalel:
LGTM

PR #5835, opened 2015-11-19
fix(autocomplete): announce item change only when a new item is selected

### LINKED ISSUE #3003: mdAutocomplete: md-selected-item-change triggers to often
md-selected-item-change triggers way to often, in the demo it triggers once on the initial click on the field ( it shouldn't happen because there is no point because nothing "changed"), and it triggers twice on clicking the X to empty the value.

### LINKED ISSUE #5276: md-selected-item-change fire 3 times when clear() the md-autocomplete
http://codepen.io/anon/pen/vNpeGO?editors=101

@robertmesserle @ThomasBurleson 
To prevent this we can do <code>if(item){//do your stuff...}</code>
But, in this case, it will be nice to have a <code>md-clear</code>

Patrik-Lundqvist:
Right now, md-selected-item-change fires when there is an update to the underlying model. The specs states that it should only fire when a new item is selected. The current behavior makes it very hard to differentiate a user initiated change from an underlying model change.

Fixes #5276
Fixes #3003

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -274,9 +274,11 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         $scope.searchText = val;
         handleSelectedItemChange(selectedItem, previousSelectedItem);
       });
+    } else {
+      if (!hasFocus) {
+        $scope.searchText = '';
+      }
     }
-
-    if (selectedItem !== previousSelectedItem) announceItemChange();
   }
 
   /**
@@ -337,7 +339,10 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     getDisplayValue($scope.selectedItem).then(function (val) {
       // clear selected item if search text no longer matches it
       if (searchText !== val) {
-        $scope.selectedItem = null;
+        if ($scope.selectedItem) {
+          $scope.selectedItem = null;
+          $mdUtil.nextTick(announceItemChange, false, $scope);
+        }
 
         // trigger change event if available
         if (searchText !== previousSearchText) announceTextChange();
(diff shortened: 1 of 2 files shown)


EladBezalel:
@topherfangio please review

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -274,9 +274,11 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         $scope.searchText = val;
         handleSelectedItemChange(selectedItem, previousSelectedItem);
       });
+    } else {
```
topherfangio: This appears unrelated to the description of the commit/PR. Can you briefly tell us what this is for? If it's a separate fix, that's okay, we just need to update the commit message to add a mention about it :-)

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -582,14 +587,20 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    * @param index
    */
   function select (index) {
+    var selectedItem = (typeof ctrl.matches[ index ] !== 'undefined') ? ctrl.matches[ index ] : null;
```
topherfangio: This is a bit long for one line. Can you perhaps refactor this into a simple function whose name describes a bit what is happening?

topherfangio:
@Patrik-Lundqvist I've added a few comments. If you could respond/make updates and ping me when you are done, I will review again. Thanks!

src/components/autocomplete/js/autocompleteController.js
Patrik-Lundqvist: This clears the search text if the underlying model is cleared. But yeah you're right, I'll move this to a separate issue/pr.

src/components/autocomplete/js/autocompleteController.js
Patrik-Lundqvist: Will do!

ThomasBurleson:

PR #5836, opened 2015-11-19
	 feat(chips): add expressions for select and deselect actions.

### LINKED ISSUE #5397: md-chips should have a callback option for de-select
I am working on an application where multiple cards are displayed below chips, and I would like selecting a chip to highlight correlated cards, but so far I have no way of knowing when the chip is deselected so I can reset the highlighted cards back to the default state.

DevVersion:
Hello,

I just started working on this, check the current JSDocs to see how it looks.
If there are any things, which I should do better, please tell me... I will fix these as fast as I can.

Fixes #5397

### DIFF
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -215,20 +215,67 @@ describe('<md-chips>', function() {
         expect(scope.removeChip.calls.mostRecent().args[1]).toBe(0);       // Index
       });
 
-
-      it('should call the select method when selecting a chip', function() {
-        var element = buildChips(CHIP_SELECT_TEMPLATE);
+      it("should call the select and deselect expressions", function() {
+        var element = buildChips(
+          '<md-chips ' +
+            'ng-model="items" ' +
+            'md-on-select="onChipSelect($index)" ' +
+            'md-on-deselect="onChipDeselect($index)">' +
+          '</md-chips>'
+        );
         var ctrl = element.controller('mdChips');
+        var scope = element.scope();
 
-        scope.selectChip = jasmine.createSpy('selectChip');
+        scope.onChipSelect = jasmine.createSpy('onChipSelect');
+        scope.onChipDeselect = jasmine.createSpy('onChipDeselect');
 
-        element.scope().$apply(function() {
-          ctrl.items = ['Grape'];
+        scope.$apply(function() {
+          ctrl.items = ['Grape', 'Banana'];
           ctrl.selectChip(0);
         });
 
-        expect(scope.selectChip).toHaveBeenCalled();
-        expect(scope.selectChip.calls.mostRecent().args[0]).toBe('Grape');
+        expect(scope.onChipDeselect).toHaveBeenCalledTimes(1);
+        expect(scope.onChipSelect).toHaveBeenCalledTimes(1);
+
+        expect(scope.onChipSelect).toHaveBeenCalledWith(0);
+        expect(scope.onChipDeselect).toHaveBeenCalledWith(-1);
+
+        scope.$apply(function() {
+          ctrl.selectChip(1);
+        });
+
+        expect(scope.onChipDeselect).toHaveBeenCalledTimes(2);
+        expect(scope.onChipSelect).toHaveBeenCalledTimes(2);
+
+        expect(scope.onChipSelect).toHaveBeenCalledWith(1);
+        expect(scope.onChipDeselect).toHaveBeenCalledWith(0);
+      });
+
+      it("should call the select and deselect expressions", function() {
+        var element = buildChips(
+          '<md-chips ' +
+          'ng-model="items" ' +
+          'md-on-select="newIndex = $index" ' +
+          'md-on-deselect="oldIndex = $index">' +
+          '</md-chips>'
+        );
+        var ctrl = element.controller('mdChips');
+        var scope = element.scope();
+
+        scope.$apply(function() {
+          ctrl.items = ['Grape', 'Banana'];
+          ctrl.selectChip(0);
+        });
+
+        expect(scope.newIndex).toBe(0);
+        expect(scope.oldIndex).toBe(-1);
+
+        scope.$apply(function() {
+          ctrl.selectChip(1);
+        });
+
+        expect(scope.newIndex).toBe(1);
+        expect(scope.oldIndex).toBe(0);
       });
 
       it('should handle appending an object chip', function() {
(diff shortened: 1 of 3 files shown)


DevVersion:
@topherfangio Any ideas, how we should provide that service?

topherfangio:
Honestly, I feel like this kinda goes against the standard Angular pattern of providing the callbacks via attributes in the HTML. This is mostly done so that the HTML provides a "blueprint" of what the code is doing so that other developers looking at it can see at a glance what methods/functions it uses.

I am also hesitant to say that having the service accept the ID of the chips is a good idea.

What is the driving reason for having a service here instead of using the HTML attributes approach that all of the others callbacks use?

DevVersion:
You're right. I will provide another solution... Thanks for your feedback :smile:

DevVersion:
I reworked the complete code, perhaps you could take a look @topherfangio :+1:

src/components/chips/chips.spec.js
```diff
@@ -213,20 +213,40 @@ describe('<md-chips>', function() {
...
+      var CHIP_COMPONENT_TEMPLATE =
+        '<md-chips ng-model="items" md-on-chip-select="onChipSelect($index)" md-on-chip-deselect="onChipDeselect($index)"></md-chips>';
 
-      it('should call the select method when selecting a chip', function() {
-        var element = buildChips(CHIP_SELECT_TEMPLATE);
+      it("should call the select and deselect expressions", function(done) {
+        var element = buildChips(CHIP_COMPONENT_TEMPLATE);
         var ctrl = element.controller('mdChips');
+        var scope = element.scope();
+        var state = 0;
```
topherfangio: What exactly is the `state` variable here? It looks like it controls the expectations below, but you should never have a conditional expectation within a test. If you need to have different expectations, these should be in different tests.

src/components/chips/js/chipsDirective.js
```diff
@@ -62,6 +62,8 @@
    *  </span>
    *
    * @param {string=|object=} ng-model A model to bind the list of items to
+   * @param {expression=} md-on-chip-select An expression to be run each time a new item got selected
```
topherfangio: Why rename this `md-on-chip-select` instead of just `md-on-select`? Is the thought that we may want to know when the chips container is selected/deselected?

Normally, I wouldn't really even mention this, but this is changing a public API very close to our 1.0 release, and I'm hesitant to do so in an RC.

src/components/chips/chips.spec.js
```diff
@@ -213,20 +213,40 @@ describe('<md-chips>', function() {
...
-        element.scope().$apply(function() {
-          ctrl.items = ['Grape'];
-          ctrl.selectChip(0);
-        });
 
-        expect(scope.selectChip).toHaveBeenCalled();
-        expect(scope.selectChip.calls.mostRecent().args[0]).toBe('Grape');
+        scope.$apply(function() {
+          ctrl.items = ['Grape', 'Banana'];
+          ctrl.selectedChip = 0;
```
topherfangio: Why was this changed from `ctrl.selectChip(0)`?

In general, you should prefer using the method that is available (since it probably has additional logic), rather than manually setting the variable.

topherfangio:
@DevVersion I made some additional comments. I hope you don't see these as too harsh, I'm just trying to understand some of your decisions.

Please keep the contributions coming :smile:

src/components/chips/chips.spec.js
DevVersion: No, its not another test, it tests a series of events that should we triggered.

The test goes as following:

**Default Select `-1`** (normal behavior, called at compiling element)
- Calls Select Event (except new index to be `-1`) 

**Select `0`** (called at beginning of test)
- Calls Deselect Event (except old index to be `-1`)
- Calls Select Event (except new index to be `0`)

**Select `1`** (called after selecting `0`)
- Calls Deselect Event (except old index to be `0`)
- Calls Select Event (except new index to be `1`)

So this thing should definitely provided in one test (I hope you understand my thinks) :smile:

src/components/chips/js/chipsDirective.js
DevVersion: I thought, these expressions will be called after a ``chip` gets selected, so I wan't to find a clear attribute for the API (and `md-on-select` can be interpreted as select event for the input field)

src/components/chips/chips.spec.js
DevVersion: Sorry, I can completely understand your thoughts, I will fix it :+1:

DevVersion:
Thanks for being hash!. That's very important for me. I like criticism, but it should be constructive -> and yours is constructive :smile: 

I try my best at helping at this project (my dream is working here :computer:)

EladBezalel:
@topherfangio please re-review

src/components/chips/js/chipsController.js
```diff
@@ -99,11 +99,23 @@ function MdChipsCtrl ($scope, $mdConstant, $log, $element, $timeout) {
...
-   * Whether to use the onSelect expression to notify the component's user
-   * after selecting a chip from the list.
-   * @type {boolean}
+   * Watch for Select Changes if expressions are present
    */
-  this.useOnSelect = false;
+  if (this.onChipSelect || this.onChipDeselect) {
+    var self = this;
+
+    $scope.$watch(function() {
```
topherfangio: I think you might be able to do this as `$mdChipsCtrl.selectedChip` which will increase performance since it won't be executing the function on every digest loop.

topherfangio:
Since we've released 1.0, I think we should change this to be `md-on-select` and `md-on-deselect`.

topherfangio:
Based on the removal of `this.onSelect({'$chip': this.items[this.selectedChip] });` and your new additions, it looks like you have changed the behavior here to pass the `$index` variable instead of the `$chip` variable. Can you refactor it to instead provide both? I think it would be helpful and backwards compatible. If so, can you also update the docs to note that those two variables are available?

Also, can you update the test to verify that this works with any expression and not just a function? I honestly don't see many people doing that, but they may have existing code that looks like the following, and I want to make sure that it works:

    <md-chips md-on-select="selectedIndex = $index">...</md-chips>

DevVersion:
@ThomasBurleson @topherfangio This is already solved, so it should need not need any work. Please review again.

src/components/chips/js/chipsDirective.js
```diff
@@ -62,6 +62,8 @@
    *  </span>
    *
    * @param {string=|object=} ng-model A model to bind the list of items to
+   * @param {expression=} md-on-select An expression to be run each time a new item got selected
```
topherfangio: LGTM, but one last change: Can we make these say

> An expression which will be called when a chips is selected. The `$chip` and `$index` variables are made available in the expression.

Similar for the deselect.

ThomasBurleson:

beaubarbour:
So did the deselect action never get implemented?

PR #5863, opened 2015-11-21
fix(toolbar): add support in scrollshrink to ngshow/hide

### LINKED ISSUE #5706: md-scroll-shrink does not work with ng-show
When using `ng-if` the scroll-shrink still works, but with `ng-show` it only works if the
toolbar has never been hidden.

Short story, this would fail if you toggle `showToolbar` twice:
```html
<md-toolbar md-scroll-shrink ng-show="showToolbar"></md-toolbar>
```

EladBezalel:
`updateToolbarHeight` was not been called when `ngShow` or `ngHide` value has change which caused the toolbar to wait for the debounce to call `updateToolbarHeight` (5 sec)

KUDOS to @topherfangio for helping (and solving) me on this!

fixes #5706

### DIFF
--- a/src/components/toolbar/toolbar.js
+++ b/src/components/toolbar/toolbar.js
@@ -101,6 +101,12 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
 
         attr.$observe('mdScrollShrink', onChangeScrollShrink);
 
+        // If the toolbar has ngShow or ngHide we need to update height immediately as it changed
+        // and not wait for $mdUtil.debounce to happen
+
+        if (attr.ngShow) { scope.$watch(attr.ngShow, updateToolbarHeight); }
+        if (attr.ngHide) { scope.$watch(attr.ngHide, updateToolbarHeight); }
+
         // If the scope is destroyed (which could happen with ng-if), make sure
         // to disable scroll shrinking again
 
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@topherfangio: :grinning:

src/components/toolbar/toolbar.js
```diff
@@ -101,6 +101,12 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
 
         attr.$observe('mdScrollShrink', onChangeScrollShrink);
 
+        // If the toolbar has ngShow or ngHide we need to update height immediately as it changed
+        // and not wait for $mdUtil.debounce to happen
+
+        if (attr.ngShow) { scope.$watch(attr.ngShow, updateToolbarHeight); }
```
ThomasBurleson: @EladBezalel - why use `scope.$watch()` instead of `attr.$observer()`?

EladBezalel: Initially it was `$observe` but than topher came with this jem:
a quick excerpt from here regarding `$observe` verses `$watch` and why we need `$watch` here: http://stackoverflow.com/questions/14876112/angularjs-difference-between-the-observe-and-watch-methods

> (If you try attrs.$observe('attr1') you'll get the string myModel.some_prop, which is probably not what you want.)

The `$observe` was returning the string ‘shouldShowToolbar’, not the value of that expression. And it was only firing once because the expression itself didn’t change.

topherfangio: As a note, I'm pretty sure the `attr.$obseve('mdScrollShrink', onChangeScrollShrink);` line above is also broken in the same way but I didn't get a chance to test.

Also, I highly doubt that people will be dynamically changing the scrollShrink attribute of the toolbar, so it's a VERY low priority to fix :-)

PR #5867, opened 2015-11-21
fix(autocomplete): store hasNotFoundTemplate in shared element

### LINKED ISSUE #5865: When multiple md-autcomplete on the same page, md-not-found doesn't work
I'm using rc-4 and when I have multiple md-autocomplete with one of them having md-not-found defined (as in demo) on one page, it never shows up. If I have only one md-autocomplete, I see this item in the list, it works fine.
Also it looks like on demo page you don't have such issue, but I'm using same JS code and layout.

DevVersion:
As before, the `hasNotFoundTemplate` Boolean was stored in the directive function so its used by all `autocomplete` directives, and that's wrong

Fixes #5865

### DIFF
--- a/src/components/autocomplete/js/autocompleteDirective.js
+++ b/src/components/autocomplete/js/autocompleteDirective.js
@@ -118,7 +118,6 @@ angular
  */
 
 function MdAutocomplete () {
-  var hasNotFoundTemplate = false;
 
   return {
     controller:   'MdAutocompleteCtrl',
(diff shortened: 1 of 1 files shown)


EladBezalel:
LGTM

skovalyov:
Apparently, it fails, if the `<md-autocomplete>` element is located inside another element, which visibility is controlled by `ngIf` directive. Then `template()` is being called first and `hasNotFoundTemplate` property is initialized with `true`, but then when `link()` is being called, `hasNotFoundTemplate` equals `undefined`. This happens due to the way how Angular processes the directive. Here is a Codepen: http://codepen.io/anon/pen/rxQOrq

DevVersion:
@skovalyov That's already a known issue. See #7035. And will be fixed in #7042.
Using an attribute will persist in the `HTMLElement` see MDN

skovalyov:
@DevVersion, great, thanks for update!

PR #5868, opened 2015-11-22
style(theme): all components looks good on dark theme

### LINKED ISSUE #5744: Dropdowns(menu, select) should be white on black theme
md-select does not depend on theming.
This codepend shows the issue : http://codepen.io/anon/pen/GpzZro

EladBezalel:
Components didn't looked like the [spec on dark theme](https://www.google.com/design/spec/style/color.html#color-themes),
Made all dialogs/menus stay white on dark theme.
Changed card color to be as spec.
Changed background color to be `50` (on light theme) as spec defines.

fixes #5744

### DIFF
--- a/src/components/sidenav/sidenav-theme.scss
+++ b/src/components/sidenav/sidenav-theme.scss
@@ -1,3 +1,5 @@
 md-sidenav.md-THEME_NAME-theme {
-  background-color: '{{background-color}}';
+  &, md-content {
+    background-color: '{{background-hue-1}}';    
+  }
 }
--- a/src/components/autocomplete/autocomplete-theme.scss
+++ b/src/components/autocomplete/autocomplete-theme.scss
@@ -1,5 +1,5 @@
 md-autocomplete.md-THEME_NAME-theme {
-  background: '{{background-50}}';
+  background: '{{background-A100}}';
   &[disabled] {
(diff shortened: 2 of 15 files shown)


ThomasBurleson:
@rschmukler - review plz.

EladBezalel:
@ThomasBurleson ping

src/components/sidenav/sidenav-theme.scss
```diff
@@ -1,3 +1,5 @@
 md-sidenav.md-THEME_NAME-theme {
-  background-color: '{{background-color}}';
+  &, md-content {
+    background-color: '{{background-hue-1}}';
```
DevVersion: Not, sure about `hue-1`. 

image

Shouldn't it be the primary background color?

image

Please correct me, if I'm wrong.

EladBezalel: it shouldn't, take a look at [themes at Material spec](https://www.google.com/design/spec/style/color.html#color-themes)

image

but i do agree that hue-1 is not the wanted color

DevVersion: Okay, perfect :+1: By the way: #7374 fixes the basic theming issue with the sidenav.
So you PR will actually depend on that.

PR #5870, opened 2015-11-22
fix(sidenav): fix delayed animation for sidenav if scrollTop != 0

### LINKED ISSUE #5850: md-sidenav was delayed

default

DevVersion:
To checkout:
- Old Version: http://codepen.io/Shyam-Chen/pen/KdYdme
- New Version: http://codepen.io/DevVersion/pen/KdYxbK

Fixes #5850

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -286,6 +286,9 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
       parent[isOpen ? 'on' : 'off']('keydown', onKeyDown);
       backdrop[isOpen ? 'on' : 'off']('click', close);
 
+      // Scroll the parent element to top, to prevent animate delay
+      parent[0].scrollTop = 0;
+
       if ( isOpen ) {
         // Capture upon opening..
         triggeringElement = $document[0].activeElement;


ThomasBurleson:
@EladBezalel - can you review please.

EladBezalel:
@DevVersion IMO this is a partial fix for a bigger problem.

as a user i won't expect the contect to jump to head when sidenav got open.

@ThomasBurleson thoughts?

DevVersion:
@EladBezalel As the current behaviour, its scrolling after the animation to the top, so the user thinks it appearing with a delay.
After this fix it scrolls at first and the animation looks perfect.

EladBezalel:
@DevVersion yes is can see that, can you make a little effort and see if there's a way of preserving the scroll? i think it would be for the best.

@ThomasBurleson LGTM

DevVersion:
@EladBezalel I'm trying to get into this, if we want to show the animation on scrolled content too, we should display the sidenav fixed in a wrapper otherwise the sidenav is fixed to the body of the page.

Any other solutions or ideas?. We should provide this temporary fix (it isn't doing something bad. it only fixes a bug with a good solution). But as you and I prefer we should provide a **PERFECT** solution.

EladBezalel:
Sidenav should have fixed position that should't affect the content..

DevVersion:
@EladBezalel This is a fix, already tested by me on Chrome, which works :smile: Can you please review that?

EladBezalel:
@DevVersion I pulled your change and tested with the codepen from above and its still seems to jump to the start

EladBezalel:
Closing as #5887 has a better outcome

Issue #4371, opened 2015-08-29
md-slider rendering issue in the case of dynamic min/max


ivan7237d:
The following CodePen illustrates (when value < min, thumb ends up outside the ruler): http://codepen.io/anon/pen/NGKVPN

karolkochan:
+1

PeWu:
A workaround is using $timeout to set the value in a separate digest cycle.
In the above CodePen, replace `$scope.value = 10;` with
`$timeout(function() { $scope.value = 10;  });`

PR #5874, opened 2015-11-22
fix(slider): use validators too if its numeric string

### LINKED ISSUE #5872: Slider gets too long if ng-model value goes higher than max
A text input and a material-slider share the same ng-model variable. The slider is set with a max value. Now if the input receives a value larger than the slider's max, the slider grows across the page. It's very ugly!

I would argue that the slider should not get longer, instead its marker should just go to the end of the bar.

DevVersion:
Normally the numeric string will be applied to the model and the slider won't check for min and max.
This checks the strings too and validates for min and max

Before: http://codepen.io/anon/pen/vNMvNZ
After: http://codepen.io/DevVersion/pen/yYrGVW

Fixes #5872


EladBezalel:
rel #4371

PR #5877, opened 2015-11-23
fix(list): add `target` to copied attributes for `md-list-item` buttons

code-tree:
This is to complement the previous fix in issue #481 that added support for `target` on `md-button`, however, that support was not carried over to an `md-button` auto-created in an `md-list-item`.


code-tree:
I signed it!

amritk:
Was this ever added in? Doesn't work on 1.0.7

Issue #5845, opened 2015-11-20
Scrolling should be disabled when sidenav is open


Wardormeur:
As of https://www.google.com/design/spec/layout/structure.html#structure-side-nav, when opening a sidenav, scrolling should be disabled in 2 ways : 
 - sidenav should be '''position:fixed'''
 - md-backdrop shouldn't be scrollable 
A good example is their own menu, even though it allows "3rd click scroll" on backdrop whereas it shouldn't (?)
Else what, codepen demo of what's missing/not working (click right side nav)
http://codepen.io/anon/pen/rObOZE

Wardormeur:
related to #5861 ?

DevVersion:
@Wardormeur Yes, seems to be the same, btw: perfect demo :+1:

EladBezalel:
@Wardormeur please write your comment there, your issue was first but #5861 has more info. thanks!

PR #5887, opened 2015-11-23
fix(sidenav): animate the sidenav on scrolled parent too.

### LINKED ISSUE #5850: md-sidenav was delayed

DevVersion:
See #5870

Closes #5870 Fixes #5850 

@EladBezalel Please review that ... Thanks in advance :+1:

### DIFF
--- a/src/components/backdrop/backdrop.scss
+++ b/src/components/backdrop/backdrop.scss
@@ -6,9 +6,10 @@ md-backdrop {
 
   position: absolute;
   top:0;
-  bottom:0;
   left: 0;
   right: 0;
+
+  height: 100%;
 
   z-index: $z-index-backdrop;
 
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -297,6 +297,11 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
       parent[isOpen ? 'on' : 'off']('keydown', onKeyDown);
       if (backdrop) backdrop[isOpen ? 'on' : 'off']('click', close);
 
+      // Align the sidenav at the current scroll position
+      var topAttribute = parent[0].scrollTop + "px";
+      element.css("top", topAttribute);
+      backdrop.css("top", topAttribute);
+
       if ( isOpen ) {
         // Capture upon opening..
         triggeringElement = $document[0].activeElement;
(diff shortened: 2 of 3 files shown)


EladBezalel:
I'm not sure your implementation is the best, did you tried to make the sidenav fixed? 
Related #5845

src/components/backdrop/backdrop.scss
```diff
@@ -6,10 +6,11 @@ md-backdrop {
 
   position: absolute;
   top:0;
-  bottom:0;
   left: 0;
   right: 0;
 
+  height: 100%;
```
EladBezalel: it might affect other components

DevVersion:
Yes, I already tried to make it fixed, but then it always shows fixed in relative to the screen. 
Later I added a relative wrapper to set the relative parent of the fixed sidenav, but that's not working too.

This should be the best solution for the `No-Scroll` Animation for the UX

src/components/backdrop/backdrop.scss
DevVersion: It shouldn't affect any other components, because it's still in relative and only dynamic to the parent height.

DevVersion:
**Update**
- Solved conflicts by rebasing to upstream/master

DevVersion:
Should be better now @EladBezalel

EladBezalel:
What is the difference between #5870 ?

DevVersion:
In this version, we prevent scrolling to zero, so the animation can run. And in #5870 we always scroll up to zero, and the animation can run too. But this version is much more user friendly.

EladBezalel:
LGTM

DevVersion:
So, this PR needs still Reviewing? :)

EladBezalel:
I still want Thomas to take a look at this before merging

DevVersion:
Perfect ;) thanks

asureshhegde:
I'm still facing the same issue nothing working for me 
Can I know the fix for this? or do I need to make any changes?
Thank you

ThomasBurleson:
@DevVersion - You have not tested this with http://localhost:8080/demo/sidenav, the Left sidenav emits an RTE!

DevVersion:
@asureshhegde Yes, the fix is outdated and throwing an RTE, when rebasing on `master`. I will rework that.

asureshhegde:
@DevVersion so is there any possible fix that can be done for now?

ThomasBurleson:
Closing as out-dated. Will not merge.

travelerspb:
@DevVersion PR sint merged, bus is still exists

PR #5939, opened 2015-11-26
Allow focus on contentEditable fields

### LINKED ISSUE #5937: contentEditable not selectable with Angular Material on Android
On Android, I can't select a contentEditable piece of text.

Without Angular material: http://codepen.io/barryvdh/pen/PPrJLr
With Angular Material: http://codepen.io/barryvdh/pen/RWzLvE

It works on desktop. Probably related to gestures/fastclick (maybe similar to https://github.com/driftyco/ionic/issues/421 ?)

FYI, I'm using https://github.com/thijsw/angular-medium-editor which used contentEditable

barryvdh:
Fixes #5937

Not sure if this is cross-browser, tested etc, but does seem to fix the issue for Android/iOS

### DIFF
--- a/src/core/services/gesture/gesture.js
+++ b/src/core/services/gesture/gesture.js
@@ -108,8 +108,8 @@ function MdGesture($$MdGestureHandler, $$rAF, $timeout) {
 
           return (element.getAttribute('tabindex') != '-1') &&
               !element.hasAttribute('DISABLED') &&
-              (element.hasAttribute('tabindex') || element.hasAttribute('href') ||
-              (focusableElements.indexOf(element.nodeName) != -1|| element.isContentEditable));
+              (element.hasAttribute('tabindex') || element.hasAttribute('href') || element.isContentEditable ||
+              (focusableElements.indexOf(element.nodeName) != -1));
         }
       }
     });
--- a/src/core/services/gesture/gesture.js
+++ b/src/core/services/gesture/gesture.js
@@ -109,7 +109,7 @@ function MdGesture($$MdGestureHandler, $$rAF, $timeout) {
           return (element.getAttribute('tabindex') != '-1') &&
               !element.hasAttribute('DISABLED') &&
               (element.hasAttribute('tabindex') || element.hasAttribute('href') ||
-              (focusableElements.indexOf(element.nodeName) != -1));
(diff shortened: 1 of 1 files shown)


barryvdh:
Any updates on this?

EladBezalel:
@barryvdh is this still happening on head?

barryvdh:
Yes, see http://codepen.io/barryvdh/full/VaXyVW/
On desktop: you can edit/type both paragraphs.
On mobile (Android, Chrome browser): You can change the text by selecting/just typing, but you can't just click to see a cursor and start editing. Textarea works as expected.

barryvdh:
So this is not a good way to fix it or ..?

ThomasBurleson:

barryvdh:
Yes but you did reopen the issue: https://github.com/angular/material/issues/5937
Or does that mean it will only be fixed in v2?

DevVersion:
@barryvdh Yeah, I reopened #5937. Please add a test for your change and I'll reopen. (Please ping me, once you've finished)

barryvdh:
Sorry, I looked but I have no idea how to write a test for this. I was looking in https://github.com/angular/material/blob/master/src/core/services/gesture/gesture.spec.js to see what to add/modify, but can't really find a test for the current behavior.
Could you point me in the right direction where the tests are for the current behaviour? (eg. just skipping hijack for input/textarea etc)

DevVersion:
@barryvdh I'm sorry, I led you into the wrong direction. I just realized that this is test will require an iOS or Android environment, so tests will always pass.

DevVersion:
@ThomasBurleson This is a reasonable PR and looks good to me.

barryvdh:
@DevVersion Ok thanks! I changed the check outside those parentheses, should be a bit more readable. now.

kpgarrod:
What is happening about this? It seems to be a simple fix but the problem still exists in 1.1.0 and it is causing major issues for production users of my shiny new, just launched application. Is there some way to patch the 1.1.0 code or is there a branch I can use in which this problem is fixed.

DevVersion:
@kpgarrod This PR has been merged into `master` and should be available in `v1.1.0`.

If you still experience issues it would be good to create a new issue for it.

PR #5943, opened 2015-11-26
fix(dialog, autocomplete): only autofocus instant if not in dialog + trigger focus events of dialog show

### LINKED ISSUE #5776: autofocus: on autocomplete inside a dialog bug
In [this Codepen](http://codepen.io/EladBezalel/pen/MaLGJe)

When clicking on the button the dialog is opened with the focused autocomplete but the list of options is opened from the button.

DevVersion:
Normally if we put a Autocomplete in a dialog and we specify the autofocus attribute. Then the autocomplete will open the list at beginning of the fade in, but it should only open if the dialog is finished with fade in effect.

Fixes #5776

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -669,4 +669,48 @@ describe('<md-autocomplete>', function() {
     }));
   });
 
+  describe('$mdDialog compatibility', function() {
+
+    beforeEach(module('material.components.dialog', 'material.components.autocomplete'));
+
+    it('should only focus the autocomplete after opening the dialog', inject(function($rootScope, $timeout, $material, $mdDialog) {
+      var parent = angular.element('<div>');
+
+      $mdDialog.show({
+        template: '<md-autocomplete\
+            md-floating-label="Some Label"\
+            md-selected-item="selectedItem"\
+            md-search-text="searchText"\
+            md-items="item in match(searchText)"\
+            md-item-text="item.display"\
+            md-autofocus="true"\
+            placeholder="placeholder">\
+              <span md-highlight-text="searchText">{{item.display}}</span>\
+          </md-autocomplete>',
+        parent: parent
+      });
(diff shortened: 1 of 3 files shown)


DevVersion:
Please review @EladBezalel

src/components/dialog/dialog.js
```diff
@@ -556,7 +556,11 @@ function MdDialogProvider($$interimElementProvider) {
       function focusOnOpen() {
         if (options.focusOnOpen) {
           var target = $mdUtil.findFocusTarget(element) || findCloseButton();
+
           target.focus();
+
+          // force trigger the focus event
```
EladBezalel: remove this comment

DevVersion:
@EladBezalel Remove unnecessary comment.

EladBezalel:
Can you add some tests please?

DevVersion:
CI is currently erroring because of `$timeout.flush()`. I will fix that tomorrow (no time today). Are there any tips for fixing that (then I don't waste time tomorrow)

DevVersion:
@EladBezalel I added a test (was quiet a bit difficult to get that compact & effective). Would be great if you review :+1:

EladBezalel:
Looks good, but i'd like to see the tests for the actual solution your'e providing.
You should test that autocomplete focus is happening after the dialog finishes its `show`

DevVersion:
@EladBezalel Done! Now it's checking for the focus after the animation is done.

For understanding: flush nextTick, check for autofocus, run and skip animation and then verify to be focused.

EladBezalel:
LGTM, @ThomasBurleson can you take a look as well?

DevVersion:
@ThomasBurleson @EladBezalel I improved that a bit. (better test, less code)

DevVersion:
Need's merge right? @EladBezalel

DevVersion:
@EladBezalel @ThomasBurleson Can't reproduce in master anymore.

PR #5954, opened 2015-11-27
style(button): changed raised color to background-900 and not background-contrast

### LINKED ISSUE #5670: md-button (md-raised) with dark theme has an unreadable text color
See attached codepen :
http://codepen.io/anon/pen/PPxBoa

EladBezalel:
Contrast made the raised button text color on dark theme to be white, as the button color,
Changing that to background-900 made it almost black no-matter the background color.

<b>NOTICE:</b> Merge this if the UX team agreed that button should always be white themed

fixes #5670


jelbourn:
LGTM

Issue #3199, opened 2015-06-10
Docs: $mdDialog Controller type listed as {string=}, but can also be function


michaelbromley:
As the title says, in the docs for the `$mdDialog.show()` method, the type of the "controller" option is [listed as `{string=}`](https://github.com/angular/material/blob/v0.9.8/src/components/dialog/dialog.js#L332), whereas it can also be a function, as is even demonstrated in the same docs [here](https://github.com/angular/material/blob/v0.9.8/src/components/dialog/dialog.js#L332) and [here](https://github.com/angular/material/blob/v0.9.8/src/components/dialog/dialog.js#L129), for example.

PR #5957, opened 2015-11-27
docs(dialog): added docs to the elements and fixed controller

### LINKED ISSUE #4728: md-dialog and md-dialog-content are not documented
The Angular/Material DIRECTIVES sidebar does not show these directives at all.

This is a complex subject that really should have detailed API and a few decent demo examples.

Perhaps someone who is not a newbie (as I am) to Angular/Material could correct this in the next release.

EladBezalel:
Added `md-dialog` directive page that contains - `md-dialog`, `md-dialog-content`, `md-dialog-actions`.
Added `function` type to controller method

fixes #4728 and #3199

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -10,6 +10,39 @@ angular
   .directive('mdDialog', MdDialogDirective)
   .provider('$mdDialog', MdDialogProvider);
 
+/**
+ * @ngdoc directive
+ * @name mdDialog
+ * @module material.components.dialog
+ *
+ * @restrict E
+ *
+ * @description
+ * `<md-dialog>` - The dialog's template must be inside this element.
+ *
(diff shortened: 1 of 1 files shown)


src/components/dialog/dialog.js
```diff
@@ -10,6 +10,39 @@ angular
...
+ *
+ * @restrict E
+ *
+ * @description
+ * `<md-dialog>` - The dialog's template must be inside this element.
+ *
+ * Inside, use an `<md-dialog-content>` element for the dialog's content, and use
+ * an `<md-dialog-actions>` element for the dialog's actions.
+ *
+ * * ## Css
```
DevVersion: We should change that text to upper case right? [As here?](https://github.com/angular/material/blob/docs/dialog-directive/src/components/list/list.js#L59)

EladBezalel: :+1:

PR #5973, opened 2015-11-29
fix(datepicker): remove calendar pane after close transition is done

DevVersion:
Currently the close transition will be ignored because we directly remove the pane from the DOM


ThomasBurleson:
@jelbourn - thoughts here ?

jelbourn:
This is on purpose; there's not supposed to be a closing animation.

PR #5974, opened 2015-11-29
feat(sidenav): added onClose callback

### LINKED ISSUE #3179: Sidenav : Is there a way to call a callback on escape or backdrop close ?
Hi,
I have an ui-view in my sidenav and I'd like to reset the router path when the user click outside the sidenav (on the backdrop) or press escape.
Is there a way to do so ?
Thanks for your help

EladBezalel:
Added onClose callback to handle close, backdrop click or escape key pressed

fixes #3179

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -61,6 +61,11 @@ angular
  * // Sync check to whether given sidenav is locked open
  * // If this is true, the sidenav will be open regardless of close()
  * $mdSidenav(componentId).isLockedOpen();
+ * // On close callback to handle close, backdrop click or escape key pressed
+ * // Callback happens BEFORE the close action occurs.
+ * $mdSidenav(componentId).onClose(function () {
+ *   $log.debug('closing');
+ * });
  * </hljs>
  */
 function SidenavService($mdComponentRegistry, $mdUtil, $q, $log) {
@@ -103,6 +108,7 @@ function SidenavService($mdComponentRegistry, $mdUtil, $q, $log) {
         toggle       : rejectFn,
         open         : rejectFn,
         close        : rejectFn,
+        onClose      : angular.noop,
         then : function(callback) {
           return waitForInstance(handle)
             .then(callback || angular.noop);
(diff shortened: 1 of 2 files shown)


src/components/sidenav/sidenav.js
```diff
@@ -55,6 +55,10 @@ angular
  * // Sync check to whether given sidenav is locked open
  * // If this is true, the sidenav will be open regardless of close()
  * $mdSidenav(componentId).isLockedOpen();
+ * // On close callback to handle close, backdrop click or escape key pressed
```
ThomasBurleson: Please note that the callback happens BEFORE the close action occurs.
Should we also support/allow a callback return value to "prevent" the close intention?

EladBezalel: We should, it's called mdIsLockedOpen

ThomasBurleson:
@EladBezalel - needs rebased with latest changes to `$mdSidenav` service.

src/components/sidenav/sidenav.js
```diff
@@ -79,6 +84,10 @@ function SidenavService($mdComponentRegistry, $q) {
       isLockedOpen: function() {
         return instance && instance.isLockedOpen();
       },
+      onClose: function (callback) {
+        if (instance) instance.onClose = callback;
```
ThomasBurleson:
```js
if (instance) instance.onClose = callback || angular.noop;
```

gkohen:
Team, this is not a P4 or a since to have in my opinion. There is no good way to have a logic run when the sidenav closes as a result of clicking on the backdrop.

topherfangio:
@EladBezalel Can you rebase this when you have a minute?

@gkohen Are you saying that it should be a higher priority?

gkohen:
@topherfangio, I'd say the answer is: 'yes' since it's an important enough functionality of which only workaround is polling for isClose.

EladBezalel:
@topherfangio updated :)

topherfangio:
@ThomasBurleson LGTM 👍

ptitdje45:
```
 $mdSidenav('right').onClose(function() {
    //blabla
      });
    });
```

A little problem about it : "SideNav 'right' is not available! Did you use md-component-id='right'?"

Issue #5704, opened 2015-11-12
uiSrefActive should be copied to child elements along with uiSref


gligoran:
You guys have done a good job including more support for uiRouter lately. I've had to hack this in previously, but with the last few relases, I noticed more and more components copy ```uiSref``` attribute to child elements where applicable. For example on ```mdListItem```.

I'd argue that [```uiSrefActive```](http://angular-ui.github.io/ui-router/site/#/api/ui.router.state.directive:ui-sref-active) should be copied too. In the case of ```mdListItem``` in particular this would allow for styling the ```md-button``` which is created because of the presence of ```uiSref```. A very basic example would be using it like so ```uiSrefActive="md-primary"```.

Same goes for directive [```uiSrefActiveEq```](http://angular-ui.github.io/ui-router/site/#/api/ui.router.state.directive:ui-sref-active-eq).

ThomasBurleson:
@EladBezalel - I think we should maintain an internal whitelist of uiRouter attributes to copy.

EladBezalel:
@ThomasBurleson this is a great idea.

ThomasBurleson:
This issue is closed as part of our ‘Surge Focus on Material 2' efforts.
      For details, see our forum posting @ http://bit.ly/1UhZyWs.

PR #5976, opened 2015-11-29
fix(list): copy data prefixed attributes too

### LINKED ISSUE #4392: md-list-item doesn't recognize data-ng-click
it does recognize ng-click, but not data-ng-click. Strange, as Angular can use them interchangeable.

DevVersion:
At the moment the attributes prefixed with `data-` will be ignored, so for example the `data-ng-click` won't be applied to the wrapped button.

Fixes #4392


Frank3K:
Note that there is also an x- prefix. [Docs] (https://code.angularjs.org/1.4.8/docs/api/ng/type/$compile.directive.Attributes).

DevVersion:
@Frank3K Fixed

EladBezalel:
related to #5704

EladBezalel:
Looks ok

DevVersion:
As discussed with Elad a long time ago, there will be a more elegant solution.

ThomasBurleson:
Closed as outdated and better solution pending.

DevVersion:
See #8121

PR #5991, opened 2015-11-30
fix(list): secondary button wasn't coping ngIf attribute

### LINKED ISSUE #5297: md-icon / md-list : secondary action cannot be hidden or disabled
jsFiddle : https://jsfiddle.net/xwn6dgyv/2/
*NB : I'm not sure the link declared in the CSS for the icon font is permanent ; in case the icons do not appear, open [https://fonts.googleapis.com/icon?family=Material+Icons] and paste the content in the CSS zone. I did not manage to import the fonts using the jsFiddle external resources manager.*

I might have missed something, but I cannot seem to properly disable the secondary action inside a list. The jsFiddle is based on the demo for the list component, so it should look familiar.

I need to display (and enable) the secondary action iff some external criteria is true. But the best I can do is to hide the icon - the blank which replaces it remains clickable.

ng-if and ng-show give the same result. Wrapping the icon inside a button and moving ng-if or ng-show to button level does even less.

EladBezalel:
fixes #5297

### DIFF
--- a/src/components/list/list.js
+++ b/src/components/list/list.js
@@ -95,7 +95,7 @@ function mdListItemDirective($mdAria, $mdConstant, $mdUtil, $timeout) {
 
       tEl[0].setAttribute('role', 'listitem');
 
-      if (tAttrs.ngClick || tAttrs.ngHref || tAttrs.href || tAttrs.uiSref || tAttrs.ngAttrUiSref) {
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
This seems like a micro-patch. What about 'ng-switch, ng-hide, etc?

ThomasBurleson:
@EladBezalel  - Please rebase and confirm CI Travis passes.

EladBezalel:
You're right. But this is the exact issue we discussed about..

PR #5992, opened 2015-11-30
style(core): change warn color from red to deep-orange as spec

### LINKED ISSUE #3973: Wrong color for validation errors
Validation errors are #F44336 (on e.g. https://material.angularjs.org/HEAD/#/demo/material.components.select). The spec says they should be #DD2C00: http://www.google.com/design/spec/patterns/errors.html#errors-user-input-errors

EladBezalel:
In #3973 OP addressed us that the color of the input error should be `#DD2C00` which is `A700` from the deep-orange palette so i changed all `warn-500` (the old value from the red palette) to `warn-A700` and change the default warn palette to deep-orange.

fixes #3973

### DIFF
--- a/src/components/datepicker/datePicker-theme.scss
+++ b/src/components/datepicker/datePicker-theme.scss
@@ -20,7 +20,7 @@ md-datepicker.md-THEME_NAME-theme {
     }
 
     &.md-datepicker-invalid {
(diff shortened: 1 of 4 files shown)


ThomasBurleson:
@EladBezalel - please rebase and confirm CI Travis passes.

PR #5993, opened 2015-11-30
fix(list): wrapping secondary if it has ng-click

### LINKED ISSUE #3928: mdIcon in mdListItem gets not wrapped with a button when mdListItem does not have a ngClick
If you build a `mdListItem` without a `ng-click` attribute the containing `mdIcons` with a `ng-click` attribute will not be wrapped with a `md-button` element. I think this is a strange behaviour. 

``` 
  <md-list-item ng-repeat="person in people">
    <p>{{ person.name }}</p>
    <md-icon md-svg-icon="communication:messenger" ng-click="doSecondaryAction($event)" aria-label="Open Chat" class="md-secondary md-hue-3" ng-class="{'md-primary': person.newMessage}"></md-icon>
  </md-list-item>
``` 

I would blame the if statement at line 80 in the [list.js](https://github.com/angular/material/blob/master/src/components/list/list.js)

[Here](http://codepen.io/anon/pen/EjOyVJ) is a codepen example.

EladBezalel:
Secondary was being wrapped only if `md-checkbox` or `md-switch` were near by or `ng-click` was applied on the list item.

fixes #3928

### DIFF
--- a/src/components/list/list.js
+++ b/src/components/list/list.js
@@ -110,6 +110,7 @@ function mdListItemDirective($mdAria, $mdConstant, $mdUtil, $timeout) {
           tEl.addClass('md-no-proxy');
         }
       }
+      wrapSecondary();
       setupToggleAria();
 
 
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@EladBezalel - please rebase and confirm that the Travis CI passes.

PR #5994, opened 2015-11-30
feat(select): support for value change through key up and key down

danjarvis:
This pull request adds support for a new attribute: `allow-keyboard-change`

When this attribute is set, key up and key down will change the selected value.

User's of data-entry applications are accustomed to key up and key down being used to change a value in a drop down field. The default behavior of `md-select` strays away from this behavior and the user is required to make a few extra clicks or use the mouse. This update aims to make `md-select` behave in a more familiar fashion.

Notes:
 - Defaults to false.
 - Ignored if `multiple` is set.

Also, I'm not exactly happy with the name of the attribute. If anyone has any ideas for a better name please let me know.

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -33,6 +33,8 @@ angular.module('material.components.select', [
  *
  * @param {expression} ng-model The model!
  * @param {boolean=} multiple Whether it's multiple.
+ * @param {boolean=} allow-keyboard-change Whether key up and key down change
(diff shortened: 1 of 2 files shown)


DevVersion:
Thanks for your contribution. Please accept the CLA and even squash your commits.

danjarvis:
@DevVersion done.

ThomasBurleson:
Improved API with Keyboard support is pending. Closing this as outdated. Will not merge.

PR #6000, opened 2015-12-01
fix(select): make aria compliant, read value in screen readers

### LINKED ISSUE #3859: md-select role should be listbox
md-select has an ARIA role of "combobox", but it's not actually a combobox; it's a listbox. A combobox has a text field for the user to type in new items.

> A presentation of a select; usually similar to a textbox where users can type ahead to select an option, or type to enter arbitrary text as a new item in the list.

-- http://www.w3.org/TR/wai-aria/roles#combobox

A combobox is required to have a child element that has role "textbox".

I think listbox is more appropriate:

> A widget that allows the user to select one or more items from a list of choices.

-- http://www.w3.org/TR/wai-aria/roles#listbox

### LINKED ISSUE #3891: md-select doesn't read currently selected item in screen readers
A native `<select>` reads aloud the currently selected item when you focus it. `<md-select>` does not. Reproduced in JAWS/Firefox and Chromevox.

This can be reproduced on the [demo page](https://material.angularjs.org/latest/#/demo/material.components.select).

### LINKED ISSUE #4914: md-select fails chrome accessibility dev tools audit
Assuming that https://chrome.google.com/webstore/detail/accessibility-developer-t/fpkknkljclfencbdbgkenhalefipecmb?hl=en is installed...

Go to https://material.angularjs.org/0.11.1/#/demo/material.components.select, open dev tools, go to audit tab and run an audit on that page.

Notice that every instance of md-select throws:
[Severe] Elements with ARIA roles must ensure required owned elements are present (6)

Perhaps it's expecting the aria roles to look more like this:
http://www.w3.org/TR/wai-aria/roles#combobox

### LINKED ISSUE #4977: mdSelect options must be rendered at load time, not just on interaction
In working on #3859 and #4914, I came across an issue that has existed for a while (I've mentioned it to @rschmukler a few times): 

The options for an mdSelect are not in the DOM at all until you click or use the keyboard to activate it. For screen readers, this means a relationship cannot be established between the select and the options it controls, causing an error in Chrome's A11y Developer Tools:
```
[Severe] Elements with ARIA roles must ensure required owned elements are present (1)
```
Here is the part of the ARIA spec this error refers to: http://www.w3.org/TR/wai-aria/roles#mustContain

Furthermore, when the options are rendered, they go into a separate part of the DOM. We can bind a relationship with `aria-owns`, however, that attribute also requires the matching element to be rendered at load time. When I went to do the work required for #3859, I got even more errors in the Chrome A11y audit stemming from this issue. So the problem isn't going away. 

Recommended fix: inject the options on page load so the required ARIA relationships can be established. We can then test `aria-owns` on mdSelect and decide if the options are accessible enough when rendered in a separate part of the DOM.

For async, we could put in at least one option to start (like a placeholder). I'm not sure how accessible this will be, since it might not refresh the accessibility tree reliably, but it's worth a try.

rschmukler:
closes #3891, closes #4914, closes #4977, closes #3859

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -132,6 +132,7 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
       element.append(angular.element('<md-content>').append(element.contents()));
     }
 
+
     // Add progress spinner for md-options-loading
     if (attr.mdOnOpen) {
 
@@ -173,17 +174,18 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
     // Use everything that's left inside element.contents() as the contents of the menu
     var multiple = angular.isDefined(attr.multiple) ? 'multiple' : '';
     var selectTemplate = '' +
-      '<div class="md-select-menu-container">' +
+      '<div class="md-select-menu-container" aria-hidden="true">' +
       '<md-select-menu {0}>{1}</md-select-menu>' +
       '</div>';
 
     selectTemplate = $mdUtil.supplant(selectTemplate, [multiple, element.html()]);
     element.empty().append(valueEl);
+    element.append(selectTemplate);
 
     attr.tabindex = attr.tabindex || '0';
 
     return function postLink(scope, element, attr, ctrls) {
-      var isDisabled;
+      var isDisabled, ariaLabelBase;
 
       // Remove event ngModel's blur listener for touched and untouched
       // we will do it ourself.
@@ -218,7 +220,7 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
 
       var selectContainer, selectScope, selectMenuCtrl;
 
-      createSelect();
+      getSelect();
       $mdTheming(element);
 
       if (attr.name && formCtrl) {
@@ -241,6 +243,7 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
       ngModelCtrl.$render = function() {
         originalRender();
         syncLabelText();
+        syncAriaLabel();
         inputCheckValue();
       };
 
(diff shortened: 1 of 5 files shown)


src/components/select/select.js
```diff
@@ -290,15 +293,17 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
...
-        setAriaLabel();
+        initAriaLabel();
         syncLabelText();
+        syncAriaLabel();
       });
 
-      function setAriaLabel() {
-        var labelText = element.attr('placeholder');
+      function initAriaLabel() {
+        var labelText = element.attr('aria-label') || element.attr('placeholder');
```
jelbourn: Will this throw spurious errors if using `aria-labelledby` ?

src/components/select/select.js
```diff
@@ -398,18 +409,15 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
...
 
       // Create a fake select to find out the label value
-      function createSelect() {
-        selectContainer = angular.element(selectTemplate);
-        var selectEl = selectContainer.find('md-select-menu');
-        selectEl.data('$ngModelController', ngModelCtrl);
-        selectEl.data('$mdSelectController', mdSelectCtrl);
-        selectScope = scope.$new();
-        $mdTheming.inherit(selectContainer, element);
+      function getSelect() {
```
jelbourn: How about `getNativeSelect` ?

src/components/select/select.js
```diff
@@ -879,14 +891,13 @@ function SelectProvider($$interimElementProvider) {
       }
 
       /**
-       * Detach the element and cleanup prior changes
+       * Clean the element up to a closed state
```
jelbourn: ...up to a closed state?

src/components/select/select.js
jelbourn: It's weird to have a function that starts with `get` that doesn't return anything.

rschmukler: It's not a native select. It's just finding the element inside itself that represents the select menu. Maybe `getSelectMenu`

src/components/select/select.js
rschmukler: Will fix.

src/components/select/select.js
jelbourn: Ah, I was thrown off by the comment. Is it still accurate?

jelbourn:
Overall LGTM aside from minor naming nitpicks.

Issue #349, opened 2014-10-01
toast: needs to support non-visual & keyboard users


marcysutton:
To be accessible, `material-toast` needs to be exposed to users who cannot see it appear. It must also be operable without a mouse and be discoverable with touch. There are a few things involved:
1. Toast contents can be read aloud in an [ARIA live region](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Live_Regions) (recommended setting: "polite"). Region must be bootstrapped on page load and be part of the page's render flow (meaning it cannot be hidden with `display: none`). Offscreen CSS works fine for this. 
2. The escape key should operate in a "stack" to close actions in the order they were triggered.
3. Any actions inside of the Toast need to be discoverable and operable from the keyboard or touch (Undo, etc.). 

Escape stack tool idea from @ajoslin: https://gist.github.com/ajoslin/498892d942e4fc5c552c

naomiblack:
@marcysutton is this still an issue...? If so, can you update with status and assign to @jelbourn if more work needed.

calebegg:
This is still an issue. Toasts are not read out loud or keyboard accessible, at least on [the demo page](https://material.angularjs.org/HEAD/#/demo/material.components.toast).

marcysutton:
I have reassigned it to @jelbourn, sorry for the delay!

jelbourn:
@marcysutton While not a complete solution to this, I plan on adding the following attributes to toasts:
```html
<span role="alert" aria-relevant="all" aria-atomic="true">
   <!-- toast content -->
</span>
```

This won't do anything for the keyboard experience, but should at least make the content announced. I know it's also better to have a global announcer for everything (re #3174); hopefully this is an incremental improvement towards that. 

Do you have any feedback on this approach for the short term?

NilsAtGoogle:
I think there are still a11y issues.  I have an internal bug filed saying it's not possible to reach a toast action via accessible navigation before the toast disappears.

Splaktar:
I posted PR https://github.com/angular/material/pull/11424 to address these issues.

Issue #3174, opened 2015-06-08
Feat: <md-aria-status> directive to be consumed by mdToast and mdAutocomplete 


marcysutton:
To enable `$mdToast` accessibility, Angular Material needs a global version of the `<aria-status>` directive used in `mdAutocomplete`. A single Live Region directive for an entire app, which could be used by multiple components as a reusable message center. It would also make screen reader messages easier to localize, per #1696. This should be part of the core functionality to ensure its inclusion for Angular Material components, though I could see a message service being useful for other live regions.

mdAutocomplete uses `aria-live="assertive"` so that messages will be announced immediately as you type, but I'm investigating whether it can use `aria-live="polite"`, which would be more appropriate for mdToast. Components using the global directive will have to share live region settings, as far as I know (role, atomic, relevant, etc.).

marcysutton:
The `announce` feature of this library could be useful: https://github.com/dequelabs/ngA11y/blob/master/src/nga11yannounce.js

naomiblack:
@jelbourn it would be really nice to do this in 1.0 bc of breaking change (and overall a11y), but I'm concerned about time needed to get it right. Do you think this is something you could squeak in along with other m12 tasks on your plate?

jelbourn:
@marcysutton Could you elaborate on the problems with having multiple live regions on a page?

marcysutton:
@jelbourn if there are multiple live regions doing the same thing on a page (and with the same politeness level), they can cause confusion in a screen reader–especially if we can't control how many components are used that require announcements. It makes sense for Angular Material to handle messages centrally to reduce complexity under the hood and optimize for assistive technology. Here is a little more info and a solution that seems to work but may need some more testing in IE: http://spa-a11y.divshot.io/#14

PR #6002, opened 2015-12-01
fix(toast): add missing a11y context. See #349

jelbourn:
cc @marcysutton

### DIFF
--- a/src/components/toast/toast.js
+++ b/src/components/toast/toast.js
@@ -213,7 +213,9 @@ function MdToastProvider($$interimElementProvider) {
         var opts = {
           template: [
             '<md-toast md-theme="{{ toast.theme }}" ng-class="{\'md-capsule\': toast.capsule}">',
-              '<span flex>{{ toast.content }}</span>',
+              '<span flex role="alert" aria-relevant="all" aria-atomic="true">' +
+                '{{ toast.content }}' +
+              '</span>',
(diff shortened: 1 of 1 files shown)


src/components/toast/toast.js
```diff
@@ -213,7 +213,9 @@ function MdToastProvider($$interimElementProvider) {
         var opts = {
           template: [
             '<md-toast md-theme="{{ toast.theme }}" ng-class="{\'md-capsule\': toast.capsule}">',
-              '<span flex>{{ toast.content }}</span>',
+              '<span flex role="alert" aria-relevant="all" aria-atomic="true">' +
```
marcysutton: This won't actually work, because the Toast is inserted and then removed–ARIA Live Regions have to be present at load time and remain in the page to be picked up by screen readers. We talked a while back about creating a global messaging directive for this purpose–could that be incorporated here? https://github.com/angular/material/issues/3174

jelbourn: My testing showed that this worked with NVDA on Firefox and VoiceOver with Safari.

marcysutton: That's good, there is some support then! I did find the toasts were not announced in JAWS with IE or Chrome and Voiceover. JAWS would be more important, while Chrome isn't used with VO as much as Safari.

PR #6003, opened 2015-12-01
fix(layout): allow layout-align without cross-axis value

### LINKED ISSUE #5996: layout-align: Resetting to default when only one argument
`layout-align` no longer works with a single argument. It will replace it with `layout-align="start-stretch"` which drastically changes layouts.

One Example
`layout-align="space-around"` no longer works since 1.0.0-rc5. It is replaced with `layout-align="start-center"` when it should default the second argument to `center` ie parse `layout-align="space-around"` as `layout-align="space-around center"`.

This also happens with the other single argument usecases like `layout-align="space-between"`. I'm suspectting the recent change to the default `layout-align` introduced this issue.

ngraef:
* fix parsing issue to allow single value
* add demo to test omitting second value

fixes #5996

### DIFF
--- a/docs/app/js/app.js
+++ b/docs/app/js/app.js
@@ -633,7 +633,8 @@ function($scope, $attrs, $location, $rootScope) {
     direction: 'row'
   };
   $scope.layoutAlign = function() {
-    return $scope.layoutDemo.mainAxis + ' ' + $scope.layoutDemo.crossAxis;
+    return $scope.layoutDemo.mainAxis +
+     ($scope.layoutDemo.crossAxis ? ' ' + $scope.layoutDemo.crossAxis : '')
   };
 }])
 
--- a/docs/app/partials/layout-alignment.tmpl.html
+++ b/docs/app/partials/layout-alignment.tmpl.html
@@ -79,6 +79,7 @@
     <div>
       <md-subheader>Alignment in Perpendicular Direction ({{layoutDemo.direction == 'column' ? 'horizontal' : 'vertical'}})</md-subheader>
       <md-radio-group ng-model="layoutDemo.crossAxis">
+        <md-radio-button value=""><em>none</em> (use default)</md-radio-button>
         <md-radio-button value="start">start</md-radio-button>
         <md-radio-button value="center">center</md-radio-button>
         <md-radio-button value="end">end</md-radio-button>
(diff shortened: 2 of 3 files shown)


ThomasBurleson:
@ngraef - this is not ideal. Allowing a mainAxis-only is fine, but the style should always fallback to defaults:

*  main-axis === 'start'
*  cross-axis === 'stretch'

If you update the demos accordingly, I will merge. 

> Meanwhile, this is already in place in the generation-2 Layout engine coming in January.

ngraef:
@ThomasBurleson I'm confused about what needs work. This doesn't change anything about the default values or the fallback behavior during parsing.

ThomasBurleson:
@ngraef - This PR is incomplete:

* Layout.js needs updated with *layout-align* improvements for fallbacks/defaults
* `$scope.layoutAlign` should **not** be checking for a blank crossAxis (that should be done in fallbacks).

> I will curl your PR and fix.

ngraef:
@ThomasBurleson 

* layout.js already has a fallback to default values in `extractAlignAxis`:

    ```javascript
    if ( ALIGNMENT_MAIN_AXIS.indexOf(axis.main) < 0 )   axis.main = "start";
    if ( ALIGNMENT_CROSS_AXIS.indexOf(axis.cross) < 0 ) axis.cross = "stretch";
    ```
I'm not sure what other improvements you want.

* My change to `$scope.layoutAlign` simply omits the trailing whitespace if the new `none (use default)` option is selected in the demo. I suppose a trim might be better:

    ```javascript
    return ($scope.layoutDemo.mainAxis + ' ' + $scope.layoutDemo.crossAxis).trim();
    ```
This was done mostly for aesthetic reasons, because the demo would previously display `layout-align="space-around "`. However, it also helped me test my changes to layout.js, because `layout-align="space-around "` and `layout-align="space-around"` were being parsed differently, which is exactly the bug this PR fixes.

ThomasBurleson:
@ngraef - I think my current Flu is confusing me; since I wrote 

```js
    if ( ALIGNMENT_MAIN_AXIS.indexOf(axis.main) < 0 )   axis.main = "start";
    if ( ALIGNMENT_CROSS_AXIS.indexOf(axis.cross) < 0 ) axis.cross = "stretch";
```

Thank you for the clear and reasonable follow-up. 
This looks good and I will merge soon.

ngraef:
@ThomasBurleson - No problem. Thanks for reviewing.

Issue #5981, opened 2015-11-30
md-dialog padding issue windows phone 8.1


graphefruit:
Hi Team,

in WP 8.1 the dialog doesn't padding the content of an confirm or alert box.

![popup](https://cloud.githubusercontent.com/assets/6260244/11465550/04145804-973a-11e5-9d8d-f362dc46305d.png)

If I modify md-dialog-content with layout="column" its working.
```
<md-dialog-content tabindex="-1" class="md-dialog-content layout-column"> </md-dialog-content>
```
This can be reproduced by taking IE and set:
- Document-Mode: IE 10
- Browse profile: Windows Phone 
- Useragent: IE10

The emulator I'm using is Windows Phone 8.1, which should work with IE 11 engine (seems like not correctly?).

![wp81](https://cloud.githubusercontent.com/assets/6260244/11465628/ae6e3c98-973a-11e5-87c0-74788086aa1c.png)

Would be awesome if you could fix this.

Thanks & Greetings
Graphefruit

Belicosus:
I'm using this fix on my current projet:

    md-dialog-content {
        display: inline-block;
    }

PR #6008, opened 2015-12-01
fix(mdDialog): chang display mode to inline-block

Belicosus:
Fixes the padding issue descriped in issue #5981


ThomasBurleson:
Will not merge.

PR #6016, opened 2015-12-01
Incomplete installation instructions

angelxmoreno:
I was a bit annoyed I had to hunt this down. I honestly was under the impression that adding the CSS file would be enough to get started just like many other UI frameworks. I was wrong; though not mentioning the module to add in the instructions those give that same impression.

### DIFF
--- a/README.md
+++ b/README.md
@@ -160,3 +160,9 @@ pull directly from the distribution GitHub
   </body>
 ```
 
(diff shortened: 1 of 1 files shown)


angelxmoreno:
I signed it!

ThomasBurleson:
From the development team's perspective this seems obvious. Obviously not the case... good catch.

PR #6017, opened 2015-12-01
fix(toast): Hide scrollbars during animation.

### LINKED ISSUE #2936: md-toast makes scroll bar appear during animation when positionned bottom
Hi guys,

Thanks for your constant enhancement of angular material design.

As you can see on [CodePen](http://codepen.io/anon/pen/YXGZQx) where i reused the [doc demo](https://material.angularjs.org/#/demo/material.components.toast), the scroll bar appears during the appear/disappear animation of a `md-toast` component positioned at bottom.
It is certainly caused by the fact that the `md-toast` component is first created at position *documents' height + its own height* and then it goes up a position *documents' height*.
On the other hand, this event doesn't appear when the toast is positioned top even if it's animation shows him appearing from a higher position.

Thank you, 

Izcalli

topherfangio:
During animation, certain browsers would show scroll bars if the
toast was positioned at the bottom of the parent container.

Fix by adding a class to the parent which sets the overflow to
hidden only during the animations.

Also, update documentation with notes about positioning and
recommendations.

Fixes #2936.

### DIFF
--- a/src/components/toast/toast.js
+++ b/src/components/toast/toast.js
@@ -40,6 +40,41 @@ function MdToastDirective($mdToast) {
   * - For a toast action, use element with class `md-action`.
   * - Add the class `md-capsule` for curved corners.
   *
+  * ## Parent container notes
+  *
+  * The toast is positioned using absolute positioning relative to it's first non-static parent
+  * container. Thus, if the requested parent container uses static positioning, we will temporarily
+  * set it's positioning to `relative` while the toast is visible and reset it when the toast is
+  * hidden.
+  *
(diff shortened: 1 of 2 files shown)


topherfangio:
@jelbourn @robertmesserle @ThomasBurleson Please review when you have a moment and provide feedback on doc updates and any potential issues you see with this fix.

robertmesserle:
@topherfangio I think a better solution would be to have a wrapper div that the toast animates within.  So when it is below the fold, it will be hidden by that div and not trigger scrolling.

PR #6019, opened 2015-12-02
fix(dialog): trap focus within dialog. Fixes #4105.

### LINKED ISSUE #4105: mdDialog: focus trap broken, incorrectly allowing keyboard navigation behind the modal

jelbourn:
@ThomasBurleson for review

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -421,6 +421,8 @@ function MdDialogDirective($$rAF, $mdTheming, $mdDialog) {
  */
 
 function MdDialogProvider($$interimElementProvider) {
+  // Elements to capture and redirect focus when the user presses tab at the dialog boundary.
+  var topFocusTrap, bottomFocusTrap;
 
   return $$interimElementProvider('$mdDialog')
     .setDefaults({
@@ -588,9 +590,12 @@ function MdDialogProvider($$interimElementProvider) {
       options.unlockScreenReader();
       options.hideBackdrop(options.$destroy);
 
+      // Remove the focus traps that we added earlier for keeping focus within the dialog.
+      topFocusTrap.parentNode.removeChild(topFocusTrap);
+      bottomFocusTrap.parentNode.removeChild(bottomFocusTrap);
+
       // For navigation $destroy events, do a quick, non-animated removal,
       // but for normal closes (from clicks, etc) animate the removal
-
       return !!options.$destroy ? detachAndClean() : animateRemoval().then( detachAndClean );
 
       /**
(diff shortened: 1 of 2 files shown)


jelbourn:
Also FYI @marcysutton

src/components/dialog/dialog.spec.js
```diff
@@ -1172,6 +1176,49 @@ describe('$mdDialog', function() {
...
+      // md-dialog element. Firefox is problematic here, as calling element.focus() inside of
+      // a focus event listener seems not to immediately update the document.activeElement.
+      // This is a behavior better captured by an e2e test.
+
+      $mdDialog.hide();
+      runAnimation();
+
+      // All of the focus traps should be removed when the dialog is closed.
+      focusTraps = document.querySelectorAll('.md-dialog-focus-trap');
+      expect(focusTraps.length).toBe(0);
```
ThomasBurleson: Nice!

src/components/dialog/dialog.js
```diff
@@ -823,6 +828,23 @@ function MdDialogProvider($$interimElementProvider) {
...
+
+      bottomFocusTrap = topFocusTrap.cloneNode(false);
+
+      // When focus is about to move out of the dialog, we want to intercept it and redirect it
+      // back to the dialog element.
+      var focusHandler = angular.bind(element, element.focus);
+      topFocusTrap.addEventListener('focus', focusHandler);
+      bottomFocusTrap.addEventListener('focus', focusHandler);
+
+      element[0].parentNode.insertBefore(topFocusTrap, element[0]);
```
ThomasBurleson: Assume this is because `element.prepend( )` is not available without jQuery ?

src/components/dialog/dialog.js
```diff
@@ -823,6 +828,23 @@ function MdDialogProvider($$interimElementProvider) {
...
+      element.append(bottomFocusTrap);
```
ThomasBurleson: You are appending to the element's parent ? or the element's child list ?

src/components/dialog/dialog.js
```diff
@@ -588,9 +590,12 @@ function MdDialogProvider($$interimElementProvider) {
       options.unlockScreenReader();
       options.hideBackdrop(options.$destroy);
 
+      // Remove the focus traps that we added earlier for keeping focus within the dialog.
+      topFocusTrap.parentNode.removeChild(topFocusTrap);
```
ThomasBurleson: Perhaps this is more consistent:

```js
var parentNode = element[0].parentNode;  // assumes same parent; consistent with append/prepend use.

parentNode.removeChild(topFocusTrap);
parentNode.removeChild(bottomFocusTrap);
```

src/components/dialog/dialog.js
jelbourn: It's not that. I actually _want_ the behavior of `insertBefore`, which is different than `prepend` (which is available in jqlite).

src/components/dialog/dialog.js
jelbourn: Each of the focus traps has different parents. The top is a sibling to the dialog element, the bottom is a child of it.

src/components/dialog/dialog.js
jelbourn: The top is a sibling to the dialog element, the bottom is a child of it.

ThomasBurleson: I thought they were different (see below) but could not quickly determine the rationale. Perhaps you can  add one sentence explaining why the parents are different.

PR #6024, opened 2015-12-02
update(layout): add corrections and warnings for IE-related issues

### LINKED ISSUE #5084: IE 11 compatibility?
When playing with offical MD demo I found that it works well in Chrome but in IE there are a couple of issues. The attached screenshot 1 shows one issue. It is Checkbox demo page. Another issues comes from Subheaders/Late Message. Check screenshot 2. I use Win8.1. 

image

image

ThomasBurleson:
* Some HTML elements can't be flex containers: read https://github.com/philipwalton/flexbugs#9-some-html-elements-cant-be-flex-containers for details.
* Add warnings (with url) for above issue.
* update checkbox demo with workaround.

Fixes #5084.

### DIFF
--- a/src/core/services/layout/layout.js
+++ b/src/core/services/layout/layout.js
@@ -229,7 +229,9 @@
           element.addClass(className);
 
           return function( scope, element ) {
-            // Wait while layout injectors configure, then uncload
+            // Wait while layout injectors configure, then uncloak
+            // NOTE: $rAF does not delay enough... and this is a 1x-only event,
+            //       $timeout is acceptable.
             $timeout( function(){
               element.removeClass(className);
             }, 10, false);
--- a/src/core/services/layout/layout.js
+++ b/src/core/services/layout/layout.js
@@ -1,7 +1,7 @@
 (function() {
   'use strict';
 
-  var $mdUtil, $interpolate;
+  var $mdUtil, $interpolate, $log;
 
   var SUFFIXES = /(-gt)?-(sm|md|lg)/g;
   var WHITESPACE = /\s+/g;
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@jelbourn - can you review please.

> The work around suggest "The simple solution to this problem is to use a wrapper element that can be a flex container (like a <div>) directly **inside** of the element that can't.". I wrapped one **outside**... but it seems to work fine for our demo use case.

src/core/services/layout/layout.js
```diff
@@ -367,6 +375,30 @@
   }
 
   /**
+   * Centralize warnings for known flexbox issues (especially IE-related issues)
+   */
+  function validateAttributeUsage(className, attr, element, $log){
```
jelbourn: I would lean towards not actually doing this with a warning, and instead just having it in the documentation. This adds payload size and runtime cost for something that isn't necessary for the feature to function.

ThomasBurleson: My worry is that many developers do not read the documentation details.
This is a nice backup for the dev console.

jelbourn: The problem is that it adds a run-time cost to every instance of a layout directive. It is small, but it could add up (entirely speculation). 

(One of the things I want to do for Angular 2 is have code blocks that we can run in dev mode but strip for production.)

ThomasBurleson: Currently this only triggers the logic for 'flex':

```js
switch(className.replace(SUFFIXES,"")) {
      case "flex":
```

jelbourn: Yeah, `className.replace(SUFFIXES,"")` isn't free.

jb29:
`$log` is undefined. See my line comment in e82ed42.

PR #6034, opened 2015-12-02
update(core): detect incompatible ngTouch usages

### LINKED ISSUE #6032: gestures:  Detect ngTouch usages
Detect if ngTouch module is loaded and warn about incompatibility with ngTouch and mdGestures.

* Update Docs section [Getting Started](https://material.angularjs.org/latest/getting-started)
  *  Add "Unsupported Options"

ThomasBurleson:
Fixes #6032.

### DIFF
--- a/src/core/core.js
+++ b/src/core/core.js
@@ -10,7 +10,24 @@ angular
     'material.core.gestures',
     'material.core.theming'
   ])
-  .config(MdCoreConfigure);
+  .config(MdCoreConfigure)
+  .run(DetectNgTouch);
+
+
+/**
+ * Detect if the ng-Touch module is also being used.
+ * Warn if detected.
+ */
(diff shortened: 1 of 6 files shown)


ThomasBurleson:
@jelbourn - can you review plz.

src/core/core.js
```diff
@@ -10,7 +10,27 @@ angular
...
+  .config(MdCoreConfigure)
+  .run(DetectNgTouch);
+
+
+/**
+ * Detect if the ng-Touch module is also being used.
+ * Warn if detected.
+ */
+function DetectNgTouch($log) {
+  try {
```
jelbourn: I think a better approach would be to use `$injector.has('$swipe')`, then you wouldn't need a try-catch.

PR #6036, opened 2015-12-02
Fix the doc - under the hood

AminaG:
https://material.angularjs.org/HEAD/Theming/05_under_the_hood:

"Angular Material dynamically generates CSS for registered themes by injecting several **`<style>`** tags into the `<head>`."

Not `<script>` tags


AminaG:
I signed it!

PR #6044, opened 2015-12-02
fix(select): fixes positioning of select menu and sets it to append t…

robertmesserle:
…o the body again


robertmesserle:
@jelbourn @ThomasBurleson @rschmukler @topherfangio please review.

PR #6060, opened 2015-12-03
fix(cards): applying zero margin only on first and last p elements

EladBezalel:
Accidentally applied zero margin to every `p` in the card, which caused multiple `p` elements to stick together.
This fix is only applies zero margin to the first and last element


jelbourn:
@EladBezalel can you go ahead and remove the `iit` in this change?

LGTM

PR #6065, opened 2015-12-03
feat(icon): observing `mdFontIcon` and `mdFontSet` attributes

### LINKED ISSUE #4961: [md-icon] bind md-font-icon
Using `md-icon` without binding `md-font-icon` like:

```html
<md-icon md-font-icon="my-icon"></md-icon>
```
works as expected,
but where trying to bind `md-font-icon` like:

```html
<md-icon md-font-icon="{{ model.icon_name }}"></md-icon>
```

doesn't, when inspecting the DOM using browser debug tools I can see the `md-font-icon` value as expected, but the icon is not shown

Tested in a directive using `ng-model`

EladBezalel:
For when the icon has changed.

fixes #4961

### DIFF
--- a/src/components/icon/js/iconDirective.js
+++ b/src/components/icon/js/iconDirective.js
@@ -182,8 +182,14 @@ function mdIconDirective($mdIcon, $mdTheming, $mdAria ) {
    */
   function postLink(scope, element, attr) {
     $mdTheming(element);
+    var lastIcon = scope.fontIcon;
+    var lastSet = $mdIcon.fontSet(scope.fontSet);
 
     prepareForFontIcon();
+
+    attr.$observe('mdFontIcon', fontIconChanged);
+    attr.$observe('mdFontSet', fontIconChanged);
+
 
     // If using a font-icon, then the textual name of the icon itself
     // provides the aria-label.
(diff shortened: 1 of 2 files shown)


src/components/icon/js/iconDirective.js
```diff
@@ -182,9 +182,15 @@ function mdIconDirective($mdIcon, $mdTheming, $mdAria ) {
    */
   function postLink(scope, element, attr) {
     $mdTheming(element);
+    var lastIcon = scope.fontIcon;
+    var lastSet = $mdIcon.fontSet(scope.fontSet);
 
     prepareForFontIcon();
 
+    attr.$observe('mdFontIcon', fontIconChanged);
```
ThomasBurleson: what if the attribute `md-font-icon` or `md-font-set` is not used/available in the markup?

ThomasBurleson:
@EladBezalel - add comment/question.

src/components/icon/js/iconDirective.js
EladBezalel: I did a quick search on the project and came with a lot of places we don't check it
Basically when we use `$observe` we don't check if the attribute has value, only on `slider` and that's because we want to apply different function if there's no attribute

ThomasBurleson:
Closing as out-dated and *not essential* to core functionality. 
Will not merge.

AdriVanHoudt:
@ThomasBurleson if I do a rebase and do a new PR ready for merge would you merge it then?

ThomasBurleson:
@AdriVanHoudt - assuming tests pass and the code review pass... then Yes! Thx.

Issue #5435, opened 2015-10-29
md-select: md-select-menu positioned above control 


fridolin-koch:
I'm using the async variant of `md-select`. Instead of the menu appearing directly over the `md-select` it appears a few pixels above. The bug only occurs if it contains only a few elements.

![screen shot 2015-10-29 at 18 37 00](https://cloud.githubusercontent.com/assets/1880828/10826875/153ca19a-7e6c-11e5-8700-77fb1cc0c2b5.png)

This bug seems to be related to my layout, am I using the containers wrong or is this a bug?

I've created a codepen http://codepen.io/anon/pen/VvdZYx

EladBezalel:
Did you tried it outside of code pen?
If we change the `body` element in your example to `div` it's [working](http://codepen.io/EladBezalel/pen/ZbRegQ)
If you didn't tried it outside please check, i'm sure it's working.

fridolin-koch:
Yep, the issue comes from the app I'm currently developing. I tried to replicate the problem inside Codepen to make the issue more feasible.

The actual code looks like that:

```html
<body layout="column">
  <stm-navbar></stm-navbar>
  <div layout="row" flex>
    <!-- Sidebar -->
    <stm-sidenav layout="column"></stm-sidenav>
    <!-- Content container -->
    <div layout="column" class="ui-view-container" flex style="overflow: auto">
      <md-progress-linear ng-show="$appLoading" md-mode="indeterminate"></md-progress-linear>
      <stm-breadcrumbs></stm-breadcrumbs>
      <div ui-view></div>
    </div>
  </div>
</body>
```

fridolin-koch:
It seems like the issue is not related to a certain layout or browser:

Chrome:

![chrome_1](https://cloud.githubusercontent.com/assets/1880828/10932660/cc51ab7c-82d2-11e5-84b4-33f0878448d0.png)
![chrome_2](https://cloud.githubusercontent.com/assets/1880828/10932659/cc519f1a-82d2-11e5-9a4b-4ae3fdd18347.png)

Firefox:

![firefox_1](https://cloud.githubusercontent.com/assets/1880828/10932657/cc4f67d6-82d2-11e5-85c5-e8e62238c580.png)
![firefox_2](https://cloud.githubusercontent.com/assets/1880828/10932658/cc507a04-82d2-11e5-8863-23e72f307bfb.png)

Here is the example code used: https://gist.github.com/fridolin-koch/2c9fc5b24a5f558a9ca1

[...]

PR #6071, opened 2015-12-04
fix(select): fix bugs introduced by keeping md-select-menu in DOM

rschmukler:
This PR addresses bugs that were introduced by keeping `md-select-menu` in the DOM at all times.

Previously `md-select-menu` was removed from the DOM and compiled by the `postLink` function of `md-select`. Now it stays in the DOM. This means that both `md-select-menu`'s and `md-option`'s post link functions fire before `md-select`'s. This broke several assumptions that were made in the code.

Obviously, our test suite should have catch these errors, but did not (and in fact, an earlier attempt at the change) broke a good number of tests. This is because tests were targeting individual components (for example, an `md-select-menu` instead of an `md-select` which contains a `md-select-menu`. I have refactored many of the tests to take a much more e2e approach which test the components within the context that they are actually in the DOM. This means use of helpers for setting up state (for example, `openSelect(el)` and `clickOption(el, index)`. Over all, this should make our tests much more robust at catching (and preventing) regressions like this in the future.

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -418,6 +418,7 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
           selectContainer[0].setAttribute('class', value);
         }
         selectMenuCtrl = selectContainer.find('md-select-menu').controller('mdSelectMenu');
+        selectMenuCtrl.init(ngModelCtrl, attr.ngModel);
       }
 
       function handleKeypress(e) {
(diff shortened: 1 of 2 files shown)


jelbourn:
cc @robertmesserle does this seem related to the issues Googlers have been seeing with select / menu?

ThomasBurleson:
@jelbourn - I think this should be in today's build

jelbourn:
This wasn't the same issue Rob had seen.

LGTM insofar as I understand menu.

jlmagee:
@rschmukler  you must have some familiarity with this code since you change the calculation of the variable top herein

         top = Math.floor(targetRect.top + targetRect.height / 2 - centeredRect.height / 2 -
-            centeredRect.top + contentNode.scrollTop) + 5;
+            centeredRect.top + contentNode.scrollTop) + 4;

interesting that it is never subsequently referenced. Discovered in my research on #5435 

Hoping for help there

PR #6072, opened 2015-12-04
fix(docs): update getting started CodePen to be usable again

Splaktar:
Re-enable the JS, HTML, CSS tabs as well as the 'Edit this on CodePen' link

Before:
screen shot 2015-12-03 at 10 03 35 pm

After:
screen shot 2015-12-03 at 10 03 45 pm


ThomasBurleson:
Fixed with SHA f6e97a02a9c88b1695094a8a85a11958924e4c61

PR #6087, opened 2015-12-04
fix(virtualRepeat): fix sizer not shrinking when items reduce in number.

### LINKED ISSUE #4435: md-virtual-repeat scroll size not updated
If the number of items shrinks, and scrollOffset > 0, it scrolls to the top, but doesn't update the scroll size.  Hence, you're left with a dirty scrollbar.

This is happening here: https://github.com/angular/material/blob/master/src/components/virtualRepeat/virtual-repeater.js#L543

Not familiar with this code, but perhaps it needs to call `this.container.setScrollSize(itemsLength * this.itemSize);`

In addition to `this.container.resetScroll();`

Easy fix?

jelbourn:
Fixes #4435


topherfangio:
@jelbourn LGTM :-)

PR #6091, opened 2015-12-04
fix(input): input error messages visible on form submit

### LINKED ISSUE #5752: RC4: ng-messages isn't shown when form is submitted, but input is untouched
Try and push the submit button in this [CodePen](http://codepen.io/chr22/pen/QjYdPX) and see that the message is set into DOM but its opacity is 0.

It will need something along the lines of: 
```CSS
form.ng-submitted [ng-messages] {
    opacity: 1;
}
```

EladBezalel:
Added check for parent form `$submitted` attribute.

Notice that `novalidate` attribute should be applied on parent form in order to make it work (Browser preventing submit when not specified)

fixes #5752

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -212,8 +212,9 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
     }
 
     var isErrorGetter = containerCtrl.isErrorGetter || function() {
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@jelbourn - thoughts?

jelbourn:
Yes, I think this should be the behavior.

@EladBezalel This change needs a unit test.

PR #6092, opened 2015-12-04
fix(ripple): calculate relative coordinates from ripple target.

### LINKED ISSUE #5917: mdInkRipple: fab button with md-icon has wrong ripple placement
The ripple is incorrectly placed when using a fab button that contains an md-icon.  A plain text fab button will have the correct ripple placement.  The ripple will incorrectly start from the top left of the button unless clicking near the bottom right of the fab.

**Browsers and OS:** Opera 33 / Chrome 46 - Windows 7 x64
**Reproduce:** [this codepen](http://codepen.io/mckenzielong/pen/BogoLj) click on the good or test button near the center to get a feel for how the ripple should be displayed.  Click on one of the icon fabs near the center and notice that the ripple will start from the top left.
**Suggested Fix:** haven't really had a chance to look... is offsetX/Y behaviour stable?

DevVersion:
If there are any children the `offsetX` and `offsetY` are wrong, because it's not relative to the ripple target.
We can fix that by calculating `layerX` and `layerY`. As the normal `layerX` and `layerY` is not a W3C Standard (and is deprecated) we need to calculate our coordinates our self.

Fixes #5917

### DIFF
--- a/src/core/services/ripple/ripple.js
+++ b/src/core/services/ripple/ripple.js
@@ -259,7 +259,17 @@ InkRippleCtrl.prototype.handleMousedown = function (event) {
   if (this.options.center) {
     this.createRipple(this.container.prop('clientWidth') / 2, this.container.prop('clientWidth') / 2);
   } else {
-    this.createRipple(event.offsetX, event.offsetY);
+
+    // We need to calculate the relative coordinates if the target is a sublayer of the ripple element
+    if (event.srcElement !== this.$element[0]) {
+      var layerRect = this.$element[0].getBoundingClientRect();
+      var layerX = event.clientX - layerRect.left;
+      var layerY = event.clientY - layerRect.top;
+
+      this.createRipple(layerX, layerY);
+    } else {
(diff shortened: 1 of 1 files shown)


DevVersion:
That's how it look after the fix

ThomasBurleson:
@robertmesserle - seems like a good fix to have in today's build. Thoughts?

robertmesserle:
@ThomasBurleson LGTM

src/core/services/ripple/ripple.js
```diff
@@ -259,7 +259,16 @@ InkRippleCtrl.prototype.handleMousedown = function (event) {
   if (this.options.center) {
     this.createRipple(this.container.prop('clientWidth') / 2, this.container.prop('clientWidth') / 2);
   } else {
-    this.createRipple(event.offsetX, event.offsetY);
+
+    if (event.srcElement !== this.$element[0]) {
```
jelbourn: Could you add a comment explaining the reasoning here?

DevVersion:
Done @jelbourn

PR #6100, opened 2015-12-05
feat(layout): add flex noshrink attribute to prevent shrinking

### LINKED ISSUE #6067: layout: add flex="noshrink" option
I've just read this thread: #2632

And based on this specification: https://gist.github.com/ThomasBurleson/88152ec57c9133dec57a

...there is no way to make an element **grow by default without shrinking** when viewport's height is too small.
The only option to prevent shrinking is `flex="none"` but then the element won't grow.

My sample code:
```html
<md-content layout="column">
  <div ng-style="{'background-image': 'url(' + getBannerFromTheme() + ')'}" alt="Branding banner"></div>
  <div flex layout="column" layout-align="center center" class="md-padding">
    [...]
  </div>
</md-content>
```

The first `<div>` has a fixed size and I want the second one to take the rest of the space. But when viewport height is too small, I don't want the second `<div>` to shrink because it breaks the design. I just want it to be hidden and accessible when scrolling, to keep the same design.
So currently I have to set `flex-shrink: 0` manually.

Having something like `[flex="fixed"] { flex: 1 0 auto; }` or `[flex="noshrink"] { flex: 1 0 auto; }` would be really handy (it's the opposite of _initial_).

DevVersion:
Fixes #6067

### DIFF
--- a/docs/app/partials/layout-children.tmpl.html
+++ b/docs/app/partials/layout-children.tmpl.html
@@ -146,6 +146,10 @@
       <td>flex="grow"</td>
       <td>Will grow and shrink as needed. Starts with a size of 100%. Same as <code>flex="100"</code>.</td>
     </tr>
+    <tr>
+      <td>flex="noshrink"</td>
+      <td>Will grow as needed, but won't shrink. Starts with a size based on it's <code>width</code> and <code>height</code> values.</td>
+    </tr>
   </table>
 
 
--- a/src/core/services/layout/layout.js
+++ b/src/core/services/layout/layout.js
@@ -6,7 +6,7 @@
   var SUFFIXES = /(-gt)?-(sm|md|lg)/g;
   var WHITESPACE = /\s+/g;
 
-  var FLEX_OPTIONS = ['grow', 'initial', 'auto', 'none'];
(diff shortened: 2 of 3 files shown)


ThomasBurleson:
:+1:

DevVersion:
Is that what you expect? @julienpa

docs/app/partials/layout-children.tmpl.html
```diff
@@ -146,6 +146,10 @@
       <td>flex="grow"</td>
       <td>Will grow and shrink as needed. Starts with a size of 100%. Same as <code>flex="100"</code>.</td>
     </tr>
+    <tr>
+      <td>flex="fixed"</td>
+      <td>Will grow as needed, but won't shrink. Starts with a size based on it's <code>width</code> and <code>height</code> values.</td>
+    </tr>
```
DevVersion: Added the attribute here too @ThomasBurleson

julienpa:
Nice :smiley: 
Good job!

DevVersion:
Done @ThomasBurleson

ThomasBurleson:
@jelbourn - I would like to add this to today's build. Thoughts ?

jelbourn:
LGTM

PR #6101, opened 2015-12-05
fix(sidenav): trigger the focus handler for autofocus target too.

### LINKED ISSUE #5665: md-autofocus does not work when md-autocomplete inside a md-sidenav
I want to make my md-autocomplete input autofocus everytime when I open a md-sidenav bar. But it doesn't seem to work:

http://codepen.io/anon/pen/bVQMmg?editors=101

DevVersion:
This fixes the current issue with autofocus for other directives like `autocomplete`

Fixes #5665

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -299,8 +299,9 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
               ])
               .then(function() {
                 // Perform focus when animations are ALL done...
-                if (scope.isOpen) {
-                  focusEl && focusEl.focus();
+                if (scope.isOpen && focusEl) {
+                  focusEl.focus();
+                  focusEl.triggerHandler('focus');
                 }
               });
     }


ThomasBurleson:
@robertmesserle, @jelbourn - Can you review for today's build?

src/components/sidenav/sidenav.js
```diff
@@ -299,8 +299,9 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
               ])
               .then(function() {
                 // Perform focus when animations are ALL done...
-                if (scope.isOpen) {
-                  focusEl && focusEl.focus();
+                if (scope.isOpen && focusEl) {
+                  focusEl.focus();
+                  focusEl.triggerHandler('focus');
```
jelbourn: I don't understand why this is necessary. Invoking `focus` on an element should always trigger any event listeners to be called. @DevVersion can you explain?

DevVersion: The same question I asked myself too. But as I figured out, `focus()` can't be applied to all type of elements. And because the sidenav is a custom angular directive it looks like, they can't be focused through the `focus()` function. That's why I trigger that focus handlers manually, and its working perfect. In my other PR, which already got reviewed by Elad, I'm using the same strategy to get that running (see #5943).

**Quote MDN**
- `The HTMLElement.focus() method sets focus on the specified element, if it can be focused.`

jelbourn: Custom elements simply need a `tabindex` in order to receive focus:
http://codepen.io/jelbourn/pen/MKYpJm?editors=101

DevVersion: All right, thanks for that codepen demonstration.

PR #6132, opened 2015-12-07
fix(select): set the focus to false if the model is undefined or if he model was added without focus.

### LINKED ISSUE #6122: Demos, select validation fails on Clear
Official doc : https://material.angularjs.org/latest/demo/select
In the last sample (Validations), when you set a value then clic on CLEAR the select is displayed with a validation error (Required) and the label is shown two times (Top + Placeholder).

Expected result : no validation error, label only in Placeholder.

I tried to call `$setPristine` and `$setUntouched`. It fixes the validation error but the label is still here.
http://codepen.io/anon/pen/BjyzYM

DevVersion:
We should set the focus to false if the model is undefined or if he model was added without focus.

Fixes #6122

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -245,6 +245,7 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
         syncLabelText();
         syncAriaLabel();
         inputCheckValue();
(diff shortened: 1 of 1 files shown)


DevVersion:
´See #6207

PR #6135, opened 2015-12-07
fix(autocomplete): always set tabindex to allow receiving focus.

### LINKED ISSUE #5665: md-autofocus does not work when md-autocomplete inside a md-sidenav

DevVersion:
At the moment we only set the `tabindex` if we specify a `tabindex` attribute.

That's why the autocomplete won't receive focus on `sidenav` autofocus.

Closes #6101 Fixes #5665

Please review @jelbourn

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -98,6 +98,29 @@ describe('<md-autocomplete>', function() {
       element.remove();
     }));
 
+    it('should allow receiving focus on the autocomplete', function() {
+      var scope = createScope(null, {inputId: 'custom-input-id'});
+      var template = '<md-autocomplete ' +
+            'md-input-id="{{inputId}}" ' +
+            'md-selected-item="selectedItem" ' +
+            'md-search-text="searchText" ' +
+            'md-items="item in match(searchText)" ' +
+            'md-item-text="item.display" ' +
+            'placeholder="placeholder">' +
+          '<span md-highlight-text="searchText">{{item.display}}</span>' +
+        '</md-autocomplete>';
+      var element = compile(template, scope);
+      var focusSpy = jasmine.createSpy('focus');
+
+      document.body.appendChild(element[0]);
+
+      element.on('focus', focusSpy);
+
+      element.focus();
+
+      expect(focusSpy).toHaveBeenCalled();
+    });
+
     it('should allow you to set an input id without floating label', inject(function() {
       var scope = createScope(null, {inputId: 'custom-input-id'});
       var template = '\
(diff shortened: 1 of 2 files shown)


src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -156,9 +156,7 @@ function MdAutocomplete () {
       // Set our variable for the link function above which runs later
       hasNotFoundTemplate = noItemsTemplate ? true : false;
 
-      if (attr.hasOwnProperty('tabindex')) {
-        element.attr('tabindex', '-1');
-      }
+      element.attr('tabindex', '-1');
```
jelbourn: It looks like someone originally went to do the right thing and forgot a `!`. It _should_ be:
```
    if (!attrs['tabindex']) {
      $element.attr('tabindex', '-1');
    }
```

DevVersion:
@jelbourn Added test for that + fixed it

src/components/autocomplete/autocomplete.spec.js
```diff
@@ -98,6 +98,31 @@ describe('<md-autocomplete>', function() {
       element.remove();
     }));
 
+    it('should allow receiving focus on the autocomplete', function() {
+      var scope = createScope(null, {inputId: 'custom-input-id'});
+      var template = '\
```
jelbourn: Prefer string concatenation instead of using `\`

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -156,9 +156,7 @@ function MdAutocomplete () {
       // Set our variable for the link function above which runs later
       hasNotFoundTemplate = noItemsTemplate ? true : false;
 
-      if (attr.hasOwnProperty('tabindex')) {
-        element.attr('tabindex', '-1');
-      }
+      if (!tabindex) element.attr('tabindex', '-1');
```
jelbourn: This doesn't work because `0` is a valid tabindex that you don't want to overwrite.

src/components/autocomplete/autocomplete.spec.js
DevVersion: I'm just trying my best in using the best code conventions. The tests above also use `\`.
But I will use string concatenation, because you wish that :smile:

src/components/autocomplete/js/autocompleteDirective.js
DevVersion: Okay, that's now a bit embarrassing ... sorry for that (too much tests this week)

src/components/autocomplete/autocomplete.spec.js
jelbourn: Yeah- typically good, but I've been hoping to eliminate the `\` stuff over time.

DevVersion: Great, I will remember that

DevVersion:
Used your suggestions now @jelbourn

PR #6139, opened 2015-12-08
fix(virtualRepeat): sets initial size for virtual repeat when the fir…

### LINKED ISSUE #5826: md-autocomplete in md-dialog have huge height also with only few values
Hi,
I'm using md-autocomplete inside md-dialog and the md-virtual-repeat-container that shows the possible choises is too high when I have only one element.
It's the right behaviour?

Thanks
schermata 2015-11-19 alle 12 50 32

robertmesserle:
…st results require shrinking

closes #5826

### DIFF
--- a/src/components/virtualRepeat/virtual-repeater.js
+++ b/src/components/virtualRepeat/virtual-repeater.js
@@ -253,7 +253,8 @@ VirtualRepeatContainerController.prototype.sizeScroller_ = function(size) {
 VirtualRepeatContainerController.prototype.autoShrink_ = function(size) {
   var shrinkSize = Math.max(size, this.autoShrinkMin * this.repeater.getItemSize());
   if (this.autoShrink && shrinkSize !== this.size) {
-    if (shrinkSize < (this.originalSize || this.size)) {
+    var currentSize = this.originalSize || this.size;
+    if (!currentSize || shrinkSize < currentSize) {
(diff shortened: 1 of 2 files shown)


robertmesserle:
@jelbourn @kseamon

src/components/virtualRepeat/virtual-repeater.js
```diff
@@ -243,7 +243,7 @@ VirtualRepeatContainerController.prototype.sizeScroller_ = function(size) {
 VirtualRepeatContainerController.prototype.autoShrink_ = function(size) {
   var shrinkSize = Math.max(size, this.autoShrinkMin * this.repeater.getItemSize());
   if (this.autoShrink && shrinkSize !== this.size) {
-    if (shrinkSize < (this.originalSize || this.size)) {
+    if (shrinkSize < (this.originalSize || this.size) || !(this.originalSize || this.size)) {
```
jelbourn: Needs a unit test

jelbourn: Would it work if it was just 
```
if (shrinkSize < (this.originalSize || this.size) || !(this.originalSize))
```
?

kseamon: That does have a different meaning, but the way that looks is still pretty awkward.

PR #6140, opened 2015-12-08
fix(tooltip): always append to body and disappear on scroll

EladBezalel:
- Default animation origin has been set to bottom that mean the origin is `top center`

@jelbourn please review

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -58,9 +58,10 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
 
     var parent        = $mdUtil.getParentWithPointerEvents(element),
         content       = angular.element(element[0].getElementsByClassName('md-content')[0]),
-        current       = $mdUtil.getNearestContentElement(element),
-        tooltipParent = angular.element(current || document.body),
+        tooltipParent = angular.element(document.body),
         debouncedOnResize = $$rAF.throttle(function () { updatePosition(); });
+
+    if ($animate.pin) $animate.pin(element, parent);
 
     // Initialize element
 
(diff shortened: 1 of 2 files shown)


src/components/tooltip/tooltip.js
```diff
@@ -145,11 +146,17 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
       };
       var elementFocusedOnWindowBlur = false;
 
+      function windowScrollHandler() {
+        setVisible(false);
+      }
+      
       ngWindow.on('blur', windowBlurHandler);
       ngWindow.on('resize', debouncedOnResize);
+      document.addEventListener('scroll', windowScrollHandler, true);
```
jelbourn: Did you test this on Firefox? Scrolling behavior is slightly different there.

kseamon:
http://www.quirksmode.org/dom/events/scroll.html

Try listening on window?

kseamon:
Hacked together a workaround: https://github.com/angular/material/compare/master...kseamon:pr/6140?expand=1

jelbourn:
Workaround seems like it would be good. @EladBezalel are you going to apply it and add a unit test?

kseamon:
An explanation for the workaround (which should probably be added in a comment):

It seems like appending and/or repositioning the menu was triggering the scroll events (In FF only, which seems like buggy behavior but we're stuck with it). These events happen more or less right away, so nextTick fires after them.

EladBezalel:
This is not working entirely, when checking the checkbox it's still disappearing.

kseamon:
Well, it's another step towards hacky, but a timeout of 100ms instead of nextTick does the trick for the checkbox. Perhaps there's a longer animation going on when it triggers from the checkbox?

jelbourn:
What do you mean by checking the checkbox?

kseamon:
Better solution: remove transform: translate3d(0,0,0); from tooltip.scss. That seems to be the source of our FF misbehavior.

jelbourn:
@kseamon that sounds like the best

EladBezalel:
@jelbourn I'm on it

kseamon:
I have not dived in with the chrome tools to look at paint performance to see if this makes any difference, but adding "will-change: opacity, height, width;" does not trigger the misbehavior but probably replaces any performance gains that the transform was accomplishing.

src/components/tooltip/tooltip.js
michahell: This is probably causing a bug on firefox (at least in 1.09) where tooltips flash and immediately get removed...

PR #6141, opened 2015-12-08
feat(autocomplete): don't focus after clicking an item from the dropdown

gbaj:
Change autocomplete to lose focus after clicking an item. This is consistent with other Google properties (search, music, inbox).


jelbourn:
The return of focus back to the text input is extremely important for accessibility. Without this, the focus would drop back to the document body when the dropdown is gone, leaving screen reader users completely lost.

PR #6151, opened 2015-12-08
Many Select / Menu Fixes

rschmukler:
This branch fixes many of the issues introduced by the side-effects caused by the accessibility enhancements in RC6, as well as some issues that came to light while debugging the issues reported in RC6.

1) Fixes an issue where menu contents could go behind the menubar if menu's position was clamped (@jelbourn saw this as the "cropping" issue we experienced while debugging accessibility) (089e33b)

2) Fix menubar breaking after a menu has been opened and closed with keyboard controls. The first open works as intended, but the following one will break. (d034e14)

3) Fix placing open selects and menus inside their parent selects. They are now moved to the `body` element when they are opened. This resolves positioning inside of dialogs, toolbars, etc. (2ea692b, cd75b22)

4) Fix accessibility issues for menubar menus. Turns out `ngAria` was setting a bunch of aria-related attributes because of our `md-menu-item` having a `ngModel` binding. This could lead to weird situations such as a menuitemcheckbox inside of a `role=checkbox`. This is now resolved and tested in Chrome w/ NVDA.  (9fe3a57) @jelbourn if you can please test w/ Firefox / JAWS

### DIFF
--- a/src/components/menuBar/js/menuItemController.js
+++ b/src/components/menuBar/js/menuItemController.js
@@ -22,11 +22,23 @@ MenuItemController.prototype.init = function(ngModel) {
     this.mode  = $attrs.type;
     this.iconEl = $element[0].children[0];
     this.buttonEl = $element[0].children[1];
-    if (ngModel) this.initClickListeners();
+    if (ngModel) {
+      // Clear ngAria set attributes
+      this.initClickListeners();
+    }
   }
 };
 
+MenuItemController.prototype.clearNgAria = function() {
+  var el = this.$element[0];
+  var clearAttrs = ['role', 'tabindex', 'aria-invalid', 'aria-checked'];
+  angular.forEach(clearAttrs, function(attr) {
+    el.removeAttribute(attr);
+  });
+};
+
 MenuItemController.prototype.initClickListeners = function() {
+  var self = this;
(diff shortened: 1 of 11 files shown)


src/components/menuBar/js/menuItemController.js
```diff
@@ -22,11 +22,23 @@ MenuItemController.prototype.init = function(ngModel) {
...
     this.buttonEl = $element[0].children[1];
-    if (ngModel) this.initClickListeners();
+    if (ngModel) {
+      // Clear ngAria set attributes
+      this.initClickListeners();
+    }
   }
 };
 
+MenuItemController.prototype.clearNgAria = function() {
```
jelbourn: Could use a comment here explaining what ngAria does and why it's a problem. (which is my understanding of what's happening here)

src/components/menuBar/js/menuItemDirective.js
```diff
@@ -10,6 +10,7 @@ angular
 function MenuItemDirective() {
   return {
     require: ['mdMenuItem', '?ngModel'],
+    priority: 210,
```
jelbourn: Could use a comment explaining the non-default priority

jelbourn:
Overall LGTM aside from minor nitpicks

PR #6156, opened 2015-12-08
Fix(md-select): md-select won't open if not in debug mode

### LINKED ISSUE #6033: [md-select][1.0.0-rc5]Cannot set property 'isOpen' of undefined
md-select throws error ``Cannot set property 'isOpen' of undefined`` around method
```
 function openSelect(e) {
        selectScope.isOpen = true;
        element.attr('aria-expanded', 'true');

        $mdSelect.show({

```

when set 

```
$compileProvider.debugInfoEnabled(false);
```
for production.

the element.scope warning appeared as well:

    element.scope() is not available when 'debug mode' == false. 
    @see https://docs.angularjs.org/guide/production!

myagoo:
Fix of md-select not opening since at least 1.0.0-rc5 when angular is not in debug mode


ThomasBurleson:
Scope to large. Please resubmit PR with distinct, small change/fix.

Issue #5837, opened 2015-11-19
ngMessage not showing until form is touched


bgoerdt:
In 1.0.0-rc4 the ngMessage for an input field is not showing up until the form is touched.

Here is a codepen demonstrating the issue: http://codepen.io/bgoerdt/pen/QjoPEP

bopm:
:+1:

andrew-cunliffe:
You can use the `md-is-error` on the input-container to force this, pretty sure its something new

e.g. something like... (or just flag as true if you always want the errors to show)

```
<md-input-container md-is-error="hasMessage(form.field)">
    ......
</md-input-container>
```

topherfangio:
This is likely a change caused by only showing after the input has the `md-input-is-invalid` class. We are making modifications to no longer do this by default, but to allow you to use `ng-show/hide` on the `ng-messages` element to better control this. Should be in master soon.

I will update this ticket when I push the PR.

bopm:
Any news on that? That blocks us from using 1.0 and reverting breaking changes to conform 0.11.0 will cost a lot.

SnakeME:
+1

bopm:
Only workaround that working for me right now is next css: ``.errors {opacity: 1 !important;}``. I am pretty angry that something so bad is needed in a project, so I'll remove it as soon as real solution will be here.

SnakeME:
md-input-container [ng-messages] {opacity: 1 !important;} worked for me.

bopm:
10 days with zero activity. Is it even going to be in 1.0? From my perspective that's huge problem as it breaks angular conventions and don't allow me to build standard user experience. Yet nothing is happening.

topherfangio:
@bopm In general, we attempt to follow Google's Material Design Spec as closely as possible. Based on the comments/images at http://www.google.com/design/spec/patterns/errors.html#errors-user-input-errors, errors are not supposed to be visible until the form is touched, so this extra CSS was added.

> Show **error text** only after user interaction with a field. 

However, we realize the need for a bit more customization (even the images in the spec show info-like messages before being touched), so we are working to make this available. We are planning to have this available in 1.0.

bopm:
[...]

PR #6164, opened 2015-12-08
fix(input): ngMessages jump with ngShow, ngIf, etc

### LINKED ISSUE #5298: ngMessages UI jump when used in combo with ng-show
Sometimes it's nice to only show an error message **after** a field is touched or submitted. So to do that we have used ng-show on the outer ng-messages element like below:

```
<div ng-messages="ctrl.form.title.$error" ng-show="ctrl.form.title.$touched || ctrl.form.$submitted" role="alert">
```

This works fine, but since some of the recent changes related to "multiple" we started seeing a small "UI jump" when leaving the field and the error message was to be shown (or not due to no errors).

To reproduce:

Add the ng-show condition on a field with ng-messages, focus the field, leave the field. The design will now jump just one pixel or so (only the first time).

Not exactly sure what would be the simplest solution, but I thought I would at least report this so someone else might be able to figure out a simple solution.

topherfangio:
When the ngMessages directive was used with the ngShow,
ngIf and other directives inside of an md-input-container,
the messages would flicker and jump rather than properly
animating.

Additionally, if the messages spanned multiple lines,
the animations were incorrect due to hard-coded values.

Fix many styles and move animations into JS so we can
properly calculate message heights.

Also, add more flexibility with the `md-auto-hide` CSS
class and associated attribute and update docs/demo
accordingly.

BREAKING CHANGE

By default, the `<md-input-container>` now automatically
hides any error messages until the component is in an
error state (determined by `md-is-error`). If you wish
to achieve the original behavior where messages are
always visible, add the `md-auto-hide="false"` attribute
to your `ng-messages` or use one of the following
visibility directives:

 - `ng-if`
 - `ng-show`/`ng-hide`
 - `ng-switch-when`/`ng-switch-default`

Fixes #5298. References #5837.

### DIFF
--- a/src/components/input/demoErrorsAdvanced/index.html
+++ b/src/components/input/demoErrorsAdvanced/index.html
@@ -1,0 +1,106 @@
+<div ng-controller="AppCtrl" layout="column" ng-cloak>
+
+  <md-content layout-padding>
+    <form name="userForm">
+      <div layout="row" layout-xs="column" layout-sm="column" layout-align="space-between center">
+        <div flex-gt-sm="80">
+          <p>
+            The <code>md-input-container</code> gives you the flexibility to display your messages
+            using many standard angular directives.
+          </p>
+
+          <p>
+            For instance, toggle the switch
+
+            <span hide-xs hide-sm>to the right</span>
+            <span hide-gt-sm>below</span>
+
+            to see the messages switch between some custom hints, and the actual error messages.
+            Note that some of the <code>ng-messages</code> containers use <code>ngIf</code> while
+            others use <code>ng-show</code> or <code>ng-hide</code>.
+          </p>
+        </div>
+
+        <md-input-container>
+          <md-switch ng-model="showHints">Showing {{showHints ? "Hints" : "Errors"}}</md-switch>
+        </md-input-container>
+      </div>
+
+      <div layout-gt-sm="row">
+
+        <md-input-container class="md-block" flex-gt-sm>
+          <label>Name</label>
+          <input md-maxlength="30" required name="name" ng-model="user.name" />
+
+          <div class="hint" ng-if="showHints">Tell us what we should call you!</div>
+
+          <div ng-messages="userForm.name.$error" ng-if="!showHints">
+            <div ng-message="required">Name is required.</div>
+            <div ng-message="md-maxlength">The name has to be less than 30 characters long.</div>
+          </div>
+        </md-input-container>
+
+        <div flex="5" hide-xs hide-sm>
+          <!-- Spacer //-->
+        </div>
+
+        <md-input-container class="md-block" flex-gt-sm>
+          <label>Social Security Number</label>
+          <input name="social" ng-model="user.social" ng-pattern="/^[0-9]{3}-[0-9]{2}-[0-9]{4}$/" />
+
+          <div class="hint" ng-if="showHints">###-##-####</div>
+
+          <div ng-messages="userForm.social.$error" ng-if="!showHints">
+            <div ng-message="pattern">###-##-#### - Please enter a valid SSN.</div>
+          </div>
+        </md-input-container>
+
+      </div>
+
+      <div layout-gt-sm="row">
+
+        <md-input-container class="md-block" flex-gt-sm>
+          <label>Email</label>
+          <input name="email" ng-model="user.email"
+                 required minlength="10" maxlength="100" ng-pattern="/^.+@.+\..+$/" />
+
+          <div class="hint" ng-show="showHints">How can we reach you?</div>
+
+          <div ng-messages="userForm.email.$error" ng-hide="showHints">
+            <div ng-message-exp="['required', 'minlength', 'maxlength', 'pattern']">
+              Your email must be between 10 and 100 characters long and look like an e-mail address.
+            </div>
+          </div>
+        </md-input-container>
+
+        <div flex="5" hide-xs hide-sm>
+          <!-- Spacer //-->
+        </div>
+
+        <md-input-container class="md-block" flex-gt-sm>
+          <label>Phone Number</label>
+          <input name="phone" ng-model="user.phone" ng-pattern="/^([0-9]{3}) [0-9]{3}-[0-9]{4}$/" />
+
+          <div class="hint" ng-show="showHints">(###) ###-####</div>
+
+          <div ng-messages="userForm.phone.$error" ng-hide="showHints">
+            <div ng-message="pattern">(###) ###-#### - Please enter a valid phone number.</div>
+          </div>
+        </md-input-container>
+
+        <style>
+          /*
+           * The Material demos system does not currently allow targeting the body element, so this
+           * must go here in the HTML.
+           */
+          body[dir=rtl] .hint {
+            right: 2px;
+            left: auto;
+          }
+        </style>
+      </div>
+
+    </form>
+  </md-content>
+
+</div>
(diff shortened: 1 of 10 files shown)


topherfangio:
@ThomasBurleson @jelbourn @EladBezalel @robertmesserle Sorry for the delay; here are my change for input messages with `ngShow`, `ngIf`, etc.

Can you all please take a look and review? I am headed to a short lunch but will be available if you find anything.

src/components/input/input.js
```diff
@@ -191,6 +242,10 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
     }
     containerCtrl.input = element;
 
+    // Add an error spacer div after our input to provide space for the char counter and any ng-messages
+    var errorsSpacer = angular.element('<div class="md-errors-spacer">');
```
EladBezalel: This spacer can't be done via pure css ?

src/components/input/input.scss
```diff
@@ -31,10 +33,19 @@ md-input-container {
...
   }
 
-  // When we have ng-messages, remove the input error height from our bottom padding, since the
-  // ng-messages wrapper has a min-height of 1 error (so we don't adjust height as often; see below)
-  &.md-input-has-messages {
-    padding-bottom: $input-container-padding;
+  // Setup a spacer that is always there as a placeholder for any messages so we don't change
+  // height with only 1 message
+  .md-errors-spacer {
+    float: right;
```
EladBezalel: you should use `@include rtl(float, right, left)`

src/components/input/input.scss
```diff
@@ -181,34 +192,29 @@ md-input-container {
...
-    right: $input-container-padding;
-    bottom: 7px;
-  }
-
-  &.md-input-invalid {
-    ng-messages, data-ng-messages, x-ng-messages,
-    [ng-messages], [data-ng-messages], [x-ng-messages] {
-      opacity: 1;
-    }
+    text-align: right;
```
EladBezalel: you should use @include rtl(text-align, right, left)

src/components/input/input.scss
```diff
@@ -181,34 +192,29 @@ md-input-container {
...
+    padding-right: $input-container-padding;
```
EladBezalel: you should use 
```
@include rtl(padding-right, $input-container-padding, 0);
@include rtl(padding-left, 0, $input-container-padding);
```

src/components/input/input.scss
```diff
@@ -181,34 +192,29 @@ md-input-container {
...
   [ng-messages], [data-ng-messages], [x-ng-messages] {
     position: relative;
     order: 4;
-    min-height: $input-error-height;
     overflow: hidden;
-    transition: $swift-ease-out;
-    transition-duration: $swift-ease-out-duration / 1.5;
-    opacity: 0;
+    min-height: $input-error-height;
     clear: left;
```
EladBezalel: Maybe add rtl support?

src/components/input/input.scss
```diff
@@ -216,51 +222,48 @@ md-input-container {
...
+
     // Add some top padding which is equal to half the difference between the expected height
     // and the actual height
-    $error-padding-top: ($input-error-height - $input-error-line-height) / 2;
     padding-top: $error-padding-top;
 
+    // For everything except the character counter (i.e. all messages)
     &:not(.md-char-counter) {
+      // Add some right padding so that the messages don't overlap the character counter
       padding-right: rem(5);
```
EladBezalel: rtl support

src/components/input/demoErrorsAdvanced/index.html
```diff
@@ -0,0 +1,95 @@
...
+            <span hide-gt-sm>below</span>
+
+            to see the messages switch between some custom hints, and the actual error messages.
+            Note that some of the <code>ng-messages</code> containers use <code>ngIf</code> while
+            others use <code>ng-show</code> or <code>ng-hide</code>.
+          </p>
+        </div>
+
+        <md-input-container>
+          <md-checkbox ng-model="showHints">Showing {{showHints ? "Hints" : "Errors"}}</md-checkbox>
```
jelbourn: I would change this to an `md-switch` since you're going between two different options instead of one thing being on/off.

src/components/input/demoErrorsAdvanced/index.html
```diff
@@ -0,0 +1,95 @@
+<div ng-controller="AppCtrl" layout="column" ng-cloak>
```
jelbourn: I pulled down this PR and ran it locally. It looks great!

src/components/input/input.js
```diff
@@ -480,6 +530,7 @@ function placeholderDirective($log) {
   }
 }
 
+var visibilityDirectives = ['ngIf', 'ngShow', 'ngHide', 'ngSwitchWhen', 'ngSwitchDefault'];
```
jelbourn: It might be worth mentioning in the docs that this understands this set of directives and that any custom hide/show directives won't behave the same.

Do we need to add things like `hide-gt-sm` ?

jelbourn:
Overall look good aside from existing comments.

src/components/input/demoErrorsAdvanced/index.html
EladBezalel: Agree!

src/components/input/demoErrorsAdvanced/index.html
topherfangio: Ah, indeed! I was thinking of using a radio button, but a switch definitely works better.

src/components/input/input.js
topherfangio: I added the following to the docs:

> We automatically disable this auto-hiding functionality if you provide any of the following visibility directives on the `ng-messages` container:
 - `ng-if`
 - `ng-show`/`ng-hide`
 - `ng-switch-when`/`ng-switch-default`
>
> You can also disable this functionality manually by adding the `md-auto-hide="false"` expression to the `ng-messages` container. This may be helpful if you always want to see the error messages or if you are building your own visibilty directive.

Is this sufficient, or should I add something else?

jelbourn: That looks good.

src/components/input/input.js
topherfangio: I tried a pure CSS solution, but ran into issues with the fact that I needed to position the `md-char-counter` at that exact spot. With pure CSS, the counter would move if the input had more than 1 message which made it bigger.

It's possible there is another way, but I was in a bit of a rush and I spent way too much time trying to find a solution before thinking of this and implementing it in about 15 minutes.

src/components/input/input.js
ThomasBurleson: please add the breakpoint attributes:

*  show, show-xs, show-gt-xs, etc.
*  hide, hide-xs, hide-gt-xs, etc.

jelbourn: I actually thought about this, and don't think it makes sense to use these responsive directives here.

ThomasBurleson: sgtm

topherfangio: @jelbourn I think they might actually be useful: consider having combined/shorter messages on mobile devices to take up less space?

That said, I'm honestly not sure how they would work at the moment. I'll need to do a tad bit of testing.

jelbourn: Don't worry about it for now. We can revisit later. It's unlikely to come up.

topherfangio:
@ThomasBurleson RTL support has been added along with a few other minor fixes.

PR #6170, opened 2015-12-08
fix(showHide): Don't set up $md-resize $broadcasting $watcher until r…

### LINKED ISSUE #5760: RC4: new feat(showHide) has a huge performance impact
On a page using lots of ng-show/ng-hide, RC4 has a monstrous performance impact. I could pinpoint the problem to this commit:
https://github.com/angular/material/commit/7355e5897b2ada38855710b7b902aaa456782d3f (aka #5579)

Does Angular-Material provide a way to disable that feature easily? Maybe something similar to what Angular Aria does:
```
$ariaProvider.config({tabindex: false});
```
Could be something like
```
$materialProvider.config({showHide: false});
```

kseamon:
…ecieving $md-resize-enable

Fixes #5760

### DIFF
--- a/src/components/virtualRepeat/virtual-repeater.js
+++ b/src/components/virtualRepeat/virtual-repeater.js
@@ -143,6 +143,7 @@ function VirtualRepeatContainerController(
       jWindow.off('resize', debouncedUpdateSize);
     });
 
+    $scope.$emit('$md-resize-enable');
     $scope.$on('$md-resize', boundUpdateSize);
   });
 }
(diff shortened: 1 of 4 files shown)


kseamon:
@jelbourn @ThomasBurleson

src/components/virtualRepeat/virtual-repeater.js
```diff
@@ -143,6 +143,7 @@ function VirtualRepeatContainerController(
       jWindow.off('resize', debouncedUpdateSize);
     });
 
+    $scope.$emit('$md-resize-enable');
```
jelbourn: Is there a way to make this optional if you want your virtual-repeat to be totally static?

kseamon: Why would you want that? In effect that means it would just not work sometimes.

PR #6172, opened 2015-12-08
fix(mdList): use correct size for md-font-icon

remcohaszing:
The width and height are set for the child md-icon element. When
md-font-icon is used, this is affected by font-size, not width or
height.


ThomasBurleson:
Will not merge. No reference issue.

Issue #1378, opened 2015-02-03
Swipe should begin off screen


gustavohenke:
I want to suggest that the swipe events begin off screen, or, in other words, at point 0 left/right (with some tolerance), for example:

![](https://s3.amazonaws.com/pushbullet-uploads/ujv8hIgQpvE-T2yyApoWl7YmJTfrHm2p2KbL25mtoZEP/Screenshot_2015-02-03-13-52-18.png)
_This one is from Inbox._

I tested this behavior in a few Android apps:

* Gmail
* Drive
* Inbox
* Hangouts
* Photos
* YouTube
* Google Now

All of them will either dismiss or change active tab if you try to swipe anywhere else.

gustavohenke:
It would be nice to add some tolerance also, ~10px for example.

ThomasBurleson:
What do you mean "begin at left:0 or right:100%" ?

gustavohenke:
I'm unable to record a video from my Android phone, so I hope this screenshot is able to sane your question:

![](https://s3.amazonaws.com/pushbullet-uploads/ujv8hIgQpvE-T2yyApoWl7YmJTfrHm2p2KbL25mtoZEP/Screenshot_2015-02-03-13-52-18.png)

Also added it to the initial comment on this issue.

robertmesserle:
Just to note: the swipe gesture absolutely should not require starting off screen.  This would have to be a new gesture.

gustavohenke:
Reading the specs is easier than trying to describe the expected behavior by myself:

> # Edge swipe
>
> A swipe that originates outside of the screen and reveals off-screen content. Invokes out-of-context content, that is, content separate and distinct from the current view.
>
> If no edge swipe action is defined, an Edge swipe gesture can default to a Paging swipe.
>
> The Edge swipe gesture is committed based on crossing a threshold.

_from "Edge swipe" in [Gestures](http://www.google.com/design/spec/patterns/gestures.html#gestures-drag-swipe-or-fling-details) specs_

DmitryEfimenko:
+1

ThomasBurleson:
This issue is closed as part of our deprecation effort.
    For details, see our post [Spring Cleaning 2016](https://groups.google.com/forum/#!topic/ngMaterial/3Ip9nd68bfk).

PR #6174, opened 2015-12-08
feat(sidenav): add gestures to sidenav to allow dragging the sidenav

### LINKED ISSUE #35: feat(sidenav): add swipe functionality
Add the same swipe functionality behavior as ion-side-menus to open sidenav not just with click but with swipe left (or right)

Awesome work!

### LINKED ISSUE #1591: Draggable sidenav
Is it possible to make sidenav draggable? Not only google apps like gmail and play music demonstrate the approach, but also MD bottom sheets.

bottomsheet

It feels more responsive that way and would be an appreciated enhancement.

DevVersion:
I'm currently working on the feature, which allows to drag the sidenav.

Fixes #1591 Fixes #35

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -197,6 +197,7 @@ function SidenavFocusDirective() {
  * @param {expression=} md-is-open A model bound to whether the sidenav is opened.
  * @param {boolean=} md-disable-backdrop When present in the markup, the sidenav will not show a backdrop.
  * @param {string=} md-component-id componentId to use with $mdSidenav service.
+ * @param {boolean=} md-disable-drag Disables the abbility to drag the sidenav
  * @param {expression=} md-is-locked-open When this expression evalutes to true,
  * the sidenav 'locks open': it falls into the content's flow instead
  * of appearing over it. This overrides the `md-is-open` attribute.
@@ -209,7 +210,7 @@ function SidenavFocusDirective() {
  *   - `<md-sidenav md-is-locked-open="$mdMedia('min-width: 1000px')"></md-sidenav>`
  *   - `<md-sidenav md-is-locked-open="$mdMedia('sm')"></md-sidenav>` (locks open on small screens)
  */
-function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, $compile, $parse, $log, $q, $document) {
+function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, $mdGesture, $parse, $log, $q, $document, $timeout) {
   return {
     restrict: 'E',
     scope: {
@@ -265,6 +266,8 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
     scope.$watch(isLocked, updateIsLocked);
     scope.$watch('isOpen', updateIsOpen);
 
+    // Enable dragging
+    if (!angular.isDefined(attr.mdDisableDrag) || !attr.mdDisableDrag) enableDragging();
 
     // Publish special accessor for the Controller instance
     sidenavCtrl.$toggleOpen = toggleOpen;
@@ -332,6 +335,101 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
         parent.css('overflow', lastParentOverFlow);
         lastParentOverFlow = undefined;
 
+      }
+    }
+
+    function enableDragging() {
+      $mdGesture.register(element, 'drag', { horizontal: true });
+
+      element
+        .on('$md.dragstart', onDragStart)
+        .on('$md.drag', onDrag)
+        .on('$md.dragend', onDragEnd);
+
+      var style = getComputedStyle(element[0]);
+      var sidenavWidth = parseInt(style.width);
+      var isRightSidenav = element.hasClass('md-sidenav-right');
+      var accelerationBound = 6;
+
+      var dragCancelled = false;
+      var dragPercentage;
+      var lastOpenState;
+      var lastDistance = 0;
+      var isQuickDrag = false;
+
+      function onDragStart() {
+        if (element.hasClass('md-locked-open')) {
+          dragCancelled = true;
+        } else {
+          lastOpenState = scope.isOpen;
+          element.css($mdConstant.CSS.TRANSITION_DURATION, '0ms');
+        }
+      }
+
+      function onDrag(ev) {
+        if (dragCancelled) return;
+
+        if (!isQuickDrag) {
+          var distance = lastDistance - ev.pointer.distanceX;
+          // When the current partial drag distance is bigger than the acceleration bound, then we can
+          // identify it as a quick drag.
+          isQuickDrag = isRightSidenav ? distance <= -accelerationBound : distance >= accelerationBound;
+        } else if (isRightSidenav && lastDistance > ev.pointer.distanceX ||
+                   !isRightSidenav && lastDistance < ev.pointer.distanceX) {
+          // When the users drags the sidenav backward, then we can reset the quick drag state.
+          isQuickDrag = false;
+        }
+
+        dragPercentage = Math.round((ev.pointer.distanceX / sidenavWidth) * 100);
+        if (!isRightSidenav) dragPercentage = 0 - dragPercentage;
+
+        if (dragPercentage > 100) dragPercentage = 100;
+        else if (dragPercentage < 0) dragPercentage = 0;
+
+        element.css($mdConstant.CSS.TRANSFORM, 'translate3d(-' + (isRightSidenav ? 100 - dragPercentage : dragPercentage) + '%,0,0)');
+        lastDistance = ev.pointer.distanceX;
+      }
+
+      function onDragEnd() {
+        if (dragCancelled) {
+          dragCancelled = false;
+          return;
+        }
+
+        var remainingPercentage = 100 - dragPercentage;
+        var animationTime = 4 * remainingPercentage;
+        var shouldClose = dragPercentage > 50 || isQuickDrag;
+
+        // This validates the correct translate value. The invert is here required, because a right
+        // aligned sidenav will transition in the other direction.
+        var endTranslate = shouldClose ? isRightSidenav ? 0 : -100 : isRightSidenav ? -100 : 0;
+
+        element.css($mdConstant.CSS.TRANSITION_DURATION, animationTime + "ms");
+        element.css($mdConstant.CSS.TRANSFORM, 'translate3d(' + endTranslate +  '%,0,0)');
+
+        // Reset drag
+        lastDistance = 0;
+        isQuickDrag = false;
+
+        // Invert shouldClose here, because we need to know if the sidenav should open.
+        $timeout(onAnimationDone, animationTime, true, !shouldClose);
+      }
+
+      function onAnimationDone(isOpen) {
+        scope.isOpen = isOpen;
+        element.css($mdConstant.CSS.TRANSFORM, '');
+        element.css($mdConstant.CSS.TRANSITION_DURATION, '');
+
+        if (isOpen) {
+          if (!lastOpenState && backdrop) {
+            $animate.enter(backdrop, element.parent());
+          }
+
+          element.removeClass('_md-closed');
+        } else {
+          if (backdrop) $animate.leave(backdrop);
+          element.addClass('_md-closed');
+        }
       }
     }
 
--- a/src/components/sidenav/demoSidenavDragging/index.html
+++ b/src/components/sidenav/demoSidenavDragging/index.html
@@ -1,0 +1,73 @@
+
+<div ng-controller="AppCtrl" layout="column" style="height:500px;" ng-cloak>
+
+  <section layout="row" flex>
+
+    <md-sidenav class="md-sidenav-left md-whiteframe-z2" md-component-id="left">
+
+      <md-toolbar class="md-theme-light">
+        <h1 class="md-toolbar-tools">Drag Enabled</h1>
+      </md-toolbar>
+      <md-content ng-controller="LeftCtrl" layout-padding>
+        <p>
+          This sidenav is able to be dragged. Through the drag gesture, it's possible to close the sidenav.
+        </p>
+        <md-button ng-click="close()" class="md-primary">
+          Close Sidenav Left
+        </md-button>
+      </md-content>
+
+    </md-sidenav>
+
+    <md-content flex layout-padding>
+
+      <div layout="column" layout-fill layout-align="top center">
+        <p>
+          You can simply close a sidenav by using a gesture. Simply drag the sidenav to close it.
+        </p>
+        <p>
+          The right sidenav allows dragging.
+        </p>
+        <p>
+          The left sidenav is disabled for dragging.
+        </p>
+
+        <div>
+          <md-button ng-click="toggleLeft()"
+            class="md-primary">
+            Toggle left
+          </md-button>
+        </div>
+
+        <div>
+          <md-button ng-click="toggleRight()"
+            ng-hide="isOpenRight()"
+            class="md-primary">
+            Toggle right
+          </md-button>
+        </div>
+      </div>
+
+      <div flex></div>
+
+    </md-content>
+
+    <md-sidenav class="md-sidenav-right md-whiteframe-z2" md-component-id="right" md-disable-drag="true">
+
+      <md-toolbar class="md-theme-indigo">
+        <h1 class="md-toolbar-tools">Drag Disabled</h1>
+      </md-toolbar>
+      <md-content layout-padding ng-controller="RightCtrl">
+        <md-button ng-click="close()" class="md-primary">
+          Close Sidenav Right
+        </md-button>
+        <p>
+          This sidenav disables the dragging gesture. It can be closed through the buttons.
+        </p>
+      </md-content>
+
+    </md-sidenav>
+
+  </section>
+
+</div>
--- a/src/components/sidenav/demoSidenavDragging/script.js
+++ b/src/components/sidenav/demoSidenavDragging/script.js
@@ -1,0 +1,27 @@
+angular
+  .module('sidenavDemo2', ['ngMaterial'])
+  .controller('AppCtrl', function ($scope, $timeout, $mdSidenav) {
+    $scope.toggleLeft = buildToggler('left');
+    $scope.toggleRight = buildToggler('right');
+
+    $scope.isOpenRight = function(){
+      return $mdSidenav('right').isOpen();
+    };
+
+    function buildToggler(navID) {
+      return function() {
+        $mdSidenav(navID).toggle()
+      }
+    }
+  })
+  .controller('LeftCtrl', function ($scope, $timeout, $mdSidenav) {
+    $scope.close = function () {
+      $mdSidenav('left').close();
+
+    };
+  })
+  .controller('RightCtrl', function ($scope, $timeout, $mdSidenav) {
+    $scope.close = function () {
+      $mdSidenav('right').close();
+    };
+  });
--- a/src/components/sidenav/sidenav.scss
+++ b/src/components/sidenav/sidenav.scss
@@ -14,6 +14,9 @@ md-sidenav {
   overflow: auto;
   -webkit-overflow-scrolling: touch;
 
+  // This allows the most browsers to setup appropriate optimizations before the animation runs.
+  will-change: transform;
+
   ul {
     list-style: none;
   }


DevVersion:
Current Progress on Mobile (ADB Recored)
- https://www.youtube.com/watch?v=FywfUAS92ns&feature=youtu.be

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,77 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
     }
 
+    function enableDragging() {
+      $mdGesture.register(element, 'drag', { horizontal: true });
+      element.on('$md.dragstart', onDragStart)
+        .on('$md.drag', onDrag)
+        .on('$md.dragend', onDragEnd);
+
+      var style = getComputedStyle(element[0]);
+      var sidenavWidth = parseInt(style.width);
```
EladBezalel: `element[0].offsetWidth` won't work better?

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,77 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+      }
+
+      function onDragEnd() {
+        if (dragCancelled) {
+          dragCancelled = false;
+          return;
+        }
+
+        var remainingPercentage = dragPercentage > 50 ? 100 - dragPercentage : dragPercentage;
+        var animationTime = 4 * remainingPercentage;
```
EladBezalel: 4?

DevVersion: That's because of my percentage calculation
- https://github.com/angular/material/blob/eb1695a0d54f6794a08f04169ad8a390b92a1849/src/core/style/variables.scss#L95

`400ms / 100% = 4ms`

So every percentage needs to run for 4ms to reach a smooth transition as always.

src/components/sidenav/sidenav.js
DevVersion: I will test it today, after school :)

DevVersion: Should be better, I will test it

DevVersion: That won't work, because the sidenav is hidden, that's why `offsetWidth` doesn't work

DevVersion:
@EladBezalel Would be great if you test the provided demo :smile:

JumpLink:
I've create a module that replaces the original mdSidenav with the version of this pull request for testing:
https://gist.github.com/JumpLink/cd2b4188a54ec3f4cc7a

DevVersion:
Thanks for your work @JumpLink, but I think we should test my PR with fetching the current patch (https://patch-diff.githubusercontent.com/raw/angular/material/pull/6174.patch).

EladBezalel:
@DevVersion I tested it locally and it feels a bit weird, i assume you based on the bottom sheet code.
Bottom sheet feels more user friendly (take in account the swipe acceleration etc..)

DevVersion:
What do you mean?
I recorded it with the Chrome  Dev Tools and there was a constant fps rate for the animation. It should drag simmilar to the bottom sheet. Please provide me more information

EladBezalel:
I mean, bottom sheet have this extra space to be dragged to, also as i said, swipe the bottom sheet down real quick and it will be dismissed..

DevVersion:
But that makes no sense to add extra space to a sidenav (as we can see for example in the Android API 23, there was never a extra space for dragging. This will create more issues with the width of the sidenav. And the bottom sheet will close itself after dragging it a few pixels down. That shouldn't be implemented in a sidenav too.

EladBezalel:
OK, about the width i can agree with you, but as i mention earlier bottom sheet calculate the drag acceleration and dismiss if its was a fast swipe and it should be also on the sidenav (take a look at the google store sidenav)

DevVersion:
Okay thanks @EladBezalel you convinced me, I will try to improve that :+1:

DevVersion:
@EladBezalel can you test the swipe acceleration? :)

EladBezalel:
@DevVersion sure, later today I'll take a look!

DevVersion:
Great thanks for your time!

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,93 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+      var dragCancelled = false;
+      var dragPercentage;
+      var lastOpenState;
+      var lastDistance = 0;
+      var isQuickDrag = false;
+
+      function onDragStart() {
+        if (element.hasClass('md-locked-open')) {
+          dragCancelled = true;
+        } else {
```
EladBezalel: please break line

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,93 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+          element.css($mdConstant.CSS.TRANSITION_DURATION, '0ms');
+        }
+      }
+
+      function onDrag(ev) {
+        if (dragCancelled) return;
+
+        if (!isQuickDrag) {
+          if (isRightSidenav) isQuickDrag = lastDistance - ev.pointer.distanceX <= -accelerationBound;
+          else isQuickDrag = lastDistance - ev.pointer.distanceX >= accelerationBound;
```
EladBezalel: wrap with brackets

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,93 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+        } else {
+          if (isRightSidenav && lastDistance > ev.pointer.distanceX) isQuickDrag = false;
+          else if (!isRightSidenav && lastDistance < ev.pointer.distanceX) isQuickDrag = false;
+        }
```
EladBezalel: i'd prefer
```js 
if (...) {
  if (...) {
  }
  else {
    ...
  }
}
else {
  ...
}
```

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,93 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+
+      function onDrag(ev) {
+        if (dragCancelled) return;
+
+        if (!isQuickDrag) {
+          if (isRightSidenav) isQuickDrag = lastDistance - ev.pointer.distanceX <= -accelerationBound;
+          else isQuickDrag = lastDistance - ev.pointer.distanceX >= accelerationBound;
+        } else {
+          if (isRightSidenav && lastDistance > ev.pointer.distanceX) isQuickDrag = false;
+          else if (!isRightSidenav && lastDistance < ev.pointer.distanceX) isQuickDrag = false;
```
EladBezalel: Same here, use ternary operator

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,93 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+        var isOpen = dragPercentage > 50 || isQuickDrag;
+        if (isRightSidenav) isOpen = !isOpen;
+
+        element.css($mdConstant.CSS.TRANSITION_DURATION, animationTime + "ms");
+
+        if (isOpen) {
+          element.css($mdConstant.CSS.TRANSFORM, 'translate3d(-100%,0,0)');
+        } else {
+          element.css($mdConstant.CSS.TRANSFORM, 'translate3d(-0%,0,0)');
+        }
```
EladBezalel: ternary operator

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,93 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+
+        // Reset drag
+        lastDistance = 0;
+        isQuickDrag = false;
+        $timeout(onAnimationDone, animationTime, true, (isRightSidenav ? isOpen : !isOpen));
```
EladBezalel: I think you should use `$mdUtil.nextTick`

DevVersion: Why should I use nextTick, if I want to wait a specific amount of time `animationTime`

src/components/sidenav/sidenav.js
DevVersion: Yes Sir, seems to be a better choice

src/components/sidenav/sidenav.js
DevVersion: Yes, that will look more clear.

src/components/sidenav/sidenav.js
DevVersion: I will use brackets ;)

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,93 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+      function onAnimationDone(isOpen) {
+        scope.isOpen = isOpen;
+        element.css($mdConstant.CSS.TRANSFORM, '');
+        element.css($mdConstant.CSS.TRANSITION_DURATION, '');
+        if (isOpen && lastOpenState == false) {
+          $animate.enter(backdrop, element.parent());
+        } else if (!isOpen) {
+          $animate.leave(backdrop);
+        }
+        element[isOpen ? 'removeClass' : 'addClass']('md-closed')
```
EladBezalel: You should do some thing like:

```js
if (isOpen) {
  if (!lastOpenState) {
    $animate.enter(backdrop, element.parent());
  }
  element.removeClass('md-closed');
} 
else {
  $animate.leave(backdrop);
  element.addClass('md-closed');
}
```

src/components/sidenav/sidenav.js
EladBezalel: Oh sorry didn't see that.. 
but what does that means:
``` js
isRightSidenav ? isOpen : !isOpen
```

isn't that just `isOpen` ?

EladBezalel:
I would also change the `right sidenav` to be non-draggable cause the first button is `left sidenav` and it's in the dragging demo so it's the first button to press..

src/components/sidenav/sidenav.js
DevVersion: I need to mirror that because the right sidenav is opening reversely to the light sidenav

DevVersion:
@EladBezalel - Should be better now :), you already tested it local? What do you think?

DevVersion:
@EladBezalel - Would be great if you review :smile:

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,100 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+          if (isRightSidenav) {
+            isQuickDrag = distance <= -accelerationBound;
+          } else {
+            isQuickDrag = distance= accelerationBound;
+          }
+        } else {
+          if (isRightSidenav && lastDistance > ev.pointer.distanceX) {
+            isQuickDrag = false;
+          } else if (!isRightSidenav && lastDistance < ev.pointer.distanceX) {
+            isQuickDrag = false;
```
EladBezalel: can be simplified to:
```js
if (isRightSidenav && lastDistance > ev.pointer.distanceX || 
    !isRightSidenav && lastDistance < ev.pointer.distanceX) {
  isQuickDrag = false
}
```

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,100 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+
+      function onDrag(ev) {
+        if (dragCancelled) return;
+
+        if (!isQuickDrag) {
+          var distance = lastDistance - ev.pointer.distanceX;
+          if (isRightSidenav) {
+            isQuickDrag = distance <= -accelerationBound;
+          } else {
+            isQuickDrag = distance= accelerationBound;
```
EladBezalel: Is `distance=` is intentional?
if so please enter space before the equal sign

DevVersion:
@EladBezalel - Not that `spacing` wasn't by intention, thanks for keeping an eye on that. It's fixed now

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,99 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+
+      function onDrag(ev) {
+        if (dragCancelled) return;
+
+        if (!isQuickDrag) {
+          var distance = lastDistance - ev.pointer.distanceX;
+          if (isRightSidenav) {
+            isQuickDrag = distance <= -accelerationBound;
+          } else {
+            isQuickDrag = distance = accelerationBound;
```
EladBezalel: @DevVersion what i meant about `intention` is that according 2 lines above you check for smaller than, so i thought you might missed a larger/smaller sign

EladBezalel:
@DevVersion also please break line before every `each`.. 
> Driving me crazy :confounded:

DevVersion:
@EladBezalel Okay acceleration bound is fixed. Sorry for all that style errors :smile:. Where should I break lines?

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,106 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+        if (!isQuickDrag) {
+          var distance = lastDistance - ev.pointer.distanceX;
+          if (isRightSidenav) {
+            isQuickDrag = distance <= -accelerationBound;
+          } else {
+            isQuickDrag = distance >= accelerationBound;
+          }
+        } else {
+          if (isRightSidenav && lastDistance > ev.pointer.distanceX ||
+              !isRightSidenav && lastDistance < ev.pointer.distanceX) {
```
EladBezalel: BTW what happens if `lastDistance == ev.pointer.distanceX` ?

EladBezalel:
@DevVersion before every each
```js
if (...) {
  ...
}
else (...) {
 ...
}
```

DevVersion:
Okay, it's your choice, I thought we should not do that. [Conventions](https://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting)

src/components/sidenav/sidenav.js
DevVersion: Nothing, that's by intention, because we should not interrupt the `quickDrag` if we don't drag back.

EladBezalel:
@DevVersion well, there's that.. go ahead and choose yourself.. 
for me it's not making sense. no where else a text is written after the bracket..
KUDOS for you to read and follow those guidelines, very thorough!

DevVersion:
@EladBezalel - I think I just use the `Conventions` in this PR, because all the other `if-else` statements in this `javascript` file are using the conventions too.

src/components/sidenav/sidenav.js
EladBezalel: So isn't that just `isRightSidenav` ? :smile:

DevVersion: Sorry if I don't get your question (really stressed due this days :smile:) - This is only happening on the right Sidenav, because as you see above I invert the `isOpen` value, because the sidenav is aligned at the other side of the page, so I mirror that. And then after I want to complete the Animation I need to get the end value, and I can do this by reverting the value back.

EladBezalel: @DevVersion Don't let this damage your other things going on.

I just said -
```js
$timeout(onAnimationDone, animationTime, true, (isRightSidenav ? isOpen : !isOpen));
```
is equals to
``` js
$timeout(onAnimationDone, animationTime, true, isRightSidenav);
```

DevVersion: No that won't work, because `isOpen` can also be `false` or `true` and needs to be inverted, so `isRightSidenav` is not dynamic. That's why I need to dynamically invert the value.

EladBezalel: Oh right :+1:

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,101 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+
+      function onDrag(ev) {
+        if (dragCancelled) return;
+
+        if (!isQuickDrag) {
+          var distance = lastDistance - ev.pointer.distanceX;
+          if (isRightSidenav) {
+            isQuickDrag = distance <= -accelerationBound;
+          } else {
+            isQuickDrag = distance >= accelerationBound;
```
``` js
isQuickDrag = isRightSidenav ? distance <= -accelerationBound : distance >= accelerationBound;
```

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,101 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+          }
+        } else {
+          if (isRightSidenav && lastDistance > ev.pointer.distanceX ||
```
EladBezalel: `else if`

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,101 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+      function onAnimationDone(isOpen) {
+        scope.isOpen = isOpen;
+        element.css($mdConstant.CSS.TRANSFORM, '');
+        element.css($mdConstant.CSS.TRANSITION_DURATION, '');
+
+        if (isOpen) {
+          if (!lastOpenState) {
+            $animate.enter(backdrop, element.parent());
+          }
+          element.removeClass('md-closed');
```
EladBezalel: you should use `$animate.removeClass`

src/components/sidenav/sidenav.js
```diff
@@ -323,6 +326,101 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+        } else {
+          $animate.leave(backdrop);
+          element.addClass('md-closed');
```
EladBezalel: you should use `$animate.addClass`

DevVersion: All the same from the other PR ;)

src/components/sidenav/sidenav.js
DevVersion: That isn't working here, because we need to run the animation in the same digest.

DevVersion:
@EladBezalel - As I already said `$animate` isn't working here. And now thanks for all your suggestions. Next PR's I will check myself for code optimization like `ternary` operator.

EladBezalel:
@DevVersion i'm happy i can help, i'm fine with leaving those without `$animate` as it's drag event and i think the possibility this animation will harm protractor users is very small

DevVersion:
@EladBezalel Does it still needs any work? Let me know if there is anything you spotted :smile:

lowshadow:
When will this be merged? Its an awesome addition to sidenav!

florent1933:
+1

code-tree:
@DevVersion Works great, thanks for this. One issue I have is that (for whatever reason) `sidenavWidth` was set to a non-integer when the app started (i.e. style.width was not an int) and broke the ability to drag. I instead moved `style` and `sidenavWidth` assignment to inside the `onDrag` function as a temporary fix (which did fix it).

My sidenav is auto-set to locking depending on the width, so it may lock in landscape but not lock in portrait.

Might be something to consider before this is merged?

code-tree:
> Should there also be an option to pull the menu out? (from closed state). This would be really complicated to behave with other sidenavs.

It sounds like there should from #1378, though it may be better to implement that separately after this is merged?

code-tree:
@DevVersion I just realised my app is suffering from same issue mentioned for md-list (#7311). After including this pull request, my sidenav no longer *touch* scrolls in iOS (works fine in Chrome though).

Any ideas?

code-tree:
[Removing preventDefault](https://github.com/code-tree/material/commit/78c0a9391daa4b6c74ff0005e80d839ade4d8684) in gestures has solved this for me, however, I'm not informed enough to know what repercussions that will have on other browsers and components.

DevVersion:
@code-tree Thanks for all the information you provided. This PR is an old PR of mine and I will rework that again, to be more clean and more.

code-tree:
Ok, sounds good. Sorry, another issue for you for when you rework it :)

Sidenav is popping back out and in for a split second after dragging closed. Seems to be an issue with transitioning from the `transform` set directly on the element back to the `transform` set by `md-closed`. Couldn't quite figure out why, but a tmp fix is to [temporarily disable the transition](https://github.com/code-tree/material/commit/ce88c26a0db06d52a97e11d0c099ed679147c8db) for the split second that it happens in.

Was occurring on both Chrome and iOS.

code-tree:
Figured it out. Was doing a double animation due to updating scope.isOpen at the same time (which triggers its own animation). Can [pull from here](https://github.com/code-tree/material/commit/f6ed0a9ac42ab8f63f93a0c71a3d8fce98e88e41) if you wish to.

ThomasBurleson:
@DevVersion - what is the status of this PR? It appears outdated ?

DevVersion:
@ThomasBurleson It is not outdated - I will rebase the next days. 

We can better say, it's currently blocked - because once this feature is landed, the siidenav will not allow scrolling for iOS anymore.

This is caused by #7311 and a fix is already pending #7857

DevVersion:
FYI: This PR will be deprecated and re-submitted, once #7857 has landend.

ThomasBurleson:

ThomasBurleson:

ThomasBurleson:

PR #6175, opened 2015-12-08
fix(chips): safety check before getting length

StefanFeederle:
If the model is not yet filled (ajax?) this functions fails with:
TypeError: Cannot read property 'length' of undefined

If we check the availability of "this.items" before accessing it we bypass this problem.

### DIFF
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -181,7 +181,7 @@ MdChipsCtrl.prototype.chipKeydown = function (event) {
  */
 MdChipsCtrl.prototype.getPlaceholder = function() {
   // Allow `secondary-placeholder` to be blank.
-  var useSecondary = (this.items.length &&
+  var useSecondary = (this.items && this.items.length &&
       (this.secondaryPlaceholder == '' || this.secondaryPlaceholder));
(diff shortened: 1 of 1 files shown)


topherfangio:
@StefanFeederle Can you please add a quick unit test which demonstrates the issue. I'd like to make sure we have a test so we don't introduce a regression later.

Thanks!

StefanFeederle:
Hey @topherfangio, here are two codepens to highlight this issue. I don't know how to write unit tests, sorry.

[error example](http://codepen.io/anon/pen/rxPdxW)
[fixed example](http://codepen.io/anon/pen/OMdvXG)

The only difference is that i added the code of this PR in the fixed one.

If databinding to a not existing object {{ctrl.form.name}} does not throw errors, md-chips should behave the same and safety-check the object. I know this is not the best coding style and could be circumvented in my code, but i think its also important that angular-material is inherently consistent. This error is really hard to track down for newbies like me.

StefanFeederle:
I signed it!

topherfangio:
@ThomasBurleson Would love a unit test to go along with this, but is a NPE check, so it's super-simple. Think we should go ahead and just merge it?

PR #6179, opened 2015-12-08
docs(input): Fix input Basic Usage demo in IE11.

topherfangio:
In IE11, the input's Basic Usage demo had odd spacing/overlap issues at lower screen widths.

Fix by using standard block layout for small screens and setting `class="md-block"` on most inputs.


topherfangio:
@ThomasBurleson @jelbourn @robertmesserle I have a solution for the input's Basic Usage demo IE11 issues, but it seems a bit counter-intuitive to write them this way.

Particularly, I had to add the `class="md-block"` to pretty much all of the containers to get it to work/look right. It still looks great when in row layout, but it seems silly to apply this to every single container.

Thoughts?

robertmesserle:
I agree.  We should have a wrapper that specifies a form layout.  Bootstrap
does something where you wrap your inputs in `inline-form` or whichever
form layout you want.

PR #6181, opened 2015-12-09
fix(list): Don't require avatar to be direct child

### LINKED ISSUE #5554: md-avatar-icon inside a clickable md-list-item
It seems that the css for the md-avatar-icon is a bit too specific.

```css
md-list-item > .md-avatar-icon {
    background-color: '{{foreground-3}}';
    color: '{{background-color}}';
}
```

When adding a click fn to the containing md-list-item, the new structure becomes:

md-list-item
-- button
----.md-list-item-inner
------.md-avatar-icon

Is there any reason to not have an md-avatar-icon inside a clickable list-item? Or is it just an omission? If the later I can submit a quick PR.

Something along these lines would fix my use case:
```css
md-list-item > .md-avatar-icon, .md-list-item-inner > .md-avatar-icon
```
**OR**
```css
md-list-item .md-avatar-icon
```

Cheers

programmist:
Fixes #5554

### DIFF
--- a/src/components/list/list-theme.scss
+++ b/src/components/list/list-theme.scss
@@ -12,7 +12,7 @@ md-list.md-THEME_NAME-theme {
     background-color: '{{background-100}}';
   }
 
-  md-list-item > .md-avatar-icon {
+  md-list-item .md-avatar-icon {
     background-color: '{{foreground-3}}';
     color: '{{background-color}}';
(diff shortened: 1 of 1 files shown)


programmist:
See example fix: http://codepen.io/programmist/pen/ZQYZay

PR #6182, opened 2015-12-09
fix(speedDial): Fix intially open bug.

### LINKED ISSUE #6111: md-fab-speed-dial showing actions when isOpen is false.
The current screenshot is from the latest material angular website.
I can see in the code that `this.isOpen=false` but the actions are visible and when we click on button, only after that it triggers the show and hide functionality.
However, in the bottom example it seems to work fine, don't know why ?

#### Can anyone elaborate on this how to hide on initial load the actions behind the big icon button.

screen shot 2015-12-07 at 2 17 45 pm

topherfangio:
Due to an animation issue, the FAB Speed Dial started in an open state when using the `md-fling` animation.

Fix by delaying the initial animation until after the component was fully rendered so that the animation's calculations were correct and adding some CSS to ensure the first animation fired instantly.

Fixes #6111.

### DIFF
--- a/src/components/fabSpeedDial/fabController.js
+++ b/src/components/fabSpeedDial/fabController.js
@@ -29,6 +29,8 @@
     setupDefaults();
     setupListeners();
     setupWatchers();
+
+    var initialAnimationAttempts = 0;
     fireInitialAnimations();
 
     function setupDefaults() {
@@ -40,6 +42,9 @@
 
       // Start the keyboard interaction at the first action
       resetActionIndex();
+
+      // Add an animations waiting class so we know not to run
+      $element.addClass('md-animations-waiting');
     }
 
     function setupListeners() {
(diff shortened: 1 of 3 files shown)


src/components/fabSpeedDial/fabController.js
```diff
@@ -135,9 +135,13 @@
 
     // Fire the animations once in a separate digest loop to initialize them
     function fireInitialAnimations() {
-      $mdUtil.nextTick(function() {
-        $animate.addClass($element, 'md-noop');
-      });
+      var delay = 100; // Delay until after the element has been appened to the DOM
```
jelbourn: I'm really not a fan of hard-coded delays like this. Is there any other way to approach this? I can consult with @matsko if it's an issue around ngAnimate.

topherfangio: @jelbourn Totally agree; let me try `$$rAF`, but the main issue is that I need the elements to be visible on-screen before the animations run because they do some calculations based on the spacing.

The **real** issue here is that the "default" non-animated state is for the element to be opened. We use the JS animations to close it. Ideally, we'd flip this so that the default state was closed and we opened it with the animations, but that would require rewriting it to not use flex positioning, which I'd like to do, but will take a while, so this is a temporary fix.

I'll let you know if `$$rAF` works or not.

jelbourn:
Closed via cfdd7cf12a00a0d47fc521468d61c9cfb3a0d9f3

PR #6207, opened 2015-12-09
fix(select): focus should behave as same as normal inputs

### LINKED ISSUE #6122: Demos, select validation fails on Clear

### LINKED ISSUE #6185: md-select label remains highlighted after lost focus
Hi

For all controls the label element goes grey after lost focus of the input. This behavior is different on the md-select where the label remains the highlighted color after the md-select has lost focus (it does not grey out like on other input controls)

Can be seen on the first demo at:
https://material.angularjs.org/latest/demo/select

When street and city are entered and the controls lose focus, the label is grayed out. The last item state (an md-select) however after selection the label remains blue.

### LINKED ISSUE #6274: Clearing md-select value (model) causes label to overlap empty select
Noticed the following layout issue. 

I use select option to filter some results and allow the user to clear filter. 

1. Selecting a md-select option
2. Clicking button to set model to "", null, or undefined
3. The md-select label positions over the select field even when it is blank.

http://codepen.io/anon/pen/YwyZXY

capture

Thanks

DevVersion:
- [x] Add test for `blur` and `focus` using `md-input-container`

Fixes  #6122 Fixes #6185 Closes #6132 Fixes #6274

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -187,12 +187,6 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
     return function postLink(scope, element, attr, ctrls) {
       var isDisabled, ariaLabelBase;
 
-      // Remove event ngModel's blur listener for touched and untouched
-      // we will do it ourself.
-      $mdUtil.nextTick(function() {
-        element.off('blur');
-      });
-
       var containerCtrl = ctrls[0];
       var mdSelectCtrl = ctrls[1];
       var ngModelCtrl = ctrls[2];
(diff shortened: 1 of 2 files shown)


src/components/select/select.js
```diff
@@ -466,6 +461,7 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
           loadingAsync: attr.mdOnOpen ? scope.$eval(attr.mdOnOpen) || true : false
         }).finally(function() {
           selectScope.isOpen = false;
+          element.focus();
```
jelbourn: I suspect this might interfere with the accessibility work that @rschmukler has recently done on this component. I'll ask him to take a look.

PR #6208, opened 2015-12-09
fix(toast): wrapping content if md-toast-content is not specified

EladBezalel:


topherfangio:
LGTM - Would be nice to have a test if possible.

EladBezalel:
I've tried to test but `$mdToast` can't be injected therefor we can't use show and specify a template that will go through the `transformTemplate`

EladBezalel:
seems to be fixed in ea60dd3b8e6ad35f5c06d8134b76c9848ce0b71b

PR #6226, opened 2015-12-10
fix(layouts): disable breaks associated with xs and xl

ThomasBurleson:
To reduce CSS file size for some deployments, create branch that has disabled the Layout breakpoints and generated mediaQuery CSS for the *-xs, -gt-xs, -gt-lg, -xl*​ breakpoints.


ThomasBurleson:
These changes are not needed for the public build of Angular Material v1.0

PR #6227, opened 2015-12-10
feat(chips): Support input tag inside md-chip-template by fixing focusing issue

OzzieOrca:
When using `<md-autocomplete>` and chips together, change onFocus to select the last `<input>` (the autocomplete input) instead of the first input which would be an input inside a `<md-chip-template>` tag.

Currently if you click anything other than exactly where the autocomplete input it, it selects whatever the first input is.

Example:
```
<md-chips ng-model="myChips" md-autocomplete-snap ng-init="myChips = []">
  <md-autocomplete
    md-items="item in ['item1', 'item2']"
    md-min-length="0">
    <md-item-template>
      <span >{{item}}</span>
    </md-item-template>
  </md-autocomplete>
  <md-chip-template>
      <div md-autocomplete-snap>
        {{$chip}}:
        <input ng-model="something">
    </div>
  </md-chip-template>
</md-chips>
```

In reality I'm putting another autocomplete inside the `<md-chip-template>` for filtering but was trying to make the example cleaner by just using an `<input>`.

### DIFF
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -436,7 +436,10 @@ MdChipsCtrl.prototype.configureNgModel = function(ngModelCtrl) {
 MdChipsCtrl.prototype.onFocus = function () {
   // Focus last input
   var inputs = this.$element[0].querySelectorAll('input');
-  inputs.length > 0 && inputs[inputs.length - 1].focus();
+  if(inputs.length > 0){
+    var lastInput = inputs[inputs.length - 1];
+    lastInput.focus();
+  }
   this.resetSelectedChip();
 };
 
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -434,8 +434,9 @@ MdChipsCtrl.prototype.configureNgModel = function(ngModelCtrl) {
 };
 
 MdChipsCtrl.prototype.onFocus = function () {
-  var input = this.$element[0].querySelector('input');
-  input && input.focus();
+  // Focus last input
+  var inputs = this.$element[0].querySelectorAll('input');
+  inputs.length > 0 && inputs[inputs.length - 1].focus();
   this.resetSelectedChip();
 };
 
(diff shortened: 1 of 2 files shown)


topherfangio:
@OzzieOrca This looks really interesting, but can you please add some tests which validates that this works? Not only do the tests help ensure we don't introduce regressions, but it's also used as a sort of API for developers so they understand the component's capabilities.

Thanks!

topherfangio:
@OzzieOrca Just wanted to ping you real fast and see if you have the time to add the requested unit tests 😄

src/components/chips/js/chipsController.js
```diff
@@ -434,8 +434,9 @@ MdChipsCtrl.prototype.configureNgModel = function(ngModelCtrl) {
 };
 
 MdChipsCtrl.prototype.onFocus = function () {
-  var input = this.$element[0].querySelector('input');
-  input && input.focus();
+  // Focus last input
+  var inputs = this.$element[0].querySelectorAll('input');
+  inputs.length > 0 && inputs[inputs.length - 1].focus();
```
ThomasBurleson: Please reformat:

```
if (inputs.length > 0) {
   var lastInput = inputs[inputs.length - 1];
   lastInput.focus();
}
```

ThomasBurleson: Also needs a unit test for this.

OzzieOrca:
I've added a test and reformatted the code a little bit. Let me know what you think.

topherfangio:
Now that I look at this a bit more, I don't think that the [docs](https://material.angularjs.org/latest/api/directive/mdChips) say anything about where the `<md-chips-template>` can be used, so I think some people may currently be using the following setup, which I don't think this will account for:

```html
<md-chips>
  <input />

  <md-chip-template><input /></md-chip-template>
</md-chips>
```

Since this would still find the wrong input, perhaps we could create an attribute directive which informs the chips which input to use for focusing, etc? Something like `<input md-chips-input />` maybe?

What are your thoughts?

OzzieOrca:
Good point. Could we select the input inside `md-chip-input-container`? It seems to be created to wrap both a normal input and an autocomplete one. I didn't realize it was there.

topherfangio:
@OzzieOrca Yeah, I think that will work. Can you give it a try?

ThomasBurleson:

ThomasBurleson:

PR #6228, opened 2015-12-10
fix(select): removes usage of `element.scope()`

### LINKED ISSUE #6033: [md-select][1.0.0-rc5]Cannot set property 'isOpen' of undefined

robertmesserle:
closes #6033

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -417,7 +417,7 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
         selectContainer = angular.element(
           element[0].querySelector('.md-select-menu-container')
         );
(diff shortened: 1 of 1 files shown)


robertmesserle:
@jelbourn @rschmukler @ThomasBurleson 

This fix should not be this easy, but the scope being fetched was the same scope that we already had a reference to.

PR #6231, opened 2015-12-10
fix(autocomplete): add message padding to progressbar with floating-label

### LINKED ISSUE #6218: md-autocomplete progressbar is misaligned when using float-label and query results
When using float label, the progress bar position is misaligned as you can see here:

http://codepen.io/anon/pen/mVJXry?editors=101

DevVersion:
**After fix**

Fixes #6218


DevVersion:
This fix is unnecessary due another fix, which includes that.

Issue #5672, opened 2015-11-11
mdInputContainer: Spacing broke in rc2


Flygenring:
This may be the problem mentioned in this commit: https://github.com/angular/material/commit/17797f001533288560854b04de81f53c552d2a80

As far I have seen it happened after changing the spacing on md-input-container from [rc1](https://github.com/angular/material/blob/v1.0.0-rc1/src/components/input/input.scss#L29) to [rc2](https://github.com/angular/material/blob/v1.0.0-rc2/src/components/input/input.scss#L26)

```
/* in rc1 */
  padding-bottom: $input-container-padding + $input-error-height;

/* in rc2 */
  margin: 18px 0;
```

This is all good and fine, but the way spacing changes for input containers with messages hasn't changed, so it still tries to cancel the bottom padding from the old method:
```
  padding-bottom: $input-container-padding;
```

This effectively means that an input container with an input and messages is 24px taller than one without messages.

julienpa:
This causes many problems to me too, as I recently updated from 0.11.4 to 1.0.0-rc4 and all my vertical forms are now broken with inconsistent spacing between input with and without messages.
Isn't a fix planned?

lievenjanssen:
Any idea when this will be fixed?

AdriVanHoudt:
This causes super spaced out forms. Any thoughts on when/id this will be fixed?

EladBezalel:
@robertmesserle moving this to your plate as you're on the PR that fixes it

PR #6258, opened 2015-12-11
fix(autocomplete): fix progressbar and messages alignment and bottom padding

### LINKED ISSUE #6218: md-autocomplete progressbar is misaligned when using float-label and query results

### LINKED ISSUE #6255: md-autocomplete with md-floating-label has too much space below it
See demo page: https://material.angularjs.org/latest/demo/autocomplete

There's the 24-ish px for the <ng-messages>, but then there's also 26px of padding that seems unnecessary and hurts the vertical rhythm of a page with md-autocomplete and md-input-container used interchangeably.

DevVersion:
- There was too much padding at the bottom (padding already done by the input container)
- Progressbar was wrong aligned due wrong bottom padding (as above said)

Fixes #6218 Closes #6231 Fixes #6255

### DIFF
--- a/src/components/autocomplete/autocomplete.scss
+++ b/src/components/autocomplete/autocomplete.scss
@@ -1,6 +1,4 @@
 $autocomplete-option-height: 48px;
-$input-container-padding: 2px !default;
-$input-error-height: 24px !default;
 
 @keyframes md-autocomplete-list-out {
   0% {
@@ -51,14 +49,7 @@ md-autocomplete {
     height: auto;
 
     md-input-container {
-      padding-bottom: $input-container-padding + $input-error-height;
-
-      // When we have ng-messages, remove the input error height from our bottom padding, since the
-      // ng-messages wrapper has a min-height of 1 error (so we don't adjust height as often; see
-      // input.scss file)
-      &.md-input-has-messages {
-        padding-bottom: $input-container-padding;
-      }
+      padding-bottom: 0px;
     }
     md-autocomplete-wrap {
       height: auto;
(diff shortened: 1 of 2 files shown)


AdriVanHoudt:
you need to fix the offset of the progress bar as well since it takes the padding into account see https://github.com/angular/material/blob/master/src/components/autocomplete/autocomplete.scss#L89

DevVersion:
Did you even run the new PR? The padding is already added to the Input Container

vlopez5:
Works perfectly for me @DevVersion !

DevVersion:
Thanks for checking it out :smile:

AdriVanHoudt:
I had a fix going for https://github.com/angular/material/issues/5672 so that messed it up sorry ^^

src/components/autocomplete/autocomplete.scss
```diff
@@ -51,14 +49,7 @@ md-autocomplete {
...
     md-input-container {
-      padding-bottom: $input-container-padding + $input-error-height;
-
-      // When we have ng-messages, remove the input error height from our bottom padding, since the
-      // ng-messages wrapper has a min-height of 1 error (so we don't adjust height as often; see
-      // input.scss file)
-      &.md-input-has-messages {
-        padding-bottom: $input-container-padding;
-      }
+      padding-bottom: 0px;
```
topherfangio: I think this should be `padding-bottom: $input-container-padding` so that it is bottom-aligned with the input. Based on the GIF you provided, it looks like the progress bar sits above the input's bottom border which is incorrect.

topherfangio: Nope; don't listen to me...apparently that bumps the progress bar below the autocomplete.

PR #6275, opened 2015-12-12
fix(select): fix blur event not triggering, focus styles in containers

rschmukler:
### DO NOT MERGE ME YET

This issue was the culprit for the select double label when invalid error that was discussed by @jelbourn and @topherfangio in slack.

I have *no* idea why this fix works. Specifically, `focus` fires but `blur` does not with the jqlite `on` methods. I'd love to hear from anyone who has ideas on why this might be (ping @jelbourn, @ThomasBurleson, @robertmesserle)

I changed the focus listener for consistency, but this feels weird and I'd much rather get to the bottom of the issue than this quick-fix.

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -275,15 +275,17 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
       };
 
       if (!isReadonly) {
-        element
-          .on('focus', function(ev) {
+        element[0]
+          .addEventListener('focus', function() {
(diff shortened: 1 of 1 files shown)


gkalpak:
It is obviously because of https://github.com/angular/material/blob/select-blur-fix/src/components/select/select.js#L192-L194.

rschmukler:
@gkalpak haha, yes obvious indeed. Sometimes a 2nd set of eyes is all it takes :P. Thanks

DevVersion:
I think that's duplicating #6207

PR #6280, opened 2015-12-13
fix(input): prevent duplicate showing duplicate messages

### LINKED ISSUE #6265: mdInputContainer: messages are displayed more than once
When an error in an `md-input-container` is triggered multiple times (e.g. when repeatedly changing the value using <kbd>backspace</kbd> and some other key) it is possible that the error message is displayed multiple times.

Issue tested on master. See the following screenshot:

ngmessage

DevVersion:
At the moment the both animation show / hide are running at same time.
That's why the client see's two messages at the same time.

Fixes #6265

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -585,13 +585,15 @@ function ngMessageDirective($mdUtil) {
   }
 }
 
+var isShowAnimating = false;
+
 function mdInputInvalidMessagesAnimation($q, $animateCss) {
   return {
     addClass: function(element, className, done) {
       var messages = getMessagesElement(element);
 
       if (className == "md-input-invalid" && messages.hasClass('md-auto-hide')) {
-        showInputMessages(element, $animateCss, $q).finally(done);
+        showInputMessages(element, $animateCss, $q, isShowAnimating).finally(done);
       }
     }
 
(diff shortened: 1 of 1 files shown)


DevVersion:
Please check that out @Frank3K

ThomasBurleson:

Frank3K:
Unfortunately the fix does not seem to work. 

Upon reading the code in the PR I see that the real work is done inside the function `hideMessage`, the other changes are only about setting and passing status variables. Upon inspection the function `hideMessage` is not triggered at all in my case (I have set a breakpoint on the first line).

DevVersion:
The fix works not perfect (I will create another commit). But it works better than the old version.. now the messages are not appearing multiple times. `hideMessage` will be triggered in the ngMessageAnimation on `leave`.

Frank3K:
You are right @DevVersion; for the demo the issue is indeed fixed.

For my own testcase the issue is not fixed, let me check what's the difference between those two...

topherfangio:
@Frank3K - This may be a separate issue. We noticed recently that the animations are firing just fine within our demo application, but not when we open it as a separate Codepen. We are currently investigating what is causing this.

@DevVersion - I already talked to Thomas about this and I think a better solution would be to debounce the model changes so that the animations don't attempt to fire so often.

Frank3K:
When trying to find out what the difference between my own testcase and the demo is I have found out that in the function `ngMessagesAnimation`:

- `enter` and `leave` do not trigger
- `addClass` and `removeClass` only trigger with `className` values `ng-active`, `ng-inactive` and `null`

I do not know if this is of any help...

DevVersion:
@topherfangio Seems to be a good idea, but I think we should still provide a force show / hide to prevent that issue when using the class selector. 

If you want me to create a Pull Request for debouncing the Invalid Model, just tell me :smile:

DevVersion:
I will close that PR, since @topherfangio is working on issues with ngMessages and debouncing the model will be even a better solution.

PR #6285, opened 2015-12-14
fix(autocomplete): fixes autocomplete required attribute parsing

bOOm-X:
The current behavior 
 - required -> value = "" -> return false (!!"" -> false)
 - required="true" -> value = "true" -> return true (!!"true" -> true)
 - required="false" -> value = "false" -> return true (!!"false" -> true)

What we want is 
 - required -> true
 - required="true" -> true
 - (optionally) required="false" -> false


bOOm-X:
I signed it!

bOOm-X:
I signed it!

PR #6292, opened 2015-12-14
feat(select): add attribute to allow permanent label showing.

### LINKED ISSUE #6286: How to keep showing the select label?
All selects (https://material.angularjs.org/latest/demo/select) hide the label when I focus.

Is it possible keep the label, for newbie users?

DevVersion:
This attribute allows the user to decide about showing the label permanently.
There is currently only a problem with bluring the select, but there is a PR by me, which fixes that too. See #6207

Fixes #6286


DevVersion:
Closing this as an outdated PR

Issue #5321, opened 2015-10-22
mdMaxlength: Always invalid for input type="number"


aaronPMorabito:
An input flagged with type="number" and md-maxlength="#" will always have a validation error.

Can be seen in this example under 'Errors': https://material.angularjs.org/latest/demo/input

Material Design v 1.0.0-rc1

topherfangio:
For clarification, the hourly rate input in the linked demo will always be invalid.

topherfangio:
@aaronPMorabito We are talking about whether or not `md-maxlength` is meaningful on a number input. Can you provide some feedback as to how you are using this?

We are planning to update the demo to not use `md-maxlength` on the Hourly Rate input to cause less confusion.

EladBezalel:
@topherfangio is this fixed?

topherfangio:
@EladBezalel This is partially fixed, we have removed the `md-maxlength` attribute from the demo so as not to cause confusion, but we still need to talk about whether or not it makes any sense for number inputs.

Was hoping to get some more feedback from @aaronPMorabito on how it might be used to see if we want to support it.

I say we close it for now and we can reopen if we get more feedback unless @ThomasBurleson gave it a P1 for a different reason.

aaronPMorabito:
@topherfangio I was only using it since it was in the demo and I thought it was fancy. I noticed that it was always causing validation errors when used in combination with <code>type="number"</code> so I reported it. I was using it to display the length of the string input and the max possible length of that string.

bitbytebot:
>We are talking about whether or not md-maxlength is meaningful on a number input. 

Why would you exclude such a function specifically for `<input type="number">`??? It does make sense and I actually need it in my projects.

You force me and many others to use `<input type="tel">` because of your exclusion.

topherfangio:
@bitbytebot We haven't specifically excluded this, it's just that based on the feedback above, we didn't see it as a high priority bug since we didn't see a good use-case.

If you have a valid use-case, please do share and we can discuss reopening it.

Also, is `<input type="tel" />` a good workaround, or does it cause problems?

Thanks!

bitbytebot:
@topherfangio I have a use for it, but now I'm using `<input type="tel" />` because it works similar to what I need.

I built a configurator for vehicle registration plates which have a maximum amount of number combinations (3). So this is a needed component, for me at least. I can think of multiple use-cases for this:
- zip code
- flight number
- vehicle identification number
- bank account number (IBAN in the SEPA-zone)

[...]

PR #6298, opened 2015-12-14
fix(input): Fix input errors CSS to properly display.

### LINKED ISSUE #5837: ngMessage not showing until form is touched

topherfangio:
Due to a CSS property that was not restrictive enough, standard error messages were hidden by default.

Restrict CSS to components with the `md-auto-hide` class. Note that this unfortunately re-introduces the message flicker bug for messages which do not use the `md-auto-hide`. This should be fixed when https://github.com/angular/angular.js/issues/12969 is merged and available.

Also, remove the `md-maxlength` requirement from the errors Hourly Rate demo to reduce confusion.

Fixes #5837. References #5321.

### DIFF
--- a/src/components/input/demoErrors/index.html
+++ b/src/components/input/demoErrors/index.html
@@ -35,7 +35,7 @@
       <md-input-container class="md-block">
         <label>Hourly Rate (USD)</label>
         <input required type="number" step="any" name="rate" ng-model="project.rate" min="800"
-               max="4999" ng-pattern="/^1234$/" md-maxlength="20" />
+               max="4999" ng-pattern="/^1234$/" />
 
         <div ng-messages="projectForm.rate.$error" multiple md-auto-hide="false">
           <div ng-message="required">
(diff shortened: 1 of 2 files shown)


michaelkrog:
Does this also fix the autocomplete issues?
https://github.com/angular/material/issues/5837#issuecomment-162059227

topherfangio:
@michaelkrog If your issue was specific to the CSS, then this should fix it as well. However, I updated your Codepen to use my local branch and it doesn't look like the messages are even in the DOM.

Can you take a quick look at your Codepen (http://codepen.io/anon/pen/yeLPVy) and see if it's missing something that you can easily fix? I'd love to test this against the Autocomplete and make sure it works.

michaelkrog:
Hi @topherfangio.

Something definitely changed in autocomplete. It used to be in the DOM in rc6, but broken because of CSS. Now it seems to be found nowhere i the DOM.

I'll create an issue for it.

PR #6354, opened 2015-12-16
fix(select): Position incorrect if selection scrolled.

### LINKED ISSUE #6190: md-select misplaced
md-select is misplaced when you select something far down the list.  The further down the list, the more misplaced.

image

1.0-rc6, 1.0-rc7 and master.

Tested in Chrome, Firefox and Edge on Windows 10.

topherfangio:
If the currently selected object scrolled into view because it was further down in the list, the position of the menu was incorrect upon opening because the `scrollTop` of the `contentNode` was always `0` due to the ARIA changes which set it to `display: none` upon closing.

Fix by modifying the display to `block` when we run the `isScrollable` check and back afterward (usually `display: none` but we account for other options).

Also fixes a slight `2px` positioning issue which made the menu not perfectly line up with the underlying selection.

Fixes #6190.

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -1276,7 +1276,7 @@ function SelectProvider($$interimElementProvider) {
      * Calculate the
      */
     function calculateMenuPositions(scope, element, opts) {
-      var 
+      var
         containerNode = element[0],
         targetNode = opts.target[0].children[0], // target the label
         parentNode = $document[0].body,
@@ -1298,13 +1298,13 @@ function SelectProvider($$interimElementProvider) {
           bottom: bounds.bottom - (targetRect.top + targetRect.height)
         },
         maxWidth = parentRect.width - SELECT_EDGE_MARGIN * 2,
-        isScrollable = contentNode.scrollHeight > contentNode.offsetHeight,
         selectedNode = selectNode.querySelector('md-option[selected]'),
         optionNodes = selectNode.getElementsByTagName('md-option'),
-        optgroupNodes = selectNode.getElementsByTagName('md-optgroup');
+        optgroupNodes = selectNode.getElementsByTagName('md-optgroup'),
+        isScrollable = calculateScrollable(element, contentNode),
+        centeredNode;
 
       var loading = isPromiseLike(opts.loadingAsync);
-      var centeredNode;
       if (!loading) {
         // If a selected node, center around that
         if (selectedNode) {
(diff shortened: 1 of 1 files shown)


topherfangio:
@rschmukler This is ready for review.

@ThomasBurleson If he happens to not be around, this is ready for review and Naomi would like it in the next release.

src/components/select/select.js
```diff
@@ -1437,5 +1438,19 @@ function SelectProvider($$interimElementProvider) {
       height: node.offsetHeight
     } : {left: 0, top: 0, width: 0, height: 0};
   }
+
+  function calculateScrollable(element, contentNode) {
+    var oldDisplay = element[0].style.display;
+
+    // Set the element's display to block so that this calculation is correct
+    element[0].style.display = 'block';
```
ThomasBurleson: this should be in a `try { } finally { }.`

src/components/select/select.scss
```diff
@@ -11,7 +11,7 @@ $select-max-visible-options: 5;
   top: 0;
   z-index: $z-index-select;
   opacity: 0;
-  display: none;
+  //TODO: Re-add - display: none;
```
ThomasBurleson: ?? Forgotten ?

rschmukler:
LGTM

src/components/select/select.scss
topherfangio: Uh, yup. Thanks! Will get this fixed.

src/components/select/select.js
```diff
@@ -1276,7 +1276,7 @@ function SelectProvider($$interimElementProvider) {
      * Calculate the
      */
     function calculateMenuPositions(scope, element, opts) {
-      var 
+      var
```
EladBezalel: Can you make all those with var? or without it cause there are var deceleration right after..

topherfangio: Note how all of the following variable declarations have a comma `,` at the end except the last one, which has a semi-colon. The semi-colon is the stop of the `var` statement basically, so yes, you can do it :-)

topherfangio:
@ThomasBurleson Updated with your suggestions. Ready for review/merging.

src/components/select/select.js
EladBezalel: It wasn't a question.. i know you can aggregate var declarations into one.

I was asking you to separate it or make the following var declarations (after the semi-colon) aggregated too.

topherfangio: Ah, sorry, I see now! I originally had the style code above/below this line which is why I separated it, but I can indeed move those back together now.

src/components/select/select.js
```diff
@@ -1437,5 +1437,24 @@ function SelectProvider($$interimElementProvider) {
...
+
+      isScrollable = contentNode.scrollHeight > contentNode.offsetHeight;
+
+      // Reset it back afterwards
+      element[0].style.display = oldDisplay;
+    } finally {
+      // Nothing to do
+    }
+    return isScrollable;
+  }
```
jelbourn: I unit test would be nice to have, but I want to get this in 1.0.1, so it can be added later.

PR #6377, opened 2015-12-17
Remove unused/commented code

levindixon:


levindixon:
I signed it!

DevVersion:
I think, we kept it, to re-add it easily. So removing isn't a good idea?

levindixon:
@DevVersion I'd disagree, keeping the code base clean should be a priority over leaving in stale code for potential usage later on.

DevVersion:
I'll discuss this with the team and let you know.

levindixon:
Sounds good, thanks!

PR #6379, opened 2015-12-17
Fix(tabs): IE is not processing the inline style for max-width

austinthedeveloper:
This issue was only found on IE:

max-width-before

IE will not evaluate this correctly. Switching to ng-style alleviates the issue:

max-width-after


austinthedeveloper:
This has been corrected. I added an email to the CLA.

ThomasBurleson:
Already fixed.

PR #6400, opened 2015-12-19
fix(input): match up all label transitions

### LINKED ISSUE #6328: Inconsistent input label animation speeds
An input's floating label animation is quicker when using a `label` tag compared to just setting the `placeholder` attribute.

Example

tjlav5:
Change all label `transitions` to be `0.25s` long, as the default is.

Fixes #6328

### DIFF
--- a/src/components/input/input.scss
+++ b/src/components/input/input.scss
@@ -112,7 +112,7 @@ md-input-container {
     @include rtl(padding-right, 0, $input-container-padding + 1px);
     z-index: 1;
     transform: translate3d(0, $input-label-default-offset + 4, 0) scale($input-label-default-scale);
-    transition: transform $swift-ease-out-timing-function 0.25s;
+    transition: transform $swift-ease-out-duration $swift-ease-out-timing-function;
 
     @include rtl(transform-origin, left top, right top);
   }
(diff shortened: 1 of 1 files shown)


tjlav5:
I signed it!

ThomasBurleson:
@topherfangio - needs review and validation with Spec.

topherfangio:
Spec-wise, I think this is compliant (I can't find anything that specifies exactly what it should be).

However, since we're using a special timing, I'd like to see it as a variable (perhaps `$md-float-animation-duration` or similar) so that we know that it is custom and we can easily change it in one place.

tjlav5:
@topherfangio good point. I'll consolidate under a variable tonight. Cheers

tjlav5:
@topherfangio considering that the transitions all use the $swift-ease-out-timing-function, should the timing duration ALSO be using $swift-ease-out-duration?

topherfangio:
@tjlav5 Yes, it should use the ease-out duration. Thanks for updating!

I am restarting the Travis build to ensure all the tests pass (there was an unrelated build issue which was causing the failure).

EladBezalel:
Bumped to Topher as it seems that he's in charge here :smile:

ThomasBurleson:
please Rebase with latest from `master`

tjlav5:
@ThomasBurleson sorry missed the rebase request. But it looks like you handled it.
Cheers

Issue #6313, opened 2015-12-15
md-autocomplete(1.0.0+): ng-messages are not shown


michaelkrog:
I just tried creating a blank pen via the getting started page(release 1.0.0).

I then added the example with ngmessages from https://material.angularjs.org/latest/api/directive/mdAutocomplete. 
```html
<form name="autocompleteForm">
  <md-autocomplete required md-input-name="autocomplete" md-selected-item="selectedItem" md-search-text="searchText" md-items="item in getMatches(searchText)" md-item-text="item.display">
    <md-item-template>
      <span md-highlight-text="searchText">{{item.display}}</span>
    </md-item-template>
    <div ng-messages="autocompleteForm.autocomplete.$error">
      <div ng-message="required">This field is required</div>
    </div>
  </md-autocomplete>
</form>
```

But the ng-messages are completely removed from the DOM.
http://codepen.io/anon/pen/ZQQObj

Nazanin1369:
@topherfangio Any one working on this? Can I go ahead?

mimani:
@Nazanin1369 I have tried to fix this in this [PR](https://github.com/angular/material/pull/6419), Let me know if it seems OK.

topherfangio:
@Nazanin1369 I am going to dig into this over the next day or two, but if you have any insights already, the knowledge would certainly be appreciated!

iamisti:
@ThomasBurleson 
Maybe we should change the milestone to `1.0.3`, since the fix is not in `1.0.2`, right?
Can you guys give me some insights when this fix will be released?

michaelkrog:
Updated pen to use 1.0.5, but it is still not possible to use ng-messages in autocomplete.
[The documentation](https://material.angularjs.org/latest/api/directive/mdAutocomplete) still states that it should be possible, but it is not.

New pen: http://codepen.io/anon/pen/obKBej

chillyspoon:
This bug is visible in the 1.0.4 (the version I'm using) and 1.0.5 Angular Material demos - see attached. The state length error shows correctly but no amount of messing with a blank/null state value can prompt the "required" message to display.

![screenshot 2016-02-27 11 31 49](https://cloud.githubusercontent.com/assets/6490625/13372679/34a3022a-dd46-11e5-84c1-5b04fde5404b.png)
[...]

Issue #6138, opened 2015-12-07
ng-messages only works in same input-container as field


dnd:
I have a form that on all display sizes displays two form elements on one row, with one of the elements being a very small "amount" input holding up to 5 characters. Because of this there is no room to put the ng-messages div in the same container as it, as it would just cause the required message to be 3 lines high. Instead I tried putting it in the container of the element next to it, since it takes up the rest of the line. The message will not show if this is done though. The form element turns red as expected, but no message shows. 

Angular material shouldn't care where I put my ng-messages div. I link it to the form element by specifying the field to bind to. Angular Material should just take the div and stick it in the space below the field. I understand that this might cause the line height to change if errors render for both fields, but especially in this situation the line height change is still less because under the larger field it's 2 lines of text vs 3 lines of text if placed under the "amount" element.

topherfangio:
Can you please try to put the `ng-messages` div outside of an input container, but still inside the form? Perhaps placing it just below both of your other input containers?

You may not get some of the fancy styling/animations that the input containers provide, but I believe this should work.

If you can provide a Codepen demoing the issue, I would be more than happy to look for a workaround, or see if something in our code is causing an issue.

dnd:
Here's a pen that shows what happens when it is outside of a container. http://codepen.io/anon/pen/KVVLpd

You can ignore the third example for now. Locally for some reason it is sort of working with the messages inside the autocomplete as of master just now. No idea why, but I can't reproduce in the pen yet.

topherfangio:
There is an existing/known issue (#6313) with the required validator of Autocomplete, so this may be causing some additional problems as well.

brianfeister:
+1

PR #6419, opened 2015-12-21
Changes to fix Autocomplete validation

mimani:
This fix is related to bug: [6138](https://github.com/angular/material/issues/6138) and [6313](https://github.com/angular/material/issues/6313), With auto-complete, error message are not getting displayed.

This is because, error messages are shown only when ng-messages are in same input-container as input field. I have made changes so that all validation messages can be put in separate element and have places it with input container.

### DIFF
--- a/src/components/autocomplete/demoFloatingLabel/index.html
+++ b/src/components/autocomplete/demoFloatingLabel/index.html
@@ -20,11 +20,13 @@
           <md-item-template>
             <span md-highlight-text="ctrl.searchText">{{item.display}}</span>
           </md-item-template>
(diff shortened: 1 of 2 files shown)


topherfangio:
At this point (since we just released 1.0.0), we really don't want to change the API (i.e. by adding a `md-error-template` element).

Additionally, `ng-messages` *should* work outside of an input container, you just won't get the fancy animations. So, if that is what is failing, then there is likely some bigger issue at play here that we want to address.

Thanks for digging into this and taking a look though!

mimani:
@topherfangio Thanks for your feedback. In that case, this may not be proper fix, but I feel going forward we may like to put ng-message related content inside some dedicated element as I have done, so that each type of html is nicely separated from each other.

PR #6425, opened 2015-12-21
feat(input): allow skip hidden inputs

### LINKED ISSUE #2153: md-input-container should ignore hidden fields as part of the inputs its counting.
I understand the purpose of having only a single *visible* input per container but this should not apply to hidden fields. Hidden fields are often used to track default parameters and therefore should be logically organized near the visible input. Consider the following example:

```html
<md-input-container class="md-default-theme">
      <input name="account[remember_me]" type="hidden" value="0" class="md-input" id="input_001">
      <input type="checkbox" value="1" name="account[remember_me]" id="account_remember_me">
      <label for="account_remember_me">Remember me</label>
</md-input-container>
```

DevVersion:
inputs with type `hidden` will be skipped by the `input-container`

#breaking

fixes #2153

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -33,6 +33,11 @@ angular.module('material.components.input', [
  * Input and textarea elements will not behave properly unless the md-input-container
  * parent is provided.
  *
+ * A single `<md-input-container>` should contain only one `<input>` element, otherwise it will throw an error.
+ *
+ * <b>Exception:</b> Hidden inputs (`<input type="hidden" />`) are ignored and will not throw an error, so
+ * you may combine these with other inputs.
+ *
  * @param md-is-error {expression=} When the given expression evaluates to true, the input container
  *   will go into error state. Defaults to erroring if the input has been touched and is invalid.
  * @param md-no-float {boolean=} When present, `placeholder` attributes on the input will not be converted to floating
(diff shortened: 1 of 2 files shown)


topherfangio:
I think this warrants a note in the docs about how we skip inputs. Otherwise, LGTM.

DevVersion:
@topherfangio - I added a note to the `<md-input-container>` directive. Is that note enough?

PS: CI is still failing due Karma (will be fixed in ws@1.0.0)

topherfangio:
Let's change the wording to this:

--
A single `<md-input-container>` should contain only one `<input>` element, otherwise it will throw an error.

**Exception:** Hidden inputs (`<input type="hidden" />`) are ignored and will not throw an error, so
you may combine these with other inputs.

DevVersion:
If I got your new wording right (with the divider and so on), it should look like this? (bold exception seems to be smaller as the actual text is). 



Please take a look @topherfangio

DevVersion:
If I got your new wording right (with the divider and so on), it should look like this? (bold exception seems to be smaller as the actual text is). 



Still needs work? @topherfangio

DevVersion:
@topherfangio - Can you review the documentation message?

topherfangio:
@ThomasBurleson LGTM; not sure which milestone you want this in. Can you please take a look?

topherfangio:
I'm going to go ahead and delay this until 1.1.0 as it could affect some of our existing users.

@DevVersion Can you please update the commit message with a BREAKING CHANGE note about how we now skip inputs. We do it in the commit message because our Changelog process automatically pulls from there when noting breaking changes.

DevVersion:
@topherfangio Done - Just updated the PR description too, so you can see it there.

PR #6430, opened 2015-12-21
chore: Add utility to take PhantomJS screenshots.

topherfangio:
Oftentimes, a PhantomJS test will fail while it passes on other browsers like Chrome or Firefox. This can be difficult to track down if it is related to layout/positioning as you cannot see what the browser looks like.

Add a new utility to capture a screenshot during your spec so that you can more easily debug the problem.

### DIFF
--- a/phantom-screenshot.js
+++ b/phantom-screenshot.js
@@ -1,0 +1,41 @@
+/*
+ * Provides a test utility for taking screenshots of PhantomJS. Add the function below to your
+ * `*.spec.js` file and call `screenshot('my-screenshot')` from within the test. Make sure to call
+ * `screenshot('my-last-screenshot', true)` to kill this script, or you can manually kill it with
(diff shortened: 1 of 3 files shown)


DevVersion:
Wow, looking forward to that new feature.

phantom-screenshot.js
```diff
@@ -0,0 +1,46 @@
...
+ *       if (typeof window.callPhantom != 'undefined') {
+ *         window.callPhantom({ filename: filename + '.png', done: done });
+ *       }
+ *     }
+ *
+ */
+var page2 = require("webpage").create();
+
+page2.viewportSize = {
+  width: 1400,
```
ThomasBurleson: These should be loaded from `karma.config.js`

ThomasBurleson:
Closing as this is a separate effort lead by @robertmesserle

Issue #6411, opened 2015-12-21
sidenav like inbox.google.com


justechn:
Hey guys,

Thanks for this great app. I have just started using it, and I have a question about the sidenav. I would like mine to behave like inbox.google.com

With inbox the sidenav opens under the toolbar, and it does not grey out the main page. how can I duplicate this in angular material?

DevVersion:
At the moment it is possible to place the sidenav under the toolbar, easily use flexbox with rows to align it under the toolbar.

The actual problem is the backdrop, maybe we should provide an attribute to disable the backdrop.

justechn:
Thanks for the suggestions. 

I have been playing with the sidenav for a couple days and it seem like it does not honor the flexbox layout unless I use the md-is-locked-open expression. Am I missing something? 

also, while aligning the sidenav under the toolbar I first had to use the layout column which causes the locked sidenav to not fill the screen vertically. I even put a layout="row" on the sidenav itself, but it still only goes down to the bottom of the content, not the bottom of the page. Is there a way to use both column and row layout at the same time so I can fill the page both horizontally (for the toolbar) and vertically (for the sidenav)?

Thanks again.

Todd-Werelius:
As a wider discussion, unless I missed this topic elsewhere, the spec provides for 5 modes with a total of 7 variants, listed below for reference ( as of today anyway ) I think the answer should always be, support the spec first, leave customization up to the user or at least defer it until the spec is completely supported.

As far as I can tell support only exists for the following in Angular Material 

1A - Permanent Full Height - but it requires overly complex strategies to work on all device sizes if you want to transition from one mode to the next, i.e. it's not at all easy, as a matter of fact a better approach for permanent is just design your own column that acts that way at the moment and hide it based on a media query ( Am I wrong about this? )   

4A - Temporary 
Really only 4A works as expected without a lot of thought needed.

As it is now I would have to work pretty hard to create a transition from permanent, to persistent, to temporary dependent on device capability 

It seems like a better semantic approach from a users perspective would be

<md-sidnav md-sidnav-mode="permanent | persistent | temporary" md-sidenav-mini="true | false">

With media query support of course, that would not make it any easier to implement, just easier to understand.  

**Material Design Spec  **

1. Permanent - Always Open - No Backdrop 
1A Full Height ( does not overlap content )   
2B Clipped under app bar  
3C Floating  
[...]

PR #6434, opened 2015-12-21
feat(sidenav, bottomsheet): allow to disable backdrop 

### LINKED ISSUE #2846: Add disableBackdrop to BottomSheet Options
Allow for the ability to disable the backdrop overlay creation when opening the bottom sheet.

### LINKED ISSUE #6473: feat(bottomsheet) open without backdrop
just like sidenav.

DevVersion:
Allows the user to disable the backdrop.

Component to #6411

Fixes #6473 Fixes #2846

### DIFF
--- a/src/components/bottomSheet/bottom-sheet.js
+++ b/src/components/bottomSheet/bottom-sheet.js
@@ -85,6 +85,7 @@ function MdBottomSheetDirective($mdBottomSheet) {
  *   of 3.
  *   - `clickOutsideToClose` - `{boolean=}`: Whether the user can click outside the bottom sheet to
  *     close it. Default true.
+ *   - `disableBackdrop` - `{boolean=}`: When set to true, the bottomsheet will not show a backdrop.
  *   - `escapeToClose` - `{boolean=}`: Whether the user can press escape to close the bottom sheet.
  *     Default true.
  *   - `resolve` - `{object=}`: Similar to locals, except it takes promises as values
@@ -143,6 +144,7 @@ function MdBottomSheetProvider($$interimElementProvider) {
       themable: true,
       onShow: onShow,
       onRemove: onRemove,
+      disableBackdrop: false,
       escapeToClose: true,
       clickOutsideToClose: true,
       disableParentScroll: true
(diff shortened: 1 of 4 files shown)


EladBezalel:
LGTM, @ThomasBurleson Do we really want to support this behavior ?

EladBezalel:
@DevVersion do you mind adding tests?

DevVersion:
@EladBezalel - There they are :) Tests are done

src/components/sidenav/sidenav.js
```diff
@@ -229,6 +230,7 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
     var lastParentOverFlow;
     var triggeringElement = null;
     var promise = $q.when(true);
+    var isBackdropDisabled = !!attr.mdDisableBackdrop;
```
DevVersion: @ThomasBurleson In general validations for attributes are not good enough (for example here only adding `md-disable-backdrop` won't work. I'd really like to update a few attributes to validate `false` and `true` and also an empty attribute. This will be easy if we have our util method, which is created in #6803

src/components/bottomSheet/bottom-sheet.js
```diff
@@ -85,6 +85,8 @@ function MdBottomSheetDirective($mdBottomSheet) {
  *   of 3.
  *   - `clickOutsideToClose` - `{boolean=}`: Whether the user can click outside the bottom sheet to
  *     close it. Default true.
+ *   - `disableBackdrop` - `{boolean=}`: When set to true, the bottomsheet will not show a backdrop.
+ *     clicking to close the bottomsheet. Default true.
```
topherfangio: Text seems incorrect here: `clicking to close the bottomsheet. Default true.`?

DevVersion: Thanks, didn't saw that, because I recently changed the attribute name from `useBackdrop` to `disableBackdrop`

antonellopasella:
any ETA about this?
Will be in 1.1?

Thankz!

ThomasBurleson:
@antonellopasella - this is already in HEAD and v1.1.0-rc.5... and will be available with the formal release of v1.1.0.

Issue #6422, opened 2015-12-21
md-toolbar with a md-warn button don't apply the default  theme


JA44:
This issue on codepen: http://codepen.io/anon/pen/RrRpeL

My second button should be red instead of white.

hewaroth:
Should be red... the same as first?

JA44:
No like the button below the md-toolbar on this example: http://codepen.io/anon/pen/RrRpeL

topherfangio:
Indeed, seems like the `md-warn` CSS selector is not specific enough.

spamguy:
What's the status of this? I'm having the same problem, but with md-accent and icon buttons.

```
<md-toolbar class="md-toolbar-tools" id="mapToolbar">
    <span flex></span>
    <md-button class="md-accent md-icon-button" aria-label="Hold" hide-xs>
        <md-icon md-svg-icon="join" class="md-accent"></md-icon>
        <md-tooltip>Hold</md-tooltip>
    </md-button>
</md-toolbar>
```

Expected: Toolbar renders in the primary colour, icon button renders as accent colour (red, in my case). Both the icon and button are md-accent to remove doubt the class is misplaced.
Actual: Button renders as white. The ripple colour is reddish, clearly influenced by md-accent.

If I swap out md-accent with a class of my own, the button is coloured correctly.

ThomasBurleson:
This issue is closed as part of our deprecation effort.
      For details, see our post [Spring Cleaning 2016](https://groups.google.com/forum/#!topic/ngMaterial/3Ip9nd68bfk).

PR #6447, opened 2015-12-22
fix(toolbar): apply the warn color, accent color

JA44:
This PR to clean this issue #6422 in order to apply the warn and accent theme on the toolbar.


JA44:
I signed it!

PR #6452, opened 2015-12-23
feat(checkbox): validate numeric booleans too

### LINKED ISSUE #2796: mdCheckbox: support values with 0,1
Current checkbox value works only with true, false. Wouldn't it be better if it works with 1,0 either? Because mostly when you get data from database you get 1,0 instead of true, false.
```
// current approach
$scope.data.cb4 = false;
$scope.data.cb5 = true;

// new approach
$scope.data.cb2 = 0;
$scope.data.cb3 = 1;
```

DevVersion:
We should format numeric booleans `0` or `1` to real booleans.

Fixes #2796

### DIFF
--- a/src/components/checkbox/checkbox.js
+++ b/src/components/checkbox/checkbox.js
@@ -127,6 +127,11 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
 
       ngModelCtrl.$render = render;
 
+      ngModelCtrl.$formatters.push(function(modelValue) {
+        if (modelValue == '1' || modelValue == '0') return !!modelValue;
+        return modelValue;
+      });
+
       function $$watchExpr(expr, htmlAttr, valueOpts) {
         if (attr[expr]) {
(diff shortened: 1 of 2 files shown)


topherfangio:
I think since we have the `ng-true-value` and `ng-false-value` parameters, this is not actually necessary. However, it is a very nice convenience. If we're going to add this and make this the default, though, then I think we should expand this PR to support any truthy/falsy values, not just `true`/`false` and `'0'`/`'1'`.

Also, looking into the future, what number/value would we use to represent a mixed-state checkbox? The first thought that comes to mind is `-1`, but if we don't currently implement that functionality, then it will start out where that value is truthy, and we'd have to change it later to be mixed...

@ThomasBurleson Thoughts?

DevVersion:
@topherfangio I agree with you. I will close that PR and update the issue and provide a more convenient solution.

ajsb85:
Example to use:
```HTML
      <md-checkbox ng-true-value="1" ng-false-value="0"/>
```

PR #6456, opened 2015-12-24
test(util): add needed test for `disableScrollAround`

### LINKED ISSUE #2305: mdUtils: unit tests needed for `disableScrollAround()`

DevVersion:
- We can append an element to the body which is higher than the actual window, so there should be space to scroll.
- So we disable scrolling with our specified method `disableScrollAround`.
- Then we try to scroll down to `y=1000px`
- So if `disableScrollAround` works, the `pageYOffset` will be zero.

Fixes #2305

### DIFF
--- a/src/core/util/util.spec.js
+++ b/src/core/util/util.spec.js
@@ -243,6 +243,23 @@ describe('util', function() {
       });
     });
 
+    describe('disableScrollAround', function() {
+      it('should prevent scrolling of the passed element', inject(function($mdUtil) {
+        var element = angular.element('<div style="height: 2000px">');
+        document.body.appendChild(element[0]);
+
+        $mdUtil.disableScrollAround(element);
+
+        window.scrollTo(0, 1000);
+
+        expect(window.pageYOffset).toBe(0);
+
+        // Revert scroll
+        window.scrollTo(0, 0);
+        document.body.removeChild(element[0]);
+      }));
+    });
+
     describe('nextTick', function() {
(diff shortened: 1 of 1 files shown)


topherfangio:
I am not 100% sure that this test works as expected. For instance, if I edit `$mdUtil.disableScrollAround()` to simply return instead of executing, the test still passes. I would expect it to fail if the code did nothing.

Similarly, if I simply remove the call to `disableScrollAround()` within the test, it still passes.

@DevVersion Can you verify that the test fails properly if the code isn't implemented?

DevVersion:
Yes @topherfangio, 
The test works as expected. If I remove the implementation of the `disableScrollAround()` the test will fail. 

Here a screenshot if removing the implemenation.

topherfangio:
@DevVersion Indeed, my apologies. I had my Karma setup to only run on PhantomJS where this passes.

@ThomasBurleson Since it fails properly on Firefox, is it okay that this test erroneously passes on PhantomJS? If so, LGTM.

DevVersion:
@topherfangio - I think the most tests will only fail in `firefox`, because it's providing a real browser and not a headless browser as `phantomjs` do.

DevVersion:
Any news on there? - I think that's normal, PhantomJS will never pass this test, because it's just a headless browser.

DevVersion:
@ThomasBurleson Ping for feedback to Topher's question.

topherfangio:
@DevVersion I have updated the milestone to 1.1.1 and I think since this is just a test, we can go ahead and merge it. Thomas will review before actually merging it, so he can ask questions/change the label then 😄

PR #6484, opened 2015-12-28
fix(bottomsheet): fix open animation for bottomsheet.

### LINKED ISSUE #6468: bottomsheet slideIn animations not working.
bottomsheet open animation sometimes (most of the time) doesnt happen.  the close animation seems to happen regularly.

https://material.angularjs.org/1.0.1/

DevVersion:
This is fixing all the issues with close and open animation.
It work's perfectly, I already took a look at the Timings in the Chrome Dev Tools (Render Rate etc.)

But to be honest (in code) I'm not satisfied with using `$timeout`. But that's the only possibility because we inserting the element into the `DOM` at `80px` and then we wan't instant run the animation from `-100%`. But that's not possible. So we need to remove the transition and transform the bottomsheet at insertion down to `-100%`, and then after it got rendered down, we can fire the animation to `80px`. 

That's the only possibility to get run with $animate (entering / leaving). 

Fixes #6468

### DIFF
--- a/src/components/bottomSheet/bottom-sheet.js
+++ b/src/components/bottomSheet/bottom-sheet.js
@@ -127,6 +127,7 @@ function MdBottomSheetDirective($mdBottomSheet) {
 function MdBottomSheetProvider($$interimElementProvider) {
   // how fast we need to flick down to close the sheet, pixels/ms
   var CLOSING_VELOCITY = 0.5;
+  var SWIFT_EASE_OUT_DURATION = 400; // same as constants css
   var PADDING = 80; // same as css
 
   return $$interimElementProvider('$mdBottomSheet')
@@ -136,7 +137,7 @@ function MdBottomSheetProvider($$interimElementProvider) {
     });
 
   /* @ngInject */
-  function bottomSheetDefaults($animate, $mdConstant, $mdUtil, $mdTheming, $mdBottomSheet, $rootElement, $mdGesture) {
+  function bottomSheetDefaults($animate, $mdConstant, $mdUtil, $mdTheming, $mdBottomSheet, $rootElement, $mdGesture, $timeout, $q) {
     var backdrop;
 
     return {
@@ -158,7 +159,7 @@ function MdBottomSheetProvider($$interimElementProvider) {
 
       if (options.clickOutsideToClose) {
         backdrop.on('click', function() {
-          $mdUtil.nextTick($mdBottomSheet.cancel,true);
+          $mdUtil.nextTick($mdBottomSheet.cancel, true);
         });
       }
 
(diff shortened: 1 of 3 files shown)


manish2535:
In order to get the animation upwards, can you not do this instead?

```
md-bottom-sheet.ng-enter-active {
    -webkit-transform: translate3d(0, 100%, 0) !important;
    transform: translate3d(0, 100%, 0) !important;
}
```

What this will do is render the BottomSheet below the view-port and animate the bottom-sheet up once the rendering is complete. I tried it and it seems to work really nicely.

DevVersion:
That's not as simple as you think. I tested several possibilities. But as I already wrote in the PR description, the element will be inserted at `80px` and at the same time we try to move that down (but there is a transition duration) and at the same time we wan't to start the open animation. All in one, the `ng-enter` and `ng-leave` animation can be only fixed with a delay and advanced css as you can see in this PR

manish2535:
Have a look at this Codepen - 

http://codepen.io/anon/pen/PZbxLY

Does this do what you were aiming for? 

Note: There seems to be some Scroll Issue when this fix is applied on Codepen. I dont get that on my App.

DevVersion:
Your animation seem's to go in the complete wrong direction. To get that running, we need to think about the steps, which will executed by the browser and the `$aniamte` service. 

- Inserts Element to DOM at (default style: 80px)
- ng-enter class tries to get the bottom-sheet down (but there is a transition duration defined [here](https://github.com/angular/material/blob/master/src/components/bottomSheet/bottom-sheet.scss#L21))
- So we need to wait by default, to get **that** version running , [the render time while inserting to dom] + [transition time to get the bottomsheet down]
- And after that we can simply run the open animation.

But my current fix is much more transparent then waiting for [the render time while inserting to dom] + [transition time to get the bottomsheet down]

My current PR Versions looks like this

src/components/bottomSheet/bottom-sheet.js
```diff
@@ -175,24 +176,37 @@ function MdBottomSheetProvider($$interimElementProvider) {
...
+            if (options.escapeToClose) {
+              options.rootElementKeyupCallback = function(e) {
+                if (e.keyCode === $mdConstant.KEY_CODE.ESCAPE) {
+                  $mdUtil.nextTick($mdBottomSheet.cancel,true);
+                }
+              };
+              $rootElement.on('keyup', options.rootElementKeyupCallback);
+            }
+          });
+      }, 10);
```
EladBezalel: 10?

src/components/bottomSheet/bottom-sheet.js
```diff
@@ -175,24 +176,37 @@ function MdBottomSheetProvider($$interimElementProvider) {
...
+
+      // Wait for element rendering below bottom
+      $timeout(function() {
+
+        // manually trigger the animation
+        $timeout(function() {
+          element.removeClass('md-prepare-enter');
+        }, SWIFT_EASE_OUT_DURATION);
+
+        $animate.enter(bottomSheet.element, options.parent)
```
EladBezalel: Promise is no longer returned from here..

src/components/bottomSheet/bottom-sheet.js
DevVersion: Yes we should give the element a free time to complete the insertion, so we can run the animation clean.

DevVersion:
@EladBezalel - now resolving promise.

DevVersion:
Should be better now - @EladBezalel

DevVersion:
@EladBezalel - This one is also improved, should be better now. Still any suggestions?

manish2535:
Thanks @DevVersion. I see your point.

src/components/bottomSheet/bottom-sheet.js
```diff
@@ -175,31 +176,48 @@ function MdBottomSheetProvider($$interimElementProvider) {
...
-            element[0].querySelector('a') ||
-            element[0].querySelector('[ng-click]')
-          );
-          focusable.focus();
-
-          if (options.escapeToClose) {
-            options.rootElementKeyupCallback = function(e) {
-              if (e.keyCode === $mdConstant.KEY_CODE.ESCAPE) {
-                $mdUtil.nextTick($mdBottomSheet.cancel,true);
+      element.addClass('md-prepare-enter');
```
EladBezalel: I think it would be better off to use $animate for adding and removing classes,
Some users wants to disable all the animations..

EladBezalel:
I truly think you can make it better with promises.
you should use $animate for adding and removing a class, by using this with the promise api this code will be both clear and beautiful at once :smile:

DevVersion:
@EladBezalel That's a really good point with the $animate classes, never thought about user who disable animations. Are animations not part of the material design?

EladBezalel:
they're but when doing a protractor testing you want to disable animations

src/components/bottomSheet/bottom-sheet.js
DevVersion: @EladBezalel - Just tested it, it's the same as in the sidenav, it needs to run in the same digest and needs to be applied before the actual timeout and `$animate.enter`, so this isn't working.

DevVersion:
@EladBezalel Are there still some things to do?

DevVersion:
I mark this PR, to require more work, because I think there must be a more elegant way, than just applying any "random" number for a timeout.

DevVersion:
Closing this as PR, because we're filtering old PR's - This one is obviously the wrong approach. 

There must be another solution to fix the animation, instead of delaying the animation.

trstowell:
Please consider re-opening and merging this fix. Bottom sheets are a critical part of our UI and the broken animation makes for a messy experience. A non-elegant solution is better than a non-functional component.

PR #6486, opened 2015-12-28
feat(sidenav): track sidenav vissibilty and enter / leave backdrop if needed

### LINKED ISSUE #4595: Sidenav component does not automatically handle entering hidden state.
If a sidenav component is in an open/active state when a resize occurs and the resize causes the sidenav to go be hidden (for example hide-gt-sm is applied), I would expect the nav handle itself needing to enter a closed/nonactive state.  Additionally, you can request to toggle the sidenav into an active state while it is hidden.

Both can be replicated here: http://codepen.io/josephperrott/pen/jbbpEq

Reproducables

Resize:
Open the page and size the window small enough for the sidenav to be available, toggle it open and then resize the page large enough for the sidenav to be hidden.  The overlay remains present, it then requires a click anywhere on the page for it to "close".

Toggle:
Open the page large enough that the sidenav cannot be displayed, then click the toggle sidenav button.

DevVersion:
- Revert scope value if sidenav is shown.
- Throttle Resize Events

Fixes #4595

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -208,7 +208,7 @@ function SidenavFocusDirective() {
  *   - `<md-sidenav md-is-locked-open="$mdMedia('min-width: 1000px')"></md-sidenav>`
  *   - `<md-sidenav md-is-locked-open="$mdMedia('sm')"></md-sidenav>` (locks open on small screens)
  */
-function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, $compile, $parse, $log, $q, $document) {
+function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, $compile, $parse, $log, $q, $document, $$rAF, $window) {
   return {
     restrict: 'E',
     scope: {
@@ -229,6 +229,9 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
     var lastParentOverFlow;
     var triggeringElement = null;
     var promise = $q.when(true);
+    var skipSidenav = false;
+    var skipNextUpdate = false;
+    var windowElement = angular.element($window);
 
     var isLockedOpenParsed = $parse(attr.mdIsLockedOpen);
     var isLocked = function() {
@@ -244,12 +247,18 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
 
     $mdTheming.inherit(backdrop, element);
 
+    var throttleResize = $$rAF.throttle(revalidateVisibility);
+    windowElement.on('resize', throttleResize);
+    revalidateVisibility();
+
     element.on('$destroy', function() {
       backdrop.remove();
       sidenavCtrl.destroy();
+      windowElement.off('resize', throttleResize);
     });
 
     scope.$on('$destroy', function(){
+      windowElement.off('resize', throttleResize);
       backdrop.remove()
     });
 
@@ -257,8 +266,28 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
     scope.$watch('isOpen', updateIsOpen);
 
 
+
     // Publish special accessor for the Controller instance
     sidenavCtrl.$toggleOpen = toggleOpen;
+
+    function revalidateVisibility() {
+      var lastValue = scope.isOpen;
+      if (attr.mdIsLockedOpen) return;
+
+      if ($mdUtil.isHidden(element, true, true) && scope.isOpen) {
+        scope.isOpen = false;
+      } else if (!$mdUtil.isHidden(element, true, true) && !scope.isOpen) {
+        scope.isOpen = true;
+      }
+
+      // If the revalidated isOpen Value got changed,
+      // we should apply it to the view without running the updateOpen watcher
+      if (lastValue != scope.isOpen) {
+        skipSidenav = true;
+        if (!scope.$$phase) scope.$apply();
+        skipSidenav = false;
+      }
+    }
 
     /**
      * Toggle the DOM classes to indicate `locked`
@@ -278,31 +307,56 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
      * Toggle the SideNav view and attach/detach listeners
      * @param isOpen
      */
-    function updateIsOpen(isOpen) {
+    function updateIsOpen(isOpen, oldValue) {
+      if (skipNextUpdate) {
+        skipNextUpdate = false;
+        return;
+      }
+
       // Support deprecated md-sidenav-focus attribute as fallback
       var focusEl = $mdUtil.findFocusTarget(element) || $mdUtil.findFocusTarget(element,'[md-sidenav-focus]') || element;
       var parent = element.parent();
 
+      // Temporary remove md-closed class (won't affect the view, because it's outside of the viewport).
+      // And then check the visbility due hide attributes. So if the isOpen variable changes
+      // and a hide attribute is active we should revert the digest change
+      if (!skipSidenav) {
+        var wasClosed = element.hasClass('md-closed');
+
+        element.removeClass('md-closed');
+        var isHidden = $mdUtil.isHidden(element, true, false);
+        element.toggleClass('md-closed', wasClosed);
+
+        if (isHidden && !skipSidenav) {
+          element.toggleClass('md-closed', wasClosed);
+          skipNextUpdate = true;
+          scope.isOpen = !!oldValue;
+          return;
+        }
+      }
+
       parent[isOpen ? 'on' : 'off']('keydown', onKeyDown);
       backdrop[isOpen ? 'on' : 'off']('click', close);
 
-      if ( isOpen ) {
+      if ( isOpen && !skipSidenav) {
         // Capture upon opening..
         triggeringElement = $document[0].activeElement;
       }
 
       disableParentScroll(isOpen);
 
-      return promise = $q.all([
-                isOpen ? $animate.enter(backdrop, parent) : $animate.leave(backdrop),
-                $animate[isOpen ? 'removeClass' : 'addClass'](element, 'md-closed')
-              ])
-              .then(function() {
-                // Perform focus when animations are ALL done...
-                if (scope.isOpen) {
-                  focusEl && focusEl.focus();
-                }
-              });
+      var actions = [isOpen ? $animate.enter(backdrop, parent) : $animate.leave(backdrop)];
+      if (!skipSidenav) {
+        actions.push(isOpen ? $animate.removeClass(element, 'md-closed') : $animate.addClass(element, 'md-closed'));
+      }
+
+      return promise = $q.all(actions)
+        .then(function() {
+          // Perform focus when animations are ALL done...
+          if (scope.isOpen) {
+            focusEl && focusEl.focus();
+          }
+        });
     }
 
     /**
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -731,6 +731,17 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
       return current;
     },
 
+    /**
+     * Checks a elements for its hidden state.
+     * Using computed style or checking for existing in the visible DOM.
+     * @param computedStyle Check Computed Style
+     * @param offsetParent Check Offset Parent
+     */
+    isHidden: function(element, computedStyle, offsetParent) {
+      return computedStyle && window.getComputedStyle(element[0]).display === 'none' ||
+        offsetParent && element[0].offsetParent == null;
+    },
+
     hasComputedStyle: hasComputedStyle
   };
 
--- a/src/core/util/util.spec.js
+++ b/src/core/util/util.spec.js
@@ -274,6 +274,41 @@ describe('util', function() {
       }));
     });
 
+    describe('isHidden', function() {
+
+      it('should return true if hidden through style', inject(function($mdUtil) {
+        var element = angular.element('<span>isHidden</span>');
+        element.css('display', 'none');
+        document.body.appendChild(element[0]);
+
+        expect($mdUtil.isHidden(element, true, false)).toBe(true);
+
+        element.remove();
+      }));
+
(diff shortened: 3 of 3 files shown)


src/components/sidenav/sidenav.js
```diff
@@ -257,9 +266,30 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
 
 
+
     // Publish special accessor for the Controller instance
     sidenavCtrl.$toggleOpen = toggleOpen;
 
+    function revalidateVisibility() {
+      var lastValue = scope.isOpen;
+      if (isHidden(element, true, true) && scope.isOpen == true) scope.isOpen = false;
+      else if (!isHidden(element, true, true) && scope.isOpen == false) scope.isOpen = true;
```
EladBezalel: Brackets, `scope.isOpen == true` is the same as `scope.isOpen` and ternary operator

src/components/sidenav/sidenav.js
```diff
@@ -257,9 +266,30 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+
+      if (lastValue != scope.isOpen) {
+        skipSidenav = true;
+        if (!scope.$$phase) scope.$apply();
+        skipSidenav = false;
```
EladBezalel: Please write a comment on what you did here

src/components/sidenav/sidenav.js
```diff
@@ -257,9 +266,30 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+      }
+    }
+
+    function isHidden(element, compute, both) {
```
EladBezalel: Compute in your uses is always true

src/components/sidenav/sidenav.js
```diff
@@ -257,9 +266,30 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+      if (!compute || both) {
```
EladBezalel: what does both means?

src/components/sidenav/sidenav.js
```diff
@@ -278,26 +308,43 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+        skipNextUpdate = false;
+        return;
+      }
+
       // Support deprecated md-sidenav-focus attribute as fallback
       var focusEl = $mdUtil.findFocusTarget(element) || $mdUtil.findFocusTarget(element,'[md-sidenav-focus]') || element;
       var parent = element.parent();
 
+      // little hack to check show abbility
+      if (!skipSidenav) {
```
EladBezalel: Please explain

src/components/sidenav/sidenav.js
DevVersion: In my new version, I renamed to `compute`, it should just allow to check the `offsetParent` and even the `computed style`. I need this, because I check the visibility due the `hide` attributes.

src/components/sidenav/sidenav.js
DevVersion: If we not skip the scope.isOpen update, I remove the `md-closed` class temporary (won't be affect the view, because it's outside of the viewport) and then I can check for a `hide` attribute.

DevVersion:
@EladBezalel - Updated the PR

src/components/sidenav/sidenav.js
EladBezalel: Please change the comment to this explanation

DevVersion:
@EladBezalel - Completed all your suggestions, thanks for reviewing!

src/components/sidenav/sidenav.js
```diff
@@ -229,6 +229,9 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
     var lastParentOverFlow;
     var triggeringElement = null;
     var promise = $q.when(true);
+    var skipSidenav = false;
+    var skipNextUpdate = false;
+    var $window = angular.element(window);
```
EladBezalel: You can use DI to get [$window ](https://docs.angularjs.org/api/ng/service/$window)

src/components/sidenav/sidenav.js
```diff
@@ -257,9 +266,36 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+      } else if (!isHidden(element, true) && !scope.isOpen) {
+        scope.isOpen = true;
+      }
+
+      // If the revalidated isOpen Value got changed,
+      // we should apply it to the view without running the updateOpen watcher
+      if (lastValue != scope.isOpen) {
+        skipSidenav = true;
+        if (!scope.$$phase) scope.$apply();
+        skipSidenav = false;
```
EladBezalel: $timeout won't do the same?

src/components/sidenav/sidenav.js
```diff
@@ -278,26 +314,45 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
 
-      return promise = $q.all([
-                isOpen ? $animate.enter(backdrop, parent) : $animate.leave(backdrop),
-                $animate[isOpen ? 'removeClass' : 'addClass'](element, 'md-closed')
-              ])
-              .then(function() {
+      var actions = [isOpen ? $animate.enter(backdrop, parent) : $animate.leave(backdrop)];
+      if (!skipSidenav) actions.push(isOpen ? $animate.removeClass(element, 'md-closed') : $animate.addClass(element, 'md-closed'));
+
+      return promise = $q.all(actions).then(function() {
```
EladBezalel: please break line before `.then` :pray:

src/components/sidenav/sidenav.js
DevVersion: That won't work because I want to set the `skipSidenav` after running `$apply`, but the normal $timeout `invokeApply` will run the `$apply` after the given function.

See here $timeout snippet
```
        try {
          deferred.resolve(fn.apply(null, args));
        } catch (e) {
          deferred.reject(e);
          $exceptionHandler(e);
        }
        finally {
          delete deferreds[promise.$$timeoutId];
        }

        if (!skipApply) $rootScope.$apply();
```

src/components/sidenav/sidenav.js
DevVersion: Will be better, yes!

DevVersion:
@EladBezalel  - Broke Line + Used Dependency Injection + Fixed Angular's Bootstrap Digest with (oldVal = undefined)

DevVersion:
@EladBezalel - Broke Line + Used Dependency Injection + Fixed Angular's Bootstrap Digest with (oldVal = undefined) - Still needs work?

DevVersion:
@EladBezalel - Fixed the suggestions + Improved `isHidden` Method (more clear and less code)

DevVersion:
@EladBezalel - Fixed the suggestions + Improved `isHidden` Method (more clear and less code) - Can you verify the new version?

src/components/sidenav/sidenav.js
```diff
@@ -257,9 +266,40 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+    }
+
+    /**
+     * Checks the sidenav's hide state
+     * @param computedStyle Check Computed Style
+     * @param offsetParent Check Offset Parent
+     */
+    function isHidden(computedStyle, offsetParent) {
+      if (computedStyle && window.getComputedStyle(element[0]).display === 'none') return true;
+      if (offsetParent && element[0].offsetParent == null) return true;
```
EladBezalel: why not doing 
``` js
if (computedStyle && window.getComputedStyle(element[0]).display === 'none' ||
   offsetParent && !element[0].offsetParent) {
  return true;
}
```

src/components/sidenav/sidenav.js
```diff
@@ -257,9 +266,40 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+      return false;
```
EladBezalel: or just:
```js
return computedStyle && window.getComputedStyle(element[0]).display === 'none' ||
       offsetParent && !element[0].offsetParent;
```

src/components/sidenav/sidenav.js
DevVersion: Yes that's also possible, but I thought we should have a clear variant, because this method is really important and needs to be easy to understand, but if you want I can simplify the code?

src/components/sidenav/sidenav.js
EladBezalel: @DevVersion i think this is how the function should look

DevVersion:
@EladBezalel Okay used your simplified method ;)

src/components/sidenav/sidenav.js
```diff
@@ -257,9 +266,39 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
 
+    function revalidateVisibility() {
+      var lastValue = scope.isOpen;
+      if (attr.mdIsLockedOpen) return;
+
+      if (isHidden(true, true) && scope.isOpen) {
+        scope.isOpen = false;
+      } else if (!isHidden(true, true) && !scope.isOpen) {
+        scope.isOpen = true;
+      }
```
EladBezalel: Isn't that equals to:
```js
scope.isOpen = !(isHidden(true, true) && scope.isOpen)
```

DevVersion: Yes that works too, and is much more professional as before.

EladBezalel: Sorry i couldn't understand, are we watching for `isOpen` changes?

DevVersion: Sorry for the late answer. Yes `updateIsOpen` (the actual method which triggers open and close animations) is called by watching `isOpen`.

src/components/sidenav/sidenav.js
```diff
@@ -257,9 +266,39 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+        skipSidenav = false;
+      }
+    }
+
+    /**
+     * Checks the sidenav's hide state
+     * @param computedStyle Check Computed Style
+     * @param offsetParent Check Offset Parent
+     */
+    function isHidden(computedStyle, offsetParent) {
```
EladBezalel: Is this function is generic enough to be on the util service?

DevVersion: Yes, definitely, should I add it?

EladBezalel: yes

DevVersion: Done :+1: Added to Util + added tests to validate the functionality.

src/components/sidenav/sidenav.js
```diff
@@ -278,31 +307,51 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
-                $animate[isOpen ? 'removeClass' : 'addClass'](element, 'md-closed')
-              ])
-              .then(function() {
-                // Perform focus when animations are ALL done...
-                if (scope.isOpen) {
-                  focusEl && focusEl.focus();
-                }
-              });
+      var actions = [isOpen ? $animate.enter(backdrop, parent) : $animate.leave(backdrop)];
+      if (!skipSidenav) actions.push(isOpen ? $animate.removeClass(element, 'md-closed') : $animate.addClass(element, 'md-closed'));
```
EladBezalel: Please break this line

DevVersion: Done :hand: I should definitely use that `conventions` (how I call it) in my next PR's so we would save much time :smile: - Already done in my newer PR's

src/components/sidenav/sidenav.js
```diff
@@ -278,31 +307,53 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
...
+      if (!skipSidenav) {
+        var wasClosed = element.hasClass('md-closed');
+        element.removeClass('md-closed');
+        if ($mdUtil.isHidden(element, true, false) && !skipSidenav) {
+          element.toggleClass('md-closed', wasClosed);
+          skipNextUpdate = true;
+          scope.isOpen = !!oldValue;
+          return;
+        }
+        element.toggleClass('md-closed', wasClosed);
```
EladBezalel: Can't you reuse that?

```js
element.removeClass('md-closed');
if ($mdUtil.isHidden(element, true, false) && !skipSidenav) {
  skipNextUpdate = true;
  scope.isOpen = !!oldValue;
}
element.toggleClass('md-closed', wasClosed);
```

src/core/util/util.js
```diff
@@ -731,6 +731,16 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
       return current;
     },
 
+    /**
+     * Checks the sidenav's hide state
```
EladBezalel: This is not a proper description

DevVersion: Will change.

src/components/sidenav/sidenav.js
DevVersion: No, won't work, because I need to revert the class before the scope watcher will be triggered.

src/core/util/util.js
DevVersion: @EladBezalel Done I'd really like to forget all that issues and create a new PR :smile: But in my newer PR's there aren't so much style improvements. Just learning the style you / the Team uses!

src/components/sidenav/sidenav.js
```js
element.removeClass('md-closed');
var isHidden = $mdUtil.isHidden(element, true, false);
element.toggleClass('md-closed', wasClosed);

if (isHidden && !skipSidenav) {
  skipNextUpdate = true;
  scope.isOpen = !!oldValue;
}
```
?

DevVersion: This still needs the `return`, but yes this will work, because we check before removing the class :+1:.

-> Updated

EladBezalel:
@ThomasBurleson please review

DevVersion:
@EladBezalel I think this PR is not really a high priority, not much developers are complaining about it. What do you think? Should we remove it from the deprecation progress?

ThomasBurleson:

ThomasBurleson:

PR #6493, opened 2015-12-29
fix(list): show item content separated to the button

### LINKED ISSUE #6488: Click on <md-list-item> with multiple <md-checkbox> not working in mozilla
Hello Team,

Facing issue in case of implementing multiple md-checkbox in md-list-item,

Tried solutions
1. used primary action to overcome click list-item issue
2. Can't use secondary action on multi checkbox, else only one appears
3. screenshot attached for given source code, works in chrome but fails in mozilla

	<md-list-item ng-repeat="item in allContacts" ng-click="contactClick()">
		{{ item.name }}
		<p>{{ item.name }}</p>
		<md-checkbox ng-model="item.read" class="md-primary" aria-label="Read">Read</md-checkbox>
		<md-checkbox ng-model="item.write" class="md-primary" aria-label="Write">Write</md-checkbox>
	</md-list-item>

material list action issue

### LINKED ISSUE #6984: list: secondary items are completely disconnected from the button
on HEAD, seems to be introduced in 3ffa0a2b9d45e648d9d50df679d46310519995b0

DevVersion:
Fixes #6984 Fixes #6488

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -25,6 +25,7 @@ angular
  * the `md-primary` class.
  *
  * @param {boolean=} md-no-ink If present, disable ripple ink effects.
+ * @param {boolean=} md-no-focus-style If present, disable focus style on button
  * @param {expression=} ng-disabled En/Disable based on the expression
  * @param {string=} md-ripple-size Overrides the default ripple size logic. Options: `full`, `partial`, `auto`
  * @param {string=} aria-label Adds alternative text to button for accessibility, useful for icon buttons.
@@ -108,9 +109,10 @@ function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
       }
     });
 
-    // restrict focus styles to the keyboard
-    scope.mouseActive = false;
-    element.on('mousedown', function() {
+    if (!angular.isDefined(attr.mdNoFocusStyle)) {
+      // restrict focus styles to the keyboard
+      scope.mouseActive = false;
+      element.on('mousedown', function() {
         scope.mouseActive = true;
         $timeout(function(){
           scope.mouseActive = false;
@@ -124,6 +126,7 @@ function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
       .on('blur', function(ev) {
         element.removeClass('md-focused');
       });
+    }
   }
 
 }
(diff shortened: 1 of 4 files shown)


src/components/button/button-theme.scss
```diff
@@ -5,7 +5,7 @@ a.md-button.md-THEME_NAME-theme,
     &:hover {
       background-color: '{{background-500-0.2}}';
     }
-    &.md-focused {
+    &.md-focused:not(.md-no-focus) {
```
EladBezalel: it would be so much clearer if you would just wrap all with 
``` css
&:not(.md-no-focus) 
```

EladBezalel:
@DevVersion can you address me what should i do to test this manually ?

EladBezalel:
@DevVersion in the issue the user claims he can't use 2 checkboxes what is your fix right here?

DevVersion:
I completely reworked the list item structure, in this version we split the button and the content, thats really important because firefox is not allowing by default click events on children. This fix works as same as before, but it only fixes the issues with firefox and other browsers which disallow click events on child nodes

DevVersion:
As you want receive more info, what info is missing?. You can see the Firefox Bug Report on Bugzilla too, https://bugzilla.mozilla.org/show_bug.cgi?id=1181763

src/components/button/button-theme.scss
DevVersion: I'm not getting what you exactly mean. You mean we should wrap that as followed:
```
    &.md-focused {
      &:not(md-no-focus) {
        background-color: '{{background-500-0.2}}';

      }
    }
```

Because I think this will add an empty selector, which is not used. (`&.md-focused`)

EladBezalel: NVM but i just feel like this is wrong, can't we apply `md-focused` only when `md-no-focus` is not available?

DevVersion: Seems to be a better a idea, I would prefer a attribute called `md-no-focus-style`. See my newest changes, I updated my PR, you should decide about `className` or `attribute`

EladBezalel:
The code looks much clearer right now.

DevVersion:
You still want to get more information? because of the label?

AnirudhaGohokar:
Hi,
Was applying md-seconday to md-input-container but it does'nt seem to work. 

http://codepen.io/anon/pen/pgmwBv

I believe its for only buttons but what if we want to have non clickable contents like input control ?

Thank You

DevVersion:
Hey @AnirudhaGohokar,

That's not related to this PR, that's related to the new `multiple` secondary items feature, there's a little bug, which got fixed in #7060

DevVersion:
@AnirudhaGohokar,

Yes you're right, as the docs say, it's only restricted for use with buttons. If you would suggest this a feature, you should create another issue.

PR #6494, opened 2015-12-29
fix(toast): strip comments from template to prevent error throwing

### LINKED ISSUE #6259: $mdToast throws exception if custom template starts with a comment.
TODO(jelbourn): fill out details here.
http://jsbin.com/sesahiyere/edit?html,js,output

DevVersion:
Fixes #6259

### DIFF
--- a/src/components/toast/toast.js
+++ b/src/components/toast/toast.js
@@ -289,18 +289,22 @@ function MdToastProvider($$interimElementProvider) {
           // Root element of template will be <md-toast>. We need to wrap all of its content inside of
           // of <div class="md-toast-content">. All templates provided here should be static, developer-controlled
           // content (meaning we're not attempting to guard against XSS).
-          var parsedTemplate = angular.element(template);
-          var wrappedContent = '<div class="md-toast-content">' + parsedTemplate.html() + '</div>';
+          var templateRoot = document.createElement('md-template');
+          templateRoot.innerHTML = template;
 
-          parsedTemplate.empty().append(wrappedContent);
+          for (var i = 0; i < templateRoot.children.length; i++) {
+            if (templateRoot.children[i].nodeName === 'MD-TOAST') {
+              var wrapper = angular.element('<div class="md-toast-content">');
+              wrapper.append(templateRoot.children[i].children);
+              templateRoot.children[i].appendChild(wrapper[0]);
+            }
+          }
 
-          // Underlying interimElement expects a template string.
-          return parsedTemplate[0].outerHTML;
+
+          return templateRoot.outerHTML;
         }
 
-        return shouldAddWrapper ?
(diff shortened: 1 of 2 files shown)


EladBezalel:
LGTM. @jelbourn thoughts?

src/components/toast/toast.js
```diff
@@ -289,18 +289,16 @@ function MdToastProvider($$interimElementProvider) {
           // Root element of template will be <md-toast>. We need to wrap all of its content inside of
           // of <div class="md-toast-content">. All templates provided here should be static, developer-controlled
           // content (meaning we're not attempting to guard against XSS).
-          var parsedTemplate = angular.element(template);
-          var wrappedContent = '<div class="md-toast-content">' + parsedTemplate.html() + '</div>';
+          var parsedTemplate = angular.element(template.replace(/<!--(.*?)-->/g, ''));
```
jelbourn: This doesn't work because Angular actually lets you apply directives as comments.

https://docs.angularjs.org/guide/directive

DevVersion:
@jelbourn - But to get our actual target we need to get a element, which allows children. Can you please take a look at my new version?

src/components/toast/toast.js
```diff
@@ -290,17 +290,23 @@ function MdToastProvider($$interimElementProvider) {
           // of <div class="md-toast-content">. All templates provided here should be static, developer-controlled
           // content (meaning we're not attempting to guard against XSS).
           var parsedTemplate = angular.element(template);
-          var wrappedContent = '<div class="md-toast-content">' + parsedTemplate.html() + '</div>';
 
-          parsedTemplate.empty().append(wrappedContent);
+          for (var i = 0; i < parsedTemplate.length; i++) {
+            if (parsedTemplate[i].nodeType != Node.COMMENT_NODE) {
+              parsedTemplate = parsedTemplate.eq(i);
```
jelbourn: Wouldn't this not work for a template like this?
```html
<span>
  An element
</span>
<span>
  A sibling element
</span>
```

DevVersion:
@jelbourn - I need one decision from you. What should we do if we got two silbings as roots. Should we pick the first node, which allows to have childs. Or should we always use `md-toast`

jelbourn:
I think a much more straightforward approach would be to change `parsedTemplate` to a `<template>` element (using `document.createElement`) and set its innerHTML instead of using `angular.element`.

DevVersion:
@jelbourn - Now it works perfectly. It allows comments, it allows sibling elements, it allows normal `md-toast` templates. It removes all the `toasts` correctly from the DOM - PS: `<template>` is not good, it is reserved for modern browsers as a directive

jelbourn:
LGTM

PR #6496, opened 2015-12-29
fix(list): fix list margin if anchor is present.

### LINKED ISSUE #5608: Ui-sref on list item breaks design
When using ui-sref instead of ng-click od list item, design is broken.

### LINKED ISSUE #6317: md-list-item with href or ng-click have different margin
`md-list-item` with href have external margin value.
image
`md-list-item` with `ng-click` will create a `button` element.
That element with `md-button.md-no-style` to remove margin value.
image
But it's with `href` doesn't remove it.

I suppose add a css called `a.md-no-style` to remove margin value inherit from `md-button`

DevVersion:
If there is currently a `md-button` anchor, we should remove the margin to order the items as the other items.

Fixes #5608 Fixes #6317

### DIFF
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -61,6 +61,7 @@ md-list-item {
       flex-direction: inherit;
       align-items: inherit;
       border-radius: 0;
+      margin: 0;
 
       & > .md-ripple-container {
         border-radius: 0;


EladBezalel:
LGTM.

EladBezalel:
@DevVersion can you address me what should i do to test this manually ?

DevVersion:
I don't think there is something to test, you should just take a look to the list-items, one applied with ng-href and one applied with ng-click

PR #6518, opened 2016-01-02
feat(input): add asterisk to input directive

### LINKED ISSUE #6511: Required inputs as spec
Looking at the latest update to the Angular Material spec ([Dec. 2015](https://www.google.com/design/spec/whats-new/whats-new.html)), I noticed an update about [Required Text Fields](https://www.google.com/design/spec/components/text-fields.html#text-fields-search-filter):

> To indicate that a field is required, display an asterisk (*) next to the field. At the bottom of the form, include a note explaining that an asterisk indicates a required field.
> 
> Helper text, or hint text, should appear below the text field and either be persistently visible, or only visible on focus.

Mind you, the given video example shows the asterisk at the end of the field label (while the spec above talks about placing it `next to the field`, but regardless...

Due to the timing of the spec update, I'm not surprised the v.1.0.0, 1.0.1 and HEAD are not yet matching this detail, but I wonder if it could be an enhancement for future version.

DevVersion:
Fixes #6511

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -135,6 +135,7 @@ function labelDirective() {
  * @param {string=} placeholder An alternative approach to using aria-label when the label is not
  *   PRESENT. The placeholder text is copied to the aria-label attribute.
  * @param md-no-autogrow {boolean=} When present, textareas will not grow automatically.
+ * @param md-no-asterisk {boolean=} When present, asterisk will not be appended to required inputs label
  * @param md-detect-hidden {boolean=} When present, textareas will be sized properly when they are
  *   revealed after being hidden. This is off by default for performance reasons because it
  *   guarantees a reflow every digest cycle.
@@ -235,6 +236,9 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
     var hasNgModel = !!ctrls[1];
     var ngModelCtrl = ctrls[1] || $mdUtil.fakeNgModel();
     var isReadonly = angular.isDefined(attr.readonly);
+    var isRequired = angular.isDefined(attr.required);
+    var mdNoAsterisk = angular.isDefined(attr.mdNoAsterisk);
+
 
     if (!containerCtrl) return;
     if (containerCtrl.input) {
@@ -248,6 +252,8 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
 
     if (!containerCtrl.label) {
       $mdAria.expect(element, 'aria-label', element.attr('placeholder'));
+    } else if (isRequired && !mdNoAsterisk) {
+      containerCtrl.label.addClass('md-required');
     }
 
     element.addClass('md-input');
--- a/src/components/input/input.scss
+++ b/src/components/input/input.scss
@@ -91,6 +91,12 @@ md-input-container {
     bottom: 100%;
     @include rtl(left, 0, auto);
     @include rtl(right, auto, 0);
+
+    &.md-required:after {
+      content: ' *';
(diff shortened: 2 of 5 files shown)


EladBezalel:
LGTM

src/components/input/input.js
```diff
@@ -135,6 +135,7 @@ function labelDirective() {
  * @param {string=} placeholder An alternative approach to using aria-label when the label is not
  *   PRESENT. The placeholder text is copied to the aria-label attribute.
  * @param md-no-autogrow {boolean=} When present, textareas will not grow automatically.
+ * @param md-no-asterisk {boolean=} Stops appending of an asterisk to the label of required inputs
```
EladBezalel: I'd change this to 
`When present, asterisk will not be appended to required inputs label`
or something like that

EladBezalel:
@DevVersion the asterisk size is a bit larger the the spec, can you try and make it smaller?

DevVersion:
@EladBezalel Okay, it's now smaller and still aligned at the top

EladBezalel:
looks better, the asterisk size shouldn't be changed but that's because the transform so fixing it will be over-kill.

Fix other comments and please add/change/ to the demo 1 input to have `md-no-asterisk`

DevVersion:
@EladBezalel - Done

Emeegeemee:
From looking at the [spec](https://www.google.com/design/spec/components/text-fields.html#text-fields-required-fields) for required fields, does this handle the label text becoming the warning color on blur?

src/components/input/input.scss
```diff
@@ -91,6 +91,12 @@ md-input-container {
     bottom: 100%;
     @include rtl(left, 0, auto);
     @include rtl(right, auto, 0);
+
+    &.md-required:after {
+      content: ' *';
```
Emeegeemee: Does the content reverse for rtl text?

DevVersion:
@Emeegeemee - Yes it does -

src/components/input/input.scss
DevVersion: Asterisk will be aligned in the same direction as the other text, so it's compatible with RTL, LTR.
But it seems that the placeholders are aligned wrong? (See [here](http://codepen.io/DevVersion/pen/zraQxj).

But in general, Yes asterisk will be reversed for RTL text

Emeegeemee:
@DevVersion From your screenshot that looks like on focus not on blur. Please watch the demo carefully, when the focus is removed from the input in an invalid state the entire label goes red it doesn't go back to the black/grey color.

EladBezalel:
@Emeegeemee you can see it for yourself on [input demo page](https://material.angularjs.org/HEAD/demo/input)

Emeegeemee:
@EladBezalel Thanks, from the site it looks to not make the label change red when focus is lost.

from the spec: 
screenshot from 2016-02-01 14-40-35

current implementation:
screenshot from 2016-02-01 14-41-57

EladBezalel:
@Emeegeemee #6964

src/components/input/input.scss
topherfangio: @DevVersion I know this is old, but I thought I would update this since I just looked at it. The labels are properly "aligned" to the right, it's just that the width is not 100% of the parent. I guess we need to use `right: 0` in addition to `text-align: right` here to fix?

topherfangio: Ah, just looked a bit more; it's actually an issue with the `scale(0.75)` transform.

Looks like we can fix it with `transform-origin: right;`.

PR #6521, opened 2016-01-03
fix(autocomplete): improve promise logic

winniehell:
The `finally`-function supported by [q] is not part of the [A+ spec]. Therefore, other promise implementations do not have it (e.g. the [promise polyfill] in [babel]). To ensure that `md-autocomplete` resets its loading state properly, I used `then` if `finally` is not available.

This pull request was joined with #6860.

[A+ spec]: https://promisesaplus.com/
[babel]: https://github.com/babel/babel/
[promise polyfill]: https://github.com/zloirock/core-js/blob/master/modules/es6.promise.js
[q]: https://github.com/kriskowal/q/

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -641,9 +641,15 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       $mdUtil.nextTick(function () {
         if (items.success) items.success(handleResults);
         if (items.then)    items.then(handleResults);
-        if (items.finally) items.finally(function () {
-          setLoading(false);
-        });
+        if (items.finally) {
+          items.finally(function () {
+            setLoading(false);
+          });
+        } else if (items.then) {
+          items.then(function () {
+            setLoading(false);
+          });
+        }
       },true, $scope);
     }
     function handleResults (matches) {


winniehell:
@googlebot I signed it!

winniehell:
Failing tests seem to be unrelated as they appear e.g. in #6515 ([build 99782625]) and #6518 ([build 99851677]) as well. Not sure what I can do about it... :alien::question:

[build 99782625]: https://travis-ci.org/angular/material/builds/99782625
[build 99851677]: https://travis-ci.org/angular/material/builds/99851677

DevVersion:
The tests of the whole project are currently failing. Thats not an issue from you ;)

winniehell:
@DevVersion you are right that it is usual to use `$q` from angular in that context&mdash;but it is not necessary to do so. In my opinion, `md-autocomplete` should work with other promise implementations as well (and besides this little issue, it seems to me that it actually does :+1:).

There are two reasons for me not to use `$q` from angular:

1. it is common to use the [Deferred API] which is not part of the [A+ spec] and deprecated for other implementations (e.g. [Gecko])
1. when combining angular with another framework (e.g. [meteor]), always wrapping existing promises with `$q` is cumbersome :cactus: 

[A+ spec]: https://promisesaplus.com/
[Deferred API]: https://docs.angularjs.org/api/ng/service/$q#the-deferred-api
[Gecko]: https://developer.mozilla.org/en-US/docs/Mozilla/JavaScript_code_modules/Promise.jsm/Deferred
[meteor]: http://www.angular-meteor.com/

DevVersion:
Can be true, I don't know about A+ etc. But in this context of code we using DI to get the $q service of Angular, so there won't be any other Deferred API, that's why I think this change is not needed.

winniehell:
Even though `$q` is injected, [items][controller-items] is [evaluated from][controller-itemExpr] the [expression][directive-itemsExpr] which is [passed to][directive-md-items] the directive. Therefore it can basically be any expression. However, `md-autocomplete` only works with array expressions and thenables. Currently, it is restricted to thenables with a `finally`-function. In my opinion, this restriction is not necessary.

[controller-itemExpr]: https://github.com/angular/material/blob/c667b5499dfd51adb47852a69d62cc5bbd19b24d/src/components/autocomplete/js/autocompleteController.js#L14-L15
[directive-md-items]: https://github.com/angular/material/blob/c667b5499dfd51adb47852a69d62cc5bbd19b24d/src/components/autocomplete/js/autocompleteDirective.js#L36-L37
[controller-items]: https://github.com/angular/material/blob/c667b5499dfd51adb47852a69d62cc5bbd19b24d/src/components/autocomplete/js/autocompleteController.js#L635
[directive-itemsExpr]: https://github.com/angular/material/blob/c667b5499dfd51adb47852a69d62cc5bbd19b24d/src/components/autocomplete/js/autocompleteDirective.js#L132

winniehell:
For clarification: the effect on `md-autocomplete` for thenables which do not have a `finally`-function is that `setLoading(false)` is never called when no matching item can be found. As a result instead of showing `md-not-found` the loading indicator remains visible.

winniehell:
Tests are now all green. :shipit:?

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -641,9 +641,15 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       $mdUtil.nextTick(function () {
```
ThomasBurleson: Why not use the `$q.when( )` to wrap angular & foreign promises:

```js
function fetchResults (searchText) {
  var items = $scope.$parent.$eval(itemExpr),
      term  = searchText.toLowerCase(),
      isList = angular.isArray(items);

  if ( isList ) handleResults(items);
  else          handleAsyncResults(items);

  function handleAsyncResults(items) {
    if ( !items ) return;

    items = $q.when(items);
    setLoading(true);

        items.then(function(){ 
          setLoading(false); 
        });
    },true, $scope);
  }
}
```

ThomasBurleson:
@robertmesserle - can you review?

src/components/autocomplete/js/autocompleteController.js
ThomasBurleson: PR #6860 also suggest not listening for `.success( )`.

robertmesserle: @ThomasBurleson Probably a good point, but this is a complaint on the original code, not this PR.

robertmesserle:
@ThomasBurleson LGTM

src/components/autocomplete/js/autocompleteController.js
ThomasBurleson: @winniehell - please incorporate my suggestions ^^ and test. This will make the merged code better moving forward.

winniehell:
@ThomasBurleson: I'll try to rework this including your comments and a fix for #6860.

winniehell:
@ThomasBurleson: Alright, I was still trying to figure out, how to write a test for it. Thank you for fixing this! :+1: :smiley:

Issue #4476, opened 2015-09-04
md-chips requires secondary-placeholder to use placeholder


Emeegeemee:
When using the md-chips directive and setting a placeholder it won't be used unless a secondary-placeholder is set. I would expect that in the secondary-placeholder would just be set to the same value as placeholder if it isn't specified.

Here's a codepen: http://codepen.io/anon/pen/OyPwjW

ericgundrum:
Likely this is an artifact of issue #2770, _secondary and primary placeholders switched in md-chips_. The code is written to prefer the secondary placeholder over the primary. The fix is trivial. (I documented it in #2770.)

ericgundrum:
The pull request @dbwhddn10 associated with #3804 should fix this issue.

ThomasBurleson:
This issue is closed as part of our deprecation effort.
    For details, see our post [Spring Cleaning 2016](https://groups.google.com/forum/#!topic/ngMaterial/3Ip9nd68bfk).

PR #6535, opened 2016-01-04
fix(md-chips placeholder): correct placeholder/secondary logic

ericgundrum:
Test and then fix the md-chips placeholder logic so that the secondary placeholder text appears only when it is documented to appear. Addresses issues #2770 and #4476.

The first commit provides test code to demonstrate the failure. The second commit changes the implementation code to pass the tests.

### DIFF
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -183,7 +183,7 @@ MdChipsCtrl.prototype.getPlaceholder = function() {
   // Allow `secondary-placeholder` to be blank.
   var useSecondary = (this.items.length &&
       (this.secondaryPlaceholder == '' || this.secondaryPlaceholder));
-  return useSecondary ? this.placeholder : this.secondaryPlaceholder;
+  return useSecondary ? this.secondaryPlaceholder : this.placeholder;
 };
 
 /**
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -362,6 +362,48 @@ describe('<md-chips>', function() {
         expect(element.find('md-chips-wrap').hasClass('md-focused')).toBe(false);
       }));
 
+      describe('placeholder', function() {
+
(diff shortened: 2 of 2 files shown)


DevVersion:
Please squash your commits, so it can be cherry picked easily.

ericgundrum:
@DevVersion, are you sure that is preferable? My point with two commits was for one to demonstrate the failure of the existing code and then a second to show a fix. By keeping them separate either one can be taken independent of the other -- they do not depend upon each other.

DevVersion:
As contributors we should always squash our commits to one commit, so Thomas Burleson, the main PR Reviewer, can easily merge the PR. But you can wait with Squashing until the PR got reviewed.

topherfangio:
@ericgundrum Thanks for the PR. LGTM. In general, we do request that you squash all commits, and Thomas may ask that before he merges, but sometimes if there are very few commits, he will squash them himself. It just saves him a step as we always squash before merging.

src/components/chips/js/chipsController.js
```diff
@@ -183,7 +183,7 @@ MdChipsCtrl.prototype.getPlaceholder = function() {
   // Allow `secondary-placeholder` to be blank.
   var useSecondary = (this.items.length &&
       (this.secondaryPlaceholder == '' || this.secondaryPlaceholder));
-  return useSecondary ? this.placeholder : this.secondaryPlaceholder;
+  return useSecondary ? this.secondaryPlaceholder : this.placeholder;
```
ThomasBurleson: @topherfangio - this logic looks wrong. Can you check.

topherfangio: Nope, our existing logic has been wrong the whole time. It currently says

    if use secondary
        use regular placeholder
    else
        use secondary placeholder

which is totally backwards. This fixes it. See #2770 for Tyler's comment about it being backwards.

PR #6051, opened 2015-12-03
WIP: Adding vertical sliders

ceoaliongroo:
#1221

Issue #4576, opened 2015-09-11
Sliders: focus and state improvements


ThomasBurleson:
MDL implements correct size of normal and click states:  http://www.getmdl.io/components/index.html#sliders-section

- sliders do not need hover.
- sliders only have ripple ring on tab focus
- on release, slider should go back to default static state.
- input field on the right could be material inputs

![screen shot 2015-09-11 at 8 51 59 am](https://cloud.githubusercontent.com/assets/210413/9816399/5ffed570-5862-11e5-8bcb-f61d1949eb35.png)

EladBezalel:
@ThomasBurleson by the spec slider do need hover,
The initial thumb size is 12px, on hover it's 14px

ThomasBurleson:
Fixed with SHA e0abeb4dc46a1753616a911fe54f14d8f1148a25

PR #6538, opened 2016-01-04
feat(slider): vertical slider, UI fixes and bug fixes

### LINKED ISSUE #4371: md-slider rendering issue in the case of dynamic min/max

EladBezalel:
- Added vertical slider
- Improved UI
- Fixed dynamic min/max with values not in range
- Added round attribute to set how many numbers should be after the dot, default is 3, maximum is 6 to prevent scientific notation.
- Made input width grow or shrink according the text length
- Added discrete readonly mode
- Added disabled on slider container

fixes #4371, #3259, #2421, #1221, #4576, #6996, #7093, #7093
closes #5874, #5872, #6051

### DIFF
--- a/src/components/slider/slider.js
+++ b/src/components/slider/slider.js
@@ -5,7 +5,84 @@
   angular.module('material.components.slider', [
     'material.core'
   ])
-  .directive('mdSlider', SliderDirective);
+  .directive('mdSlider', SliderDirective)
+  .directive('mdSliderContainer', SliderContainerDirective);
+
+/**
+ * @ngdoc directive
+ * @name mdSliderContainer
+ * @module material.components.slider
+ * @restrict E
+ * @description
+ * The `<md-slider-container>` contains slider with two other elements.
+ *
+ *
+ * @usage
+ * <h4>Normal Mode</h4>
+ * <hljs lang="html">
+ * </hljs>
+ */
+function SliderContainerDirective() {
+  return {
+    controller: function () {},
+    compile: function (elem) {
+      var slider = elem.find('md-slider');
+
+      if (!slider) {
+        return;
+      }
+
+      var vertical = slider.attr('md-vertical');
+
+      if (vertical !== undefined) {
+        elem.attr('md-vertical', '');
+      }
+
+      if(!slider.attr('flex')) {
+        slider.attr('flex', '');
+      }
+
+      return function (scope, element, attr, ctrl) {
+
+        // We have to manually stop the $watch on ngDisabled because it exists
+        // on the parent scope, and won't be automatically destroyed when
+        // the component is destroyed.
+        function setDisable(isDisabled) {
+          element.children().attr('disabled', isDisabled);
+          element.find('input').attr('disabled', isDisabled);
+        }
+
+        var stopDisabledWatch = angular.noop;
+        if (attr.disabled) {
+          setDisable(true);
+        }
+        else if (attr.ngDisabled) {
+          stopDisabledWatch = scope.$watch(attr.ngDisabled, function (isDisabled) {
+            setDisable(isDisabled);
+          });
+        }
+
+        scope.$on('$destroy', function () {
+          stopDisabledWatch();
+        });
+
+        var initialMaxWidth;
+
+        ctrl.fitInputWidthToTextLength = function (length) {
+          var input = element.find('md-input-container');
+          var computedStyle = getComputedStyle(input[0]);
+          var minWidth = parseInt(computedStyle['min-width']);
+          var padding = parseInt(computedStyle['padding']) * 2;
+          initialMaxWidth = initialMaxWidth || parseInt(computedStyle['max-width']);
+
+          var newMaxWidth = Math.max(initialMaxWidth, minWidth + padding + (minWidth / 2 * length));
+
+          input.css('max-width', newMaxWidth + 'px');
+        };
+      }
+    }
+  }
+}
 
 /**
  * @ngdoc directive
@@ -44,26 +121,29 @@
  * @param {number=} step The distance between values the user is allowed to pick. Default 1.
  * @param {number=} min The minimum value the user is allowed to pick. Default 0.
  * @param {number=} max The maximum value the user is allowed to pick. Default 100.
+ * @param {number=} round The amount of numbers after the decimal point, maximum is 6 to prevent scientific notation. Default 3.
  */
-function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdTheming, $mdGesture, $parse, $log) {
+function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdTheming, $mdGesture, $parse, $log, $timeout) {
   return {
     scope: {},
-    require: '?ngModel',
+    require: ['?ngModel', '?^mdSliderContainer'],
     template:
       '<div class="_md-slider-wrapper">' +
-        '<div class="_md-track-container">' +
-          '<div class="_md-track"></div>' +
-          '<div class="_md-track _md-track-fill"></div>' +
-          '<div class="_md-track-ticks"></div>' +
-        '</div>' +
-        '<div class="_md-thumb-container">' +
-          '<div class="_md-thumb"></div>' +
-          '<div class="_md-focus-thumb"></div>' +
-          '<div class="_md-focus-ring"></div>' +
-          '<div class="_md-sign">' +
-            '<span class="_md-thumb-text"></span>' +
+        '<div class="_md-slider-content">' +
+          '<div class="_md-track-container">' +
+            '<div class="_md-track"></div>' +
+            '<div class="_md-track _md-track-fill"></div>' +
+            '<div class="_md-track-ticks"></div>' +
           '</div>' +
-          '<div class="_md-disabled-thumb"></div>' +
+          '<div class="_md-thumb-container">' +
+            '<div class="_md-thumb"></div>' +
+            '<div class="_md-focus-thumb"></div>' +
+            '<div class="_md-focus-ring"></div>' +
+            '<div class="_md-sign">' +
+              '<span class="_md-thumb-text"></span>' +
+            '</div>' +
+            '<div class="_md-disabled-thumb"></div>' +
+          '</div>' +
         '</div>' +
       '</div>',
     compile: compile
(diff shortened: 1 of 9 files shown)


src/components/slider/slider.js
```diff
@@ -5,7 +5,70 @@
...
+ *
+ * @usage
+ * <h4>Normal Mode</h4>
+ * <hljs lang="html">
+ * </hljs>
+ */
+function SliderContainerDirective() {
+  return {
+    compile: function (elem) {
+      var slider = angular.element(elem[0].getElementsByTagName('md-slider'));
```
topherfangio: I believe you can replace this with

    var slider = elem.find('md-slider');

since you are finding an element, and not using a CSS query selector.

src/components/slider/slider.js
```diff
@@ -5,7 +5,70 @@
...
+      }
+
+      return {
+        post: function (scope, element, attr) {
+          // We have to manually stop the $watch on ngDisabled because it exists
+          // on the parent scope, and won't be automatically destroyed when
+          // the component is destroyed.
+          function setDisable(isDisabled) {
+            element.children().attr('disabled', isDisabled);
+            angular.element(element[0].getElementsByTagName('input')).attr('disabled', isDisabled);
```
topherfangio: Should also be able to do `element.find('input').attr..` here as well.

topherfangio: Should also be able to do `element.find('input').attr()` here as well.

src/components/slider/slider.scss
```diff
@@ -49,16 +74,22 @@ md-slider {
 
   height: $slider-height;
   position: relative;
-  display: block;
   margin-left: 4px;
   margin-right: 4px;
   padding: 0;
+  justify-content: center;
+  display: flex;
```
topherfangio: Could this be moved to a wrapper? I'm concerned users might attempt to make this block/inline and it would mess up your justification.

src/components/slider/slider.scss
```diff
@@ -101,14 +133,15 @@ md-slider {
...
+    transition: all .4s cubic-bezier(.25,.8,.25,1);
+    transition-property: left, bottom;
   }
   .md-thumb {
     z-index: 1;
 
     // Positioning the outer area of the thumb 6px bigger than it needs to be keeps
     // the :after area being clipped by the background of the focus-thumb animation.
-    @include slider-thumb-position($slider-thumb-width + 6, $slider-thumb-height + 6);
+    @include slider-thumb-position($slider-thumb-width, $slider-thumb-height);
```
topherfangio: Is the comment above this line relevant any longer?

topherfangio:
Aside from my line-comments above, I have seen a few visual issues when running the demos:

1. On the Vertical Sliders demo, before interacting with them, a very tiny number id displayed inside the slider's circle. Seems a bit distracting, and it disappears after you focus/touch it.

2. Upon focusing a slider via keyboard interaction, the slider is drastically larger than it should be (I think the bigger circle is simply 100% opacity instead of 38% or something)

3. Focusing the Rating sliders via keyboard makes the circle appear and then disappear whereas clicking  seems to have the proper functionality.

src/components/slider/slider.js
EladBezalel: you're right

src/components/slider/slider.js
EladBezalel: :+1:

src/components/slider/slider.scss
EladBezalel: Add an other wrapper?

topherfangio: No, I'm wondering if that display/justification can be put on one of the existing wrappers instead of the main element. By putting it on the element, you force the user to use flex with this component instead of allowing block or inline.

EladBezalel: I see, trying to solve it.

EladBezalel:
@topherfangio Talked to thomas about the
``` css
disaplay: flex,
justify-content: center
```

and he said it's ok but just add `flex-direction: row`

src/components/slider/slider.scss
ThomasBurleson: @EladBezalel - 

```html
<md-slider ....>
   <div class="md-slider-wrapper">
 
    ....
   </div>
</md-slider>
```

As @topherfangio  mentioned, you should make the mdSlider component `display: inline-block` so it can be easily used/integrated with other elements. And then use the Layout API for the `md-slider-wrapper` and its children.

screen shot 2016-02-06 at 7 08 02 am

src/components/slider/slider.js
```diff
@@ -237,9 +336,9 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
       }
 
       var changeAmount;
-      if (ev.keyCode === $mdConstant.KEY_CODE.LEFT_ARROW) {
+      if (vertical ? ev.keyCode === $mdConstant.KEY_CODE.DOWN_ARROW : ev.keyCode === $mdConstant.KEY_CODE.LEFT_ARROW) {
```
topherfangio: I'm curious if we should just always associated left/down with negative and right/up with positive; no matter what orientation it's in.

Thoughts?

src/components/slider/slider.js
```diff
@@ -254,6 +353,29 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
...
 
+    function mouseDownListener() {
+      redrawTicks();
+
+      scope.mouseActive = true;
+      wrapper.removeClass('md-focused');
+
+      $timeout(function() {
+        scope.mouseActive = false;
+      }, 100);
```
topherfangio: Why `100ms`? We generally try to avoid having a hard-coded timeout if possible. Maybe we could use `$mdUtil.nextTick()` instead?

src/components/slider/slider.js
```diff
@@ -280,8 +404,8 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
     function stepValidator(value) {
       if (angular.isNumber(value)) {
         var formattedValue = (Math.round((value - min) / step) * step + min);
-        // Format to 3 digits after the decimal point - fixes #2015.
-        return (Math.round(formattedValue * 1000) / 1000);
+        // Format to 6 digits after the decimal point.
```
topherfangio: I think I've seen this value changed like 3 times now. Perhaps we should make it configurable?

src/components/slider/slider.js
```diff
@@ -365,9 +488,7 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
      * @param x
      */
     function doSlide( x ) {
-      scope.$evalAsync( function() {
-        setModelValue( percentToValue( positionToPercent(x) ));
-      });
+      setModelValue( percentToValue( positionToPercent(x) ));
```
topherfangio: I am always weary to remove existing code; especially a `$evalAsync()`. Did this fix a bug/issue you were having?

src/components/slider/slider.scss
```diff
@@ -174,18 +207,11 @@ md-slider {
...
-    animation: sliderFocusThumb 0.4s linear;
-  }
   .md-focus-ring {
     @include slider-thumb-position($slider-focus-thumb-width, $slider-focus-thumb-height);
-    transform: scale(0);
-    transition: all 0.2s linear;
-    opacity: 0.26;
+    transform: scale(.7);
+    opacity: 0;
+    transition: all ($slider-thumb-focus-duration / 2) $swift-ease-in-out-timing-function;
```
topherfangio: When providing a custom duration, it helps to have a short comment as to why we don't use the default.

src/components/slider/slider.spec.js
```diff
@@ -23,13 +23,30 @@ describe('md-slider', function() {
       'getBoundingClientRect'
     ).and.returnValue(angular.extend({
       width: 100,
+      height: 100,
```
topherfangio: I am pretty sure you do not need width/height if you set the top/bottom/left/right all to `0`. Perhaps we can remove it?

src/components/slider/slider.spec.js
```diff
@@ -23,13 +23,30 @@ describe('md-slider', function() {
...
+      top: 0
     }, dimensions || {}));
 
     return slider;
   }
 
+  function setupContainer(attrs, sliderAttrs) {
+    var container;
+
+    container = $compile('<md-slider-container ' + (attrs || '') + '>' +
```
topherfangio: Can you refactor the HTML here into a `template` variable? With the embedded parentheses, it makes it a bit hard to read :smile:

src/components/slider/slider.js
DevVersion: I think the `100ms` timeout is the same as in the `button` component. This is the default timeout for detecting the mouse interaction. 

>    // We only *show* the focus style when focus has come to the button via the keyboard.
    // The Material Design spec is silent on this topic, and without doing this, the
    // button continues to look :active after clicking.
    // @see http://marcysutton.com/button-focus-hell/

EladBezalel: Wow this is weird! i usually don't do those kind of stuff cause i hate it!
great catch, thanks!

src/components/slider/slider.js
EladBezalel: Agree

src/components/slider/slider.spec.js
EladBezalel: This is for vertical proposes

src/components/slider/slider.spec.js
EladBezalel: lol your'e right.

src/components/slider/slider.js
EladBezalel: I can't really remember right now, i will revert it and see if anything has changed

src/components/slider/slider.scss
EladBezalel: basically it's supposed to be as close as possible to the spec, i will write it there

topherfangio:
I think this should be a 1.1.0 change as it modifies the structure of the the slider and makes a lot of changes that might potentially be breaking.

EladBezalel:
I don't think postponing this to 1.1.0 is smart, there're more additions than breaking changes.. I agree that this shouldn't be on the minor branch but on master for sure, more than that postponing it will make the rebasing even more hard for me.

src/components/slider/slider.js
EladBezalel: Mmm i think it's a valid thought but how would you suggest doing it?

src/components/slider/slider.js
EladBezalel: After reading @marcysutton post, i think it's necessary although it seems to be working with `$mdUtil.nextTick()`

topherfangio:
@Elad master is 1.1.0 :-)

EladBezalel:
ah :+1: didn't realized it lol

src/components/slider/slider.js
topherfangio: @EladBezalel Instead of checking Vertical/Horizontal in the line above, just always associate DOWN/LEFT with negative and UP/RIGHT with positive.

topherfangio:
@EladBezalel Just left one more comment re: the arrow keys, but I think that can be a separate discussion/PR if we want to do it later.

ThomasBurleson:
Fixed with SHA e0abeb4dc46a1753616a911fe54f14d8f1148a25

Issue #6552, opened 2016-01-06
(dialog): Default behavior of reopen dialog. 


Jacksunwei:
Currently if we reopen a dialog, the default behavior is $mdDialog.hide(), which resolves the previous promise. However, I think it makes more sense to call $mdDialog.cancel(), because the previous dialog is closed unexpectedly. 

A sample codepen is here: 
http://codepen.io/Jacksunwei/pen/VepWvW

ThomasBurleson:
API will not be changed.

PR #6570, opened 2016-01-07
fix(interimElement): default behavior of show interim element when there is already one. 

Jacksunwei:
I think it should be more natural to reject the previous promise (call cancel) instead of to resolve, because  in most case, the previous interim element isn't aware of the future unexpected interruption. 
I also open an issue(#6552), which is specific to $mdDialog. 
What do you think?


ThomasBurleson:
Will not change latest code in v1.1.0-rc4.

PR #6572, opened 2016-01-07
fix(docs): menu toggle is not open on the current page

danlevan:
Fix docs nav toggle to open the current page when refreshing the page or navigating directly to a section.


danlevan:
I signed it!

danlevan:
Bug description: When navigating to `https://material.angularjs.org/latest/demo/whiteframe` for example, the menu is not selected on the left (the toggle icon is correct, but the sub menu is not shown).

Wrapped getTargetHeight() in a $timeout in order to get the proper ul height (instead of 0).

PR #6589, opened 2016-01-07
fix(autocomplete): disable scroll events for autocomplete wrap layer

### LINKED ISSUE #5230: md-autocomplete + scroll
0.11.4

result box does not stick to element when window is scrolled.

untitled screencast

### LINKED ISSUE #5890: md-autocomplete, scrolling over input while menu opened, shifts dropdown list position
You could easily reproduce it on demo page.

### LINKED ISSUE #6585: md-scroll-mask might not take effect in time with md-autocomplete
There would be a small window when search in md-autocomplete and make a page scrolling that causes md-autocomplete suggestion box's position changed with scrolling. There is a screenshot:

screen shot 2016-01-07 at 1 30 28 pm

### LINKED ISSUE #6716: md-autocomplete - when dropdown list is open, attempting to scroll while hovering over the input, shifts the dropdown
Hello,

When hovering over the autocomplete input, attempting to scroll while the mouse is over the input shifts the dropdown list. This happens in Master branch and in version 1.0.2. This happens in Chrome, Firefox, Safari on MAC. It seems like it happens on all browsers.

Please let me know if you need more info.

Here's a snapshot of it:
autocomplete_bug

DevVersion:
At the moment we set the z-index of the `autocomplete-wrap` to `51`, that's why it is usable while an open scrollmask
So we only need to disable the scroll events to prevent scrolling when hovering over the `autocomplete-wrap`

Fixes #6585 Fixes #5230 Fixes #5890 Fixes #6716

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -19,7 +19,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       selectedItemWatchers = [],
       hasFocus             = false,
       lastCount            = 0,
-      fetchesInProgress    = 0;
+      fetchesInProgress    = 0,
+      enableWrapScroll     = null;
 
   //-- public variables with handlers
   defineProperty('hidden', handleHiddenChange, true);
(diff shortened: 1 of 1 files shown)


hitigon:
nice, it works, resolved #6585

EladBezalel:
@topherfangio do you mind assign it to you?

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -241,16 +242,36 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       if (elements) {
         $mdUtil.nextTick(function () {
           $mdUtil.disableScrollAround(elements.ul);
+          revertWrapScrollBlocking = disableElementScrollEvents(angular.element(elements.wrap));
         }, false, $scope);
       }
     } else if (hidden && !oldHidden) {
       $mdUtil.nextTick(function () {
         $mdUtil.enableScrolling();
+        if (revertWrapScrollBlocking) revertWrapScrollBlocking();
```
topherfangio: Shouldn't this also set `revertWrapScrollBlocking` to null so it doesn't potentially try to do it twice?

DevVersion: I don't that's necessary, because the `off` events are bound to an unique event function, but I will update it, because it seems to be logical.

topherfangio:
LGTM if it can be rebased :-)

PR #6611, opened 2016-01-10
perf(md-tooltip): do not run tooltip toggle logic if tooltip is already in that state

### LINKED ISSUE #6557: (perf) md-tooltip triggers digest cycles on scroll
In the latest (1.0.1) it seems that md-tooltip is very greedy about scroll event - basically any scroll on page is triggering digest cycle (no throttling etc. = a lot of digests!). At the moment we disabled tooltips because of that but in my opinion this needs to be addressed.

Here is a very simple codepen based on md-tooltip demo page.
http://codepen.io/anon/pen/rxyJaZ

I added small vanilla JS counter that is triggered whenever my watcher function is triggered. You can see that any scroll on the page creates whole lot of digests.
It does seem that tooltip is using runs "setVisible(false)"
https://github.com/angular/material/blob/master/src/components/tooltip/tooltip.js#L150

And it uses "$mdUtil.nextTick" to set "scope.visible = false;" so that it coalesces calls. Unfortunately it seems it is not enough for desktop browser scroll (this is tested on OSX) and processQueue is called multiple times per scroll. Maybe some kind of debounced function as windowScroll would be better here?

kwypchlo:
Do not run tooltip toggle logic if tooltip is already in that state.

ps. is there any reason why onscroll event is not throttled like the resize event?
```
ngWindow.on('resize', debouncedOnResize);
document.addEventListener('scroll', windowScrollHandler, true);
```

closes #6557

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -187,6 +187,9 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
     }
 
     function setVisible (value) {
+      // break if passed value is already in queue or there is no queue and passed value is current in the scope
+      if (setVisible.queued && setVisible.visible === !!value || scope.visible === !!value) return;
+      
       setVisible.value = !!value;
       if (!setVisible.queued) {
         if (value) {


src/components/tooltip/tooltip.js
```diff
@@ -187,6 +187,9 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
     }
 
     function setVisible (value) {
+      if ((setVisible.queued && setVisible.visible === !!value) || (!setVisible.queued && scope.visible === !!value)) {
```
ThomasBurleson: Is this not better logic?

```js
if (setVisible.visible === !!value) return;
```

kwypchlo: Nope because `setVisible.visible` can be still undefined and `scope.visible` is directive param that can be set from outside. There are 2 way this goes down:
1. `setVisible.queued` is set which means that someone already called `setVisible` and the change on `scope.visible` is pending. While `scope.visible` is not yet set correctly ($timeout not resolved), `setVisible.visible` is already set to the value that `scope.visible` will be after the $timeout resolves hence if `setVisible.visible === !!value` then we don't have to do nothing here.
2. `setVisible.queued` is not set, which means that `scope.visible` has the most current view value, hence we should compare the value passed to the function with `scope.value`

kwypchlo: @ThomasBurleson by the way, this perf issue made me disable tooltips in my company's project for the time being. I would be glad if we could push some kind of perf fix here as soon as possible.

EladBezalel:
LGTM

kwypchlo:
@ThomasBurleson @EladBezalel any chance to push it in the next release?

EladBezalel:
@DevVersion Please review also

src/components/tooltip/tooltip.js
DevVersion: @kwypchlo I'm trying to understand your logic here.

Won't work the following snippet?
```js
if (setVisible.queued && setVisible.value === value || scope.visible === value) return;
```

It checks if the visibility is in queue and then compares the temporary stored value.
- If `true`: Cancel the logic, because it's the same value and a queue is running -> unnecessary.

If the first condition has passed (no queue / no same value), then we will check the current scope value against the new value.
- If `true`: Cancel the logic, because the tooltip scope has already the assigned value. -> unnecessary step.


But (OR operator) if there is no queue in progress and the tooltip is already set to the same value.

kwypchlo:
@DevVersion could you review this PR before 1.1 release? It's quite important performance fix tbh

DevVersion:
@kwypchlo Oh I already did, but you didn't reply to my comment. https://github.com/angular/material/pull/6611#discussion_r56657414

kwypchlo:
@DevVersion oh sorry, It got appended on the top of the page and I must have missed it when receiving notification. And I was checking this issue twice a day for last 2 weeks waiting for any action :D 
Indeed this what you suggested is exactly the same logic and it will still work. I will replace my condition with yours and add necessary comment (just to understand the logic concept because it god pretty complicated along the way).

kwypchlo:
@EladBezalel @DevVersion @ThomasBurleson reviews were done so if this PR is ok, I'll be glad if we could merge it before 1.1, thanks!

ThomasBurleson:
@kwypchlo - We will merge this today. Thank you.

PR #6631, opened 2016-01-12
fix(ripple): remove translate from ripple animation

alxdnlnko:
`translate` transformation causes the ripple animation transition to be blocked by js code.
Set the ripple position in js and remove translate from css to prevent animation lag.

You can see the example of animation lag in anglar-material docs: https://material.angularjs.org/latest/demo/chips
When you select another page in the left sidenav menu, you can see, how the ripple of menu item lags when new page inits.

After removing `transform: translate(-50%, -50%)` from ripple classes, js code doesn't block the transition any more.

### DIFF
--- a/src/core/services/ripple/ripple.js
+++ b/src/core/services/ripple/ripple.js
@@ -365,8 +365,8 @@ InkRippleCtrl.prototype.createRipple = function (left, top) {
   var color       = this.calculateColor();
 
   ripple.css({
-    left:            left + 'px',
-    top:             top + 'px',
+    left:            (left - size / 2) + 'px',
+    top:             (top - size / 2) + 'px',
     background:      'black',
     width:           size + 'px',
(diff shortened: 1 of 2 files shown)


topherfangio:
Bumping this to the next milestone since we are already in Release Candidate stage and we're not sure what this may affect.

@robertmesserle Is this something we can test against existing internal apps to see if it causes issues?

robertmesserle:
@topherfangio I believe all PRs will now be tested against existing apps going forward

ThomasBurleson:
@alxdnlnko - why not use `translate3d()` as that uses the GPU?

alxdnlnko:
@ThomasBurleson - why use css translations for centering, if we set `left` and `top` in js and and we can adjust it there right away? Maybe, i miss something?

ThomasBurleson:

PR #6677, opened 2016-01-14
md-datepicker disabled only input

marquinhoalm:


marquinhoalm:
eu assinei!

ThomasBurleson:
Will not merge; must have associated issue and CodePen discussing the problem.

PR #6682, opened 2016-01-14
fix(datepicker): fix max date validation. Fixes #6622

### LINKED ISSUE #6622: md-datepicker shows error if default date == max date
The documentation indicates that max date is inclusive, but it does not behave this way. Also, if I pick an older date, then the current date, the error goes away.

http://codepen.io/anon/pen/zrwPMB

It appears the bug is due to minDate and maxDate including the time-of-day, whereas they should reference the beginning and end of each respective day.

AndriIushchuk:
Allow date before max date (inclusive) not depending on time of date.
Fixes https://github.com/angular/material/issues/6622

### DIFF
--- a/src/components/datepicker/datePicker.js
+++ b/src/components/datepicker/datePicker.js
@@ -378,9 +378,10 @@
       }
 
       if (this.dateUtil.isValidDate(this.maxDate)) {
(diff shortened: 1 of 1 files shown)


jelbourn:
The problem has been fixed via #6705

PR #6684, opened 2016-01-14
fix(interim, toast): add hide queue to prevent multiple interims at same time

### LINKED ISSUE #4822: Toasts overlap
Try http://jsfiddle.net/gm1o55nt/

Expected
5 toasts appear consecutively.

Actual:
You can see only the last toast ('4').

From the spec:
>There is only ever one snackbar displayed. When a second snackbar is triggered while the first is displayed, the first should start the contraction motion downwards before the second one animates upwards.

https://material-design.storage.googleapis.com/publish/material_v_4/material_ext_publish/0B6Okdz75tqQsX3VhUHNoaWVEbDQ/components_snackbars_usage_consecutivesbdifferenttype_005.webm

### LINKED ISSUE #6633: mdToast removed early
If a toast is created when an existing toast's remove animation is underway will remove the newly created toast early.

See example at codepen:
http://codepen.io/csvn/pen/XXgBvm?editors=101

DevVersion:
After this patch, the interim stack will pick up the first added element from the stack which
is not in hide process. At the moment we always take the first added (not caring about current hide process). 

Also the stack always removes the elements immediately, which causes errors when showing multiple interims (for example spamming toasts). After this fix, we will remove the elements after the hide-process is done (like a stack queue)

Fixes #6633 Fixes #4822

### DIFF
--- a/src/core/services/interimElement/interimElement.js
+++ b/src/core/services/interimElement/interimElement.js
@@ -257,7 +257,7 @@ function InterimElementProvider() {
        * A service used to control inserting and removing an element into the DOM.
        *
        */
-      var service, stack = [];
+      var service, stack = [], hideQueue = 0;
 
       // Publish instance $$interimElement service;
       // ... used as $mdDialog, $mdToast, $mdMenu, and $mdSelect
@@ -291,6 +291,10 @@ function InterimElementProvider() {
         // This hide()s only the current interim element before showing the next, new one
         // NOTE: this is not reversible (e.g. interim elements are not stackable)
 
+        if (!hideExisting) {
+          return $q.reject('Too many interim elements in queue!');
+        }
+
         hideExisting.finally(function() {
 
           stack.push(interimElement);
@@ -323,28 +327,41 @@ function InterimElementProvider() {
        */
       function hide(reason, options) {
         if ( !stack.length ) return $q.when(reason);
+        if (stack.length - 1 - hideQueue < 0) return;
         options = options || {};
 
         if (options.closeAll) {
-          var promise = $q.all(stack.reverse().map(closeElement));
-          stack = [];
-          return promise;
-        } else if (options.closeTo !== undefined) {
-          return $q.all(stack.splice(options.closeTo).map(closeElement));
-        } else {
-          var interim = stack.pop();
-          return closeElement(interim);
+          return $q.all(stack.reverse().map(closeElement));
         }
 
+        if (options.closeTo) {
+          return $q.all(stack.slice(options.closeTo).map(closeElement));
+        }
+
+        // Poll the first added interim element, which isn't currently hiding.
+        return closeElement(stack[stack.length - 1 - hideQueue]);
+
         function closeElement(interim) {
+          hideQueue++;
           interim
             .remove(reason, false, options || { })
             .catch(function( reason ) {
               //$log.error("InterimElement.hide() error: " + reason );
               return reason;
+            })
+            .finally(function() {
+              handleElementHide(interim);
             });
           return interim.deferred.promise;
         }
+      }
+
+      function handleElementHide(element) {
+        var index = stack.indexOf(element);
+        if (index != -1) {
+          stack.splice(index, 1);
+        }
+        hideQueue--;
       }
 
       /*
@@ -360,14 +377,18 @@ function InterimElementProvider() {
        *
        */
       function cancel(reason, options) {
-        var interim = stack.shift();
+        var interim = stack.pop();
         if ( !interim ) return $q.when(reason);
 
+        hideQueue++;
         interim
           .remove(reason, true, options || { })
           .catch(function( reason ) {
             //$log.error("InterimElement.cancel() error: " + reason );
             return reason;
+          })
+          .finally(function() {
+            hideQueue--;
           });
 
         return interim.deferred.promise;
@@ -378,7 +399,7 @@ function InterimElementProvider() {
        * Note: interim elements are in "interim containers"
        */
       function destroy(target) {
-        var interim = !target ? stack.shift() : null;
+        var interim = !target ? stack.splice(hideQueue, 1)[0] : null;
         var cntr = angular.element(target).length ? angular.element(target)[0].parentNode : null;
 
         if (cntr) {


src/core/services/interimElement/interimElement.js
```diff
@@ -291,6 +291,12 @@ function InterimElementProvider() {
         // This hide()s only the current interim element before showing the next, new one
         // NOTE: this is not reversible (e.g. interim elements are not stackable)
 
+        if (!hideExisting) {
+          return $q(function(resolve, reject) {
+            reject("Too many interim elements in queue!");
```
DevVersion: Not sure about the rejection, but that's the best way to prevent the interim element to stuck on showing. Any suggestions?

GongT:
cant get your point
But `var hideExisting = !options.skipHide && stack.length ? service.hide() : $q.when(true);` will constantly True. you must misunderstand `stack`.

DevVersion:
@GongT - I think you just didn't read the complete code. I will talk through the steps with you...
If we show an interim element, we check if have already an element in the stack, if so we close the previous element in stack, calling the hide method, their we will return the stack element (but not remove that **instantly** as before), so the stack will work as desired. In the current version we always removed the interim element instantly from the stack and returned the instance, that's why multiple interim elements will be removed in a wrong order. 

In this PR, we only remove the element from the stack if the hide is completed. 
If we destroy the interim element through the api, we remove the latest element in stack using our queue count, so the order will be always right!

So any questions?

GongT:
now I know your solution can work !
but I think the most important change in your code  ( if only talk about wrong hide order ) is this line:

355 -          var interim = stack.pop();
340 +          return closeElement(stack[hideQueue]);

you changed pop to "shift" ( hideQueue increase from 0, so it's hiding item from low-index, same with shift )

hideQueue means "how many items are in hiding queue". I don't think this can fix wrong "order".

(of course you are not only fixing wrong hide order, but I'm still not understand why add hideQueue)

DevVersion:
@GongT - I will try to combine your issue with this PR, I think this really near to a solution for your issue too.

src/core/services/interimElement/interimElement.js
EladBezalel:
``` js
return $q.reject('Too many interim elements in queue!');
```

src/core/services/interimElement/interimElement.js
```diff
@@ -323,30 +329,40 @@ function InterimElementProvider() {
...
         if ( !stack.length ) return $q.when(reason);
+        if (hideQueue >= stack.length) return;
         options = options || {};
 
         if (options.closeAll) {
-          var promise = $q.all(stack.reverse().map(closeElement));
-          stack = [];
-          return promise;
+          return $q.all(stack.reverse().map(closeElement));
         } else if (options.closeTo !== undefined) {
```
EladBezalel: can you fix it to be just `options.closeTo`

src/core/services/interimElement/interimElement.js
```diff
@@ -323,30 +329,40 @@ function InterimElementProvider() {
...
-          return $q.all(stack.splice(options.closeTo).map(closeElement));
+          return $q.all(stack.slice(options.closeTo).map(closeElement));
```
EladBezalel: Why use slice here while all the other functions using splice?

DevVersion: Yes, that's by intention, because we don't want to remove items from stack when the close animation starts. They will be removed later, after the animation is done.

src/core/services/interimElement/interimElement.js
```diff
@@ -323,30 +329,40 @@ function InterimElementProvider() {
...
+              handleElementHide(interim);
             });
           return interim.deferred.promise;
         }
       }
 
+      function handleElementHide(element) {
+        var index = stack.indexOf(element);
+        if (index != -1) {
+          stack.splice(index, 1);
```
DevVersion: @EladBezalel - Here the items will be removed from the stack, after the animation is done.

src/core/services/interimElement/interimElement.js
```diff
@@ -378,7 +398,7 @@ function InterimElementProvider() {
        * Note: interim elements are in "interim containers"
        */
       function destroy(target) {
-        var interim = !target ? stack.shift() : null;
+        var interim = !target ? stack.splice(hideQueue, 1)[0] : null;
```
DevVersion: @EladBezalel When destroying, we don't care anymore about the animation (it won't run), so the item will be removed instantly.

src/core/services/interimElement/interimElement.js
DevVersion: Thanks, definitely better!

src/core/services/interimElement/interimElement.js
DevVersion: Updated, but I don't get what you mean with the else statement?

```
if (options.closeAll) {
  return $q.all(stack.reverse().map(closeElement));
} else if (options.closeTo) {
  return $q.all(stack.slice(options.closeTo).map(closeElement));
} else {
  return closeElement(stack[hideQueue]);
}
```
Every check will be used. Mostly this `return closeElement(stack[hideQueue]);` when showing / hiding a toast

EladBezalel: what i meant is some thing like this: 

``` js
if (options.closeAll) {
  return $q.all(stack.reverse().map(closeElement));
} 

if (options.closeTo) {
  return $q.all(stack.slice(options.closeTo).map(closeElement));
}

return closeElement(stack[hideQueue]);
```

DevVersion: Ahh got it :smile: Done!

DevVersion:
@EladBezalel - Updated the description of the PR (also waiting for #7053 for being merged). Just made this PR compatible with the new changes from #7053. 

> After this patch, the interim stack will pick up the first added element from the stack which
is not in hide process. At the moment we always take the first added (not caring about current hide process).

> Also the stack always removes the elements immediately, which causes errors when showing multiple interims (for example spamming toasts). After this fix, we will remove the elements after the hide-process is done (like a stack queue)

DevVersion:
Closing this, because I will improve that, when we starting to support multiple dialogs.

That means, we need to take a more accurate look at the interim element stack.

PR #6691, opened 2016-01-15
feat($mdThemeProvider): allow the user to define a nonce attribute…

### LINKED ISSUE #6602: Inline <style> elements violates style-src Content Security Policy
See https://developer.mozilla.org/en-US/docs/Web/Security/CSP/CSP_policy_directives#style-src

In order to address this, there would need to be build-time tooling to let people generate their own theme css outputs.

Emeegeemee:
…for generated theme style tags

Fixes #6602

### DIFF
--- a/src/core/services/theming/theming.js
+++ b/src/core/services/theming/theming.js
@@ -14,6 +14,13 @@ angular.module('material.core.theming', ['material.core.theming.palette'])
 
 /**
  * @ngdoc method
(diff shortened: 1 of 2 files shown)


Emeegeemee:
@jelbourn Here's that pull request I mentioned I would get for you.

jelbourn:
LGTM, great PR. Thanks!

PR #6701, opened 2016-01-15
feat(input): add directive to auto select text on input focus

### LINKED ISSUE #6679: pre-select text in input on md-autofocus?
Shouldn't the text in an input box be selected when using md-autofocus on an input box?

```javascript
<md-input-container>
  <label>Name</label>
  <input name="name" type="text" value="abc" md-autofocus>
</md-input-container>
```

Currently the focus is set on the input box, but the cursor is placed at the end of text.

I'd prefer to have the text pre-selected, that's what happens when you use TAB key to move to another input box.

If you still think this behavior is fine what do you recommend to use in order to have the text pre-selected?

DevVersion:
@ThomasBurleson Is that something you thought about? A directive which selects the text on focus? 

Fixes #6679

### DIFF
--- a/src/components/input/demoBasicUsage/index.html
+++ b/src/components/input/demoBasicUsage/index.html
@@ -83,7 +83,7 @@
 
         <md-input-container class="md-block">
           <label>Biography</label>
-          <textarea ng-model="user.biography" columns="1" md-maxlength="150" rows="5"></textarea>
+          <textarea ng-model="user.biography" md-maxlength="150" rows="5" md-select-on-focus></textarea>
         </md-input-container>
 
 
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
let' modify in the input demo textarea to auto-select-all.

src/components/input/demoBasicUsage/index.html
```diff
@@ -83,7 +83,7 @@
 
         <md-input-container class="md-block">
           <label>Biography</label>
-          <textarea ng-model="user.biography" columns="1" md-maxlength="150" rows="5"></textarea>
+          <textarea ng-model="user.biography" md-maxlength="150" rows="5" md-select-on-focus></textarea>
```
DevVersion: @ThomasBurleson - Done. 

PS: `columns` seems not be a valid attribute for a native textarea, also for the `md <textarea>`.

PR #6705, opened 2016-01-15
fix(datepicker): use time-insensitive comparison for min/max.

### LINKED ISSUE #6622: md-datepicker shows error if default date == max date

jelbourn:
@mhchen for review

Fixes #6622


mhchen:
LGTM!

Issue #5196, opened 2015-10-15
md-autocomplete escape behavior is inconsistent with modern browsers (Chrome)


clshortfuse:
When using Chrome's built in autocomplete/autofill feature on forms, escape will close the autocomplete box and the cursor will stay focused on the last character used. Pressing escape again will pass the event back up to event. For example, if I'm using an md-input field in md-dialog, it would close the dialog (as expected).

With md-autocomplete, escape will close the autocomplete box and for some odd reason clear the entire field and then lose focus. Also, pressing escape won't propagate the event up. This means you can't close a dialog from an md-autocomplete input field ever.

I have fixed this in my personal fork, but I believe it should be discussed a bit before I supply a pull request.

I'm assuming the idea was to copy the Chrome address bar behavior, but that's not how autocomplete should work on actual pages.

clshortfuse:
The changes would be from this:

    case $mdConstant.KEY_CODE.ESCAPE:
        event.stopPropagation();
        event.preventDefault();
        clearValue();

        // Force the component to blur if they hit escape
        doBlur(true);

to this:

    case $mdConstant.KEY_CODE.ESCAPE:
        if (ctrl.hidden && !ctrl.loading) //fixes escape event not being sent up
            return;
        event.stopPropagation();
        event.preventDefault();

        // Clear and remove results
        ctrl.matches = [];
        ctrl.hidden  = true;
        ctrl.index   = getDefaultIndex();

Again, if you guys want to have a different style user interaction from Chrome's Autofill, that's totally fine, just we should have the option to allow events to pass up, remove the forced blur, and close the dialog on escape.

elubin:
+1

francittadini:
@ThomasBurleson Hey Thomas, was this bug fixed? If so, in what release will the fix be? Thanks!

PR #6707, opened 2016-01-15
fix(autocomplete) improved ESCAPE key interaction

clshortfuse:
If you press ESCAPE inside an autocomplete control, the input will always be captured, always clear the selection, and always blur.

The first change

        if (ctrl.hidden && !ctrl.loading) //fixes escape event not being sent up
            return;

 only responds when the autocomplete dropdown is hidden and not loading, which means the key event is sent upwards. This allows you to close dialogs when focused on an autocomplete control.

The removal of

        clearValue();

        // Force the component to blur if they hit escape
        doBlur(true);

and replaced with 

        ctrl.matches = [];
        ctrl.hidden  = true;
        ctrl.index   = getDefaultIndex();

is the EXACT same Google Chrome Autocomplete interaction. Escape will close the dropdown while the dropdown stays in focus. ESCAPE never clears the value of the window in Google Chrome. If the user wants to go one step further to provide this interaction (clear and blur focus), this is actually now supported because the ESCAPE key event is passed up.

This would keep user input consistent with the commonly used browser (Google Chrome). I'm not sure how Safari, Firefox, Edge, etc work, but I would assume they work the same.

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -434,13 +434,13 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         select(ctrl.index);
         break;
       case $mdConstant.KEY_CODE.ESCAPE:
+        if (ctrl.hidden && !ctrl.loading) //fixes escape event not being sent up
+            return;
         event.stopPropagation();
         event.preventDefault();
-        clearValue();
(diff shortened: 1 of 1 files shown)


clshortfuse:
fixes https://github.com/angular/material/issues/5196

clshortfuse:
closing for https://github.com/angular/material/pull/6707 which isn't a breaking change

PR #6714, opened 2016-01-16
fix(select): fix boolean validation multiple attribute

### LINKED ISSUE #6698: md-select attribute multiple not working
Hi,

```html
<md-select ng-model="xxxxx" placeholder="xxx" multiple="false">
    <md-optgroup label="xxxxxxxxx">
        <md-option ng-value="xxxxx" ng-repeat="xxxxx in xxxxxx">{{xxxxxx}}   </md-option>
    </md-optgroup>
</md-select>
```

In this case multiple="false" not work.

DevVersion:
Currently we just ignore the `false` <boolean> string.
That's why we should check for a `false` attribute string.
A truthy check won't work here.

Fixes  #6698

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -171,14 +171,16 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
       element.parent().append(autofillClone);
     }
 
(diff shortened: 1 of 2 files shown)


EladBezalel:
LGTM

DevVersion:
Going with #6803, which offers a util method and tests for that new validation.

PR #6715, opened 2016-01-16
fix(switch): fix switch drag functionality.

### LINKED ISSUE #2338: mdSwitch click and drag
Clicking the switch or text works great!  But if you attempt to click and slide the switch you can experience some problems.  I hate to admit that this is what I tried to do with the mouse when using the switch for the first time.

Low priority in my opinion.

### LINKED ISSUE #4719: md-switch bounces back when dragged
The following happens on [the demo page](https://material.angularjs.org/latest/#/demo/material.components.switch) and also in my own code. I'm using Chrome 44 on Mac OS X, Angular Material 0.10.1 and 0.11.

If I just click on the switch, it toggles to the other position.

If I click and hold and drag the switch from off to on, then release the mouse, the switch bounces back to off.

(But dragging the switch from on to off works; also, if I drag the switch from off to on then while continuing to hold the mouse button down, move the mouse outside the md-switch element's bounding box, THEN release the mouse button, the switch does not bounce back to off.)

However, of all these, the one that I find the most natural behavior is just dragging it from off to on and that's the one that doesn't work.

DevVersion:
This fixes the complete switch drag functionality.
Fix ported from my previous work at material2 (https://github.com/angular/material2/pull/7/files)

Fixes #4719 Fixes #2338

### DIFF
--- a/src/components/checkbox/checkbox.js
+++ b/src/components/checkbox/checkbox.js
@@ -150,7 +150,9 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
         }
       }
       function listener(ev) {
-        if (element[0].hasAttribute('disabled')) {
+        // skipToggle boolean is used by the switch directive to prevent the click event
+        // when releasing the drag. There will be always a click if releasing the drag over the checkbox
+        if (element[0].hasAttribute('disabled') || scope.skipToggle) {
           return;
         }
 
--- a/src/components/switch/switch.js
+++ b/src/components/switch/switch.js
@@ -48,7 +48,7 @@ angular.module('material.components.switch', [
  *
  * </hljs>
  */
-function MdSwitch(mdCheckboxDirective, $mdUtil, $mdConstant, $parse, $$rAF, $mdGesture) {
+function MdSwitch(mdCheckboxDirective, $mdUtil, $mdConstant, $parse, $$rAF, $mdGesture, $timeout) {
   var checkboxDirective = mdCheckboxDirective[0];
 
   return {
@@ -141,11 +141,17 @@ function MdSwitch(mdCheckboxDirective, $mdUtil, $mdConstant, $parse, $$rAF, $mdG
 
         // We changed if there is no distance (this is a click a click),
         // or if the drag distance is >50% of the total.
-        var isChanged = ngModel.$viewValue ? drag.translate > 0.5 : drag.translate < 0.5;
+        var isChanged = ngModel.$viewValue ? drag.translate < 0.5 : drag.translate > 0.5;
         if (isChanged) {
           applyModelValue(!ngModel.$viewValue);
         }
         drag = null;
+
+        // Wait for incoming mouse click
+        scope.skipToggle = true;
+        $timeout(function() {
+          scope.skipToggle = false;
+        }, 1);
       }
 
       function applyModelValue(newValue) {
--- a/src/components/switch/switch.spec.js
+++ b/src/components/switch/switch.spec.js
@@ -65,4 +65,18 @@ describe('<md-switch>', function() {
     parentScope.$apply();
     expect(element.attr('tabindex')).toEqual('-1');
   });
+
+  it('should skip click event if releasing drag over element', function() {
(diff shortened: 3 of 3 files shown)


src/components/checkbox/checkbox.js
```diff
@@ -150,7 +150,7 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
         }
       }
       function listener(ev) {
-        if (element[0].hasAttribute('disabled')) {
+        if (element[0].hasAttribute('disabled') || scope.skipToggle) {
```
DevVersion: @EladBezalel Should I add a comment here to explain what `skipToggle` is used for?

src/components/switch/switch.js
```diff
@@ -141,11 +141,17 @@ function MdSwitch(mdCheckboxDirective, $mdUtil, $mdConstant, $parse, $$rAF, $mdG
 
         // We changed if there is no distance (this is a click a click),
         // or if the drag distance is >50% of the total.
-        var isChanged = ngModel.$viewValue ? drag.translate > 0.5 : drag.translate < 0.5;
+        var isChanged = ngModel.$viewValue ? drag.translate < 0.5 : drag.translate > 0.5;
```
EladBezalel: :+1: noticed this too

src/components/switch/switch.js
```diff
@@ -141,11 +141,17 @@ function MdSwitch(mdCheckboxDirective, $mdUtil, $mdConstant, $parse, $$rAF, $mdG
...
         if (isChanged) {
           applyModelValue(!ngModel.$viewValue);
         }
         drag = null;
+
+        // Wait for incoming mouse click
+        scope.skipToggle = true;
+        $timeout(function() {
```
EladBezalel: I really don't like this, iv'e tried to fix this issue myself but also encountered the mouse click issue.
We shouldn't solve the problem, we should find the source..

DevVersion: The source is clear, we are starting the mousedown and leaving the mouseup on the switch element, thats why a click got triggered after the drag got released over the switch.

EladBezalel: yet, applying timeout here is just a hack. we should find a way to suspend clicks during drag and re-apply them when the drag ended

DevVersion: There will be always the same way, to register an click event listener on the overlaying element. I think the more lightweight solution, is to use the current `hack` (I wont call it a hack - because the click is a default behavior of the browser) or we directly prevent the click from the drag service. Thoughts?

src/components/checkbox/checkbox.js
ThomasBurleson: Yes plz

src/components/checkbox/checkbox.spec.js
```diff
@@ -136,6 +136,19 @@ describe('mdCheckbox', function() {
     expect(checkbox[0]).not.toHaveClass('md-focused');
   });
 
+  it('should skip toggle if scope property is set', function() {
+    var checkbox = compileAndLink('<md-checkbox ng-model="blue"></md-checkbox>');
+    var scope = checkbox.scope();
+
+    scope.skipToggle = true;
```
ThomasBurleson: Why would this be set to 'true' ?

DevVersion: Just to test if the checkbox skips the click, this is needed for the switch. Bust just rethought, it would make more sense to add this to the switch tests, using the switch extending the checkbox.

src/components/checkbox/checkbox.js
DevVersion: Done!

DevVersion:
@ThomasBurleson It's very hard for me to detect your reason.. What particular line or *block* needs work?

DevVersion:
@ThomasBurleson I agree, thx. @EladBezalel Can you respond to my last comment? You didn't answer yet.

EladBezalel:
@ThomasBurleson feels odd to me that we need to use `$timeout` but it seems that there's no other option, other than that lgtm

PR #6717, opened 2016-01-16
fix(dialog): prevent duplicate ids for md-dialog and md-dialog-content

### LINKED ISSUE #6339: md-dialog duplicating IDs
Consider this CodePen:

http://codepen.io/berikin/pen/gPPzRK

When the md-dialog is created, his child md-dialog-content receives the same id of the parent.

Duplicated id

Even if we force a different id for md-dialog-content, as seen in this CodePen:

http://codepen.io/berikin/pen/JGGvyE

DevVersion:
I thought about the current version, there we always have two times the same id, and not desired.
That's why I prefixed the id, so the user can still access the md-dialog-content through id.

Fixes #6339

> #breaking

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -23,8 +23,12 @@ angular
  * Inside, use an `<md-dialog-content>` element for the dialog's content, and use
  * an `<md-dialog-actions>` element for the dialog's actions.
  *
- * * ## CSS
+ * ## CSS
  * - `.md-dialog-content` - class that sets the padding on the content as the spec file
+ *
+ * ## Notes
+ * - If you specify an `id` for the `<md-dialog>`, the `<md-dialog-content>` will have the same `id`
+ * prefixed with `dialogContent_`.
  *
  * @usage
  * ### Dialog template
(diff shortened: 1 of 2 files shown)


src/components/dialog/dialog.js
```diff
@@ -842,7 +846,7 @@ function MdDialogProvider($$interimElementProvider) {
 
       var role = (options.$type === 'alert') ? 'alertdialog' : 'dialog';
       var dialogContent = element.find('md-dialog-content');
-      var dialogId = element.attr('id') || ('dialog_' + $mdUtil.nextUid());
+      var dialogContentId = ('dialog_' + element.attr('id')) || ('dialog_' + $mdUtil.nextUid());
```
ThomasBurleson: let's use `'dialogContent_'+element.attr('id'))`

src/components/dialog/dialog.spec.js
```diff
@@ -137,6 +137,30 @@ describe('$mdDialog', function() {
...
+          parent: parent
+        })
+      );
+
+      runAnimation();
+
+      var dialog = parent.find('md-dialog');
+      var content = parent[0].querySelector('md-dialog-content');
+
+      expect(content.id).toBe('dialog_' + dialog[0].id);
```
ThomasBurleson: See above change to id value prefix

src/components/dialog/dialog.js
DevVersion: Done :)

PR #6721, opened 2016-01-17
chore(): fix native angular dependencies in closure output

### LINKED ISSUE #6704: Don't add `goog.require` for ngAnimate in Closure output
When running `gulp build-all-modules --mode=closure`, the outputted `core.js` file contains the line `goog.require('ng.ngAnimate');`. The `ngAnimate` module should be excluded from generation of `goog.require` statements.

DevVersion:
We should only require our self created material components.

Fixes #6704

### DIFF
--- a/scripts/gulp-utils.js
+++ b/scripts/gulp-utils.js
@@ -219,7 +219,8 @@ exports.addClosurePrefixes = function() {
     if (module) {
       var closureModuleName = moduleNameToClosureName(module.name);
       var requires = (module.dependencies || []).sort().map(function(dep) {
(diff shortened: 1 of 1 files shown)


DevVersion:
@ThomasBurleson  - I will close this one, because you assigned robbermesserle before :)

ThomasBurleson:
@robertmesserle, @jelbourn - is this still a valid, necessary change ?

robertmesserle:
@ThomasBurleson Yes:
https://github.com/angular/bower-material/blob/master/modules/closure/core/core.js#L12

PR #6723, opened 2016-01-17
feat(autocomplete): cache elements for a specific lifetime in milliseconds

### LINKED ISSUE #2705: md-autocomplete: Add a cache lifetime attribute
`md-autocomplete` has an `md-no-cache` option to completely disable the cache. But it'd be nice to meet in the middle and let users specify a cache lifetime for results.

A use case would be searching through items which a team of people can create/edit. Caching for 30s or even a few minutes is a lot better then no cache because while searching a cache is helpful. But things aren't likely to change within 5 seconds depending on your team size. Also if we cache for ever then you would have to refresh the page to see new items from the search.

DevVersion:
Fixes #2705

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -8,7 +8,7 @@ var ITEM_HEIGHT   = 41,
     INPUT_PADDING = 2; // Padding provided by `md-input-container`
 
 function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming, $window,
-                             $animate, $rootElement, $attrs, $q) {
+                             $animate, $rootElement, $attrs, $q, $timeout) {
   //-- private variables
   var ctrl                 = this,
       itemParts            = $scope.itemsExpr.split(/ in /i),
@@ -19,7 +19,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       selectedItemWatchers = [],
       hasFocus             = false,
       lastCount            = 0,
-      promiseFetch         = false;
+      promiseFetch         = false,
+      lifetimeQueue        = [];
 
   //-- public variables with handlers
   defineProperty('hidden', handleHiddenChange, true);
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@robertmesserle - not sure I like this extension to mdAutoComplete. Can you review?

robertmesserle:
@ThomasBurleson It feels clunky and a bit specific, but the use-case brought up in #2705 is a valid one.  I'm fine with this feature being added.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -684,6 +686,20 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
   /**
+   * Removes the given term after the specified cache lifetime
+   * @param term
+   */
+  function removeAfterLifetime(term) {
+    if ($scope.cacheLifetime && lifetimeQueue.indexOf(term) === -1) {
+      lifetimeQueue.push(term);
+      $timeout(function() {
+        cache[term] = null;
+        lifetimeQueue.splice(lifetimeQueue.indexOf(term), 1);
```
robertmesserle: Should we store the index since it is referenced more than once?

robertmesserle: Or would a hashmap make more sense for this with boolean true/false values?

DevVersion: The thing is, the index will change if a previous queue task will remove an item from the list.

So we need to retrieve the index on removing.
I think indexOf is a good solution.

DevVersion:
@robertmesserle What do you mean with clunky? Is there anything I can improve? :)

src/components/autocomplete/js/autocompleteController.js
robertmesserle: Ah, right - didn't notice it was inside a timeout.  Disregard. =p

robertmesserle:
@DevVersion Oh, no, the implementation is fine.  Clunky probably wasn't the right word, but in the past, I was on the fence about this feature because it felt very edge-casey.

DevVersion:
@ThomasBurleson Rebased to master and resolved appeared merge conflict (from [here](https://github.com/angular/material/blob/52a519ead04f2faeba227fbc51a7214253cbf2ba/src/components/autocomplete/js/autocompleteController.js#L22).

ThomasBurleson:
Solution is too specifc and not central to ngMaterial. Rejecting and closing.

PR #6728, opened 2016-01-17
perf(icon): avoid multiple renders on frequent changes

### LINKED ISSUE #5598: mdIconDirective: race condition sometimes renders multiple icons when the source value changes quickly
There appears to be a subtle [race condition in `mdIconDirective`](https://github.com/angular/material/blob/747ef8264be3ac3f4d47cea4c9dfe7ffbc6d2f40/src/components/icon/js/iconDirective.js#L216) that can cause it to render multiple icons when the source value (in my case, `mdSvgIcon`) quickly changes.

Essentially, I think what happens is that the `attr.$observe` callback gets executed twice, attempting to clear the container element, *before* the [`$mdIcon` service's promise](https://github.com/angular/material/blob/747ef8264be3ac3f4d47cea4c9dfe7ffbc6d2f40/src/components/icon/js/iconDirective.js#L215) gets resolved. This results in 2 SVGs getting appended to the element, without it being cleared.

The fix may be as simple as moving the call to `element.empty()` inside the promise callback, right before `element.append(svg);`.

EladBezalel:
When the source value quickly changes it causing multiple icons rendering,
The `attr.$observe` callback gets executed twice, attempting to clear the container element, before the $mdIcon service's promise gets resolved.
This results in 2 SVGs getting appended to the element, without it being cleared.

Calling empty when the promise is resolved.

KUDOS to @voithos

fixes #5598

### DIFF
--- a/src/components/icon/js/iconDirective.js
+++ b/src/components/icon/js/iconDirective.js
@@ -212,9 +212,11 @@ function mdIconDirective($mdIcon, $mdTheming, $mdAria ) {
 
         element.empty();
         if (attrVal) {
-          $mdIcon(attrVal).then(function(svg) {
-            element.append(svg);
-          });
+          $mdIcon(attrVal)
+            .then(function(svg) {
+              element.empty();
+              element.append(svg);
+            });
         }
 
       });


src/components/icon/js/iconDirective.js
```diff
@@ -210,11 +210,12 @@ function mdIconDirective($mdIcon, $mdTheming, $mdAria ) {
       // Use either pre-configured SVG or URL source, respectively.
       attr.$observe(attrName, function(attrVal) {
 
-        element.empty();
```
ThomasBurleson: Should we not keep this here also? To force clear all content WHILE loading the SVG?

EladBezalel: Sounds legit

PR #6729, opened 2016-01-17
fix(dialog): changed translate3d to translate

### LINKED ISSUE #4544: md-dialog: IE11 / Edge blurry text and buttons after animating onto screen
I don't have a ready CodePen for this at the moment, but I do have some screenshots of what I'm describing. When IE11 or Edge finishes animating in an `md-dialog` element, the resulting dialog appears fuzzy and blurred when compared to Chrome and Firefox:

image
**IE 11 Image**

image
**Chome Image**

image
**Edge Image**

image
**Firefox Image**

I imagine this is an artifact of the CSS styling that is done to animate the `md-dialog` into place, and it isn't likely to be something easy to fix or a high priority either. I just wanted to report it because in our limited testing for our apps here, we've been getting users reporting it (along with complaints about how they think their eyesight is going bad until they switch to Firefox or Chrome).

EladBezalel:
fixes #4544

### DIFF
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -21,12 +21,12 @@ md-dialog {
   &.md-transition-in {
     opacity: 1;
     transition: $swift-ease-out;
-    transform: translate3d(0,0,0) scale(1.0);
+    transform: translate(0,0) scale(1.0);
   }
   &.md-transition-out {
     opacity: 0;
     transition: $swift-ease-out;
-    transform: translate3d(0,100%,0) scale(0.2);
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@EladBezalel - See http://stackoverflow.com/questions/22111256/translate3d-vs-translate-performance.
So I think for only IE should we use translate (instead of `translate3d`).

EladBezalel:
Ran the jsprdf both on my pc and phone with chrome and both showed better results with translate..

ThomasBurleson:
@EladBezalel - sgtm. thx.

PR #6730, opened 2016-01-18
refactor(bottomSheet,dialog,list,select,input,menu,menuBar): denote private classes with _md prfix

rschmukler:
**Note**: This commit will break input until it has been updated with private classes. I modified some of the classes as they interact with `md-select`. Presumably @topherfangio changes will result in the same changes as I made resulting in full compatibility between select and input.

### DIFF
--- a/docs/app/partials/menu-link.tmpl.html
+++ b/docs/app/partials/menu-link.tmpl.html
@@ -3,7 +3,7 @@
     ng-href="{{section.url}}"
     ng-click="focusSection()">
   {{section | humanizeDoc}}
-  <span class="md-visually-hidden"
+  <span class="_md-visually-hidden"
     ng-if="isSelected()">
     current page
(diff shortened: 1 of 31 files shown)


ThomasBurleson:
@jelbourn - Can you review these to confirm these changes. The goal is to also use these changes as 'standards' for CSS changes in the other components.

topherfangio:
@rschmukler Looks like we did make the same changes. Thanks for the heads-up.

LGTM

jelbourn:
@rschmukler For all the classes that _aren't_ private, we're able to say that they won't see any major changes in the foreseeable future?

ThomasBurleson:
@rschmukler - ping for @jelbourn question.

ThomasBurleson:
Duplicate. See aggregation into PR #7248.

PR #6736, opened 2016-01-18
fix(button): add aria tests to button and uses by default the text content as aria-label

### LINKED ISSUE #6731: button does not infer aria-label properly
mbButton does not use $mdAria correct to set the aria-label based on the label/content text.

DevVersion:
Fixes #6731


ThomasBurleson:
@DevVersion - please contact one of the team before working on any more PRs... as a team member may already be producing one. Let's not duplicate efforts.

DevVersion:
Sorry can't know about the paralell work. How should I contact any one? I have only contact to Elad through Gitter.

PR #6737, opened 2016-01-18
fix(tabs): Change updatePagingWidth to handle subpixel tab width

mcosand:
updatePagingWidth now uses getBoundingClientRect() instead of offsetWidth
to compute total width of tabs. offsetWidth returns only the integer part
of the tab width which can cause the pagination-wrapper to be to narrow,
causing the last tab to wrap and become not visible.

Addresses https://github.com/angular/material/issues/5439


susu:
:+1:

ThomasBurleson:
@mcosand - lgtm

ThomasBurleson:

PR #6741, opened 2016-01-18
fix(select): made select line height aligned with input

### LINKED ISSUE #5524: The text in text inputs and in selects has a slightly different baseline
This is regardless of browser. This screenshot from https://material.angularjs.org/HEAD/demo/select demonstrates: 

screen shot 2015-09-14 at 10 49 23 pm

Originally discussed in #4357.

EladBezalel:
* Made pixel perfect fixes to input and select according the spec
* Aligned select label line height with input

fixes #5524

### DIFF
--- a/src/components/input/input.scss
+++ b/src/components/input/input.scss
@@ -2,14 +2,14 @@ $input-container-padding: 2px !default;
 
 $input-label-default-offset: 24px !default;
 $input-label-default-scale: 1.0 !default;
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
@EladBezalel  - Do we have a sample/demo that shows this?

EladBezalel:
No and i don't think we should have specifically for that

PR #6756, opened 2016-01-19
fix(icon, menuBar, slider, tabs, util): use CSP compliant element.style…

### LINKED ISSUE #6602: Inline <style> elements violates style-src Content Security Policy

Emeegeemee:
…instead of style attribute

Fixes #6602

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -163,6 +163,9 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
 
   function handleMaxTabWidth (newWidth, oldWidth) {
     if (newWidth !== oldWidth) {
+      angular.forEach(elements.tabs, function(tab) {
+        tab.style.maxWidth = newWidth + 'px';
+      });
       $mdUtil.nextTick(ctrl.updateInkBarStyles);
     }
   }
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -247,8 +247,7 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
         var scrollMask = angular.element(
           '<div class="md-scroll-mask">' +
           '  <div class="md-scroll-mask-bar"></div>' +
-          '</div>');
-        scrollMask.css('z-index', zIndex);
+          '</div>').css('z-index', zIndex);
         element.appendChild(scrollMask[0]);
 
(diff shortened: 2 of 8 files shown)


src/components/tabs/js/tabsController.js
```diff
@@ -163,6 +163,9 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
 
   function handleMaxTabWidth (newWidth, oldWidth) {
     if (newWidth !== oldWidth) {
+      angular.forEach(elements.tabs, function(tab) {
+        tab.style.maxWidth = newWidth + 'px';
```
jelbourn: Does this update if the max-width changes? 

(I'm not actually sure if it needs to. cc @robertmesserle)

src/core/util/util.js
```diff
@@ -321,7 +322,14 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
     },
     floatingScrollbars: function() {
       if (this.floatingScrollbars.cached === undefined) {
-        var tempNode = angular.element('<div style="width: 100%; z-index: -1; position: absolute; height: 35px; overflow-y: scroll"><div style="height: 60px;"></div></div>');
+        var tempNode = angular.element('<div><div></div></div>');
```
jelbourn: I believe you could write this as:
```js
var tempNode = angular.element('<div><div></div></div>').css({
  width: '100%',
  'z-index': -1,
  ...
});
```

src/components/tabs/js/tabsController.js
Emeegeemee: It does, handleMaxTabWidth is the setter function for the maxTabWidth property on the [ctrl](https://github.com/Emeegeemee/material/blob/619a7a3adffa05e1340157792ee0064cf1b74ac7/src/components/tabs/js/tabsController.js#L25). It does need to handle the max width changing or else the tabs content get cut off, from my manual testing.

src/core/util/util.js
Emeegeemee: Done and updated all css() calls.

jelbourn:
LGTM

jelbourn:
Thanks for the contribution! It should be part of the 1.0.3 release later this week.

PR #6769, opened 2016-01-19
feat($mdDialog): add prompt

### LINKED ISSUE #4995: add prompt dialog?
Addition to $mdDialog. just a suggestion.

zbjornson:
Add preconfigured dialog for prompt. Fixes #4995.

I didn't see the bit about "not actively reviewing unsolicited PRs" until after I did this :(. This is relatively minor, only 20 lines of code; remaining 160 are docs and tests.

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -366,6 +366,32 @@ function MdDialogDirective($$rAF, $mdTheming, $mdDialog) {
 
 /**
  * @ngdoc method
+ * @name $mdDialog#prompt
+ *
+ * @description
+ * Builds a preconfigured dialog with the specified message and input box. You can call show and the promise returned
+ * will be resolved only if the user clicks the prompt action on the dialog, passing the input value as the first argument.
+ *
+ * @returns {obj} an `$mdDialogPreset` with the chainable configuration methods:
+ *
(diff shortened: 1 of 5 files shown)


src/components/dialog/dialog.js
```diff
@@ -500,11 +534,16 @@ function MdDialogProvider($$interimElementProvider) {
...
         this.hide = function() {
-          $mdDialog.hide(true);
+          $mdDialog.hide(this.$type === 'prompt' ? this.result : true);
         };
         this.abort = function() {
           $mdDialog.cancel();
         };
+        this.keypress = function($event) {
+          if ($event.keyCode === $mdConstant.KEY_CODE.ENTER) {
+            $mdDialog.hide(this.result)
```
ThomasBurleson: Do you not want to call `this.hide()` instead in order to account for type == 'prompt'?

ThomasBurleson:
Nice work and enhancement. Like the tests also.

src/components/dialog/dialog.js
zbjornson: `ng-keypress` is only set if `dialog.$type == 'prompt'` on line 520, so it won't make a difference, if I understand your question.

zbjornson:
Thanks for the review!

PR #6770, opened 2016-01-19
fix(util/datepicker): Remove usages of scope().

### LINKED ISSUE #6474: $mdutil.findFocusTarget uses scope() which will not work if debug information is turned off
function scanForFocusable(target, selector) {
        var elFound, items = target[0].querySelectorAll(selector);

        // Find the last child element with the focus attribute
        if ( items && items.length ){
          var EXP_ATTR = /\s*\[?([\-a-z]*)\]?\s*/i;
          var matches = EXP_ATTR.exec(selector);
          var attribute = matches ? matches[1] : null;

          items.length && angular.forEach(items, function(it) {
            it = angular.element(it);

            // If the expression evaluates to FALSE, then it is not focusable target
            var focusExpression = it[0].getAttribute(attribute);
            var isFocusable = !focusExpression || !$mdUtil.validateScope(it) ? true :
                              (it.scope().$eval(focusExpression) !== false );

            if (isFocusable) elFound = it;
          });
        }
        return elFound;
      }

The scope() function is only available if debug information is available! Evaluating the expression assigned to the focus attribute will break as soon as this is turned off in production.

topherfangio:
A few places use `element.scope()` to retrieve the proper scope,
however, this is only available when Angular's debugging option
is enabled and thus fails in production.

Replace usages of `.scope()` in `util.js` and `datepicker.js`
with appropriate alternatives. Remove unused function
`$mdUtil.validateScope()`.

Fixes #6474.

### DIFF
--- a/src/components/datepicker/datePicker.js
+++ b/src/components/datepicker/datePicker.js
@@ -330,7 +330,7 @@
     if (this.$attrs['ngDisabled']) {
       // The expression is to be evaluated against the directive element's scope and not
       // the directive's isolate scope.
-      var scope = this.$mdUtil.validateScope(this.$element) ? this.$element.scope() : null;
+      var scope = this.$scope.$parent;
 
       if (scope) {
         scope.$watch(this.$attrs['ngDisabled'], function(isDisabled) {
--- a/src/core/util/autofocus.js
+++ b/src/core/util/autofocus.js
@@ -1,0 +1,108 @@
+angular.module('material.core')
+  .directive('mdAutofocus', MdAutofocusDirective)
+
+  // Support the deprecated md-auto-focus and md-sidenav-focus as well
+  .directive('mdAutoFocus', MdAutofocusDirective)
+  .directive('mdSidenavFocus', MdAutofocusDirective);
+
+/**
+ * @ngdoc directive
+ * @name mdAutofocus
+ * @module material.core.util
+ *
+ * @description
+ *
+ * `[md-autofocus]` provides an optional way to identify the focused element when a `$mdDialog`,
+ * `$mdBottomSheet`, or `$mdSidenav` opens or upon page load for input-like elements.
+ *
+ * When one of these opens, it will find the first nested element with the `[md-autofocus]`
+ * attribute directive and optional expression. An expression may be specified as the directive
+ * value to enable conditional activation of the autofocus.
+ *
+ * @usage
+ *
+ * ### Dialog
+ * <hljs lang="html">
+ * <md-dialog>
+ *   <form>
+ *     <md-input-container>
+ *       <label for="testInput">Label</label>
+ *       <input id="testInput" type="text" md-autofocus>
+ *     </md-input-container>
+ *   </form>
+ * </md-dialog>
+ * </hljs>
+ *
+ * ### Bottomsheet
+ * <hljs lang="html">
+ * <md-bottom-sheet class="md-list md-has-header">
+ *  <md-subheader>Comment Actions</md-subheader>
+ *  <md-list>
+ *    <md-list-item ng-repeat="item in items">
+ *
+ *      <md-button md-autofocus="$index == 2">
+ *        <md-icon md-svg-src="{{item.icon}}"></md-icon>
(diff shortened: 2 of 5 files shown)


topherfangio:
@ThomasBurleson Can you take a quick look at this PR and provide feedback on the following:

1. **Overall approach** - Does it look valid/acceptable?
2. **Breaking autofocus out** - Should we do this/is there a better way/place to put it?
3. **Docs** - I didn't update the docs; do we need to make any modifications?
4. **Datepicker** - Can you think of any case where `this.$scope.$parent` won't work?

src/core/util/autofocus.js
```diff
@@ -0,0 +1,105 @@
...
+    link: postLink
+  }
+}
+
+function postLink(scope, element, attrs) {
+  var attr = attrs.mdAutoFocus || attrs.mdAutofocus || attrs.mdSidenavFocus;
+
+  // Setup a watcher on the proper attribute to update a class we can check for in $mdUtil
+  scope.$watch(attr, function(canAutofocus) {
+    element.toggleClass('_md-can-autofocus', canAutofocus);
```
ThomasBurleson: why not store this value in the element.data where it is more hidden?

ThomasBurleson: Does this support expression evaluation; should we not use attrs.$observe() instead?

If we use this approach to add a 'signature' flag to the element, it we made it public then  users could also attach styling to the element. That may be desirable... if so, then this class should be public `md-autofocus`.

> I think `md-autofocus` class is better than `md-can-autofocus`.

src/components/datepicker/datePicker.js
```diff
@@ -330,7 +330,7 @@
     if (this.$attrs['ngDisabled']) {
       // The expression is to be evaluated against the directive element's scope and not
       // the directive's isolate scope.
-      var scope = this.$mdUtil.validateScope(this.$element) ? this.$element.scope() : null;
+      var scope = this.$scope.$parent;
```
ThomasBurleson: `$mdUtil.validateScope` is only used by `datePicker.js` and `util.js` 
And in util.js, it is only used to scan for focusable.

Anyway to fix the util.js or **remove** that fragile utility?

src/core/util/autofocus.js
topherfangio: `attrs.$observe()` is used for interpolated values; we're not using interpolation here, so I think `$watch` is better: (see https://docs.angularjs.org/api/ng/type/$compile.directive.Attributes and http://stackoverflow.com/questions/14876112/angularjs-difference-between-the-observe-and-watch-methods for more info; I had to re-read them to check)

Basically, the value we get from using `$observe()` is the string `"shouldAutoFocus==1"` which we would then have to evaluate and setup a watcher for. I guess we don't cover the case of someone changing the actual expression (like making it `shouldAutoFocus==2`), but I doubt that people will do that.

I'm adding some additional tests to ensure it works for all types of expressions.

Regarding the `md-autofocus`, I'm not opposed to renaming it (I don't like "can" either), however, I can't think of a reason to make it public (instead of private). This class would be applied to any element that "can" automatically receive focus, not one that currently has focus (I imagine it will get `md-focused` in that case). So, I don't know why people would style it unless they want to make all autofocus-capable inputs stand out a bit.

If we don't make it public, then we can indeed set it on `element.data` instead of a class; I generally just use classes so developers can see it visually in the inspector pane a bit more easily.

src/components/datepicker/datePicker.js
topherfangio: I have removed `$mdUtil.validateScope()` as it was not called anywhere else.

ThomasBurleson:
@topherfangio - see my comments.

topherfangio:
@ThomasBurleson Sorry; looks like GitHub hid my comments after I pushed: https://github.com/angular/material/pull/6770#discussion_r50466619

src/core/util/autofocus.js
ThomasBurleson: Let's use `_md-autoFocus` class and support interpolation on the attribute string. This gives us the maximum flexibility.

ThomasBurleson:
LGTM

PR #6782, opened 2016-01-20
fix(select): fix set form to pristine if ng-model for multiple select is predefined.

### LINKED ISSUE #6556: Multiple md-select , adding one resets pristine to true
Ok this is my first message here.

Using lastest version of angularjs/material , it seems i found out an issue using md-select.

See this jsfiddle : http://jsfiddle.net/ypjgdzxp/3/

What you can see here is that i'm adding rows on the fly wich contains an md-select. The problem is that adding a new md-select to the Dom resets the form pristine to true , even if it was false/dirty before.
In my exemple , juste put anything in the upper input "Nom de couche" wich will make the form dirty , then add a new row using the "+" button at the bottom.

DevVersion:
Bug introduced in 09bd5a3b52f29dfe530192099d2c405f204ec3ad

Fixes #6556

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -228,9 +228,12 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
         });
       }
 
-      if (formCtrl) {
+      if (formCtrl && angular.isDefined(attr.multiple)) {
         $mdUtil.nextTick(function() {
-          formCtrl.$setPristine();
+          var hasModelValue = ngModelCtrl.$modelValue || ngModelCtrl.$viewValue;
+          if (hasModelValue) {
+            formCtrl.$setPristine();
+          }
         });
       }
 
--- a/src/components/select/select.spec.js
+++ b/src/components/select/select.spec.js
@@ -387,6 +387,19 @@ describe('<md-select>', function() {
         expect(selectedOptions(el).length).toBe(0);
       }));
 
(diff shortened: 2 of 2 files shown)


src/components/select/select.js
```diff
@@ -228,9 +228,10 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
         });
       }
 
-      if (formCtrl) {
+      if (formCtrl && angular.isDefined(attr.multiple)) {
```
ThomasBurleson: So if we are **not using multiple**, then we **never `$setPristine()`? Is that valid logic?

DevVersion: Yes that is valid, as you can see in 09bd5a3b52f29dfe530192099d2c405f204ec3ad, we should only set the form to pristine if `multiple` objects are preselected in the `ngModel`.
I manually tested that too, if there is a predefined value for a `single select`, we don't need to set the form pristine. It only sets the form to dirty if we render `multiple`

ThomasBurleson:
@DevVersion - for 99% of all PRs, each one will need 1 or more unit tests. Can you provide a unit test to validate and verify these changes?

DevVersion:
@ThomasBurleson - Yes, I will add a Unit Test for both cases.

DevVersion:
@ThomasBurleson - Added two tests for the `$pristine` behavior. also when not using **multiple**

PR #6482, opened 2015-12-28
feat(tabs): allow disabling select click event by adding `md-no-selec…

epelc:
…t-click` attribute

This adds a new attribute(md-no-select-click) to disable the click event firing on select. Therefore allowing external code to manage the selected tab via `md-active` and `ng-click/ui-sref` on each `md-tab`. Normally this works fine but if you have a child state and reload the page `md-active="$state.includes('somestate')"` will fire the click event for your ui-sref kicking you back up to your tabs main state(see the 2 issues for more info on this). Adding `md-no-select-click` to your `md-tabs` fixes this by not firing the click event.

Closes #5351 #5930

PR #6784, opened 2016-01-20
feat(layout): Added `*-print` attributes for all layout and flex attributes

epelc:
This adds support for controlling `flex` and `layout` attributes when printing. Previously printing would use your layout for the `sm` breakpoint. Now you can override this by using attributes like `layout-print="row"` or `flex-offset-print="2"` etc.

I'm suspecting this will primarily be of use for enterprise apps as enterprise users tend to print a lot of things.

@ThomasBurleson Can you take a look? I really needed this for a while now as my users print a lot. It reduced my primary view from 3-4 pages of paper down to 2.

### DIFF
--- a/src/core/services/layout/layout.scss
+++ b/src/core/services/layout/layout.scss
@@ -132,10 +132,12 @@
   // Do not use unitless flex-basis values in the flex shorthand because IE 10-11 will error.
   // Also use 0% instead of 0px since minifiers will often convert 0px to 0 (which is unitless and will have the same problem).
   // Safari, however, fails with flex-basis : 0% and requires flex-basis : 0px
-  @media screen\0 {
-      .#{$flexName} {
-        flex: 1 1 0%;
-      }
+  @if $name != '-print' { 
+    @media screen\0 {
+        .#{$flexName} {
+          flex: 1 1 0%;
+        }
+    }
   }
 
 
@@ -512,6 +514,7 @@
 @media (min-width: $layout-breakpoint-lg) {
   @include layouts_for_breakpoint(gt-lg);
   @include layouts_for_breakpoint(xl);
+  @include layouts_for_breakpoint(print);
 
   // BIGGER THAN LARGE SCREEN
   .hide, .hide-gt-xs, .hide-gt-sm, .hide-gt-md, .hide-gt-lg {
@@ -525,3 +528,11 @@
 
 }
 
+@media print {
+  // PRINT
+  @include layouts_for_breakpoint(print);
+  
+  .hide-print:not(.show-print):not(.show) {
+    display: none;
+  }
+}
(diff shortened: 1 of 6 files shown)


ThomasBurleson:
@epelc - review now.

epelc:
@ThomasBurleson Thank you!

1 more pr(#6482) after this and I'll be back on master :)

src/core/services/layout/layout.scss
```diff
@@ -512,6 +514,7 @@
 @media (min-width: $layout-breakpoint-lg) {
   @include layouts_for_breakpoint(gt-lg);
   @include layouts_for_breakpoint(xl);
+  @include layouts_for_breakpoint(print);
```
ThomasBurleson: Why are you doing this here?

epelc: @ThomasBurleson I forgot to remove that sorry.

I was having problems because of the ie hack using `screen\0` [here](https://github.com/angular/material/pull/6784/files#diff-afb8b370e6204859a801fc83d3b0f2f4R145). It basically returned from that block when it got to the screen part for the `print` styles. I didn't know this so I tried to get print styles working for something but I must of forgotten about it.

I removed it in a commit below.

ThomasBurleson:
@epelc - you realize this overrides the current breakpoints to forces styles normally used in the default alias?

e.g.

default with `layout="row" flex="10"`

```css
.layout-row > .flex-10, .layout-print-row > .flex-10 {
    -webkit-flex: 1 1 10%;
        -ms-flex: 1 1 10%;
            flex: 1 1 10%;
    max-width: 10%;
    max-height: 100%;
    box-sizing: border-box; 
}
```

And the css for with `layout="row" flex-sm="30"`

```css
.layout-row > .flex-sm-30, .layout-print-row > .flex-sm-30 {
    -webkit-flex: 1 1 30%;
        -ms-flex: 1 1 30%;
            flex: 1 1 30%;
    max-width: 30%;
    max-height: 100%;
    box-sizing: border-box; 
}
```

So when display in small mode the layout and flex looks good on screen, but the print options (as you defined) would *fallback** to defaults:

```css
  .layout-row > .flex-print-10, .layout-print-row > .flex-print-10 {
    -webkit-flex: 1 1 10%;
        -ms-flex: 1 1 10%;
            flex: 1 1 10%;
    max-width: 10%;
    max-height: 100%;
    box-sizing: border-box; 
}
```

Is this what you expect ?

ThomasBurleson:
Would you not want the print to use the current screen layout view?

epelc:
@ThomasBurleson I know it overrides it while printing. It's meant too.

The regular breakpoints might work well if your printing on different paper sizes. But 99% of the time your using 8.5x11inch paper and it is using the 'sm' breakpoint for this which makes everything way too big.

I know it might not be 100% for all imaginable cases. But I think it covers most real world situations well.

ThomasBurleson:
@epelc  - thx for this PR. I created a new PR #6923 (based on this one) with many additional changes.
Closing this one.

epelc:
@thomasburleson I see and you really went over the top. Great job! Thanks for working on this.

PR #6785, opened 2016-01-20
fix(tabs): fix dynamic height demo for tabs

### LINKED ISSUE #6780: Angular-Material Demo Codepen examples don't run
None of the links to codepen examples on the Angular-Material site work.  I believe this is because the link to angular-material.js is bad. If you look in the settings on the demo codepen examples, it is pointing to:

http://localhost:8080/angular-material.js

image

DevVersion:
This issue was introduced in d8602747f597879532ff2d1bb23c7a9bec6e6b88.

Because of the new `svg-asset-cache` script, we need to init the module ourself.
The tabs demo was always working, because the `asset cache` initialized the module (unintentional).

Fixes #6780

### DIFF
--- a/src/components/tabs/demoDynamicHeight/script.js
+++ b/src/components/tabs/demoDynamicHeight/script.js
@@ -1,0 +1,1 @@
+angular.module('tabsDemoDynamicHeight', ['ngMaterial']);


ThomasBurleson:
@DevVersion - good catch.

ThomasBurleson:
BTW - this does NOT fix #6780.

DevVersion:
@ThomasBurleson - In my case the demo is now working if I append the new change. (Waiting for the demo to built up) - Check here http://codepen.io/DevVersion/pen/VeXVPy

PR #6803, opened 2016-01-21
fix(autocomplete, select, virtualrepeat, input): correctly validate attribute values

### LINKED ISSUE #5831: ng-required does not work with md-autocomplete with floating label
using ng-required is not working with md-auto complete
here is the code pen
http://codepen.io/anon/pen/jbJXjW

### LINKED ISSUE #5896: md autocomplete required not working
Tag required in md autocomplete elements is not working. I need to change it to ng-required="true" or to required="true"?

### LINKED ISSUE #6393: md-autocomplete 'required' validation broken in 1.0.1
The Floating Label Autocomplete demo is broken in versions 1.0.0, 1.0.1 and HEAD, although it was working in 0.11.4. Leaving the field empty does not cause the form to have the 'required' error.  

Other validations (min, max) seem to be working.

### LINKED ISSUE #6415: Codepen and validation in demo is not working
Hi,

Codepen for [autocomplete demo: FloatingLabel] (https://material.angularjs.org/latest/demo/autocomplete) is not working. Also this demo has validation, but errors are not show when we leave name field blank.

### LINKED ISSUE #6698: md-select attribute multiple not working

### LINKED ISSUE #7063: mdVirtualRepeat: does not respect md-on-demand boolean value
I'm using angular-material 1.0.4 and md-virtual-repeat does not seem to consider the value of md-on-demand when I use the following code (`numbers` is a plain old array)
```html
<md-virtual-repeat-container id="list-container">
      <div md-virtual-repeat="num in numbers" md-on-demand="false">
        {{num}}
      </div>
</md-virtual-repeat-container>
```
I see this error:
```
Error: When md-on-demand is enabled, the Object passed to md-virtual-repeat
must implement functions getItemAtIndex() and getLength()
```
[Plunkr](http://plnkr.co/edit/YJUovKA0SaunlckEafeH?p=preview)

Removing the attribute solves the problem. I'd consider doing that, but in my case, I'm wrapping md-virtual-repeat within a directive that determines if the list is paginated (`md-on-demand=true`), or a bulk array of elements (`md-on-demand=false`, like the vertical usage example in the demo page) based on some factors.

### LINKED ISSUE #7105: Support asterisk on input with `ng-required` in the same way as with `required`
Currently only `<input required>`is support but a lot of uses involve `<input ng-required="foo.bar">`.

DevVersion:
Fixes the validation of the boolean attributes.

### Autocomplete
**Attributes:**
<table>
  <td>required</td>
  <td>disabled</td>
</table>

- `required` ignored an empty attribute value (`<md-autocomplete required>`)
- `disabled` ignored an empty attribute value (`<md-autocomplete disabled>`)

---
### Input
**Attributes:**
<table>
  <td>md-no-asterisk</td>
</table>
- `md-no-asterisk` is ignoring false attribute value (`<input md-no-asterisk="false">`)

**Extra Changes:**
- `required` is not updating the `md-required` class (now observing for `ng-required`)
- Wrong test-case for `md-no-asterisk`

---
### Select
**Attributes:**
<table>
  <td>multiple</td>
</table>
- `multiple` is ignoring false attribute value (`<md-select multiple="false">`)

---
### Virtual Repeat
**Attributes**
<table>
  <td>md-on-demand</td>
</table>
- `md-on-demand` is ignoring false attribute value (`<md-virtual-repeat md-on-demand="false">`)

**Extra Changes:**
- Missing attribute test for `md-on-demand`

---

### Notes
- Validating `negative` attribute values for `required`, `readonly`, `disabled` will be ignored. 
- Validating `0` as falsy value is also a useful feature and should be implemented too, because sometimes your backend will return a `boolean` in numeric style (See [here](https://en.wikipedia.org/wiki/Boolean_data_type)). 

Fixes #5831 Fixes #7063 Fixes #6415 Fixes #6393 Fixes #5896 Fixes #6698 Fixes #7105

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -164,8 +164,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    */
   function configureWatchers () {
     var wait = parseInt($scope.delay, 10) || 0;
-    $attrs.$observe('disabled', function (value) { ctrl.isDisabled = !!value; });
-    $attrs.$observe('required', function (value) { ctrl.isRequired = !!value; });
+    $attrs.$observe('disabled', function (value) { ctrl.isDisabled = $mdUtil.parseAttributeBoolean(value, false); });
+    $attrs.$observe('required', function (value) { ctrl.isRequired = $mdUtil.parseAttributeBoolean(value, false); });
     $scope.$watch('searchText', wait ? $mdUtil.debounce(handleSearchText, wait) : handleSearchText);
     $scope.$watch('selectedItem', selectedItemChange);
     angular.element($window).on('resize', positionDropdown);
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -642,6 +642,19 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
       return current;
     },
 
+    /**
+     * Parses an attribute value, mostly a string.
+     * By default checks for negated values and returns `false´ if present.
+     * Negated values are: (native falsy) and negative strings like:
+     * `false` or `0`.
+     * @param value Attribute value which should be parsed.
+     * @param negatedCheck When set to false, won't check for negated values.
+     * @returns {boolean}
+     */
+    parseAttributeBoolean: function(value, negatedCheck) {
+      return value === '' || !!value && (negatedCheck === false || value !== 'false' && value !== '0');
+},
+
     hasComputedStyle: hasComputedStyle
   };
 
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -856,6 +856,55 @@ describe('<md-autocomplete>', function() {
 
       element.remove();
     });
+
+    it('should validate an empty `required` as true', function() {
+      var scope = createScope();
+      var template = '\
+          <md-autocomplete\
+              md-selected-item="selectedItem"\
+              md-search-text="searchText"\
+              md-items="item in match(searchText)"\
+              md-item-text="item.display"\
+              md-min-length="0" \
+              required\
+              placeholder="placeholder">\
+            <span md-highlight-text="searchText">{{item.display}}</span>\
+          </md-autocomplete>';
+      var element = compile(template, scope);
+      var ctrl = element.controller('mdAutocomplete');
+
+      expect(ctrl.isRequired).toBe(true);
+    });
+
+    it('should correctly validate an interpolated `ng-required` value', function() {
+      var scope = createScope();
+      var template = '\
+          <md-autocomplete\
+              md-selected-item="selectedItem"\
+              md-search-text="searchText"\
+              md-items="item in match(searchText)"\
+              md-item-text="item.display"\
+              md-min-length="0" \
+              ng-required="interpolateRequired"\
+              placeholder="placeholder">\
+            <span md-highlight-text="searchText">{{item.display}}</span>\
+          </md-autocomplete>';
+      var element = compile(template, scope);
+      var ctrl = element.controller('mdAutocomplete');
+
+      expect(ctrl.isRequired).toBe(false);
+
+      scope.interpolateRequired = false;
+      scope.$apply();
+
+      expect(ctrl.isRequired).toBe(false);
+
+      scope.interpolateRequired = true;
+      scope.$apply();
+
+      expect(ctrl.isRequired).toBe(true);
+    });
+
   });
 
   describe('md-highlight-text', function() {
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -264,8 +264,7 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
     var hasNgModel = !!ctrls[1];
     var ngModelCtrl = ctrls[1] || $mdUtil.fakeNgModel();
     var isReadonly = angular.isDefined(attr.readonly);
-    var isRequired = angular.isDefined(attr.required);
-    var mdNoAsterisk = angular.isDefined(attr.mdNoAsterisk);
+    var mdNoAsterisk = $mdUtil.parseAttributeBoolean(attr.mdNoAsterisk);
 
 
     if (!containerCtrl) return;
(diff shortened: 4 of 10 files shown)


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -163,8 +163,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    */
   function configureWatchers () {
     var wait = parseInt($scope.delay, 10) || 0;
-    $attrs.$observe('disabled', function (value) { ctrl.isDisabled = !!value; });
-    $attrs.$observe('required', function (value) { ctrl.isRequired = !!value; });
+    $attrs.$observe('disabled', function (value) { ctrl.isDisabled = value !== 'false'; });
```
ThomasBurleson: Could the value not interpolate to '0' also ? This solutions seems limited.

DevVersion:
@ThomasBurleson - It now supports interpolated `0` values. Added a util method for attribute parsing, this will be helpful for #6714 too (if a util method isn't desired let me know :) ) + accompanying tests.

DevVersion:
-- Now fixing #6698 too, after merging my old PR into this by using the new util method.

src/core/util/util.js
```diff
@@ -739,6 +739,17 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
     },
 
+    /**
+     * Parses an attribute value to false if the value is falsy but not an empty value
+     * otherwise it will be true.
+     * Will be false if
+     * @param value Interpolated Attribute Value
+     * @returns {boolean}
+     */
+    parseAttributeBoolean: function(value) {
```
ThomasBurleson: @jelbourn - I think we should regex match on the `value` string. Thoughts or code suggestions ?

DevVersion: Won't do a negated match for `/(false|0)/g`?

jelbourn: I wouldn't consider the "zero" string to be falsy. Without that the check is fine as-is.

robertmesserle: @DevVersion Can you remove the check for '0' and update the PR?  Other than that, it looks good to me.

DevVersion: @robertmesserle - Yes will do, just thought I need that because Thomas asked about the interpolated `zero` as a falsy value.

DevVersion:
@ThomasBurleson @robertmesserle @jelbourn - Okay removed the `zero` check (but kept the initial check structure). But in my opinion I think `zero` should be used as a `falsy` value too.

robertmesserle:
Just to clarify, the use-case would be something like:

```
<md-autocomplete ... required="0"></md-autocomplete>
```

Is that right?

DevVersion:
Okay, just wan't to answer your question :smile: .
- Yes exactly this will allow using `zero` as false value too. This would be really useful when using something like a backend which is returning booleans in numeric style.

DevVersion:
@ThomasBurleson At the moment in this PR, in most cases we support `required="false"` `required="true"`, so the `required` attribute is accepting a boolean as value. But this is not a default behavior in HTML from W3C.

The users should better use `ng-disabled`, `ng-readonly` and `ng-required`.
So IMO we should not parse the attribute values from these type of attributes.
- `disabled`
- `readonly`
- `required`

jelbourn:
While treating these as boolean attributes is _correct_, it would be a breaking API change.

DevVersion:
@jelbourn Sorry, I didn't get it. Should we continue this.. or should I remove that?

robertmesserle:
@ThomasBurleson @jelbourn Is this ready to merge?  LGTM

jelbourn:
Sorry, I wasn't clear. I don't think we should change this API since it will break people that are depending on the current behavior.

DevVersion:
@jelbourn Okay, then I will remove the `required`, `readonly` and `disabled` attribute (and surely the references to (#5896, #6415). Please correct me if I understood that wrong. To be honest I'm still not sure, which question you answered.. Haha

You talking about this one?
> Just to clarify, the use-case would be something like:
>
><md-autocomplete ... required="0"></md-autocomplete>
>Is that right?

>I don't have a strong opinion one way or the other, I'll defer to @ThomasBurleson and @jelbourn for this.

Or about the boolean validation for the already-said attributes?

DevVersion:
@jelbourn @ThomasBurleson Updated the description and added more information to the given attributes. Can you please review a second time and read the description? Thanks in advance.

EladBezalel:
@jelbourn according what I've read this PR doesn't seems like a breaking API, it just enables wider range of values, more than that the old behavior won't break.

jelbourn:
@EladBezalel We would be changing the behavior of, for example `<md-select disabled="false">`. Someone might be depending on this already, so changing it would be a breaking API change.

DevVersion:
@jelbourn I've changed that a few days ago, it's currently only a issue of the autocomplete directive (`required`, `disabled`). Should I add a parameter to the util function which excludes the negative checks (this can be used for attributes like `readonly`, `disabled`, `required`), because in general I need the negative checks for attributes like `md-no-asterisk` or `multiple`. I think this will be a good solution. Thoughts?

wembernard:
@DevVersion May I suggest to add `required` attribute in `select` section?

DevVersion:
@wembernard I would like to have this PR merge, before looking at the actual `select` issue, because this may not relate to the actual boolean validation.

BTW: As discussed with @jelbourn I removed the breaking attribute changes. PTAL

leocaseiro:
Will this PR fix the issue with **multiple="false"**?

If yes, I vote +1

DevVersion:
@leocaseiro Yes, that will be a good workaround for that, and yes this PR adds a basic boolean validation for `multiple`, this should be added, because the only attributes which should not validate for negated boolean values are `readonly`, `required`, `disabled`...

leocaseiro:
how about ng-required, ng-multiple, ng-disabled and ng-readonly?

DevVersion:
@leocaseiro Yes exactly that's the right usage for these attributes. By default `readonly`, `required`, `disabled` (which are native attributes - without the `md-` prefix) are not interpolating a boolean value. That's why you should use the `ng-` prefixed attributes for these three native attributes, when you wan't to interpolate the boolean.

PR #6809, opened 2016-01-21
fix(input): check if parent form is undefined before checking if it's…

### LINKED ISSUE #6807: dialog: form with required input causes error when dialog is closed
If you have a dialog with a form and an `md-input-container` that is using `required` it will cause an error when you close the dialog if your input hasn't been touched.

image
image

epelc:
… submitted

This fixes an issue I had with requiring inputs in form dialogs causing errors when you close the dialog. It should also fix the same issue in other situations as this is a fix for `md-input-container`. It was just triggered with a dialog in my case.

Fixes #6807

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -273,7 +273,11 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
     var isParentFormSubmitted = function () {
       var parent = $mdUtil.getClosest(element, 'form');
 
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
Thanks @epelc.

tonyawad88:
I had the same issue and had to remove the <form> element from the dialog template. I guess now after this fix I can put the form tag back in :)

PR #6827, opened 2016-01-23
fix(layout): change last `start` to `flex-start`

### LINKED ISSUE #6818: Bug report: Justify Content: start, should be flex-start;
I was working with angular material and I needed the property layout-align="start start", I wasn't getting the expected result, so I checked the developer tools. There, I noticed that justify-content has a syntax issue, instead of "justify-content: flex-start;" it is "justify-content: start;" 

I am using the latest Angular Material version.

DevVersion:
Fixes #6818


ThomasBurleson:
FYI - This file will be deprecated in the near future.

DevVersion:
Thanks for the information :)

PR #6840, opened 2016-01-25
fix(mdSelect): Selected label shouldn't copy the ripple container in …

### LINKED ISSUE #6819: md-select violates CSP
The `md-select` directive violates CSP when an item is clicked and the select closes. Upon closing, there is still some remaining styling form the ink-ripple which is applied when the item is clicked. The following HTML is added to the DOM (new lines added for readability):

``` html
<div class="md-text ng-binding">some text value</div>
<div class="md-ripple-container" style="background-color: rgba(0, 0, 0, 0.0980392);">
<div class="md-ripple md-ripple-placed md-ripple-scaled md-ripple-active" style="left: 52px; top: 22px; width: 147.472px; height: 147.472px; border-color: rgb(0, 0, 0); background: rgb(0, 0, 0);">
</div>
</div>
```

Violating call in [select.js](https://github.com/angular/material/blob/de5237f1b7ad8185e1af143f467fae01edaa33e5/src/components/select/select.js#L255).

References #6602, #6691, #6756.

Emeegeemee:
…the md-option

fixes #6819

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -648,7 +648,12 @@ function SelectMenuDirective($parse, $mdUtil, $mdTheming) {
         var mapFn;
 
         if (mode == 'html') {
-          mapFn = function(el) { return el.innerHTML; };
+          mapFn = function(el) {
+            var html = el.innerHTML;
+            // Remove the ripple container from the selected option, copying it would cause a CSP violation.
+            var rippleContainer = el.querySelector('.md-ripple-container');
+            return rippleContainer ? html.replace(rippleContainer.outerHTML, '') : html;
+          };
         } else if (mode == 'aria') {
           mapFn = function(el) { return el.hasAttribute('aria-label') ? el.getAttribute('aria-label') : el.textContent; };
         }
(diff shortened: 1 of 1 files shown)


Emeegeemee:
@jelbourn Please take a look when you have the chance, Thanks!

src/components/select/select.js
```diff
@@ -648,7 +648,12 @@ function SelectMenuDirective($parse, $mdUtil, $mdTheming) {
         var mapFn;
 
         if (mode == 'html') {
-          mapFn = function(el) { return el.innerHTML; };
+          mapFn = function(el) {
```
jelbourn: I don't quite follow what this map function is doing- could you add a bit more explanation?

jelbourn:
cc @rschmukler

src/components/select/select.js
Emeegeemee: Done

Emeegeemee:
Thanks for the review, PTAL

jelbourn:
Merged via b7073759dbb321dce941d95b56a01123a32e6379

PR #6852, opened 2016-01-26
fix(menu): prevent container scaling if still animating

### LINKED ISSUE #6806: <md-menu> shrinks on iOS safari
If an `md-menu` is opened when iOS safari is in "minimal UI" mode (after scrolling down), it will shrink to a tiny size. Example:

md-menu-bug

This doesn't affect the demo site since the behavior is prevented by restricting overflow scrolling to a `div`, rather than the `body` element.

DevVersion:
This fixes issues with minimal UI for Safari / IOS

*Already tested, but this needs more manual testing*

Fixes #6806

### DIFF
--- a/src/components/menu/js/menuServiceProvider.js
+++ b/src/components/menu/js/menuServiceProvider.js
@@ -118,6 +118,7 @@ function MenuProvider($$interimElementProvider) {
       // Return the promise for when our menu is done animating in
       return showMenu()
         .then(function(response) {
+          opts.isAnimating = false;
           opts.alreadyOpen = true;
           opts.cleanupInteraction = activateInteraction();
           return response;
@@ -134,6 +135,9 @@ function MenuProvider($$interimElementProvider) {
           var position = calculateMenuPosition(element, opts);
 
           element.removeClass('md-leave');
+
+          // Prevent container scaling while running animation
+          opts.isAnimating = true;
 
           // Animate the menu scaling, and opacity [from its position origin (default == top-left)]
           // to normal scale.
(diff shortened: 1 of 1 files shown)


src/components/menu/js/menuServiceProvider.js
```diff
@@ -476,7 +480,7 @@ function MenuProvider($$interimElementProvider) {
         top: Math.round(position.top),
         left: Math.round(position.left),
         // Animate a scale out if we aren't just repositioning
-        transform: !opts.alreadyOpen ? $mdUtil.supplant('scale({0},{1})', [scaleX, scaleY]) : undefined,
+        transform: !opts.alreadyOpen && !opts.isAnimating ? $mdUtil.supplant('scale({0},{1})', [scaleX, scaleY]) : undefined,
```
ThomasBurleson: I worry that this is not a fix but a bandaid. The issue is not the animation but the scale X/Y calculations when scrolled right ?

DevVersion: Yes, the actual problem is, that on IOS the minimal-ui will trigger a orientantionchange / resize event after the menu animation is started. 

I already thought about adding this up to the actual $$rAF but it would make more sense to let it update the position except the `scale` from `undefined` to `scale x/y`

jpike88:
Copy pasted that code into my own, didn't seem to work for me. Chromium on a Samsung Tab 3

DevVersion:
I'm not sure you pasted everything correctly. But this fix works in my case for **IOS** Safari

jpike88:
Wasn't that just one line of code?

DevVersion:
Nope, that were about 6 lines

jpike88:
Oh damn my bad. Can't believe I didn't notice. Will let you know results

topherfangio:
Reassigned to myself to do some testing/review.

topherfangio:
Just tried to do some manual testing but couldn't reproduce. Do we have a Codepen that is failing that we can test against? I tried the one in #7704 but it didn't seem to do it on my phone (iPhone 6; iOS 9.1).

DevVersion:
@topherfangio This is weird to test, you need a page which runs not inside of a codepen or something else, it should be a standalone page which is able to scroll.

And then you should try opening a menu, when being in Apple's Minimal UI-Mode.

See Issue #6806

DevVersion:
@topherfangio This PR is now deprecated. I'm feeling like the solution is not the correct way to continue with. What do you think? If you think otherwise, let me know :)

ThomasBurleson:

PR #6858, opened 2016-01-27
fix(slider): clamp md-slider's initialisation

whitneyit:
Previously when a model was set with a value outside of the `min` or
the `max` of the slider, it would call `setSliderPercent` with the
initial value and no check would occur to make sure that the value of
the model could not extend the range of the slider.

The below screenshot demonstrates what the slider looks like when it has
been initialised with a value of `-1`.

screenshot

As you can see, the thumb is floating way off to the left. This is due
to the `left: -25%` style that has been applied to the element. It is
worth noting that the `min` and `max` are `0` and `4` respectively.


ThomasBurleson:
lgtm.

Issue #1798, opened 2015-03-05
md-autocomplete : $http promise fails to work


kschwidder:
When the items are returned from an $http promise the md-autocomplete fails to work. 
From a $http promise the data items are returned as response.data. md-autocomplete expects the data is passed back as an array.
The array is contained in the response.data object but md-autocomplete is reading response as the array which fails.

```js
$scope.loadSkillTags = function (query) {
    var data = {qData: query};
    return SkillService.querySkills(data);
};
```

```html
<md-autocomplete
	     md-selected-item="data.selectedItem"
	     md-search-text="data.searchText"
	     md-items="item in loadSkillTags(data.searchText)"
	     md-item-text="item"
	     placeholder="Talent">
	 <span md-highlight-text="data.searchText">{{item.name}}</span>
 </md-autocomplete>
 ```

kschwidder:
Workaround is like this

                                $scope.loadSkillTags = function (query) {
                                    var data = {qData: query};
                                    return SkillService.querySkills(data).then(
                                        function(data) {return data.data;}, function(){return [];});
                                };

ThomasBurleson:
@robertmesserle - I think we should check the return value from the promise and check it is is the standard $http `response` object... just to add a little smarts to the component.

robertmesserle:
Agreed, will add this when I get a chance.

kschwidder:
Great. Thx.

sombriks:
would be great if the docs get updated too
https://material.angularjs.org/latest/#/api/material.components.autocomplete/directive/mdAutocomplete

to support promises is a great thing and it's not mentioned in md-items

balamanigandan:
md-autocomplete : $http promise never fails to work, it will work perfectly. 

Kindly refer the post: here is the live code is available: http://stackoverflow.com/questions/35652828/http-issue-values-cant-be-returned-before-a-promise-is-resolved-in-angular-m

ashishsiddhu:
-You can use Angularjs promises to get data from $http call.
        <body>
                <form name="searchForm" ng-submit="searchForm.$valid && submit()" novalidate>
                    <md-autocomplete flex-gt-sm="35"
                		      	required md-input-name="autocompleteField" 
                		      	md-input-maxlength="80" 
                		      	md-selected-item="selectedItem"
                		        md-search-text="formdata.searchText" 
                		        md-items="item in querySearch(formdata.searchText)" 
                		        md-item-text="item.display" 
                		        md-min-length="0"
[...]

PR #6860, opened 2016-01-27
fix(autocomplete): do not use .then() for $http promise

m1ch3lcl:
$http returns promise with both .then() and .success() methods :
- `.then()` : calls back with http response object with following properties : data, status, config, headers, statusText (cf https://docs.angularjs.org/api/ng/service/$http#general-usage)
- `.success()` : calls back with data coming from server

currently, autocomplete use both methods, and when `handleResults` is called from `.then(...)`, items are replaced by http response object

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -639,9 +639,9 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     } else if (items) {
       setLoading(true);
       $mdUtil.nextTick(function () {
-        if (items.success) items.success(handleResults);
-        if (items.then)    items.then(handleResults);
-        if (items.finally) items.finally(function () {
+        if (items.success)   items.success(handleResults);
+        else if (items.then) items.then(handleResults);
+        if (items.finally)   items.finally(function () {
           setLoading(false);
         });
       },true, $scope);


m1ch3lcl:
I signed it!

ThomasBurleson:
Related and fixed in #6521. Close this PR.

winniehell:
I think I don't agree that this is fixed by #6521. In my opinion these are two different issues.

winniehell:
This issues is actually what I asked in my comment on [de423ae48d593c08f0277376b7e3e80571a3ba48](https://github.com/angular/material/commit/de423ae48d593c08f0277376b7e3e80571a3ba48#commitcomment-15232172). That commit actually removed `$q.when` (what @ThomasBurleson suggested to use in #6521) for fixing #1798.

m1ch3lcl:
I double checked the content of both PR and I am pretty sure, as @winniehell says, that both issues are completely different : the problem here is that when using autocomplete with $http promise, we use successively both .success() and .then() methods : 
```
if (items.success) items.success(handleResults);
if (items.then)    items.then(handleResults);
```
this part of the code is not changed by the PR you mentionned : it lacks an `else if` statement, making it unusable with $http promise (data is replaced with http response object)

@ThomasBurleson : I didn't say that we should not listen for `.success()`, it's the opposite : i think whenever `.success()` is available, we should listen for it instead of `.then()`

PR #6878, opened 2016-01-27
fix(dialog): fix dialog resizing on window resize

### LINKED ISSUE #6876: $mdDialog position not updating when escapeToClose set to false
You can see it in action on plunkr: https://plnkr.co/edit/EC9CexUlBa41jrrpybdJ?p=preview

1. click on escapeToClose: true, resize window height and dialog stays centered and its height updating as you resize window. as expected.

2. click on escapeToClose: false, resize window height and dialog do not update its height and not staying centered.

Is this a bug or am i missing something?

DevVersion:
The current resize listeners for the window are misplaced inside of the condition of `escapeToClose == true`.

Fixes #6876


ThomasBurleson:
Good catch.

PR #6888, opened 2016-01-27
<fix(datepicker)>: change mdDateUtil.isDateWithinRange to ignore the …

### LINKED ISSUE #6887: datepicker: inconsistency between allowed dates in UI and text field
Commit 0e334cd fixed the issue where the validation for the datepicker was respecting the time component of the dates in md-min-date and md-max-date. However, the logic that decides whether to show a date as selectable in the popup still looks at the time component.

This means that e.g. for md-min-date="new Date('2015-05-02 12:00:00')", the date '2015-05-02' typed into the text field will be valid, but on the popup date picker, 2015-05-02 will not be enabled.

I'll send a PR shortly.

ademuri:
…time component of the date

Previously, mdDateUtil.isDateWithinRange did a simple comparison on the given dates. Since this
component is used for dates (and not times or datetimes), this leads to inconsistent behavior.

Fixes #6887

### DIFF
--- a/src/components/datepicker/dateUtil.js
+++ b/src/components/datepicker/dateUtil.js
@@ -204,8 +204,9 @@
      * Creates a date with the time set to midnight.
      * Drop-in replacement for two forms of the Date constructor:
      * 1. No argument for Date representing now.
-     * 2. Single-argument value representing number of seconds since Unix Epoch.
-     * @param {number=} opt_value
+     * 2. Single-argument value representing number of seconds since Unix Epoch
+     * or a Date object.
+     * @param {number|Date=} opt_value
      * @return {Date} New date with time set to midnight.
      */
     function createDateAtMidnight(opt_value) {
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@jelbourn - can you review ?

src/components/datepicker/dateUtil.js
```diff
@@ -220,15 +221,18 @@
...
       * If minDate or maxDate are not dates, they are ignored.
       * @param {Date} date
       * @param {Date} minDate
       * @param {Date} maxDate
       */
      function isDateWithinRange(date, minDate, maxDate) {
-       return (!angular.isDate(minDate) || minDate <= date) &&
-           (!angular.isDate(maxDate) || maxDate >= date);
+       var dateAtMidnight = createDateAtMidnight(date);
+       var minDateAtMidnight = angular.isDate(minDate) ? createDateAtMidnight(minDate) : null;
```
jelbourn: Instead of `angular.isDate`, use `isValidDate`

jelbourn:
LGTM aside from one comment. Great contribution, thanks!

src/components/datepicker/dateUtil.js
ademuri: Done.

ademuri:
PTAL

PR #6893, opened 2016-01-28
update(theme): add Green hue 500 contrast color

pdore-netfore:
Fixes https://github.com/angular/material/issues/6885

> #breaking


pdore-netfore:
I signed it!

Issue #6884, opened 2016-01-27
Palette 'blue-grey' should include hue 700 as a contrast light color


pdore-netfore:

ThomasBurleson:
PR requested.

PR #6894, opened 2016-01-28
update(theme): add Blue-Grey hue 700 contrast color

pdore-netfore:
Fixes https://github.com/angular/material/issues/6884

> #breaking


pdore-netfore:
I signed it!

PR #6897, opened 2016-01-28
feat(chips): allow to specify a maximum of chips that can be added through user input

### LINKED ISSUE #6864: md-chips apply Limit to chips maximum chips like md-input-maxlength
Hi
I want to apply the limit on md-chips. like a user cannot add more than 3 chips as input
Thanks

DevVersion:
The current behavior on chips limit reaching is that there will happen nothing (no input clear).
Any suggestions for the behavior?

Fixes #6864

### DIFF
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -409,6 +409,71 @@ describe('<md-chips>', function() {
         }));
       });
 
+      describe('md-max-chips', function() {
+
+        beforeEach(function() {
+          // Clear default items to test the max chips functionality
+          scope.items = [];
+        });
+
+        it('should not add a new chip if the max-chips limit is reached', function() {
+          var element = buildChips('<md-chips ng-model="items" md-max-chips="1"></md-chips>');
+          var ctrl = element.controller('mdChips');
+
+          element.scope().$apply(function() {
+            ctrl.chipBuffer = 'Test';
+            simulateInputEnterKey(ctrl);
+          });
+
+          expect(scope.items.length).toBe(1);
+
+          element.scope().$apply(function() {
+            ctrl.chipBuffer = 'Test 2';
+            simulateInputEnterKey(ctrl);
+          });
+
+          expect(scope.items.length).not.toBe(2);
+        });
+
+        it('should update the md-max-chips model validator for forms', function() {
+          var template =
+            '<form name="form">' +
+            '<md-chips name="chips" ng-model="items" md-max-chips="1"></md-chips>' +
+            '</form>';
+
+          var element = buildChips(template);
+          var ctrl = element.find('md-chips').controller('mdChips');
+
+          element.scope().$apply(function() {
+            ctrl.chipBuffer = 'Test';
+            simulateInputEnterKey(ctrl);
+          });
+
+          expect(scope.form.chips.$error['md-max-chips']).toBe(true);
+        });
+
+        it('should not reset the buffer if the maximum is reached', function() {
+          var element = buildChips('<md-chips ng-model="items" md-max-chips="1"></md-chips>');
+          var ctrl = element.controller('mdChips');
+
+          element.scope().$apply(function() {
+            ctrl.chipBuffer = 'Test';
+            simulateInputEnterKey(ctrl);
+          });
+
+          expect(scope.items.length).toBe(1);
+
+          element.scope().$apply(function() {
+            ctrl.chipBuffer = 'Test 2';
+            simulateInputEnterKey(ctrl);
+          });
+
+          expect(ctrl.chipBuffer).toBe('Test 2');
+          expect(scope.items.length).not.toBe(2);
+        });
+
+      });
+
       describe('md-autocomplete', function() {
         var AUTOCOMPLETE_CHIPS_TEMPLATE = '\
           <md-chips ng-model="items">\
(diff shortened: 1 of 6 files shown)


src/components/chips/js/chipsController.js
```diff
@@ -225,8 +228,14 @@ MdChipsCtrl.prototype.getAdjacentChipIndex = function(index) {
  * call out to the md-transform-chip method, if provided.
  *
  * @param newChip
+ * @return {boolean|undefined} Returns true if the max chips limit is reached,
+ *   otherwise will always return undefined
  */
 MdChipsCtrl.prototype.appendChip = function(newChip) {
+  // Don't append the chip if the maxmimum is reached
+  // Return true, to show that the limit is reached
+  if (this.items.length >= this.maxChips) return true;
```
DevVersion: I can swap the return values, it would make more sense to return `false` if the chip can't be added, but then I need to return `false` if the chip is a duplicate... and in this case it should clear the buffer. (so I think the current solution is the most simple)

DevVersion: --- Improved.

ThomasBurleson:
We should support ng-messages with custom errors when the limit is reached. (much like input with errors).

ThomasBurleson:
Without ngMessages support, this feature creates more issues than it solves. Removed from the merge-queue until feature is finished.

DevVersion:
Okay, I will work on that :+1:

DevVersion:
@ThomasBurleson - Should this work as same as input do? with the animations, then we better should make it usable with the input container?

DevVersion:
@ThomasBurleson All done :+1: Would be great if you review again.

src/components/chips/js/chipsDirective.js
```diff
@@ -67,6 +67,9 @@
    *    displayed when there is at least on item in the list
    * @param {boolean=} readonly Disables list manipulation (deleting or adding list items), hiding
    *    the input and delete buttons
+   * @param {number=} md-max-chips The maximum number of chips allowed to add through user input.
+   *    <br/><br/>The validation property `md-max-chips` can be used when the max chips
+   *    amount is reached.
```
DevVersion: Please review here @ThomasBurleson

ThomasBurleson:
The demo is broken and does not work: entire a 4th chip and the error message shows.
@DevVersion - did you test this ?

DevVersion:
@ThomasBurleson - I was sure everything works. Will check it now!

DevVersion:
@ThomasBurleson - Update :smile: - Forgot adding ngMessages as dependency.

DevVersion:
@ThomasBurleson - Now fixed!

ThomasBurleson:
@DevVersion:

* what about the error text color?
* Is the error message too close to the chip bottom border?

DevVersion:
I don't know, I thought I should just add ngMessages support - but I can create a custom styling just for the styling? Support for ngMessages for the complete chips will be another issue / PR, (would like to that, if needed :smile: )

ThomasBurleson:
Also this logic appears wrong:

```js
  var isMaximumReached = !(this.maxChips && this.items.length >= this.maxChips);
  this.ngModelCtrl.$setValidity('md-max-chips', isMaximumReached);
};
```

And `this.items.length` is a number while `this.maxChips` is === "5".

Please use this code:

```js
MdChipsCtrl.prototype.hasMaxChips = function() {
  this.maxChips = (angular.isString(this.maxChips) ) ? parseInt(this.maxChips,10) || 0 :
                  (angular.isUndefined(this.maxChips) ) ? 0 : this.maxChips;

  return (this.maxChips > 0) && (this.items.length >= this.maxChips);
};
```

ThomasBurleson:
full Support for ngMessages is overkill for this PR. Yet if we use ngMessages to show an error, then that message should be **colored** accordingly IMO.

DevVersion:
this.maxChips won't be a string, it will be interpolated from the scope, so it will be an integer?

ThomasBurleson:
It is a string... inspect in the dev console.

DevVersion:
@ThomasBurleson - Sorry wrong approach.. Updated it, added style to the custom ngMessages in the demo? and used your string validation for `maxChips`

ThomasBurleson:
@DevVersion - you need to **thoroughly** test each PR before submitting. Otherwise the impacts could be huge.

src/components/chips/chips.spec.js
```diff
@@ -409,6 +409,71 @@ describe('<md-chips>', function() {
         }));
       });
 
+      describe('md-max-chips', function() {
```
ThomasBurleson: @DevVersion - are these passing for you ?

DevVersion:
@ThomasBurleson - Yes, next time I will test it more thoroughly. Just wasn't sure about ngMessages, but in my case everything worked perfect until the ngMessages support was requested. So next time I will check more thoroughly when updating a PR.

src/components/chips/chips.spec.js
DevVersion: Now Passing (just said in the comment before :smile:, so please don't be to harsh). Updated the `check` to ->

```
MdChipsCtrl.prototype.hasMaxChips = function() {
  if (angular.isString(this.maxChips)) this.maxChips = parseInt(this.maxChips, 10) || 0;

  return this.maxChips > 0 && this.items.length >= this.maxChips;
};
```

If undefined the it will return false, because `undefined > 0` will be false.

PR #6908, opened 2016-01-28
feat: Initial $mdPanel implementation.

ErinCoughlan:
@gmoothart @DerekLouie @KarenParker 

This is the initial files for the $mdPanel implementation. Please be picky about style and documentation since we will continue with it in future commits.

@jelbourn I'm also unsure how to get the $mdPanelRef to appear on the material site. It's not a service, so there's no page for it. But I do want the specific definition, properties, and methods defined.

Current version: panel_ref

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -8,14 +8,19 @@ angular
     ])
     .service('$mdPanel', MdPanelService);
 
+
+/***************************************************************************************
+ *                              PUBLIC DOCUMENTATION                                   *
+ ***************************************************************************************/
+
 /**
  * @ngdoc service
  * @name $mdPanel
  * @module material.components.panel
  *
  * @description
- * `$mdPanel` is a general component that can be used for any panels on the screen,
- * such as a tooltip, dialog, or custom popup.
+ * `$mdPanel` is a robust, low-level service for creating floating panels on
+ * the screen. It can be used to implement tooltips, dialogs, pop-ups, etc.
  *
  * @usage
  * <hljs lang="js">
@@ -24,56 +29,42 @@ angular
  *
  *   angular
  *       .module('demoApp', ['ngMaterial'])
- *       .controller('PizzaBuilderController', PizzaBuilderController)
- *       .controller('PizzaDialogController', DialogController);
+ *       .controller('DemoDialogController', DialogController);
  *
- *   function PizzaBuilderController($element, $mdPanel) {
- *     var panelRef;
- *     var toppings = [
- *       'pepperoni',
- *       'sausage',
- *       'ham',
- *       'peppers',
- *       'onions',
- *       'pineapple'
- *     ];
+ *   var panelRef;
  *
- *     function showPizzaPanel($event) {
- *       var config = {
- *         parent: angular.element(document.body),
- *         controller: DialogController,
- *         controllerAs: 'ctrl',
- *         position: $element,
- *         targetEvent: $event,
- *         template:
- *           '<div class="pizza-panel">' +
- *           '  <h2>Choose your favorite topping.</h2>' +
- *           '  <md-list>' +
- *           '    <md-list-item ng-repeat="topping in ctrl.toppings">' +
- *           '      <md-button ng-click="ctrl.favoriteTopping = topping">' +
- *           '        {{ topping }}' +
- *           '      </md-button>' +
- *           '    </md-list-item>' +
- *           '  </md-list>' +
- *           '  <div class="actions">' +
- *           '    <md-button ng-click="ctrl.closeDialog()">CANCEL</md-button>' +
- *           '  </div>' +
- *           '</div>',
- *         clickOutsideToClose: true,
- *         escapeToClose: true,
- *         focusOnOpen: true
- *       }
- *       panelRef = $mdPanel.create(config);
- *       panelRef.open()
- *           .finally(function() {
- *             panelRef = undefined;
- *           });
+ *   function showPanel($event) { *
+ *     var panelPosition = $mdPanelPosition
+ *         .absolute()
+ *         .top('50%')
+ *         .left('50%');
+ *
+ *     var panelAnimation = $mdPanelAnimation
+ *         .targetEvent($event)
+ *         .defaultAnimation('md-panel-animate-fly')
+ *         .closeTo('.show-button');
+ *
+ *     var config = {
+ *       attachTo: angular.element(document.body),
+ *       controller: DialogController,
+ *       controllerAs: 'ctrl',
+ *       position: panelPosition,
+ *       animation: panelAnimation,
+ *       targetEvent: $event,
+ *       template: 'dialog-template.html',
+ *       clickOutsideToClose: true,
+ *       escapeToClose: true,
+ *       focusOnOpen: true
  *     }
+ *     panelRef = $mdPanel.create(config);
+ *     panelRef.open()
+ *         .finally(function() {
+ *           panelRef = undefined;
+ *         });
  *   }
  *
  *   function DialogController($mdPanelRef, toppings) {
  *     var toppings;
- *     var favoriteTopping;
  *
  *     function closeDialog() {
  *       $mdPanelRef.close();
@@ -98,17 +89,13 @@ angular
  *   - `groupName` - `{string=}`: Name of panel groups. This group name is used for
  *     configuring the number of open panels and identifying specific behaviors for
  *     groups. For instance, all tooltips will be identified using the same groupName.
- *   - `maxOpen` - `{number=}` : Maximum number of panel in a specific group that
- *     can be opened at a given time. If multiple panels have the same groupName, but
- *     different values of maxOpen, only the first one will be considered. A warning
- *     will be logged in the console.
  *   - `template` - `{string=}`: HTML template to show in the dialog. This **must** be
  *     trusted HTML with respect to Angular’s
  *     [$sce service](https://docs.angularjs.org/api/ng/service/$sce).
  *   - `templateUrl` - `{string=}`: The URL that will be used as the content of
  *     the panel.
  *   - `controller` - `{(function|string)=}`: The controller to associate with the
- *     panel. The controller will be injected with a reference to the returned
+ *     panel. The controller can inject a reference to the returned
  *     panelRef, which allows the panel to be closed, hidden, and shown.
  *     Any fields passed in through locals or resolve will be bound to the controller.
  *   - `controllerAs` - `{string=}`: An alias to assign the controller to on the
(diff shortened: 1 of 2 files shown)


jelbourn:
cc @ThomasBurleson also

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+    .service('$mdPanel', MdPanelService);
+
+/**
+ * @ngdoc service
+ * @name $mdPanel
+ * @module material.components.panel
+ *
+ * @description
+ * `$mdPanel` is a general component that can be used for any panels on the screen,
+ * such as a tooltip, dialog, or custom popup.
```
jelbourn:
```
`$mdPanel` is a robust, low-level service for creating floating panels on the screen. It can be used to implement tooltips, dialogs, pop-ups, etc.
```

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+ * <hljs lang="js">
+ * (function(angular, undefined) {
+ *   ‘use strict’;
+ *
+ *   angular
+ *       .module('demoApp', ['ngMaterial'])
+ *       .controller('PizzaBuilderController', PizzaBuilderController)
+ *       .controller('PizzaDialogController', DialogController);
+ *
+ *   function PizzaBuilderController($element, $mdPanel) {
```
jelbourn: Why is this here?

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+ *
+ *     function showPizzaPanel($event) {
+ *       var config = {
+ *         parent: angular.element(document.body),
+ *         controller: DialogController,
+ *         controllerAs: 'ctrl',
+ *         position: $element,
+ *         targetEvent: $event,
+ *         template:
+ *           '<div class="pizza-panel">' +
```
jelbourn: It's okay to use a fake `templateUrl` instead of a whole template for the example.

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+ *       'sausage',
+ *       'ham',
+ *       'peppers',
+ *       'onions',
+ *       'pineapple'
+ *     ];
+ *
+ *     function showPizzaPanel($event) {
+ *       var config = {
+ *         parent: angular.element(document.body),
```
jelbourn: Maybe `attachTo` instead of `parent`? What do other people think?

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+ *         controller: DialogController,
+ *         controllerAs: 'ctrl',
+ *         position: $element,
```
jelbourn: Is `position` something that will change later?

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+ * @param opt_config {Object=} Specific configuration object that may contain the
+ * following properties:
+ *
+ *   N/A
+ *
+ * TODO(ErinCoughlan): Add the following config options.
+ *   - `groupName` - `{string=}`: Name of panel groups. This group name is used for
+ *     configuring the number of open panels and identifying specific behaviors for
+ *     groups. For instance, all tooltips will be identified using the same groupName.
+ *   - `maxOpen` - `{number=}` : Maximum number of panel in a specific group that
```
jelbourn: I think `maxOpen` isn't an option for a single instance, but a setting on the service itself. Something like
```ts
$mdPanel.setGroupMax('tooltips', 1);
```

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+ *     can be opened at a given time. If multiple panels have the same groupName, but
+ *     different values of maxOpen, only the first one will be considered. A warning
+ *     will be logged in the console.
+ *   - `template` - `{string=}`: HTML template to show in the dialog. This **must** be
+ *     trusted HTML with respect to Angular’s
+ *     [$sce service](https://docs.angularjs.org/api/ng/service/$sce).
+ *   - `templateUrl` - `{string=}`: The URL that will be used as the content of
+ *     the panel.
+ *   - `controller` - `{(function|string)=}`: The controller to associate with the
+ *     panel. The controller will be injected with a reference to the returned
```
jelbourn: `will be injected with` > `can inject`

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+ *     scope.
+ *   - `locals` - `{Object=}`: An object containing key/value pairs. The keys will be
+ *     used as names of values to inject into the controller. For example,
+ *     `locals: {three: 3}` would inject `three` into the controller, with the value
+ *     3.
+ *   - `resolve` - `{Object=}`: Similar to locals, except it takes promises as
+ *     values. The panel will not open until all of the promises resolve.
+ *   - `parent` - `{Element=}`: The element to append the dialog to. Defaults to
+ *     appending to the root element of the application.
+ *   - `position` - `{(string|Element|Object)=}`: The query selector, DOM element,
```
jelbourn: What do you think about this API shape instead:
```js
postion: $mdPanelPosition.relativeTo(someElement).offsetX(-20).offsetY(50);
// or
postion: $mdPanelPosition.absolute().top('50%').left('50%');
```
The end result is that `$mdPanel` will always get a position of a certain shape rather than having to interpret everything.

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+ *   - `closeTo` - `{(string|Element|Object)=}`: The query selector, DOM element, or
+ *     the Rect object that is used to determine the bounds (top, left, height, width)
+ *     from which the panel will close to.
+ *   - `hasBackdrop` - `{boolean=}`: Whether there should be an opaque backdrop behind
+ *     the panel. Defaults to false.
+ *   - `escapeToClose` - `{boolean=}`: Whether the user can press escape to close the
+ *     panel. Defaults to false.
+ *   - `clickOutsideToClose` - `{boolean=}`: Whether the user can click outside the
+ *     panel to close it. Defaults to false.
+ *   - `disableParentScroll` - `{boolean=}`: Whether the user can scroll the page
```
jelbourn: We've run into a lot of problems with the behavior before, where the scrolling doesn't occur where the library assumes it will be. I wonder if we can avoid this by making the user have to be more explicit.

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+ *
+ * @returns {$mdPanelRef} panelRef
+ */
+
+
+
+/**
+ * A service that is used for controlling/displaying panels on the screen.
+ *
+ * @param $rootElement
```
jelbourn: Add JsDoc types for params / properties. Though they're not _enforced_, they're still quite helpful.

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+ * @param $injector
+ * @final
+ * @constructor
+ * @ngInject
```
jelbourn: Can be one line:
```
@final @ngInject @constructor
```

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+  };
+
+  this.config_ = this.defaultConfigOptions_;
+  this.injector_ = $injector;
+}
+
+
+MdPanelService.prototype.create = function(opt_config) {
+  var configSettings = opt_config || {};
+  for (var key in configSettings) {
```
jelbourn: Use `angular.forEach`

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+ *     open() action is finished.
+ *   - `onRemoving` - `{function=}`: Callback function used to announce the
+ *     close/hide() action is starting. This allows developers to run custom
+ *     animations in parallel the close animations.
+ *
+ * @param $injector
+ * @param config
+ *
+ */
+function $mdPanelRef($injector, config) {
```
jelbourn: Class name should be `MdPanelRef`

Missing `@constructor`

src/components/panel/panel.js
```diff
@@ -0,0 +1,284 @@
...
+ * @ngInject
+ */
+function MdPanelService($rootElement, $injector) {
+  // Default config options for the panel.
+  this.defaultConfigOptions_ = {
+    parent: $rootElement
+  };
+
+  this.config_ = this.defaultConfigOptions_;
+  this.injector_ = $injector;
```
jelbourn: You can keep the `$` in the property names here :)

jelbourn:
Completed my first pass!
(github doesn't have a draft > send workflow)

src/components/panel/panel.js
gmoothart: +1 for "attachTo"

src/components/panel/panel.js
ErinCoughlan: Oh awesome! I actually had them and then removed them when I noticed they were rarely included in existing code.

@jelbourn For nullability, the indicators were removed when I ran the demo site, so I didn't include them in the actual code. Is this something we can still use? I think it's helpful to have them in the types, but it's also strange to me to include them only in the developer docs, not the public ones.

src/components/panel/panel.js
ErinCoughlan: I think "attachTo" makes more sense. Changed to that (here and design doc).

src/components/panel/panel.js
ErinCoughlan: This works a lot better. There were a lot of edge cases with the option version, which I didn't like.

I added it to the design doc, but am leaving it out of this PR for now. We'll add it when we work on groups.

src/components/panel/panel.js
ErinCoughlan: @gmoothart @KarenParker @DerekLouie I've moved all the public documentation to the top of the file. When we add new public methods, we need to update the docs there, but within the implementation section, we can follow our existing coding style.

src/components/panel/panel.js
jelbourn: Go ahead with makes most sense in the source code, we can deal with the docs generation as a separate issue.

src/components/panel/panel.js
ErinCoughlan: @jelbourn I'm not sure if I understand the change you are suggesting here, if any. This would be making the user be explicit about whether they want scrolling or not.

src/components/panel/panel.js
ErinCoughlan: I like the idea of defining an object like that and that the options relate directly to the css we need to apply.

I defined the $mdPanelPosition in this PR, but left the animations and size for a future PR.

ErinCoughlan:
I believe I have address all of the previous comments (I didn't mark "Done" on all of them since that's a lot of email spam). Please take a look when you get a chance.

src/components/panel/panel.js
jelbourn: The problem is *how / where* to disable scrolling. You could assume `document.body`, or `$rootElement`, or the closest `md-content`... but you're never going to have something that works for every application.

src/components/panel/panel.js
```diff
@@ -0,0 +1,365 @@
...
+
+/**
+ * Creates a panel with the specified options.
+ * @param {!Object=} opt_config Configuration object for the panel.
+ * @returns {!$mdPanelRef}
+ */
+MdPanelService.prototype.create = function(opt_config) {
+  var configSettings = opt_config || {};
+  angular.forEach(configSettings, function(value, key) {
+    if (configSettings.hasOwnProperty(key)) {
```
jelbourn: `angular.forEach` does it's own filtering on `hasOwnProperty`, so you don't have to do it here.

gmoothart:
LGTM

src/components/panel/panel.js
```diff
@@ -0,0 +1,363 @@
...
+ * (function(angular, undefined) {
+ *   ‘use strict’;
+ *
+ *   angular
+ *       .module('demoApp', ['ngMaterial'])
+ *       .controller('DemoDialogController', DialogController);
+ *
+ *   var panelRef;
+ *
+ *   function showPanel($event) { *
```
DerekLouie: nit: there seems to be a straggling asterisk here.

DerekLouie:
LGTM!

ThomasBurleson:
Closed with SHA 1a6b201f5758a075d

PR #6911, opened 2016-01-29
fix(whiteframe): update breakpoints in basic class demo.

DevVersion:
Use the same breakpoints as in attribute demo.
This changes the breakpoints to the current layout breakpoints.
And annotated the media queries with the given breakpoint.


ThomasBurleson:
LGTM. :+1:

DevVersion:
There are now conflicts, should I resolve them?

DevVersion:
Wow, that's so embarrassing. Excuse me that was a strange rebase :smile:

PR #6921, opened 2016-01-29
layoutNowrap: Inconsistency in directive/css implementation

isaaclyman:
Updated layout-no-wrap to layout-nowrap (which is what's in the documentation). Previously, using layout-nowrap-gt-sm didn't work because the directive was layout-no-wrap, whereas the CSS was layout-nowrap (without the second hyphen).


ThomasBurleson:
@isaaclyman - since we already released support for `layout-no-wrap`, we should support both.
LGTM>

PR #6923, opened 2016-01-30
feat(layout, print): add support for `-print` media queries with Layout API

ThomasBurleson:
This adds support for print in the presence of the Layout API: controlling flex and layout attributes when printing.  Previously printing would use your layout for the `-sm` breakpoint. Now you can printing will support attributes like `layout="row"` or `flex-offset-xl="2"`, `hide-print`, `show-print-gt-sm`,  etc.

* added support for Layout API with print
* updated docs to support printing of demos with sideNav
* update build to deploy `angular-material.print.css`: the full Layout API with `print` media queries

thx @epelc. 
refs #6784.


ThomasBurleson:
@robertmesserle, @topherfangio - can you review please.

ThomasBurleson:
Merged by accident. Please continue review.

PR #6927, opened 2016-01-30
fix(autocomplete): stop loading if last promise got resolved

### LINKED ISSUE #6907: mdAutocomplete: md-not-found template flashes before items are displayed
The `md-not-found` template is briefly (as long as half a second) shown before the `md-autocomplete` displays its result set.  

**See this in action here: https://jsfiddle.net/NathanFriend/Lctcxjqr/**

Expected result: the `md-not-found` template should never be shown if the query returns at least one result.  The examples at https://material.angularjs.org/latest/demo/autocomplete don't exhibit this issue.

Some additional information:

Angular Material Version: 1.0.3
Angular Version: 1.4.9
Browser and OS: Chrome 47.0.2526.111 on Windows 8.1

DevVersion:
As said in #6907:
> Just investigated a bit, the problem is, the autocomplete will set the loading state to false if a previous promise is resolved. So when you search for A-R-I-Z-O-N-A (normally typed - not pasted) there will be a promise for each char. So the state of setLoading will change to false. This can be solved by adding a simple fetching queue. But only check the queue on the promise finally to store the retrieved results in cache.

-- Little notice.
> I would like to submit a test for that, but that is complicated for two promises, where the first promise should be flushed (only the first) or resolved. But that will be easy to solve, just resolving the promises manually. But then there will be another problem, because there is another timeout running from the `autoComplete` controller, which validates the promises in the next tick. So flushing and manually resolving won't work, to test this new feature / issue.

Fixes #6907

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -19,7 +19,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       selectedItemWatchers = [],
       hasFocus             = false,
       lastCount            = 0,
-      promiseFetch         = false;
+      fetchesInProgress    = 0;
 
   //-- public variables with handlers
   defineProperty('hidden', handleHiddenChange, true);
@@ -648,15 +648,16 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       if ( !items ) return;
 
       items = $q.when(items);
+      fetchesInProgress++;
       setLoading(true);
-      promiseFetch = true;
 
       $mdUtil.nextTick(function () {
           items
             .then(handleResults)
             .finally(function(){
-              setLoading(false);
-              promiseFetch = false;
+              if (--fetchesInProgress === 0) {
+                setLoading(false);
+              }
             });
       },true, $scope);
     }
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@robertmesserle - please review/test/authorize.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -638,14 +639,17 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
       promiseFetch = true;
       $mdUtil.nextTick(function () {
         if (items.success) items.success(handleResults);
         if (items.then)    items.then(handleResults);
         if (items.finally) items.finally(function () {
-          setLoading(false);
-          promiseFetch = false;
+          if (--fetchesInProgress === 0) {
+            setLoading(false);
+            promiseFetch = false;
```
kseamon: I think it would be cleaner to have only one variable that tracks the fetching state. I'd get rid of promiseFetch and just use fetchesInProgress. If need be, add a getter method that just returns fetchesInProgress === 0

DevVersion: Yes definitely :+1: Just had no chance to look over, because of the rebased 52a519ead04f2faeba227fbc51a7214253cbf2ba commit.

DevVersion: @kseamon Just want to let you know :smile: - Done!

kseamon:
@ThomasBurleson LGTM

DevVersion:
Rebased for 4d59a611e30a6c3d27c4089bc1d1adb27ec3cc6b

PR #6935, opened 2016-01-31
feat(autocomplete): allow select on match to be case insensitive.

### LINKED ISSUE #5782: md-select-on-match is case sensitive
When typing on autocomplete ,
eg: "chennai" the doropdown shows "Chennai" ('C' is in caps) and it is not selecting the value

screenshot from 2015-11-17 20 29 28

DevVersion:
Fixes #5782

### DIFF
--- a/src/components/autocomplete/demoBasicUsage/index.html
+++ b/src/components/autocomplete/demoBasicUsage/index.html
@@ -12,6 +12,7 @@
           md-items="item in ctrl.querySearch(ctrl.searchText)"
           md-item-text="item.display"
           md-min-length="0"
+          md-select-on-match="ctrl.selectOnMatch"
           placeholder="What is your favorite US state?">
         <md-item-template>
           <span md-highlight-text="ctrl.searchText" md-highlight-flags="^i">{{item.display}}</span>
@@ -25,6 +26,7 @@
       <md-checkbox ng-model="ctrl.simulateQuery">Simulate query for results?</md-checkbox>
       <md-checkbox ng-model="ctrl.noCache">Disable caching of queries?</md-checkbox>
       <md-checkbox ng-model="ctrl.isDisabled">Disable the input?</md-checkbox>
+      <md-checkbox ng-model="ctrl.selectOnMatch">Select on Match?</md-checkbox>
 
       <p>By default, <code>md-autocomplete</code> will cache results when performing a query.  After the initial call is performed, it will use the cached results to eliminate unnecessary server requests or lookup logic. This can be disabled above.</p>
     </form>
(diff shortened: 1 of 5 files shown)


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -738,7 +738,12 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         matches    = ctrl.matches,
         item       = matches[ 0 ];
     if (matches.length === 1) getDisplayValue(item).then(function (displayValue) {
-      if (searchText == displayValue) select(0);
+      var isMatching = searchText == displayValue;
+      if ($scope.matchInsensitive && !isMatching) {
+        isMatching = searchText.toLowerCase() == displayValue.toLowerCase();
```
DevVersion: Not using a ternary operator here, because it will be really unclear. 
`isMatching == false` is just saving up an unnecessary step.

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -124,25 +126,26 @@ function MdAutocomplete () {
...
+      inputMinlength:   '@mdInputMinlength',
+      inputMaxlength:   '@mdInputMaxlength',
+      searchText:       '=?mdSearchText',
+      selectedItem:     '=?mdSelectedItem',
+      itemsExpr:        '@mdItems',
+      itemText:         '&mdItemText',
+      placeholder:      '@placeholder',
+      noCache:          '=?mdNoCache',
+      selectOnMatch:    '=?mdSelectOnMatch',
+      matchInsensitive: '=?mdMatchCaseInsensitive',
```
ThomasBurleson: Let's invert the logic here. Better name is `matchCase` which should have default value == `false`.

DevVersion: @ThomasBurleson If understood everything right?
- By default it will ignore case
- Allows caring about case when using `md-match-case="true"`

Updated it to behave as this. See specs for example.

src/components/autocomplete/demoBasicUsage/index.html
```diff
@@ -12,6 +12,7 @@
           md-items="item in ctrl.querySearch(ctrl.searchText)"
           md-item-text="item.display"
           md-min-length="0"
+          md-select-on-match="ctrl.selectOnMatch"
```
DevVersion: @ThomasBurleson Added little demo possibility here. Thoughts?

DevVersion:
@ThomasBurleson is that by intention, that you merged the old version? without the change `md-match-case`?

DevVersion:
Would be great if you give a small statement to this :+1:

gigablox:
CodePen? I can't get this too work.

DevVersion:
@gigablox Just apply the new attribute `md-match-case="false"`.

PR #6955, opened 2016-02-01
docs(select): remove fixed height for select demo

### LINKED ISSUE #6943: md-select partially disabled can't be clicked in some situations
The md-select field can't be clicked in some situations. There is only a small stripe clickable.
An example is also on the demo site. Select -> Options With Async Search
After the first load this select (Assign to user) is not clickable.

DevVersion:
At the moment we are fixing the height, which will caught issues with the select because it provides the content and a margin of `20px 0 26px`, plus the message which will be shown when the user selects one option. 

That mean's that won't fit in, and the clickable area will be shrinked.
Better remove that fixed height and use the margin of `26px`.

Fixes #6943


ThomasBurleson:
@DevVersion  - Did you test this on safari and IE. `layout=column` presents issues on those browsers.

DevVersion:
@ThomasBurleson - Yes works still on Safari (`iOS 9`) and IE. But IE has these scrollbars appearing when the spinners shows up (but that's also on master) so that may be another issue.

Issue #6958, opened 2016-02-01
HEAD(MASTER) in angular material throws error


jayeshanandani:
https://material.angularjs.org/HEAD/demo/autocomplete throws error when clicked on demos. 
`Unknown provider: componentProvider <- component <- DemoCtrl <- DemoCtrl <- DemoCtrl <- DemoCtrl`

DevVersion:
Yes, can reproduce. Investigated, it's caused by 82625cf4edeba84ef38fb816f1b32ded902c8064

DevVersion:
Fixed in #6959.. read more there.

jayeshanandani:
Commit has fixed this. Closing this issue.

PR #6959, opened 2016-02-01
fix(demo): fix demo interpolating at controller startup

DevVersion:
Issue introduced in 82625cf by myself. Sorry for that :-1: 

The actual problem is the `demoInclude` directive, because it's not observing the `module` attribute. 
See [here](https://github.com/angular/material/blob/0412152e9ea76387b8e4f474283ed369e88c65c9/docs/app/js/demoInclude.js#L18-L18). 
So we need to interpolate the module on controller startup to prevent an error, when evaling the module.

#6958

### DIFF
--- a/docs/app/js/demo.js
+++ b/docs/app/js/demo.js
@@ -1,9 +1,7 @@
 DocsApp
 .directive('layoutAlign', function() { return angular.noop; })
 .directive('layout', function() { return angular.noop; })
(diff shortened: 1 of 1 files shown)


DevVersion:
@ThomasBurleson - Please review

DevVersion:
Closed in da53c00874fd227118096ceadc853e8db28c4461

PR #6960, opened 2016-02-01
feat(select): allow to apply an offset to the dropdown

### LINKED ISSUE #6957: Offset for md-select directive
Would be great to have a possibility to set position offset for `md-select ` the same way we can do for `md-menu` with `md-offset` (example: https://github.com/angular/material/blob/f6e97a02a9c88b1695094a8a85a11958924e4c61/src/components/menu/demoMenuPositionModes/index.html#L28).

DevVersion:
Fixes #6957

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -41,6 +41,8 @@ angular.module('material.components.select', [
  * explicit label is present.
  * @param {string=} md-container-class Class list to get applied to the `.md-select-menu-container`
  * element (for custom styling).
+ * @param {string} md-offset An offset to apply to the dropdown after positioning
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
Rejected.

DevVersion:
Is only the PR rejected or in general the issue? :smile:

james-bell:
Rejected and closed with no explanation?  I first checked the MD specs to see if there was an explicit reason but I don't see select controls mentioned specifically, only "menus".  And sometimes the MD menus are positioned to hide their trigger control and sometimes they are not, which is exactly what an offset would allow us to do.

I also have a hard time understanding given that the menu control has an offset.

Issue #5839, opened 2015-11-19
autocomplete select on match behavior should be more intuitive


jrenee42:
right now, autocomplete select on match only autoselects the item if the user types the *entire* text; which is case sensitive.

how I expect/want the interaction to work:

1) case insensitive matching
2) when typing, as soon as the user narrows the field to one item; typing return should add that item  (why make the user type it all out?)

I'm happy to do the coding and submit a pull request for this new interaction.  

could add two attributes: 
case-insensitive-matching = true (false by default; which is the current behavior)
select-on-narrowing-field = true (also false by default; which is the current behavior)

(this feature request/issue also satisfies  issue #5782 )

binzma:
looking forward for this!

ThomasBurleson:
You are encouraged to submit a Pull Request with this enhancement.

jrenee42:
noted!

jrenee42:
pull req here: https://github.com/angular/material/pull/6962

PR #6962, opened 2016-02-01
  proposal for #5839: better auto-complete matching

jrenee42:
added 2 more attributes to md-auto-complete:
    md-case-insensitive-matching
    md-select-incomplete-on-match

    also updated the basic demo to use both of these new attributes


DevVersion:
@jrenee42 The case insensitive is a duplicate of my PR #6935? What should we do?

ThomasBurleson:
This is an API extension that is not core to ngMaterial. Will not merge.

PR #6979, opened 2016-02-02
fix(input): smart icon support should notice other input tags too

### LINKED ISSUE #6977: mdIcon overlaps mdSelect inside mdInputContainer
Hi, given this example http://codepen.io/anon/pen/jWpzyp you can see that mdIcon shows **over** the mdSelect, following the example shown at https://material.angularjs.org/HEAD/demo/input .

mdIcon works as expected with an input, but not with a mdSelect. Adding the class ```md-icon-left``` to the mdInputContainer fixes this, although I don't remember setting this previously so it could be a regression.

Regards.

DevVersion:
Textarea: 

Select:

@EladBezalel Please review.
- Added no tests for left + right aligned icons because then it will be like a duplicate. Should I add them too?

Fixes #6977

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -54,6 +54,9 @@ angular.module('material.components.input', [
  * </hljs>
  */
 function mdInputContainerDirective($mdTheming, $parse) {
+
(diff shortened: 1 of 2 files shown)


EladBezalel:
No need, LGTM

PR #7014, opened 2016-02-03
update(panel): Adds/removes a panel from the DOM.

ErinCoughlan:
@gmoothart @KarenParker @DerekLouie @jelbourn Please review. This adds an element to the DOM on open and removes it on close.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -63,11 +63,11 @@ angular
  *         });
  *   }
  *
- *   function DialogController($MdPanelReference, toppings) {
+ *   function DialogController(MdPanelRef, toppings) {
  *     var toppings;
  *
  *     function closeDialog() {
- *       $MdPanelReference.close();
+ *       MdPanelRef.close();
  *     }
  *   }
  * })(angular);
@@ -129,7 +129,7 @@ angular
  *     Only disable if focusing some other way, as focus management is required for
  *     panels to be accessible. Defaults to true.
  *
- * @returns {$MdPanelReference} panelRef
+ * @returns {MdPanelRef} panelRef
  */
 
 /**
@@ -144,13 +144,13 @@ angular
 
 
 /***************************************************************************************
- *                                   $MdPanelReference                                       *
+ *                                       MdPanelRef                                    *
  ***************************************************************************************/
 
 
 /**
  * @ngdoc object
- * @name $MdPanelReference
+ * @name MdPanelRef
  * @description
  * A reference to a created panel. This reference contains a unique id for the panel,
  * along with the following properties:
@@ -174,7 +174,7 @@ angular
 
 /**
  * @ngdoc method
- * @name $MdPanelReference#open
+ * @name MdPanelRef#open
  * @description
  * If the panel is not visible, opens an already created and configured panel.
  *
@@ -183,10 +183,10 @@ angular
 
 /**
  * @ngdoc method
- * @name $MdPanelReference#close
+ * @name MdPanelRef#close
  * @description
  * If the panel is visible, closes the panel, resolving the promise that is returned
- * from `$MdPanelReference#open`. This method destroys the reference to the panel. In order
+ * from `MdPanelRef#open`. This method destroys the reference to the panel. In order
  * to open the panel again, a new one must be created.
  */
 
@@ -294,16 +294,25 @@ angular
 /**
  * A service that is used for controlling/displaying panels on the screen.
  * @param {!angular.JQLite} $rootElement
+ * @param {!angular.Scope} $rootScope
  * @param {!angular.$injector} $injector
  * @final @constructor @ngInject
  */
-function MdPanelService($rootElement, $injector) {
-  // Default config options for the panel.
+function MdPanelService($rootElement, $rootScope, $injector) {
+  /**
+   * Default config options for the panel.
+   * @private {!Object}
+   */
   this._defaultConfigOptions = {
-    attachTo: $rootElement
+    attachTo: $rootElement,
+    scope: $rootScope.$new(true),
+    bindToController: true
   };
 
+  /** @private {!Object} */
   this._config = this._defaultConfigOptions;
+
+  /** @private {!angular.$injector} */
   this._$injector = $injector;
 }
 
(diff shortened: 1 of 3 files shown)


src/components/panel/panel.js
```diff
@@ -329,21 +332,25 @@ MdPanelService.prototype.create = function(opt_config) {
  * A reference to a created panel. This reference contains a unique id for the panel,
  * along with properties/functions used to control the panel.
  *
- * @param {!angular.$injector} $injector
  * @param {!Object} config
+ * @param {!angular.$injector} $injector
  * @final @constructor
  */
-function MdPanelReference(config,$q) {
+function MdPanelReference(config, $injector) {
```
jelbourn: What do you think of renaming this to `MdPanelRef` ?

src/components/panel/panel.js
```diff
@@ -129,7 +129,7 @@ angular
  *     Only disable if focusing some other way, as focus management is required for
  *     panels to be accessible. Defaults to true.
  *
- * @returns {$MdPanelReference} panelRef
+ * @returns {$mdPanelReference} panelRef
```
jelbourn: Why change this from PascalCase? I think the ideal is just `MdPanelRef` (without the `$`)

src/components/panel/panel.js
```diff
@@ -351,56 +358,65 @@ function MdPanelReference(config,$q) {
...
-    this._openPromise = this._$q(function(resolve, reject) {
-      // Open an instance and animation the instance
-      // ...
-
 
+    var self = this;
+    this._openPromise = this._$q(function(resolve, reject) {
+      self._openReject = reject;
+      self._createPanel();
+      resolve('panel created');
```
jelbourn: Instead of the string, I would pass along the ref instance.

src/components/panel/panel.js
```diff
@@ -351,56 +358,65 @@ function MdPanelReference(config,$q) {
...
     });
-
   }
 
-  return this._openPromise.promise;
+  return this._openPromise;
```
jelbourn: Maybe add a comment like
```
// If the panel is already open, re-use the (already resolved) promise from when it was opened.
```
(same for `close()`)

src/components/panel/panel.js
ErinCoughlan: I initially had it as MdPanelRef. @ThomasBurleson changed it to MdPanelReference. I'm okay with either answer. Thomas?

src/components/panel/panel.js
```diff
@@ -351,56 +358,65 @@ function MdPanelReference(config,$q) {
...
-  } else {
-
     if ( !this._closePromise ) {
+      var self = this;
       this._closePromise = this._$q(function(resolve, reject) {
-
-        // Animate and close the instance
-        // ....
-
+        // Animate and close the instance.
```
jelbourn: Is anything being animated now? I think you just want a 
```
// TODO(ErinCoughlan): kick-off animation
```

src/components/panel/panel.js
```diff
@@ -329,21 +332,25 @@ MdPanelService.prototype.create = function(opt_config) {
...
-  this.id     = 'panel_' + this.id;
+  this.id = config.id;
   this.isOpen = false;
 
   // Private variables.
+  this._config = config;
   this._openPromise;
+  this._openReject;
+  this._closePromise;
+  this._panelEl;
```
jelbourn: Add JsDoc for these properties where it's not completely obvious (e.g., is `panelEl` an `angular.JQLite` or an `Element`?)

src/components/panel/panel.spec.js
```diff
@@ -8,37 +8,82 @@ describe('$mdPanel', function() {
...
+    $rootEl = $injector.get('$rootElement');
   };
 
   beforeEach(function() {
     module('material.components.panel', 'ngSanitize');
 
     inject(injectLocals);
+
+    // By default, the panel is attached to $rootElement, so add it to the DOM.
+    angular.element(document.body).append($rootEl);
```
jelbourn: You need to clean up the DOM in an `afterEach`

src/components/panel/panel.spec.js
```diff
@@ -8,37 +8,82 @@ describe('$mdPanel', function() {
...
   });
 
   it('creates a basic panel', function() {
```
jelbourn: I missed this last time, but specs are traditionally of the for `it('should ...', ...`

src/components/panel/panel.spec.js
```diff
@@ -8,37 +8,82 @@ describe('$mdPanel', function() {
...
+      closeResolved = true;
+    });
     $rootScope.$apply();
 
-    expect(resolved).toBe(true);
+    expect(closeResolved).toBe(true);
     expect(panelRef.isOpen).toEqual(false);
   });
+
+  it('adds a panel to the DOM', function() {
```
jelbourn: add _and_ remove

src/components/panel/panel.spec.js
```diff
@@ -8,37 +8,82 @@ describe('$mdPanel', function() {
...
+      attachTo: angular.element(parentEl),
+      template: template
+    };
+
+    var panelRef = $mdPanel.create(config);
+    panelRef.open();
+    $rootScope.$apply();
+
+    var panelEl = document.querySelector('#panel');
+    expect(panelEl.parentElement).toBe(parentEl);
```
jelbourn: And then remove

src/components/panel/panel.js
```diff
@@ -311,17 +314,17 @@ function MdPanelService($rootElement, $injector) {
...
  */
 MdPanelService.prototype.create = function(opt_config) {
   var configSettings = opt_config || {};
 
   angular.extend(this._config, configSettings);
 
-  var instanceID     = $injector.get('$mdUtil').nextUid();
-  var instanceConfig = angular.extend({ }, this._config, {id : instanceID } );
+  var instanceId = 'panel_' + this._$injector.get('$mdUtil').nextUid();
+  var instanceConfig = angular.extend({ }, this._config, {id : instanceId } );
```
DerekLouie: (nit) Why not: 

var instanceConfig = angular.extend({id : instanceId }, this._config);

src/components/panel/panel.js
ErinCoughlan: @jelbourn When is there a `$` vs just PascalCase. If there's the `$`, I would assume we would use camelCase, such as with `$mdDialog` or `$mdMenu`. If there isn't, then I understand the PascalCase. My distinction was something being injected, which the panelRef would be injected into the controller, but that's also slightly different.

jelbourn: When using a `$`, I think it would be easy to mistake the value for a singleton that you might inject anywhere. I also want to follow the widely-accepted JS standard that constructors / types are PascalCase.

@ThomasBurleson do you have a strong counter to this?

ErinCoughlan:
I have fixed all of the previously mentioned comments.

src/components/panel/panel.js
```diff
@@ -311,96 +320,138 @@ function MdPanelService($rootElement, $injector) {
...
- * @param {!angular.$injector} $injector
  * @param {!Object} config
+ * @param {!angular.$injector} $injector
  * @final @constructor
  */
-function MdPanelReference(config,$q) {
+function MdPanelRef(config, $injector) {
   // Injected variables.
-  this._$q = $q;
+  /** @private {!angular.$q} @const */
```
jelbourn: You can do 
```js
/** @private @const {type} */
```
Which I think is much more readable

src/components/panel/panel.spec.js
```diff
@@ -8,37 +9,105 @@ describe('$mdPanel', function() {
...
+  });
+
+  /**
+   * Attached an element to document.body. Keeps track of attached elements
+   * so that they can be removed in an afterEach.
+   * @param el
+   */
+  function attachToBody(el) {
+    var element = angular.element(el);
+    angular.element(document.body).append(element);
```
jelbourn: You don't have to change anything here, but in general I'm in favor of using native DOM APIs instead of angular's JQLite APIs whenever possible. The JQLite stuff is a remnant of an era when even things like `addEventListener` weren't totally standard; these days the native APIs will be what you need 90% of the time.

jelbourn:
LGTM aside from one last comment on your JsDocs

ErinCoughlan:
Done.

DerekLouie:
LGTM

KarenParker:
LGTM

PR #7015, opened 2016-02-04
fix(autocomplete): fixes the sizing math

### LINKED ISSUE #6212: md-tabs last tab sometimes not visible
- Depending upon the Tab Title, md-tabs will not always render the last tab in the Array when used with ng-repeat.
 - When it gets in this state it also is unable to scroll left, right when the tab window becomes small

See codepen here:
http://codepen.io/scottreisdorf/pen/EPjXzN

#### A Fix using the v1.0.0-rc7 code ####
Using  v1.0.0-rc7 I was able to fix this by changing the updatePagingWidth method to add 1 more px.
See code below adding 1 more px at the end:
```javascript 
 width +=1;
````
Full Method:
```javascript
      function updatePagingWidth() {
        var width = 1;
        angular.forEach(getElements().dummies, function (element) {
          width += Math.ceil(element.offsetWidth);
        });
        width +=1;
        angular.element(elements.paging).css('width', width + 'px');
      }
````

### PR DESCRIPTION
**Pull request by robertmesserle**
closes #6212

@jelbourn @ThomasBurleson

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -545,7 +545,9 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
   function updatePagingWidth() {
     var width = 1;
     angular.forEach(getElements().dummies, function (element) {
-      width += Math.ceil(element.offsetWidth);
+      // uses `getBoundingClientRect().width` rather than `offsetWidth` to include decimal values
+      // when calculating the total width
+      width += Math.ceil(element.getBoundingClientRect().width);
     });
     angular.element(elements.paging).css('width', width + 'px');
   }

### CODE REVIEW

File: src/components/tabs/js/tabsController.js [hunk 1]
```diff
@@ -545,7 +545,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
   function updatePagingWidth() {
     var width = 1;
     angular.forEach(getElements().dummies, function (element) {
-      width += Math.ceil(element.offsetWidth);
+      width += Math.ceil(element.getBoundingClientRect().width);
```
jelbourn: Add a comment saying why you use this specifically?

src/components/tabs/js/tabsController.js
```diff
@@ -545,7 +545,9 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
   function updatePagingWidth() {
     var width = 1;
     angular.forEach(getElements().dummies, function (element) {
-      width += Math.ceil(element.offsetWidth);
+      // uses `getBoundingClientRect().width` rather than `offsetWidth` to include decimal values
+      // when calculating the total width
+      width += Math.ceil(element.getBoundingClientRect().width);
```
ThomasBurleson: Do we not want to round-up to the nearest integer?

ThomasBurleson:
Please assign back to me when ready.

ngraef:
Looks like this supersedes #6737 and also fixes #5439.
And a note for when the changelog gets generated: this is a tabs fix, not autocomplete as the title says.

danielgranat:
This fix worked when the tabs initializes,  showing the 2 tabs i have. But when clicking on the second tab, it gets bigger (by 2px on firefox and 7px on safari), but the calculation is using the dummy which remained the same size.
Any ideas?

FriOne:
@danielgranat Same thing, and to all that last tab is not visible.

PR #5811, opened 2015-11-18
fix(toolbar): apply accent color to ripple and icons in toolbar.

DevVersion:
The toolbar won't apply the accent colors to the ripple and the md-icon.

Fixes #5341

PR #7041, opened 2016-02-05
docs(): overwrite fill color too.

DevVersion:
Since we are setting the fill color for the icons too, we need to overwrite the fill attribute too.


DevVersion:
@ThomasBurleson @EladBezalel This Travis CI, can't be related to my PR, seems there are problems with the CI?

EladBezalel:
Related to #5811

PR #7042, opened 2016-02-05
fix(autocomplete): fix not found template detection when element is hidden

### LINKED ISSUE #7035: md-not-found doesn't work if there is a ng-if on the md-autocomplete
There is some caching behavior on the MDAutoComplete controller which assumes the element() object will be the same in the lifecycle. This makes the check for ctrl.hasNotFound being reported as undefined. seems like this check should be done with no caching in the link() function and not in the template() function.

http://codepen.io/anon/pen/GoYKOb

### LINKED ISSUE #7142: md-not-found won't show when using md-autocomplete in ng-repeat scopes (since 1.0.5)
Took me some time to figure out that it's not me - it's you :)

Since angular-material 1.0.5, using `md-autocomplete` in a repeating element won't show a text when nothing is found. The normal dropdown for regular results seems to work fine.

I created a [codepen](http://codepen.io/anon/pen/PZVGGz) to demonstrate the bug - when you change the material version in the script tag to 1.0.4, `md-not-found` will be rendered properly.

Probably the same bug as in #7035.

DevVersion:
When the element is hidden at compilation time through a **ng-if** or **ng-repeat** directive for example.
The stored variable as currently used will be trashed, so it will be undefined.

So there a possibilities to store that state, using a variable in the directive (ex. hashmaps with ids) or storing it as an attribute / class.

Fixes #7035 Fixes #7142

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -488,6 +488,33 @@ describe('<md-autocomplete>', function() {
       element.remove();
     }));
 
+    it('properly sets hasNotFound when element is hidden through ng-if', inject(function() {
+      var scope = createScope();
+      var template1 =
+        '<div>' +
+          '<md-autocomplete ' +
+              'md-selected-item="selectedItem" ' +
+              'md-search-text="searchText" ' +
+              'md-items="item in match(searchText)" ' +
+              'md-item-text="item.display" ' +
+              'placeholder="placeholder" ' +
+              'ng-if="showAutocomplete">' +
+            '<md-item-template>{{item.display}}</md-item-template>' +
+            '<md-not-found>Sorry, not found...</md-not-found>' +
+          '</md-autocomplete>' +
+        '</div>';
+      var element = compile(template1, scope);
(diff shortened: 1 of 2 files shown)


topherfangio:
Rerunning the build to see if we can merge this as the the build failure doesn't appear related to the code changes.

DevVersion:
Thanks @topherfangio!

DevVersion:
@topherfangio Need to rebase this PR due the PhantomJS2 changes.

topherfangio:
@DevVersion Did you already do the rebase? It looks like the tests are passing again. If so, I'll make it as needs merge. Thanks!

DevVersion:
@topherfangio Yes, already rebased. Thanks for tracking :+1:

topherfangio:
@ThomasBurleson LGTM, should be a simple non-breaking change.

cleever:
thanks @DevVersion.

bluebirdtech:
thanks @DevVersion fixes it nicely for me

ThomasBurleson:
@DevVersion - :+1:

Issue #6941, opened 2016-01-31
Datepicker demo: ngMessage color is slightly off-spec


MariskaS:
Will the validation have been fixed datepicker ?
(color validation and transition for error)

![input](https://cloud.githubusercontent.com/assets/5942101/12701846/72041bdc-c828-11e5-9e7c-c6162ecc3ce0.png)

DevVersion:
The color of the ngMessage is set manually in the demo. See [here](https://github.com/angular/material/blob/c2e17ad6ce0a0f16b2169986adb278083ad22eb8/src/components/datepicker/demoBasicUsage/style.css#L8-L8). @jelbourn Is that by intention?

jelbourn:
I just made it some arbitrary red.

MariskaS:
Thanks for the answer. More interested in the appearance of the animation block error. currently the field is different and has to build upon css

DevVersion:
@MariskaS Can you provide a codepen from your `aniamtion block` version? Currently the datepicker doesn't support its own ngMessages animations.

MariskaS:
/***
 * Exammple c-datepicker:
 *  <div layout-gt-sm="row" class="c-datepicker_container">
 *      <label class="c-datepicker_container__label">Available</label>
 *      <div class="c-datepicker_container__date">
 *          <md-datepicker name="availableFrom" ng-model="vm.houseType.availableFrom" ng-required="true"></md-datepicker>
 *          <div class="c-validation-messages" ng-messages="editForm.availableFrom.$error">
 *              <div ng-message="valid">Please enter valid date</div>
 *          </div>
 *      </div>
 *      <div hide-xs="" hide-sm="" class="c-datepicker_container__icon">
 *          <label>&nbsp;&mdash;&nbsp;</label>
 *      </div>
 *      <div class="c-datepicker_container__date">
 *          <md-datepicker name="availableTo" ng-model="vm.houseType.availableTo"></md-datepicker>
 *          <div class="c-validation-messages" ng-messages="editForm.availableTo.$error">
 *              <div ng-message="valid">Please enter valid date</div>
 *          </div>
 *      </div>
 *      <div class="c-datepicker_container__icon">
 *          <md-icon md-svg-icon="icon-warning" ng-show="vm.isExpired()">
 *              <md-tooltip>House type is expired</md-tooltip>
 *          </md-icon>
 *      </div>
 *  </div>
 */

Less
`.c-datepicker_container {
    display: block;
    position: relative;
[...]

PR #7071, opened 2016-02-08
fix(datepicker): style demo ngMessage as same as input messages.

DevVersion:
At the moment there was just some randomly picked color for the ngMessage in the datepicker demo.
We should use a color which looks as same as the actual input ngMessages (to have a flat ngMessages design),
and even the font-size should be the same.

References #6941

### DIFF
--- a/src/components/datepicker/demoBasicUsage/style.css
+++ b/src/components/datepicker/demoBasicUsage/style.css
@@ -4,7 +4,7 @@ md-content {
 }
 
 .validation-messages {
-  font-size: 11px;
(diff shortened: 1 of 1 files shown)


DevVersion:
@EladBezalel Can you take a look at this PR when you got time? :)

EladBezalel:
Please take a look the the CI

DevVersion:
@EladBezalel Yes you're right, that's why this PR needs a rebase - I guess at the time I submitted the PR, the CI had some errors.

So will do that asap.

PR #7076, opened 2016-02-08
fix(tabs): fixes background flicker in iOS Safari

### LINKED ISSUE #6290: md-tabs(static-tabs): Tabs and content becoming invisible on ios safari
The static tabs demo(and if you use the method in your own app) of rendering content outside of the tabs is very buggy on ios safari(I couldn't reproduce on os x but it always shows on ios). The tabs partially become invisible and the tab content totally disappears. I'm not sure why and it has been like this for a while.

See the `static tabs` demo here at the very bottom (here)[https://material.angularjs.org/latest/demo/tabs] on ios using safari(I'm not sure if it affects ipads but it definitely affects iphones).

This is what it looks like sometimes when loading but most of the time after you switch tabs(note you may have to switch multiple times to see it). It requires a refresh of the page once this happens no matter what.
img_0262

Tested on an iphone 6s running ios 9.2
Note it does not show up in os x safari

robertmesserle:
closes #6290

@ThomasBurleson @jelbourn

### DIFF
--- a/src/components/tabs/tabs.scss
+++ b/src/components/tabs/tabs.scss
@@ -75,6 +75,8 @@ md-tabs {
 md-tabs-wrapper {
   display: block;
   position: relative;
+  // transform is needed for iOS Safari to prevent content from disappearing on scroll
+  transform: translate3d(0, 0, 0);
   md-prev-button, md-next-button {
     height: 100%;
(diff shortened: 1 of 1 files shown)


jelbourn:
LGTM

PR #7088, opened 2016-02-09
fix(checkbox): fix IE11 focus for checkbox

### LINKED ISSUE #7086: IE11 md-checkbox ng-model-options not working with blur
If I add ng-model-options="{updateOn:'blur'}"  to a md-checkbox it is not updating the ng-model after i click on something else.  It works fine in firefox and chrome, but not in IE11.
You can reproduce this easly using the codepen that can be reached at https://material.angularjs.org/latest/demo/checkbox the one in basic usage, just add ng-model-options="{updateOn:'blur'}" to the first checkbox.

DevVersion:
At the moment IE11 is always selecting the div.md-container, but that will cause issues with the ngModel (options).
- So we need to block pointer events for the container, because we will manage all from actual `<md-checkbox>`

Fixes #7086

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -59,6 +59,10 @@ md-checkbox {
     @include rtl(left, 0, auto);
     @include rtl(right, auto, 0);
 
+    // Disable pointer events, because IE11 is always focusing the child elements instead of the
+    // md-checkbox element.
+    pointer-events: none;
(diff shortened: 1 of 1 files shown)


DevVersion:
@EladBezalel Can you review?

EladBezalel:
LGTM

ThomasBurleson:
:+1:

PR #7089, opened 2016-02-09
fix(autocomplete): only handle results if it's an array or a promise

### LINKED ISSUE #7074: md-autocomplete binding to md-items
Binding md-autocompletes md-items to an empty array or object, then clicking into md-autocomplete will cause md-autocomplete to not function property even after md-items are populated until a reload.

http://codepen.io/anon/pen/Gowjmv?editors=1010

DevVersion:
At the moment the autocomplete will handle the results wrong.
So if we specify for example an empty JSON-Object, the autocomplete will handle the results as an async promise.

Fixes #7074

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -635,10 +635,11 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   function fetchResults (searchText) {
     var items = $scope.$parent.$eval(itemExpr),
         term  = searchText.toLowerCase(),
-        isList = angular.isArray(items);
+        isList = angular.isArray(items),
+        isPromise = !!items.then; // Every promise should contain a `then` property
 
-    if ( isList ) handleResults(items);
-    else          handleAsyncResults(items);
+    if (isList) handleResults(items);
+    else if (isPromise) handleAsyncResults(items);
 
     function handleAsyncResults(items) {
       if ( !items ) return;
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -206,6 +206,36 @@ describe('<md-autocomplete>', function() {
       element.remove();
     }));
 
+    it('should not show a loading progress when the items object is invalid', inject(function() {
+      var scope = createScope(null, {
+        match: function() {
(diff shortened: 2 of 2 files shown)


EladBezalel:
LGTM

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -635,10 +635,11 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
     var items = $scope.$parent.$eval(itemExpr),
         term  = searchText.toLowerCase(),
-        isList = angular.isArray(items);
+        isList = angular.isArray(items),
+        isPromise = !!items.then; // Every promise should contain a `then` property
 
-    if ( isList ) handleResults(items);
-    else          handleAsyncResults(items);
+    if (isList) handleResults(items);
+    else if (isPromise) handleAsyncResults(items);
```
ThomasBurleson: Messing scenario:  what if **not** a list and **not** a Promise?

DevVersion: I thought about doing nothing, because it makes no sense, and otherwise it will prevent the autocomplete from working. See the demo of the actual issue. http://codepen.io/anon/pen/Gowjmv?editors=1010

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -635,10 +635,11 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   function fetchResults (searchText) {
     var items = $scope.$parent.$eval(itemExpr),
         term  = searchText.toLowerCase(),
-        isList = angular.isArray(items);
+        isList = angular.isArray(items),
+        isPromise = !!items.then; // Every promise should contain a `then` property
```
crisbeto: This is PR is kinda old, but just to throw my 2 cents: Angular has an internal function, called `isPromiseLike` that checks whether an object is a promise like this: `obj && isFunction(obj.then)`, maybe we could use the same?

ThomasBurleson: @DevVersion - Can you use `isPromiseLike( )` also?

crisbeto: It's not a public function but the check isn't that large.

PR #7090, opened 2016-02-09
feat(dialog): add default value to prompt preset.

### LINKED ISSUE #7046: $mdDialog.prompt should have a defaultValue option
I missed this when I made this config...

To be a full replacement for window.prompt, we should be able to provide a default value.

https://developer.mozilla.org/en-US/docs/Web/API/Window/prompt

DevVersion:
Fixes #7046

### DIFF
--- a/src/components/dialog/demoBasicUsage/script.js
+++ b/src/components/dialog/demoBasicUsage/script.js
@@ -40,13 +40,14 @@ angular.module('dialogDemo1', ['ngMaterial'])
   $scope.showPrompt = function(ev) {
     // Appending dialog to document.body to cover sidenav in docs app
     var confirm = $mdDialog.prompt()
-          .title('What would you name your dog?')
-          .textContent('Bowser is a common name.')
-          .placeholder('dog name')
-          .ariaLabel('Dog name')
-          .targetEvent(ev)
-          .ok('Okay!')
-          .cancel('I\'m a cat person');
+      .title('What would you name your dog?')
+      .textContent('Bowser is a common name.')
+      .placeholder('Dog name')
+      .ariaLabel('Dog name')
+      .initialValue('Buddy')
+      .targetEvent(ev)
+      .ok('Okay!')
+      .cancel('I\'m a cat person');
 
     $mdDialog.show(confirm).then(function(result) {
       $scope.status = 'You decided to name your dog ' + result + '.';
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -421,6 +421,7 @@ function MdDialogDirective($$rAF, $mdTheming, $mdDialog) {
  * - $mdDialogPreset#htmlContent(string) - Sets the prompt message as HTML. Requires ngSanitize
  *     module to be loaded. HTML is not run through Angular's compiler.
  * - $mdDialogPreset#placeholder(string) - Sets the placeholder text for the input.
+ * - $mdDialogPreset#initialValue(string) - Sets the initial value for the prompt input.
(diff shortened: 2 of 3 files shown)


src/components/dialog/demoBasicUsage/script.js
```diff
@@ -40,13 +40,13 @@ angular.module('dialogDemo1', ['ngMaterial'])
...
     // Appending dialog to document.body to cover sidenav in docs app
     var confirm = $mdDialog.prompt()
-          .title('What would you name your dog?')
-          .textContent('Bowser is a common name.')
-          .placeholder('dog name')
-          .ariaLabel('Dog name')
-          .targetEvent(ev)
-          .ok('Okay!')
-          .cancel('I\'m a cat person');
+      .title('What would you name your dog?')
```
DevVersion: Changed indention from `tab` to `2-spaces`

src/components/dialog/dialog.js
```diff
@@ -536,8 +538,15 @@ function MdDialogProvider($$interimElementProvider) {
         '</md-dialog>'
       ].join('').replace(/\s\s+/g, ''),
       controller: function mdDialogCtrl() {
+        var isPrompt = this.$type === 'prompt';
+
+        if (isPrompt && this.defaultValue) {
+          // Initialize the prompt result with the specified default value.
```
DevVersion: Not sure about relevance of this comment :P

src/components/dialog/dialog.js
```diff
@@ -521,7 +522,8 @@ function MdDialogProvider($$interimElementProvider) {
         '      <p>{{::dialog.mdTextContent}}</p>',
         '    </div>',
         '    <md-input-container md-no-float ng-if="::dialog.$type == \'prompt\'" class="md-prompt-input-container">',
```
zbjornson: @DevVersion should this be changed to `===` instead of `==` while you're in the area?

PR looks good, thanks for doing this.

DevVersion: @zbjornson - I don't think this is really relevant, but I will change it :+1:

src/components/dialog/dialog.js
```diff
@@ -383,6 +383,7 @@ function MdDialogDirective($$rAF, $mdTheming, $mdDialog) {
  * - $mdDialogPreset#title(string) - Sets the prompt title.
  * - $mdDialogPreset#textContent(string) - Sets the prompt message.
  * - $mdDialogPreset#htmlContent(string) - Sets the prompt message as HTML. Requires ngSanitize
+ * - $mdDialogPreset#defaultValue(string) - Sets the default value for the prompt input.
```
EladBezalel: wrong position, you accidentally spliced the line above

DevVersion: Fixed, thank you :+1:

src/components/dialog/dialog.js
```diff
@@ -503,7 +504,7 @@ function MdDialogProvider($$interimElementProvider) {
       options: advancedDialogOptions
     })
     .addPreset('prompt', {
-      methods: ['title', 'htmlContent', 'textContent', 'content', 'placeholder', 'ariaLabel',
+      methods: ['title', 'htmlContent', 'textContent', 'defaultValue', 'content', 'placeholder', 'ariaLabel',
```
ThomasBurleson: I think `initialValue` is a better name.

DevVersion: I respected the `window.prompt` reference here.
- https://developer.mozilla.org/de/docs/Web/API/window/prompt

DevVersion: @ThomasBurleson Finally updated. Sorry for the delay ;)

PR #7091, opened 2016-02-09
update(panel): Adds tests for component logic.

ErinCoughlan:
@gmoothart @KarenParker @DerekLouie @jelbourn Please review.

These test verify that we can create a panel with a controller and the associated logic.

### DIFF
--- a/src/components/panel/panel.spec.js
+++ b/src/components/panel/panel.spec.js
@@ -1,5 +1,6 @@
 describe('$mdPanel', function() {
-  var $mdPanel, $rootScope, $rootEl;
+  var $mdPanel, $rootScope, $rootEl, $templateCache, $q;
+  var panelRef;
   var attachedElements = [];
 
   /**
@@ -10,6 +11,8 @@ describe('$mdPanel', function() {
     $mdPanel = $injector.get('$mdPanel');
     $rootScope = $injector.get('$rootScope');
     $rootEl = $injector.get('$rootElement');
+    $templateCache = $injector.get('$templateCache');
+    $q = $injector.get('$q');
   };
 
   beforeEach(function() {
@@ -26,6 +29,8 @@ describe('$mdPanel', function() {
       el.remove();
     });
     attachedElements = [];
+
+    panelRef && panelRef.close();
   });
 
   it('should create a basic panel', function() {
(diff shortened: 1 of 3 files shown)


src/components/panel/panel.spec.js
```diff
@@ -117,6 +120,110 @@ describe('$mdPanel', function() {
     expect(parentEl.childElementCount).toEqual(0);
   });
 
+  describe('component logic: ', function() {
+    it('should allow templateUrl to specify content',
+        inject(function($templateCache) {
```
jelbourn: Rather than inject `$templateCache` for this one test, can you move it to a file-level var? I prefer keeping `inject` out of all individual tests so that later, when someone else adds a test, you don't end up with 30 different tests that all` inject` the same thing.

src/components/panel/panel.spec.js
```diff
@@ -117,6 +120,110 @@ describe('$mdPanel', function() {
...
+          var htmlContent = 'Hello World';
```
jelbourn: For each of these tests that use `htmlContent`, could you use a different string? This will protect against the tests accidentally passing because some DOM didn't get cleaned up from a previous test.

src/components/panel/panel.spec.js
```diff
@@ -117,6 +120,110 @@ describe('$mdPanel', function() {
...
+
+      panelRef = $mdPanel.open(config);
+      $rootScope.$apply();
+
+      expect('#panel').toContainHtml(htmlContent);
+    });
+
+    it('should bind resolves to the controller', function() {
+      var htmlContent = 'Hello World';
+      var template = '<div id="panel">{{ ctrl.content }}</div>';
```
jelbourn: Should also have a test where the value isn't immediately resolved, and thus assert that, when calling `$apply` _before_ the promise is resolved, the element isn't in the DOM yet.

src/components/panel/panel.spec.js
```diff
@@ -117,6 +120,110 @@ describe('$mdPanel', function() {
...
+      expect('#panel').toContainHtml(htmlContent);
+    });
+
+    it('should bind locals to the controller', function() {
+      var htmlContent = 'Hello World';
+      var template = '<div id="panel">{{ ctrl.content }}</div>';
+
+      var config = {
+        template: template,
+        controller: function Ctrl() {
```
jelbourn: Should there also be a test where the controller injects (via constructor) something from locals?

ErinCoughlan:
Updated the CL with all the comments. PTAL.

KarenParker:
LGTM

jelbourn:
LGTM!

PR #7107, opened 2016-02-10
feat(autocomplete): support readonly attribute

### LINKED ISSUE #5507: md-autocomplete and ng-readonly
Is it possible to use ng-readonly with md-autocomplete?
I can still change the value in md-autocomplete with ng-readonly="true"

DevVersion:
@robertmesserle Please review :)

Fixes #5507

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -36,6 +36,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   ctrl.id         = $mdUtil.nextUid();
   ctrl.isDisabled = null;
   ctrl.isRequired = null;
+  ctrl.isReadonly = null;
   ctrl.hasNotFound = false;
 
   //-- public methods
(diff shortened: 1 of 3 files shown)


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -166,6 +167,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     var wait = parseInt($scope.delay, 10) || 0;
     $attrs.$observe('disabled', function (value) { ctrl.isDisabled = !!value; });
     $attrs.$observe('required', function (value) { ctrl.isRequired = !!value; });
+    $attrs.$observe('readonly', function (value) { ctrl.isReadonly = !!value; });
```
DevVersion: This attribute needs to be fixed too (heopfully using the util method of #6803)

robertmesserle: Could we use `ctrl.isReadonly = value !== null` for now?  I think that should work.

DevVersion: @robertmesserle Done - re-enabled test. I will rebase #6803 if this PR is merged.

robertmesserle: Yea, once #6803 gets merged, these will all use your utility function, right?

DevVersion: Yes, then the attributes will also work with an empty attribute value. And a little side-effect will be, that `required="false"` also works. This won't break any existing versions, because it just adds `false` and `0` as a negative value.

ThomasBurleson:
@DevVersion - please rebase with `master`. I think another PR (readonly validations) conflicts with this one.

DevVersion:
@ThomasBurleson Done - CI failing till #7307 is merged.

PR #7114, opened 2016-02-10
update(panel): Sets up SCSS for the panel.

ErinCoughlan:
Also wraps the user-provided template in an md-panel-container, which will allow us to control positioning, animations, and similar panel properties. I came up with a few ways to implement this, so if you have a better solution, let me know.

@ThomasBurleson @jelbourn @gmoothart @KarenParker @DerekLouie Please review.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -9,9 +9,9 @@ angular
     .service('$mdPanel', MdPanelService);
 
 
-/***************************************************************************************
- *                              PUBLIC DOCUMENTATION                                   *
- ***************************************************************************************/
+/*****************************************************************************
+ *                            PUBLIC DOCUMENTATION                           *
+ *****************************************************************************/
 
 /**
  * @ngdoc service
@@ -109,7 +109,6 @@ angular
  *     used for configuring the number of open panels and identifying specific
  *     behaviors for groups. For instance, all tooltips will be identified
  *     using the same groupName.
- *
  *   - `position` - `{$mdPanelPosition=}`: An $mdPanelPosition object that
  *     specifies the alignment of the panel. For more information, see
  *     `$mdPanelPosition`.
@@ -168,9 +167,9 @@ angular
  */
 
 
-/***************************************************************************************
- *                                       MdPanelRef                                    *
- ***************************************************************************************/
+/*****************************************************************************
+ *                                 MdPanelRef                                *
+ *****************************************************************************/
 
 
 /**
@@ -218,9 +217,9 @@ angular
  */
 
 
-/***************************************************************************************
- *                                 $mdPanelPosition                                    *
- ***************************************************************************************/
+/*****************************************************************************
+ *                               $mdPanelPosition                            *
+ *****************************************************************************/
 
 
 /**
@@ -243,9 +242,9 @@ angular
  * @ngdoc method
  * @name $mdPanelPosition#absolute
  * @description
- * Positions the panel absolutely relative to the parent element. If the parent is
- * document.body, this is equivalent to positioning the panel absolutely within the
- * viewport.
+ * Positions the panel absolutely relative to the parent element. If the parent
+ * is document.body, this is equivalent to positioning the panel absolutely
+ * within the viewport.
  * @returns {$mdPanelPosition}
  */
 
@@ -254,7 +253,8 @@ angular
  * @name $mdPanelPosition#relativeTo
  * @description
  * Positions the panel relative to a specific element.
- * @param {!angular.JQLite} element Element to position the panel with respect to.
+ * @param {!angular.JQLite} element Element to position the panel with
+ *     respect to.
  * @returns {$mdPanelPosition}
  */
 
@@ -313,9 +313,9 @@ angular
  */
 
 
-/***************************************************************************************
- *                                   IMPLEMENTATION                                    *
- ***************************************************************************************/
+/*****************************************************************************
+ *                                 IMPLEMENTATION                            *
+ *****************************************************************************/
 
 
 /**
@@ -332,8 +332,9 @@ function MdPanelService($rootElement, $rootScope, $injector) {
    */
   this._defaultConfigOptions = {
     attachTo: $rootElement,
+    bindToController: true,
     scope: $rootScope.$new(true),
-    bindToController: true
+    transformTemplate: angular.bind(this, this.wrapTemplate_)
   };
 
   /** @private {!Object} */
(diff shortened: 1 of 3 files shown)


DerekLouie:
LGTM

src/components/panel/panel.js
```diff
@@ -374,6 +378,25 @@ MdPanelService.prototype.open = function(opt_config) {
 
 
 /**
+ * Adds a surrounding <div> to the user-provided template, which allows
```
gmoothart: Not just a div; adda sentence about auto wrap?

src/components/panel/panel.js
```diff
@@ -103,13 +103,14 @@ angular
  *     values. The panel will not open until all of the promises resolve.
  *   - `attachTo` - `{Element=}`: The element to attach the panel to. Defaults
  *     to appending to the root element of the application.
+ *   - `autoWrap` - `{boolean=}`: Whether to automatically wrap the template
```
gmoothart: Why make this an option? What if we just always add <md-panel>? And if we do, do we also need to wrap that in a containing div?

src/components/panel/panel.js
```diff
@@ -374,6 +378,25 @@ MdPanelService.prototype.open = function(opt_config) {
...
 /**
+ * Adds a surrounding <div> to the user-provided template, which allows
+ * control over positioning, animations, and similar properties.
+ * @param {string} template The template to verify.
+ * @param {!Object} config The config object passed to the compiler.
+ * @returns {string} The transformed template.
+ * @private
+ */
+MdPanelService.prototype.transformTemplate_ = function(template, config) {
+  template = template || '';
```
gmoothart: input params shouldn't be mutated IMO; can you use a new local var for this?

src/components/panel/panel.spec.js
```diff
@@ -63,69 +64,121 @@ describe('$mdPanel', function() {
...
+        template: template,
+        autoWrap: true
+      };
+
+      panelRef = $mdPanel.create(config);
+
+      expect('md-panel').not.toExist();
+      expect('#panel').not.toExist();
+
+      panelRef.open();
```
gmoothart: Add helper methods for open/apply and close/apply patterns?

src/components/panel/panel.spec.js
```diff
@@ -63,69 +64,121 @@ describe('$mdPanel', function() {
...
+      $rootScope.$apply();
+
+      expect('md-panel').toExist();
+      expect('#panel').toExist();
 
-    expect(parentEl.childElementCount).toEqual(0);
+      panelRef.close();
```
gmoothart: 2 ideas:
- do `panelRef.close()` in an afterEach, so you don't have to include it in tests that don't care.
- extract these last lines into a new test, `it should remove panel from the dom on close` or somesuch, and remove from these two tests.

src/components/panel/panel.js
ErinCoughlan: 1. I debated adding it as an option. The reason I went with it is because it allows the user to create a completely custom component. Doing so however would mess with animations and positioning, so I would be okay with removing it. What do you think?

2. We need to be able to control two elements to get positioning correct. The first element is absolutely positioned and will cover the entire width/height of the attachTo element. The second element is the actual panel, which will be positioned according to the mdPanelPosition object. The panel is what animates, and animations would be relative to the attachTo element as well.

src/components/panel/panel.spec.js
ErinCoughlan: I already have panelRef.close() in the afterEach, so I removed them from these tests. (I already have a test that it opened and closed correctly.)

src/components/panel/panel.js
jelbourn: I think it's okay to always have an `.md-panel` element.

src/components/panel/panel.js
```diff
@@ -374,6 +378,25 @@ MdPanelService.prototype.open = function(opt_config) {
...
 
 /**
+ * Adds a surrounding <div> to the user-provided template, which allows
+ * control over positioning, animations, and similar properties.
+ * @param {string} template The template to verify.
+ * @param {!Object} config The config object passed to the compiler.
+ * @returns {string} The transformed template.
+ * @private
+ */
+MdPanelService.prototype.transformTemplate_ = function(template, config) {
```
jelbourn: I would call the method `wrapTemplate`, since `transform` could be anything.

src/components/panel/panel.js
```diff
@@ -374,6 +378,25 @@ MdPanelService.prototype.open = function(opt_config) {
...
+  template = template || '';
+
+  if (config.autoWrap && !/<\/md-panel>/g.test(template)) {
+    template = '<md-panel>' + template + '</md-panel>';
```
jelbourn: Add comment explaining why you need both elements.

src/components/panel/panel.scss
```diff
@@ -0,0 +1,20 @@
+.md-panel-container {
+  align-items: center;
+  display: flex;
```
jelbourn: The `.md-panel-container` going to be the thing that holds the arbitrary `template`, right? If so, I wouldn't make it `display:flex`; they layout of the panel's content should be determined by the template being loaded.

ErinCoughlan: I was thinking the other way around, like this:

```
<div class="md-panel-container">
  <div class="md-panel">
    {{ template }}
  </div>
</div>
```

I removed the display:flex from the panel element. Do you think it makes more sense to swap those two classes?

jelbourn: How about:
```html
<div class="md-panel-outer-wrapper">
  <div class="md-panel"> ... </div>
</div>
```
A bit more specific.

jelbourn:
LGTM

KarenParker:
LGTM

src/components/panel/panel.js
```diff
@@ -332,8 +332,9 @@ function MdPanelService($rootElement, $rootScope, $injector) {
    */
   this._defaultConfigOptions = {
     attachTo: $rootElement,
+    bindToController: true,
     scope: $rootScope.$new(true),
-    bindToController: true
+    transformTemplate: angular.bind(this, this.wrapTemplate_)
```
gmoothart: rename this `transformTemplate` to `wrapTemplate` too?

ErinCoughlan: I can't change that one. transformTemplate is something mdCompiler calls once it has the template loaded.

gmoothart:
:+1:

PR #7119, opened 2016-02-10
feat(autocomplete): allow disabling asterisk on floating label

DevVersion:
- This allows the user to disable the asterisk on the child input of the `autocomplete`.


ThomasBurleson:
Why do we have this PR. What issue is it resolving?

Issue #3815, opened 2015-07-19
$mdMedia as a directive wrapper


wlingke:
There are many cases where I'd like to specify a certain class based on the screen size. It would be super convenient if we could do something like:

`<div md-class-sm="small-class" md-class-gt-sm="gt-sm-class">stuff</div>`

And yes, I know that I can do this with ng-class or write my own directive, but I thought it was a common enough use case that it should be supported out of the box.

ThomasBurleson:
A Pull Request for this feature would be welcomed.

STOpandthink:
I really wanted this too, but then realized that in most cases you'll have multiple identical checks for "gt-sm", for example, but really you should have a variable, e.g. isSingleColumn, that you set, and then check that variable. It actually leads to better code and readability.
But yes, for prototype development, this would be nice to have.

ThomasBurleson:
Thx to @DevVersion for reminding me this feature was already available. See [CodePen](http://codepen.io/team/AngularMaterial/pen/YweVWZ)

```html
<md-card ng-repeat="it in products.catalog" flex="22" flex-sm="46" flex-xs="80" ng-class="{shaded:$mdMedia('xs') }">
```

```js
angular.controller('ProductController', function( $scope, $products, $mdMedia) {
    $scope.$mdMedia           = $mdMedia;
}
```

PR #7121, opened 2016-02-11
feat(media): add media directive as $mdMedia wrapper for classes

DevVersion:
This directive allows the user to add specific classes on specific screen-sizes / breakpoints.

**Notes**
- Tests are currently manually (not dynamic using a loop - I can change that if desired)
- mdMedia directive is currently listed under directives / components (Is that the right way?) 

@EladBezalel @ThomasBurleson - Please correct me if the directive should be located where-else (and about the dynamic tests)

PS: Work in progress :+1:

### DIFF
--- a/src/components/media/media.js
+++ b/src/components/media/media.js
@@ -1,0 +1,141 @@
+(function() {
+
+  var PREFIX_REGEXP = /^((?:x|data)[\:\-_])/i;
+  var SPECIAL_CHARS_REGEXP = /([\:\-\_]+(.))/g;
+  var DIRECTIVE_PREFIX = "md-media-";
+  var BREAKPOINTS = ["xs", "gt-xs", "sm", "gt-sm", "md", "gt-md", "lg", "gt-lg", "xl", "print"];
+
+  /**
+   * @ngdoc module
+   * @name material.components.media
+   * @description
+   * Media module
+   */
+  registerMediaDirectives(angular.module('material.components.media', ['material.core']));
+
+  function registerMediaDirectives(module) {
+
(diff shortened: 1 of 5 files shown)


src/components/media/media.js
```diff
@@ -0,0 +1,141 @@
...
+      }
+    }];
+  }
+
+  /**
+   * Converts snake_case to camelCase.
+   * Also there is special case for Moz prefix starting with upper case letter.
+   * @param name Name to normalize
+   */
+  function directiveNormalize(name) {
```
DevVersion: And I suggest this method to be added to `$mdUtil`

ThomasBurleson:
This is the approach we want to pursue with the Layout API. 

PleaseCurrently we support breakpoint triggers for classes using this:

```html
<div ng-class="{$mdMedia('gt-sm'):'blue'}"`> 
    Testing breakpoint changes to trigger class changes
</div>
```

DevVersion:
@ThomasBurleson So whats now with #3815?

ThomasBurleson:
@DevVersion - thanks for the reminder. Let me update that issue. To be honest I had forgotten that this feature already exists. Your PR made me realize the above solution:

http://codepen.io/team/AngularMaterial/pen/YweVWZ

```html
<md-card ng-repeat="it in products.catalog" flex="22" flex-sm="46" flex-xs="80" ng-class="{shaded:$mdMedia('xs') }">
```

```js
angular.controller('ProductController', function( $scope, $products, $mdMedia) {
    $scope.$mdMedia           = $mdMedia;
}
```

PR #7132, opened 2016-02-12
chore(doc-gen): enable `type` docs to be accessed

petebacondarwin:


src/components/panel/panel.js
```diff
@@ -173,8 +173,9 @@ angular
 
 
 /**
- * @ngdoc object
+ * @ngdoc type
```
jelbourn: Is the end result here to put a whole new section for Types in the sidenav, or just add this these docs to the `$mdPane` page? (the former would be desirable)

petebacondarwin:
Well, this PR creates a new section, but we could create a processor that connected the type to the `$mdPanel` page

docs/app/js/app.js
```diff
@@ -286,6 +286,10 @@ function(SERVICES, COMPONENTS, DEMOS, PAGES, $location, $rootScope, $http, $wind
       pages: apiDocs.service.sort(sortByName),
       type: 'toggle'
     },{
+      name: 'Types',
```
ThomasBurleson: @petebacondarwin - IMO, we do not want a new 'Types' section (that will be shown in the Sidenav). What were your thoughts here ?

petebacondarwin: If you don't want the types section then you could do one of a few things...

a) just bundle types and services into the same section (where everything gets it own page)
b) manually move the information about the return type into the service doc
c) build a dgeni processor that combines the service and type into the same doc

ThomasBurleson:
This impacts the Panel component which has not yet been merged into `master`; I will merge this PR after the mdPanel simultaneously with the mdPanel/

ErinCoughlan:
@ThomasBurleson I can merge this PR with the panel branch today, so we can make sure the docs are right before they get merged to master. Then it's not extra work to merge this.

jelbourn:
@ErinCoughlan I am in favor of that

PR #7146, opened 2016-02-12
fix(input): ngMessages height changing.

### LINKED ISSUE #7072: ng-messages pushing the others elements down
Is this the desired behavior?
if not, the css is lacking of a
`
md-input-container [ng-messages] {
    float: left;
}
`

topherfangio:
When using `ng-messages` inside of an input container, the height
of the messages will change when appearing, which pushes other
elements out of the way. This was a regression from the previous
behavior which allowed space for a one-line message before pushing
other content out of the way.

This was caused by some changes to the icons inside of an input.

Fix by finding an alternate method to address the icon issue.

Fixes #7072.

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -65,7 +65,15 @@ angular.module('material.components.input', [
  */
 function mdInputContainerDirective($mdTheming, $parse) {
 
-  var INPUT_TAGS = ['INPUT', 'TEXTAREA', 'MD-SELECT'];
+  var INPUT_TAGS = ['INPUT', 'TEXTAREA', 'SELECT', 'MD-SELECT'];
+
+  var LEFT_SELECTORS = INPUT_TAGS.reduce(function(selectors, isel) {
+    return selectors.concat(['md-icon ~ ' + isel, '.md-icon ~ ' + isel]);
+  }, []).join(",");
+
+  var RIGHT_SELECTORS = INPUT_TAGS.reduce(function(selectors, isel) {
+    return selectors.concat([isel + ' ~ md-icon', isel + ' ~ .md-icon']);
+  }, []).join(",");
 
   return {
     restrict: 'E',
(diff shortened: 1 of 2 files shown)


topherfangio:
@EladBezalel Can you please review my changes when you get a moment? Everything looks good in the icons demo, but I wanted to double-check with you.

topherfangio:
@EladBezalel NVM...just noticed an alignment issue with the icons that I didn't see before :/

src/components/input/input.js
```diff
@@ -65,7 +65,15 @@ angular.module('material.components.input', [
  */
 function mdInputContainerDirective($mdTheming, $parse) {
 
-  var INPUT_TAGS = ['INPUT', 'TEXTAREA', 'MD-SELECT'];
+  var INPUT_TAGS = ['INPUT', 'TEXTAREA', 'SELECT', 'MD-SELECT'];
```
EladBezalel: Shouldn't this be in a const?

EladBezalel:
Nice improvement @topherfangio LGTM.
Checking locally

src/components/input/input.js
```diff
@@ -65,7 +65,15 @@ angular.module('material.components.input', [
...
+
+  var LEFT_SELECTORS = INPUT_TAGS.reduce(function(selectors, isel) {
+    return selectors.concat(['md-icon ~ ' + isel, '.md-icon ~ ' + isel]);
+  }, []).join(",");
```
EladBezalel: You can use `.toString()` instead of `.join(',')` they're equivalent

topherfangio:
@EladBezalel Just wanted to touch base, is this ready for merge, or does it still need manual testing? :-)

EladBezalel:
i've tested it manually and it's working good, what about the `.toString`?

topherfangio:
@EladBezalel I never knew that `toString()` would automatically add the commas, however, `.join()` is more clear in my opinion, and it's a character shorter, so unless there is a compelling reason to use `.toString()`, I would prefer to stick with join.

EladBezalel:
BTW `.join(',')` is equals to `.join()`

ThomasBurleson:
@topherfangio - :+1:

PR #7148, opened 2016-02-12
update(panel): Adds a basic demo of a dialog using mdPanel.

ErinCoughlan:
Also adds custom css class to the panel.

@jelbourn @KarenParker @gmoothart Please review.


src/components/panel/demoBasicUsage/script.js
```diff
@@ -0,0 +1,34 @@
+angular.module('panelDemo', ['ngMaterial'])
+    .controller('BasicDemoCtrl', BasicDemoCtrl)
+    .controller('DialogCtrl', DialogCtrl);
+
+
+function BasicDemoCtrl($mdPanel) {
+  this._mdPanel = $mdPanel;
+  this._panelRef;
```
gmoothart: Can you initialize this or remove it?

gmoothart:
:beer:

jelbourn:
Where does the image come from and what is its license?

jelbourn:
The image is also really large (73 kB); something smaller would be better.

src/components/panel/demoBasicUsage/script.js
```diff
@@ -0,0 +1,33 @@
...
+    .controller('BasicDemoCtrl', BasicDemoCtrl)
+    .controller('DialogCtrl', DialogCtrl);
+
+
+function BasicDemoCtrl($mdPanel) {
+  this._mdPanel = $mdPanel;
+}
+
+
+BasicDemoCtrl.prototype.showDialog = function() {
```
jelbourn: `showDialog` or `showPane` ?

src/components/panel/panel.js
```diff
@@ -381,14 +383,16 @@ MdPanelService.prototype.open = function(opt_config) {
...
  * @private
  */
-MdPanelService.prototype.wrapTemplate_ = function(origTemplate) {
+MdPanelService.prototype.wrapTemplate_ = function(origTemplate, config) {
   var template = origTemplate || '';
+  var panelClass = config['panelClass'] || '';
 
   return '<div class="md-panel-outer-wrapper">' +
-            '<div class="md-panel">' +
+            '<div class="md-panel ' + panelClass + '">' +
```
jelbourn: Don't append the CSS class into the template string. While somewhat oblique, this is an XSS risk.

test/angular-material-spec.js
```diff
@@ -119,11 +119,17 @@
...
-     * Add special matchers used in the Angular-Material specs
-     *
+     * Add special matchers used in the Angular-Material spec.
      */
     jasmine.addMatchers({
 
+      /**
+       * Asserts that an element has a given class name.
+       * Accepts any of:
+       *   {string} - A query selector.
```
jelbourn: `CSS selector`

test/angular-material-spec.js
```diff
@@ -246,6 +254,17 @@
 
     });
 
+    /**
+     * Returns the angular element associated with a css selector or element.
+     * @param el {string|!angular.JQLite|!Element}
+     * @returns {!angular.JQLite}
+     */
+    function getElement(el) {
+      var queryResult = typeof el == 'string' ?
```
jelbourn: Use `angular.isString`

ErinCoughlan:
The image is from https://pixabay.com/en/donut-white-icing-frosting-glaze-522431/

It's CC0 Public Domain, so free for commercial use and doesn't need attribution. (Which is why I picked that one.)

I'll grab a smaller size.

src/components/panel/demoBasicUsage/script.js
ErinCoughlan: I was thinking showDialog because the specific example is for dialogs. I intend to add more demos for other uses of the panel, such as a menu.

Using showDialog will allow me to have a different function for these other examples, but if you think it will lead to confusion, I can call it something else.

ErinCoughlan:
I fixed all the comments. Please take a look when you get a chance.

@jelbourn The image is now 44.7 kb. If that's still too big, I can find a different image to use.

src/components/panel/demoBasicUsage/style.css
```diff
@@ -0,0 +1,25 @@
...
+}
+
+.dialog-content img {
+  height: 300px;
+  margin: auto;
+}
+
+.dialog-button {
+  width: 100%;
+}
```
jelbourn: Just thought of this- could you prefix these css classes with `demo-`, so that it's clear what is from the demo and what is part of the underlying component?

jelbourn:
LGTM

DerekLouie:
LGTM

PR #7149, opened 2016-02-13
docs(input): document numeric input md-maxlength validation

### LINKED ISSUE #5321: mdMaxlength: Always invalid for input type="number"

DevVersion:
As discussed with @EladBezalel, we should not parse the numeric model to a string but mention  it in the docs.

Fixes #5321

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -157,7 +157,9 @@ function labelDirective() {
  *   specified, a character counter will be shown underneath the input.<br/><br/>
  *   The purpose of **`md-maxlength`** is exactly to show the max length counter text. If you don't
  *   want the counter text and only need "plain" validation, you can use the "simple" `ng-maxlength`
- *   or maxlength attributes.
+ *   or maxlength attributes.<br/><br/>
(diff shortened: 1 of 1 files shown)


EladBezalel:
LGTM

src/components/input/input.js
```diff
@@ -154,7 +154,8 @@ function labelDirective() {
  * allows you to build complex forms for data entry.
  *
  * @param {number=} md-maxlength The maximum number of characters allowed in this input. If this is
- *   specified, a character counter will be shown underneath the input.<br/><br/>
+ *   specified, a character counter will be shown underneath the input.<br/>
+ *   The `md-maxlength` validation is restricted for the use of inputs, which accepts strings.<br/><br/>
```
topherfangio: I might shorten this to

> **Note:** Only valid for text/string inputs (not numeric).

DevVersion: I like it :+1:

PR #7151, opened 2016-02-13
fix(sticky): compile cloned element in the same scope

### LINKED ISSUE #4979: mdSticky: directive with scope params lose assignment
Hi,

i have a directive which has different kinds of scopes. If the mdSticky component clone the directive, the scope values will be lost. But it is nessecary that the scope state stay active.

That is my directive:

```javascript
subnavigation.directive('subNavigation', function($mdSticky){
	return {
		restrict: 'E',
		replace: true,
		scope: {
			navigate: '&',
			activeComponents: '=',
			versions: '=',
			tariffs: '=',
			settings: '='
		},
		templateUrl: 'Resources/Public/js/app/shared/directives/subnavigation/subnavigation.html',
		link: function (scope, element) {
			$mdSticky(scope, element);
		}
	};
});
```

```html
<md-content>
			<sub-navigation navigate="bootstrap.navigateTo(component)" versions="versions"
			                active-components="activeComponents" tariffs="tariffs" settings="settings"></sub-navigation>
...
</md-content>
```

Do i something wrong or is this a bug?

DevVersion:
Currently never apply a scope to the cloned element, which will cause problems with directives, which have bindings in their template.

Also added the documentation for a common use-case

Closes #4979

### DIFF
--- a/src/components/sticky/sticky.js
+++ b/src/components/sticky/sticky.js
@@ -20,6 +20,45 @@ angular
  * @description
  * The `$mdSticky`service provides a mixin to make elements sticky.
  *
+ * By default the `$mdSticky` service compiles the cloned element in the same scope as the actual element lives.
+ *
+ * <h3>Notes</h3>
+ * When using an element which is containing a compiled directive, which changed its DOM structure during compilation,
+ * you should compile the clone yourself using the plain template.<br/><br/>
+ * See the right usage below:
+ * <hljs lang="js">
+ *   angular.module('myModule')
+ *     .directive('stickySelect', function($mdSticky, $compile) {
+ *       var SELECT_TEMPLATE =
+ *         '<md-select ng-model="selected">' +
+ *           '<md-option>Option 1</md-option>' +
+ *         '</md-select>';
+ *
+ *       return {
+ *         restrict: 'E',
+ *         replace: true,
+ *         template: SELECT_TEMPLATE,
+ *         link: function(scope,element) {
+ *           $mdSticky(scope, element, $compile(SELECT_TEMPLATE)(scope));
+ *         }
+ *       };
+ *     });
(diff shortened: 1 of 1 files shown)


EladBezalel:
Code LGTM

DevVersion:
@EladBezalel Yep, just referenced now :)

Sriram143:
this is missing

  MdSticky.$inject = ["$document", "$mdConstant", "$$rAF", "$mdUtil", **"$compile"]**;

DevVersion:
@Sriram143 No need, this will be automatically added through the build process, using ngAnnotate.

Sriram143:
@DevVersion  Thanks for the update.

david-meza:
Thanks for the contribution. I am running into this issue at the moment. Will this still work with more complex templates that use transclusion and/or inner components which have to compile (say, `md-icon`). Thank you.

DevVersion:
@david-meza The new changes won't change anything huge in the service. As mentioned in the documentation, we will compile the cloned directive by default into the provided scope. At this point there can be problems with already-compiled elements which change their structure during compilation. 

So for example if I use a directive template like:
```html
<my-directive>
  Hello
</my-directive>
``` 
and the actual directive will modify the element tree, to look like this:
```html
<my-directive>
  <div class="wrap-container">
    Hello
  </div>
</my-directive>
```
then we'll have a problem with compiling that back into the element tree. The `compilation` of the cloned element is required, to reapply the specified scope to the cloned element.

If this is your case, that the directive can't detect the actual content anymore - then you have to use the plain template as shown in the example.

So finally I can say, this is a good question and isn't easy to solve, because we have no way to detect the initial DOM structure or copying all the needed `controllers` / `directives` back to the cloned element.

PR #7161, opened 2016-02-15
fix(list): fix converted space when target is content editable.

### LINKED ISSUE #5406: Editable divs do not work in a md-list-item
Pressing the spacebar does not result in a space in a editable divs (e.g. TextAngular) in md-list-item nodes.
INPUT and TEXTAREA are already ignored (see #2668) but there is no check for editable divs.

Patch to fix the behaviour:
https://github.com/Pluxbox/material/commit/6cb9a373c90bb437620b86ca05187c8162b25bd9

DevVersion:
As this is a feature which is depending on the browsers behaviour, we need to dispatch native events to the editable element, otherwise the test won't fail when the `editableContent` check is removed.

If you don't trust me :), checkout the PR and use jqLite for triggering the event and remove the `isContentEditable` check.

Fixes #5406

### DIFF
--- a/src/components/list/list.js
+++ b/src/components/list/list.js
@@ -282,7 +282,7 @@ function mdListItemDirective($mdAria, $mdConstant, $mdUtil, $timeout) {
 
         if (!hasClick && !proxies.length) {
           firstChild && firstChild.addEventListener('keypress', function(e) {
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@topher - can you review. This lgtm.

topherfangio:
@ThomasBurleson LGTM

ThomasBurleson:
@DevVersion - :+1:

PR #7198, opened 2016-02-17
update(panel): Adds config option to set z-index on the panel-container.

### LINKED ISSUE #7115: mdPanel: Add z-index to the config object and set it via JS.
The user should be able to specify the z-index for their components.

This task involves:
* Updating documentation for z-index.
* Adding z-index to the config object.
* Setting the z-index for the md-panel-container, using $z-index-dialog as a default.

ErinCoughlan:
Fixes #7115 

@gmoothart @KarenParker @DerekLouie @jelbourn Please review when you get a chance.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -320,6 +320,10 @@ angular
  *****************************************************************************/
 
 
+// Default z-index for the panel.
+var defaultZIndex = 80;
+
+
 /**
  * A service that is used for controlling/displaying panels on the screen.
  * @param {!angular.JQLite} $rootElement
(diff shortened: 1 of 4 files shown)


src/components/panel/panel.js
```diff
@@ -527,5 +531,16 @@ MdPanelRef.prototype._createPanel = function() {
...
       });
 };
+
+
+/**
+ * Adds the styles for the panel, such as positioning and z-index.
+ * @private
+ */
+MdPanelRef.prototype._addStyles = function() {
+  this._panelContainer.css('z-index', this._config['zIndex'] || defaultZIndex);
```
jelbourn: If `config['zIndex'] == 0` it will go to the default index, when it's technially possible that someone would want to set a `0` index.

Instead you could do something like
```js
var configZindex = this._config['zIndex'];
this._panelContainer.css('z-index', configZIndex != null ? configZIndex : defaultZIndex);
```

jelbourn:
LGTM aside from `0` case

src/components/panel/panel.js
ErinCoughlan: I set the default z-index in the default config object. This way, if the user sets z-index to 0, it will overwrite the default value and I don't need to do the checks here. Added a test for this case as well.

jelbourn:
LGTM

PR #7201, opened 2016-02-18
Problem with virtual-repeat

ililienthal:
After the patch https://github.com/angular/material/commit/9a36112c2a2604dbfd2100deb2daa661a773ef06 I got problems with scrolling in a virtual-repeat container.

When you use the mousewheel to scroll down the virtual-repeat container, you can scroll lower than possible. The patch https://github.com/angular/material/commit/9a36112c2a2604dbfd2100deb2daa661a773ef06 catches the case that the user scroll lower than bottom and exit the function.

With my patch the behaviour is that when the user scrolls lower than the bottom, than the container do scroll to the end of the container.


ililienthal:
I signed it!

ThomasBurleson:
@ililienthal - the unit tests are failing. 

screen shot 2016-02-26 at 6 32 49 am

ThomasBurleson:
Closing as out-of-date and missing issue reference.

PR #7210, opened 2016-02-19
feat(chips): Make chips edittable on demand.

### LINKED ISSUE #2320: mdChips: editing existing chip
*  Enter/Spacebar key on active chip should start edit mode on that chip item.
*  Double click on chip should auto-edit `unlocked` chip item

> UX specification details is still pending from the UX Material team.

mertdeg2:
Description: closes #2320

### DIFF
--- a/src/components/chips/chips-theme.scss
+++ b/src/components/chips/chips-theme.scss
@@ -5,7 +5,7 @@ md-chips.md-THEME_NAME-theme {
       box-shadow: 0 2px '{{primary-color}}';
     }
   }
-  .md-chip {
+  md-chip {
     background: '{{background-300}}';
     color: '{{background-800}}';
     &.md-focused {
--- a/src/components/chips/chips-theme.scss
+++ b/src/components/chips/chips-theme.scss
@@ -17,7 +17,7 @@ md-chips.md-THEME_NAME-theme {
     }
 
     &.md-chip-editing {
-      background: #fff;
+      background:transparent;
       color: '{{background-800}}';
     }
   }
--- a/src/components/chips/chips-theme.scss
+++ b/src/components/chips/chips-theme.scss
@@ -15,6 +15,11 @@ md-chips.md-THEME_NAME-theme {
         color: '{{primary-contrast}}';
       }
     }
+
+    &.md-chip-editing {
+      background: #fff;
+      color: '{{background-800}}';
+    }
   }
   md-chip-remove {
     .md-button {
--- a/src/components/chips/js/chipController.js
+++ b/src/components/chips/js/chipController.js
@@ -19,6 +19,11 @@ function MdChipCtrl ($scope, $element, $mdConstant, $attrs, $timeout) {
   this.$scope = $scope;
 
   /**
+   * @type {$scope}
+   */
+  this.$scope = $scope;
+
+  /**
    * @type {$element}
    */
   this.$element = $element;
--- a/src/components/chips/js/chipController.js
+++ b/src/components/chips/js/chipController.js
@@ -45,7 +45,7 @@ function MdChipCtrl ($scope, $element, $mdConstant, $attrs, $timeout) {
 
   if(this.enableChipEdit) {
     this.$element.on('keydown', this.chipKeyDown.bind(this));
-    this.$element.on('dblclick', this.doubleClicked.bind(this));
+    this.$element.on('mousedown', this.chipMouseDown.bind(this));
     this.getChipContent().addClass('_md-chip-content-edit-is-enabled');
   }
 }
@@ -69,13 +69,21 @@ MdChipCtrl.prototype.getContentElement = function() {
 
 
 /**
+ * @return {number}
+ */
+MdChipCtrl.prototype.getChipIndex = function() {
+    return parseInt(this.$element.attr('index'));
+};
+
+
+/**
  * Presents an input element to edit the contents of the chip.
  */
 MdChipCtrl.prototype.goOutOfEditMode = function() {
   this.isEditting = false;
   this.$element.removeClass('md-chip-editing');
   this.getChipContent()[0].contentEditable = 'false';
-  var chipIndex = parseInt(this.$element.attr('index'));
+  var chipIndex = this.getChipIndex();
 
   var content = this.getContentElement().text();
   if (content === "") {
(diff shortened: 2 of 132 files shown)


mertdeg2:
(Hoping npm run watch won't crush on me) Demo: http://mertdeg.sea.corp.google.com:8080/demo/chips

src/components/chips/js/chipController.js
```diff
@@ -0,0 +1,165 @@
...
+     */
+    this.isEditting = false;
+
+    /**
+     * @type {boolean}
+     */
+    this.enableChipEdit = $scope.$parent && $scope.$parent.enableChipEdit == 'true';
+
+    if(this.enableChipEdit) {
+        this.$element.on('keydown', this.chipKeyDown.bind(this));
```
DevVersion: You should indent your code with two spaces to be aligned with the other components.

src/components/chips/js/chipController.js
```diff
@@ -0,0 +1,165 @@
...
+    return metrics.width;
+};
+
+
+/**
+ * Adjusts the input width to match its contents
+ */
+MdChipCtrl.prototype.adjustEditInputWidth = function() {
+    var currentText = this.getEditInput().val();
+    var width = this.getTextWidth(currentText, '16px Roboto');
```
DevVersion: Not every browsers supports `Roboto` and ngMaterial isn't injecting it by itself.

src/components/chips/js/chipsDirective.js
```diff
@@ -194,6 +194,7 @@
       scope: {
         readonly: '=readonly',
         placeholder: '@',
+        enableChipEdit: '@',
```
DevVersion: If you add a new attribute, you should document that in the `ngDocs` comment.

src/components/chips/js/chipController.js
mertdeg2: Good catch. I have replaces four spaces into two

src/components/chips/js/chipsDirective.js
mertdeg2: Added a comment for this

src/components/chips/js/chipController.js
mertdeg2: I have added a change to get computed font-family & font-size instead of hardcoding them.

mertdeg2:
Looking for UX design input. You can find a quick demo here

src/components/chips/js/chipsDirective.js
DevVersion: Ah just rechecked the code and noticed the attribute name. Whenever we provide custom attributes, we should prefix them with `md-` like `md-enable-chip-edit`

src/components/chips/js/chipController.js
```diff
@@ -0,0 +1,181 @@
...
+
+/**
+ * Presents an input element to edit the contents of the chip.
+ */
+MdChipCtrl.prototype.goInEditMode = function() {
+  this.isEditting = true;
+  this.$element.addClass('md-chip-editing');
+  var currentText = this.getContentElement().text();
+  this.getContentElement().addClass('ng-hide');
+  this.getChipContent().append('<input type="text" value="'+currentText+'" />');
```
DevVersion: Just thought a bit about the input. It's just an idea, can we not use a contend editable div? This will work really good and doesn't need all the text measures. Then we just need to watch for the content editable blur and update the actual chip model.

mertdeg2: I like this idea. I have used this approach in the next iteration.

src/components/chips/js/chipController.js
```diff
@@ -0,0 +1,149 @@
...
+};
+
+
+/**
+ * Presents an input element to edit the contents of the chip.
+ */
+MdChipCtrl.prototype.goInEditMode = function() {
+  this.isEditting = true;
+  this.$element.addClass('md-chip-editing');
+  this.getChipContent().attr('contenteditable', 'true');
```
DevVersion: Hey, sorry for tracking all these things, just trying to find a perfect solution.

I think using the native `HTMLElement` property will do a better job ([mdn](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/contentEditable) - the syntax will be: 
```js
element.contentEditable = "true"
````

Comment (mertdeg2) [inline src/components/chips/js/chipController.js]: No problem! I appreciate your feedback. I have changed the code to use the native HTMLElement property.

**Comment by DevVersion** [comment]
@mertdeg2 Just run it myself, works perfect :+1:

 - About the editing style, I'm not sure, do you think this is the right way?

  

- Also I think, you should check for an empty text, because it's not good if the chip still persists, when the text is empty.

**Comment by mertdeg2** [comment]
Thanks Paul! Good catch, I have fixed the second issue. We had a discussion
about your first point.

It seemed reasonable to follow gmail's chips implementation like below.
Having said that, I am open to suggestions about this.
​​

File: src/components/chips/chips-theme.scss [hunk 6]
```diff
@@ -15,6 +15,11 @@ md-chips.md-THEME_NAME-theme {
         color: '{{primary-contrast}}';
       }
     }
+
+    &.md-chip-editing {
+      background: #fff;
```
conniew: Should this color be hard-coded? Is there a variable for the general chipbar background color?

mertdeg2:
I signed it!

ThomasBurleson:
@char0n - why are you commenting on a branch change that has nothing to do with `master` or the weekly patch releases (like 1.0.6) ?

ThomasBurleson:
@char0n - it was reverted because Chrome fixed in it v48. Are you  saying this is STILL an issue with Firefox?  If so, then our apologies, we will restore the fix.

char0n:
@ThomasBurleson, sorry I don't know the exact process, I found the commit that caused the issue and commented. Yes I confirm that Chrome v48 fixes the issue, but the issue is still present in FireFox 44.0.2

DevVersion:
@char0n @ThomasBurleson GitHub automatically shows comments from commits which are attached to the PR (See for example: https://github.com/JonFerrera/material/pull/1 - referenced by https://github.com/angular/material/commit/179dc1989c3fa695433b1faea6b05f3be921948d)

char0n:
@ThomasBurleson Issue created: https://github.com/angular/material/issues/7382

@DevVersion thanks for info.

Issue #7220, opened 2016-02-20
mdMenu: mdButton that becomes an <a> tag will not take focus on hover


snaumets:
It seems like normally, hovering over an mdButton item in the menu will also give focus to that element, but when the mdButton becomes an `<a>` tag, it no longer takes that focus, and the `<button>` above it in the menu keeps focus, which seems like incorrect/unintentional behavior.
A live example can be seen [here](http://codepen.io/anon/pen/MKdxjx?editors=101). In it, when you hover over any of the 'somethingX' items, it also takes the focus, but when you hover over the 'google' item which is actually a link, it will not steal the focus.

In a mdMenu that consists of mdButtons, each mdButton becomes either a `<button>` or an `<a>` tag (when given either an href or ng-href for example). However, the `<a>` tags dont take the focus when the element is hovered on. It seems that only non-disabled buttons can take the focus.

jelbourn:
cc @rschmukler

topherfangio:
Closed in 0627d9e98320c60158de13385ea26208aa8567ab.

PR #7221, opened 2016-02-20
fix(handleMenuItemHover): allow any mdButton to be a focusableTarget

snaumets:
Allow any mdButton element (`<button>` or `<a>` tag) to take focus when it is hovered over.
Previously, only a non-disabled button could be a focusable target when being hovered over.
Now, any non-disabled element with the md-button class can be a focusable target and take focus.

Closes issue #7220


snaumets:
I signed it!

jelbourn:
LGTM

PR #7236, opened 2016-02-22
fix(sidenav, backdrop): fill the backdrop with its parents height

### LINKED ISSUE #5861: scrolling when sideNav open
screen shot 2015-11-20 at 10 51 04 pm
See in this copy of the angular material website sideNav codepen demo:

http://codepen.io/anon/pen/ZbZyZN

- Open the right sideNav
- Scroll down

Result: You see the backdrop and the right sidenav scrolling up
Expected: nothing should scroll

Note: removing the `height: 500px;` on the body hides the issue

The workaround I have been using:

    md-backdrop, md-sidenav {
	position: fixed;
    }

DevVersion:
> As there is the fixed height on the second parent, everything works as expected, except that the backdrop should fill its parent, when we are not locking the whole body

@topherfangio Can you take a look?

Fixes #5861

### DIFF
--- a/src/components/backdrop/backdrop.js
+++ b/src/components/backdrop/backdrop.js
@@ -19,50 +19,69 @@
 
 angular
   .module('material.components.backdrop', ['material.core'])
-  .directive('mdBackdrop', function BackdropDirective($mdTheming, $animate, $rootElement, $window, $log, $$rAF, $document) {
-    var ERROR_CSS_POSITION = "<md-backdrop> may not work properly in a scrolled, static-positioned parent container.";
+  .directive('mdBackdrop', MdBackdropDirective)
+  .controller('$mdBackdropController', MdBackdropController);
 
-    return {
-      restrict: 'E',
-      link: postLink
-    };
 
-    function postLink(scope, element, attrs) {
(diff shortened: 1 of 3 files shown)


src/components/backdrop/backdrop.js
```diff
@@ -19,50 +19,69 @@
...
-      // adjust the 'backdrop' height to account for the fixed 'body' top offset
-      var body = $window.getComputedStyle($document[0].body);
-      if (body.position == 'fixed') {
-        var hViewport = parseInt(body.height, 10) + Math.abs(parseInt(body.top, 10));
-        element.css({
-          height: hViewport + 'px'
-        });
-      }
+  .directive('mdBackdrop', MdBackdropDirective)
+  .controller('$mdBackdropController', MdBackdropController);
```
DevVersion: FYI: Don't wonder about the diff, this is because I moved the anonymous directive function into a declared - should be better for the controller etc.

ThomasBurleson:
@topherfangio - ping.

DevVersion:
Closing this one as outdated - as well. Seems to be the wrong approach.

PR #7239, opened 2016-02-22
refactor(bottomSheet,dialog,list,select,input,menu,menuBar): denote private classes with _md prfix

topherfangio:
Description copied from Ryan's PR #6730 which this will close when merged.

> Note: This commit will break input until it has been updated with private classes. I modified some of the classes as they interact with md-select. Presumably @topherfangio changes will result in the same changes as I made resulting in full compatibility between select and input.


ThomasBurleson:
Duplicate. See aggregation into PR #7248.

PR #7246, opened 2016-02-23
feat(toast): add highlight class method + improved docs design + align demo with specs

### LINKED ISSUE #6607: mdToast (docs): examples don't align with Material Spec
Currently the [docs](https://material.angularjs.org/latest/demo/toast) show an 'OK' button and a 'Close' button in the two snackbar (toast) examples (Show With Action, Show Custom).

Based on the [spec](https://www.google.com/design/spec/components/snackbars-toasts.html#snackbars-toasts-usage) having this kind of a 'Dismiss', 'OK', 'Close' button in a snackbar is a "Don't".

The docs should not show examples of things that go against the spec as it creates [confusion](http://stackoverflow.com/questions/28481029/how-can-i-change-the-color-of-toast-depends-on-message-type-in-angular-material/32241428?noredirect=1#comment56994080_32241428) with users.

DevVersion:
BREAKING CHANGE: Toasts now use accent color by default for a highlighted action button.

-  Add `highlightClass` method to toast (very useful! - and won't break anything)
-  Improved the docs design for the possible methods for the #simple method.
- Changed demo to align with specs, No Dismiss buttons.
- Highlight Color should use by default the accent color.

@topherfangio As discussed changed the demo and the default highlight color.
- So about the `highlightClass`, we didn't discuss that, but it seems very useful, when aligning the toast to the official specs.

**Little preview of the new docs changes**

New Docs

**Preview of the old doc design**

Old

I think the new design looks much more clearer and also looks better with the new `highlightClass` method.

Fixes #6607

### DIFF
--- a/src/components/toast/toast-theme.scss
+++ b/src/components/toast/toast-theme.scss
@@ -1,15 +1,20 @@
 md-toast.md-THEME_NAME-theme {
+
   .md-toast-content {
     background-color: #323232;
     color: '{{background-50}}';
 
     .md-button {
       color: '{{background-50}}';
+
       &.md-highlight {
-        color: '{{primary-A200}}';
-        &.md-accent {
-          color: '{{accent-A200}}';
+        // By default the toast should use the accent color as the primary highlight color
+        color: '{{accent-A200}}';
+
+        &.md-primary {
+          color: '{{primary-A200}}';
         }
+
         &.md-warn {
           color: '{{warn-A200}}';
         }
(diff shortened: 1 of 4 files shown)


src/components/toast/toast-theme.scss
```diff
@@ -1,15 +1,20 @@
...
 
     .md-button {
       color: '{{background-50}}';
+
       &.md-highlight {
-        color: '{{primary-A200}}';
-        &.md-accent {
-          color: '{{accent-A200}}';
+        // By default the toast should use the accent color as the primary highlight color
+        color: '{{accent-A200}}';
```
topherfangio: @DevVersion This is a breaking change since we are changing the default from primary to accent. I believe it is a correct breaking change, but we need to make sure that we note it in the commit message (and PR description) so that users will know. Can you update the commit message accordingly?

DevVersion: Thank's for remembering, just updated :)

PR #7252, opened 2016-02-23
update(sticky): only compile the self-cloned element in the given scope.

DevVersion:
As discussed with @topherfangio, we should not take advantage of the `scope()` method, because these will be only available in Angular's debug mode.
So we both agreed, with removing that and just compiling the cloned element from service into the given scope.


ThomasBurleson:
Nice work with the unit tests!

PR #7270, opened 2016-02-24
fix(docs): docs stylesheet should not prevent scrolling in code pens

### LINKED ISSUE #7216: List Controls Demos don't have scroll when opened in codepen
http://codepen.io/anon/pen/QyRZpM
i checked both Chrome and Firefox 
image

### LINKED ISSUE #7269: codepen: demos are not scrollable
At the moment there are situations when scrolling in a codepen demo isn't possible.

-  Codepen demos from **HEAD** are not affected, because the `demo.css`script is not injected probably.
- Relases are affected, because the `demo.css` is referenced / injected correctly. 
 
 Little example: https://material.angularjs.org/latest/demo/card

---
For example in HEAD builds the `docs.css` will be injected using the current commit SHA and not the latest release tag. (Like so - https://material.angularjs.org/master-9e7937e/docs.css)

So the actual problem on releases' demo is:
```
body {
  overflow: hidden;
  max-width: 100%;
  max-height: 100%;
}
```

I'm not suggesting to remove that `overflow` property or neither stopping injecting the `docs.css`.
A really elegant solution will be, to add a unique identifier to the CSS selector which will be only used in the `docs`, not in the demos.

Includes - #7216

DevVersion:
- See more in #7269 

Fixes #7269 Fixes #7216

### DIFF
--- a/docs/app/css/style.css
+++ b/docs/app/css/style.css
@@ -7,7 +7,8 @@ body {
   font-size: 1.6rem;
 }
 
-body {
+/* This styling should be not applied to codepen demos, so we need to add a unique identifier. */
+body.docs-body {
   overflow: hidden;
   max-width: 100%;
   max-height: 100%;
--- a/docs/config/template/index.template.html
+++ b/docs/config/template/index.template.html
@@ -12,7 +12,7 @@
 <link rel="stylesheet" href="angular-material.min.css">
 <link rel="stylesheet" href="docs.css">
 </head>
-<body layout="row" ng-cloak>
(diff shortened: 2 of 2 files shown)


ThomasBurleson:
This does not fix #7216 codepen:  http://codepen.io/anon/pen/QyRZpM

DevVersion:
@ThomasBurleson Oh it does, I just removed the `docs.css` from the codepen to simulate my PR changes. http://codepen.io/DevVersion/pen/VaZWWY

PR #7290, opened 2016-02-25
docs(autocomplete): better describe the `md-autofocus` attribute

### LINKED ISSUE #7283: md-autocomplete: md-autofocus not focusing input.
Passing md-autofocus="true" to the md-autocomplete directive has no effect: the input doesn't get focused.

Material version 1.0.5

Live example: http://codepen.io/anon/pen/ZQgZgp

thanks.

DevVersion:
The current description for the autocomplete attribute was really confusing, and the user was wondering why the autocomplete wasn't retrieving focus automatically.
But they didn't know about the other behaviour of the actual `md-autofocus` directive.
The autocomplete attribute "wrap" is just a little addition which focus the related input instead of the element.

Fixes #7283

### DIFF
--- a/src/components/autocomplete/js/autocompleteDirective.js
+++ b/src/components/autocomplete/js/autocompleteDirective.js
@@ -49,7 +49,9 @@ angular
  *     make suggestions
  * @param {number=} md-delay Specifies the amount of time (in milliseconds) to wait before looking
  *     for results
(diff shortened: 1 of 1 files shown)


flea89:
Nice @DevVersion, that's much clearer now. 

Personally I think that a reference ( or just pointing the user ) to read the actual md-autofocus directive doc might be ace

PR #7291, opened 2016-02-25
fix(autocomplete): clean up md-scroll-mask element on destroy

### LINKED ISSUE #7049: Changing route in response to md-autocomplete clear creates md-scroll-mask
Assign model to $scope.
$watch model.
Create a md-autocomplete, bind to a model, select an value, clear the value.
In the model $watcher, change route.

there will be a stale md-scroll-mask attached to the body.

### LINKED ISSUE #7128: md-autocomplete triggering an ng-route change makes page unusable
I'm running into a few issues when using an autocomplete list which triggers a $location.path() call on an md-selected-item-change.

Fiddle: http://jsfiddle.net/999t17Lw/12/

note that when selecting B or C, the page gets successfully routed, but then nothing can be interacted with due to an md-scroll-mask.  The hacky workaround of adding $mdUtil.enableScrolling() to the controller allows the page to be usable upon routing, but the autocomplete directive becomes useless on the new page.

cah-alexsukhodolsky:
* added test

Fixes #7049
Fixes #7128

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -365,6 +365,49 @@ describe('<md-autocomplete>', function() {
       element.remove();
     }));
 
+    it('should remove the md-scroll-mask on cleanup', inject(function($mdUtil, $timeout, $material) {
+      spyOn($mdUtil, 'enableScrolling');
+
+      var scope = createScope();
+      var template =
+          '<md-autocomplete' +
+          '   md-selected-item="selectedItem"' +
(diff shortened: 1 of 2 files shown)


cah-alexsukhodolsky:
I signed it!

ilovett:
Yep, I ran into this issue on 1.0.5 -- http://codepen.io/anon/pen/ZWWqrM

Seems to be working on 1.0.6 -- http://codepen.io/anon/pen/RaaeMa

Prozi:
does it work with ngRouter and latest master - I don't think so :/

PR #7298, opened 2016-02-26
fix(contactChipsDirective): Remplement the filter-selected attribute

### LINKED ISSUE #4313: md-date-picker: Usability scrolling

jamesfer:
- Reimplement the `filter-selected` as an optional attribute on mdContactChips.
  The value of `filter-selected` is also used as the value for the `md-no-cache` attribute on the
  autocomplete inside the directive so that when filtering is enabled, caching is disabled,
  and vice versa.
  This was the behaviour before the attribute was disabled [here](https://github.com/angular/material/commit/48c7c448fc100540af8cd2b27bd710825be0346e#diff-a49333b5f28f4db760c64a42a04866fcL60)
- Add checkbox to the contact chips demo to enable/disable filtering.
- Update contact chips unit tests
  - There are two tests that involve filtering, the first one expected 3 matches when it should
    have expected 2 ([changed here](https://github.com/angular/material/commit/786b0ed3652b7460c2c802efa1aa79972bd96f5d#diff-26f31de42de579310d01694fc34904bbL63)).
    The second one was commented out as filtering was previously unable to be turned off.

Note: In the demo, the behaviour of the `filter-selected` checkbox is a little funky:
Any searches while the checkbox is selected, will be cached without the selected contacts in the results. When the checkbox is unselected these cached searches will be used and therefore, selected contacts will still not appear even though they are not being filtered out. Even though this is unintuitive I thought it wasn't a big problem as it is very unlikely the `filter-selected` setting will be changed while the directive is in use.

Closes #4313

### DIFF
--- a/src/components/chips/contact-chips.spec.js
+++ b/src/components/chips/contact-chips.spec.js
@@ -83,7 +83,7 @@ describe('<md-contact-chips>', function() {
     });
 
     describe('filtering selected items', function() {
-      it('should filter', inject(function() {
+      it('should filter by default', inject(function() {
         scope.querySearch = jasmine.createSpy('querySearch').and.callFake(function(q) {
           return scope.allContacts;
         });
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
Will not merge.

PR #7307, opened 2016-02-26
fix(sticky): tests relaid on private classes

EladBezalel:
- Tests was using `md-sticky-clone` instead of `_md-sticky-clone`


DevVersion:
Nice catch, Thanks!

PR #7308, opened 2016-02-26
fix(sticky): fix travis unit test failure

DevVersion:
After 78ccef6354cbe8898bfe3c489e3b7ea1d43bbffc we missed to update the `sticky` test
- The test failed because it used a CSS class from `$mdSticky` service, which was changed in 78ccef6354cbe8898bfe3c489e3b7ea1d43bbffc


DevVersion:
As I was in stress with fixing that, I didn't checked for duplicate PR's. So we will go with definitely with @EladBezalel's PR - #7307

PR #7318, opened 2016-02-27
feat(toast): add dynamic `start` and `end` positions.

### LINKED ISSUE #7263: FR: $mdToast 'position' should suport rtl
Currently, we can specify `left` or `right` for the `$mdToast` `position` attribute.

In web pages that has both rtl and ltr versions it is sometimes necessary to specify `start`/`end` position instead of `left`/`right`. Those should specify the position in relation to the the container direction. That is, when `direction=ltr` then `start=left,end=right`, and when `direction=rtl`, it's `start=right,end=left`.

DevVersion:
The `start` and `end` positions are dynamic and are useful when using a bi-directional application with RTL and LTR

Fixes #7263

### DIFF
--- a/src/components/toast/toast.scss
+++ b/src/components/toast/toast.scss
@@ -173,6 +173,15 @@ md-toast {
       top: 0;
     }
 
+    // Support for RTL alignment
+    &._md-start {
+      @include rtl-prop(left, right, 0);
+    }
+
+    &._md-end {
+      @include rtl-prop(right, left, 0);
+    }
+
     /*
    * When the toast doesn't take up the whole screen,
    * make it rotate when the user swipes it away
--- a/src/components/toast/toast.js
+++ b/src/components/toast/toast.js
@@ -198,8 +198,10 @@ function MdToastDirective($mdToast) {
   *   - `hideDelay` - `{number=}`: How many milliseconds the toast should stay
   *     active before automatically closing.  Set to 0 or false to have the toast stay open until
   *     closed manually. Default: 3000.
-  *   - `position` - `{string=}`: Where to place the toast. Available: any combination
-  *     of 'bottom', 'left', 'top', 'right'. Default: 'bottom left'.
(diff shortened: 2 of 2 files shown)


src/components/toast/toast.scss
```diff
@@ -173,6 +173,15 @@ md-toast {
       top: 0;
     }
 
+    // Support for RTL alignment
+    &._md-start {
```
EladBezalel: Where are we using those classes?

DevVersion: It's a new feature, there should be two new position properties for the toast - `start` and `end`. The prefix will be automatically applied in the controller. `'md-' + positionFlag`

EladBezalel: so where are the new position properties?

DevVersion: There're already registered through the CSS.
The toast interim automatically applies the prefix and uses the given css classes.

See the mapping line.
```js
element.addClass(isSmScreen ? '_md-bottom' : options.position.split(' ').map(function(pos) {
  return '_md-' + pos;
}).join(' '));
```
[more code](https://github.com/DevVersion/material/blob/feat/toast-rtl-positions/src/components/toast/toast.js#L389)

EladBezalel: ah, ok sorry for that

DevVersion: No worries, thanks for reviewing :)

ThomasBurleson:
@ErinCoughlan - This is related to mdPanel and positioning... what are your thoughts ?

ErinCoughlan:
`start` and `end` are supported in mdPanel. In fact, we are not supporting `left` and `right` as start/end covers those cases and the "correct" behavior is to support RTL and LTR. (Those two are easy to add back though if necessary.)

For toast, we're not currently using the mdPanel, and I don't know when the implementation will get converted. For consistency, I think they should be added.

DevVersion:
@ThomasBurleson Ping.

PR #7319, opened 2016-02-27
fix(docs): retrieve source descriptor dynamically.

### LINKED ISSUE #7131: Github links for some API docs are broken
For example: https://material.angularjs.org/latest/api/service/$mdInkRipple
The github links points to
```
https://github.com/angular/material/blob/master/src/components/gle/home/jelbourn/ng-material/src/core/services/ripple/ripple.js
```

Cause:
The fuction `publicDocData` in `componentData.js` just checks whether the docType is `"directive"`. And, if so, assume the thing will live under `components/`. But ripple (and a few other things) are marked as `directive` and live in `core/`. So an `indexOf` check on my local path, while cutting a release, messes up. E.g., the path to the file is
```
/usr/local/google/home/jelbourn/ng-material/src/core/services/ripple/ripple.js
```
And the substring is removing everything up to `goo` instead of everything up to `core/`

Thanks to @petebacondarwin for pointing this out.

DevVersion:
At the moment the doc descriptor is retrieved through hard-coding.
This causes problems when a component is not located in the hard-coded folder `src/components`
After this fix, the doc descriptor will be dynamically parsed (using the last possible descriptor criteria), so this allows components to be located everywhere in the project.
This fixes the wrong GitHub links for the ripple service / component.

Fixes #7131

### DIFF
--- a/docs/config/processors/componentsData.js
+++ b/docs/config/processors/componentsData.js
@@ -5,7 +5,17 @@ var buildConfig = require('../../../config/build.config.js');
 // add many kilobytes of loading overhead.
 function publicDocData(doc, extraData) {
   var options = _.assign(extraData || {}, { hasDemo: (doc.docType === 'directive') });
-  return buildDocData(doc, options, 'components');
+
+  // This RegEx always retrieves the last source descriptor.
+  // For example it retrieves from `/opt/material/src/core/services/ripple/ripple.js` the following
+  // source descriptor: `src/core/`.
+  // This is needed because components are not only located in `src/components`.
+  var descriptor = doc.fileInfo.filePath.toString().match(/src\/.*?\//g).pop();
+  if (descriptor) {
+    descriptor = descriptor.substring(descriptor.indexOf('/') + 1, descriptor.lastIndexOf('/'));
+  }
+
+  return buildDocData(doc, options, descriptor || 'components');
 }
 
 function coreServiceData(doc, extraData) {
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@EladBezalel - can you review ?

EladBezalel:
LGTM

docs/config/processors/componentsData.js
```diff
@@ -5,7 +5,17 @@ var buildConfig = require('../../../config/build.config.js');
 // add many kilobytes of loading overhead.
 function publicDocData(doc, extraData) {
   var options = _.assign(extraData || {}, { hasDemo: (doc.docType === 'directive') });
-  return buildDocData(doc, options, 'components');
+
+  // This RegEx always retrieves the last source descriptor.
+  // For example it retrieves from `/opt/material/src/core/services/ripple/ripple.js` the following
+  // source descriptor: `src/core/`.
+  // This is needed because components are not only located in `src/components`.
+  var descriptor = doc.fileInfo.filePath.toString().match(/src\/.*?\//g).pop();
```
jelbourn: I feel like the node `fs` api might be able to do this more easily, but I'm not that familiar with it.
@robertmesserle  ?

robertmesserle: Nothing comes to mind, but I'm not sure I fully understand the context.

DevVersion: I don't think the `fs` / `path`api has anything for that, it's a really unspecific case.

jelbourn:
LGTM

PR #7322, opened 2016-02-28
feat(progresCircular): Full rewrite to SVG

crisbeto:
Full rewrite of the `progressCircular` directive to use SVG for rendering the circle. 

TODO:
- [x] Finishing and hooking up the indeterminate mode.
- [x] Re-doing the unit tests.


src/components/progressCircular/js/progressCircularDirective.js
```diff
@@ -0,0 +1,184 @@
+"use strict";
```
DevVersion: `use strict` is not needed here, because it will be automatically added in the build process.

src/components/progressCircular/js/progressCircularDirective.js
```diff
@@ -0,0 +1,184 @@
...
+
+      // The viewBox has to be applied via setAttribute, because it is
+      // case-sensitive. If jQuery is included in the page, `.attr` lowercases
+      // all attribute names.
+      svg[0].setAttribute('viewBox', '0 0 ' + diameter + ' ' + diameter);
+
+      // No need to animate it if the values are the same
+      if(animateTo === animateFrom){
+        path.attr('d', getSvgArc(animateTo, diameter, pathDiameter));
+      }else{
```
DevVersion: Little nitpicky, but can you space the else from curly braces?
See in the Google Code Conventions [here](https://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting)

src/components/progressCircular/js/progressCircularDirective.js
```diff
@@ -0,0 +1,184 @@
...
+        (function animation() {
```
DevVersion: Question: Isn't the first animation-tick running synchronously in the ` `link` Function?

src/components/progressCircular/js/progressCircularDirective.js
crisbeto: Good to know, thanks.

src/components/progressCircular/js/progressCircularDirective.js
crisbeto: Sure, I tried to be consistent but apparently this slipped through.

src/components/progressCircular/js/progressCircularProvider.js
```diff
@@ -0,0 +1,66 @@
...
+    $get : function() { return progressConfig; }
+  };
+
+  function linearEase(t, b, c, d) { return c * t / d + b; }
+  function materialEase(t, b, c, d) {
+    // via http://www.timotheegroleau.com/Flash/experiments/easing_function_generator.htm
+    // with settings of [0, 0, 1, 1]
+    var ts=(t/=d)*t;
+    var tc=ts*t;
+    return b+c*(6*tc*ts + -15*ts*ts + 10*tc);
```
DevVersion: Nitpicky: I would reformat the code, instead of just using the exported code.
I mean, you should space the arithmetic operators.

See quote of the conventions:
> BE CONSISTENT.

> If you're editing code, take a few minutes to look at the code around you and determine its style. If they use spaces around all their arithmetic operators, you should too. If their comments have little boxes of hash marks around them, make your comments have little boxes of hash marks around them too.

> The point of having style guidelines is to have a common vocabulary of coding so people can concentrate on what you're saying rather than on how you're saying it. We present global style rules here so people know the vocabulary, but local style is also important. If code you add to a file looks drastically different from the existing code around it, it throws readers out of their rhythm when they go to read it. Avoid this.

src/components/progressCircular/js/progressCircularDirective.js
crisbeto: Yes, it's always (technically) synchronous. The condition in here is for cases where the watcher gets fired by a change in the diameter or any other (in the future) properties that aren't being animated with JS.

src/components/progressCircular/js/progressCircularProvider.js
```diff
@@ -0,0 +1,66 @@
+"use strict";
```
DevVersion: Just to remember you about this file too.

src/components/progressCircular/js/progressCircularDirective.js
DevVersion: Yes, I was just wondering, maybe we can initialize the animation also by using `$$rAF` 
Like so 
```js
$$rAF(function(animate) {
```

crisbeto: I don't think it's necessary since it's a single frame.

ThomasBurleson:
@crisbeto - can you rebase from master and squash your commits? Thx.

PR #7327, opened 2016-02-29
fix: Layout issue with flex and layout breakpoints #6528

### LINKED ISSUE #6528: Layout issue with flex and layout breakpoints
Hi,

we ran into an issue with the layout changes in Angular Material 1.0.0 or 1.0.1.
We want a column layout on mobile device and a row layout on desktop, where the second div has a 40% width  and 100% height on row-layout and 40% height and 100% width on column-layout.

Example Code ( http://codepen.io/anon/pen/yegQBm ):
<div layout="column" layout-gt-sm="row" layout-fill>
  <div flex></div>
  <div flex="40"></div>
</div>

If the wrapper has layout="column" and layout-gt-sm="row", the max-width is not set correctly on  larger width than sm.

The same problem appears if I change the layout to 
layout-xs="column" layout-sm="column" layout="row" ( Code: http://codepen.io/anon/pen/pgRQvX ) or
layout-xs="column" layout-sm="column" layout-gt-sm="row"

keenondrums:
fix: Layout issue with flex and layout breakpoints

Change class definition order in flex-properties-for-name mixin.

Closes #6528

### DIFF
--- a/src/core/style/layout.scss
+++ b/src/core/style/layout.scss
@@ -152,6 +152,16 @@
       box-sizing: border-box;
     }
 
+    .layout-row {
+	    > .#{$flexName}-33   , > .#{$flexName}-33     {  flex: 1 1 33.33%;  max-width: 33.33%;  max-height: 100%; box-sizing: border-box; }
+	    > .#{$flexName}-66   , > .#{$flexName}-66     {  flex: 1 1 66.66%;  max-width: 66.66%;  max-height: 100%; box-sizing: border-box; }
+	  }
+	  
+	  .layout-column {
+	    > .#{$flexName}-33   , > .#{$flexName}-33     {  flex: 1 1 33.33%;  max-width: 100%;  max-height: 33.33%; box-sizing: border-box; }
+	    > .#{$flexName}-66   , > .#{$flexName}-66     {  flex: 1 1 66.66%;  max-width: 100%;  max-height: 66.66%; box-sizing: border-box; }
+	  }
+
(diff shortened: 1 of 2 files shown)


keenondrums:
I signed it!

src/core/style/layout.scss
```diff
@@ -135,7 +135,23 @@
       box-sizing: border-box;
     }
 
-    .layout-row > .#{$flexName}-#{$i * 5},
+    .layout-row > .#{$flexName}-#{$i * 5} {
```
ThomasBurleson: This will not work in you have markup like this:

```html
<div layout="row" layout-sm="column">
  <div flex="33"  flex-sm="95">
  <div flex="33"  flex-sm="95">
  <div flex="33"  flex-sm="95">
</div>
```

keenondrums: Sorry, my bad! Kind of never user 33 and 66. FIxed it up.

keenondrums:
Sorry, my bad! Kind of never user 33 and 66. FIxed it up.

ThomasBurleson:
@keenondrums - why did you change the class definition order in flex-properties-for-name mixin.

keenondrums:
@ThomasBurleson, maybe my mind is not at the right place but that was kind of the whole issue behind the bug described in the related issue https://github.com/angular/material/issues/6528. Please correct me if I'm wrong. I'd be happy to answer any questions via Skype if that's more convenient for you. My skype id is keenondrums.

PR #7329, opened 2016-02-29
fix(datePicker): ensure one is able to override default sass variables values

sryzycki:


sryzycki:
I signed it!

DevVersion:
LGTM

ThomasBurleson:
@jelbourn - I assume you had specific reasons why you used `!default`... so their removal would be problematic ?

jelbourn:
LGTM. I just forgot to add `!default` when I wrote them.

PR #7348, opened 2016-02-29
fix(grid-list): validate that row height is defined

### LINKED ISSUE #6870: gridList: md-row-height is required but throws unhelpful error
`md-grid-list` should either handle not having `md-row-height` set or throw a more helpful error that it is required. The other grid list attributes give nice errors but `md-row-height` doesn't.

Here is a demo showing it
http://codepen.io/anon/pen/dGeoYP
Just add back `md-row-height="3:2"` to show it working besides this.

Currently you get the following
image
image

EladBezalel:
- throwing more informative error about missing row height attribute

fixes #6870


ThomasBurleson:
@EladBezalel - unit tests ?

PR #7354, opened 2016-03-01
fix(autocomplete): probably clear the autocomplete.

### LINKED ISSUE #7180: autocomplete requiere double click in text
https://material.angularjs.org/latest/demo/autocomplete

tested in chrome in autocomplete demo and in my project

select item in autocomplete requiere double click if you click in text

image: https://drive.google.com/file/d/0B4BrpMT6tZxtQlJUWU1GQXNCZGc/view?usp=sharing

PD: only occurs when clean from the button

DevVersion:
- At the moment we call the `select(-1)` function when clearing, but that will cause problems, because it will stuck with the matches.

- There is no need of calling the `select(-1)` function, we should just clear the loading indicator on an synchronous query.
On an async query, the loading indicator will be set by default to false on completion.

This fixes the issue, which required two clicks on a list item after cleared the autocomplete through the button.

Fixes #7180

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -16,12 +16,17 @@ describe('<md-autocomplete>', function() {
     items = items || ['foo', 'bar', 'baz'].map(function(item) {
         return {display: item};
       });
-    inject(function($rootScope) {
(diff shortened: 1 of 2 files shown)


robertmesserle:
LGTM

PR #7358, opened 2016-03-01
fix(mdFabSpeedDial): styles on nested speed dials

### LINKED ISSUE #7356: mdFabSpeedDial: nested speed dial style broken
Nested speed dials doesn't always get the correct style applied.

v1.0.6

IcanDivideBy0:
fixes #7356
The `md-fab-actions` component should always be a direct child of the `md-fab-speed-dial` component anyways.

### DIFF
--- a/src/components/fabSpeedDial/fabSpeedDial.js
+++ b/src/components/fabSpeedDial/fabSpeedDial.js
@@ -129,13 +129,13 @@
 
       var el = element[0];
       var ctrl = element.controller('mdFabSpeedDial');
-      var items = el.querySelectorAll('.md-fab-action-item');
+      var items = findDirectChildren(findDirectChild(el, 'md-fab-actions'), '.md-fab-action-item');
 
       // Grab our trigger element
-      var triggerElement = el.querySelector('md-fab-trigger');
+      var triggerElement = findDirectChild(el, 'md-fab-trigger');
 
       // Grab our element which stores CSS variables
-      var variablesElement = el.querySelector('._md-css-variables');
+      var variablesElement = findDirectChild(el, '._md-css-variables');
 
       // Setup JS variables based on our CSS variables
       var startZIndex = parseInt(window.getComputedStyle(variablesElement).zIndex);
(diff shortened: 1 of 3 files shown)


IcanDivideBy0:
I signed it!

src/components/fabSpeedDial/fabSpeedDial.js
```diff
@@ -129,13 +129,13 @@
 
       var el = element[0];
       var ctrl = element.controller('mdFabSpeedDial');
-      var items = el.querySelectorAll('.md-fab-action-item');
+      var items = el.querySelectorAll(':scope > md-fab-actions > .md-fab-action-item');
```
DevVersion: Not every browsers supports scoped stylesheets. See [the MDN reference](https://developer.mozilla.org/de/docs/Web/CSS/:scope#Browser_compatibility)

Especially all IE browsers aren't supported, but `ngMaterial` is supporting IE11

IcanDivideBy0: Aw, IE... I'll try to find a workaround.

DevVersion: Also please take a look at the tests after you found a workaround.

IcanDivideBy0: Sure, i wanted to validate the approach first, I'll fix the test then.

IcanDivideBy0:
I'm not totally conviced by the approach I used... let me know what do you think about it. Without the `:scope` pseudo element, I can't find a way to get direct children of an element with the querySelector

IcanDivideBy0:
IE and PhantomJS does not implements `Element.matches` (but IE yet have `msMatchesSelector` https://developer.mozilla.org/en-US/docs/Web/API/Element/matches).
What should I do ? is there any place to put a polyfill ? Should I develop another utility function ?

ThomasBurleson:
@topherfangio - please review

ThomasBurleson:
Nested Speed dials will not be supported.

PR #7366, opened 2016-03-01
refactor: Refactor BDO styles to be cleaner and less specific.

topherfangio:
Refactor per discussion with Thomas.

### DIFF
--- a/src/core/style/structure.scss
+++ b/src/core/style/structure.scss
@@ -170,3 +170,14 @@ input {
     padding: 16px;
   }
 }
+
+// Bi-directional support
+
+bdo[dir=rtl] {
+  direction: rtl;
+  unicode-bidi: bidi-override;
+}
+bdo[dir=ltr] {
+  direction: ltr;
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@topherfangio - do we also want this:

```scss
@mixin rtl($prop, $ltr-value, $rtl-value) {
  #{$prop}: $ltr-value;
  unicode-bidi: embed;

  [dir=rtl] & {
    #{$prop}: $rtl-value;
  }

  bdo[dir=rtl] {
    direction: rtl;
    unicode-bidi: bidi-override;
  }
  bdo[dir=ltr] {
    direction: ltr;
    unicode-bidi: bidi-override;
  }

}
```

src/core/style/structure.scss
```diff
@@ -170,3 +170,14 @@ input {
     padding: 16px;
   }
 }
+
+// Bi-directional support
+
+bdo[dir=rtl] {
```
DevVersion: This is really nice :+1:

PR #7368, opened 2016-03-02
feat(chips): Make chips editable

mertdeg2:
26064e3c-d956-11e5-83e1-fcc6c644fc6f


ThomasBurleson:
@DevVersion - Can you review this... ?

src/components/chips/chips-theme.scss
```diff
@@ -15,6 +15,11 @@ md-chips.md-THEME_NAME-theme {
         color: '{{primary-contrast}}';
       }
     }
+
+    &.md-chip-editing {
+      background:transparent;
```
DevVersion: Nitpicky: Can you space the property from the value?

src/components/chips/js/chipController.js
```diff
@@ -0,0 +1,174 @@
...
+  /**
+   * @type {$scope}
+   */
+  this.$scope = $scope;
+
+  /**
+   * @type {$scope}
+   */
+  this.$scope = $scope;
+
```
DevVersion: Above this comment are three `$scope` definitions. - One should be enough?

src/components/chips/js/chipController.js
```diff
@@ -0,0 +1,174 @@
...
+   * @type {boolean}
+   */
+  this.isEditting = false;
+
+  /**
+   * @type {boolean}
+   */
+  this.enableChipEdit = this.$scope.$parent && this.$scope.$parent.enableChipEdit == 'true';
+
+  if(this.enableChipEdit) {
```
DevVersion: Nitpicky: Just be consistent, can you space the brace from the `if`

src/components/chips/js/chipController.js
```diff
@@ -0,0 +1,174 @@
...
+ * Presents an input element to edit the contents of the chip.
+ */
+MdChipCtrl.prototype.goOutOfEditMode = function() {
+  this.isEditting = false;
+  this.$element.removeClass('md-chip-editing');
+  this.getChipContent()[0].contentEditable = 'false';
+  var chipIndex = this.getChipIndex();
+
+  var content = this.getContentElement().text();
+  if (content === "") {
```
DevVersion: I think `if (content) {` would be more consistent and clear.

src/components/chips/js/chipsController.js
```diff
@@ -50,6 +51,26 @@ function MdChipsCtrl ($scope, $mdConstant, $log, $element, $timeout) {
   /** @type {boolean} */
   this.hasAutocomplete = false;
 
+  /** @type {string} */
+  this.$scope.enableChipEdit = $attrs['mdEnableChipEdit'] || '';
```
DevVersion: This variable should be automatically injected to scope. So just use the variable `$scope.mdEnableChipEdit`, it's already interpolated.

Use it like so:
```js
$scope.mdEnableChipEdit = $mdUtil.parseAttributeValue($scope.mdEnableChipEdit);
```

src/components/chips/js/chipController.js
```diff
@@ -0,0 +1,174 @@
...
+    this.$scope.$parent.removeChipAtIndex(chipIndex);
+  } else {
+    this.$scope.$parent.updateChipContents(
```
DevVersion: I'm not a fan of using `$scope.$parent`, it's really un-flexible / dynamic. Better inject the `chips` controller into the `chip` controller, by using the `require` property of a directive.

src/components/chips/js/chipsController.js
```diff
@@ -50,6 +51,26 @@ function MdChipsCtrl ($scope, $mdConstant, $log, $element, $timeout) {
...
+    }
+  }.bind(this);
+
+  /** @type {function} */
+  this.$scope.removeChipAtIndex = function(chipIndex){
+    this.removeChip(chipIndex);
+  }.bind(this);
+
+  /** @type {function} */
+  this.$scope.getSelectedChipIndex = function(){
```
DevVersion: When using the `injected` controller instead of the `$scope.$parent` (as above said), then you should append these methods to the prototype.

PR #7369, opened 2016-03-02
fix(release): updates the release script to prevent git conflicts

robertmesserle:
@ThomasBurleson @jelbourn

### DIFF
--- a/release.js
+++ b/release.js
@@ -19,11 +19,11 @@
   var dryRun;
 
   header();
-  write('Is this a dry-run? {{"[yes/no]".cyan}} ');
+  write(`Is this a dry-run? ${"[yes/no]".cyan} `);
   dryRun = prompt() !== 'no';
 
   if (dryRun) {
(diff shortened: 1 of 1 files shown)


release.js
```diff
@@ -19,11 +19,11 @@
   var dryRun;
 
   header();
-  write('Is this a dry-run? {{"[yes/no]".cyan}} ');
+  write(`Is this a dry-run? ${"[yes/no]".cyan} `);
```
ThomasBurleson: This uses the support for template strings in NodeJS 4.2.x ?

ThomasBurleson:
@robertmesserle - are these changes required due to the upgrade to NodeJS 4.2.x?

robertmesserle:
No they are unrelated.

release.js
robertmesserle: Yes, it uses the v4 features in place of the manual method I used before.

PR #7379, opened 2016-03-02
feat(chips): Make chips editable

mertdeg2:
26064e3c-d956-11e5-83e1-fcc6c644fc6f

### DIFF
--- a/src/components/chips/js/chipController.js
+++ b/src/components/chips/js/chipController.js
@@ -103,6 +103,8 @@ MdChipCtrl.prototype.getChipIndex = function() {
  * Presents an input element to edit the contents of the chip.
  */
 MdChipCtrl.prototype.goOutOfEditMode = function() {
+  if (!this.isEditting) return;
+
   this.isEditting = false;
   this.$element.removeClass('_md-chip-editing');
   this.getChipContent()[0].contentEditable = 'false';
@@ -115,7 +117,7 @@ MdChipCtrl.prototype.goOutOfEditMode = function() {
         this.getContentElement().text()
     );
   } else {
-    this.parentController.removeChip(chipIndex);
+    this.parentController.removeChipAndFocusInput(chipIndex);
   }
 };
 
--- a/src/components/chips/js/chipController.js
+++ b/src/components/chips/js/chipController.js
@@ -53,7 +53,7 @@ function MdChipCtrl ($scope, $element, $mdConstant, $timeout) {
 /**
  * @param {MdChipsCtrl} controller
  */
-MdChipCtrl.prototype.setParentController = function(controller) {
+MdChipCtrl.prototype.init = function(controller) {
   this.parentController = controller;
   this.enableChipEdit = this.parentController.enableChipEdit;
 
--- a/src/components/chips/js/chipController.js
+++ b/src/components/chips/js/chipController.js
@@ -1,0 +1,187 @@
+angular
+  .module('material.components.chips')
+  .controller('MdChipCtrl', MdChipCtrl);
+
+/**
+ * Controller for the MdChip component. Responsible for handling keyboard
+ * events and editting the chip if needed.
+ *
+ * @param $scope
+ * @param $element
+ * @param $mdConstant
+ * @param $timeout
+ * @constructor
+ */
+function MdChipCtrl ($scope, $element, $mdConstant, $timeout) {
+  /**
+   * @type {$scope}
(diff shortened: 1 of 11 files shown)


mertdeg2:
Hi @DevVersion sorry I messed up something with my pull request again. It was showing others changes. I have created this one, and addressed your feedback.

Could you take a look at this when you get the chance?

src/components/chips/js/chipController.js
```diff
@@ -0,0 +1,187 @@
...
+   * @type {boolean}
+   */
+  this.enableChipEdit = false;
+}
+
+
+/**
+ * @param {MdChipsCtrl} controller
+ */
+MdChipCtrl.prototype.setParentController = function(controller) {
```
DevVersion: I suggest calling that `init()` - currently the function name is not related to the things what happen in there.

DevVersion:
Besides one suggestion, LGTM

src/components/chips/js/chipController.js
mertdeg2: I have changed the function name.

src/components/chips/js/chipController.js
```diff
@@ -0,0 +1,187 @@
...
+  var chipIndex = this.getChipIndex();
+
+  var content = this.getContentElement().text();
+  if (content) {
+    this.parentController.updateChipContents(
+        chipIndex,
+        this.getContentElement().text()
+    );
+  } else {
+    this.parentController.removeChip(chipIndex);
```
DevVersion: Just got into manual testing again, and I figured a few focus issues. Use here: 
```js   
 this.parentController.removeChipAndFocusInput(chipIndex);
```

src/components/chips/js/chipController.js
```diff
@@ -0,0 +1,187 @@
...
+  this.$element.removeClass('_md-chip-editing');
+  this.getChipContent()[0].contentEditable = 'false';
+  var chipIndex = this.getChipIndex();
+
+  var content = this.getContentElement().text();
+  if (content) {
+    this.parentController.updateChipContents(
+        chipIndex,
+        this.getContentElement().text()
+    );
```
DevVersion: Here, if you update the text / model, then the `body` will retrieve the focus, and that's weird. Use here:

```js
    this.$mdUtil.nextTick(function() {
      this.parentController.focusChip(chipIndex);
    }.bind(this));
```
Remember to inject `$mdUtil`
The delay of the next tick is needed, because otherwise the body will be focused after our focus completed. So we should wait the tick.

src/components/chips/js/chipController.js
mertdeg2: I have made this change.

src/components/chips/js/chipController.js
mertdeg2: I wasn't able to reproduce this. Which browser are you using? Can you elaborate what you mean by body will retrieve focus?

Addition of this causes more problems. If the user clicks on another chip while editing current one, the blur function also calls goOutOfEditMode. And chip retains focus even though the user wanted to focus another chip.

DevVersion: Indeed, that seems to be a create a new problem, when clicking on another chip while editing. 

I will tell you at first the steps to reproduce.
- Select Input
- Backspace to Focus Chip (ready to delete)
- Press Space
- Write something new
- Press enter to update chip
- And then press space again to edit again
- But here it **breaks** - because body will be selected

I've just find a way to fix that issue, without introducing a new problem (as you mentioned, when clicking on another chip)

```js
    this.$mdUtil.nextTick(function() {
      if (this.parentController.selectedChip === chipIndex) {
        this.parentController.focusChip(chipIndex);
      }
    }.bind(this));
```

It's important that the check also runs after the tick.

mertdeg2: Good catch! I have added this code to my changelist.

On Wed, Mar 2, 2016 at 12:21 PM, Paul Gschwendtner <notifications@github.com

DevVersion: Done, that's all I found. :+1:

DevVersion:
@mertdeg2 Can you squash your commits into one? 

I suggest the following `git alias`
```
[alias]
  squash = "!f(){ git reset --soft HEAD~${1} && git commit --edit -m\"$(git log --format=%B --reverse HEAD..HEAD@{1})\"; };f"
```

PR #7395, opened 2016-03-03
fix(list): make dense and normal showcase responsive.

DevVersion:
Currently the dense / normal demo won't shrink on mobile screen sizes.
After this commit, the two showcases will wrap on a small screen size.

Also notice, the divider is now splitting the toolbar too, instead of only the content.

*FYI:*
 > The diff is really unclear, so I just mention it here, just made the demo root-container using a `row` layout when bigger than breakpoint `-sm`, otherwise it uses a `column` layout.
> Created two containers, which flexes to 50% on `gt-sm`devices, otherwise 100%
> So now every container (two containers) contains now one toolbar and the associated content (normal and dense)

### DIFF
--- a/src/components/list/demoBasicUsage/index.html
+++ b/src/components/list/demoBasicUsage/index.html
@@ -1,129 +1,146 @@
-<div ng-controller="AppCtrl" ng-cloak>
-  <md-toolbar layout="row" class="md-hue-2">
-    <div flex class="md-toolbar-tools">
-      <span>Normal</span>
-    </div>
-    <div flex class="md-toolbar-tools">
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
lgtm

EladBezalel:
I didn't want the divider to split the toolbar too but that's nvm.

thanks. lgtm

DevVersion:
@EladBezalel Yes, extra mentioned that because it was some kind of a side-effect. Just kept it, because otherwise there would be a much more complex flex layout. So I think it still looks fine.

PR #7402, opened 2016-03-03
fix(progressCircular): switch to using SVG rotation

crisbeto:
Internet Explorer seems to have a hard time rotating SVG paths using `transform`. This causes a slight wobble that ruins the animation. Below is an example of the wobble (left is CSS rotation, right is SVG):

ffs

This PR switches to using SVG transforms and `$$rAF` to animate the rotation.

### DIFF
--- a/src/components/progressCircular/js/progressCircularDirective.js
+++ b/src/components/progressCircular/js/progressCircularDirective.js
@@ -47,17 +47,15 @@ angular
     '$$rAF',
     '$window',
     '$mdProgressCircular',
-    '$interval',
     '$mdUtil',
     '$log',
     MdProgressCircularDirective
   ]);
 
-function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $interval, $mdUtil, $log) {
+function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdUtil, $log) {
   var DEGREE_IN_RADIANS = $window.Math.PI / 180;
   var MODE_DETERMINATE = 'determinate';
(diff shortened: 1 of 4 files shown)


src/components/progressCircular/js/progressCircularDirective.js
```diff
@@ -158,7 +146,7 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $inter
       if (animateTo === animateFrom) {
         path.attr('d', getSvgArc(animateTo, diameter, pathDiameter, rotation));
       } else {
-        (function animation() {
+        $$rAF(function animation() {
```
DevVersion: :+1: - Finally haha :smile:

crisbeto: Oh, I though that before you were referring to the case a little bit above where `animateTo === animateFrom`.

src/components/progressCircular/js/progressCircularDirective.js
```diff
@@ -193,11 +181,51 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $inter
       endIndeterminate = -temp;
     }
 
-    function cleanupIndeterminate() {
+    function startIndeterminateAnimation() {
+      if (!interval) {
+        var startTime = new $window.Date();
+        var animationDuration = $mdProgressCircular.rotationDurationIndeterminate;
+        var radius = getSize(scope.mdDiameter)/2;
```
DevVersion: To be consistent, please space the arithmetic operator.

PR #7408, opened 2016-03-04
fix(dialog): moved bottom focus trap to be a sibling

### LINKED ISSUE #7407: dialog: focus trap in FF seems to be broken in demos
Quoting @ErinCoughlan 
> On the demo site in Firefox, focus trap seems to be broken for alert and confirm dialogs.

EladBezalel:
- bottomFocusTrap was appended to the dialog, in FF it seems like it didn't trapped the focus, ,moving it to be a sibling solves it.

@jelbourn please take a look 

fixes #7407


jelbourn:
LGTM

PR #7409, opened 2016-03-04
fix(bidi-demos): fixed broken demos

### LINKED ISSUE #7399: Bi-di issues with demos
After the commits for improved Bi-Directional support, the following demos still need work:

screen shot 2016-03-02 at 8 52 12 am

screen shot 2016-03-02 at 8 52 30 am

EladBezalel:
- fixed `rtl-prop` mixin to override `ltr` property when applied
- fixed checkbox demos
- fixed datepicker triangle alignment
- fixed divider demos
- fixed inset divider in list
- fixed tooltip demos

fixes #7399

### DIFF
--- a/src/core/style/mixins.scss
+++ b/src/core/style/mixins.scss
@@ -114,6 +114,7 @@
 @mixin rtl-prop($ltr-prop, $rtl-prop, $value) {
   #{$ltr-prop}: $value;
   [dir=rtl] & {
+    #{$ltr-prop}: auto;
     #{$rtl-prop}: $value;
   }
 }
--- a/src/components/checkbox/demoBasicUsage/style.css
+++ b/src/components/checkbox/demoBasicUsage/style.css
@@ -5,9 +5,7 @@ div.flex-xs {
 .checkboxDemo1 div {
   clear: both;
 }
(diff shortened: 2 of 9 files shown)


src/core/style/mixins.scss
```diff
@@ -114,6 +114,7 @@
 @mixin rtl-prop($ltr-prop, $rtl-prop, $value) {
   #{$ltr-prop}: $value;
   [dir=rtl] & {
+    #{$ltr-prop}: auto;
```
DevVersion: Not every CSS property supports `auto` as a value?

EladBezalel: You're right, i thought that when a property is getting a wrong value it behaves as the initial behavior, silly me!

EladBezalel: As [initial](http://caniuse.com/#feat=css-initial-value) isn't supported on IE i think i might have to do something like:
```css
#{$ltr-prop}: auto;
#{$ltr-prop}: initial;
```

DevVersion: @EladBezalel It seems like the PR is already merged, only with `auto` as value.

PR #7414, opened 2016-03-04
fix(progressCircular): SVG transforms for indeterminate mode and performance improvements

crisbeto:
* Avoids unnecessary digest cycles during the indeterminate animation by using `setInterval`, instead of `$interval`.
* Avoids creating multiple `Date` objects by using the performance API and the timestamp from `$$rAF`.
* Avoids unnecessary DOM manipulation and layout by only setting the element dimensions when the diameter changes.
* Avoids unnecessary layout by pulling the SVG element out of the flow.
* Fixes wobble in indeterminate animation on IE by switching to animating the SVG transform attribute. Below is an example of the wobble (left is CSS rotation, right is SVG):
ffs

### DIFF
--- a/src/components/progressCircular/js/progressCircularDirective.js
+++ b/src/components/progressCircular/js/progressCircularDirective.js
@@ -47,17 +47,16 @@ angular
     '$$rAF',
     '$window',
     '$mdProgressCircular',
-    '$interval',
     '$mdUtil',
+    '$interval',
     '$log',
     MdProgressCircularDirective
   ]);
 
-function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $interval, $mdUtil, $log) {
+function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdUtil, $interval, $log) {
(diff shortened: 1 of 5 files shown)


src/components/progressCircular/js/progressCircularDirective.js
```diff
@@ -93,73 +91,66 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $inter
...
       if (mode !== MODE_DETERMINATE && mode !== MODE_INDETERMINATE) {
-        cleanupIndeterminate();
+        cleanupIndeterminateAnimation();
         element.addClass('ng-hide');
       } else {
         element.removeClass('ng-hide');
 
-        if (mode === MODE_INDETERMINATE){
-          if (!interval) {
-            interval = $interval(
```
DevVersion: Just curious, you said, that you wan't to stop the digests for $interval. I would rather use 
```js
$interval(fn, delay, 0, false);
```

crisbeto: Good point, I'll put it back in.

PR #7420, opened 2016-03-05
fix(progressCircular) aria in indeterminate, timestamp polyfill, IE10 visibility, test cleanup

crisbeto:
* Adds a polyfill for `performance.now`.
* Doesn't set the `aria-valuenow` attribute in indeterminate mode.
* Fixes the circle being invisible in IE10.
* Cleans up the `progressCircular` testing setup and adds a test for the stroke width.

### DIFF
--- a/src/components/progressCircular/js/progressCircularDirective.js
+++ b/src/components/progressCircular/js/progressCircularDirective.js
@@ -92,8 +92,8 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdUti
   };
 
   function MdProgressCircularLink(scope, element) {
-    var svg = element[0].querySelector('svg');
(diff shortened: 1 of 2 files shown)


src/components/progressCircular/js/progressCircularDirective.js
```diff
@@ -123,35 +126,38 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdUti
...
-      element.css({
-        width: diameter + 'px',
-        height: diameter + 'px'
-      });
     });
 
     function renderCircle(animateFrom, animateTo, easing, duration, rotation) {
       var id = ++lastAnimationId;
-      var startTime = $window.performance.now();
+      var startTime = getTimestamp();
```
DevVersion: Nice, this fixes the issue on jenkins.

PR #7421, opened 2016-03-05
feat(autocomplete): add public method to clear the cache.

### LINKED ISSUE #7411: Allow to clear md-autocomplete cache
I would like to use the caching provided by `md-autocomplete` but need to clear/invalidate the cache upon certain events.

I propose to let the application remain in charge of when the cache should be cleared, so it could be done on timeout or (as in our case) when something specific occurs.

Example usage:
```<md-autocomplete md-cache-invalid="cacheInvalid">```

Where `cacheInvalid` is a scope variable that the application sets to true as needed.  `md-autocomplete` would see it was now true, clear the cache and set it to false.

If anyone has a workaround, please post it.  I'm able to get the `$mdAutocompleteCtrl` object but none of the public functions are of help here.

### PR DESCRIPTION
**Pull request by DevVersion**
This adds a public method to the controller, which allows the user to dynamically clear the cache.
Also added a complex test which tests the cache functionality by using the new method.

Fixes #7411

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -8,7 +8,7 @@ var ITEM_HEIGHT   = 41,
     INPUT_PADDING = 2; // Padding provided by `md-input-container`
 
 function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming, $window,
-                             $animate, $rootElement, $attrs, $q) {
+                             $animate, $rootElement, $attrs, $q, $mdComponentRegistry) {
   //-- private variables
   var ctrl                 = this,
       itemParts            = $scope.itemsExpr.split(/ in /i),
@@ -17,6 +17,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       cache                = {},
       noBlur               = false,
       selectedItemWatchers = [],
+      removeFromRegistry   = null,
       hasFocus             = false,
       lastCount            = 0,
       promiseFetch         = false;
@@ -64,6 +65,17 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     $mdUtil.initOptionalProperties($scope, $attrs, { searchText: null, selectedItem: null });
     $mdTheming($element);
     configureWatchers();
+
+    if ($scope.componentId) {
+      removeFromRegistry = $mdComponentRegistry.register({
+        // Restricted API interface for the component registry.
+        // We don't want to provide the whole controller as API.
+        clearCache: function() {
+          cache = [];
+        }
+      }, $scope.componentId);
+    }
+
     $mdUtil.nextTick(function () {
       gatherElements();
       moveDropdown();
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

**Comment by EladBezalel** [comment]
So if the user want's to clear the cache he should get the controller from the element and call this action? I'm this is not a good solution

**Comment by DevVersion** [comment]
No that's not the complete solution. 
It's just an <kbd>API</kbd> access / public method for the controller, as same as `select(index)`, `clear()`, `registerSelectedItemWatcher(cb)` etc.

File: src/components/autocomplete/js/autocompleteController.js [hunk 1]
```diff
@@ -64,6 +66,18 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
     $mdTheming($element);
     configureWatchers();
+
+    if ($scope.componentId) {
+      removeFromRegistry = $mdComponentRegistry.register({
+        // Restricted API interface for the component registry.
+        // We don't want to provide the whole controller as API.
+        clearCache: function() {
+          // Run the clearCache method in the controller context.
+          ctrl.clearCache();
```
DevVersion: @EladBezalel Please review again - and about this specific line:
-  It is needed to provide a new object with all the needed functions, which should accessable in the component instance.

Also I think, you're wondering about the extra `clearCache` method in the controller, but I think it's good to implement this as a new function in the controller. But if you have other thoughts about the `clearCache` controller function, let me know - I'm up to move this in-line.

EladBezalel: Why not just:
```js
{
  clearCache: ctrl.clearCache
}
```

DevVersion: Had that before, and it worked, but then I won't be able to test that in jasmine, because the spy can't access this one. But I think, I better change that back and remove that particular test. The behavior of the component registry is covered in other tests too, so it will be still fine.

src/components/autocomplete/js/autocompleteService.js
```diff
@@ -0,0 +1,64 @@
...
+    };
+
+    /**
+     * Deferred lookup of component instance using $component registry
+     */
+    function waitForInstance() {
+      return $mdComponentRegistry
+        .when(handle)
+        .then(function(it) {
+          instance = it;
```
EladBezalel: i guess this is copied from sidenav right?

DevVersion: Yes, the deferred lookup is a useful function, so I thought I'm implementing that in `$mdAutocomplete` too.

src/components/autocomplete/js/autocompleteController.js
EladBezalel: Our code shouldn't be written to make jasmine work, it supposed to be opposite. 
I think the component registry is important and i do think there're ways to bypass that

EladBezalel:
I have a feeling that this is a code replica and there's a better way of making it more generic, something like `interimElement`

DevVersion:
@EladBezalel Re-structuring the way of providing `$mdComponentRegistry` instances, may be another PR, which should be done after this implementation.

src/components/autocomplete/js/autocompleteController.js
DevVersion: Yes, you're right. I decided to you the in-line implementation. Because a controller function would be unnecessary, when `$mdAutocomplete` is present.

DevVersion:
This PR needs to be reworked after #7736

DevVersion:
Closing this, as we need to wait for $$mdAPI to be accepted / discussed.

clshortfuse:
What's the status of this feature? Abandoned?

PR #7423, opened 2016-03-05
fix(rtl-mixin): changed ltr override to initial

EladBezalel:
- Changed ltr override to `initial` as most of the properties doesn't support `auto`
- Added `auto` fallback for IE as it doesn't support [initial](http://caniuse.com/#feat=css-initial-value)

related #7409

### DIFF
--- a/src/core/style/mixins.scss
+++ b/src/core/style/mixins.scss
@@ -114,7 +114,12 @@
 @mixin rtl-prop($ltr-prop, $rtl-prop, $value) {
   #{$ltr-prop}: $value;
   [dir=rtl] & {
+    // fallback for auto
+    #{$ltr-prop}: 0;
+    // fallback for IE
     #{$ltr-prop}: auto;
+    #{$ltr-prop}: initial;
+
     #{$rtl-prop}: $value;
   }
 }


src/core/style/mixins.scss
```diff
@@ -114,7 +114,12 @@
 @mixin rtl-prop($ltr-prop, $rtl-prop, $value) {
   #{$ltr-prop}: $value;
   [dir=rtl] & {
+    // fallback for auto
+    #{$ltr-prop}: 0;
+    // fallback for IE
     #{$ltr-prop}: auto;
+    #{$ltr-prop}: initial;
```
ThomasBurleson: Does this not overwrite L118 ?

EladBezalel: This is by intention,
What we really want: `initial`
What we feel meh about: `auto`
What we settle for: `0`

ThomasBurleson: Is `0` always overwritten by `initial` ?

EladBezalel: On chrome and firefox yes, on IE no

PR #7427, opened 2016-03-06
fix(theming): only insert default theme if it's still used

### LINKED ISSUE #7425: default theme is injected even when we use setDefaultTheme with another value
In function generateAllThemes we have teh following code

```javascript
 angular.forEach(THEMES, function(theme) {
    if (!GENERATED[theme.name]) {
      generateTheme(theme.name);
    }
  });
```

this code is called also for 'default' theme even when actually i used setDefaultTheme with another theme name.
This should not happen.
The result is that have twice style tags as i would really want.

DevVersion:
Currently if I specify a default theme by using #setDefaultTheme, then the `default` theme will be still inserted.

Fixes #7425

### DIFF
--- a/src/core/services/theming/theming.js
+++ b/src/core/services/theming/theming.js
@@ -473,7 +473,7 @@ function parseRules(theme, colorType, rules) {
 var rulesByType = {};
 
 // Generate our themes at run time given the state of THEMES and PALETTES
-function generateAllThemes($injector) {
+function generateAllThemes($injector, $mdTheming) {
   var head = document.head;
   var firstChild = head ? head.firstElementChild : null;
   var themeCss = $injector.has('$MD_THEME_CSS') ? $injector.get('$MD_THEME_CSS') : '';
@@ -528,7 +528,7 @@ function generateAllThemes($injector) {
   if (generateOnDemand) return;
 
   angular.forEach(THEMES, function(theme) {
-    if (!GENERATED[theme.name]) {
+    if (!GENERATED[theme.name] && !(theme.name !== $mdTheming.defaultTheme() && theme.name === 'default')) {
       generateTheme(theme.name, nonce);
     }
   });


src/core/services/theming/theming.js
```diff
@@ -528,7 +528,7 @@ function generateAllThemes($injector) {
   if (generateOnDemand) return;
 
   angular.forEach(THEMES, function(theme) {
-    if (!GENERATED[theme.name]) {
+    if (!GENERATED[theme.name] && !(theme.name !== $mdTheming.defaultTheme() && theme.name === 'default')) {
```
EladBezalel:
```js
!(theme.name !== $mdTheming.defaultTheme() && theme.name === 'default')
```
is equals to
```js
theme.name !== 'default'
```

DevVersion: No, that's wrong. Imagine you have set the default theme to `myDefaultTheme`
And it's iterating through the themes, and their is `default` now in progress.

```js
theme.name !== 'default'
// 'default' !== 'default' === FALSE
```
So in this case it works.

But now, if you have no default theme set, then it will never insert the `default` theme and the theming breaks.

EladBezalel: it's just that 
```js
!(theme.name !== $mdTheming.defaultTheme() && theme.name === 'default')
```

doesn't makes sense, you are checking if `theme.name` isn't equals to something AND equals to something else, if you know that you expect it to be `default` than why do you even bother to check if it's not equals to something else

DevVersion: Oh it makes sense. An `if-expression` won't continue the check if it already failed. 
For example:
```js
true && false && true
-> The check will immediately cancel itself, because the second `AND` is false.
```

The theme should be skipped if the theme name is equal to  `default` and if the default theme has been changed. 

But maybe we can find a more elegant check, but your logic in the check isn't valid.

EladBezalel: I'm trying to figure when your addition will be valid,
```js
!(theme.name !== $mdTheming.defaultTheme() && theme.name === 'default')
```
let's say `theme.name !== $mdTheming.defaultTheme()` is true, than for sure `theme.name === 'default'` is false. - invalid
`theme.name !== $mdTheming.defaultTheme()` is false and `theme.name === 'default'` is false (or true) - invalid

it will only be valid if both are `true` which means `$mdTheming.defaultTheme() !== 'default'`

DevVersion: Let's go through a few cases.

Case 1
-----
We have the following mocked theme array
```js
THEMES = ['default', ' blue']
```
And my default theme is set to `blue`

So the iteration will check:
```js
// Replicates the iteration - case for `default`
var themeName = THEMES[0]; // `default`
// Negate the check, so if both conditions are true, 
// then the theme won't be inserted into the DOM
if (!(
  themeName !== 'blue' // true -> continue with the checks `AND`
  && themeName === 'default' // true -> both conditions are true
)
```
So in this case, only the theme `blue` will inserted into the DOM - <kbd>Valid</kbd>

Case 2
---
We have the following mocked theme array
```js
THEMES = ['default', ' blue']
```
And my default theme is set to `default` (default behavior)

So the iteration will check:
```js
// Replicates the iteration - case for `default`
var themeName = THEMES[0]; // `default`
// Negate the check, so if both conditions are true, 
// then the theme won't be inserted into the DOM
if (!(
  themeName !== 'default' // false -> cancel check and and return false
  && themeName === 'default' // Not reachable
)
// The check will pass, because of the negated expression 
// !(false) ==> true
```
So in this case, both themes will be inserted to the DOM

EladBezalel:
First of all let's use discrete to simplify the conditions

``` js
!(a != z && b == z) 
// => 
(a == z || b != z)
```

Ok so let's go through your cases.

Case 1
-----
We have the following mocked theme array
```js
THEMES = ['default', ' blue']
```
And my default theme is set to `blue`

So the iteration will check:
```js
// Replicates the iteration - case for `default
var themeName = THEMES[0]; // `default`
// if one conditions is true, the theme won't be inserted into the DOM
if (themeName === 'blue' // false
  || themeName !== 'default' // false -> both conditions are false
)
```
So in this case, only the theme `blue` will inserted into the DOM - <kbd>Valid</kbd>

Case 2
---
We have the following mocked theme array
```js
THEMES = ['default', ' blue']
```
And my default theme is set to `default` (default behavior)
So the iteration will check:
```js
// Replicates the iteration - case for `default`
var themeName = THEMES[0]; // `default`
// if one conditions is true, the theme won't be inserted into the DOM
if (themeName === 'default' // true
  || themeName !== 'default' //  false
)
// The check will pass
```
So in this case, both themes will be inserted to the DOM

Finally
---
As you can see, in both cases 
``` js
themeName !== 'default' //  false
```
Than it's clearly redundant..

DevVersion:
But imagine, as I said these are examples have a custom default theme set. So if you haven't changed the default theme, but you registered another theme.

So now take a look at `Case 2`.  If we don't check both, as in mine implementation, then you'll have the following execution.

```js
// Replicates the iteration - case for `default`
var themeName = THEMES[0]; // `default`

// This is the check, which can be simplified in your opinion.
if (themeName !== 'default' // FALSE
)
// The check won't pass
```

So this case won't be valid. Because here `'default'` is our default theme and won't be injected into the DOM, because the check didn't pass.

We won't add the iterated theme when:
-The default theme has been changed by the user
- And the iterated theme is `'default`

EladBezalel:
I don't say you should use:
```js
// Replicates the iteration - case for `default`
var themeName = THEMES[0]; // `default`

// This is the check, which can be simplified in your opinion.
if (themeName !== 'default' // FALSE
)
// The check won't pass
```
i say you should use:
``` js
// Replicates the iteration - case for `default`
var themeName = THEMES[0]; // `default`
// if one conditions is true, the theme won't be inserted into the DOM
if (themeName === 'default')
```

ThomasBurleson:
@DevVersion, @EladBezalel - what is the status here? It appears that Elad's logic is valid. Will it pass all the test cases ?

DevVersion:
@ThomasBurleson We're still discussing. 

I mean Elad says that, I should use this code:
```js
if (themeName === 'default') // theme.name === $mdTheming.defaultTheme()
```
But isn't valid. Imagine if the user changed the default theme, then it's no longer `default`.

ThomasBurleson:
**tl;dr**
@DevVersion - what is the status of this PR?

DevVersion:
@ThomasBurleson We never had our dicussion. @EladBezalel Let's talk about that soon okay?

DevVersion:
Closing this, as we both have different opinions about the logic. Let's revisit this anytime later.

PR #7428, opened 2016-03-06
fix(chips): apply right theme styles

### LINKED ISSUE #7104: MdChips with not applying theme to input
When using a `md-chips` ~~with autocomplete and non-floating~~ input there is no theme applied to the input.

DevVersion:
Currently there is incorrect theming on the chips component.
- Wrong Border Shadow Color
- No Theming for the Icon (`delete`)
- Input had no theming

Fixes #7104


DevVersion:
This PR is partially depending on  #7432, because of the mixin fix.

PR #7430, opened 2016-03-06
fix(list): ui-sref support for md-secondary

ruant:
md-secondary using ui-sref wasn't being moved outside the primary container


DevVersion:
@ruant About which version are you talking? If you talk about `master` / `HEAD`, then please file an issue, because then it won't be only `ui-sref`

ruant:
I've added a issue, #7431

DevVersion:
Thanks :+1:

PR #7432, opened 2016-03-06
fix(input, mixing): fix input placeholder mixing and fix label theming.

### LINKED ISSUE #7429: input: input placeholder mixin isn't working
The theming for the floating label is different to the normal placeholder

That's caused by the placeholder mixin, because it isn't working.
 So the placeholder is always using the default placeholder color `darkgrey`

DevVersion:
At the moment the placeholder mixing isn't working, so the the placeholder is always using the browsers default color (`darkgrey`).

This commit fixes the invalid mixin for the placeholder, so the floating label and normal placeholder won't look different anymore.

Also removed the `text-shadow`, because in the specs is no inset shadow, this makes the label look weird.

So now the labels will look as same as in the specs, except that the `foreground-3` color is currently not correct. (in the specs - because it has a opacity of 0.26, but it should have 0.38)

Fixes #7429

### DIFF
--- a/src/components/input/input-theme.scss
+++ b/src/components/input/input-theme.scss
@@ -3,7 +3,6 @@ md-input-container.md-THEME_NAME-theme {
     @include input-placeholder-color('{{foreground-3}}');
     color: '{{foreground-1}}';
     border-color: '{{foreground-4}}';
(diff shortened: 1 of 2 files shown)


EladBezalel:
LGTM

PR #7433, opened 2016-03-06
fix(theming): updates text opacities to the newest specs

DevVersion:
This updates the foreground opacities to be aligned with the specs.

https://www.google.com/design/spec/style/color.html#color-text-background-colors

As discussed with @EladBezalel, created a PR for the wrong opacities.


EladBezalel:
@DevVersion Please check ci

DevVersion:
@EladBezalel Thanks, didn't notice it yet. :+1:

Issue #6796, opened 2016-01-21
md-icon / accesibility / validation usage with svg (i.e: https://www.google.com/design/icons/) duplicate id's issue


noamamit92:
When using the same icon twice in a page the html validation fails due to 2 elements having the same ID. 

Since md-icon adds the svg element with the attrbute : md-svg-icon which accepts an id as an input, Is it possible to add the svg element without it's id  - so the validation error wont occur?

a codepen :
 
http://codepen.io/noamamit92/pen/adGjYp#0

notice that if you run the accessibility Audit using your chrome accessibility plugin you will get a duplicate id error for the g element under the svg

ThomasBurleson:
Please provide a CodePen or Plunkr that demonstrates this issue. Here are some starter demo templates that you can use/fork:

*  CodePen template: http://codepen.io/pen?template=bEGJdd
*  Plunkr template: http://plnkr.co/edit/yP0gB9iTacIQb48thud8?p=preview.

noamamit92:
I added a demo:

http://codepen.io/noamamit92/pen/adGjYp#0

PR #7440, opened 2016-03-07
fix(md-icon): accessibility issue with unique ids

david-gang:
fix(md-icon): accessibility issue with unique ids

When the same svg icon is placed multiple times in the same page, it raises an aria warning that every element should have an  [unique id](https://www.w3.org/TR/WCAG20-TECHS/F77).
The solution is to remove the id from the svg and the path children.

closes: #6796

### DIFF
--- a/src/components/icon/js/iconService.js
+++ b/src/components/icon/js/iconService.js
@@ -104,7 +104,7 @@
     * @param {string} id Icon name/id used to register the iconset
     * @param {string} url specifies the external location for the data file. Used internally by `$http` to load the
     * data or as part of the lookup in `$templateCache` if pre-loading was configured.
-    * @param {number=} viewBoxSize Sets the width and height of the viewBox of all icons in the set. 
+    * @param {number=} viewBoxSize Sets the width and height of the viewBox of all icons in the set.
     * It is ignored for icons with an existing viewBox. All icons in the icon set should be the same size.
     * Default value is 24.
     *
@@ -134,7 +134,7 @@
     *
     * @param {string} url specifies the external location for the data file. Used internally by `$http` to load the
     * data or as part of the lookup in `$templateCache` if pre-loading was configured.
-    * @param {number=} viewBoxSize Sets the width and height of the viewBox of all icons in the set. 
+    * @param {number=} viewBoxSize Sets the width and height of the viewBox of all icons in the set.
     * It is ignored for icons with an existing viewBox. All icons in the icon set should be the same size.
     * Default value is 24.
     *
@@ -247,7 +247,7 @@
      config.defaultViewBoxSize = viewBoxSize;
      return this;
    },
-   
+
    /**
     * Register an alias name associated with a font-icon library style ;
     */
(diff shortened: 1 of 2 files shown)


src/components/icon/js/iconService.js
```diff
@@ -422,8 +422,25 @@
...
+
+       //clear id attributes to prevent aria issues
+       var elem = icon.element;
+       elem.removeAttribute('id');
+
+       var children = elem.childNodes;
+
+       for (var i=0; i<children.length;i++) {
+         var child = children[i];
+         if(child.nodeName==='path') {
```
DevVersion: Why only on the `path` node? `circle` etc, can also hold an `id` attribute?

david-gang: You are right. I will remove the if.

DevVersion: @david-gang I suggest, walking recursively through the children and remove the `id` if present.
Like so
```js
(function recursiveIteration(element) {
  angular.forEach(element.childNodes, function(item) {
    item.removeAttribute('id');
    if (item.childNodes) {
      recursiveIteration(item)
    }
  });
})(myIconElement);
```

Or another way, would be the `querySelector`. Like so
```js
angular.forEach(myIconElement.querySelectorAll('[id]'), function(item) {
  item.removeAttribute('id');
});
```

But I'm not sure what's the most elegant / performant solution.

david-gang: @DevVersion The second option looks very nice. It should also be performant because it is a native function and the call stack is flat.

DevVersion: @david-gang Yes definitely, it looks more elegant, but I heard some negative about `attribute selectors` in IE and other browsers.

david-gang: @DevVersion ie is always a pain in the neck :-)
I tried to use jsperf http://jsperf.com/remove-id-by-queryselectorall-vs-recursion but suddenly the site does not work and i cannot edit the test case. Maybe this is due to the fact that i am behind a firewall. The recursion function throwed an exception. I am pretty sure that the queryselectorall is better but i cannot prove it. May you please check on ie what is faster?

david-gang: @DevVersion  I found a similar test case https://jsperf.com/document-vs-element-queryselectorall-performance . queryselectorall looks very performant when it is narrowed on a certain element. In this test case it did 215,335 operations in ie per second

DevVersion: Just updated your test - now it works. See the results here: http://jsperf.com/recursive-iteration-vs-queryselectorall

david-gang: thanks :+1:

DevVersion: No problem. Look here at IE11

image

david-gang: The code is just called once per icon type, unconnected to the number of times i use the icon in the application, and even in ie11 the performance is 10^5 ops per second, so this should be fine.

DevVersion: It's okay :smile: Just wanted to share with you the performance differences.

Issue #3244, opened 2015-06-12
Checkboxes in md-select when multiple selection is possible


FDIM:
Hi,
I am doing a rapid prototyping and I have heard multiple complains/suggestions from colleagues/clients/stakeholders that in select dropdown it is not clear when you can choose multiple items and when only one.

I propose to add a checkbox to each item for md-select[multiple]. While it is simple enough to add, it would be nice if something like this would be included in default distribution.

![image](https://cloud.githubusercontent.com/assets/1985429/8130371/f4136bc4-1111-11e5-95ba-9963ae432cb6.png)

Result in the picture can be achieved with this less snippet:

```   
// multi select dropdown with checkboxes
md-select-menu[multiple] {
    md-option {
        padding-left: 48px;
        &:before {
            transition: 240ms;
            position: absolute;
            top: 15px;
            left: 16px;
            width: 18px;
            height: 18px;
            border: 2px solid;
            border-radius: 2px;
            content: ' ';
            border-color: rgba(0, 0, 0, 0.54);
            z-index: 1;
        }
        &[selected] {
            &:before {
                background-color: rgba(30, 136, 229, 0.87);
                border-color: rgba(30, 136, 229, 0.87);
            }
            &:after {
                -webkit-transform: rotate(45deg);
                transform: rotate(45deg);
                position: absolute;
                top: 17px;
                left: 22px;
                display: table;
                width: 6px;
                height: 12px;
                border: 2px solid;
                border-top: 0;
                border-left: 0;
                content: ' ';
                z-index: 2;
                border-color: rgb(255, 255, 255);
            }
        }
    }
}
```

Unfortunately there is more work involved to support themes.

ahmadhsalim:
Agreed. This should be possible for `md-menu` too.

sitic:
+1, I've received complaints from users about the non-obvious multiple select.

phynweb:
Per @ThomasBurleson, moving comments from [#3610](https://github.com/angular/material/issues/3610) 
*******
A few ideas for enhancing the md-select directive when multiple is true. All use-cases derived from the Option Groups select demo. Codepen with multiple set to true on toppings: http://codepen.io/anon/pen/rVdzzb

While the current selection is quite clear for single selections since the dropdown closes on select and populates the select text with the selection, this is not the case for multiple selections. The color difference between selected and unselected items is a bit too subtle if we contemplate a large list (e.g. 100+ items). A check icon or some other visual cue would be helpful while the user scrolls through the list.

[...]

PR #7455, opened 2016-03-08
feat(select): Adding check boxes to the multi-select menu

DerekLouie:
@ErinCoughlan @jelbourn @KarenParker @gmoothart 

Please Review. 

Adding check boxes to the multi-select menu.

selectcheckboxes

References issue #3244.

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -6,6 +6,107 @@ $checkbox-border-width: 2px !default;
 $checkbox-text-margin: 10px !default;
 $checkbox-top: 12px !default;
 
+// This mixin allows a user to use the md-checkbox css outside of the
+// md-checkbox directive.
+// See src/components/select/select.scss for an example.
+@mixin checkbox-container(
+  $checkedSelector: '.md-checked',
+  $width: $checkbox-width,
+  $height: $checkbox-height,
+  $border-width: $checkbox-border-width,
+  $border-radius: $checkbox-border-radius) {
+  ._md-container {
+    position: absolute;
+    top: 50%;
+    transform: translateY(-50%);
+
+    box-sizing: border-box;
+    display: inline-block;
+
+    width: $width;
+    height: $height;
+    @include rtl(left, 0, auto);
+    @include rtl(right, auto, 0);
+
+    &:before {
+      box-sizing: border-box;
+      background-color: transparent;
+      border-radius: 50%;
+      content: '';
+      position: absolute;
+      display: block;
+      height: auto;
+      left: 0;
+      top: 0;
+      right: 0;
+      bottom: 0;
+      transition: all 0.5s;
+      width: auto;
+    }
+
+    &:after {
+      box-sizing: border-box;
+      content: '';
+      position: absolute;
+      top: -10px;
+      right: -10px;
+      bottom: -10px;
+      left: -10px;
+    }
+
+    .md-ripple-container {
+      position: absolute;
+      display: block;
+      width: auto;
+      height: auto;
+      left: -15px;
+      top: -15px;
+      right: -15px;
+      bottom: -15px;
+    }
+  }
+
+  // unchecked
+  ._md-icon {
+    box-sizing: border-box;
+    transition: 240ms;
+    position: absolute;
+    top: 0;
+    left: 0;
+    width: $width;
+    height: $height;
+    border-width: $border-width;
+    border-style: solid;
+    border-radius: $border-radius;
+  }
+
+  &#{$checkedSelector} ._md-icon {
+    border: none;
+
+    &:after {
+      box-sizing: border-box;
+      transform: rotate(45deg);
+      position: absolute;
+      left: $width / 3;
+      top: $width / 9;
+      display: table;
+      width: $width / 3;
+      height: $width * 2 / 3;
+      border-width: $border-width;
+      border-style: solid;
+      border-top: 0;
+      border-left: 0;
+      content: '';
+    }
+  }
+
+  // disabled
+  &[disabled] {
+    cursor: default;
+  }
+
+}
+
 .md-inline-form {
   md-checkbox {
     margin: 19px 0 18px;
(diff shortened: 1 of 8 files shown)


src/components/select/select-theme.scss
```diff
@@ -75,3 +75,17 @@ md-select-menu.md-THEME_NAME-theme {
     }
   }
 }
+
+md-select-menu[multiple] {
+    md-option {
```
DevVersion: To be consistent, please indent with two spaces, instead of one tab

src/components/select/select.scss
```diff
@@ -228,3 +228,41 @@ md-optgroup {
...
   }
 }
+
+md-select-menu[multiple] {
+    md-option {
+        height: $select-option-height;
+        padding: 0 $select-option-padding 0 ($select-option-padding / 2);
+        padding-left: ($select-option-padding * 2) + ($select-option-padding / 2);
+
+        &:before {
```
DevVersion: @DerekLouie Why are we not using the `md-checkbox` component? I know it will be a bit more complex, to integrate the `md-checkbox` component, but I think this would be better way, instead of mirroring the checkbox. Thoughts?

jelbourn: At the very least it should be reusing CSS from the checkbox.

jelbourn:
The checkbox borders in the screenshot seem exceptionally dark (black?) to me.

src/components/select/select.scss
DerekLouie: @DevVersion 

I believe that this can all be done using CSS and introducing the `md-checkbox` into the mix adds more overhead ,regarding watching values, than necessary.

But you are right, consistency is always appreciated, I will change this to mirror the `md-checkbox` css!

Muchos gracias for the feedback so far, more code to come.

DevVersion: I'm just a bit curious, what happens, when we change something in the `md-checkbox` style, then I guess, mostly the *mirror* checkbox will be forgotten to update.

jelbourn: Agreed- copying is a pain for maintenance. Have you explored refactoring the checkbox css such that you can use the same classes directly?

src/components/select/select-theme.scss
jelbourn: Yes- the indent should be +2 spaces

DerekLouie:
@ErinCoughlan @jelbourn @KarenParker @gmoothart

Please Review.

I have refactored the select menu checkbox CSS to match the md-checkbox css (wherever applicable).

selectcheckboxv2

References issue #3244.

src/components/select/select-theme.scss
DerekLouie: ahh apologies, my vimrc did not recognize scss and defaulted to 4 spaces. Corrected.

DerekLouie:
Going to explore refactoring the checkbox css such that you can use the same classes directly.

Question @jelbourn: Where would prefer I put this css that will be used in both the select, checkbox, and potentially other components?

jelbourn:
@DerekLouie there should be one single place where the styles for a checkbox lives, and then each other component should consume those styles.

DerekLouie:
Hello All,

I have done some scss wizardry to make the applicable md-checkbox styles re-usable in md-select (and beyond?).

Please Review.

Screenshot with RTL look and feel included for completeness. 

checkboxesv3

KarenParker:
LGTM

ErinCoughlan:
LGTM

src/components/checkbox/checkbox.scss
```diff
@@ -6,46 +6,12 @@ $checkbox-border-width: 2px !default;
...
-    }
-
-    &:not(.md-checked) {
-      ._md-container:before {
-        background-color: rgba(0, 0, 0, 0.12);
-      }
-    }
-  }
-
+@mixin checkbox-container(
```
jelbourn: Add comment explaining what this mixin does

src/components/checkbox/checkbox.scss
```diff
@@ -112,15 +74,31 @@ md-checkbox {
...
-    border-radius: $checkbox-border-radius;
+    border-radius: $border-radius;
   }
 
-  &.md-checked ._md-icon {
+  &#{$checkedSelector} ._md-icon {
     border: none;
+
+    &:after {
+        box-sizing: border-box;
```
jelbourn: Too much indent

src/components/select/select.scss
```diff
@@ -228,3 +231,17 @@ md-optgroup {
     border: 1px solid #fff;
   }
 }
+
+md-select-menu[multiple] {
+    md-option[md-checkbox-enabled] {
+        @include rtl(padding-left, $select-option-padding * 2.5, $select-option-padding);
```
jelbourn: Too much indent

src/components/select/select.js
```diff
@@ -776,6 +776,11 @@ function OptionDirective($mdButtonInkRipple, $mdUtil) {
     var optionCtrl = ctrls[0];
     var selectCtrl = ctrls[1];
 
+    if (selectCtrl.isMultiple) {
+      element.attr('md-checkbox-enabled', '');
+      element.prepend('<div class="_md-container"><div class="_md-icon"></div></div>');
```
jelbourn: Instead of using a string here, define a single const that contains this "template element" and then clone it in this `prepend` call; element cloning will be significantly faster than parsing the HTML and re-creating the DOM structure every time.

src/components/select/select.js
```diff
@@ -776,6 +776,11 @@ function OptionDirective($mdButtonInkRipple, $mdUtil) {
...
+    }
```
jelbourn: There should be a (hopefully simple) unit test for this.

ThomasBurleson:
@DerekLouie - we definitely need unit test(s) with this PR.

DerekLouie:
Hello All,

I have added unit tests. I ran into an issue where *element*.find('._md-container') returned an empty sqlite object, my wild shot in the dark is that the [underscore in the class name](https://developer.mozilla.org/en-US/docs/Web/Guide/CSS/Underscores_in_class_and_ID_Names) messes something up?

I ended up for looping and using some good old native JS to verify the md-checkbox markup is correctly added.

Please Review.

Best,

Derek

jelbourn:
@DerekLouie angular.element's `.find` function only works for class names (which is surprising if you're used to using jQuery). `querySelector` is the way to go.

src/components/select/select.js
```diff
@@ -776,6 +777,14 @@ function OptionDirective($mdButtonInkRipple, $mdUtil) {
     var optionCtrl = ctrls[0];
     var selectCtrl = ctrls[1];
 
+    if (selectCtrl.isMultiple) {
+      element.attr('md-checkbox-enabled', '');
+      if (!checkboxElement) {
+        checkboxElement = angular.element('<div class="_md-container"><div class="_md-icon"></div></div>');
```
jelbourn: This isn't there yet; you're still performing the html -> dom operation for _every single item, everywhere_. The goal is to only do that operation once _ever_, and then clone the result for each item.

You want to do something like 
```js
var CHECKBOX_SELECTION_INDICATOR = 
    angular.element('<div class="_md-container"><div class="_md-icon"></div></div>');
```
right inside of `OptionDirective` before the `return`, and then clone that.

src/components/select/select.spec.js
```diff
@@ -241,10 +241,25 @@ describe('<md-select>', function() {
                               '</md-select>' +
                             '</md-input-container>')($rootScope).find('md-select');
       var label = select.find('md-select-value');
+      var options = select.find('md-option');
+      var numberOfOptions = options.length;
+      var container;
```
jelbourn: Declare variables where they're used.

src/components/select/select.spec.js
```diff
@@ -241,10 +241,25 @@ describe('<md-select>', function() {
...
+      var numberOfChildren;
+
       $rootScope.$digest();
 
       expect(label.text()).toBe('Two');
       expect(label.hasClass('md-select-placeholder')).toBe(false);
+
+
+      // Ensure every md-option element does not have a checkbox prepended to it.
+      for (var i = 0; i < numberOfOptions; i++) {
```
jelbourn: It's fine to just use `options.length` here since it's not a real node list.

src/components/select/select.js
DerekLouie: I believe the behaviour you are describing is, in fact, what is going on here.

Actually if I clone element inside of the `OptionDirective` right before the `return`  statement a new clone will be made every time the `md-option` directive is invoked (if you put a `debugger` on line 760 and navigate to the select demo page, you will notice you hit that `debugger` breakpoint several times).

I have declared the variable at the very top of the file so there should only be one reference present in the code (kept alive in the parent closure), and it is instantiated IFF we have a multi-select (as close to lazy loading as I could manage).

We never clone the element again (if you put a `debugger` on line 782 and navigate to the select demo page, you will notice you hit that `debugger` breakpoint only once).

I agree though that I should rename the variable because my intention was to create a constant. 

I believe putting that constant at the top of the `OptionDirective` initially un-initialized but keeping the if statement would follow more of the "declare variables where they are used" philosophy, but then detecting if the clone has been created already in other `OptionDirective`'s on the page becomes more difficult..

jelbourn: Ah, I just misread the code. I wouldn't worry too much about the lazy initializing here; the cost of creating the element once at start-up should be pretty small.

DerekLouie:
Hello All,

I have fixed all the aforementioned comments!

Please Review.

Best,

Derek

jelbourn:
LGTM

Bouazizi-Ezzeddine:
Hi,
In which version these changes are integrated ?

torontom:
Can I do it with anguler-material 1.0.0? I only want to make minimum changes to our project, and get the multiple select with check box...

ThomasBurleson:
This feature is available in v1.1.0 or higher. See the [Select Demo](https://material.angularjs.org/latest/demo/select) Pizza Toppings which uses the *multiple* attribute: `<md-select ng-model="selectedToppings" multiple>` 

> Version 1.0 is almost a year old and many bugs have been fixed since the v1.0.0 release.

Splaktar:
This works for our standard `multiple` API, but it doesn't work for the undocumented, untested [`ng-multiple` API](https://github.com/angular/material/commit/9956f62d8652a4d335bfe2d759b12634ef8691c8) which was introduced in `0.9.0`.

I've opened https://github.com/angular/material/issues/11759 to track supporting checkboxes when using `ng-multiple="true"`

PR #7460, opened 2016-03-08
fix(progress-circular): removed dependency on `ng-hide` class

### LINKED ISSUE #7454: ng-show is not working on md-progress-circular
See the first progress-circular on the page:
http://codepen.io/jelbourn/pen/RareKa

cc @calebegg

EladBezalel:
- If the progress mode wasn't indeterminate or determinate it would use `ng-hide` class to hide it, and if it's a valid mode, `ng-hide` was removed, what caused `ng-hide` not to preform as wanted.
Now when mode is not indeterminate or determinate set it default to indeterminate

fixes #7454

### DIFF
--- a/src/components/progressCircular/js/progressCircularDirective.js
+++ b/src/components/progressCircular/js/progressCircularDirective.js
@@ -91,7 +91,7 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdUti
     }
   };
 
-  function MdProgressCircularLink(scope, element) {
+  function MdProgressCircularLink(scope, element, attrs) {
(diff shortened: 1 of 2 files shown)


DevVersion:
@EladBezalel You need to update the demo too, because it uses `md-mode=""` to hide the progress circulars. See [here](https://github.com/angular/material/blob/fix/progress-circular-ng-hide/src/components/progressCircular/demoBasicUsage/index.html)

EladBezalel:
@DevVersion :+1:

ThomasBurleson:
@EladBezalel - much better logic here. Thx.

Issue #7116, opened 2016-02-10
mdPanel: Implement MdPanelRef.show and MdPanelRef.hide


ErinCoughlan:
Unlike open/close, show and hide do not add or remove the panel from the DOM. Instead, they manipulate visibility. Both of these methods should return a promise that resolves when the panel is hidden/shown.

This task involves:
* Updating public documentation.
* Adding two methods to the MdPanelRef (adding TODOs for animations and canceling in-progress open/close/show/hide).
* Adding an "isHidden" boolean to indicate whether the panel is visible to users.

ErinCoughlan:
Don't add isHidden (remove it from the documentation). Having isHidden forces anyone having to check whether the panel is open to check two variables. The show/hide promises should be enough to indicate when the states change. And the isOpen says whether the panel is attached.

PR #7461, opened 2016-03-08
update(panel): Add show and hide to mdPanelRef. Cleaned up open and close.

KarenParker:
@gmoothart @ErinCoughlan @jelbourn @DerekLouie 

Please review.

Fixes issue #7116

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -196,12 +196,10 @@ angular
  *     when a panel was interacted with.
  *   - `config` - `{Object=}`: The entire config object that was used in
  *     create.
- *   - `isOpen` - `{boolean}`: Whether the panel is attached to the DOM.
- *     Visibility to the user does not factor into isOpen.
+ *   - `isAttached` - `{boolean}`: Whether the panel is attached to the DOM.
+ *     Visibility to the user does not factor into isAttached.
  *
  * TODO(ErinCoughlan): Add the following properties.
- *   - `isHidden` - `{boolean}`: Whether the panel is attached to the DOM, but
- *     not visible to the user.
  *   - `onDomAdded` - `{function=}`: Callback function used to announce when
  *     the panel is added to the DOM.
  *   - `onOpenComplete` - `{function=}`: Callback function used to announce
@@ -215,19 +213,66 @@ angular
  * @ngdoc method
  * @name MdPanelRef#open
  * @description
- * If the panel is not visible, opens an already created and configured panel.
+ * Attaches and shows the panel.
  *
- * @returns {angular.$q.Promise} A promise that is resolved when the panel is
- * closed.
+ * @returns {!angular.$q.Promise} A promise that is resolved when the panel is
+ * opened.
  */
 
 /**
  * @ngdoc method
  * @name MdPanelRef#close
  * @description
- * If the panel is visible, closes the panel, resolving the promise that is
- * returned from `MdPanelRef#open`. This method destroys the reference to the
- * panel. In order to open the panel again, a new one must be created.
+ * Hides and detaches the panel. This method destroys the reference to the panel.
+ * In order to open the panel again, a new one must be created.
+ *
+ * @returns {!angular.$q.Promise} A promise that is resolved when the panel is
+ * closed.
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#attachOnly
+ * @description
+ * Create the panel elements and attach them to the DOM. The panel will be
+ * hidden by default. Call toggleVisibility() or setVisible(true) to make the
+ * panel visible.
+ *
+ * @returns {!angular.$q.Promise} A promise that is resolved when the panel is
+ * attached.
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#detach
+ * @description
+ * Removes the panel from the DOM. If the panel is visible, hide the panel,
+ * then detach.
+ *
+ * @returns {!angular.$q.Promise} A promise that is resolved when the panel is
+ * detached.
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#setVisibililty
+ * @description
+ * Sets the visibility of the panel. On creation, the panel is hidden. Visible=false
+ * is equivalent to adding "display: none".
+ *
+ * @param {boolean} visible Whether the panel should be visible or hidden.
+ * @returns {!angular.$q.Promise} A promise that is resolved when the panel's
+ * visibility is updated and animations are completed.
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#toggleVisibility
+ * @description
+ * Toggles the visibility of the panel. On creation, the panel is hidden.
+ *
+ * @returns {!angular.$q.Promise} A promise that is resolved when the panel's
+ * visibility is updated and animations are completed.
  */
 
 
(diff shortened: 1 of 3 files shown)


src/components/panel/panel.js
```diff
@@ -464,19 +484,34 @@ function MdPanelRef(config, $injector) {
  * is opened and animations finish.
  */
 MdPanelRef.prototype.open = function() {
-  if (!this.isOpen) {
-    this.isOpen = true;
+  if (this.isOpen) {
+    // Ignore open. We don't want chaining to happen here.
+    return this._$q.reject('Panel is already opened.');
```
jelbourn: Why reject? Why not hang on to the (already resolved) promise from when it was opened and just return that?

KarenParker: This was a design decision, but are discussing it more, we combined this if case with the next one and removed the finally.

jelbourn: What was the reasoning behind it? Why is calling open on an opened panel considered a failure rather than a no-op?

src/components/panel/panel.js
```diff
@@ -464,19 +484,34 @@ function MdPanelRef(config, $injector) {
...
+
+  // TODO(ErinCoughlan) - Cancel any in-progress `closing`.
 
-    var self = this;
-    this._openPromise = this._$q(function(resolve, reject) {
-      self._openReject = reject;
+  this.isOpen = true;
+  var self = this;
+  this._openPromise = this._$q(function(resolve, reject) {
+    try {
```
jelbourn: What's the reasoning behind the try-catch here? This will eat any errors that are thrown while opening so that they're not logged to the console.

src/components/panel/panel.js
```diff
@@ -488,33 +523,115 @@ MdPanelRef.prototype.open = function() {
  * closed and animations finish.
  */
 MdPanelRef.prototype.close = function() {
-  if (this.isOpen) {
-    this.isOpen = false;
+  if (!this.isOpen) {
+    // Ignore close. We don't want chaining to happen here.
+    return this._$q.reject('Panel is already closed.');
+  }
```
jelbourn: Same comment about no-op vs. fail.

src/components/panel/panel.js
```diff
@@ -488,33 +523,115 @@ MdPanelRef.prototype.open = function() {
...
+  // TODO(ErinCoughlan) - Cancel any in-progress `opening`.
 
-      this._closePromise.finally(function() {
-        // Clear for next open/close pairing.
-        self._closePromise = undefined;
-      });
+  this.isOpen = false;
+  var self = this;
+  this._closePromise = this._$q(function(resolve, reject) {
+    try {
```
jelbourn: Same comment about try-catch

src/components/panel/panel.js
```diff
@@ -488,33 +523,115 @@ MdPanelRef.prototype.open = function() {
...
 };
 
 
 /**
+ * Show the panel.
+ *
+ * @returns {!angular.$q.Promise} A promise that is resolved when the panel is
+ * shown and animations finish.
+ */
+MdPanelRef.prototype.show = function() {
```
jelbourn: From an API design standpoint, that the difference between open/close and show/hide is rather unclear. I'm thinking perhaps if we make it `setVisible` instead that might be a little more obvious? Thoughts?

KarenParker:
Updated based on comments and refactored based on a discuss offline.

src/components/panel/panel.js
```diff
@@ -215,19 +215,69 @@ angular
...
+ * Removes the panel from the DOM. If the panel is visible, hide the panel,
+ * then detach.
+ *
+ * @returns {!angular.$q.Promise} A promise that is resolved when the panel is
+ * detached.
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#show
```
jelbourn: I talked a lot about this with @kara; we thought that having both `open` / `close` and `show` / `hide` is going to lead users to doing the wrong thing or just being confused. So, as an alternative, how about we simply do `open` / `close` as they are now, and then have `addClass` / `removeClass` / `toggleClass` to accomplish the setting and unsetting of `display: none` ?

KarenParker:
Comment addressed. I added set/toggle visibility to add/remove/toggle the class.

src/components/panel/panel.js
```diff
@@ -215,19 +213,66 @@ angular
...
+ * Removes the panel from the DOM. If the panel is visible, hide the panel,
+ * then detach.
+ *
+ * @returns {!angular.$q.Promise} A promise that is resolved when the panel is
+ * detached.
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#setVisibililty
```
jelbourn: I can live with this, but I think having `addClass` / `removeClass` / `toggleClass` (with a `display: none` class) methods are more generally useful and lead to a smaller API surface area.

src/components/panel/panel.spec.js
```diff
@@ -1,9 +1,10 @@
-describe('$mdPanel', function() {
+fdescribe('$mdPanel', function() {
```
jelbourn: `fdescribe`

jelbourn:
LGTM

src/components/panel/panel.js
ErinCoughlan: I think I agree with @jelbourn now that I understand the API suggestion. What pushed me towards his version was the fact that we really don't know _how_ the user wants to hide the panels. With our version, we say they have to use `display: none`, but they may prefer `visibility: hidden` or `opacity: 0`. Since the panel is so low-level, we want to make it as flexible as possible.

jelbourn: To be clear, I mean adding `addClass` / `removeClass` / `toggleClass` to `MdPanelRef`

KarenParker:
I updated the visibility functions to be addClass/removeClass/toggleClass. I still have a concern about it becoming visible as soon as attach is called (attach now is essentially an attach and show). If we don't care about this case, that is also fine.

jelbourn:
LGTM

src/components/panel/panel.js
```diff
@@ -571,54 +630,181 @@ MdPanelRef.prototype.open = function() {
...
+ * @returns {!angular.$q.Promise} A promise that is resolved when the panel has
+ * the new class and animations finish.
+ */
+MdPanelRef.prototype.addClass = function(newClass) {
+  if (!this._panelContainer) {
+    return this._$q.reject(
+        'Panel does not exist yet. Call open() or attach().');
+  }
+
+  if (this._panelContainer.hasClass(newClass)) {
```
gmoothart: Can you add a check that the promise already exists, to be safe?

src/components/panel/panel.spec.js
```diff
@@ -96,6 +68,139 @@ describe('$mdPanel', function() {
...
+      expect(panelRef.isAttached).toEqual(false);
+    });
+
+    it('should resolve when opening/closing', function () {
+      var openResolved = false;
+      var closeResolved = false;
+
+      expect(panelRef.id).toBeDefined();
+      expect(panelRef.open).toBeOfType('function');
+      expect(panelRef.close).toBeOfType('function');
```
gmoothart: what are you testing here?

gmoothart:
:+1:

src/components/panel/panel.spec.js
KarenParker: This code was here before, I just moved it into this describe.

yanndupuy: This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  dupuyyann@gmail.com
    Domain zyapps.com has exceeded the max emails per hour (417/400 (104%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
Received: from github-smtp2-ext3.iad.github.net ([192.30.252.194]:53732 helo=github-smtp2b-ext-cp1-prd.iad.github.net)
	by serveur.ex2-p10.com with esmtps (TLSv1.2:ECDHE-RSA-AES256-GCM-SHA384:256)
	(Exim 4.86_1)
	(envelope-from <noreply@github.com>)
	id 1afmMd-000ajn-V1
	for yann.dupuy@zyapps.com; Tue, 15 Mar 2016 11:38:22 +0100
Date: Mon, 14 Mar 2016 13:30:39 -0700
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=github.com;
	s=pf2014; t=1457987439;
	bh=utXgOOJgFx8VFH9hX/UXFueY/iKH8nbYhrftdy1YiXg=;
	h=From:Reply-To:To:In-Reply-To:References:Subject:List-ID:
	 List-Archive:List-Post:List-Unsubscribe:From;
	b=nQzTYZRk1OWCbcSNN+zPF6kz5owMx2LFcM8ptZsMn35MRctpOPPPhQjx/Bd/DIozK
	 eO9qdzKZNkBgq2wHK/n+t3/SdgaQFj7RUsPV5X86OZa+tPRt75prkMv5AulOB5WxzO
	 f2XOXlP/RHfRH5oC1AK+BhO7p9KHGCR8RAleMRYA=

ThomasBurleson:
@KarenParker - `angular/material:feat/panel` has also been updated to include the latest changes in master. I recommend that everyone do the following:

```console
git co feat/panel
git pull --rebase origin feat/panel
```

Issue #7467, opened 2016-03-08
md-icon: svg <clipPath> no longer works


mpiasta-ca:
After the commit 810c4f3, any svg with `<clipPath id="...">` fails to render the clipPath, because line 433 @ "src/components/icon/js/iconService.js" removes all svg child id, but this id is required for clipPath to work.

Created after issue report: https://github.com/angular/material/issues/7381

I tested with the change from 810c4f3, and here is screenshot to show the breaking change. In this example, the five square svg icons are actually supposed to appear as circles (the only reason the "Canada" icon works is because it doesn't use a clipPath).

![clip-path-id-issue](https://cloud.githubusercontent.com/assets/672103/13610754/28df32da-e514-11e5-9b2b-d207abd5aa59.png)

Potential solution: instead of removing ID from the svg's child elements, need to prefix a random string to all child element id's at render (not at caching). If done at caching, issue 7381 will still persist (svg breaks when the icon is rendered more than once, because IDs are no longer unique [Firefox, IE]). So best to attach this random prefix string in `<md-icon>` right before the svg is rendered (one string that is attached to all IDs of that svg, so that they can still reference each other through`xml:link="..."` and `clip-path: url(#...)`).

DevVersion:
@mpiasta-ca I like the solution. This seems to be a elegant way.

DevVersion:
The cached icons id's are now always prefixed with a incrementing prefix.
-  67bd35d

PR #7468, opened 2016-03-08
fix(icon): rename nodes id's from cache to prevent duplicates.

DevVersion:
Currently we always rename the icons id's on cache initialization, but that causes problems, when using the cached icon multiple times in the app.
A super elegant solution, is to transform the ids, when the icon is requested from the cache.

Referencing #7467

@david-gang Ping.


EladBezalel:
@topherfangio please take a look as well

david-gang:
Hi @DevVersion ,

The fix looks cool :+1: 
I think that the test case should have some positive expectation rather than not.toBe but maybe it is an issue of taste :-)

topherfangio:
@EladBezalel LGTM :+1:

PR #7471, opened 2016-03-08
fix(progress-circular): animation timing

crisbeto:
There were cases where the circle rendering animation would run for one iteration longer than it was supposed to.
This caused a bug where the determinate value circle would get drawn at the wrong place in the end of a longer animation.
Also replaces the `getTimestamp` method with `$mdUtil.now`.

### DIFF
--- a/src/components/progressCircular/js/progressCircularDirective.js
+++ b/src/components/progressCircular/js/progressCircularDirective.js
@@ -144,7 +144,7 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdUti
 
     function renderCircle(animateFrom, animateTo, easing, duration, rotation) {
       var id = ++lastAnimationId;
-      var startTime = getTimestamp();
+      var startTime = $mdUtil.now();
       var changeInValue = animateTo - animateFrom;
       var diameter = getSize(scope.mdDiameter);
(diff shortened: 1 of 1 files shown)


src/components/progressCircular/js/progressCircularDirective.js
```diff
@@ -156,7 +156,7 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdUti
         path.attr('d', getSvgArc(animateTo, diameter, pathDiameter, rotation));
       } else {
         $$rAF(function animation(now) {
-          var currentTime = (now || getTimestamp()) - startTime;
+          var currentTime = $window.Math.min((now || getTimestamp()) - startTime, animationDuration);
```
DevVersion: Just a question, why are you using `$window.Math`? `Math` alone works too.

crisbeto: I find it easier to locate global variables that are being used if I reference then through `$window`.

DevVersion: Just asked, because to be consistent, we always use it without `$window`. [See here](https://github.com/angular/material/blob/master/src/components/input/input.js#L433)

But it's your choice :wink:

PR #7482, opened 2016-03-09
feat(sidenav): add attribute to disable escape to close

### LINKED ISSUE #7022: md-sidenav should have a clickOutsideToClose option.
There's no easy way to disable the click or Esc to close behavior of the Sidenav. It would be great to have something similar to Dialog's clickOutsideToClose option to disable/enable this behavior.

DevVersion:
- Fixes tests for `md-disable-backdrop`
- Adds attribute to disable the escapeToClose action.

I decided to use `md-escape-to-close` instead of `md-disable-escape-to-close`, because it seems more user-friendly.

If you have any suggestions about the attribute name, let me know ;)

Fixes #7022

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -196,6 +196,8 @@ function SidenavFocusDirective() {
  *
  * @param {expression=} md-is-open A model bound to whether the sidenav is opened.
  * @param {boolean=} md-disable-backdrop When present in the markup, the sidenav will not show a backdrop.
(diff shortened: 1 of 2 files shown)


DevVersion:
Note: We're currently discussing about the implementation. I'll keep that PR open meanwhile.

DevVersion:
I will close that PR, because I don't think that the escapeToClose attribute makes any sense, we need to discuss first, how we want to implement it correctly. See #7022

Issue #7178, opened 2016-02-16
mdPanel: Create an mdPanelPosition object and pass it with the config.


ErinCoughlan:
This task involves:

* Creating the mdPanelPosition object.
* Adding public documentation for mdPanelPosition.
* Implementing the following API methods, including configuring the CSS in JS to make it work:
  * absolute()

ErinCoughlan:
Re-designing this added a 2.

PR #7494, opened 2016-03-09
update(panel): Adds absolute positioning.

ErinCoughlan:
Fixes issues #7178 and #7191.

@jelbourn @KarenParker @gmoothart @DerekLouie Please review.


src/components/panel/panel.js
```diff
@@ -544,4 +651,188 @@ MdPanelRef.prototype._createPanel = function() {
  */
 MdPanelRef.prototype._addStyles = function() {
   this._panelContainer.css('z-index', this._config['zIndex']);
+
+  var positionConfig = this._config['position'];
+
+  if (!positionConfig) { return; }
+
+  var isAbsolute = positionConfig.getAbsolute();
+  var panelPosition = isAbsolute ? 'fixed' : 'relative';
```
jelbourn: I vaguely recall `position: fixed` for this use case having some issues on one of the mobile platforms, but I don't remember any of the details. Might be something to just watch out for.

src/components/panel/panel.js
```diff
@@ -544,4 +651,188 @@ MdPanelRef.prototype._createPanel = function() {
...
+ *   position: panelPosition
+ * });
+ *
+ * @final @constructor
+ */
+function MdPanelPosition() {
+  /** @private {boolean} */
+  this._absolute = false;
+
+  /** @private {string|undefined} */
```
jelbourn: Use empty string instead of undefined?

src/components/panel/panel.js
```diff
@@ -544,4 +651,188 @@ MdPanelRef.prototype._createPanel = function() {
...
+};
+
+
+/**
+ * Centers the panel in the x axis. Note: This is the default option,
+ * so we don't set any CSS values.
+ * @returns {MdPanelPosition}
+ */
+MdPanelPosition.prototype.centerHorizontally = function() {
+  return this;
```
jelbourn: Maybe throw an error for not-yet-implemented methods?

ErinCoughlan: This gets a little confusing here. The default position of the panel is centered, both horizontally and vertically. So this method doesn't need to set any variables.

I could leave the center methods out from the API, but it felt incomplete doing so and I liked being able to explicitly write .absolute().center() when specifying the position.

What do you think? Remove the methods? Remove all default behavior? Leave as is? Something else?

jelbourn: Ah, I was thinking of times where you'd want to _change_ a position. Is that something you plan on supporting?

ErinCoughlan: That wasn't something I was thinking of, but going over it more, I can see that being useful in cases where you want to reuse a single panel instance for multiple popups. (For example, having a single tooltip where you can change the position and content without creating it again.)

I don't know how common that use case would be though.

jelbourn: I'd omit the method then, since they don't do anything now.

ErinCoughlan: Removed the centering methods.

src/components/panel/panel.js
ErinCoughlan: I'll keep an eye out for cases where this won't work. It looks okay here: http://caniuse.com/#feat=css-fixed

jelbourn:
LGTM

KarenParker:
LGTM

PR #7504, opened 2016-03-10
fix(autocomplete): fix incorrect height for progress-bar

### LINKED ISSUE #7497: Autocomplete progress bar height is wrong
Progress bar height is too large on autocomplete

Just try to write something in the autocomplete
http://codepen.io/anon/pen/mPPZYK

DevVersion:
The current indeterminate class wasn't privatized and that caused the progress bar to have a height of 5px, instead of 3px

@topherfangio Seems like you forgot to privatize the indeterminate class here. Can you review this PR?

Fixes #7497

### DIFF
--- a/src/components/autocomplete/autocomplete.scss
+++ b/src/components/autocomplete/autocomplete.scss
@@ -82,7 +82,7 @@ md-autocomplete {
         left: 2px;
         width: auto;
       }
-      .md-mode-indeterminate {
+      ._md-mode-indeterminate {
         position: absolute;
         top: 0;
         left: 0;


gpopovic:
DevVersion, i suspect same error happened with select.

Loading on select is almost two times larger than before.
http://codepen.io/anon/pen/oxLbgo

DevVersion:
@gpopovic Just checked, it's not related to any privatizing. Can you please file another issue for the `select`, because this is related to 640b55db400cb5c45381edce8b75058a75399c14

crisbeto:
I've replied to your issue @gpopovic, this PR isn't related to it.

topherfangio:
LGTM

PR #7506, opened 2016-03-10
fix(select): spinner size

### LINKED ISSUE #7505: Loading on select is almost two times larger than before
v1.0.6.
http://codepen.io/anon/pen/Wwxrbq

master
http://codepen.io/anon/pen/oxLbgo

crisbeto:
Due to the new changes in the `progressCircular` component,
the spinner was being rendered bigger than usual. This change
reverts to the old size.

Fixes #7505

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -87,9 +87,9 @@ angular.module('material.components.select', [
  * </hljs>
  *
  * At first one might expect that the select should be populated with "Bob" as the selected user. However,
- * this is not true. To determine whether something is selected, 
+ * this is not true. To determine whether something is selected,
  * `ngModelController` is looking at whether `$scope.selectedUser == (any user in $scope.users);`;
- * 
+ *
  * Javascript's `==` operator does not check for deep equality (ie. that all properties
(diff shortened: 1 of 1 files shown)


src/components/select/select.js
```diff
@@ -144,7 +144,7 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
         .find('md-content')
         .prepend(angular.element(
           '<div>' +
-          ' <md-progress-circular md-mode="{{progressMode}}" ng-hide="$$loadingAsyncDone"></md-progress-circular>' +
+          ' <md-progress-circular md-mode="{{progressMode}}" ng-hide="$$loadingAsyncDone" md-diameter="25px"></md-progress-circular>' +
```
DevVersion: Shouldn't that be `md-diameter="25"`? 
I know it will still work with the `px` suffix because the float will be parsed and ignores the text.

crisbeto: I wanted it to be as obvious as possible.

DevVersion: :+1:

PR #7515, opened 2016-03-10
fix(mixins): unquote is causing issues with module compilation.

DevVersion:
`unquote` caused issues with theming and module compilation.

Fixes #7514

PR #7524, opened 2016-03-11
fix(build): fix theme strings becoming unquoted

jelbourn:


ThomasBurleson:
@jelbourn  - Do we still need https://github.com/angular/material/pull/7515/files

PR #7531, opened 2016-03-11
fix(list): item-inner shouldn't be wider than the container

### LINKED ISSUE #7525: Md-list-item with ng-click still broken in master
It appears that text overflow is still not working on md-list-item with ng-click
http://codepen.io/anon/pen/KzMyGY

Solution this time is
`._md-list-item-inner {
    overflow: hidden;
}
`

### LINKED ISSUE #7551: Text disappears in md-list-item if background-color is set
Hi,
if you want to set a background-color in a md-list-item (with ng-click), the text inside the list-item disappears.

http://codepen.io/anon/pen/dMpWpQ

Best Regards
Frank

DevVersion:
This fixes a overflow issue - when being in a flex layout.

image

image

* Also fixes the issue, when using  a custom background color on the list-item.

Fixes #7551 Fixes #7525

### DIFF
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -138,11 +138,13 @@ md-list-item {
       align-items: center;
       justify-content: flex-start;
 
-      padding: 0;
+      padding: 0 16px;
       margin: 0;
 
+      // The button should not inherit the parents background color.
+      background-color: initial;
+
       font-weight: 400;
-      background-color: inherit;
       @include rtl(text-align, left, right);
       border: medium none;
 
@@ -151,16 +153,16 @@ md-list-item {
         position: absolute;
         top: 0;
         left: 0;
+        height: 100%;
 
         margin: 0;
         padding: 0;
-        height: 100%;
       }
 
       ._md-list-item-inner {
         // The list item content should fill the complete width.
-        flex: 1 1 auto;
-        padding: 0 16px;
+        width: 100%;
+        height: 100%;
       }
 
     }


EladBezalel:
LGTM needs to manual tested

DevVersion:
@EladBezalel Thanks :+1: This should be tested manually as soon as possible, because the overflow seems to be broken in some specific layouts.

src/components/list/list.scss
```diff
@@ -138,11 +138,13 @@ md-list-item {
       align-items: center;
       justify-content: flex-start;
 
-      padding: 0;
+      padding: 0 16px;
       margin: 0;
 
+      // The button should not inherit the parents background color.
+      background-color: initial;
```
ThomasBurleson: Why do you have this here. Is the rule specified in the Design specifictions?

DevVersion: No, that's just our button wrap, it's just executing the ripple and the click attributes. It should not inherit the background, otherwise the text content will be invisible.

http://codepen.io/anon/pen/dMpWpQ

ThomasBurleson: I do not understand the **need** for this setting. I see no difference with or without this setting.

DevVersion: The button wrap, has the `_md-no-style` class applied, which removes all the needed stylings.

image

And `_md-no-style` applies the background color inheritance to the button wrap, which is a bug, because the button wrap, should be transparent, as it is by default.

So now if remove the inheritance, the background-color won't be red anymore.
image

This is a bug, because this is our `executor and ripple` wrap, which overlays over the actual list. So that means, if it has the inherited background-color, the complete list will be invisible.

EladBezalel: `initial` is not supported in IE

DevVersion: Already talked about that with Thomas, IE11 doesn't need this property, because the bug isn't present there. So this will be just an unnecessary property on IE11

yanndupuy:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  dupuyyann@gmail.com
    Domain zyapps.com has exceeded the max emails per hour (420/400 (105%)) allowed.  Message will be reattempted later

------- This is a copy of the message, including all the headers. ------
Received: from github-smtp2-ext6.iad.github.net ([192.30.252.197]:54259 helo=github-smtp2b-ext-cp1-prd.iad.github.net)
	by serveur.ex2-p10.com with esmtps (TLSv1.2:ECDHE-RSA-AES256-GCM-SHA384:256)
	(Exim 4.86_1)
	(envelope-from <noreply@github.com>)
	id 1afmMi-000akp-Um
	for yann.dupuy@zyapps.com; Tue, 15 Mar 2016 11:38:25 +0100
Date: Mon, 14 Mar 2016 10:00:52 -0700
DKIM-Signature: v=1; a=rsa-sha256; c=relaxed/relaxed; d=github.com;
	s=pf2014; t=1457974852;
	bh=mPDmjQpIrRPd1zv//ouhe9O1agwcxwj6fr1MzO1JpoE=;
	h=From:Reply-To:To:In-Reply-To:References:Subject:List-ID:
	 List-Archive:List-Post:List-Unsubscribe:From;
	b=UISPVXehDnoxyp0op/vr6mEiPClFToOyWEQKIGpkhFTGp/eT2986GvMg8nbRnMp0C
	 zCT4Gj+CQnhAC919VB/Bs8NYaMPai07FfO8huWBOQJ2kAoGo/tdMPrhskFzGbaiMFe
	 H8Jukd9Xqq9/oM/MVvv1zt/0UVUnEJibatNIA4I4=

PR #7539, opened 2016-03-11
Md fab fixes

garmoshka-mo:
.md-fab - fix bug (duplicate scss declaration) and 
change usage of color palette to more pastel


EladBezalel:
Is this aligned with the spec? If so please refer

garmoshka-mo:
@EladBezalel those changes made not by specification. 
Reason, why they are made is:
- duplicated style definition created confusion (second definition overrides first one and they declare same logic)
- color change - is question of perception though

I see acceptance process is very strict, actually it is not so significant changes, can be ignored

EladBezalel:
You can update your PR to include the duplication fix but our colors are aligned with the spec and aren't a subject to our considerations

PR #7541, opened 2016-03-11
fix(checkbox): pointer events disable ripple events too

### LINKED ISSUE #7538: md-checkbox missing ripples on click
md-checkbox seems to be missing ripples on click.

See:
https://material.angularjs.org/HEAD/demo/checkbox
vs
https://material.angularjs.org/HEAD/demo/radioButton

https://www.google.com/design/spec/components/selection-controls.html#selection-controls-radio-button

DevVersion:
We should not disable the ripple events by using `pointer-events: none`.
IE11 always focuses the div, instead of the `md-checkbox`, and that's causing issues with ngModel etc.

I good trick is, to redirect the focus, instead of preventing the whole focus / pointer events.
This won't interfere with the ripple, but it will fix the IE11 issue.

https://connect.microsoft.com/IE/feedback/details/1028411/

cc @ThomasBurleson 

Fixes #7538

### DIFF
--- a/src/components/checkbox/checkbox.js
+++ b/src/components/checkbox/checkbox.js
@@ -68,6 +68,7 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
   // **********************************************************
 
   function compile (tElement, tAttrs) {
+    var container = tElement.children();
 
     tAttrs.type = 'checkbox';
     tAttrs.tabindex = tAttrs.tabindex || '0';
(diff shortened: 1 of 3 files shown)


EladBezalel:
LGTM

ThomasBurleson:
I am confused can we not use tabindex= 0 or -1 to solve the issue without programmatically redirecting focus?

DevVersion:
@ThomasBurleson That won't work, because that's the actual IE11 bug. 
Normally when no `tabIndex` is set, then the element should be not focusable (in no cases). But IE11 decided to make `DIV` and `SPAN` always focusable. (no caring about tabIndex etc.)

- https://connect.microsoft.com/IE/feedback/details/1028411/

ThomasBurleson:
@DevVersion - That is what I needed to know. Thx.

PR #7544, opened 2016-03-12
fix(datepicker): enable scrolling when scope destroyed.

### LINKED ISSUE #7542: md-datepicker-- md-scroll-mask remains open on navigation back
If you open the md-datepicker that is on a secondary route in the app and then hit the back button while the datepicker is still open, the scroll mask remains open and preventing scroll until the user clicks.

DevVersion:
Currently when we change the route, while the pane is open, the scroll mask won't be removed.
That's why we should re-enable scrolling on scope destruction.

In this case, we also have to check for the pane to be open, because otherwise we may accidentally remove another scroll mask.

@jelbourn Can you also review please?

Fixes #7542

### DIFF
--- a/src/components/datepicker/datePicker.js
+++ b/src/components/datepicker/datePicker.js
@@ -527,6 +527,10 @@
     this.calendarPane.classList.remove('md-pane-open');
     this.calendarPane.classList.remove('md-datepicker-pos-adjusted');
 
+    if (this.isCalendarOpen) {
+      this.$mdUtil.enableScrolling();
+    }
+
     if (this.calendarPane.parentNode) {
       // Use native DOM removal because we do not want any of the angular state of this element
       // to be disposed.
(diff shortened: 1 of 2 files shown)


EladBezalel:
LGTM

src/components/datepicker/datePicker.js
```diff
@@ -570,11 +574,10 @@
   /** Close the floating calendar pane. */
   DatePickerCtrl.prototype.closeCalendarPane = function() {
     if (this.isCalendarOpen) {
-      this.isCalendarOpen = false;
       this.detachCalendarPane();
+      this.isCalendarOpen = false;
```
jelbourn: I'm not a fan of this because now the person working on this needs to know that `isCalendarOpen` _must_ be set after detaching; this is a weird coupling of behaviors that makes brittle code. It would be much better to simple call `enableScrolling()` in the `$destroy` handler in addition to `detachPane`

DevVersion: I can understand your thoughts. I had the behavior, you desired, before, but I thought calling `enableScrolling` is to generic, because `detachPane` will be always called after `hidePane`, so it made sense to me, to move the enable scroll function completely into the `detachPane` function.

PR #7550, opened 2016-03-13
fix(chips): fix md-max-chips for autocomplete item watch

### LINKED ISSUE #7549: Md-max-chips doesn't work with autocomplete
Md-max-chips works with only input, but doesn't work when autocomplete is used with chips.

On codepen, ,d-max-chips is set to 1 on both autocomplete and input.
Chips with input will not allow adding another chip, while autocomplete will just ignore that tag.
http://codepen.io/anon/pen/GZjrQE

DevVersion:
The chips didn't check for the maxmium, when using an autocomplete.
That's why the chip always got appended, when reached the chips max.

Fixes #7549

### DIFF
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -556,6 +556,56 @@ describe('<md-chips>', function() {
           expect(ctrl.chipBuffer).toBe('Test 2');
           expect(scope.items.length).not.toBe(2);
         });
+
+        it('should not append the chip when maximum is reached and using an autocomplete', function() {
+          var template =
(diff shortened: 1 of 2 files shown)


mariohmol:
i dont know about you.. but here when I use md-max-chips="1" it limits to 1 but autocomplete still enabled so you can type a text and add on it.

To avoid that i had to make a ng-show on md-autocomplete to hide when nd-model element is filled

DevVersion:
Hey @mariohmol. Please file a new issue and I'll investigate.

mariohmol:
Here is , https://github.com/angular/material/issues/8332 . thanks!

PR #7553, opened 2016-03-13
fix(textarea): scrolling, text selection, reduced DOM manipulation.

### LINKED ISSUE #7487: md-input textarea is broken as earlier
md-input textarea is broken as earlier in #2007 
Bug is even in 1.06
1. scroll-bar is not appearing
2. when in md-no-autogrow just one line of text appears (even if rows= 5, and one has 3 lines of text)
3. md-select-on-focus: just flashes the select and then select disappears.

crisbeto:
* Fixes some jumping when the textarea is being resized.
* Fixes `mdSelectOnFocus` not working in Edge and being unreliable in Firefox.
* Fixes `textarea` not being scrollable once it is past it's minimum number of rows.
* Fixes `textarea` not being scrollable if `mdNoAutogrow` is specified.
* Tries to reduce the number of event listeners and the amount of DOM manipulation when resizing the `textarea`.

Closes #7487.

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -250,7 +250,7 @@ function labelDirective() {
  *
  */
 
-function inputTextareaDirective($mdUtil, $window, $mdAria) {
+function inputTextareaDirective($mdUtil, $window, $mdAria, $timeout) {
   return {
     restrict: 'E',
     require: ['^?mdInputContainer', '?ngModel'],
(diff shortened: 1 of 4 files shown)


src/components/input/input.js
```diff
@@ -629,12 +631,28 @@ function mdSelectOnFocusDirective() {
...
+      // The debounce is here for Edge's sake, otherwise the selection doesn't work.
+      $timeout(function() {
+        element[0].select();
+      }, 1, false);
+
+      // Prevents the default action of the first `mouseup` after a focus.
+      // This is necessary, because browsers fire a `mouseup` right after the element
+      // has been focused. In some browsers (Firefox in particular) this can clear the
+      // selection. There are examples of the problem in issue #7487.
+      element.on('mouseup', preventMouseUp);
```
DevVersion: Wouldn't it make more sense to keep the `mouseup` event registered and just change a state in the directive?.

crisbeto: That would mean adding another handler for `blur`, though.

src/components/input/input.js
```diff
@@ -626,15 +628,37 @@ function mdSelectOnFocusDirective() {
...
+        .off('focus', onFocus)
+        .off('mouseup', onMouseUp);
     });
 
     function onFocus() {
-      // Use HTMLInputElement#select to fix firefox select issues
-      element[0].select();
+      // Use HTMLInputElement#select to fix firefox select issues.
+      // The debounce is here for Edge's sake, otherwise the selection doesn't work.
+      preventMouseUp = true;
```
DevVersion: What happens if I focus the element using keyboard interaction? The `mouseup` event can't be triggered.

crisbeto: The `mouseup` happens after the focus. E.g. you click, `focus` fires, `mouseup` fires. I didn't have issues with this while testing.

crisbeto: Good point, it actually prevents the first click afterwards. Not sure how we could distinguish between the two though.

DevVersion: When is the `mouseup` event usually called? Before the 1ms delay?

crisbeto: I don't have it running atm either, but from what I remember it's afterwards. I'll take a closer look at it later.

crisbeto: It's good now with resetting it from inside the `$timeout`. Also had time to test it out in most browsers.

DevVersion: That's exactly what I've thought about. Great :+1:

src/components/input/input.js
```diff
@@ -365,84 +365,81 @@ function inputTextareaDirective($mdUtil, $window, $mdAria) {
...
-        min_rows = parseInt(node.getAttribute('rows'));
-      }
-
-      var onChangeTextarea = $mdUtil.debounce(growTextarea, 1);
-
-      function pipelineListener(value) {
-        onChangeTextarea();
-        return value;
-      }
+      var minRows = attr.hasOwnProperty('rows') ? parseInt(attr.rows) : NaN;
```
DevVersion: How can I now let the textarea grow?

crisbeto: If you don't specify the rows. It works in the same way as before, I just made it a one-liner.

DevVersion: Ah, just rechecked with the latest version. Everything works fine :+1:

dkirsanov:
Hope that PR solves that issue too:

PR #7562, opened 2016-03-14
chore(travis): travis cache phantomjs2 binary.

DevVersion:
This caches the `phantomjs2` installation. There appeared problems, with too many requests on Travis, while downloading the installation on each change.

I decided to use two variables on top of the script, to be able to update the version easily.


ThomasBurleson:
Very nice!

PR #7564, opened 2016-03-14
perf(tabs): use transform instead of left & right positioning ( md-ink-bar )

corentin-gautier:


ThomasBurleson:
Closing. Change risk is too high to existing customers.

PR #7568, opened 2016-03-15
feat(menu): check for ngClick and ngMouseenter data attributes

bretkikehara:
ensure alternative for ngClick and ngMouseenter definitions are selected.


ThomasBurleson:
More generic solution in progress.

PR #7573, opened 2016-03-15
fix(list): fix compatibility with IE11

### LINKED ISSUE #7557: Padding on item-list with ng-click wrong on ie 11
On ie11, padding is wrong ...
ebb2b408-af11-4f9f-a786-99b293c3e12f

DevVersion:
* Added a mixin / workaround for the flexbox bug for Internet Explorer 11.

The most common problems in the list component are now fixed, so the only missing fix for now is the compatibility with IE11. So that means we should test this thoroughly.

PS: I decided, not to make this mixin generic, because the use-cases for the `min-height` flexbug are really rarely.

Fixes #7557.

@ThomasBurleson Please take a look as well.

### DIFF
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -37,6 +37,20 @@ $list-item-dense-header-font-size: 1.3rem !default;
 $list-item-dense-font-size: 1.2rem !default;
 $list-item-dense-line-height: 1.25rem !default;
 
+// This is a mixin, which fixes IE11's vertical alignment issue, when using `min-height`.
+// This mixin isn't that much generic to be used in other cases.
+// It can be only used for the list-item's button container.
+// See https://connect.microsoft.com/IE/feedback/details/816293/
+@mixin ie11-min-height-flexbug($min-height) {
+  div.md-button:first-child {
+    &::before {
+      content: '';
+      min-height: $min-height;
+      visibility: hidden;
+    }
+  }
+}
+
 md-list {
   display: block;
   padding: $list-padding-top $list-padding-right $list-padding-bottom $list-padding-left;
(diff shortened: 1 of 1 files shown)


src/components/list/list.scss
```diff
@@ -37,6 +37,20 @@ $list-item-dense-header-font-size: 1.3rem !default;
...
+// This is a mixin, which fixes IE11's vertical alignment issue, when using `min-height`.
+// This mixin isn't that much generic to be used in other cases.
+// It can be only used for the list-item's button container.
+// See https://connect.microsoft.com/IE/feedback/details/816293/
+@mixin ie11-min-height-flexbug($min-height) {
+  div.md-button:first-child {
+    &::before {
+      content: '';
+      min-height: $min-height;
+      visibility: hidden;
```
ThomasBurleson: What element is this hiding?

DevVersion: The pseudo element, which sets our height. This is our actual workaround. IE11 now thinks, that the actual button container has a fixed height, but behind the scenes (::before pseudo) it still has a min-height.

ThomasBurleson: Notice that we have a `/core/services/layout/ie_fixes.css`. These are appended to the `/dist/angular-material.css` contents as **raw** css.

Perhaps we should add this ^^ (as CSS instead of scss) to that file?

DevVersion: Oh, I don't think that's a good idea. It should be located in the list component, since the component can be exported as a (Closure)-/ Module.

In addition to that, it will require much modifying, whenever we change a height of the list. 
Also it's not really an IE fix, it's rather a workaround.

Issue #6443, opened 2015-12-22
ngMessages have multiple bugs


gligoran:
Hello,

I've [created a pen](http://codepen.io/gligoran/pen/GoqdPy), with multiple cases where ngMessages don't work as they should.

There seem to be 2 core issues:
1. The auto hide functionality (ngMessages that doesn't use md-auto-hide, ngIf, ngShow, or any other directive that disables auto hide) has trouble switching between different errors.
2. Custom $asyncValidators display multiple of the same message at once.

jaGarcia:
+1  to point 1

jaspersorrio:
+1 to point 1

Here is my <a href="http://codepen.io/jaspersorrio/pen/LGxyNV?editors=101">pen</a>!

I noticed if the user typed too fast, the errors do not appear.
And @gligoran 2nd point, i am facing that too.

Odahi:
I've also noticed that using ngMessagesInclude can change the show/hide behavior in messages.

wcjr:
I've noticed the nodes for the messages are there but they have an `opacity: 0` when it should be `1`.  Does this have something with ngAnimate?

Clarity-89:
+1 to wcjr's issue

wcjr:
I took a closer look at issue [6610](https://github.com/angular/material/issues/6610) referenced by @grosem and it is fully explained there.  Unfortunately it seems @ThomasBurleson labeled it for Q2 consideration :-(  I noticed it was also labeled as an "Edge" case...that seems under-prioritized to me personally if I'm understanding the team's approach to solving this one.

chadjaros:
+1
The required (number one) bug also seems to prevent other error messages (such as pattern) from showing up properly. Very annoying.

wcjr:
looks like these are related issues that are being worked now:  [6859](https://github.com/angular/material/issues/6859) and [6399] (https://github.com/angular/material/issues/6399)

topherfangio:
Hi all, we are indeed working on these issues; just trying to make sure we get some proper tests, and since this is related to animations, we're working with the core Angular team to make sure we have the proper approach. This is definitely a priority on our list!

wcjr:
Thanks for letting us know @topherfangio

ThomasBurleson:
This issue is closed as part of our ‘Surge Focus on Material 2' efforts.
        For details, see our [forum posting](http://bit.ly/1UhZyWs).

ThomasBurleson:
[...]

PR #7583, opened 2016-03-15
fix(input): detect ng-message if container is already invalid.

### LINKED ISSUE #6399: ng-messages does not show message
reproduce:

1 - open http://codepen.io/topherfangio/pen/eJpoZN
2 - enter input>5 and focus out => error message appears
3 - clear the input and focus out => expected "required" message _does not_ appear

Note that it seems that the message div is there but stuck on animation. Manually removing it's `md-input-message-animation` class in chrome devtools reveals the message.

### LINKED ISSUE #6767: md-input-container not showing error ng-messages
Hi,
There are cases when the error message (ng-message) not showing, try to play with the codepen below.
the input field is required and minlength set to 2:
http://codepen.io/vlio20/pen/KVybvx?editors=110

Steps to reproduce:
1. focus on input.
2. enter 1 char & focus out (blur) -> minlength error message is displayed.
3. delete the char from step 2 -> no message is displayed, when the required message should have been displayed.

Note-1: The form is invalid with correct message

Note-2: in happens for other validations as well (other then required and minlength).
 If I am overriding the styles like bellow, it seems to be working, but the animations are not good when the error message is changed:

```
md-input-container.md-input-invalid .md-auto-hide .md-input-message-animation:not(.ng-animate) {
  opacity: 1;
  margin-top: 0;
}

md-input-container.md-input-invalid [ng-messages] {
  max-height: 19px;
}
```

I am using 1.0.2 version, but it seems to be broken since 1.0.0

DevVersion:
Currently we only check for the `md-input-invalid` class, but that won't work, because if you're input is already invalid, and another invalid criterium appears, then we can't detect the difference.
The input container still thinks, it's only one message, from the start.

@topherfangio Please take a look as well.

FYI: This is still not fixing the duplicate messages bug (-> this is in progress)

Fixes #6767. Fixes #6399. References #6443.

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -723,16 +723,18 @@ function ngMessageDirective($mdUtil) {
 function mdInputInvalidMessagesAnimation($q, $animateCss) {
   return {
     addClass: function(element, className, done) {
-      var messages = getMessagesElement(element);
+      if (className == 'md-input-invalid') {
 
-      if (className == "md-input-invalid" && messages.hasClass('md-auto-hide')) {
-        showInputMessages(element, $animateCss, $q).finally(done);
-      } else {
-        done();
+        if (isAutoHide(element)) {
+          showInputMessages(element, $animateCss, $q).finally(done);
+          return;
+        }
       }
+
+      done();
     }
 
-    // NOTE: We do not need the removeClass method, because the message ng-leave animation will fire
+    // NOTE: We do not need the removeClass method, because the message ng-leave animation will fire.
   };
 }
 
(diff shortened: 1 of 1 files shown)


topherfangio:
@DevVersion Apologies, but I've been working on this issue already. We're trying to get some really good tests written so that we can make sure we fix these issues and don't regress in the future. I'm actually meeting with someone in about 5 minutes to hopefully finalize these tests.

DevVersion:
@topherfangio No worries, it's not a problem :smile: I should have asked you, but as always I consider this as a good experience.

So looking forward to see your changes!

PR #7603, opened 2016-03-16
fix(datepicker): ensure the element always blends in with it's parent

### LINKED ISSUE #7502: md-date-picker no longer transparent
My datapicker used to be transparent (i have it in a card that has a white background):

image

No it has a slight grey color to it..

crisbeto:
fix(datepicker): ensure the element always blends in with it's parent
* Makes the datepicker component transparent so it always blends in with it's parent.
* Prevents the content around a datepicker from jumping when it gets opened.

Closes #7502.

### DIFF
--- a/src/components/datepicker/datePicker-theme.scss
+++ b/src/components/datepicker/datePicker-theme.scss
@@ -1,15 +1,10 @@
 /** Theme styles for mdDatepicker. */
 
-md-datepicker.md-THEME_NAME-theme {
-  background: '{{background-hue-1}}';
-}
-
 .md-THEME_NAME-theme {
 
   .md-datepicker-input {
     @include input-placeholder-color('\'{{foreground-3}}\'');
     color: '{{foreground-1}}';
-    background: '{{background-color}}';
   }
 
   .md-datepicker-input-container {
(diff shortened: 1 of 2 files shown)


topherfangio:
I am not sure this is ready quite yet. If I manually set the background of my container to something like pink, there is still some part of the datepicker that is set to an incorrect color.

datepicker-background

crisbeto:
I'm not quite sure that we'd be able to accommodate for all of these cases though (e.g. a grandparent having a background etc.). The alternative could be to set it as transparent and have it inherit only when it's open?

topherfangio:
I think transparent is okay, but I currently don't understand why we're even setting the background for input elements at all. Is there ever a case in the spec where it doesn't have the same background as the parent container?

crisbeto:
I don't think so. I'll make it transparent and try out a few more scenarios just in case.

crisbeto:
I've updated it @topherfangio. Now it's transparent when closed and when it's open, it uses the body background. I also managed to fix an issue where the datepicker would push it's surrounding content when it gets opened.

topherfangio:
LGTM :+1:

PR #7605, opened 2016-03-16
fix(sidenav): animation when width is explicitly defined

### LINKED ISSUE #7483: mdSideNav: Animation is broken when specifing "style="width: ...px" attribute
I have a scenario when a sidenav has ability for resizing dynamicly.

For that purpose I add **"style="width: ...px"** attribute for **<'md-sidenav'>** element.
Resizing works perfectly. 

But after attribute is added animation of the toggle is broken. Without it animation works.

Here is [Codepen](http://codepen.io/anon/pen/qZZxQB)

Thanks!

EladBezalel:
- Made `width` and `min-width` important for the animation

fixes #7483


ThomasBurleson:
@EladBezalel  - The slideIn navigation for the left sideNav is broken with this change: https://material.angularjs.org/latest/demo/sidenav

Issue #7464, opened 2016-03-08
mdPanel: Add TrapFocus config option.


KarenParker:
Whether focus should be trapped within the panel. If `trapFocus` is true, the user will not be able to interact with the rest of the page until the panel is dismissed. Defaults to false.

See mdDialog implementation [here](https://github.com/angular/material/blob/a2eca807832ad31797fd5ad7801ea594ed01ff52/src/components/dialog/dialog.js#L924).

Note: Make sure to remove the focus traps when the panel is removed.

------------
This task involves:

* Updating public documentation
* Adding elements to the beginning and end of the panel (as siblings), which can trap focus.
* Event focus handling for the panel when the focus traps are hit.

PR #7606, opened 2016-03-17
feat(panel): Add TrapFocus config option to the panel.

KarenParker:
@gmoothart @ErinCoughlan @DerekLouie @jelbourn 

Please review.

Fixes issue #7464

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -112,6 +112,10 @@ angular
  *     outside the panel to close it. Defaults to false.
  *   - `escapeToClose` - `{boolean=}`: Whether the user can press escape to
  *     close the panel. Defaults to false.
+ *   - `trapFocus` - `{boolean=}`: Whether focus should be trapped within the
+ *     panel. If `trapFocus` is true, the user will not be able to interact
+ *     with the rest of the page until the panel is dismissed. Defaults to
+ *     false.
  *
  * TODO(ErinCoughlan): Add the following config options.
  *   - `groupName` - `{string=}`: Name of panel groups. This group name is
@@ -128,10 +132,6 @@ angular
  *   - `fullScreen` - `{boolean=}`: Whether the panel should be full screen.
  *     Applies the class `.md-panel-fullscreen` to the panel on open. Defaults
  *     to false.
- *   - `trapFocus` - `{boolean=}`: Whether focus should be trapped within the
- *     panel. If `trapFocus` is true, the user will not be able to interact
- *     with the rest of the page until the panel is dismissed. Defaults to
- *     false.
  *   - `focusOnOpen` - `{boolean=}`: An option to override focus behavior on
  *     open. Only disable if focusing some other way, as focus management is
  *     required for panels to be accessible. Defaults to true.
@@ -465,6 +465,7 @@ function MdPanelService($rootElement, $rootScope, $injector) {
     escapeToClose: false,
     scope: $rootScope.$new(true),
     transformTemplate: angular.bind(this, this.wrapTemplate_),
+    trapFocus: false,
     zIndex: defaultZIndex
   };
 
@@ -605,6 +606,12 @@ function MdPanelRef(config, $injector) {
 
   /** @private {Array<function()>} */
   this._removeListeners = [];
+  
+  /** @private {!angular.JQLite|undefined} */
+  this._topFocusTrap;
+
+  /** @private {!angular.JQLite|undefined} */
+  this._bottomFocusTrap;
 }
 
 
@@ -695,6 +702,16 @@ MdPanelRef.prototype.detachOnly = function() {
   var self = this;
   this._detachPromise = this._$q(function(resolve) {
     self._removeEventListener();
+
+    // Remove the focus traps that we added earlier for keeping focus within the panel.
+    if (self._topFocusTrap && self._topFocusTrap.parentNode) {
+      self._topFocusTrap.parentNode.removeChild(self._topFocusTrap);
+    }
+
+    if (self._bottomFocusTrap && self._bottomFocusTrap.parentNode) {
+      self._bottomFocusTrap.parentNode.removeChild(self._bottomFocusTrap);
+    }
+
     self._panelContainer.remove();
     self.isAttached = false;
     resolve(self);
@@ -812,6 +829,7 @@ MdPanelRef.prototype._createPanel = function() {
           }
 
           self._addStyles();
+          self._configureAria();
           resolve(self);
         }, reject);
   });
@@ -896,6 +914,16 @@ MdPanelRef.prototype._configureEscapeToClose = function() {
 
 
 /**
+ * Add ARIA-specific attributes appropriate for panels.
+ */
+MdPanelRef.prototype._configureAria = function() {
+  this._panelEl.attr('tabIndex', '-1');
+
+  this._configureTrapFocus();
+};
+
+
+/**
  * Setup the clickOutsideToClose event listeners.
  * @private
  */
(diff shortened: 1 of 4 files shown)


src/components/panel/panel.js
```diff
@@ -832,6 +850,47 @@ MdPanelRef.prototype._addStyles = function() {
 };
 
 
+/**
+ * Add ARIA-specific attributes appropriate for panels.
+ */
+MdPanelRef.prototype._configureAria = function() {
```
jelbourn: nit: "ARIA" refers to the extra attributes on top of html that screen readers understand. Focus trapping is unrelated to ARIA; it just _also_ improves accessibility. I would just call `configureFocusTrap` directly (and set the `tabindex` there).

src/components/panel/panel.js
```diff
@@ -832,6 +850,47 @@ MdPanelRef.prototype._addStyles = function() {
...
+ * @private
+ */
+MdPanelRef.prototype._configureTrapFocus = function() {
+  if (this._config.trapFocus) {
+    var element = this._panelEl;
+    // Set up elements before and after the panel to capture focus and
+    // redirect back into the panel.
+    this._topFocusTrap = document.createElement('div');
+    this._topFocusTrap.classList.add('_md-panel-focus-trap');
+    this._topFocusTrap.tabIndex = 0;
```
jelbourn: I'd create one constant `HTMLElement` like `FOCUS_TRAP_TEMPLATE` with these properties already set and then clone it for both the top and bottom.

src/components/panel/panel.js
```diff
@@ -832,6 +850,47 @@ MdPanelRef.prototype._addStyles = function() {
...
+    var focusHandler = function () {
+      element.focus();
+    };
+    this._topFocusTrap.addEventListener('focus', focusHandler);
+    this._bottomFocusTrap.addEventListener('focus', focusHandler);
+
+    // The top focus trap inserted immeidately before the md-panel element (as a sibling).
+    // The bottom focus trap is inserted at the very end of the md-panel element (as a child).
+    element[0].parentNode.insertBefore(this._topFocusTrap, element[0]);
+    element.after(this._bottomFocusTrap);
```
jelbourn: Didn't @ErinCoughlan fix a bug in $mdDialog where having the bottom focus trap as a child caused some bug?

src/components/panel/panel.js
```diff
@@ -832,6 +850,47 @@ MdPanelRef.prototype._addStyles = function() {
...
+    this._topFocusTrap = document.createElement('div');
+    this._topFocusTrap.classList.add('_md-panel-focus-trap');
+    this._topFocusTrap.tabIndex = 0;
+
+    this._bottomFocusTrap = this._topFocusTrap.cloneNode(false);
+
+    // When focus is about to move out of the panel, we want to intercept it and redirect it
+    // back to the panel element.
+    var focusHandler = function () {
+      element.focus();
```
jelbourn: Might be worth mentioning that this _wraps_ the bottom from the bottom, but simply "sticks" it when shift-tabbing, which isn't the _best_ approach.

(I recently had someone tell me that the only perfect way of focus trapping is to actually set the `tabIndex` on everything _outside_ the trapped area to `-1` and then set it back when you're done trapping).

src/components/panel/panel.js
KarenParker: I was matching mdDialog since we will likely have more content here later, but I am fine with removing it for now.

src/components/panel/panel.js
KarenParker: This is copied directly from mdDialog so if there was a fix, this is it.

KarenParker:
Comments addressed.

jelbourn:
LGTM

src/components/panel/panel.spec.js
```diff
@@ -91,6 +91,52 @@ describe('$mdPanel', function() {
     expect(PANEL_WRAPPER_CLASS).not.toHaveClass(HIDDEN_CLASS);
   });
 
+  it('should trap focus inside of the panel with trapFocus=true', function() {
```
gmoothart: Can you rename this test? Something like "should create and clean up focus traps"; as the TODO mentions you can't actually test the focus behavior but the test description makes it sound as if you are.

src/components/panel/panel.js
```diff
@@ -938,6 +958,35 @@ MdPanelRef.prototype._configureClickOutsideToClose = function() {
 };
 
 
+/**
+ * Setup the focus traps. These traps will wrap focus when tabbing past the
+ * panel. When shift-tabbing, the focus will stick in place.
+ * @private
+ */
+MdPanelRef.prototype._configureTrapFocus = function() {
+  this._panelEl.attr('tabIndex', '-1');
```
gmoothart: Can you explain with this is necessary?

src/components/panel/panel.spec.js
```diff
@@ -91,6 +91,52 @@ describe('$mdPanel', function() {
...
+    var isPanelFocused = false;
+    panel[0].addEventListener('focus', function() {
+      isPanelFocused = true;
+    });
+
+    // Both of the focus traps should be in the normal tab order.
+    expect(topTrap.tabIndex).toBe(0);
+    expect(bottomTrap.tabIndex).toBe(0);
+
+    // TODO(jelbourn): Find a way to test that focusing the traps redirects focus to the
```
ErinCoughlan: Change the TODO to your name or someone on our team?

src/components/panel/panel.spec.js
```diff
@@ -91,6 +91,52 @@ describe('$mdPanel', function() {
...
+    // md-dialog element. Firefox is problematic here, as calling element.focus() inside of
+    // a focus event listener seems not to immediately update the document.activeElement.
+    // This is a behavior better captured by an e2e test.
+
+    closePanel();
+
+    // All of the focus traps should be removed when the dialog is closed.
+    focusTraps = document.querySelectorAll('._md-panel-focus-trap');
```
ErinCoughlan: Perhaps add the class to the list at the top so it's easy to change if needed?

src/components/panel/panel.spec.js
```diff
@@ -91,6 +91,52 @@ describe('$mdPanel', function() {
     expect(PANEL_WRAPPER_CLASS).not.toHaveClass(HIDDEN_CLASS);
   });
 
+  it('should trap focus inside of the panel with trapFocus=true', function() {
+    var template = '<div>Hello World!</div>';
```
ErinCoughlan: optional: We've now used this template probably a dozen tests. What do you think of moving it to the top of the file as a constant? DEFAULT_TEMPLATE perhaps? It doesn't actually matter for the tests.

src/components/panel/panel.spec.js
ErinCoughlan: Should these tests be under the config options describe block? Or as a separate describe for a11y?

src/components/panel/panel.spec.js
```diff
@@ -91,6 +91,52 @@ describe('$mdPanel', function() {
...
+    // md-dialog element. Firefox is problematic here, as calling element.focus() inside of
+    // a focus event listener seems not to immediately update the document.activeElement.
+    // This is a behavior better captured by an e2e test.
+
+    closePanel();
+
+    // All of the focus traps should be removed when the dialog is closed.
+    focusTraps = document.querySelectorAll('._md-panel-focus-trap');
+    expect(focusTraps.length).toBe(0);
+  })
```
ErinCoughlan: Missing semicolon here and the next test.

src/components/panel/panel.js
KarenParker: In truth, I have no clue why it works with this line and doesn't work without it. This was code specified in the focus traps of dialog.

KarenParker:
Comments addressed.

ErinCoughlan:
LGTM. Make sure to squash the 3 commits before merging to feat/panel. It also looks like there's a comment update from Jeremy.

PR #7608, opened 2016-03-17
Add @ngInject to core.js files.

mhansen:
Otherwise the closure compiler's renaming will break the injection.


mhansen:
#7610

crisbeto:
LGTM :+1:

Issue #7610, opened 2016-03-17
Many injected functions don't have @ngInject


mhansen:
A lot of functions that have injected services are annotated with @ngInject, but a lot aren't.

Closure compiler renames function arguments, so you need to add @ngInject to these for them to work after compilation.

topherfangio:
@crisbeto Can you take a look at this one and see how much work is required to do this?

crisbeto:
Sure. Also would be a good chance to make them all consistent (we have some that are in the doc strings and apparently that shows a warning).

topherfangio:
Awesome, thanks!

mhansen:
Has that fix been tested by running it thru the closure compiler? I thought
the closure compiler only interpreted jsdoc annotations - starting with two
stars (/** */), though I may be wrong on this.

crisbeto:
I think your changes got merged. We determined that the changes I did in #7744 aren't necessary.

PR #7611, opened 2016-03-17
Add @ngInject to $mdUtil

mhansen:
Without this, closure compiler renaming breaks injection. #7610


crisbeto:
LGTM :+1:

DevVersion:
@crisbeto @mhansen I think it would be better if you combine all these changes into one PR, having four PR's for all the same issue (with one line change) seems for me really weird. Maybe @mhansen can combine all the changes into one. (And the typo you detected @crisbeto)

PR #7613, opened 2016-03-17
fix(menu): typo in @ngInject annotation

crisbeto:
It was labeled as `@ngInjdect` instead of `@ngInject`.

Relates to #7610.


ThomasBurleson:
Fixed with #7613

PR #7619, opened 2016-03-17
fix(dialog): dialog should support skipCompile from interim element.

### LINKED ISSUE #7566: mdDialog: is there any way to provide already existing content (domNode)?
versions 1.0.6, 1.1.0-rc1

I've been playing around the mdDialog that I'd like to use as a UI for the settings of the application. The content of the settings uses other Angular Material components (buttons, switches, selects), let's say up to 12 components.

It works reasonably well on the non IE browsers, but on IE it takes a really long time to render the template and inject it with all the scope values and to render the inner components. If the template is a simple text it's realtively quick.

I was wondering, is there any way I could provide the mdDialog already rendered content (with display: none; ) and keep the reference to this dialog somewhere (or I could place this content some elsewhere on hide)? Or does the mdDialog always constructs the inner content and destroys it?

Thanks

DevVersion:
* Makes dialog compatible with `options.skipCompile` from the interim element service.

Fixes #7566.

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -891,6 +891,11 @@ function MdDialogProvider($$interimElementProvider) {
      * Inject ARIA-specific attributes appropriate for Dialogs
      */
     function configureAria(element, options) {
+      // If the element is undefined, then there was no <md-dialog> element present.
+      // This can be caused by using the options.skipCompile and options.element properties.
+      // In this case, the template processing and compilation will be skipped and wont
+      // add the <md-dialog> automatically.
+      if (!element.length) return;
 
       var role = (options.$type === 'alert') ? 'alertdialog' : 'dialog';
       var dialogContent = element.find('md-dialog-content');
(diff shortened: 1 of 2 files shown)


src/components/dialog/dialog.js
```diff
@@ -891,6 +891,11 @@ function MdDialogProvider($$interimElementProvider) {
      * Inject ARIA-specific attributes appropriate for Dialogs
      */
     function configureAria(element, options) {
+      // If the element is undefined, then there was no <md-dialog> element present.
+      // This can be caused by using the options.skipCompile and options.element properties.
+      // In this case, the template processing and compilation will be skipped and wont
+      // add the <md-dialog> automatically.
+      if (!element.length) return;
```
topherfangio: I feel like this would be clearer if it specifically checked for the options and returned instead of checking element length and adding a comment about how it could get into this state.

topherfangio:
@DevVersion No rush; just wanted to ping you on this to see if you had any questions or needed any guidance 😄

DevVersion:
@topherfangio Let's close this as the old PR - I need to start from scratch again for this feature.

Issue #7465, opened 2016-03-08
mdPanel: Add FocusOnOpen (autofocus) config option


KarenParker:
An option to override focus behavior on open. Only disable if focusing some other way, as focus management is required for panels to be accessible. Defaults to true.

Focus should be chained to the open and show promises.

Consider following the mdDialog implementation [here](https://github.com/angular/material/blob/a2eca807832ad31797fd5ad7801ea594ed01ff52/src/components/dialog/dialog.js#L641).

-----------
This task involves:

* Updating public documentation
* Adding focus to the panel config.

PR #7623, opened 2016-03-17
feat(panel): Add focusOnOpen config option to the panel.

KarenParker:
@gmoothart @DerekLouie @jelbourn @ErinCoughlan 

Please review.

Fixes issue #7465

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -116,6 +116,9 @@ angular
  *     panel. If `trapFocus` is true, the user will not be able to interact
  *     with the rest of the page until the panel is dismissed. Defaults to
  *     false.
+ *   - `focusOnOpen` - `{boolean=}`: An option to override focus behavior on
+ *     open. Only disable if focusing some other way, as focus management is
+ *     required for panels to be accessible. Defaults to true.
  *
  * TODO(ErinCoughlan): Add the following config options.
  *   - `groupName` - `{string=}`: Name of panel groups. This group name is
@@ -132,9 +135,6 @@ angular
  *   - `fullScreen` - `{boolean=}`: Whether the panel should be full screen.
  *     Applies the class `.md-panel-fullscreen` to the panel on open. Defaults
  *     to false.
- *   - `focusOnOpen` - `{boolean=}`: An option to override focus behavior on
- *     open. Only disable if focusing some other way, as focus management is
- *     required for panels to be accessible. Defaults to true.
  *
  * @returns {MdPanelRef} panelRef
  */
@@ -282,6 +282,13 @@ angular
  *
  * @returns {!angular.$q.Promise} A promise that is resolved when the panel has
  * toggled the class and animations are completed.
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#focusOnOpen
+ * @description
+ * Focuses the panel content if the focusOnOpen config value is true.
  */
 
 
@@ -465,6 +472,7 @@ function MdPanelService($rootElement, $rootScope, $injector) {
     bindToController: true,
     clickOutsideToClose: false,
     escapeToClose: false,
+    focusOnOpen: true,
     scope: $rootScope.$new(true),
     transformTemplate: angular.bind(this, this.wrapTemplate_),
     trapFocus: false,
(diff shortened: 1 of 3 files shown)


src/components/panel/panel.js
```diff
@@ -661,6 +676,9 @@ MdPanelRef.prototype.attach = function() {
   var self = this;
   this._attachPromise = this._$q(function(resolve, reject) {
     self._createPanel().then(function() {
+      self._$rootScope.$applyAsync(function() {
```
jelbourn: Why do you need to trigger a digest cycle?

ThomasBurleson:
@KarenParker - would you mind merging directly into `upstream feat/panel` and close this PR?

jelbourn:
Review wasn't finished yet

src/components/panel/panel.js
KarenParker: I am using md-autofocus and it hasn't added the class _md-autofocus by this point. Without that class, the mdUtil.findFocusTarget can't find the right spot. I don't know if we need to fix something in createPanel to wait longer, but this is the fix I found. Do you have a different suggestion for how to wait longer to determine what to focus?

jelbourn: Can you add that explanation as a comment? I'm okay with it aside from that.

DerekLouie:
LGTM

src/components/panel/panel.js
```diff
@@ -559,7 +569,13 @@ function MdPanelRef(config, $injector) {
   this._$mdCompiler = $injector.get('$mdCompiler');
 
   /** @private */
```
ErinCoughlan: Should this have a type and the `@const` annotation?

src/components/panel/panel.spec.js
```diff
@@ -627,6 +651,7 @@ describe('$mdPanel', function() {
       panelRef = $mdPanel.open();
     }
     $rootScope.$apply();
+    $rootScope.$apply();
```
ErinCoughlan: Should there be two `$rootScope.$apply()` here?

src/components/panel/panel.spec.js
```diff
@@ -344,6 +344,30 @@ describe('$mdPanel', function() {
       // expect(panelRef).toBeUndefined();
       expect(PANEL_EL).not.toExist();
     }));
+
+    it('should focus on open', function() {
+      var htmlContent = 'Donuts';
```
ErinCoughlan: This htmlContent doesn't matter for this test. You can just add it directly to the template. Same with the below test.

src/components/panel/panel.js
```diff
@@ -284,6 +284,15 @@ angular
  * toggled the class and animations are completed.
  */
 
+/**
+ * @ngdoc method
+ * @name MdPanelRef#focusOnOpen
+ * @description
+ * Focuses the panel content if the focusOnOpen config value is true. This should
```
ErinCoughlan: I wonder if we should call this method automatically in the addClass state (if focusOnOpen is true). My concern with having the user do it is that they will forget and then the panel will not be accessible.

I know you are using addClass in createPanel, but I think that executing the applyAsync and focus any time after compile should also work.

src/components/panel/panel.spec.js
KarenParker: Yes, this handles the extra digest that is sometimes needed.

src/components/panel/panel.js
KarenParker: I agree, but the problem is do we call it on addClass or removeClass? If you have a class that is "display:none" you would need it on the remove class. I also can't check the visibility/display/etc of the panel since I don't know how the user will be hiding it.

Do you have a solution?

KarenParker:
Comments addressed.

src/components/panel/panel.js
ErinCoughlan: I wonder if we should just do it on both? Focusing on something with `display: none` or `visibility: hidden` doesn't do anything. The only issue would be `opacity: 0`, but that element is accessible to screen readers anyways, so they can still get to it and interact with it (therefore setting the focus doesn't seem too bad).

KarenParker: I added it to both locations.

ErinCoughlan:
LGTM

KarenParker:
@jelbourn Any more comments?

jelbourn:
LGTM

PR #7625, opened 2016-03-17
feat(datepicker): add date-range attributes to md-datepicker directive

hsueing3:
Initial refactoring work to add md-datepicker support for both the ngModel and md-begin-date/md-end-date use-case in preparation for date-range picker functionality.

@jelbourn 

Note: This change will not result in begin/end date selection functionality via the md-calendar.


hsueing3:
I signed it!

jelbourn:
I will try to review this later today, but I'm currently primary on the angular2 merge/sync rotation, so I might not get a chance.

src/components/datepicker/datePicker.js
```diff
@@ -21,16 +21,20 @@
    * @module material.components.datepicker
    *
    * @param {Date} ng-model The component's model. Expects a JavaScript Date object.
+   * @param {Date} md-begin-date The begin date of a date range. Expects a JavaScript Date object.
```
jelbourn: I'd say `start-date` instead of `begin-date`

src/components/datepicker/datePicker.js
```diff
@@ -603,10 +913,23 @@
   /**
    * Sets whether the input is currently focused.
    * @param {boolean} isFocused
+   * @param {boolean} isBeginDate
    */
-  DatePickerCtrl.prototype.setFocused = function(isFocused) {
-    if (!isFocused) {
-      this.ngModelCtrl.$setTouched();
+  DatePickerCtrl.prototype.setFocused = function(isFocused, isBeginDate) {
```
jelbourn: Have a function where the signature is `(boolean boolean)` is generally something you want to avoid. Can we break this up into two functions?

src/components/datepicker/datePicker.js
```diff
@@ -82,26 +90,41 @@
...
+            throw Error('md-datepicker should not be placed inside md-input-container.');
+          }
+
+          mdDatePickerCtrl.configureNgModel(ngModelCtrl, scope);
+        } else {
+          var mdDatePickerCtrl = controllers[0];
 
-        mdDatePickerCtrl.configureNgModel(ngModelCtrl);
+          var mdInputContainer = controllers[1];
+          if (mdInputContainer) {
```
jelbourn: This check and error don't need to be duplicated.

src/components/datepicker/datePicker.js
```diff
@@ -59,8 +65,10 @@
           '</md-button>' +
           '<div class="md-datepicker-input-container" ' +
               'ng-class="{\'md-datepicker-focused\': ctrl.isFocused}">' +
-            '<input class="md-datepicker-input" aria-haspopup="true" ' +
-                'ng-focus="ctrl.setFocused(true)" ng-blur="ctrl.setFocused(false)">' +
+            '<input class="md-datepicker-input begin" aria-haspopup="true" ' +
```
jelbourn: All CSS classes must be prefixed, and ideally be self-explanatory. It should be `md-datepicker-start-date`.

src/components/datepicker/datePicker.js
```diff
@@ -179,11 +202,23 @@
     this.ngModelCtrl = null;
 
     /** @type {HTMLInputElement} */
-    this.inputElement = $element[0].querySelector('input');
+    this.inputElement = $element[0].querySelector('input.begin');
```
jelbourn: When you change the CSS class, this can just be `querySelector('.md-datepicker-start-date');`

src/components/datepicker/datePicker.js
```diff
@@ -179,11 +202,23 @@
...
 
     /** @final {!angular.JQLite} */
     this.ngInputElement = angular.element(this.inputElement);
 
+    /** @type {HTMLInputElement} */
+    this.beginInputElement = $element[0].querySelector('input.begin');
+
+    /** @final {!angular.JQLite} */
+    this.ngBeginInputElement = angular.element(this.beginInputElement);
```
jelbourn: Do you really need both the native and the wrapped elements? I'd prefer to just use the native elements throughout the code.

src/components/datepicker/datePicker.js
```diff
@@ -211,6 +246,29 @@
     /** @type {Date} */
     this.date = null;
 
+    /** @type {Date} */
+    this.beginDate = null;
```
jelbourn: `startDate` (here and elsewhere)

src/components/datepicker/datePicker.js
```diff
@@ -211,6 +246,29 @@
...
+
+    /** @type {Date} */
+    this.endDate = null;
+
+    /** @type {string} */
+    this.beginDateViewValue = null;
```
jelbourn: Need explanation for what these "view values" are and why you need them

src/components/datepicker/datePicker.js
```diff
@@ -211,6 +246,29 @@
...
+
+    /** @type {string} */
+    this.endDateViewValue = null;
+
+    /** @type {Object} */
+    this.dateRangeValidity = {
```
jelbourn: Need explanation of what this is and why it's needed.

src/components/datepicker/datePicker.js
```diff
@@ -211,6 +246,29 @@
...
+      valid: true
+    };
+
+    /** @type {boolean} */
+    this.isBeginDateTouched = false;
+
+    /** @type {boolean} */
+    this.isEndDateTouched = false;
```
jelbourn: Add comments explaining what these ("touched") are for.

src/components/datepicker/datePicker.js
```diff
@@ -221,6 +279,9 @@
     /** @type {boolean} Whether the date-picker's calendar pane is open. */
     this.isCalendarOpen = false;
 
+    /** @type {boolean} Whether or not the datepicker is being used as a date-range picker. */
+    this.isDateRangePicker = null;
```
jelbourn: Can you not set this here and now?

src/components/datepicker/datePicker.js
```diff
@@ -260,22 +321,62 @@
...
   DatePickerCtrl.prototype.configureNgModel = function(ngModelCtrl) {
-    this.ngModelCtrl = ngModelCtrl;
-
     var self = this;
-    ngModelCtrl.$render = function() {
-      var value = self.ngModelCtrl.$viewValue;
 
-      if (value && !(value instanceof Date)) {
-        throw Error('The ng-model for md-datepicker must be a Date instance. ' +
+    if (ngModelCtrl == null && self.begin != undefined && self.end != undefined) {
```
jelbourn: This function should be broken up into smaller functions; at least one for the ngModel setup and one for the range setup, with possibly other functions for anything shared between them.

src/components/datepicker/datePicker.js
```diff
@@ -286,19 +387,55 @@
...
-      self.closeCalendarPane();
-      self.resizeInputElement();
-      self.updateErrorState();
-    });
-
-    self.ngInputElement.on('input', angular.bind(self, self.resizeInputElement));
-    // TODO(chenmike): Add ability for users to specify this interval.
-    self.ngInputElement.on('input', self.$mdUtil.debounce(self.handleInputEvent,
+    // Cannot use isDateRangePicker flag here as this even listener is wired up before the flag gets assigned a value.
+    if (self.$scope.ctrl.begin == undefined && self.$scope.ctrl.end == undefined) {
```
jelbourn: `self.$scope.ctrl` > `self` (or `this` if it's not in a function expression)

jelbourn: Use `== null` instead of `== undefined`

src/components/datepicker/datePicker.js
```diff
@@ -286,19 +387,55 @@
...
-      self.inputElement.value = self.dateLocale.formatDate(date);
-      self.closeCalendarPane();
-      self.resizeInputElement();
-      self.updateErrorState();
-    });
-
-    self.ngInputElement.on('input', angular.bind(self, self.resizeInputElement));
-    // TODO(chenmike): Add ability for users to specify this interval.
-    self.ngInputElement.on('input', self.$mdUtil.debounce(self.handleInputEvent,
+    // Cannot use isDateRangePicker flag here as this even listener is wired up before the flag gets assigned a value.
```
jelbourn: This function also needs to be broken up into smalle rfunctions

src/components/datepicker/datePicker.js
```diff
@@ -307,13 +444,30 @@
...
-    self.ngInputElement.on('keydown', function(event) {
-      if (event.altKey && event.keyCode == keyCodes.DOWN_ARROW) {
-        self.openCalendarPane(event);
-        $scope.$digest();
-      }
-    });
+    if (!self.isDateRangePicker) {
+      // Add event listener through angular so that we can triggerHandler in unit tests.
+      self.ngInputElement.on('keydown', function(event) {
+        if (event.altKey && event.keyCode == keyCodes.DOWN_ARROW) {
```
jelbourn: These `keydown` handlers are all exactly the same. They shouldn't be duplicated.

src/components/datepicker/datePicker.js
```diff
@@ -351,7 +516,12 @@
    */
   DatePickerCtrl.prototype.setDisabled = function(isDisabled) {
     this.isDisabled = isDisabled;
-    this.inputElement.disabled = isDisabled;
+    if (!this.isDateRangePicker) {
```
jelbourn: Why check? Why not just always enable/disable them all?

src/components/datepicker/datePicker.js
```diff
@@ -365,55 +535,108 @@
...
 
-    // Clear any existing errors to get rid of anything that's no longer relevant.
-    this.clearErrorState();
+      // Clear any existing errors to get rid of anything that's no longer relevant.
+      this.clearErrorState();
 
-    if (this.dateUtil.isValidDate(date)) {
-      // Force all dates to midnight in order to ignore the time portion.
-      date = this.dateUtil.createDateAtMidnight(date);
+      if (this.dateUtil.isValidDate(date)) {
```
jelbourn: There is a lot of logic duplicated in this method. It can be cleaned up a lot by taking the function as-is and simply grabbing the min date/max date from the appropriate field.

src/components/datepicker/datePicker.js
```diff
@@ -365,55 +535,108 @@
...
   };
 
   /** Clears any error flags set by `updateErrorState`. */
   DatePickerCtrl.prototype.clearErrorState = function() {
     this.inputContainer.classList.remove(INVALID_CLASS);
-    ['mindate', 'maxdate', 'filtered', 'valid'].forEach(function(field) {
-      this.ngModelCtrl.$setValidity(field, true);
-    }, this);
+    if (!this.isDateRangePicker) {
+      ['mindate', 'maxdate', 'filtered', 'valid'].forEach(function(field) {
```
jelbourn: No reason to duplicate this array. And why check `isDateRange` vs. always clearing both?

src/components/datepicker/datePicker.js
jelbourn: Can be shortened to just `isDateRange`

jelbourn:
@hsueing3 finished my first pass; let me know when you're ready for a second.

src/components/datepicker/datePicker.js
hsueing3: The HTMLInputElement does have native impl's for oninput and onkeydown (which is what the wrapped elements are being used for), however the function serving as the event handler does not have the event passed as a param. This event param is needed by the functions currently being passed in. I think this is why this approach was initially used.  e.g. --> http://www.w3schools.com/jsref/event_onkeydown.asp

src/components/datepicker/datePicker.js
hsueing3: I don't think $element[0]/tagName would help as this value ("MD-DATEPICKER") would always be the same regardless of the context under which the the md-datepicker is being used (remeber we are not creating a separate directive). I can however use something like this --> $element[0].hasAttribute('ng-model'). So if that attribute is missing, I can assume that the date-range-picker context is being used. Great suggestion!

src/components/datepicker/datePicker.js
hsueing3: As you mentioned, this check is no longer needed given the impl of your prior mentioned suggestion around initially setting the isDatePicker value.

src/components/datepicker/datePicker.js
hsueing3: Addressed locally.

src/components/datepicker/datePicker.js
hsueing3: Addressed locally.

src/components/datepicker/datePicker.js
hsueing3: Addressed locally.

src/components/datepicker/datePicker.js
hsueing3: Addressed locally.

src/components/datepicker/datePicker.js
hsueing3: Addressed locally.

src/components/datepicker/datePicker.js
hsueing3: Addressed locally.

src/components/datepicker/datePicker.js
hsueing3: Addressed locally.

src/components/datepicker/datePicker.js
hsueing3: Addressed locally.

src/components/datepicker/datePicker.js
hsueing3: Addressed locally.

src/components/datepicker/datePicker.js
hsueing3: Addressed locally.

src/components/datepicker/datePicker.js
hsueing3: Addressed locally.

src/components/datepicker/datePicker.js
hsueing3: Because depending on the context of the directive, this.inputElement or this.startInputElement/this.endInputElement will be null. This means the .disabled call on a null object will be undefined.

src/components/datepicker/datePicker.js
hsueing3: Addressed locally.

src/components/datepicker/datePicker.js
hsueing3: Similarly to a prior comment, depending on the context of the directive, this.ngModelCtrl could be null. This means the $setValidity call on a null object will be an error. Furthermore, i am using the approach throughout this file to fork the code flow, and therefore usage of certain groups of vars depending on the context, to make it clear to the others what vars are being used consistently depending on whether or not the context isDatePicker or not. If I remove the if checks which are based on the isDatePicker, it could be confusing to the next guy or gal who comes behind me to update this file and is trying to understand what vars are actually being used in either the datePicker or dateRangePicker context.

src/components/datepicker/datePicker.js
hsueing3: As mentioned elsewhere on this pull request, I guess I am struggling with the dilemma of duplication with simplicity or DRY with complexity. Tough call. Will discuss in person.

hsueing3:
Addressed feedback on https://github.com/hsueing3/material/commit/bd792ea8e82607b68093df60df906d1c2fa1f09d. Should this pull request be updated or closed after another is created?

ThomasBurleson:
@hsueing3 - What is the status of this PR ?

hsueing3:
@ThomasBurleson I have addressed the feedback comments provided on a later commit in my hsueing3/material branch. What are the typical next steps to close this PR out?

jelbourn:
@ThomasBurleson he's waiting on me to finally get around to a second pass of review comments :)

ThomasBurleson:

hsueing3:
FYI, I'm shifting focus to TypeScript and Angular2 learning/development. I will no longer be working on this feature.

Issue #1912, opened 2015-03-16
feat(mdCheckbox): Add indeterminate state


breandr:
I just replaced a native checkbox with mdCheckbox and realised it doesn't have an indeterminate state.

epelc:
If anyone doesn't know what this is have a look below

https://css-tricks.com/indeterminate-checkboxes/
![image](https://cloud.githubusercontent.com/assets/5204642/6666869/9bb8f78e-cbbe-11e4-88ea-86a9090ee93b.png)

marcysutton:
We have passed this question along to the UX team and will report back.

marcysutton:
I'm still unclear what the use-case for this would be, and the native versions can only be added with JavaScript (not HTML attributes). That said, I think a line icon just like the Chrome version would communicate effectively.

breandr:
Use-case: a select-all check box which is checked whem all children are
checked; unchecked when all children are unchecked; and intermediate when
at least 1 but not all children are selected.

epelc:
@marcysutton My use case would be for user specific permissions. Specifically for overriding defaults from a users role. The indeterminate state would be used to default to the parent role for each permission.

marcysutton:
Cool! thanks for your replies, those help.

ThomasBurleson:
The UX Material team is adding this feature to the specification.

breandr:
Thanks!

rscafi:
+1
Use case: Row selection in tables determine state of "select all" checkbox.

http://www.google.com/design/spec/components/data-tables.html#data-tables-interaction

YuriyVlasenko:
+1

langdonx:
PRs welcome?

	md-checkbox.md-indeterminate .md-icon:after {
		box-sizing: border-box;
		transform: rotate(45deg);
		position: absolute;
		left: $checkbox-width / 3;
		top: $checkbox-width / 9;
		display: table;
		width: $checkbox-width / 3;
		height: $checkbox-width * 2 / 3;
		border-width: $checkbox-border-width;
		border-style: solid;
		border-top: 0;
		border-left: 0;
		content: '';
		transform: rotate(0deg);
		border-right: 0;
		top: -2px;
		width: 10px;
		left: 4px;
	}

[...]

PR #7643, opened 2016-03-19
feat(checkbox): add indeterminate checkbox support

DerekLouie:
@gmoothart @ErinCoughlan @jelbourn @ThomasBurleson 

This adds the indeterminate state to md-checkbox.

Please Review. 

References issue #1912

indetcheckbox
indetunselectall
indetselectall

### DIFF
--- a/src/components/checkbox/checkbox.js
+++ b/src/components/checkbox/checkbox.js
@@ -27,7 +27,12 @@ angular
  * @param {string=} ng-change Angular expression to be executed when input changes due to user interaction with the input element.
  * @param {boolean=} md-no-ink Use of attribute indicates use of ripple ink effects
  * @param {string=} aria-label Adds label to checkbox for accessibility.
- * Defaults to checkbox's text. If no default text is found, a warning will be logged.
+ *     Defaults to checkbox's text. If no default text is found, a warning will be logged.
+ * @param {expression=} md-indeterminate This determines when the checkbox should be rendered as 'indeterminate'.
+ *     If a truthy expression or no value is passed in the checkbox renders in the md-indeterminate state.
+ *     If falsy expression is passed in it just looks like a normal unchecked checkbox.
+ *     The indeterminate, checked, and unchecked states are mutually exclusive. A box cannot be in any two states at the same time. 
+ *     When a checkbox is indeterminate that overrides any checked/unchecked rendering logic.
  *
  * @usage
  * <hljs lang="html">
@@ -55,7 +60,7 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
     transclude: true,
     require: '?ngModel',
     priority: 210, // Run before ngAria
-    template: 
+    template:
       '<div class="_md-container" md-ink-ripple md-ink-ripple-checkbox>' +
         '<div class="_md-icon"></div>' +
       '</div>' +
@@ -69,6 +74,7 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
 
   function compile (tElement, tAttrs) {
     var container = tElement.children();
+    var mdIndeterminateStateEnabled = tAttrs.hasOwnProperty('mdIndeterminate');
 
     tAttrs.type = 'checkbox';
     tAttrs.tabindex = tAttrs.tabindex || '0';
@@ -89,8 +95,13 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
     });
 
     return function postLink(scope, element, attr, ngModelCtrl) {
+      var isIndeterminate;
       ngModelCtrl = ngModelCtrl || $mdUtil.fakeNgModel();
       $mdTheming(element);
+      if (mdIndeterminateStateEnabled) {
+        setIndeterminateState();
+        scope.$watch(attr.mdIndeterminate, setIndeterminateState);
+      }
 
       if (attr.ngChecked) {
         scope.$watch(
@@ -156,6 +167,7 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
           listener(ev);
         }
       }
+
       function listener(ev) {
         if (element[0].hasAttribute('disabled')) {
           return;
@@ -171,11 +183,19 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
       }
 
       function render() {
-        if(ngModelCtrl.$viewValue) {
+        if(ngModelCtrl.$viewValue && !isIndeterminate) {
           element.addClass(CHECKED_CSS);
         } else {
           element.removeClass(CHECKED_CSS);
         }
+      }
+
+      function setIndeterminateState(newValue) {
+        isIndeterminate = newValue !== false;
+        if (isIndeterminate) {
+          element.attr('aria-checked', 'mixed');
+        }
+        element.toggleClass('md-indeterminate', isIndeterminate);
       }
     };
   }
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -9,7 +9,6 @@ $checkbox-margin: 16px !default;
 $checkbox-text-margin: 10px !default;
 $checkbox-top: 12px !default;
 
-
 .md-inline-form {
   md-checkbox {
     margin: 19px 0 18px;
--- a/src/components/checkbox/checkbox.spec.js
+++ b/src/components/checkbox/checkbox.spec.js
@@ -1,6 +1,7 @@
 
 describe('mdCheckbox', function() {
   var CHECKED_CSS = 'md-checked';
+  var INDETERMINATE_CSS = 'md-indeterminate';
   var $compile, $log, pageScope, $mdConstant;
 
   beforeEach(module('ngAria', 'material.components.checkbox'));
@@ -245,6 +246,56 @@ describe('mdCheckbox', function() {
       checkbox.triggerHandler('click');
       expect(isChecked(checkbox)).toBe(false);
       expect(checkbox.hasClass('ng-invalid')).toBe(true);
+    });
+
+    describe('with the md-indeterminate attribute', function() {
+
+      it('should set md-indeterminate attr to true by default', function() {
+        var checkbox = compileAndLink('<md-checkbox md-indeterminate></md-checkbox>');
+
+        expect(checkbox).toHaveClass(INDETERMINATE_CSS);
+      });
+
+      it('should be set "md-indeterminate" class according to a passed in function', function() {
+        pageScope.isIndeterminate = function() { return true; };
+
+        var checkbox = compileAndLink('<md-checkbox md-indeterminate="isIndeterminate()"></md-checkbox>');
+
+        expect(checkbox).toHaveClass(INDETERMINATE_CSS);
+      });
+
+      it('should set aria-checked attr to "mixed"', function() {
+        var checkbox = compileAndLink('<md-checkbox md-indeterminate></md-checkbox>');
+
+        expect(checkbox.attr('aria-checked')).toEqual('mixed');
+      });
+
+      it('should never have both the "md-indeterminate" and "md-checked" classes at the same time', function() {
+        pageScope.isChecked = function() { return true; };
+
+        var checkbox = compileAndLink('<md-checkbox md-indeterminate ng-checked="isChecked()"></md-checkbox>');
+
+        expect(checkbox).toHaveClass(INDETERMINATE_CSS);
+        expect(checkbox).not.toHaveClass(CHECKED_CSS);
+      });
+
+      it('should change from the indeterminate to checked state correctly', function() {
+        var checked = false;
+        pageScope.isChecked = function() { return checked; };
+        pageScope.isIndet = function() { return !checked; };
+
+        var checkbox = compileAndLink('<md-checkbox md-indeterminate="isIndet()" ng-checked="isChecked()"></md-checkbox>');
+
+        expect(checkbox).toHaveClass(INDETERMINATE_CSS);
+        expect(checkbox).not.toHaveClass(CHECKED_CSS);
+
+        checked = true;
+        pageScope.$apply();
+
+        expect(checkbox).not.toHaveClass(INDETERMINATE_CSS);
+        expect(checkbox).toHaveClass(CHECKED_CSS);
+      });
+
     });
   });
 });
--- a/src/components/checkbox/demoBasicUsage/index.html
+++ b/src/components/checkbox/demoBasicUsage/index.html
@@ -43,6 +43,18 @@
             Checkbox (md-primary): No Ink
           </md-checkbox>
         </div>
+        <div flex-xs flex="50">
+          <md-checkbox md-indeterminate
+              aria-label="Checkbox Indeterminate" class="md-primary">
+            Checkbox: Indeterminate
+          </md-checkbox>
+        </div>
+        <div flex-xs flex="50">
+          <md-checkbox md-indeterminate aria-label="Checkbox Disabled Indeterminate"
+              ng-disabled="true" class="md-primary">
+            Checkbox: Disabled, Indeterminate
+          </md-checkbox>
+        </div>
       </div>
     </fieldset>
 
--- a/src/components/checkbox/demoSelectAll/script.js
+++ b/src/components/checkbox/demoSelectAll/script.js
@@ -1,0 +1,37 @@
+
+angular.module('checkboxDemo3', ['ngMaterial'])
+
+.controller('AppCtrl', function($scope) {
+  $scope.items = [1,2,3,4,5];
+  $scope.selected = [1];
+  $scope.toggle = function (item, list) {
+    var idx = list.indexOf(item);
+    if (idx > -1) {
+      list.splice(idx, 1);
+    }
+    else {
+      list.push(item);
+    }
+  };
+
+  $scope.exists = function (item, list) {
+    return list.indexOf(item) > -1;
+  };
+
+  $scope.isIndeterminate = function() {
+    return ($scope.selected.length !== 0 &&
+        $scope.selected.length !== $scope.items.length);
+  };
+
+  $scope.isChecked = function() {
+    return $scope.selected.length === $scope.items.length;
+  };
+
+  $scope.toggleAll = function() {
+    if ($scope.selected.length === $scope.items.length) {
+      $scope.selected = [];
+    } else if ($scope.selected.length === 0 || $scope.selected.length > 0) {
+      $scope.selected = $scope.items.slice(0);
+    }
+  };
+});
--- a/src/components/checkbox/demoSelectAll/style.css
+++ b/src/components/checkbox/demoSelectAll/style.css
@@ -1,0 +1,15 @@
+.demo {
+  &-legend {
+    color: #3F51B5;
+  }
+
+  &-fieldset {
+    border-style: solid;
+    border-width: 1px;
+    height: 100%;
+  }
+
+  &-select-all-checkboxes {
+    padding-left: 30px;
+  }
+}
--- a/src/core/style/mixins.scss
+++ b/src/core/style/mixins.scss
@@ -251,4 +251,21 @@
     cursor: default;
   }
 
+  &.md-indeterminate ._md-icon {
+    &:after {
+      box-sizing: border-box;
+      position: absolute;
+      top: 50%;
+      left: 50%;
+      transform: translate(-50%, -50%);
+      display: table;
+      width: $width * 0.6;
+      height: $border-width;
+      border-width: $border-width;
+      border-style: solid;
+      border-top: 0;
+      border-left: 0;
+      content: '';
+    }
+  }
 }
--- a/src/core/style/themes.scss
+++ b/src/core/style/themes.scss
@@ -39,4 +39,10 @@ html, body {
   &#{$checkedSelector} ._md-icon:after {
     border-color: '{{primary-contrast-0.87}}';
   }
+
+  & .md-indeterminate[disabled] {
+    ._md-container {
+      color: '{{foreground-3}}';
+    }
+  }
 }
--- a/test/angular-material-spec.js
+++ b/test/angular-material-spec.js
@@ -8,8 +8,8 @@
 (function() {
 
 
-  // Patch since PhantomJS does not implement click() on HTMLElement. In some 
-  // cases we need to execute the native click on an element. However, jQuery's 
+  // Patch since PhantomJS does not implement click() on HTMLElement. In some
+  // cases we need to execute the native click on an element. However, jQuery's
   // $.fn.click() does not dispatch to the native function on <a> elements, so we
   // can't use it in our implementations: $el[0].click() to correctly dispatch.
   // Borrowed from https://stackoverflow.com/questions/15739263/phantomjs-click-an-element
@@ -119,11 +119,17 @@
     };
 
     /**
-     * Add special matchers used in the Angular-Material specs
-     *
+     * Add special matchers used in the Angular-Material spec.
      */
     jasmine.addMatchers({
 
+      /**
+       * Asserts that an element has a given class name.
+       * Accepts any of:
+       *   {string} - A CSS selector.
+       *   {angular.JQLite} - The result of a jQuery query.
+       *   {Element} - A DOM element.
+       */
       toHaveClass: function() {
         return {
           compare: function(actual, expected) {
@@ -131,7 +137,7 @@
             var classes = expected.trim().split(/\s+/);
 
             for (var i = 0; i < classes.length; ++i) {
-              if (!angular.element(actual).hasClass(classes[i])) {
+              if (!getElement(actual).hasClass(classes[i])) {
                 results.pass = false;
               }
             }
@@ -141,7 +147,7 @@
             results.message = "";
             results.message += "Expected '";
             results.message += angular.mock.dump(actual);
-            results.message += negation + "to have class '" + expected + "'.";
+            results.message += "'" + negation + "to have class '" + expected + "'.";
 
             return results;
           }
@@ -194,9 +200,20 @@
             return results;
           }
         };
-      }
+      },
 
     });
+
+    /**
+     * Returns the angular element associated with a css selector or element.
+     * @param el {string|!angular.JQLite|!Element}
+     * @returns {!angular.JQLite}
+     */
+    function getElement(el) {
+      var queryResult = angular.isString(el) ?
+          document.querySelector(el) : el;
+      return angular.element(queryResult);
+    }
 
   });
 
--- a/src/components/checkbox/checkbox-theme.scss
+++ b/src/components/checkbox/checkbox-theme.scss
@@ -1,5 +1,3 @@
-
-
 md-checkbox.md-THEME_NAME-theme {
   .md-ripple {
     color: '{{accent-600}}';
--- a/src/components/checkbox/demoSelectAll/index.html
+++ b/src/components/checkbox/demoSelectAll/index.html
@@ -1,0 +1,29 @@
+<div ng-controller="AppCtrl" class="md-padding demo">
+  <div layout="row" layout-wrap>
+    <div flex="100" layout="column">
+      <div>
+        <!--
+          In IE, we cannot apply flex directly to <fieldset>
+          @see https://github.com/philipwalton/flexbugs#9-some-html-elements-cant-be-flex-containers
+        -->
+        <fieldset class="demo-fieldset" >
+          <legend class="demo-legend">Using &lt;md-checkbox&gt; with the 'indeterminate' attribute </legend>
+          <div layout="row" layout-wrap flex>
+          <div flex-xs flex="50">
+            <md-checkbox aria-label="Select All"
+                         ng-checked="isChecked()"
+                         md-indeterminate="isIndeterminate()"
+                         ng-click="toggleAll()">
+              <span ng-if="isChecked()">Un-</span>Select All
+            </md-checkbox>
+          </div>
+            <div class="demo-select-all-checkboxes" flex="100" ng-repeat="item in items">
+              <md-checkbox ng-checked="exists(item, selected)" ng-click="toggle(item, selected)">
+               {{ item }}
+              </md-checkbox>
+            </div>
+          </div>
+        </fieldset>
+      </div>
+    </div>
+</div>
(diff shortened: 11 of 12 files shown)


ThomasBurleson:
@DerekLouie - please note that we have a [PR and issue naming convention](https://github.com/angular/material/blob/master/CONTRIBUTING.md#-commit-message-format).

e.g. (for this PR)  `feat(checkbox): add indeterminate checkbox support`

src/components/checkbox/checkbox.js
```diff
@@ -171,12 +197,48 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
+
         if(ngModelCtrl.$viewValue) {
           element.addClass(CHECKED_CSS);
         } else {
           element.removeClass(CHECKED_CSS);
         }
       }
+
+      function setIndeterminateState() {
+        var isIndeterminate = element.attr('indeterminate') === 'true' || false;
```
DevVersion: @DerekLouie I think it would be better, if you use 
```js
$mdUtils.parseAttributeValue(attr.indeterminate)
```

src/components/checkbox/checkbox.js
```diff
@@ -171,12 +197,48 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
+          } else {
+            element.removeClass(CHECKED_CSS);
+          }
+        }
+
+        if (isIndeterminate) {
+          element[0].indeterminate = true;
+        } else {
+          element[0].indeterminate = undefined;
+        }
```
DevVersion: I think a ternary operator will be better here:

```js
element[0].indeterminate = isIndeterminate ? true : undefined;
```

src/components/checkbox/checkbox.scss
```diff
@@ -105,6 +105,24 @@ $checkbox-top: 12px !default;
     cursor: default;
   }
 
+  &[indeterminate="true"] ._md-icon{
```
DevVersion: Can you not apply a custom class to the element (like `.md-indeterminate`)? 

This will be more compatible with the attribute value.

src/components/checkbox/checkbox.js
```diff
@@ -69,6 +80,7 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
 
   function compile (tElement, tAttrs) {
     var container = tElement.children();
+    var indeterminateStateEnabled = tElement[0].hasAttribute('indeterminate');
```
crisbeto: You can replace this one with `tAttrs.hasOwnProperty('indeterminate')` so you don't have to touch the DOM.

src/components/checkbox/demoSelectAll/script.js
```diff
@@ -0,0 +1,41 @@
...
+    }
+  };
+
+  $scope.isChecked = function() {
+    if ($scope.selected.length === $scope.items.length) {
+      return true;
+    } else {
+      return false;
+    }
+  };
```
crisbeto: This whole check can be replaced with `return $scope.selected.length === $scope.items.length`.

src/components/checkbox/checkbox.js
```diff
@@ -156,11 +175,18 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
       }
+
       function listener(ev) {
         if (element[0].hasAttribute('disabled')) {
           return;
         }
 
+        if (indeterminateStateEnabled) {
+          if (attr.indeterminateClick) {
+            scope.$apply(element.attr('indeterminate-click'));
```
crisbeto: You should be able to rewrite this one as `scope.$apply(attr.indeterminateClick);`.

src/components/checkbox/checkbox.js
```diff
@@ -156,11 +175,18 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
           listener(ev);
         }
       }
+
       function listener(ev) {
         if (element[0].hasAttribute('disabled')) {
```
crisbeto: This check can be rewritten as `attr.hasOwnProperty('disabled')`.

src/components/checkbox/checkbox.js
```diff
@@ -171,12 +197,48 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
+        // mutually exclusive states.
+        if (attr.indeterminateCheckedWhen) {
+          var isChecked = scope.$eval(attr.indeterminateCheckedWhen);
+          if (isChecked) {
+            element.addClass(CHECKED_CSS);
+            isIndeterminate = false;
+          } else {
+            element.removeClass(CHECKED_CSS);
+          }
+        }
```
crisbeto: This whole block could be shortened down to:

```javascript
var isChecked = scope.$eval(attr.indeterminateCheckedWhen);

element.toggleClass(CHECKED_CSS, isChecked);
isIndeterminate = !isChecked;
```

src/components/checkbox/checkbox.js
DevVersion: Also I think, that using `$apply` for that is not the right way.

Why aren't you using `$parse(attr.indeterminateClick)(scope)`?

src/components/checkbox/demoSelectAll/script.js
```diff
@@ -0,0 +1,41 @@
...
+    return list.indexOf(item) > -1;
+  };
+
+  $scope.isIndeterminate = function() {
+    if ($scope.selected.length !== 0 &&
+        $scope.selected.length !== $scope.items.length) {
+      return true;
+    } else {
+      return false;
+    }
```
crisbeto: This check can be replaced with `return $scope.selected.length !== 0 && $scope.selected.length !== $scope.items.length;`.

src/components/checkbox/demoSelectAll/script.js
```diff
@@ -0,0 +1,41 @@
+
+angular.module('checkboxDemo3', ['ngMaterial'])
+
+.controller('AppCtrl', function($scope) {
+  $scope.items = [1,2,3,4,5];
+  $scope.selected = [1];
+  $scope.toggle = function (item, list) {
+    var idx = list.indexOf(item);
+    if (idx > -1) list.splice(idx, 1);
+    else list.push(item);
```
crisbeto: Not sure whether writing an `if/else` like this matches the guidelines.

src/components/checkbox/checkbox.js
crisbeto: Isn't that the same as `scope.$eval(attr.indeterminateClick)`?

DevVersion: :+1: Oh I think an `$evalAsync` is better here, because it's evaling in an listener. Thoughts?

crisbeto: Looks like it. Can't be sure since I haven't used `$evalAsync` myself.

src/components/checkbox/demoSelectAll/script.js
DerekLouie: Yeah I noticed that (this style got Apple in a lot of trouble not too long ago..), this code is pasted from [this file](https://github.com/angular/material/blob/master/src/components/checkbox/demoSyncing/script.js)

Should I can change it in both places?

src/components/checkbox/checkbox.js
DerekLouie: I do not believe this is logically equivalent

crisbeto: Fair enough, I'm probably missing a bit of context.

src/components/checkbox/checkbox.js
DerekLouie: Yes, eval would be the correct thing to use here, and I should have noticed the `if` statement had the `attr` I needed :+1:.

As a sidenote: I tried substituting `$evalAsync` in here but it breaks necessary indet state logic.

This is because if this runs asynchronously it does not skip the code I need it to in the `render` function which ensures when a checkbox is indeterminate it does not adhere to the normal "checked' / 'unchecked' click logic.

topherfangio: I ran into some issues with the Speed Dial when I was using `$apply` instead of `$evalAsync` because there were cases when the code could be called during an existing digest and throw an error.

If I recall, this occurred when a user added a custom `ng-click` to the component, so you may want to double-check that case.

src/components/checkbox/checkbox.js
DerekLouie: Ahh so the issue with this is that `$mdUtil.parseAttributeBoolean` will interpret an empty string as `true` which we want instead to be interpreted as `false` (in the case no value is passed in).

DevVersion: @DerekLouie Yes, that's by intention. Shouldn't `<md-checkbox inderterminate>` work?

topherfangio: The convention we usually follow is that if the attribute is found with no value, then it is treated as `true` as this follows standard HTML attributes (readonly, disabled, etc).

So,

 - `intermediate` == true
 - `intermediate="true"` == true
 - `intermediate="false"` == false

@EladBezalel @ThomasBurleson Please correct me if I'm wrong here.

DevVersion: @topherfangio I agree with you. That's exactly what `$mdUtil.parseAttributeValue` does.

But notice, attributes like `readonly`, `disabled` or `required` should ignore false values.
That's why you should use in that case
```js
$mdUtil.parseAttributeValue(attr, false)
```

DerekLouie:
Hello All,

I did a minor refactor. 

The appearance of the indeterminate state is now class based instead of attribute selector based. This allowed me to get rid of the "indeterminate-when" attribute and simplified the api a little.

I also fixed the aforementioned comments.

Please Review.

Best,

Derek

src/components/checkbox/checkbox.spec.js
```diff
@@ -246,5 +247,115 @@ describe('mdCheckbox', function() {
       expect(isChecked(checkbox)).toBe(false);
       expect(checkbox.hasClass('ng-invalid')).toBe(true);
     });
+
+    describe('with the indeterminate attribute', function() {
+
+      it('should set indeterminate attr to false by default', function() {
```
topherfangio: I disagree with this test; I believe it should be set to true by default to follow our existing conventions.

src/components/checkbox/checkbox.js
DerekLouie: Ahh so it does seem like `parseAttributeValue` does not exist. BUT `parseAttributeBoolean` does exist. Unfortunately because we can accept a function now as well as a string, we can't use `parseAttributeBoolean` :(

src/components/checkbox/checkbox.spec.js
DerekLouie: Hmm.. that does make a lot of sense :).

Changed, thank you.

src/components/checkbox/checkbox.js
DerekLouie: Added an ng-click to the demo locally, both functions fired correctly, so we are all good on this front!

src/components/checkbox/checkbox.js
```diff
@@ -27,6 +27,16 @@ angular
...
  * @param {string=} aria-label Adds label to checkbox for accessibility.
+ * @param {expression=} indeterminate This determines when the checkbox
+ * should be rendered as 'indeterminate'.
+ * If a truthy expression or no value is passed in the checkbox renders in the indeterminate state.
+ * If falsy expression is passed in it just looks like a normal unchecked checkbox.
+ * To use the other indeterminate-* attributes this attribute must be present.
+ * @param {expression=} indeterminate-checked-when This determines when the
+ * indeterminate checkbox should appear checked. The 'indeterminate' state is mutually
+ * exclusive with the 'checked' and 'unchecked' states.
+ * @param {expression=} indeterminate-click This gets triggered when the
```
ThomasBurleson: I do not understand the feature here. At first glance, this seems to interfere with ng-click. Or is this an expression that should be invoked with the indeterminant state changes?

e.g.  `md-change-indeterminant`

What is the use case here ?

src/components/checkbox/checkbox.js
```diff
@@ -27,6 +27,16 @@ angular
  * @param {string=} ng-change Angular expression to be executed when input changes due to user interaction with the input element.
  * @param {boolean=} md-no-ink Use of attribute indicates use of ripple ink effects
  * @param {string=} aria-label Adds label to checkbox for accessibility.
+ * @param {expression=} indeterminate This determines when the checkbox
+ * should be rendered as 'indeterminate'.
+ * If a truthy expression or no value is passed in the checkbox renders in the indeterminate state.
+ * If falsy expression is passed in it just looks like a normal unchecked checkbox.
+ * To use the other indeterminate-* attributes this attribute must be present.
+ * @param {expression=} indeterminate-checked-when This determines when the
```
ThomasBurleson: This appears to be a confusing API. Perhaps:

`md-indeterminant="<expression>"`

when the `expression` becomes true forces the checkbox to indeterminant state.

src/components/checkbox/checkbox.js
DerekLouie: You are correct, this is not a good attribute.

Previously when I was unsure if a check box could be selected AND indeterminate I thought the click function would be useful for special toggle cases, but in retrospect this was a poor api design decision. Removing it now.

src/components/checkbox/checkbox.js
DerekLouie: I have made some naming changes. Not completely sure if this is what you meant?

Please give the new code a look-over, any notes on how to give the api more clarity are appreciated!

src/components/checkbox/checkbox.spec.js
```diff
@@ -246,5 +247,96 @@ describe('mdCheckbox', function() {
...
+        var element = compileAndLink(
+            '<div>' +
+            '<md-checkbox md-indeterminant></md-checkbox>' +
+            '</div>');
+
+        var checkbox = element.find('md-checkbox');
+
+        pageScope.$apply();
+
+        expect(checkbox.hasClass(INDETERMINATE_CSS)).toBe(true);
```
ErinCoughlan: Is `toHaveClass` copied over from our branch? This would be a lot easier to read as:

`expect(checkbox).toHaveClass(INDETERMINATE_CSS);`

src/components/checkbox/checkbox.spec.js
```diff
@@ -246,5 +247,96 @@ describe('mdCheckbox', function() {
       expect(isChecked(checkbox)).toBe(false);
       expect(checkbox.hasClass('ng-invalid')).toBe(true);
     });
+
+    describe('with the md-indeterminant attribute', function() {
+
+      it('should set md-indeterminant attr to true by default', function() {
+        var element = compileAndLink(
```
ErinCoughlan: Do you need the wrapper `div`? Would `var checkbox = compileAndLink('<md-checkbox ... </md-checkbox>');` work?

src/components/checkbox/checkbox.spec.js
```diff
@@ -246,5 +247,96 @@ describe('mdCheckbox', function() {
...
+            '<div>' +
+            '<md-checkbox md-indeterminant></md-checkbox>' +
+            '</div>');
+
+        var checkbox = element.find('md-checkbox');
+
+        pageScope.$apply();
```
ErinCoughlan: `compileAndLink` calls `pageScope.$apply();`. Do you need another one?

src/components/checkbox/demoBasicUsage/index.html
```diff
@@ -43,6 +43,18 @@
...
         </div>
+        <div flex-xs flex="50">
+          <md-checkbox md-indeterminant
+            aria-label="Checkbox Indeterminate" class="md-primary">
+            Checkbox: Indeterminate
+          </md-checkbox>
+        </div>
+        <div flex-xs flex="50">
+          <md-checkbox md-indeterminant aria-label="Checkbox Disabled Indeterminate"
+            ng-disabled="true" class="md-primary">
```
ErinCoughlan: nit: indent to line up with md-indeterminate or indent 4 spaces instead of 2 to separate attributes from the checkbox content. (Ditto above)

DerekLouie:
Hello All!

Did a significant re-write (thank you @ErinCoughlan for the help). 

Please check it out!

Best,

Derek

src/components/checkbox/checkbox.js
```diff
@@ -27,7 +27,12 @@ angular
  * @param {string=} ng-change Angular expression to be executed when input changes due to user interaction with the input element.
  * @param {boolean=} md-no-ink Use of attribute indicates use of ripple ink effects
  * @param {string=} aria-label Adds label to checkbox for accessibility.
- * Defaults to checkbox's text. If no default text is found, a warning will be logged.
+ *     Defaults to checkbox's text. If no default text is found, a warning will be logged.
+ * @param {expression=} md-indeterminant This determines when the checkbox
```
jelbourn: Spelling: `indeterminate` here and everywhere else

src/components/checkbox/checkbox.js
```diff
@@ -27,7 +27,12 @@ angular
...
+ *     should be rendered as 'indeterminant'.
+ *     If a truthy expression or no value is passed in the checkbox renders in the md-indeterminant state.
+ *     If falsy expression is passed in it just looks like a normal unchecked checkbox.
+ *     To use the other md-indeterminant-* attributes this attribute must be present.
```
jelbourn: Documentation needs to explain the interaction between `indeterminate` and `checked`

src/components/checkbox/checkbox.js
```diff
@@ -89,8 +95,14 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
     });
 
     return function postLink(scope, element, attr, ngModelCtrl) {
+      var isIndeterminant;
       ngModelCtrl = ngModelCtrl || $mdUtil.fakeNgModel();
       $mdTheming(element);
+      if (mdIndeterminantStateEnabled) {
+        // Bootstrap initial state
```
jelbourn: This comment doesn't say anything useful.

src/components/checkbox/checkbox.js
```diff
@@ -171,12 +184,18 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
-        if(ngModelCtrl.$viewValue) {
+        if(ngModelCtrl.$viewValue && !isIndeterminant) {
           element.addClass(CHECKED_CSS);
         } else {
           element.removeClass(CHECKED_CSS);
         }
       }
+
+      function setIndeterminantState() {
+        isIndeterminant = scope.$eval(attr.mdIndeterminant) === false ? false : true;
```
jelbourn: `$scope.$watch` evaluates the expression already, so running `$eval` again is doing extra, unnecessary work.

jelbourn: This assignment is kind of clunky. I'd do
```js
isIndeterminant = indeterminateValue || indeterminateValue === '';
```

src/components/checkbox/checkbox.js
```diff
@@ -171,12 +184,18 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
+        element[0].indeterminate = isIndeterminant ? true : undefined;
```
jelbourn: Why is setting a property to the element necessary, and why does it need to be either `true` or `undefined`?

src/components/checkbox/checkbox.spec.js
```diff
@@ -246,5 +247,45 @@ describe('mdCheckbox', function() {
...
+
+    describe('with the md-indeterminant attribute', function() {
+
+      it('should set md-indeterminant attr to true by default', function() {
+        var checkbox = compileAndLink('<md-checkbox md-indeterminant></md-checkbox>');
+
+        expect(checkbox).toHaveClass(INDETERMINATE_CSS);
+      });
+
+      it('should set its HTMLInputElement.indeterminate to be undefined when not md-indeterminant', function() {
```
DevVersion: I would rather say: `should not set the indeterminate property when md-indeterminate is false.`

src/components/checkbox/demoSelectAll/style.css
```diff
@@ -0,0 +1,13 @@
+legend {
```
jelbourn: Prefix demo styles with `demo`-

src/core/style/mixins.scss
```diff
@@ -251,4 +251,21 @@
     cursor: default;
   }
 
+  &.md-indeterminant ._md-icon{
+    &:after {
+        box-sizing: border-box;
```
jelbourn: Too much indent. You're IDE is likely configured to the wrong indent for `.scss` files.

src/components/checkbox/checkbox.js
DevVersion: In addition to that, I thought `md-indeterminat-*` attribute have been dropped? So maybe remove that line.

src/components/checkbox/checkbox.spec.js
```diff
@@ -246,5 +247,45 @@ describe('mdCheckbox', function() {
...
+
+      it('should set its HTMLInputElement.indeterminate to be "true" when md-indeterminant', function() {
+        var checkbox = compileAndLink('<md-checkbox md-indeterminant></md-checkbox>');
+
+        expect(checkbox[0].indeterminate).toBe(true);
+      });
+
+      it('should be set "md-indeterminant" class according to a passed in function', function() {
+        pageScope.isIndeterminate = function() { return true; } ;
+
```
DevVersion: @DerekLouie Maybe you can also add a test, which changes the `indeterminate` expression on runtime as well?

src/components/checkbox/checkbox.js
DerekLouie: I believe this is to leverage some native browser implementation features. @ErinCoughlan?

src/components/checkbox/checkbox.js
DerekLouie: Hmm so, the $scope.$watch will pass in the new value as the first parameter which will either be - `undefined`, `true`, or `false`.

If it returns `false` we want to set `isIndeterminate` to `false`. 

All other cases we want to set  `isIndeterminate` to `true`.

I cannot seem to find a better way to achieve that, but I am open to suggestions!

src/core/style/mixins.scss
```diff
@@ -251,4 +251,21 @@
     cursor: default;
   }
 
+  &.md-indeterminate ._md-icon{
```
EladBezalel: Why is that in here and not in `checkbox.scss`?

src/core/style/themes.scss
```diff
@@ -39,4 +39,10 @@ html, body {
   &#{$checkedSelector} ._md-icon:after {
     border-color: '{{primary-contrast-0.87}}';
   }
+
+  & .md-indeterminate[disabled] {
```
EladBezalel: Why is that in here and not in `checkbox.scss`?

src/core/style/mixins.scss
ErinCoughlan: The checkbox CSS is used both within the checkbox and the select menu (for multi-select case).

src/core/style/themes.scss
DerekLouie: We created a mixin for the checkbox's css so it could be used in other components. This is where all the css goes now that might be used by multiple components.

DerekLouie:
Hello,

I have addressed the aforementioned comments, please review!

Best,

Derek

test/angular-material-spec.js
```diff
@@ -194,10 +200,71 @@
...
+
+      /**
+       * Asserts that a given element contains a given substring in
+       * its innerHTML property.
+       * Accepts any of:
+       *   {string} - A CSS selector.
+       *   {angular.JQLite} - The result of a jQuery query.
+       *   {Element} - A DOM element.
+       */
+      toContainHtml: function() {
```
DevVersion: Why has that been added?

ErinCoughlan: I'm okay with waiting to add this method until we merge feat/panel. It was added here for completeness and convenience, but it's not being used yet.

ErinCoughlan:
LGTM.

Make sure to change your commit message to be:
feat(checkbox): add indeterminate checkbox support

DerekLouie:
Changed the commit message, please review again at your earliest convenience!

test/angular-material-spec.js
DerekLouie: I believe I these methods will be useful, and am adding all of them now (instead of just the `toHaveClass` method changes) so later down the line we don't get merge conflicts when we merge feat/panel.

src/components/checkbox/checkbox.js
```diff
@@ -171,12 +183,18 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
-        if(ngModelCtrl.$viewValue) {
+        if(ngModelCtrl.$viewValue && !isIndeterminate) {
           element.addClass(CHECKED_CSS);
         } else {
           element.removeClass(CHECKED_CSS);
         }
       }
+
+      function setIndeterminateState(newValue) {
+        isIndeterminate = (newValue === false) ? false : true;
```
jelbourn: isIndeterminate = newValue !== false;

test/angular-material-spec.js
DevVersion: Didn't see actual point, because we rarely change the spec shims. So there won't be any merge conflicts. But it's okay.

src/components/checkbox/checkbox.js
```diff
@@ -171,12 +183,18 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
+        element[0].indeterminate = isIndeterminate ? true : undefined;
```
jelbourn: It's still not clear why you need to set an `indeterminate` property on the DOM element.

src/components/checkbox/checkbox.spec.js
```diff
@@ -246,5 +247,64 @@ describe('mdCheckbox', function() {
...
+      it('should set md-indeterminate attr to true by default', function() {
+        var checkbox = compileAndLink('<md-checkbox md-indeterminate></md-checkbox>');
+
+        expect(checkbox).toHaveClass(INDETERMINATE_CSS);
+      });
+
+      it('should not set the HTMLInputElement.indeterminate property when md-indeterminate is false', function() {
+        var checkbox = compileAndLink('<md-checkbox md-indeterminate="false"></md-checkbox>');
+
+        expect(checkbox[0].indeterminate).toBe(undefined);
```
jelbourn: `toBeUndefined()`

src/components/checkbox/demoSelectAll/style.css
```diff
@@ -0,0 +1,15 @@
...
+  fieldset.standard {
+    border-style: solid;
+    border-width: 1px;
+    height: 100%;
+  }
+
+  .select-all-checkboxes {
+    padding-left: 30px;
+  }
+}
```
jelbourn: Just do these as
```css
.demo-legend { ... }
.demo-fieldset { ... }
.demo-select-all-checkboxes { .. }
```

test/angular-material-spec.js
jelbourn: It's better to not add things until they're used. Plus you'll prevent merge conflicts.

DevVersion: Yep, I agree. These test helpers, should be just merged within the `feat/panel` branch. Partially merging code snippets doesn't make any sense.

ErinCoughlan:
@jelbourn Should there be aria-checked somewhere in here? I didn't see it in the initial code, so is it not helpful?

test/angular-material-spec.js
DerekLouie: Fair point, taking out the non-used additions!

src/components/checkbox/checkbox.js
DerekLouie: Apologies, I miss-understood the spec I was looking at. This is for when we move to the native checkbox. The only benefit of this would be that users can use the `:indeterminate` pseudo selector to find the indeterminate `md-checkbox`es.

@jelbourn I am okay with removing this code, do you care whether or not the directive is query-able using the  `:indeterminate` pseudo selector?

DerekLouie:
Took out some extra code and also added some code to dictate the 'aria-checked' attribute.

Please review.

src/components/checkbox/checkbox.js
```diff
@@ -171,12 +183,20 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
           element.addClass(CHECKED_CSS);
         } else {
           element.removeClass(CHECKED_CSS);
         }
       }
+
+      function setIndeterminateState(newValue) {
+        isIndeterminate = newValue !== false;
+        if (isIndeterminate) {
+          element.attr('aria-checked', 'mixed');
```
ErinCoughlan: aria-checked is never set for checked or unchecked checkboxes either. If you're going to set it to "mixed", you also should set it to true/false where appropriate.

DerekLouie: It actually is set, and accounted for in the [tests](https://github.com/DerekLouie/material/blob/inteterminateCheckbox/src/components/checkbox/checkbox.spec.js#L43). It was a little curious though that the code isn't found in this file..

jelbourn: I believe it's already handled by ngAria

DerekLouie:
Hello All,

I have addressed all the aforementioned comments, please review at your earliest convenience!

Best,

Derek

src/core/style/mixins.scss
DevVersion: @DerekLouie: Nitpicky -  Please space bracket from `._md-icon`

DevVersion:
LGTM now, aside one comment.

DerekLouie:
Apologies, there have been quite a few comments, I will happily address it, which comment?

DerekLouie:
Ahh thank you @DevVersion, fixed the comment.

jelbourn:
LGTM

src/core/style/themes.scss
EladBezalel: Indeterminate isn't used anywhere else, only on progress and this seems not related, 
Also don't you think this class should be private?

src/core/style/mixins.scss
```diff
@@ -251,4 +251,21 @@
     cursor: default;
   }
 
+  &.md-indeterminate ._md-icon {
```
EladBezalel: This is not related here as well, this is very specific for checkbox

src/core/style/themes.scss
topherfangio: @EladBezalel In response to private: I had a discussion with Thomas where we thought that CSS classes representing the state of a component should probably be public as they are likely the ones to be used most.

crisbeto:
LGTM

ThomasBurleson:
@DerekLouie - really nice feature. I LOVE this demo:

screen shot 2016-03-30 at 12 27 27 pm

src/core/style/mixins.scss
DerekLouie: @EladBezalel I could remove the `&`. The rule would still be valid and it would be one class less specific?

PR #7646, opened 2016-03-19
fix(mdSelect): fix flickering of ngMessages on mdSelect opening

### LINKED ISSUE #7636: mdSelect: flickering with ngMessages - regression
mdSelect error message provided by ngMessages flickers when opening the select pop up on Firefox and Chrome on Linux, other browsers weren't tested.

Demo can be found at http://codepen.io/anon/pen/grmWOe.

Steps to reproduce:
1. Open select.
Expected result: "required" message should be not visible before select closes without the selection.
Current result: "required" message is visible in background, when the select is opened.

This issue was already fixed once in https://github.com/angular/material/issues/5879, but those changes were removed in https://github.com/angular/material/commit/dc8f388a8cb2abe1d49455ab28c96e2fa231a5ec. After that, fix for this issue was added in different form in https://github.com/angular/material/commit/8d4e7c2e0595387fd8326ac829c7d1f4895aba10, but I don't think that it works properly.

montgomery1944:
To properly prevent ngMessages from showing up on select opening,
we have to set untouched, in the $apply triggered by ngModel, after ngModel sets touched.
To be close to mdInput behaviour, we set untouched only, when mdSelect was opened.

closes #7636

### DIFF
--- a/src/components/select/select.spec.js
+++ b/src/components/select/select.spec.js
@@ -158,6 +158,23 @@ describe('<md-select>', function() {
     expect($rootScope.myForm.select.$touched).toBe(true);
   }));
 
+  it('should remain untouched during opening', inject(function($compile, $rootScope) {
+    var form = $compile('<form name="myForm">' +
+                        '<md-select name="select" ng-model="val">' +
+                        '<md-option>1</md-option>' +
+                        '</md-select>' +
+                        '</form>')($rootScope);
+    var unwatch = $rootScope.$watch('myForm.select.$touched',
+      function(touched) {
+        expect(touched).toBe(false);
+      });
+    var select = form.find('md-select');
+    openSelect(select);
+    unwatch();
+    closeSelect();
+    expect($rootScope.myForm.select.$touched).toBe(true);
+  }));
+
   it('restores focus to select when the menu is closed', inject(function($document) {
     var select = setupSelect('ng-model="val"').find('md-select');
     openSelect(select);
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -299,19 +299,19 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
             }
           });
 
-        // Wait until postDigest so that we attach after ngModel's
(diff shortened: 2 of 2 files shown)


topherfangio:
@montgomery1944 Thanks! This looks helpful!

Any chance you could provide a test that shows why we need this?

montgomery1944:
@topherfangio, sure, I will write some tests covering this change as soon as possible.

montgomery1944:
@topherfangio, I wrote a test and optimized this solution a little bit. This test will fail without the fix.

src/components/select/select.spec.js
```diff
@@ -153,6 +153,22 @@ describe('<md-select>', function() {
     expect($rootScope.myForm.select.$touched).toBe(true);
   }));
 
+  it('should remain untouched during opening', inject(function($compile, $rootScope) {
+    var form = $compile('<form name="myForm">' +
+                        '<md-select name="select" ng-model="val">' +
+                        '<md-option>1</md-option>' +
+                        '</md-select>' +
+                        '</form>')($rootScope);
+    var unwatch = $rootScope.$watch('myForm.select.$touched',
```
topherfangio: Can you explain why this is in a `$watch` expression instead of just calling `$rootScope.$eval('myForm.select.$touched')` within the test code below?

montgomery1944: Because we want in this test to check, whether mdSelect remains untouched during opening, to prevent ngMessages flicker. The easiest way to do that, is to watch the value of $touched property and assert, that the value is false. Without the assertion in $watch, this test would be the same as the 'should set touched only after closing' test, which was passing previously without the fix.

topherfangio: Gothca; thanks for clarifying.

topherfangio:
@EladBezalel This looks pretty good to me, but I'd love a second pair of eyes.

montgomery1944:
@topherfangio, based on findings from https://github.com/angular/material/pull/7778 I've changed the solution from $evalAsync to stopImmediatePropagation. The latter is slightly better.

PR #7650, opened 2016-03-20
fix(list): dense list font size

### LINKED ISSUE #7552: md-list: .md-dense - wrong font-size
`md-list.md-dense` has wrong font-size set. According to the spec it should be 13px, but it actually goes up to ~21px.

Demo [here](http://codepen.io/anon/pen/JXRyNW).

Note that the official demo has some additional css added (`html: font-size: 62.5%;`) which makes it look good there.

crisbeto:
The dense lists were using rems for the font size, whereas the normal lists were using pixels,
which meant that if the root font size is changed, the two list types would look inconsistent.
This makes them consistent by switching the dense list to using pixels.

@EladBezalel I've made sure that the lists look the same as the ones in master, but could you take a look at this since you added the original dense styling?

Closes #7552.


EladBezalel:
LGTM, needs to be manual tested

ThomasBurleson:
@EladBezalel  - did you finish the manual testing for this PR?

EladBezalel:
LGTM

PR #7651, opened 2016-03-20
fix(list): make secondary items static

### LINKED ISSUE #2759: Long md-list-item text goes through secondary button
If the text is long, it overflows into the secondary button.

capture

### LINKED ISSUE #7500: Text overflow doesn't work on list-item with secondary buttons
Secondary buttons are ignored and text goes over them

http://codepen.io/anon/pen/YqqmXL

DevVersion:
* The secondary items always had an absolute position, which causes issues with overflowing.
* This commit refactors the list to make the secondary items static.
* Fixes multiple secondary padding (as in specs - consistent)
* Fixes Proxy Scan for secondary items container
* Improved tests for the list component (more clear and more safe)

@crisbeto Maybe you can also review

Fixes #7500. Fixes #2759.

### DIFF
--- a/src/components/list/list.js
+++ b/src/components/list/list.js
@@ -88,10 +88,12 @@ function mdListItemDirective($mdAria, $mdConstant, $mdUtil, $timeout) {
     restrict: 'E',
     controller: 'MdListController',
     compile: function(tEl, tAttrs) {
+
       // Check for proxy controls (no ng-click on parent, and a control inside)
       var secondaryItems = tEl[0].querySelectorAll('.md-secondary');
       var hasProxiedElement;
       var proxyElement;
+      var itemContainer = tEl;
 
       tEl[0].setAttribute('role', 'listitem');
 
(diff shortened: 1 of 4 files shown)


src/components/list/list.js
```diff
@@ -190,20 +182,12 @@ function mdListItemDirective($mdAria, $mdConstant, $mdUtil, $timeout) {
...
-          // orderd correctly in the list-item
-          if (hasSecondaryItemsWrapper) {
-            secondaryItem.classList.remove('md-secondary');
-          }
-          tEl.addClass('md-with-secondary');
-          container.append(secondaryItem);
-        }
+        // When using multiple secondary items we need to remove their secondary class to be
+        // orderd correctly in the list-item
+        secondaryItem.classList.remove('md-secondary');
```
crisbeto: Why not wrap this as a jqLite element and use `removeClass` here?

DevVersion: There is no need, because, when iterating through a jqLite array, we retrieve always plain DOM elements.
So wrapping the element for one expression, is not needed in my opinion. In addition to that, I just transferred this statement from the previous child block.

crisbeto: Sounds good, we should just bear in mind that `classList` can be a little inconsistent in IE. The way you're using it here should be fine though.

crisbeto:
LGTM :+1:

src/components/list/list.js
DevVersion: Just rechecked that code snippet, because you mentioned me on the IE10/9 issues with `classList` and figured out, that there's no need to remove that class anymore, because we are now always using a secondary container with static children. :+1:

EladBezalel:
LGTM, with that said, let this be one of the last PRs about list and focus on the new implementation,
@ThomasBurleson

DevVersion:
I already wrote with Elad about the implementation, and I'm looking forward with helping to plan the new implementation.

Just want to summarize a few things here again:
- The secondary items are now static (this was a bug since many months)
  I compared the new secondary items with the old items and focused on the specs.
  That means, padding etc is all the same.

- The Proxy scan wasn't incorrect in the previous version. I've updated the scan logic here (no breaking change) and it works very well. (tested that with a lot user applications - codepens etc.)

- I've improved the tests, to be more clear
  That means, now the tests are more accurate and easier to read.

So I can say, I tested these things thoroughly, which I've listed before.

The only thing about I'am a bit unsure, is the `flexFiller`, which moves the secondary items to the end of the list.

PR #7654, opened 2016-03-20
fix(mdInput): fix missing error message with md-auto-hide option

### LINKED ISSUE #7653: mdInput: error message is not shown with md-auto-hide option
When mdInput has more than one error message in ngMessages and the input remains invalid during the change of message, the message is not shown. This is caused by an issue in ngMessageAnimation which does not fire showMessage, when md-auto-hide option in turned on. I think that ngMessageAnimation should show the message when the mdInput has md-input-invalid class.

Demo can be found at http://codepen.io/anon/pen/zqZZex.
To reproduce this issue:
1. Type "a" in input
2. Leave the input.
3. Enter the input.
4. Delete "a".
Expected result: "required" message should be visible.
Current result: no message is shown.

Original implementation was added in https://github.com/angular/material/commit/cf7f21aa6c633d9f0bd22013e9a79f99b82b16af.

montgomery1944:
In ngMessageAnimation if we have the md-auto-hide class and input is not invalid, we have to skip animation.
Otherwise, we have to fire animation to show message, because md-input-invalid animation does not fire when the input remains invalid during the change of message.

closes #7653

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -768,9 +768,12 @@ function ngMessageAnimation($animateCss) {
   return {
     enter: function(element, done) {
       var messages = getMessagesElement(element);
+      var input = getInputElement(element);
 
-      // If we have the md-auto-hide class, the md-input-invalid animation will fire, so we can skip
-      if (messages.hasClass('md-auto-hide')) {
+      // If we have the md-auto-hide class and input is not invalid, we have to skip animation.
+      // Otherwise, we have to fire animation to show message, because md-input-invalid animation does not fire
+      // when the input remains invalid during the change of message.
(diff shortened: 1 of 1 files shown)


DevVersion:
Thanks @montgomery1944 for your contribution. 

@topherfangio is already working on a fix for that issue, which includes a huge amount of tests.
That's why we need to reject these type of PR's - a PR from Topher is coming soon.

Quote from [Here](https://github.com/angular/material/pull/7583#issuecomment-197018180)
> Apologies, but I've been working on this issue already. We're trying to get some really good tests written so that we can make sure we fix these issues and don't regress in the future. I'm actually meeting with someone in about 5 minutes to hopefully finalize these tests.

Issue #6342, opened 2015-12-16
md-tabs:md-enable-disconnect doesn't work


Hotell:
`md-enable-disconnect` doesn't render tab content lazy via ngIf.

all content is rendered on boot and not destroyed on tab change.

demo includes a component which is inside `md-tab` and should console.log on scope `$destroy`, which is unfortunately never called.

demo:
http://codepen.io/Hotell/pen/QyyKjY?editors=101

krasevych:
I have the same problem(

PR #7665, opened 2016-03-21
fix(md-tabs): fix mdEnableDisconnect flag by inverting some booleans

benelliott:
The md-enable-disconnect flag, designed to allow tab content scope to be destroyed when the tab is not displayed, was not working because the logic for hiding the tab content in ngIf was incorrect.

Fixes issue #6342

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -395,7 +395,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
           isActive:     function () { return this.getIndex() === ctrl.selectedIndex; },
           isLeft:       function () { return this.getIndex() < ctrl.selectedIndex; },
           isRight:      function () { return this.getIndex() > ctrl.selectedIndex; },
-          shouldRender: function () { return !ctrl.noDisconnect || this.isActive(); },
+          shouldRender: function () { return ctrl.noDisconnect || this.isActive(); },
           hasFocus:     function () {
             return !ctrl.lastClick
                 && ctrl.hasFocus && this.getIndex() === ctrl.focusIndex;
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -30,7 +30,6 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
   defineBooleanAttribute('dynamicHeight', handleDynamicHeight);
   defineBooleanAttribute('noPagination');
   defineBooleanAttribute('swipeContent');
-  defineBooleanAttribute('noDisconnect');
   defineBooleanAttribute('autoselect');
   defineBooleanAttribute('noSelectClick');
   defineBooleanAttribute('centerTabs', handleCenterTabs, false);
@@ -395,7 +394,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
           isActive:     function () { return this.getIndex() === ctrl.selectedIndex; },
           isLeft:       function () { return this.getIndex() < ctrl.selectedIndex; },
           isRight:      function () { return this.getIndex() > ctrl.selectedIndex; },
-          shouldRender: function () { return ctrl.noDisconnect || this.isActive(); },
+          shouldRender: function () { return this.isActive(); },
           hasFocus:     function () {
             return !ctrl.lastClick
                 && ctrl.hasFocus && this.getIndex() === ctrl.focusIndex;
--- a/src/components/tabs/js/tabsDirective.js
+++ b/src/components/tabs/js/tabsDirective.js
@@ -191,7 +191,7 @@ function MdTabs () {
                 'md-tabs-template="::tab.template" ' +
                 'md-connected-if="tab.isActive()" ' +
                 'md-scope="::tab.parent" ' +
-                'ng-if="$mdTabsCtrl.enableDisconnect || tab.shouldRender()"></div> ' +
+                'ng-if="!$mdTabsCtrl.enableDisconnect || tab.shouldRender()"></div> ' +
           '</md-tab-content> ' +
         '</md-tabs-content-wrapper>';
     },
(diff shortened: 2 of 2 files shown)


src/components/tabs/js/tabsController.js
```diff
@@ -395,7 +395,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
           isActive:     function () { return this.getIndex() === ctrl.selectedIndex; },
           isLeft:       function () { return this.getIndex() < ctrl.selectedIndex; },
           isRight:      function () { return this.getIndex() > ctrl.selectedIndex; },
-          shouldRender: function () { return !ctrl.noDisconnect || this.isActive(); },
+          shouldRender: function () { return ctrl.noDisconnect || this.isActive(); },
```
DevVersion: I'm wondering, why there is another attribute `md-no-disconnect`, which disabled the `disconnect`functionality.

So that means, that we have two attributes (which are both reversed)
- `md-enable-disconnect` (Documented)
- `md-no-disconnect` (Not documented) 

@robertmesserle Do you know anything about that?

robertmesserle: I am not up to date on this component, but md-no -disconnect was used in the past when the scope was disconnected by default.

benelliott:
If you like, I can amend the pull request after running through the code and deleting any mention of mdNoDisconnect.

DevVersion:
@benelliott Yes I think it would make much sense to remove that.

So the that means, that the tab `shouldRender`:
- When it's active

and it should be only added to the DOM, when
- `shouldRender()` is true (above)
- `md-enable-disconnect` is set to false

In my opinion that should be the correct behavior. And it won't break anything, because `md-no-disconnect` isn't actually documented. 

So notice: It isn't changing anything in the API, it is just improving / fixing the attribute use.

cc. @robertmesserle

benelliott:
Okay cool, I'll get to it tomorrow!

benelliott:
Would you mind clarifying the difference between the tab rendering and the tab being added to the DOM? From what I can see in the current code, `tab.shouldRender()` is only read in the `ng-if` on the `div` inside `md-tab-content` - so at the moment whether or not the tab should render seems to be equivalent to whether or not it should be added to the DOM.

Additionally, as of your comment, `tab.shouldRender()` is just returning the value of `tab.isActive()`.

This actually suggests that there is no need for `tab.shouldRender()`, and the `ng-if` expression can become `ng-if="!$mdTabsCtrl.enableDisconnect || tab.isActive()"` (assuming that I interpreted your specification correctly). I guess removing `shouldRender` would change the API though which isn't ideal.

DevVersion:
@benelliott Thanks for looking into it.

Sure, I will try to explain my thoughts a bit more accurate. 

- We should drop the `ctrl.noDisconnect`

You are right, the `shouldRender` function should just return `tab.isActive()` 
```js
shouldRender: function () { return this.isActive(); },
```

And as you said, removing the `shouldRender` function (because it's unnecessary) won't be a good idea - I agree with you. Let's keep the `shouldRender` function and just check in the `ng-if`
```html
<x ng-if="!$mdTabsCtrl.enableDisconnect || tab.shouldRender()">
```

benelliott:
@DevVersion I've pushed the changes, please take a look when you get a chance.

DevVersion:
@benelliott Just tested it out and it seems, that the logic is valid. 

But the scopes will be never destroyed, the tab will just retrieve a new scope. 
So I'm a bit wondering, what we should do with the scopes when `md-enable-disconnect` is enabled.

I mean, #6342 says, that the scopes should be destroyed, but the attribute documentation only says that the scopes will be disconnected (and that seems to work, with that PR)

So what are your thoughts?

benelliott:
@DevVersion this is actually something I only realised myself today - the initial issue is based on a misunderstanding of what mdEnableDisconnect is supposed to do (i.e. that the scope will be destroyed rather than just disconnected), but it also seemed the case that the intended behaviour was not being fulfilled.

Looking at the comments in the code, it seems at some point it was decided that disconnecting the scope offered a nice middle-ground between destroying the scope completely and not doing anything at all, since it prevents the scope's watches and events from firing without forcing it to be recreated the next time it is in view. This makes sense as (I would assume) most of the overhead of keeping an inactive tab's scope around comes from the CPU work of processing its watches and events on every digest cycle, rather than the memory its scope objects are consuming.

benelliott:
Hi @DevVersion, do you have any more thoughts on this?

DevVersion:
@benelliott I agree, I think we should keep it as it now is, in your PR. Maybe you can update the `mdEnableDisconnect` attribute description, to explain more how it works.

benelliott:
@DevVersion done!

src/components/tabs/js/tabsDirective.js
```diff
@@ -61,7 +61,7 @@
  * @param {boolean=} md-center-tabs When enabled, tabs will be centered provided there is no need for pagination
  * @param {boolean=} md-no-pagination When enabled, pagination will remain off
  * @param {boolean=} md-swipe-content When enabled, swipe gestures will be enabled for the content area to jump between tabs
- * @param {boolean=} md-enable-disconnect When enabled, scopes will be disconnected for tabs that are not being displayed.  This provides a performance boost, but may also cause unexpected issues and is not recommended for most users.
+ * @param {boolean=} md-enable-disconnect When enabled, scopes will be disconnected for tabs that are not being displayed. When a scope is disconnected it is removed from the scope hierarchy so that its events and watches are not fired, but its state is not destroyed. This can result in a performance boost but may have unexpected behaviour.
```
DevVersion: Sorry, for being so nitpicky :smile:, what do you think about also mentioning, that this causes a new scope for every new tab. That means, if I change tab, then it will always create a new scope, instead of using the old.

benelliott:
@DevVersion If that's the case, is there any advantage to disconnecting the scopes as opposed to destroying them?

ThomasBurleson:
ngMaterial2 will have a much better solution for non-active tab content.

ThomasBurleson:

ThomasBurleson:

Issue #7666, opened 2016-03-21
mdSlider: md-reverse feature request


soooooot:
```html
<md-slider md-reverse min="0" max="100"></md-slider>
```
it makes min value to top/right, and max value to bottom/left 
(regularly  min value at bottom/left, max value at top/right)

I have implemented it in [CodePen](http://codepen.io/anon/pen/bpqNYv). 
I don't know is anybody else need the feature or not, anyway, I will make a PR soon.

topherfangio:
@soooooot What is the purpose of the reverse mode? Is it to support RTL languages? If so, I think it would be better if the component supported the `[dir="rtl"]` attribute.

@EladBezalel We had a similar discussion recently correct? Any updates/thoughts?

soooooot:
@topherfangio my case is not a RTL case but a pagination slider, which should be reversed.

EladBezalel:
@topherfangio this is just to make the slider reversed, legit request

ThomasBurleson:
This issue is closed as part of our ‘Surge Focus on Material 2' efforts.
      For details, see our forum posting @ http://bit.ly/1UhZyWs.

PR #7667, opened 2016-03-21
feat(slider): md-reverse

soooooot:
make min value to top/right, max value to bottom/left, details on #7666

### DIFF
--- a/src/components/slider/slider.js
+++ b/src/components/slider/slider.js
@@ -116,8 +116,14 @@ function SliderContainerDirective() {
  * <md-slider md-discrete ng-model="myDiscreteValue" step="10" min="10" max="130">
  * </md-slider>
  * </hljs>
+ * <h4>Invert Mode</h4>
+ * <hljs lang="html">
+ * <md-slider md-invert ng-model="myValue" step="10" min="10" max="130">
+ * </md-slider>
+ * </hljs>
  *
  * @param {boolean=} md-discrete Whether to enable discrete mode.
+ * @param {boolean=} md-invert Whether to enable invert mode.
  * @param {number=} step The distance between values the user is allowed to pick. Default 1.
  * @param {number=} min The minimum value the user is allowed to pick. Default 0.
  * @param {number=} max The maximum value the user is allowed to pick. Default 100.
@@ -202,6 +208,7 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
     var DEFAULT_ROUND = 3;
     var vertical = angular.isDefined(attr.mdVertical);
     var discrete = angular.isDefined(attr.mdDiscrete);
+    var invert = angular.isDefined(attr.mdInvert);
     angular.isDefined(attr.min) ? attr.$observe('min', updateMin) : updateMin(0);
     angular.isDefined(attr.max) ? attr.$observe('max', updateMax) : updateMax(100);
     angular.isDefined(attr.step)? attr.$observe('step', updateStep) : updateStep(1);
@@ -356,6 +363,7 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
       } else if (vertical ? ev.keyCode === $mdConstant.KEY_CODE.UP_ARROW : ev.keyCode === $mdConstant.KEY_CODE.RIGHT_ARROW) {
         changeAmount = step;
       }
+      changeAmount = invert ? -changeAmount : changeAmount;
       if (changeAmount) {
         if (ev.metaKey || ev.ctrlKey || ev.altKey) {
           changeAmount *= 4;
@@ -404,7 +412,7 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
 
       ngModelCtrl.$viewValue = minMaxValidator(ngModelCtrl.$viewValue);
 
-      var percent = (ngModelCtrl.$viewValue - min) / (max - min);
+      var percent = valueToPercent(ngModelCtrl.$viewValue);
       scope.modelValue = ngModelCtrl.$viewValue;
       element.attr('aria-valuenow', ngModelCtrl.$viewValue);
       setSliderPercent(percent);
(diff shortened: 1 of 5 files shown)


DevVersion:
Very well implemented. LGTM.

src/components/slider/slider.js
```diff
@@ -558,11 +560,13 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
...
      */
     function percentToValue( percent ) {
-      return (min + percent * (max - min));
+      var adjustedPercent = reverse ? (1 - percent) : percent;
+      return (min + adjustedPercent * (max - min));
     }
 
     function valueToPercent( val ) {
-      return (val - min)/(max - min);
+      var percent = (val - min)/(max - min);
```
DevVersion: Just one little thing I saw (nitpicky): The arithmetic operator should be spaced, just to be consistent.

EladBezalel:
@soooooot thanks for this, please next time open an issue just to make order,
This seems good, please fix @DevVersion comment, squash into 1 commit and follow our commit guideline.

thanks again, LGTM.

@DevVersion do you mind manual test it?

soooooot:
@EladBezalel I fired an issue here issue #7666, next time I will mark it clear if I have a PR to create.

There's a [CodePen](http://codepen.io/anon/pen/bpqNYv) link posted in issue, maybe it can help you guys to manual test it.

sorry for my limited experience on operating PR. 
Should I  rebase the 2 commit  into one and  push **--force**,  the PR can automatically update it or not?
Or I need to make a new branch and create a new PR?

DevVersion:
@EladBezalel Just tested it locally and it works like a charm.

topherfangio:
I'd like to have a bit more discussion about this before we merge it. I'm not opposed to adding this functionality, but I think the terminology "reverse" is a bit confusing and would like to discuss some other options.

When I first saw this PR, I assumed it meant that the entire component would be reversed as if it were being used in an RTL setting. However, visually, the slider works identically to the existing slider, it's just that it counts downward with the value instead of upward.

So, rather than `md-reverse`, I'm thinking it might be more appropriate to call it `md-invert-value` or `md-count-down`.

@EladBezalel @DevVersion Thoughts?

EladBezalel:
Oh, isn't that as RTL?

DevVersion:
@topherfangio Yes, `md-invert-value` sounds for me very good. But what about `md-invert` this sounds even better for me and is not confusingly

topherfangio:
@EladBezalel In my mind, RTL would start the slider at the right all grey, and the accent color (pink in our demo) would grow from the right to the left. Currently, it's still left-to-right even in reverse:

screen shot 2016-03-24 at 2 59 51 pm

That said, I unfortunately don't know much about how visual elements are represented in RTL cultures.

@DevVersion I worry that without the `-value`, people may think it affects the visual display of the element as well and thus has a similar problem to `md-reverse` (IMO :smile:).

soooooot:
`md-invert-value` makes more sense to me.

I have no any experience  on RTL too. I describe my use-case FYI. 

what I wanna do is **a vertical, value increasing top-to-down, pagination slider**. unfortunately the current md-slider does not support it. For such a small code change, I have to copy all the md-slider js code and paste it to another new `md-reverse-slider` (maybe `md-invert-value-slider` is better, :laughing: ). And as you know, the *css* is generated and inject in run-time, and the css is all base on `md-slider` tagname, it hard for me to put my new `md-reverse-slider` directive into the run-time css injection list. Also it will takes time to copy all the css rules. So I did a little bit hack stuff in my `compile` function to change my `md-reverse-slider` tagname to `md-slider` to get the styles. and last to override the `_md-track-fill`(the accent pink part) color to transparent.

that's my story.

topherfangio:
For reference, this is how I assume the RTL slider would work: http://demos.telerik.com/kendo-ui/slider/right-to-left-support

If you can update this PR to use `md-invert-value`, I think we can run it by Thomas and see if we can get it merged :smile:

soooooot:
it seems like `md-invert-value` will be temporarily stuff, sooner or later If we implemented `rtl`, it will be deprecated or obsolete. 

I made a new [CodePen](http://codepen.io/soooooot/pen/oxwYLx) to make it visual reversed( the accent pink part) just like the demo @topherfangio posted. 

And I also updated the PR, so you guys can check the changes or you can check on [my personal branch commit](https://github.com/soooooot/material/commit/de4fc47470d352468ccd73b86d971a839841acb0) to get the visual part changes.

I didn't change the name yet, since we should rename it depends on the behaviour right? `rtl`, `invert` or `reverse`.

topherfangio:
@soooooot If we implement real RTL support, could you use it on just that single slider for your purposes? If so, I think that would definitely be a better solution.

soooooot:
@topherfangio yep, I think RTL single slider can implement a pagination slider.
 
The PR I updated yesterday did implement the RTL feature in my opinion, or I missed something?

topherfangio:
@soooooot Sorry for the delay (was on vacation). True RTL support just looks at the `dir="rtl"` attribute on a parent element (usually the body) rather than using a `md-reverse` or `md-invert-value` attribute on the actual component.

For instance, it should work if you did something like this:

```html
<div dir="rtl">
    <md-slider min="0" max="255"></md-slider>
</div>
```

@EladBezalel Does this sound correct to you?

EladBezalel:
@topherfangio it won't work, the dir attribute is only taking place with our code if it's being applied on body or html..

soooooot:
@topherfangio After some days to think about the issue about, I found that even in `rtl` mode, `md-invert` is also useful to make a `regular ltr` slider in some cases. So, can I just update the PR to rename `md-reverse` to `md-invert`. (I perfer to the one without `-value` since now it will be inverted both in value and visual display).  Can we **end up the PR with the `md-invert` feature only**? if we need more discussion on the `rtl` mode, it's better to fire an new issue.  Although the `md-invert` is related to `rtl` mode, but we need it in `rtl` mode too, we cannot merge the `md-invert` to `rtl`.

@EladBezalel I have a rough idea to make 'md-invert' to work with `rtl` mode, we should **invert** it depends on `rtl XOR md-invert`....(I just make a start if you don't agree to discuss the `rtl` in another issue..)

I will rename it as soon as I see your reply with a confirmed name. `md-invert`, `md-invert-value` or whatever.

topherfangio:
@soooooot I think `md-invert` is good.

@EladBezalel Are you sure about RTL on body only? When I just tested it, all I did was add `dir="rtl"` to the demo wrapper (from within the Chrome inspector) and the Switch component swapped into RTL mode. Based on https://github.com/angular/material/blob/master/src/core/style/mixins.scss#L111, it looks like all of the RTL mixins use the attribute selector only and don't specify the body.

EladBezalel:
@topherfangio so i guess it's fixed, rtl mixin was once only on html and body.. goo for us! :D

ThomasBurleson:

PR #7671, opened 2016-03-21
Select All feature for Multi Select component.

praskarnam:
The multi select component of angular material does not have the below

- Select All
- When there are more than 1 selected then it keeps adding all of them to the selected label hence growing the component in size. Missing a label saying XX +n more as selected.

This commit adds the above functionality to the select component.

The below code shows how a md select can be written the feature of check all

`
      <md-input-container style="width:200px;">
        <label>State</label>
        <md-select ng-model="userState1" multiple id="hello" select-all>
          <md-option ng-repeat="state in states" value="{{state.abbrev}}" select-all>
            {{state.abbrev}}
          </md-option>
        </md-select>
      </md-input-container>
`

If you notice there is a attribute **select-all** which is added to the md-select which would add the check all option. Further you can see that the same attribute is added to the md-option to just display the check box for all the options. Adding to the md-option is optional where are adding to the md-select is required. The output would be as below

selectall-1

selectall-2

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -118,7 +118,9 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
     restrict: 'E',
     require: ['^?mdInputContainer', 'mdSelect', 'ngModel', '?^form'],
     compile: compile,
-    controller: function() {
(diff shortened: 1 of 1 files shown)


praskarnam:
I signed it!

Issue #7383, opened 2016-03-02
mdSelect: (feat) Add md-select-header as a template to the existing select component.


ErinCoughlan:
This is a step to supporting search/filtering in the select menu.

Template should be added to the top of the existing select to add customizable header section (example use: search box).

API:
```
<md-input-container>
  <md-select
      ng-model="selectedOption"
      placeholder="Select an option"
      selected-text=”{{ myTextFn() }}”
      multiple>
    <md-select-header>
      <input ng-model=”searchTerm” type=”text” class=”searchbox”>
    </md-select-header>
    <md-option ng-repeat="opt in ctrl.filteredOptions" ng-value="opt.value">
      {{ opt.text }}
    </md-option>
  </md-select>
</md-input-container>
```

PR #7674, opened 2016-03-21
feat(select): changed md-select to have select all feature

praskarnam:
@gmoothart @ErinCoughlan @jelbourn @ThomasBurleson , Please review.

The multi select component of angular material does not have the below

- Select All
- When there are more than 1 selected then it keeps adding all of them to the selected label hence growing the component in size. Missing a label saying XX +n more as selected.

This commit adds the above functionality to the select component.

The below code shows how a md select can be written the feature of check all

```html
      <md-input-container style="width:200px;">
        <label>State</label>
        <md-select ng-model="userState1" multiple id="hello" select-all>
          <md-option ng-repeat="state in states" value="{{state.abbrev}}" select-all>
            {{state.abbrev}}
          </md-option>
        </md-select>
      </md-input-container>
```

If you notice there is a attribute **select-all** which is added to the md-select which would add the check all option. Further you can see that the same attribute is added to the md-option to just display the check box for all the options. Adding to the md-option is optional where are adding to the md-select is required. The output would be as below

selectall-1

selectall-2

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -118,7 +118,9 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
     restrict: 'E',
     require: ['^?mdInputContainer', 'mdSelect', 'ngModel', '?^form'],
     compile: compile,
-    controller: function() {
+    controller: function($scope) {
+      $scope.checkAll = false; // Select all binding
+      $scope.checkClass='';  // toggle class for the check box     
     } // empty placeholder controller to be initialized in link
   };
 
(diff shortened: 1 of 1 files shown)


ErinCoughlan:
@praskarnam This will be possible with the new changes to select currently in progress.

Adding checkboxes to each item in the multiple select case is PR #7455, which was already merged to master.

Allowing custom selected text (like "AL [+49 more]") is PR #7721, which also allows custom text like "Select your state" when the select is closed.

Adding custom header information, like a searchbar or select-all/none functionality is Issue #7383, which is in progress. Material UX has not specified how "select all/none" should appear, so that adds more flexibility for devs to customize it to their apps, for instance by adding a "Check All" checkbox above the item content, like you did here.

PR #7675, opened 2016-03-21
feat(datepicker): md-hide-button, md-readonly, md-floating-label

simonepri:
Hello,
since the datepicker directive doesn't allow the developer to customize the template, I added some parameters that allow to better adapt the module to different situations.
These are the parameters:
```
    @param {boolean=} md-hide-button Whether to hide the datepicker button or not.
    @param {boolean=} md-readonly Whether the datepicker is readonly (The datapicker will be opened on focus automatically).
    @param {String=} md-floating-label This will add a floating label to datepicker
```

demo

I also refactored the htmls structure of the module in order to make it equal to other input directives.
The input field of the datepicker had different size than the other fields, worsening the UX. Now it have the same look of the other modules.

I hope It was helpful.
Simone

### DIFF
--- a/src/components/datepicker/datePicker-theme.scss
+++ b/src/components/datepicker/datePicker-theme.scss
@@ -6,35 +6,51 @@ md-datepicker.md-THEME_NAME-theme {
 
 .md-THEME_NAME-theme {
 
-  .md-datepicker-input {
-    @include input-placeholder-color('\'{{foreground-3}}\'');
-    color: '{{foreground-1}}';
-    background: '{{background-color}}';
-  }
-
-  .md-datepicker-input-container {
-    border-bottom-color: '{{background-300}}';
-
-    &.md-datepicker-focused {
-      border-bottom-color: '{{primary-500}}'
+//  .md-datepicker-input {
+//    @include input-placeholder-color('\'{{foreground-3}}\'');
+//    color: '{{foreground-1}}';
+//    background: '{{background-color}}';
+//  }
+//
+//  .md-datepicker-input-container {
+//    border-bottom-color: '{{background-300}}';
+//
+//    &.md-datepicker-focused {
+//      border-bottom-color: '{{primary-500}}'
+//    }
+//
+//    &.md-datepicker-invalid {
+//      border-bottom-color: '{{warn-A700}}';
+//    }
+//  }
+  md-datepicker {
+    md-input-container {
+      &.md-datepicker-invalid {
+        .md-input {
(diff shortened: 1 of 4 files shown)


jaredklewis:
:+1:

simonepri:
I know that the way i've done it, is pretty dirty.
The use of the already styled md-input it seemed to me the less messy way to do it.

crisbeto:
@simonepri I'll close this PR, because:

1. It introduces some APIs that are either unnecessary (e.g. the `md-hide-button` easily be done by the developer with CSS) or redundant (e.g. `readonly` which behaves in a similar way to `disabled` which is already supported).

2. This doesn't follow the pattern that other components use to integrate with `md-input-container` and floating labels. 

We have another PR that does this over at https://github.com/angular/material/pull/8083 that does the same as this one, but without such drastic changes to the component structure and styles.

simonepri:
@crisbeto 
I agree with points 2 and 3.
Let's talk about the point 1.
'disabled' means that the user isn't able to select the date at all.
'readonly' means that the user is able to select the date from the datepicker while is not able to select it by typing directly in the input.

crisbeto:
You're right about 1, but the current datepicker doesn't prevent you from selecting text when it's disabled either.

simonepri:
https://material.angularjs.org/latest/demo/datepicker
I'm not able to select the text in the disabled datepicker, what you are talking about?

crisbeto:
I can do it by double clicking, but this might depend on the browser:

asd

simonepri:
However for "select" i was meaning "pick the date", not "select the text". I'm sorry for the ambiguity.
So is not a problem if i'm able to select the text while it's disabled, the point is that i want to be able to pick a date while keeping the input not manually modificable (readonly)
aa8xpcyz0t
As you can see i'm not able to edit the date with the keyboard but i'm able to do it with the picker.

topherfangio:
@simonepri I don't believe that `readonly` should allow you to alter/pick the date. The functionality you describe sounds more like a `md-force-select` or `md-no-manual` option which we don't currently support. Honestly, I'm not sure what the value of forcing them to select it is. This makes it very difficult for keyboard users (i.e. mostly accessibility-related) to alter the date as they would have to use the keyboard to navigate through the entire list.

ThomasBurleson:
This issue is closed. I am locking the conversation. Please post further discussions in the [Angular Material Forum](https://groups.google.com/forum/#!forum/ngmaterial)

PR #7677, opened 2016-03-21
fix(sidenav): fix incorrect disable-backdrop tests

DevVersion:
The tests for the `md-disable-backdrop` attribute are currently really useless, because the sidenav is actually never opening.

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -228,6 +228,7 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
    */
   function postLink(scope, element, attr, sidenavCtrl) {
     var lastParentOverFlow;
+    var backdrop;
     var triggeringElement = null;
     var promise = $q.when(true);
     var isLockedOpenParsed = $parse(attr.mdIsLockedOpen);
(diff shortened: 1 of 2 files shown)


src/components/sidenav/sidenav.js
```diff
@@ -240,7 +241,11 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
         $mdMedia: $mdMedia
       });
     };
-    var backdrop = !angular.isDefined(attr.mdDisableBackdrop) ? $mdUtil.createBackdrop(scope, "_md-sidenav-backdrop md-opaque ng-enter") : undefined;
+
+    // Only create the backdrop if the backdrop isn't disabled.
+    if (!$mdUtil.parseAttributeBoolean(attr.mdDisableBackdrop)) {
```
ThomasBurleson: *  Why parse for boolean value, the `md-disable-backdrop` implies a TRUE. 
*  What need do we have to parse the expression value?

src/components/sidenav/sidenav.spec.js
```diff
@@ -52,18 +52,18 @@ describe('mdSidenav', function() {
...
 
       var backdrop = el.parent().find('md-backdrop');
-      expect(backdrop.length).toBe(0);
+      expect(backdrop.length).toBe(1);
     }));
 
-    it('should show no backdrop if disabled', inject(function($rootScope, $material, $timeout) {
-      var el = setup('md-disable-backdrop');
+    it('should not show a backdrop if md-disable-backdrop is set to true', inject(function($rootScope, $material) {
+      var el = setup('md-is-open="show" md-disable-backdrop="true"');
```
ThomasBurleson: I do not like this requirement `md-disable-backdrop="true"` seem redundant.
`md-disable-backdrop` seems sufficient.

src/components/sidenav/sidenav.js
DevVersion: @ThomasBurleson You are right, but I thought parsing for a boolean value is more consistent and provides a little advantage. This allows the use a interpolated boolean as a value for the attribute.

But I agree with you, negated values doesn't make any sense.

src/components/sidenav/sidenav.spec.js
DevVersion: Yes, `md-disable-backdrop` is also working. I just used a value in the test, to clarify the current use case.

DevVersion:
Continuing with #7790

PR #7686, opened 2016-03-21
fix(theme): remove default background theme (grey) `1000` value

### LINKED ISSUE #7600: theming dark default background md-hue-3 is an not a valid hue
https://github.com/angular/material/blob/a653122cadf5feba15ca6da8fb876098997892dc/src/core/services/theming/theming.js#L112

Value is `'1000'` but is not listed in
```
var VALID_HUE_VALUES = [
  '50', '100', '200', '300', '400', '500', '600',
  '700', '800', '900', 'A100', 'A200', 'A400', 'A700'
];
```

EladBezalel:
- Other palettes doesn't have `1000` therefor when registering a new background palette 1000 couldn't be found

fixes #7600


src/core/services/theming/theme.palette.js
```diff
@@ -330,9 +330,8 @@ angular.module('material.core.theming.palette', [])
     '700': '#616161',
     '800': '#424242',
     '900': '#212121',
-    '1000': '#000000',
     'A100': '#ffffff',
-    'A200': '#eeeeee',
+    'A200': '#000000',
```
ThomasBurleson: Is this a deviation from the Material Design color palette ?

EladBezalel: grey doesn't have accent colors, we use it to host `white`, `black` and other colors from the spec

PR #7689, opened 2016-03-21
fix(card): add `color` property to `md-card.md-THEME_NAME-theme` style

### LINKED ISSUE #7687: `md-card.md-THEME_NAME-theme` does not set `color` property
```
md-card.md-THEME_NAME-theme {
  background-color: '{{background-hue-1}}';
  border-radius: $card-border-radius;
```

Should define:
```
md-card.md-THEME_NAME-theme {
  color: '{{foreground-1}}';
  background-color: '{{background-hue-1}}';
  border-radius: $card-border-radius;
```

ilovett:
Depends on #7693
Depends on #7686

Closes #7687

### DIFF
--- a/src/components/card/demoBasicUsage/index.html
+++ b/src/components/card/demoBasicUsage/index.html
@@ -1,7 +1,7 @@
 <div ng-controller="AppCtrl" ng-cloak >
   <md-content class="md-padding" layout-xs="column" layout="row">
     <div flex-xs flex-gt-xs="50" layout="column">
-      <md-card>
+      <md-card md-theme="{{ showDarkTheme ? 'dark-grey' : 'default' }}" md-theme-watch>
         <md-card-title>
           <md-card-title-text>
             <span class="md-headline">Card with image</span>
@@ -16,10 +16,10 @@
           <md-button>Action 2</md-button>
         </md-card-actions>
       </md-card>
-      <md-card>
+      <md-card md-theme="{{ showDarkTheme ? 'dark-orange' : 'default' }}" md-theme-watch>
         <md-card-title>
           <md-card-title-text>
-            <span class="md-headline">Card with  image</span>
+            <span class="md-headline">Card with image</span>
             <span class="md-subhead">Extra Large</span>
           </md-card-title-text>
         </md-card-title>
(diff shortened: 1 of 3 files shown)


ilovett:
preview:

image

src/components/card/demoBasicUsage/index.html
```diff
@@ -1,7 +1,7 @@
 <div ng-controller="AppCtrl" ng-cloak >
   <md-content class="md-padding" layout-xs="column" layout="row">
     <div flex-xs flex-gt-xs="50" layout="column">
-      <md-card>
+      <md-card md-theme="dark">
```
DevVersion: I'm not sure, it looks a bit weird in the demos, when the first card is dark. In my opinion it would be awesome if we have a checkbox for toggle the dark theme.

DevVersion:
@ilovett Tested it locally and it doesn't work, because all card demos use the same module, and that's a bug. (See #7693).

src/components/card/demoBasicUsage/index.html
ilovett: Yes I like that and perhaps with multiple themes with different light/dark contrasting values

DevVersion: Yes, I suggest only `default` and `default-dark`

ilovett:
Cool thanks, should I write "Depends on" or anything to signify that?

DevVersion:
@ilovett No there is no need to do that :)

ilovett:
Here's new preview

image

ilovett:
Want to mention that I'm using grey backgroundPalette due to #7600.

Once #7686 is merged I will happily add different color dark themes.

DevVersion:
@ilovett Great, so this PR is depending on two others? Maybe you can update the description if so.

ilovett:
Rebased on upstream/master after merge of #7686 , added different background colors.

image

ilovett:
While looking into this I noticed perhaps the color and background color SHOULD be something more like
```
color: {{ background-contrast }}
background: {{ background-color }}
```
Where the contrast ends up being the calculated transparent white or #FFF -- and then in addition, if palettes have `contrastStrongDarkColors` or `contrastStrongLightColors`

I would suggest a dynamic weight variable:

```
font-weight: {{ contrast-weight }};
```

Where weight is either `500` or `800` or `bold`, etc.

I read somewhere a while back that the entirety of dark themes will be revisited soon, so I've avoided making bigger changes than just changing the foreground text color.

PR #7705, opened 2016-03-22
fix(button): reduce selector length

crisbeto:
The `a.md-button.md-THEME_NAME-theme` in the button theme didn't really do anything, however
because there's quite a lot of nesting going on, it almost doubled the amount of selectors.

This was introduced in https://github.com/angular/material/commit/0178b8955f6bf2120a3a32fa8c51398557c9c059.


DevVersion:
LGTM

PR #7712, opened 2016-03-23
docs(progressCircular): missing Provider typo

soooooot:
typo: `$mdProgressCircular` ->  `$mdProgressCircularProvider`
otherwise angular can not find the module named `$mdProgressCircular`


crisbeto:
Nice catch! LGTM.

crisbeto:
I'm actually looking at it again and for some reason I marked `endIndeterminate`, `rotationDurationIndeterminate` and `easeFnIndeterminate` as arguments instead of properties. Would be a nice additions to change those as well. If you don't want to bother with it I can put in a PR for it later.

soooooot:
@crisbeto  updated. so pity I didn't catch it before..

Issue #7700, opened 2016-03-22
mdPanel: Create an MdPanelAnimation object and pass it with the general panel config.


ErinCoughlan:
This task involves:

* Creating the mdPanelAnimation object.
* Adding public documentation for mdPanelAnimation (copy from design doc).
* Adding a method to MdPanel to create a new MdPanelAnimation object.
* Implementing the following API methods:
    * `openFrom(string|!Element|!jQuery.Event| {top: string, left: string, height: string, width: string}) : MdPanelAnimation`
    * `closeTo(string|!Element| {top: string, left: string, height: string, width: string}) : MdPanelAnimation`

For both of those methods, ignore adding the animations, but make sure to store the BoundingRect. See [mdDialog](https://github.com/angular/material/blob/32951a7d1b17bb62938f3740d71504a70cf01f8d/src/components/dialog/dialog.js#L717).

PR #7718, opened 2016-03-23
feat(panel): Create an MdPanelAnimation object.

KarenParker:

Please review.

Fixes issue #7700

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -478,7 +478,19 @@ angular
  * @ngdoc object
  * @name MdPanelAnimation
  * @description
- * Object for configuring panel animations.
+ * Animation configuration object. To use, create an MdPanelAnimation with the
+ * desired properties, then pass the object as part of $mdPanel creation.
+ *
+ * Example:
+ *
+ * var panelAnimation = new MdPanelAnimation()
+ *     .openFrom(myButtonEl)
+ *     .closeTo('.my-button')
+ *     .withAnimation('');
+ *
+ * $mdPanel.create({
+ *   animation: panelAnimation
+ * });
  */
 
 /**
@@ -1347,6 +1359,10 @@ MdPanelAnimation.prototype.openFrom = function(openFrom) {
   openFrom = openFrom.target ? openFrom.target : openFrom;
 
   this._openFrom = this._getBoundingClientRect(this._getElement(openFrom));
+
+  if (angular.isUndefined(this._closeTo)) {
+    this._closeTo = angular.copy(this._openFrom);
+  }
   return this;
 };
 
@@ -1389,20 +1405,15 @@ MdPanelAnimation.prototype._getElement = function(el) {
  * @private
  */
 MdPanelAnimation.prototype._getBoundingClientRect = function(element) {
-  var source = angular.element((element || {}));
-  if (source && source.length) {
-    // Compute and save the target element's bounding rect, so that if the
-    // element is hidden when the dialog closes, we can shrink the dialog
-    // back to the same position it expanded from.
-    //
-    // Checking if the source is a rect object or a DOM element
+  if (element && element.length) {
     var bounds = { top:0, left:0, height:0, width:0 };
-    var hasFn = angular.isFunction(source[0].getBoundingClientRect);
+    // "element" may just be a JQLite wrapper around the object with top/left/height/width.
+    var hasFn = angular.isFunction(element[0].getBoundingClientRect);
 
     return {
-      element : hasFn ? source : undefined,
-      bounds  : hasFn ? source[0].getBoundingClientRect() : angular.extend({}, bounds, source[0]),
-      focus   : angular.bind(source, source.focus),
+      element : hasFn ? element : undefined,
+      bounds  : hasFn ? element[0].getBoundingClientRect() : angular.extend({}, bounds, element[0]),
+      focus   : angular.bind(element, element.focus),
     };
   }
 };
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -122,15 +122,15 @@ angular
  *   - `fullscreen` - `{boolean=}`: Whether the panel should be full screen.
  *     Applies the class `._md-panel-fullscreen` to the panel on open. Defaults
  *     to false.
+ *   - `animation` - `{MdPanelAnimation=}`: An MdPanelAnimation object that
+ *     specifies the animation of the panel. For more information, see
+ *     `MdPanelAnimation`.
  *
  * TODO(ErinCoughlan): Add the following config options.
  *   - `groupName` - `{string=}`: Name of panel groups. This group name is
  *     used for configuring the number of open panels and identifying specific
  *     behaviors for groups. For instance, all tooltips will be identified
  *     using the same groupName.
- *   - `animation` - `{MdPanelAnimation=}`: An MdPanelAnimation object that
- *     specifies the animation of the panel. For more information, see
- *     `MdPanelAnimation`.
  *   - `hasBackdrop` - `{boolean=}`: Whether there should be an opaque backdrop
  *     behind the panel. Defaults to false.
  *   - `disableParentScroll` - `{boolean=}`: Whether the user can scroll the
@@ -177,6 +177,17 @@ angular
  * the position config object.
  *
  * @returns {MdPanelPosition} panelPosition
+ */
+
+
+/**
+ * @ngdoc method
+ * @name $mdPanel#newPanelAnimation
+ * @description
+ * Returns a new instance of the MdPanelAnimation object. Use this to create
+ * the animation config object.
+ *
+ * @returns {MdPanelAnimation} panelAnimation
  */
 
 
@@ -459,6 +470,46 @@ angular
 
 
 /*****************************************************************************
+ *                               MdPanelAnimation                            *
+ *****************************************************************************/
+
+
+/**
+ * @ngdoc object
+ * @name MdPanelAnimation
+ * @description
+ * Object for configuring panel animations.
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelAnimation#openFrom
+ * @description
+ * Specifies where to start the open animation. `openFrom` accepts a
+ * click event object, query selector, DOM element, or a Rect object that
+ * is used to determine the bounds. When passed a click event, the location
+ * of the click will be used as the position to start the animation.
+ *
+ * @param {string|!Element|!jQuery.Event|
+ *     {top: string, left: string, height: string, width: string}}
+ * @returns {MdPanelAnimation}
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelAnimation#closeTo
+ * @description
+ * Specifies where to animate the dialog close. `closeTo` accepts a
+ * query selector, DOM element, or a Rect object that is used to determine
+ * the bounds.
+ *
+ * @param {string|!Element|
+ *     {top: string, left: string, height: string, width: string}}
+ * @returns {MdPanelAnimation}
+ */
+
+
+/*****************************************************************************
  *                                IMPLEMENTATION                             *
  *****************************************************************************/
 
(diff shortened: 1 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -1160,3 +1222,129 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+
+
+
+/*****************************************************************************
+ *                               MdPanelAnimation                            *
+ *****************************************************************************/
+
+
+/**
+ * Animation configuration object. To use, create an MdPanelAnimation with the
```
ErinCoughlan: Let's add all this documentation to the public documentation section above. I think it would be useful for users to know how to use the component. (Less useful for our internal docs.)

src/components/panel/panel.js
```diff
@@ -1160,3 +1222,129 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+/**
+ * Identify the bounding RECT for the target element.
+ * @param {!angular.JQLite} element
+ * @returns {!Object}
+ * @private
+ */
+MdPanelAnimation.prototype._getBoundingClientRect = function(element) {
+  var source = angular.element((element || {}));
+  if (source && source.length) {
+    // Compute and save the target element's bounding rect, so that if the
```
ErinCoughlan: Not sure this comment belongs here. You're saving the values in the other function.

src/components/panel/panel.js
```diff
@@ -1160,3 +1222,129 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+
+
+/**
+ * Identify the bounding RECT for the target element.
+ * @param {!angular.JQLite} element
+ * @returns {!Object}
+ * @private
+ */
+MdPanelAnimation.prototype._getBoundingClientRect = function(element) {
+  var source = angular.element((element || {}));
```
ErinCoughlan: nit: extra parens inside angular.element

ErinCoughlan: You're forcing element to be angular.JQLite, so this line is actually not necessary.

src/components/panel/panel.js
```diff
@@ -1160,3 +1222,129 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+ *     .withAnimation('');
+ *
+ * $mdPanel.create({
+ *   animation: panelAnimation
+ * });
+ *
+ * @final @constructor
+ */
+function MdPanelAnimation() {
+  /** @private {!Object} */
```
ErinCoughlan: {top: string, left: string, height: string, width: string} as the type? It's a little more specific as to what this object is.

src/components/panel/panel.spec.js
```diff
@@ -650,6 +650,83 @@ describe('$mdPanel', function() {
     });
   });
 
+  describe('animation logic', function() {
+    var config;
+    var mdPanelAnimation;
+    var myButton;
+
+    beforeEach(function() {
+      config = {
```
ErinCoughlan: Your other PR was merged, so you use the DEFAULT_CONFIG instead. (You may just need to rebase to get that change into this client.)

src/components/panel/panel.spec.js
```diff
@@ -650,6 +650,83 @@ describe('$mdPanel', function() {
...
+      };
+
+      mdPanelAnimation = $mdPanel.newPanelAnimation();
+
+      myButton = '<button>myButton</button>';
+      attachToBody(myButton);
+      myButton = angular.element(document.querySelector('button'));
+    });
+
+    describe('should openFrom', function() {
```
ErinCoughlan: nit: This doesn't sound like a sentence. I think you're missing a word somewhere:

"MdPanel animation logic should openFrom provided a selector"

src/components/panel/panel.js
```diff
@@ -1160,3 +1222,129 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+  return this._openFrom;
+};
+
+
+/**
+ * Gets the boundingClientRect for the closing animation.
+ * @returns {!Object}
+ */
+MdPanelAnimation.prototype.getCloseTo = function() {
+  return this._closeTo;
```
ErinCoughlan: I'm unsure where you plan on doing this, but if no closeTo is provided, but openFrom is, the panel should close to the same place it opened from.

Possibly check that here, but I'm also good if you plan on doing the checks when you add the animation.

src/components/panel/panel.js
KarenParker: Ah I realize the confusion. We don't always have an element. Sometimes it is a object with top/left/height/width. I updated the param type.

KarenParker:
Comments addressed.

src/components/panel/panel.js
```diff
@@ -1160,3 +1234,128 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+ */
+MdPanelAnimation.prototype._getBoundingClientRect = function(element) {
+  if (element && element.length) {
+    var bounds = { top:0, left:0, height:0, width:0 };
+    // "element" may just be a JQLite wrapper around the object with top/left/height/width.
+    var hasFn = angular.isFunction(element[0].getBoundingClientRect);
+
+    return {
+      element : hasFn ? element : undefined,
+      bounds  : hasFn ? element[0].getBoundingClientRect() : angular.extend({}, bounds, element[0]),
```
ErinCoughlan: super nit: I think this lines is too long. This file is using 80 characters, like the google style guide. (Same with the comment above and a few of your expects in the tests.)

ErinCoughlan:
LGTM

KarenParker:
Comment addressed.

@jelbourn any comments?

src/components/panel/panel.js
```diff
@@ -1160,3 +1239,131 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+ * @returns {!Object}
+ */
+MdPanelAnimation.prototype.getOpenFrom = function() {
+  return this._openFrom;
+};
+
+
+/**
+ * Gets the boundingClientRect for the closing animation.
+ * @returns {!Object}
```
jelbourn: Is it actually an instance of `ClientRect` ?

src/components/panel/panel.js
```diff
@@ -1160,3 +1239,131 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+ * click event object, query selector, DOM element, or a Rect object that
+ * is used to determine the bounds. When passed a click event, the location
+ * of the click will be used as the position to start the animation.
+ *
+ * @param {string|!Element|!jQuery.Event|
+ *     {top: string, left: string, height: string, width: string}} openFrom
+ * @returns {MdPanelAnimation}
+ */
+MdPanelAnimation.prototype.openFrom = function(openFrom) {
+  // Check if 'openFrom' is a jQuery.Event.
```
jelbourn: jQuery event or just an event?

src/components/panel/panel.js
```diff
@@ -1160,3 +1239,131 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+  openFrom = openFrom.target ? openFrom.target : openFrom;
+
+  this._openFrom = this._getBoundingClientRect(this._getElement(openFrom));
+
+  if (angular.isUndefined(this._closeTo)) {
+    this._closeTo = angular.copy(this._openFrom);
```
jelbourn: Why do you need to perform a deep copy?

src/components/panel/panel.js
```diff
@@ -1160,3 +1239,131 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+ * @param {!angular.JQLite} element
+ * @returns {!Object}
+ * @private
+ */
+MdPanelAnimation.prototype._getBoundingClientRect = function(element) {
+  if (element && element.length) {
+    var bounds = { top:0, left:0, height:0, width:0 };
+    // "element" may just be a JQLite wrapper around the object with
+    // top/left/height/width.
+    var hasFn = angular.isFunction(element[0].getBoundingClientRect);
```
jelbourn: You should be able to use `element instanceof angular.element`

src/components/panel/panel.js
KarenParker: Not quite. It can be, but essentially it is an object with {top: string, left: string, height: string, width: string}. I updated the Objects to specify this.

src/components/panel/panel.js
KarenParker: This is a $event. Do you know what type that is? We have referred to it in the past as a jQuery.Event.

src/components/panel/panel.js
ErinCoughlan: It also has `bottom` and `right`, which I will be using in the relative positioning CL. Should be make a typedef like thing somewhere in this file to specify it?

src/components/panel/panel.js
KarenParker: I don't think so. If the object of top/left/width/height is passed, it previously gets wrapped by an angular.element.

KarenParker:
Comments addressed.

src/components/panel/panel.js
KarenParker: The bottom and right aren't always there. If you need them for positioning,
we can add it in that CL.

src/components/panel/panel.js
jelbourn: ngMaterial doesn't depend on jQuery. Is there anything in the jqlite event that's not on the native event? If not, I think the type can just be `MouseEvent` (since it says it's always click)

src/components/panel/panel.js
jelbourn: I don't follow; the `@param` says `element` is always a non-nullable JQLite element. Under what circumstances would `getBoundingClientRect` not be present?

jelbourn: Oh, I see your comment. I think wrapping any non-element in `angular.element` is a bad idea; it strikes me as very unexpected and prone to doing the wrong thing.

src/components/panel/panel.js
KarenParker: I am switching to just an Event.

src/components/panel/panel.js
KarenParker: Fixed

KarenParker:
Comments addressed

src/components/panel/panel.js
```diff
@@ -1167,3 +1246,139 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+ *
+ * @param {string|!Element|!Event|
+ *     {top: string, left: string, height: string, width: string}} openFrom
+ * @returns {MdPanelAnimation}
+ */
+MdPanelAnimation.prototype.openFrom = function(openFrom) {
+  // Check if 'openFrom' is an Event.
+  openFrom = openFrom.target ? openFrom.target : openFrom;
+
+  if (!!openFrom.top || !!openFrom.left || !!openFrom.height || !!openFrom.width) {
```
jelbourn: You don't need the `!!` in a conditional, only when you're working against an API that explicitly accepts or returns `boolean`.

src/components/panel/panel.js
```diff
@@ -1167,3 +1246,139 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+    this._openFrom = {
+      element: undefined,
+      bounds: angular.extend({top: 0, left: 0, height: 0, width: 0}, openFrom),
+      focus: undefined
+    };
+  } else {
+    this._openFrom = this._getBoundingClientRect(this._getElement(openFrom));
+  }
+
+  if (angular.isUndefined(this._closeTo)) {
```
jelbourn: Would it work to just say `!this._closeTo` ?

src/components/panel/panel.js
```diff
@@ -1167,3 +1246,139 @@ MdPanelPosition.prototype.right = function(opt_right) {
...
+ * @param {string|!Element|!Event|
+ *     {top: string, left: string, height: string, width: string}} openFrom
+ * @returns {MdPanelAnimation}
+ */
+MdPanelAnimation.prototype.openFrom = function(openFrom) {
+  // Check if 'openFrom' is an Event.
+  openFrom = openFrom.target ? openFrom.target : openFrom;
+
+  if (!!openFrom.top || !!openFrom.left || !!openFrom.height || !!openFrom.width) {
+    this._openFrom = {
```
jelbourn: Would it make sense to introduce a new class for this? Something like `PanelAnimationPoint`?

src/components/panel/panel.js
jelbourn: This check (and the object creation) could also be broken out into a helper function since it's duplicated here and in the next function

src/components/panel/panel.js

src/components/panel/panel.js
KarenParker: Instead of dealing with a new class, I specified what you expect with the object.

KarenParker:
Comments addressed.

jelbourn:
LGTM

PR #7719, opened 2016-03-23
refactor(interim) cleanup interim and split it up.

DevVersion:
I recently played with the interim elements and try to understand them.

But it's really hard, because the interim elements are really unstructured developed. 
So I just rewrote the complete interim element provider, but let's better say refactored and optimized.

That means, I've splitted the interim element up, into the provider, service and the base.
That's really clear and it's very easy to understand the code, since some function names have been changed as well.

But notice: The API is still the same, no breaking changes - Tests still passing.

I know, I didn't file an issue about that, so I'm fine, when this PR will be rejected. I just did it, to improve my comprehension of the interim elements.

**Todo**
- [ ] Reintroduce JSDocs for the most functions (didn't add it yet - to keep the source clear)


src/core/services/interimElement/interimElementBase.js
```diff
@@ -0,0 +1,217 @@
...
+      remove: _runRemoveExecution,
+      options: options,
+      deferred: $q.defer()
+    };
+
+    function _runRemoveExecution(response, isCancelled, opts) {
+      // Abort and reject when the previous show failed.
+      if (!element) return $q.when(false);
+
+      options = angular.extend(options, opts || {});
```
crisbeto: You don't need to pass in an empty object here, `extend` won't do anything if `opts` is falsey.

src/core/services/interimElement/interimElementBase.js
```diff
@@ -0,0 +1,217 @@
...
+
+      } catch (err) {
+        defer.reject(err.message);
+      }
+
+      return defer.promise;
+    }
+
+    function hideElement() {
+      var defer = $$q.defer();
```
crisbeto: What's the difference between `$$q` and `$q`. I can see that you've only used `$$q` in a couple of places.

src/core/services/interimElement/interimElementBase.js
DevVersion: True, but including an extra check, which checks for falsey is also a unnecessary step. Also imported from the previous code. This is needed, when the options object is undefined.

src/core/services/interimElement/interimElementBase.js
```diff
@@ -0,0 +1,217 @@
...
+
+      if (angular.isFunction(parent)) {
+        parent = parent(options.scope, element, options);
+      } else if (angular.isString(parent)) {
+        parent = angular.element($document[0].querySelector(parent));
+      } else {
+        parent = angular.element(parent);
+      }
+
+      if (!(parent || {}).length) {
```
crisbeto: This one seems a little strange, instead you could do `if (parent && !parent.length)`.

src/core/services/interimElement/interimElementBase.js
```diff
@@ -0,0 +1,217 @@
...
+        var possibleParent;
+        if ($rootElement[0] && $rootElement[0].querySelector) {
```
crisbeto: Is there a case where `$rootElement[0]` wouldn't be defined?

src/core/services/interimElement/interimElementBase.js
DevVersion: @crisbeto Great question, had the same question today and asked Thomas about that.

Quote from Thomas:
> $$q does NOT depend upon the $digest() cycle nor does it trigger $digest()

See `$$q` in the source. https://github.com/angular/angular.js/blob/master/src/ng/q.js#L228

src/core/services/interimElement/interimElementProvider.js
```diff
@@ -0,0 +1,278 @@
+angular
+  .module('material.core')
+  .provider('$$interimElement', InterimElementProvider);
+
+function InterimElementProvider() {
+  function test () {
+    console.log("LOL");
```
crisbeto: lol

src/core/services/interimElement/interimElementBase.js
DevVersion: Yep, had the same thoughts before, but then no parent wasn't found. So I changed it back.

src/core/services/interimElement/interimElementProvider.js
DevVersion: Hahaha :smile:

src/core/services/interimElement/interimElementProvider.js
```diff
@@ -0,0 +1,278 @@
...
+
+    return provider;
+  }
+
+  function parseDefinition(definition, name) {
+    definition.methods = definition.methods || [];
+    definition.options = definition.options || function() { return {}; };
+
+    if (name) {
+      if (/^cancel|hide|show$/.test(name)) {
```
crisbeto: Using a regex here might be overkill.

src/core/services/interimElement/interimElementProvider.js
```diff
@@ -0,0 +1,278 @@
...
+        throw new Error("Method '_options' in " + interimName + " is reserved!");
+      }
+    }
+
+    return definition;
+  }
+
+  function factory($$interimElement, $injector) {
+    // This is needed, otherwise ng-annotator can't detect this function as a factory $get.
+    'ngInject';
```
crisbeto: I believe it works if it's in a comment as well.

src/core/services/interimElement/interimElementBase.js
crisbeto: Good to know :+1:

src/core/services/interimElement/interimElementBase.js
DevVersion: I don't think so, I will change that. Great catch.

src/core/services/interimElement/interimElementBase.js
crisbeto: I mean that Angular just won't do anything if there's only one argument, you don't need to do any extra checks.

DevVersion: Just took a look at the source code of `angular.extend` and you're right. Thanks!

src/core/services/interimElement/interimElementBase.js
```diff
@@ -0,0 +1,217 @@
...
+    return API = {
+      show: createAndShowElement,
+      remove: _runRemoveExecution,
+      options: options,
+      deferred: $q.defer()
+    };
+
+    function _runRemoveExecution(response, isCancelled, opts) {
+      // Abort and reject when the previous show failed.
+      if (!element) return $q.when(false);
```
crisbeto: This promise is being resolved, not rejected.

src/core/services/interimElement/interimElementProvider.js
DevVersion: Using a comment, in the same line of the function, also looks really weird. :disappointed:

src/core/services/interimElement/interimElementBase.js
DevVersion: Very nice catch, but resolve is actually correct, I just messed the comment up.

src/core/services/interimElement/interimElementBase.js
```diff
@@ -0,0 +1,217 @@
...
+
+        onRemove: function transitionOut(scope, element) {
+          return element && $animate.leave(element) || $q.when();
+        }
+      }, options);
+    }
+
+
+    function compileElement(options) {
+      return options.skipCompile ? mockCompile() : $mdCompiler.compile(options);
```
crisbeto: Considering that this is all in the same function, you just inline it.

src/core/services/interimElement/interimElementProvider.js
```diff
@@ -0,0 +1,274 @@
...
+
+      interim.remove(reason, true, options || {});
+
+      return interim.deferred.promise;
+    }
+
+    function _destroyInterim(target) {
+      var interim = !target ? showQueue.shift() : null;
+
+      var cntr = angular.element(target).length ? angular.element(target)[0].parentNode : null;
```
crisbeto: You can replace `angular.element(target)[0].parentNode` with `target.parent()`.

src/core/services/interimElement/interimElementBase.js
DevVersion: Yes I agree, that's possible, I just thought, that it would be clearer without inlining.

src/core/services/interimElement/interimElementProvider.js
DevVersion: Very nice. That's not really a smart step. (I wasn't that :smile: )

src/core/services/interimElement/interimElementProvider.js
crisbeto: I'm not too familiar with how these annotations work, but in other modules I've seen `@ngInject` being thrown into the doc string.

DevVersion: Yes, indeed, using that in the doc string causes warning like these:
image

src/core/services/interimElement/interimElementBase.js
DevVersion: Took a look again, should be `!parent || !parent.length`

src/core/services/interimElement/interimElementProvider.js
```diff
@@ -0,0 +1,274 @@
...
+
+      interim.remove(reason, true, options || {});
+
+      return interim.deferred.promise;
+    }
+
+    function _destroyInterim(target) {
+      var interim = !target ? showQueue.shift() : null;
+
+      var cntr = angular.element(target).length ? target.parentNode : null;
```
crisbeto: You can also ditch the `angular.element` call here and just check if `target` is present.

src/core/services/interimElement/interimElementProvider.js
```diff
@@ -0,0 +1,274 @@
...
+        var filtered = showQueue.filter(function(entry) {
+          var currNode = entry.options.element[0];
+          return currNode === cntr;
+        });
+
+        // Note: this function might be called when the element already has been removed, in which
+        // case we won't find any matches. That's ok.
+        if (filtered.length > 0) {
+          interim = filtered[0];
+          showQueue.splice(showQueue.indexOf(interim), 1);
```
crisbeto: It could be a good idea to checkout if the index here is bigger than -1. If the `interim` isn't in the `showQueue`, the splice will still execute and return an array with only the last element.

src/core/services/interimElement/interimElementProvider.js
```diff
@@ -0,0 +1,274 @@
...
+      }
+
+      // Example: $mdDialog.alert() will return a new alert preset
+      publicService[name] = function(arg) {
+        // If argOption is supplied, eg `argOption: 'content'`, then we assume
+        // if the argument is not an options object then it is the `argOption` option.
+        //
+        // @example `$mdToast.simple('hello')`
+
+        if (arguments.length && definition.argOption && !angular.isObject(arg) && !angular.isArray(arg)) {
```
crisbeto: Note that `isObject` will return true for arrays as well.

DevVersion: Correct, that's why we check later for an Array.

DevVersion:
Closing this, because it seems like it is a really dangerous change.

Issue #7384, opened 2016-03-02
mdSelect: (feat) Adds an attribute, selected-text, to allow custom text when items are selected.


ErinCoughlan:
Examples:
100 selected items
Choose one...

`selected-text` should be a '=?' binding. Users can look at the scope to see what items are selected and format the text that way.

One example on the demo site should include the empty text case, where no items are selected.

PR #7721, opened 2016-03-24
feat(select): Adding selected-text to md-select.

DerekLouie:
@gmoothart @ErinCoughlan @jelbourn @ThomasBurleson

This adds the attribute `selected-text` to the md-select component.

Please Review.

References issue #7384

selectedtext1

selectedtext2

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -38,6 +38,8 @@ angular.module('material.components.select', [
  * @param {expression=} md-on-close Expression to be evaluated when the select is closed.
  * @param {expression=} md-on-open Expression to be evaluated when opening the select.
  * Will hide the select options and show a spinner until the evaluated promise resolves.
+ * @param {expression=} selected-text Expression to be evaluated that will return a string
+ * to be displayed as a placeholder in the select input box when it is closed.
  * @param {string=} placeholder Placeholder hint text.
  * @param {string=} aria-label Optional label for accessibility. Only necessary if no placeholder or
  * explicit label is present.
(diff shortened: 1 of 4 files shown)


DevVersion:
@DerekLouie Can you please check the CI, it seems like your changes broke one test.

src/components/select/select.js
```diff
@@ -116,6 +116,9 @@ angular.module('material.components.select', [
 function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $parse) {
   return {
     restrict: 'E',
+    scope: {
+      selectedText: '=?selectedText'
```
DevVersion: Why are you introducing here a new scope? Currently we didn't have any scope, this may break some applications.

src/components/select/select.spec.js
```diff
@@ -258,6 +258,26 @@ describe('<md-select>', function() {
       }
     }));
 
+    it('displays selected-text when specified', inject(function($rootScope) {
+      var selectedText = "Hello World";
+      $rootScope.getSelectedText = function() { return selectedText; };
+
+      var select = setupSelect('ng-model="someVal", selected-text="getSelectedText()"', null, true).find('md-select');
```
DevVersion: Won't work a simple interpolated variable? Instead of calling a function?

```js
$rootScope.selectedText = "Hello World"
```

src/components/select/select.js
DerekLouie: Ahh yes, this is why the test broke, I have now changed the code to copy the same methods that are used for `mdOnClose` / `mdOnOpen`

src/components/select/select.spec.js
DerekLouie: That works too, I figured if a test shows a function will be executed it should also cover the passed in string case, but not necessarily vice versa.

src/components/select/select.js
```diff
@@ -257,9 +259,15 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
...
         mdSelectCtrl.setIsPlaceholder(!text);
-        // Use placeholder attribute, otherwise fallback to the md-input-container label
-        var tmpPlaceholder = attr.placeholder || (containerCtrl && containerCtrl.label ? containerCtrl.label.text() : '');
-        text = text || tmpPlaceholder || '';
+
+        if (attr.selectedText) {
+          text = $parse(attr.selectedText)(scope);
+        } else {
+          // Use placeholder attribute, otherwise fallback to the md-input-container label
+          var tmpPlaceholder = attr.placeholder || (containerCtrl && containerCtrl.label ? containerCtrl.label.text() : '');
```
ErinCoughlan: nit: some line breaks here will make this easier to read

ErinCoughlan:
LGTM

jelbourn:
LGTM

crisbeto:
LGTM

ThomasBurleson:
@DerekLouie - I changed the attribute name to `md-selected-text` (per our namespacing conventions).

PR #7727, opened 2016-03-24
fix(mdToolbar): Prevent scrollbar from hiding when using scroll schrink

maximethebault:
As it can be seen on the demo, when scroll shrink is enabled and the toolbar is visible, the bottom of the scrollbar is not visible. This PR fixes that bug.

### DIFF
--- a/src/components/toolbar/toolbar.js
+++ b/src/components/toolbar/toolbar.js
@@ -169,6 +169,7 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
 
           element.css($mdConstant.CSS.TRANSFORM, translateY([-y * shrinkSpeedFactor]));
           contentElement.css($mdConstant.CSS.TRANSFORM, translateY([(toolbarHeight - y) * shrinkSpeedFactor]));
+          contentElement.css('margin-bottom', (toolbarHeight - y) * shrinkSpeedFactor);
 
           prevScrollTop = scrollTop;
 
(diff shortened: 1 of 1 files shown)


src/components/toolbar/toolbar.js
```diff
@@ -169,6 +169,7 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
 
           element.css($mdConstant.CSS.TRANSFORM, translateY([-y * shrinkSpeedFactor]));
           contentElement.css($mdConstant.CSS.TRANSFORM, translateY([(toolbarHeight - y) * shrinkSpeedFactor]));
+          contentElement.css('margin-bottom', (toolbarHeight - y) * shrinkSpeedFactor);
```
DevVersion: You used the same operation / value twice, this can be refactored into a variable.

src/components/toolbar/toolbar.js
```diff
@@ -214,11 +215,11 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
...
           // As the user scrolls down, the content will be transformed up slowly
           // to put the content underneath where the toolbar was.
-          var margin = (-toolbarHeight * shrinkSpeedFactor) + 'px';
+          var margin = (-toolbarHeight * shrinkSpeedFactor);
 
           contentElement.css({
-            "margin-top": margin,
-            "margin-bottom": margin
+            "margin-top": margin + 'px',
+            "margin-bottom": -margin + 'px'
```
DevVersion: Additionally you broke the tests, please check the CI.

maximethebault:
Good news! It looks like it's fixed by a more recent PR, much more complete than this one :)
Go see #7763

Issue #7733, opened 2016-03-24
mdPanel: Switch from addClass/removeClass/toggleClass to show/hide.


KarenParker:
addClass/removeClass/toggleClass forces certain settings, like hasBackdrop, to be controlled by the user. We don't want to force the user to remember each setting they have to control when they want to hide a panel.

PR #7735, opened 2016-03-24
update(panel): Add show/hide and comment that addClass/removeClass/toggleClass should not show/hide.

KarenParker:

Please review.

Fixes issue #7733.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -273,6 +273,33 @@ angular
 
 /**
  * @ngdoc method
+ * @name MdPanelRef#addClass
+ * @description
+ * Adds a class to the panel. DO NOT use this to hide/show the panel.
+ *
+ * @param {string} newClass Class to be added.
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#removeClass
+ * @description
+ * Removes a class from the panel. DO NOT use this to hide/show the panel.
+ *
+ * @param {string} oldClass Class to be removed.
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#toggleClass
+ * @description
+ * Toggles a class on the panel. DO NOT use this to hide/show the panel.
+ *
+ * @param {string} toggleClass Class to be toggled.
+ */
+
+/**
+ * @ngdoc method
  * @name MdPanelRef#focusOnOpen
  * @description
  * Focuses the panel content if the focusOnOpen config value is true.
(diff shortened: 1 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -254,34 +253,22 @@ angular
 
 /**
  * @ngdoc method
- * @name MdPanelRef#addClass
```
jelbourn: Why not leave `addClass` et al here, since they're separately useful from show/hide?

KarenParker: Done

src/components/panel/panel.js
```diff
@@ -254,34 +253,49 @@ angular
...
  * @ngdoc method
  * @name MdPanelRef#removeClass
  * @description
- * Removes a class from the panel. This can be used to hide/show the panel.
+ * Removes a class from the panel. DO NOT use this to hide/show the panel.
  *
- * @param {string} oldClass Class to be removed.
- * @returns {!angular.$q.Promise} A promise that is resolved when the panel has
- * removed the old class and animations are completed.
+ * @param {string} odlClass Class to be removed.
```
jelbourn: `odl`

src/components/panel/panel.js
```diff
@@ -738,29 +761,28 @@ MdPanelRef.prototype.detachOnly = function() {
...
- * Add a class to the panel. This can control the panel's visibility.
+ * Shows the panel.
  *
- * @param {string} newClass Class to be added.
  * @returns {!angular.$q.Promise} A promise that is resolved when the panel has
- * the new class and animations finish.
+ * shown and animations finish.
  */
-MdPanelRef.prototype.addClass = function(newClass) {
+MdPanelRef.prototype.show = function() {
```
jelbourn: Internally, don't `show` and `hide` simply need to call `addClass` and do some extra stuff?
```js
MdPanelRef.prototype.show = function() {
  return this.removeClass('internal-hidden-class').then( /** show backdrop? */ );
}
```

src/components/panel/panel.js
```diff
@@ -769,59 +791,85 @@ MdPanelRef.prototype.addClass = function(newClass) {
...
+ *
+ * @param {string} newClass Class to be added.
+ */
+MdPanelRef.prototype.addClass = function(newClass) {
+  if (!this._panelContainer) {
+    throw new Error('Panel does not exist yet. Call open() or attach().');
+  }
+
+  !this._panelContainer.hasClass(newClass) &&
+  this._panelContainer.addClass(newClass);
```
jelbourn: This addClass function should no-op if the class is already there.

src/components/panel/panel.js
KarenParker: Yes, but it doesn't return a promise.

src/components/panel/panel.js
KarenParker: It does already.... I reformatted it for clarity.

jelbourn:
LGTM

gmoothart:
:+1:

ErinCoughlan:
LGTM

PR #7676, opened 2016-03-21
fix(button): icon in icon-button not centered vertically

aldeso:
`md-icon-button` currently the icon is  not centered vertically

Fix line-height value to 100%

PR #7738, opened 2016-03-24
feat(checkbox): add classes md-full and md-right; change styling  

aldeso:
- refactor css styling for better support alignment with text (also with empty label), tested on IE11, Edge, Chrome, Safari
- add class md-full to change inline-flex to flex
- add class md-right to change the places of _md-container and _md-label

Please Review and thanks for some helpful comments...

René


DevVersion:
@aldeso Seems like you committed the button alignment fix as well. #7676

ThomasBurleson:
Scope of change is too large.

PR #7741, opened 2016-03-25
fix(icon): Use data URLS instead of $templateCache

### LINKED ISSUE #6531: (Re)cache used icons like "md-tabs-arrow'" after $templateCache.removeAll()
Hi guys,

For my app we need to delete the entire templateCache after creating or updating some GUI forms. The problem is the MdIconProvider with function "preloadIcons" cached some icons in the templatecache.

After $templateCache.removeAll() the cached icons are gone ;)

Is there an option to rerun this "preloadIcons" function on runtime (in a controler)? 

Would be nice!

Grtz Dennis

programmist:
Fixes #6531

### DIFF
--- a/src/components/icon/js/iconService.js
+++ b/src/components/icon/js/iconService.js
@@ -1,6 +1,14 @@
-angular
-  .module('material.components.icon')
-  .provider('$mdIcon', MdIconProvider);
+  angular
+    .module('material.components.icon')
+    .constant('$$mdSvgRegistry', {
+        'mdTabsArrow':   'data:image/svg+xml;base64,PHN2ZyB2ZXJzaW9uPSIxLjEiIHg9IjBweCIgeT0iMHB4IiB2aWV3Qm94PSIwIDAgMjQgMjQiPjxnPjxwb2x5Z29uIHBvaW50cz0iMTUuNCw3LjQgMTQsNiA4LDEyIDE0LDE4IDE1LjQsMTYuNiAxMC44LDEyICIvPjwvZz48L3N2Zz4=',
+        'mdClose':       'data:image/svg+xml;base64,PHN2ZyB2ZXJzaW9uPSIxLjEiIHg9IjBweCIgeT0iMHB4IiB2aWV3Qm94PSIwIDAgMjQgMjQiPjxnPjxwYXRoIGQ9Ik0xOSA2LjQxbC0xLjQxLTEuNDEtNS41OSA1LjU5LTUuNTktNS41OS0xLjQxIDEuNDEgNS41OSA1LjU5LTUuNTkgNS41OSAxLjQxIDEuNDEgNS41OS01LjU5IDUuNTkgNS41OSAxLjQxLTEuNDEtNS41OS01LjU5eiIvPjwvZz48L3N2Zz4=',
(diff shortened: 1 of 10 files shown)


programmist:
@ThomasBurleson: Could you have a look when you get a minute? Thanks!

src/components/icon/js/iconService.js
```diff
@@ -1,5 +1,13 @@
   angular
     .module('material.components.icon' )
+    .constant('svgRegistry', {
```
DevVersion: @programmist What do you think about privatizing that a bit more. Something like `$$mdSvgRegistry`?

For example, `$$interimElement` is also an internal service for ngMaterial, so I think the `$$` prefix should be used consistent. Thoughts?

programmist: Good suggestion! I wouldn't want anyone accidentally trying to use this.

DevVersion: Yes, that was my thought :)

programmist: Done. Commit amended.

src/components/icon/js/iconService.js
```diff
@@ -1,5 +1,13 @@
...
+        'mdCancel':      'data:image/svg+xml;base64,' + btoa('<svg version="1.1" x="0px" y="0px" viewBox="0 0 24 24"><g><path d="M12 2c-5.53 0-10 4.47-10 10s4.47 10 10 10 10-4.47 10-10-4.47-10-10-10zm5 13.59l-1.41 1.41-3.59-3.59-3.59 3.59-1.41-1.41 3.59-3.59-3.59-3.59 1.41-1.41 3.59 3.59 3.59-3.59 1.41 1.41-3.59 3.59 3.59 3.59z"/></g></svg>'),
+        'mdMenu':        'data:image/svg+xml;base64,' + btoa('<svg version="1.1" x="0px" y="0px" viewBox="0 0 24 24"><path d="M3,6H21V8H3V6M3,11H21V13H3V11M3,16H21V18H3V16Z" /></svg>'),
+        'mdToggleArrow': 'data:image/svg+xml;base64,' + btoa('<svg version="1.1" x="0px" y="0px" viewBox="0 0 48 48"><path d="M24 16l-12 12 2.83 2.83 9.17-9.17 9.17 9.17 2.83-2.83z"/><path d="M0 0h48v48h-48z" fill="none"/></svg>'),
+        'mdCalendar':    'data:image/svg+xml;base64,' + btoa('<svg xmlns="http://www.w3.org/2000/svg" width="24" height="24" viewBox="0 0 24 24"><path d="M19 3h-1V1h-2v2H8V1H6v2H5c-1.11 0-1.99.9-1.99 2L3 19c0 1.1.89 2 2 2h14c1.1 0 2-.9 2-2V5c0-1.1-.9-2-2-2zm0 16H5V8h14v11zM7 10h5v5H7z"/></svg>')
```
Frank3K: While this is good for readability for developers, would you want to let the browser do this?

Frank3K:
Just wondering, is there any performance impact when changing from template cache to base64 data? In my experience base64 is slower, probably because the browser needs to perform a decode step and maybe because it is less optimized since it's not as widely used.

src/components/icon/js/iconService.js
programmist: I'm not too concerned with it, but I do agree that using escaped, unencoded strings would be better. I'll submit that change shortly.

ThomasBurleson:
@programmist - is this PR still valid and ready ?

programmist:
@ThomasBurleson: It was probably ok as-is, but I'll went ahead and pre-encoded those SVG strings rather than relying on the browser to do it at runtime.

ThomasBurleson:
@programmist - Thx Tony.

ThomasBurleson:
@programmist  - can you rebase with `master` please ?

programmist:
@ThomasBurleson: Sure, I'll comment here once rebase is done.

programmist:
@ThomasBurleson: Rebased and ready to merge at your convenience.

PR #7744, opened 2016-03-25
fix(build): add @ngInject annotations everywhere

### LINKED ISSUE #7610: Many injected functions don't have @ngInject

crisbeto:
* Adds @ngInject to everything that was missing it.
* Removes any instances of @ngInject that were in the doc strings.
* Makes all the @ngInject declarations consistent with each other.
* Adds newlines at the ends of files that didn't have them and trims any excess whitespace.

Closes #7610.

This doesn't really need a whole lot of reviewing, but it might be a good idea to run a build locally, just in case and see whether Closure complains. It all seems fine locally with `gulp build-all-modules --mode=closure`, but I'm not too familiar with the build setup.

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -7,6 +7,7 @@ var ITEM_HEIGHT   = 41,
     MENU_PADDING  = 8,
     INPUT_PADDING = 2; // Padding provided by `md-input-container`
 
+/* @ngInject */
 function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming, $window,
                              $animate, $rootElement, $attrs, $q) {
   //-- private variables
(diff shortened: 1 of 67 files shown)


topherfangio:
Regarding the portions where you moved it out of the ngdoc comment, can you revert those back so that they match the [style guide](https://google.github.io/styleguide/angularjs-google-style.html#nginject)?

crisbeto:
Sure, but the ngdocs task throws warnings for some of them.

topherfangio:
What is the warning?

crisbeto:
That it's not a valid tag/annotation (I don't remember the exact wording).

crisbeto:
Alright, I've put the ones that were in doc string back. I'll only move the one in `media.js` out because it throws the warning. Also here's what I was talking about, I think it only complains for those that get generated into pages:

```
warn:    Invalid tags found - doc "$mdMedia" (service)  - from file "src/core/util/media.js" - starting at line 4, ending at line 89
Line: 87: @ngInject ...
```

ThomasBurleson:
@jelbourn - Is this an issue for you when do Closure builds on G3?

ThomasBurleson:
This is not currently an issue within our build processes.

PR #7748, opened 2016-03-25
fix(chips): do not trim the input model.

### LINKED ISSUE #7243: md-chips input : issuse with space characters inputed
1 - Remove all items in input chips
2 - Press space button to input space character
=> Can't press BackSpace button to remove/delete, same for many spaces.

DevVersion:
* The chips input should not trim the text, because otherwise the buffer will be always falsey, even when there are spaces in the input, and this would cause the backspace not to work.

Fixes #7243

### DIFF
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -388,7 +388,58 @@ describe('<md-chips>', function() {
           expect(enterEvent.preventDefault).toHaveBeenCalled();
         }));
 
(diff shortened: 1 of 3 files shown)


topherfangio:
Looks good, but do you think we should trim any new chips so that users don't accidentally add spaces into a chip?

DevVersion:
@topherfangio I like the idea. I will implement that and let you know.

DevVersion:
@topherfangio Done :)

DevVersion:
@topherfangio Please review again.

PR #7749, opened 2016-03-25
fix(virtualRepat): make demos responsive.

### LINKED ISSUE #7333: Virtual repeat: Container does not fit on screen on mobile device
How to reproduce:
1. Go to Virtual Repeat in the Demo area
2. Observe that the container is wider than the viewable area (and thus it's possible to scroll horisontally).

Tested on
Angular Material version 1.0.5
Model: Motorola Nexus 6
Browser: Chrome version 47
Operating System: Android 5.1.1 (Lollipop)

DevVersion:
* The demos had currently fixed widths and that causes overflow issues on mobile devices.

Fixes #7333

### DIFF
--- a/src/components/virtualRepeat/demoDeferredLoading/style.css
+++ b/src/components/virtualRepeat/demoDeferredLoading/style.css
@@ -1,6 +1,7 @@
 #vertical-container {
   height: 292px;
-  width: 400px;
+  width: 100%;
+  max-width: 400px;
 }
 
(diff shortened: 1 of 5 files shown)


topherfangio:
LGTM, but is there any way that you can add a fix for the Horizontal Demo so that the bottom border is visible? It's really inconsistent and awkward without it :smile: 

screen shot 2016-03-25 at 1 32 39 pm

PR #7750, opened 2016-03-25
Add md-autocomplete-snap="width"

aperl:
This allows a `md-autocomplete-wrap` to be applied to an ancestor of a `md-autocomplete` and cause that element to be used for evaluating the width of the autocomplete instead of the the `md-autocomplete-wrap` element. This addresses #4450 which constrains the width of the menu that pops up to the width of the auto complete wrapper.
autocomplete
When using chips and possibly in other situations it would be desirable to have the menu be the width of the chips component.
autocompleteoverride
Adding support for this attribute would allow for this and other similar behavior.

A working version of this change can be seen at https://codepen.io/allenperl/pen/KzmQwr.

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -257,7 +257,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       scroller: $element[0].querySelector('.md-virtual-repeat-scroller'),
       ul:    $element.find('ul')[0],
       input: $element.find('input')[0],
-      wrap:  $element.find('md-autocomplete-wrap')[0],
+      wrap:  getWrapTarget(),
       root:  document.body
     };
 
@@ -267,6 +267,17 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
 
     inputModelCtrl = elements.$.input.controller('ngModel');
   }
+
+	/**
+	 * Finds the correct element that will determine the width
+	 * @returns {*}
+	 */
+	function getWrapTarget() {
+    for (var element = $element; element.length; element = element.parent()) {
+      if (angular.isDefined(element.attr('md-autocomplete-wrap-override'))) return element[ 0 ];
+    }
+    return $element.find('md-autocomplete-wrap')[0];
+	}
 
   /**
    * Finds the element that the menu will base its position on
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -268,16 +268,16 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     inputModelCtrl = elements.$.input.controller('ngModel');
   }
 
-	/**
-	 * Finds the correct element that will determine the width
-	 * @returns {*}
-	 */
-	function getWrapTarget() {
+  /**
+   * Find the target for the wrap element which determines menu width
+   * @returns {*}
+   */
+  function getWrapTarget() {
     for (var element = $element; element.length; element = element.parent()) {
       if (angular.isDefined(element.attr('md-autocomplete-wrap-override'))) return element[ 0 ];
     }
     return $element.find('md-autocomplete-wrap')[0];
-	}
+  }
 
   /**
    * Finds the element that the menu will base its position on
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -251,43 +251,51 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    * Gathers all of the elements needed for this controller
    */
   function gatherElements () {
+
+    var snapWrap = gatherSnapWrap();
+
     elements = {
       main:  $element[0],
       scrollContainer: $element[0].querySelector('.md-virtual-repeat-container'),
       scroller: $element[0].querySelector('.md-virtual-repeat-scroller'),
       ul:    $element.find('ul')[0],
       input: $element.find('input')[0],
-      wrap:  getWrapTarget(),
+      wrap:  snapWrap.wrap,
+      snap:  snapWrap.snap,
       root:  document.body
     };
 
     elements.li   = elements.ul.getElementsByTagName('li');
-    elements.snap = getSnapTarget();
     elements.$    = getAngularElements(elements);
 
     inputModelCtrl = elements.$.input.controller('ngModel');
   }
 
   /**
-   * Find the target for the wrap element which determines menu width
-   * @returns {*}
+   * Gathers the snap and wrap elements
+   *
    */
-  function getWrapTarget() {
-    for (var element = $element; element.length; element = element.parent()) {
-      if (angular.isDefined(element.attr('md-autocomplete-wrap-override'))) return element[ 0 ];
+  function gatherSnapWrap() {
+    var element;
+    var value;
+    for (element = $element; element.length; element = element.parent()) {
+      value = element.attr('md-autocomplete-snap');
+      if (angular.isDefined(value)) break;
     }
-    return $element.find('md-autocomplete-wrap')[0];
-  }
 
-  /**
-   * Finds the element that the menu will base its position on
-   * @returns {*}
-   */
-  function getSnapTarget () {
-    for (var element = $element; element.length; element = element.parent()) {
-      if (angular.isDefined(element.attr('md-autocomplete-snap'))) return element[ 0 ];
+    if (element.length) {
+      var snapWidth = value.toLowerCase() === 'width';
+      return {
+        snap: element[0],
+        wrap: snapWidth ? element[0] : $element.find('md-autocomplete-wrap')[0]
+      };
     }
-    return elements.wrap;
+
+    var wrap = $element.find('md-autocomplete-wrap')[0];
+    return {
+      snap: wrap,
+      wrap: wrap
+    };
   }
 
   /**
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -284,10 +284,9 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     }
 
     if (element.length) {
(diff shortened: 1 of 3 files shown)


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -207,6 +207,17 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     elements.$    = getAngularElements(elements);
   }
 
+	/**
+	 * Finds the correct element that will determine the width
```
DevVersion: That's not a valid description. `getWrapTarget` will be not used to only determine the width. As the function says, it's just scanning for the `wrap` target.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -207,6 +207,17 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+	 * @returns {*}
+	 */
+	function getWrapTarget() {
+    for (var element = $element; element.length; element = element.parent()) {
+      if (angular.isDefined(element.attr('md-autocomplete-wrap-override'))) return element[ 0 ];
+    }
+    return $element.find('md-autocomplete-wrap')[0];
```
DevVersion: You should fix the indentation here and above. Using 2 spaces for one indentation.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -207,6 +207,17 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     elements.$    = getAngularElements(elements);
   }
 
+	/**
+	 * Finds the correct element that will determine the width
+	 * @returns {*}
+	 */
+	function getWrapTarget() {
+    for (var element = $element; element.length; element = element.parent()) {
```
DevVersion: What about using `querySelector` instead of a iteration?

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -207,6 +207,17 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+      if (angular.isDefined(element.attr('md-autocomplete-wrap-override'))) return element[ 0 ];
```
DevVersion: The attribute name, should be documented when we publish that feature.

DevVersion:
@aperl Thanks for the PR, but please provide more details in your Pull Request description. For example, what's the use-cases etc.

src/components/autocomplete/js/autocompleteController.js
aperl: The `warp` is used to calculate the width of the menu. I"ll update the description accordingly.

src/components/autocomplete/js/autocompleteController.js
aperl: `querySelector` regretfully won't allow me to search just the ancestors. This was modeled after the `getSnapTarget ()` in this same directive.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -208,6 +208,17 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   }
 
   /**
+   * Find the target for the wrap element which determines menu width
```
DevVersion: I'm still not happy with the function description. The wrap is not specializing on determining the menu width.

aperl:
@DevVersion I've updated the initial comment to add use case and further description. I'll look now for the appropriate place to add this to documentation.

src/components/autocomplete/js/autocompleteController.js
aperl: while the name `warp` describes what the element is, as far as the controller is concerned it only uses the wrap to calculate the width which can be found in `positionDropdown ()`. I could change the comment to something like "Find the target for the wrap element." But I thought more description help convey why this code is there. Again the description is similar to that of `getSnapTarget()`.

topherfangio:
I like the principle of this PR, but it definitely needs some tests. I am rerunning the Travis build to see if it passes existing tests; something threw off the original build.

aperl:
@topherfangio, what sort of test would we be looking for?

topherfangio:
Regarding tests; I think it could be relatively simple. Just test both cases:

1. When not using a wrapper, the autocomplete's width is the same as the `md-autocomplete` component's width
2. When using a wrapper, it should be the same as the wrapper

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -208,6 +208,17 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+   * @returns {*}
+   */
+  function getWrapTarget() {
+    for (var element = $element; element.length; element = element.parent()) {
```
topherfangio: I'm wondering if there is a better way to accomplish this using a directive. Perhaps the `md-autocomplete-wrap-override` could be a full directive that searches for child autocomplete's (perhaps using some combination of `.querySelector()` and `.controller()`) and register with them as the wrapper so that we don't have to traverse the entire tree of parents all the way back up the DOM.

@aperl You think you could give this a shot?

aperl: @topherfangio It would be that hard. I"ll hop on it.

aperl: @topherfangio So the `md-autocomplete-wrap-override` seemed like an easy way to do it but then it got fragile when I looked at a situation where the `md-autocomplete` element resided inside and `ng-if`. In this case the `md-autocomplete-wrap-override` can't find the `md-autocomplete` controller to register to.

aperl: I suspect it is for similar reasons `md-autocomplete-snap` is found by walking up the DOM.

topherfangio: Hmm, good point; okay, let me think about it a bit more. Please ping me early next week if I forget to respond though :smile:

aperl: Another way to go would be to piggy back on the `md-autocomplete-snap` attribute and give it an optional value `md-autocomplete-snap="width"`. This would avoid adding another attribute and then we could only have to walk up the DOM once.

topherfangio: That's not a bad idea; although I am a bit confused as to how this doesn't already affect the width of the item. Are positioning and width somehow separate?

(I didn't actually know `md-autocomplete-snap` existed since it's not in the docs)

aperl: For whatever reason the `md-autocomplete-snap` was only made to effect the vertical position.

topherfangio: Okay, if it's relatively easy to do; why don't you go ahead with the `md-autocomplete-snap="width"` approach. That will give us the most flexibility.

If you can also add some docs to the Autocomplete about it's usage, that would be great!

aperl: I'll update the PR

aperl:
@topherfangio where would I go to add documentation about `md-autocomplete-wrap`.

topherfangio:
@aperl I would just add it to the existing autocomplete docs. They can be found at the top of `autocompleteDirective.js` and I would probably add a short paragraph right before the "Validation" section.

aperl:
@topherfangio the documentation changes are in.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -193,29 +193,48 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+
+    if (element.length) {
+      var snapWidth = value.toLowerCase() === 'width';
+      return {
+        snap: element[0],
+        wrap: snapWidth ? element[0] : $element.find('md-autocomplete-wrap')[0]
+      };
+    }
+
+    var wrap = $element.find('md-autocomplete-wrap')[0];
```
topherfangio: You could move this above the `if` on line 225 so that you could reuse it on line 229.

aperl: done

aperl:
@topherfangio I've added the tests

topherfangio:
@ThomasBurleson LGTM 👍

aperl:
@ThomasBurleson, I was just wondering why the the request was moved from 1.1.1 to the backlog.

Undrium:
+1 Awesome feature.

ThomasBurleson:

ThomasBurleson:

topherfangio:
Reopening since this was marked as `needs: merge`. Will put through the pre-submit process. If it passes, it should be good to go in, but if not, it may need additional work.

topherfangio:
@aperl It looks like there are a few conflicts now. Is there any chance you could rebase soon? We're hoping to build a 1.1.2 next week, so if we can get this done this week, that would be great!

aperl:
@topherfangio the conflicts were simple. The PR has been rebased.

topherfangio:
@aperl Thanks so much for the super-fast turn around!

aperl:
@topherfangio is there anything else I need to do?

topherfangio:
@aperl Nope. It's labelled as presubmit, so it will go through Google's presubmit process where they run it against internal tests. If all the tests pass, then it should get immediately merged. If they fail, we should get some more info on the PR with the failure so we can investigate.

If it happens to break a TON of tests, it's possible that it may get rejected (with a note), but since this is a small addition/non-breaking change, I wouldn't expect that.

aperl:
@topherfangio thanks for the info

jelbourn:
@aperl can you rebase?

aperl:
@jelbourn Rebase done.

barryvdh:
This seems to work very well like this:
```
 <md-chips md-autocomplete-snap="width" style="width: 100%">
            <md-autocomplete ..
```

But without both the `="width" ` and the style it doesn't work, is this intentional? I would have guessed `md-autocomplete-snap` would just make the chips full width already. But perhaps I'm not understanding correctly.

>  We're hoping to build a 1.1.2 next week, so if we can get this done this week, that would be great!

Would be nice to have this in a release indeed :)

aperl:
@barryvdh md-autocomplete-snap was already a feature, though perhaps not well documented. It would cause the drop down to snap the the specified element. For reverse compatibly it was decided to add the `="width"` value to force it to snap to the width as well.

PR #7754, opened 2016-03-25
update(panel): Adds relative positioning config.

### LINKED ISSUE #7466: mdPanel: Implement PanelXPosition and PanelYPosition.
See panel documentation for a description of each of these values:

xPosition = center | align-left | align-right | align-start | align-end | offset-left | offset-right | offset-start | offset-end

yPosition = center | align-tops | align-bottoms | above | below

-------------------------------------------------------------------------------------------------------------------------
This task involves:
* CSS for the 14 values above (using Math from the relative element to get the correct left/right/top/bottom values)
* Methods for setting/getting those values
* Updating panel positioning method to include relative positioning.

ErinCoughlan:
@jelbourn @DerekLouie @gmoothart @KarenParker Please review.

Fixes #7466

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -323,8 +323,9 @@ angular
 
 
 /**
- * @ngdoc object
+ * @ngdoc type
  * @name MdPanelPosition
+ * @module material.components.panel
  * @description
  * Object for configuring the position of the panel. Examples:
  *
@@ -359,8 +360,8 @@ angular
  * @name MdPanelPosition#relativeTo
  * @description
  * Positions the panel relative to a specific element.
- * @param {!angular.JQLite} element Element to position the panel with
- *     respect to.
+ * @param {string|!Element|!angular.JQLite} element Query selector,
+ *     DOM element, or angular element to position the panel with respect to.
  * @returns {MdPanelPosition}
  */
 
@@ -1015,16 +1016,11 @@ MdPanelRef.prototype._addStyles = function() {
 
   if (!positionConfig) { return; }
 
-  var isAbsolute = positionConfig.getAbsolute();
-  var panelPosition = isAbsolute ? 'fixed' : 'relative';
-  this._panelEl.css('position', panelPosition);
-
-  if (isAbsolute) {
-    this._panelEl.css('top', positionConfig.getTop());
-    this._panelEl.css('bottom', positionConfig.getBottom());
-    this._panelEl.css('left', positionConfig.getLeft());
-    this._panelEl.css('right', positionConfig.getRight());
-  }
+  this._panelEl.css('position', 'fixed');
+  this._panelEl.css('top', positionConfig.getTop(this._panelEl));
+  this._panelEl.css('bottom', positionConfig.getBottom(this._panelEl));
+  this._panelEl.css('left', positionConfig.getLeft(this._panelEl));
+  this._panelEl.css('right', positionConfig.getRight(this._panelEl));
 };
 
 
@@ -1182,6 +1178,12 @@ function MdPanelPosition() {
   /** @private {boolean} */
   this._absolute = false;
 
+  /** @private {!DOMRect} */
+  this._relativeToRect;
+
+  /** @private {boolean} */
+  this._panelPositionCalculated = false;
+
   /** @private {string} */
   this._top = '';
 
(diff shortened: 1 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -1101,69 +1129,263 @@ MdPanelPosition.prototype.top = function(opt_top) {
...
+      this._xPosition = xPosition;
+      validPosition = true;
+      break;
+    }
+  }
+
+  if (!validPosition) {
+    throw new Error('withPanelXPosition only accepts the following values:\n' +
+        'center | align-left | align-right | align-start | align-end | ' +
+        'offset-left | offset-right | offset-start | offset-end');
```
DevVersion: Isn't that a bit hard coded? What do you think about generating those possible values from the array?

src/components/panel/panel.js
```diff
@@ -1056,6 +1051,12 @@ function MdPanelPosition() {
   /** @private {boolean} */
   this._absolute = false;
 
+  /** @private {!Object} */
```
jelbourn: Can we tighten-up the type on this?

src/components/panel/panel.js
```diff
@@ -1067,25 +1068,52 @@ function MdPanelPosition() {
...
+  CENTER: 'center',
+  ALIGN_LEFT: 'align-left',
+  ALIGN_RIGHT: 'align-right',
+  ALIGN_START: 'align-start',
+  ALIGN_END: 'align-end',
+  OFFSET_LEFT: 'offset-left',
+  OFFSET_RIGHT: 'offset-right',
+  OFFSET_START: 'offset-start',
+  OFFSET_END: 'offset-end'
+};
```
jelbourn: Why have the items that are `left` and `right` at all as opposed to just `start` and `end` ?

src/components/panel/panel.js
```diff
@@ -1101,69 +1129,263 @@ MdPanelPosition.prototype.top = function(opt_top) {
...
-MdPanelPosition.prototype.getLeft = function() {
-  return this._left;
+MdPanelPosition.prototype.withPanelXPosition = function(xPosition) {
+  if (!this._relativeToRect) {
+    throw new Error('withPanelXPosition can only be used with relative' +
+        'positioning. Set relativeTo first.');
+  }
+
+  var validPosition = false;
+  for (var key in MdPanelPosition.xPosition) {
```
jelbourn: Should do
```ts
if (MdPanelPosition.xPosition[key])
```
It's generally considered an anti-pattern to use the `in` operator for checking if an object has a property.

src/components/panel/panel.js
```diff
@@ -1101,69 +1129,263 @@ MdPanelPosition.prototype.top = function(opt_top) {
...
+  if (this._absolute || this._panelPositionCalculated) {
+    return;
+  }
+
+  // TODO(ErinCoughlan): Update position on scroll.
+
+  // Indicate that the position is calculated so it can be skipped next time.
+  this._panelPositionCalculated = true;
+
+  var panelBounds = getBoundingClientRect(panelEl).bounds;
```
ErinCoughlan: Rewrite this. We know it's an element.

`panelEl[0].getBoundingClientRect()`

src/components/panel/panel.js
jelbourn: Err, ignore that last snippet, I wasn't reading carefully. You should iterate over `Object.keys`

src/components/panel/panel.js
```diff
@@ -1101,69 +1129,263 @@ MdPanelPosition.prototype.top = function(opt_top) {
...
+ * panel element.
+ * @param {!angular.JQLite} panelEl
+ */
+MdPanelPosition.prototype._calculatePanelPosition = function(panelEl) {
+  // Only calculate the position if necessary.
+  if (this._absolute || this._panelPositionCalculated) {
+    return;
+  }
+
+  // TODO(ErinCoughlan): Update position on scroll.
```
jelbourn: Add TODO for keeping the pane on-screen?

src/components/panel/panel.js
```diff
@@ -1101,69 +1129,263 @@ MdPanelPosition.prototype.top = function(opt_top) {
...
   return this._right;
 };
+
+
+/**
+ * Calculates the panel position for relative based on the created
+ * panel element.
+ * @param {!angular.JQLite} panelEl
+ */
+MdPanelPosition.prototype._calculatePanelPosition = function(panelEl) {
```
jelbourn: FYI: this is a good starting point, but the positioning can get pretty complicated. I'd suggest checkout out the function `attachCalendarPane` in the datepicker to see some of the issue I've run into with positioning the panel next to the trigger.

You may also want to consider: instead of setting one position (like centered-beneath), setting a prioritized list of preferred positions. That way, if the desired position would go off-screen, you can do a more intelligent fallback (eventually falling back to the "this doesn't fit anywhere" case).

ErinCoughlan: Sounds good. I knew it could get a lot more complicated from the menu positioning code. I didn't want to tackle it in this initial PR though. The prioritized list of positions is an interesting idea. I'll consider it in a future PR.

ErinCoughlan:
I've addressed all comments. PTAL.

jelbourn:
LGTM

gmoothart:
:+1:

DerekLouie:
LGTM

PR #7758, opened 2016-03-26
fix(gridList): grid tiles have exceeded the max columns.

### LINKED ISSUE #7190: md-grid-list first render is broken with limited width
I've been playing around with the md-grid-list component. It usually works fine, but when the width of the screen is small, the first rendering is broken. If you resize the window and then get it back at the original width, it scales well, but when you refresh the page, the rendering is broken again. 

You can see it in the [gridList demo](https://material.angularjs.org/latest/demo/gridList)

You can see the (not very well done) video here: https://goo.gl/Um7ZL6

DevVersion:
Fixes #7190


topherfangio:
LGTM! Thanks!

PR #7763, opened 2016-03-26
feat(tabs): tabs + toolbar shrinking behaviour

corentin-gautier:
also fixes pagination color issue & add md-button-offset for toolbar with nav button

tabs


src/components/tabs/js/tabsController.js
```diff
@@ -97,6 +99,92 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+      bindScrollListener(ctrl.selectedIndex);
+    }
+  }
+
+  function bindScrollListener (newIndex, oldIndex) {
+    if (!DOMtabs.length) return;
+    if (oldIndex !== undefined && ctrl.tabs[oldIndex]) {
+      angular.element(DOMtabs[oldIndex]).off('scroll');
+      $timeout(function () {
+        scrollTop(oldIndex);
```
crisbeto: Does this `$timeout` need to do a digest? Otherwise you can skip the digest by passing `false` as the third argument.

src/components/tabs/js/tabsController.js
```diff
@@ -97,6 +99,92 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+    DOMtabs = $element.find('md-tab-content');
+    if (ctrl.shrinkToolbar) {
+      bindScrollListener(ctrl.selectedIndex);
+    }
+  }
+
+  function bindScrollListener (newIndex, oldIndex) {
+    if (!DOMtabs.length) return;
+    if (oldIndex !== undefined && ctrl.tabs[oldIndex]) {
+      angular.element(DOMtabs[oldIndex]).off('scroll');
```
crisbeto: You're unbinding all scroll listeners here, including anything that the user might have added.  Try unbinding only a method that you have a reference to.

src/components/tabs/js/tabsController.js
```diff
@@ -97,6 +99,92 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+      $timeout(function () {
+        scrollTop(oldIndex);
+      }, 1000);
+    }
+
+    prevScrollTop = 0;
+    angular.element(DOMtabs[newIndex || 0])
+      .off('scroll')
```
crisbeto: You're also unbinding all listeners here.

src/components/tabs/js/tabsController.js
```diff
@@ -97,6 +99,92 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+      $element.css({
+        "margin-top": - margin + 'px',
+        "margin-bottom": ((toolbarHeight - y) * shrinkSpeedFactor) + 'px'
+      });
+
+      if (contentScroll >= toolbarHeight - y ) {
+        angular.element(elements.wrapper).addClass(ZCLASS);
+      } else {
+        angular.element(elements.wrapper).removeClass(ZCLASS);
+      }
```
crisbeto: This whole expression is the same as `angular.element(elements.wrapper).toggleClass(ZCLASS, contentScroll >= toolbarHeight - y);`

src/components/tabs/js/tabsController.js
```diff
@@ -145,6 +233,9 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
   function cleanup () {
     destroyed = true;
     angular.element($window).off('resize', handleWindowResize);
+    if (ctrl.shrinkToolbar) {
+      angular.element(DOMtabs[ctrl.selectedIndex]).off('scroll');
```
crisbeto: Again removing all of the scroll listeners.

src/components/tabs/js/tabsController.js
```diff
@@ -219,17 +310,24 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+    if (ctrl.shrinkToolbar) {
+      bindScrollListener(newValue, oldValue);
+    }
     $scope.$broadcast('$mdTabsChanged');
     ctrl.tabs[ oldValue ] && ctrl.tabs[ oldValue ].scope.deselect();
     ctrl.tabs[ newValue ] && ctrl.tabs[ newValue ].scope.select();
   }
 
+  function scrollTop (index) {
+    document.querySelector('#tab-content-' + ctrl.tabs[index].id).scrollTop = 0;
```
crisbeto: This is the same as `document.getElementById('tab-content-' + ctrl.tabs[index].id).scrollTop = 0;`.

src/components/tabs/tabs.scss
```diff
@@ -178,6 +184,7 @@ md-tabs-content-wrapper {
   right: 0;
   bottom: 0;
   overflow: hidden;
+  z-index: 1;
```
DevVersion: Our z-indexes are always defined in the `variables.scss` file. Please take a look at those. We are having a strict use-case of z-indexes. Applying random z-indexes may break some other components.

src/components/tabs/js/tabsController.js
```diff
@@ -97,6 +99,92 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+  var shrinkSpeedFactor = .5;
+  var contentScroll = 0;
+  var toolbar;
+  var toolbarHeight = 0;
+  var y = 0;
+  var ZCLASS = 'md-whiteframe-z1';
+
+  function shrinkToolbar (e) {
+    if (!toolbar) {
+      elements.toolbar = angular.element($element.parent()[0].querySelector('md-toolbar'));
```
DevVersion: I'm curious, the toolbar itself already has a functionality like that. What is the point of creating a dupe of this functionality?

Can you not apply the `scroll-shrink` attribute on the toolbar?

corentin-gautier: @DevVersion the current "toolbar shrink" only listens for scroll on a parent md-content which does not apply when using tabs. The shrink behavior is actually very linked to the tabs themselves (keeping the scroll when changing tab, only listening to the current tab scroll etc.) But if you think it can be done in the toolbar component then feel free to use this PR as inspiration only :)

DevVersion: @corentin-gautier And that's what I've said. 

gif

I've transformed the toolbar scroll shink functionality, to accept a target element. And as you can see in the GIF it works like a charm. (Took me about 7lines)

So now if I continue transforming, to provide a controller API, then I'm able to transform the tabs header as well.

corentin-gautier: Let's do it this way then :)

DevVersion: @corentin-gautier Awesome, please remember to keep the old changes (just for the future). 

I suggest, creating a controller for the sidenav (which is not documented, so only for internal use).

corentin-gautier: For the sidenav ?

DevVersion: @corentin-gautier Sorry, definitely meant toolbar. How embarrassing :)

corentin-gautier: @DevVersion ahah, it's ok :) Do you want to do it or do you want me to do it ?

DevVersion: @corentin-gautier Of course, you can do that :)

corentin-gautier: @DevVersion can you push your code or just show me here what you've done ?

DevVersion: @corentin-gautier Already discarded those changes, ping me on Gitter, then we can chat there.

corentin-gautier: @DevVersion newbie alert here :) I don't know what that is ! Investigating now ...

corentin-gautier:
@DevVersion please review the last commit ( also tests are not passing anymore and I don't quite understand how they work, would appreciate some help :) )

src/components/tabs/js/tabsController.js
```diff
@@ -85,6 +93,39 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+
+  function getToolbar () {
+    var toolbar = angular.element($element.parent()[0].querySelector('md-toolbar'));
+
+    if (!toolbar.length) return angular.noop;
+
+    elements.toolbar = toolbar;
+
+    if (!toolbarController) {
+      toolbarController = elements.toolbar.scope().$mdToolbarCtrl;
```
DevVersion: `#scope()` is only available in debug mode. Let's use 

```js
elements.toolbar.controller('mdToolbar');
```

src/components/toolbar/js/toolbarDirective.js
```diff
@@ -0,0 +1,72 @@
...
+
+function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
+
+  return {
+    template: '',
+
+    restrict: 'E',
+
+    controller: 'MdToolbarController',
+    controllerAs: '$mdToolbarCtrl',
```
DevVersion: If you use the approach from above, with `controller('mdToolbar');` then you can drop this line.

src/components/tabs/demoToolbarTabs/readme.html
```diff
@@ -0,0 +1 @@
+<p>The Toolbar Tabs demo shows how toolbar can be used with tabs</p>
```
DevVersion: This `readme.html` file is not needed. Would be great if you remove this file :)

src/components/toolbar/js/toolbarDirective.js
corentin-gautier: @DevVersion the controllerAs ?

DevVersion: yes

corentin-gautier:
@DevVersion please review ( again :) )

src/components/tabs/js/tabsController.js
```diff
@@ -85,6 +93,36 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+        loadTabElements();
+        getToolbar();
+      }
+    });
+  }
+
+  function getToolbar () {
+    var toolbar = angular.element($element.parent()[0].querySelector('md-toolbar'));
+
+    if (!toolbar.length) return angular.noop;
```
DevVersion: Just returning should be enough here. `return;`

src/components/tabs/js/tabsDirective.js
```diff
@@ -64,6 +64,7 @@
  * @param {boolean=} md-enable-disconnect When enabled, scopes will be disconnected for tabs that are not being displayed.  This provides a performance boost, but may also cause unexpected issues and is not recommended for most users.
  * @param {boolean=} md-autoselect When present, any tabs added after the initial load will be automatically selected
  * @param {boolean=} md-no-select-click When enabled, click events will not be fired when selecting tabs
+ * @param {boolean=} md-button-offset If present, add an offset to the tabs (when using a toolbar with a button)
```
DevVersion: I think this needs a bit more information. As a user you won't get the actual behavior of that attribute.

corentin-gautier: @DevVersion Open to suggestions here ;)

ThomasBurleson:
@corentin-gautier - please squash your commits and rebase from master.

src/components/toolbar/js/toolbarController.js
```diff
@@ -0,0 +1,206 @@
+angular
+  .module('material.components.toolbar')
+  .controller('MdToolbarController', MdToolbarController);
```
DevVersion: Let's prefix the controller with `$mdToolbarController`, to avoid complications and to be consistent with other internal controllers.

corentin-gautier: @DevVersion I should do that on the tabsController as well then.

src/components/toolbar/js/toolbarController.js
```diff
@@ -0,0 +1,206 @@
...
+    // When using ng hide there's a delay caused by ngAnimate
+    var isHidden = false;
+
+    if ($attrs.hasOwnProperty('ngShow')) {
+      isHidden = !$scope.$eval($attrs.ngShow);
+    }
+
+    if ($attrs.hasOwnProperty('ngHide')) {
+      isHidden = $scope.$eval($attrs.ngHide);
+    }
```
DevVersion: Can we not skip that checks above? It would make much more sense, to check the DOM visibility. 

Like I did in another PR: [Here](https://github.com/DevVersion/material/blob/feat/sidenav-track-visibility/src/core/util/util.js#L735-L743) (Notice: This is not in master)

src/components/toolbar/js/toolbarDirective.js
```diff
@@ -0,0 +1,71 @@
...
+
+angular.module('material.components.toolbar')
+  .directive('mdToolbar', mdToolbarDirective);
+
+function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
+
+  return {
+    template: '',
+
+    restrict: 'E',
```
DevVersion: Please remove those empty lines between the JSON object.

src/components/toolbar/toolbar.spec.js
```diff
@@ -15,74 +15,65 @@ describe('<md-toolbar>', function() {
...
     $$rAF.flush();
 
-    expect(toolbarCss[$mdConstant.CSS.TRANSFORM]).toEqual('translate3d(0,0px,0)');
-    expect(contentCss[$mdConstant.CSS.TRANSFORM]).toEqual('translate3d(0,100px,0)');
+    toolbarStyles = getComputedStyle(element.find('md-toolbar')[0]);
+    contentStyles = getComputedStyle(element.find('md-content')[0]);
+    transformToolbar = getTransform(toolbarStyles);
+    transformContent = getTransform(contentStyles);
+
+    expect(transformToolbar).toEqual('matrix(1, 0, 0, 1, 0, 0)');
```
DevVersion: Please explain, why you use `matrix` here?

src/components/toolbar/js/toolbarController.js
DevVersion: Indeed, then we can keep it. Ignore my comment here :)

src/components/toolbar/toolbar.spec.js
```diff
@@ -291,10 +283,19 @@ describe('<md-toolbar>', function() {
...
     });
   }
+
+  function getTransform (st) {
+    return st.getPropertyValue("-webkit-transform") ||
+           st.getPropertyValue("-moz-transform") ||
+           st.getPropertyValue("-ms-transform") ||
+           st.getPropertyValue("-o-transform") ||
+           st.getPropertyValue("transform") ||
+           'none';
```
DevVersion: This function can be replaced with an util constant.

https://github.com/angular/material/blob/master/src/core/util/constant.js#L39
`$mdConstant.CSS.TRANSFORM`

src/components/toolbar/toolbar.spec.js
corentin-gautier: because that's what the transform property returns

src/components/toolbar/toolbar.spec.js
corentin-gautier: this doesn't work with phantom ... that's why I didn't used it

src/components/toolbar/js/toolbarController.js
corentin-gautier: I don't understand what you want me to do :)

DevVersion: I mean, you could implement that method, and remove all the `ngShow`, `ngHide` checks and just use:
```js
var isHidden = $mdUtil.isHidden($element, true, true)
```

corentin-gautier: @DevVersion oh ok. but I need your PR to be merged before I can change that

corentin-gautier:
@ThomasBurleson squashed and rebased :)

src/components/toolbar/js/toolbarController.js
DevVersion: Yes, it's okay, just copy it from my PR.

src/components/toolbar/toolbar.spec.js
DevVersion: Oh, it should work without matrix as well. Just as before, by using the `$mdConstant` and computed style. Matrix is not that fancy.

corentin-gautier: no I mean thats what the getPropertyValue returns:

image

src/components/toolbar/js/toolbarController.js
corentin-gautier: I just tried it but it doesn't work because, as my comment mentions: there's a delay caused by ngAnimate -> the isHidden function will return false because it takes a few ms for the element to actually hide (because of ngAnimate I believe)

corentin-gautier:
```javascript
if ($attrs.ngHide) { $scope.$watch($attrs.ngHide, determineToolbarHeight); }
```

The solution would be to wait until the element is actually hidden and then to run determineToolbarHeight. I don't know how to do that :/

DevVersion: @corentin-gautier If you use `isHidden()` and ignore the delay by `ngAnimate`, then $$rAF should just later determine the height again, and then `isHidden()` will be true.

src/components/toolbar/toolbar.spec.js
DevVersion: Seems valid, but it's still weird. But I agree, it is much better than the old mock function.

src/components/toolbar/js/toolbarController.js
corentin-gautier: @DevVersion yeah but there's not $$rAF later :)

DevVersion: I know, just searching for possibilities, what about adding 
```js
  if (!toolbarHeight && !isHidden || !isHidden) {
```

corentin-gautier: @DevVersion It'll constantly check for the toolbar height when it's visible and not scroll (because ot the return)

corentin-gautier: @DevVersion Aren't there any event we can use that is triggered after the ngAnimate animation ?

DevVersion: @corentin-gautier Wow, that's also weird. 

What about this approach, when the `$watch` get updated by `ngShow` / `ngHide`, then there will be the parsed new value as an first parameter.

So, you use:
```js
function determineToolbarHeight(watchVal) {
// watchVal is now the ngHide or ngShow value <-- WEIRD
```

So you may have noticed, that those values are inverted for ngHide and ngShow, and you can't differentiate between those.

That's why you should use in the `$watch` caller for `ngShow`:
```js
if ($attrs.ngShow) { $scope.$watch($attrs.ngShow, , function(val) {
  determineToolbarHeight(!val);
})}

```

corentin-gautier: @DevVersion hum yeah it seams to work with

```javascript
function determineToolbarHeight(watchValue) {
  if (!angular.isNumber(watchValue)) {
    isHidden = watchValue;
  }
```

Does that look good to you ?

DevVersion: @corentin-gautier Yes that looks good to me.

Also add an comment on the `isNumber` check, explaining that `$$rAF` is providing a random number.

corentin-gautier: @DevVersion Ok done :)

corentin-gautier:
fixes #1259

ThomasBurleson:
@EladBezalel - can you review ?

ThomasBurleson:
This is a large change set and requires careful review and testing.

EladBezalel:
@ThomasBurleson IMO we should have shrinking toolbars that support md-tabs and not the other way

corentin-gautier:
@EladBezalel that's what I did :)

ThomasBurleson:

ThomasBurleson:

PR #7765, opened 2016-03-27
feat(select): add hover styles and fix disabled

### LINKED ISSUE #7518: select: Options have no hover effect on 1.1.0-rc1
When hovering a select option, there is no longer a hover effect. Likely due to recent privatization changes.

Can be viewed on the select demo page for latest/HEAD.

crisbeto:
* Adds hover styles to the select component.
* Fixes the disabled styles that were being applied to the wrong element.

Closes #7518.


crisbeto:
Tagging this as "needs: merge" as per our Slack discussion a couple of days ago.

PR #7770, opened 2016-03-27
fix(radioButton): replaced inherit margin values with 0

### LINKED ISSUE #7740: Replacing inherited value with 0 for radio button
Hi,

Currently, radio button uses value inherit which is not the proper value I guess and should be replaced by 0. It was done for the checkbox in #6472 by @EladBezalel. I think same can be done for the radio button.

Just for the reference:

md-checkbox: https://github.com/angular/material/blob/master/src/components/checkbox/checkbox.scss#L30

md-radio-buttton:
https://github.com/angular/material/blob/master/src/components/radioButton/radio-button.scss#L133

jayeshanandani:
Inherit is not the proper value, 0 should be used instead
Closes #7740

### DIFF
--- a/src/components/radioButton/radio-button.scss
+++ b/src/components/radioButton/radio-button.scss
@@ -130,12 +130,12 @@ md-radio-group {
     md-radio-button {
       margin-top: 0;
       margin-bottom: 0;
-      @include rtl(margin-left, inherit, $radio-margin);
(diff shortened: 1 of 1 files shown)


DevVersion:
LGTM

EladBezalel:
LGTM

PR #7775, opened 2016-03-27
fix(progress-circular): add the dark theme background in the demo

crisbeto:
Fixes the `progressCircular` demo showing the dark theme colors, but on a light background. Also makes the padding consistent with the other demos.


ThomasBurleson:
@crisbeto - this is VERY nice! :+1:

PR #7778, opened 2016-03-28
fix(select): disabled state, invalid html in unit tests

### LINKED ISSUE #7773: mdSelect: disabled select is set to touched when clicking at it
Disabled select is set to touched when clicking at it and clicking somewhere else on the page. This causes the error message to be shown. This is not consistent with behaviour of disabled mdInput, which is not set to touched when clicking at it.

Demo can be found at: http://codepen.io/anon/pen/MyoVwY.

I can provide a pull request with fixes for this issue, including tests, if this is going to help.

crisbeto:
fix(select): disabled state, invalid html in unit tests
Fixes:
* Disabled `mdSelect` elements being marked as `touched` when they get blurred.
* Click and key events being bound to elements with empty `disabled` attributes.
* Invalid, comma-separated attributes in the `mdSelect` unit tests.

Cheers to @montgomery1944 for the tip about removing the tabindex.
Closes #7773;

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -188,7 +188,9 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
     element.empty().append(valueEl);
     element.append(selectTemplate);
 
-    attr.tabindex = attr.tabindex || '0';
+    if(!attr.tabindex){
+      attr.$set('tabindex', 0);
+    }
 
     return function postLink(scope, element, attr, ctrls) {
       var untouched = true;
@@ -376,22 +378,24 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
         }
         isDisabled = disabled;
         if (disabled) {
-          element.attr({'tabindex': -1, 'aria-disabled': 'true'});
-          element.off('click', openSelect);
-          element.off('keydown', handleKeypress);
+          element
+            .attr({'aria-disabled': 'true'})
+            .removeAttr('tabindex')
+            .off('click', openSelect)
+            .off('keydown', handleKeypress);
         } else {
-          element.attr({'tabindex': attr.tabindex, 'aria-disabled': 'false'});
-          element.on('click', openSelect);
-          element.on('keydown', handleKeypress);
+          element
+            .attr({'tabindex': attr.tabindex, 'aria-disabled': 'false'})
+            .on('click', openSelect)
+            .on('keydown', handleKeypress);
         }
       });
 
-      if (!attr.disabled && !attr.ngDisabled) {
-        element.attr({'tabindex': attr.tabindex, 'aria-disabled': 'false'});
+      if (!attr.hasOwnProperty('disabled') && !attr.hasOwnProperty('ngDisabled')) {
+        element.attr({'aria-disabled': 'false'});
         element.on('click', openSelect);
         element.on('keydown', handleKeypress);
       }
-
 
       var ariaAttrs = {
         role: 'listbox',
@@ -1229,7 +1233,7 @@ function SelectProvider($$interimElementProvider) {
             }
             newOption = optionsArray[index];
             if (newOption.hasAttribute('disabled')) newOption = undefined;
-          } while (!newOption && index < optionsArray.length - 1 && index > 0)
+          } while (!newOption && index < optionsArray.length - 1 && index > 0);
           newOption && newOption.focus();
           opts.focusedNode = newOption;
         }
--- a/src/components/select/select.spec.js
+++ b/src/components/select/select.spec.js
@@ -31,8 +31,13 @@ describe('<md-select>', function() {
   }));
 
   it('should preserve tabindex', function() {
-    var select = setupSelect('tabindex="2", ng-model="val"').find('md-select');
+    var select = setupSelect('tabindex="2" ng-model="val"').find('md-select');
     expect(select.attr('tabindex')).toBe('2');
+  });
+
+  it('should set a tabindex if the element does not have one', function() {
+    var select = setupSelect('ng-model="val"').find('md-select');
+    expect(select.attr('tabindex')).toBeDefined();
   });
 
   it('supports non-disabled state', function() {
@@ -41,7 +46,7 @@ describe('<md-select>', function() {
   });
 
   it('supports disabled state', inject(function($document) {
-    var select = setupSelect('disabled="disabled", ng-model="val"').find('md-select');
+    var select = setupSelect('disabled ng-model="val"').find('md-select');
     openSelect(select);
     expectSelectClosed(select);
     expect($document.find('md-select-menu').length).toBe(0);
@@ -49,7 +54,7 @@ describe('<md-select>', function() {
   }));
 
   it('supports passing classes to the container', inject(function($document) {
-    var select = setupSelect('ng-model="val", md-container-class="test"').find('md-select');
+    var select = setupSelect('ng-model="val" md-container-class="test"').find('md-select');
     openSelect(select);
 
     var container = $document[0].querySelector('._md-select-menu-container');
@@ -58,7 +63,7 @@ describe('<md-select>', function() {
   }));
 
   it('supports passing classes to the container using `data-` attribute prefix', inject(function($document) {
-    var select = setupSelect('ng-model="val", data-md-container-class="test"').find('md-select');
+    var select = setupSelect('ng-model="val" data-md-container-class="test"').find('md-select');
     openSelect(select);
 
     var container = $document[0].querySelector('._md-select-menu-container');
@@ -67,7 +72,7 @@ describe('<md-select>', function() {
   }));
 
   it('supports passing classes to the container using `x-` attribute prefix', inject(function($document) {
-    var select = setupSelect('ng-model="val", x-md-container-class="test"').find('md-select');
+    var select = setupSelect('ng-model="val" x-md-container-class="test"').find('md-select');
     openSelect(select);
 
     var container = $document[0].querySelector('._md-select-menu-container');
@@ -77,7 +82,7 @@ describe('<md-select>', function() {
 
   it('sets aria-owns between the select and the container', function() {
     var select = setupSelect('ng-model="val"').find('md-select');
-    var ownsId = select.attr('aria-owns'); 
+    var ownsId = select.attr('aria-owns');
     expect(ownsId).toBeTruthy();
     var containerId = select[0].querySelector('._md-select-menu-container').getAttribute('id');
     expect(ownsId).toBe(containerId);
@@ -88,7 +93,7 @@ describe('<md-select>', function() {
     $rootScope.onClose = function() {
       called = true;
     };
-    var select = setupSelect('ng-model="val", md-on-close="onClose()"', [1, 2, 3]).find('md-select');
+    var select = setupSelect('ng-model="val" md-on-close="onClose()"', [1, 2, 3]).find('md-select');
     openSelect(select);
     expectSelectOpen(select);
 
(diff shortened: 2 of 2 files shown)


DevVersion:
LGTM. 

FYI: @crisbeto and I had a discussion about the `blur` handler, to stop propagation, and we both agreed, that this is the only way to probably prevent the `$touched` change.

Also notice, that native select elements are not setting `$touched` as well.

montgomery1944:
@DevVersion, this is not the only solution to prevent from setting untouched. I've already implemented touched setting prevention in different case using $evalAsync in https://github.com/angular/material/pull/7646. This solution does not have the downside of preventing all blur event handlers from firing. I think that for both cases the same implementation should be used to be consistent.

crisbeto:
@montgomery1944 I was also considering something similar (re-using the event handler and removing the `nextTick` wrapper), but in that case it's being set as touched, then back as untouched. That might still cause something to flicker or some other unexpected behavior.

montgomery1944:
@crisbeto, because of how $evalAsync works, it is not going to cause any flicker. Async evals are run just after the regular eval, where the touched is set, and just before the digest loop, so in the digest all places in the code using touched property will see false value. As I said, consistency is important, so if your solution will be the preffered one, I will have to change my pull request.

crisbeto:
Alright, let me try your solution and see how it goes. We might have to work out a way of combining these since whatever gets merged first will definitely cause a conflict in the other.

crisbeto:
@montgomery1944 Just tried out your solution with `$evalAsync`. It works in blocking the first `blur`, but it doesn't work on the next. Here's what I mean:

blur

montgomery1944:
@crisbeto, works for me, just don't put your code inside "if (untouched) {" condition, it is needed only in my case and this is why your prevention works only once. The exact code working for me:
`
element.on('blur', function() {
  if (untouched) {
    untouched = false;
    if (selectScope.isOpen) {
      scope.$evalAsync(function() {
        ngModelCtrl.$setUntouched();
      });
    }
  }

  if (element[0].hasAttribute('disabled')) {
    scope.$evalAsync(function() {
      ngModelCtrl.$setUntouched();
    });
  }

  if (selectScope.isOpen) return;
  containerCtrl && containerCtrl.setFocused(false);
  inputCheckValue();
});
`

crisbeto:
Alright, that works but we can reduce it to something like:

```javascript
element.on('blur', function() {
  if (element[0].hasAttribute('disabled') || untouched && selectScope.isOpen) {
    scope.$evalAsync(function() {
      ngModelCtrl.$setUntouched();
    });
  }

  if (untouched) {
    untouched = false;
  }

  if (selectScope.isOpen) return;
  containerCtrl && containerCtrl.setFocused(false);
  inputCheckValue();
});
```

How should we go about this?

montgomery1944:
@crisbeto, I think that there is a better solution to this issue. The root cause of the problem is the tabindex with -1 value on mdSelect element, which is removing the element from tabbing order, but it makes the element focusable by clicking. By default, mdSelect element is not focusable, so I think that changing "element.attr({'tabindex': -1, 'aria-disabled': 'true'});" to "element.attr({'aria-disabled': 'true'});" will fix the issue without the need for setting untouched.

crisbeto:
Good point, that works even better! I suppose instead we should remove the tabindex altogether when the element becomes disabled.

crisbeto:
So I think we should combine our fixes into 1 PR. What do you think @montgomery1944?

montgomery1944:
@crisbeto, if you are going to fix this issue by removing tabindex, then there is no conflict between my and your pull request. Still, if you want to combine those fixes in one pull request, feel free to include changes from my pull request in yours.

crisbeto:
Alright, I'll include the tabindex stuff in my PR. I'll also try to see if I can do anything about the events not being bound if the select goes from disabled to not disabled.

src/components/select/select.js
```diff
@@ -288,19 +288,16 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
...
-            if (untouched) {
-              untouched = false;
-              ngModelCtrl.$setUntouched();
-            }
 
-            if (selectScope.isOpen) return;
-            containerCtrl && containerCtrl.setFocused(false);
-            inputCheckValue();
-          });
+        element.on('blur', function() {
```
montgomery1944: After removal of tabindex this change is not needed, because non focusable element won't get a blur event.

src/components/select/select.js
```diff
@@ -370,29 +367,32 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
...
+            .attr({'aria-disabled': 'true'})
+            .removeAttr('tabindex')
+            .off('click', openSelect)
+            .off('keydown', handleKeypress);
         } else {
-          element.attr({'tabindex': attr.tabindex, 'aria-disabled': 'false'});
-          element.on('click', openSelect);
-          element.on('keydown', handleKeypress);
+          element
+            .attr({'tabindex': attr.tabindex || 0, 'aria-disabled': 'false'})
```
montgomery1944: Similar code is run in line 191: "attr.tabindex = attr.tabindex || '0';". Isn't it redundant?

src/components/select/select.js
```diff
@@ -370,29 +367,32 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
...
           return;
         }
         isDisabled = disabled;
         if (disabled) {
-          element.attr({'tabindex': -1, 'aria-disabled': 'true'});
-          element.off('click', openSelect);
-          element.off('keydown', handleKeypress);
+          element
+            .attr({'aria-disabled': 'true'})
+            .removeAttr('tabindex')
```
montgomery1944: Are we sure, that the template will always be cloned, so that removal of tabindex here is not going to remove "attr.tabindex" from the template? I think that we mustn't lose the original tabindex value set by user.

src/components/select/select.js
```diff
@@ -370,29 +367,32 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
...
+            .attr({'tabindex': attr.tabindex || 0, 'aria-disabled': 'false'})
+            .on('click', openSelect)
+            .on('keydown', handleKeypress);
         }
       });
 
-      if (!attr.disabled && !attr.ngDisabled) {
-        element.attr({'tabindex': attr.tabindex, 'aria-disabled': 'false'});
+      if (!attr.hasOwnProperty('disabled') && !attr.hasOwnProperty('ngDisabled')) {
+        element.attr({'aria-disabled': 'false'});
```
montgomery1944: Why setting "attr.tabindex" is no longer needed here?

src/components/select/select.spec.js
```diff
@@ -153,6 +153,16 @@ describe('<md-select>', function() {
     expect($rootScope.myForm.select.$touched).toBe(true);
   }));
 
+  it('should not set touched if the element is disabled', inject(function($compile, $rootScope) {
```
montgomery1944: I think that this test is not testing the primary change in this PR. In the test blur event is triggered, while element without tabindex is not focusable, so it won't get blur event at all.

src/components/select/select.js
crisbeto: Yes, but I believe that this is necessary if you're switching between an enabled and disabled element. It broke a few unit tests if I remove it.

src/components/select/select.js
crisbeto: Angular doesn't keep track of the attributes, `attr` keeps the original value that was set.

src/components/select/select.js
crisbeto: This one restores the tabindex if it was removed while the select was disabled. As for the `|| 0`, it's to make sure that the element is focusable afterwards.

src/components/select/select.js
crisbeto: Because this is only being run on load. The element already has the tabindex.

src/components/select/select.spec.js
crisbeto: Good point, this one is useless after the latest changes. It probably makes more sense to test whether the tabindex gets removed.

src/components/select/select.js
montgomery1944: Weird, I've checked out your branch, reverted this change, while keeping the rest of the changes, and all tests are passing. Futhermore, I am still not sure how this is going to work, when we are not going to get any blur events without tabindex.

src/components/select/select.js
montgomery1944: OK, but because in line 191 we are setting a fallback value, we don't need to have "|| 0" in here, because "attr.tabindex" will be always truthy.

src/components/select/select.js
montgomery1944: True. :+1:

src/components/select/select.js
crisbeto: Commenting out that whole handler makes these fail:
* `<md-select> should set touched only after closing`.
*  ` <md-select> should remove the input-container focus state FAILED`.

As for the event not firing, my point is that the disabled attribute can be removed at a later point and it'll need to fire then.

src/components/select/select.js
crisbeto: Alright, that makes sense.

src/components/select/select.js
montgomery1944: This line is indeed needed, because user can create mdSelect without any "tabindex". In such case, mdSelect won't get any tabindex and won't be focusable.

crisbeto: But if the element doesn't have a tabindex, this won't set one either since the value in `attr` will be undefined. Also as you mentioned on another diff, the tabindex is being set on line 191 anyway.

src/components/select/select.js
montgomery1944: Yes, commenting the whole handler will indeed make those tests fail. I meant, that the change to $evalAsync is not needed. In later point of time, when user enables disabled mdSelect, everything should work as before. Problems with blur on mdSelect will be fixed in https://github.com/angular/material/pull/7646. If you want to include those changes in your PR, as we talked about, then please include it as it is there, including the test and condition on "isOpen".

src/components/select/select.js
crisbeto: Oh I just looked again, I didn't notice that it was being set on the attr object and not the element itself. I'd say it makes more sense to do it with `attr.$set` which will set it both on the object and the element.

src/components/select/select.js
crisbeto: Removing the `$evalAsync` makes the `<md-select> should set touched only after closing` fail.

montgomery1944: Yes, if you're going to remove the whole line: "scope.$evalAsync(ngModelCtrl.$setUntouched);", then yes, it's going to fail. What I meant, is that the whole change done to those lines are not needed for the fix based on tabindex to work. Furthermore, as I said, those changes are not completely correct, because they are missing the "isOpen" condition, which is already included in https://github.com/angular/material/pull/7646. If you'are going to change those lines to the previous state:
```
        // Wait until postDigest so that we attach after ngModel's
        // blur listener so we can set untouched.
          element.on('blur', function() {
            if (untouched) {
              untouched = false;
              ngModelCtrl.$setUntouched();
            }

            if (selectScope.isOpen) return;
            containerCtrl && containerCtrl.setFocused(false);
            inputCheckValue();
          });
        });
```
then no test is going to fail.

crisbeto: Got it. I've reverted those changes from my PR and tried it out to make sure everything passes and it still works.

ThomasBurleson:
@crisbeto - CI is failing.

crisbeto:
@ThomasBurleson seems like a Node error in Travis, they ran fine locally:

```
npm ERR! network read ECONNRESET
npm ERR! network This is most likely not a problem with npm itself
npm ERR! network and is related to network connectivity.
npm ERR! network In most cases you are behind a proxy or have bad network settings.
npm ERR! network 
npm ERR! network If you are behind a proxy, please make sure that the
npm ERR! network 'proxy' config is set properly.  See: 'npm help config'
npm ERR! Linux 3.13.0-40-generic
npm ERR! argv "/home/travis/.nvm/versions/node/v4.2.3/bin/node" "/home/travis/.nvm/versions/node/v4.2.3/bin/npm" "install"
npm ERR! node v4.2.3
npm ERR! npm  v2.14.7
npm ERR! code ECONNRESET
npm ERR! errno ECONNRESET
npm ERR! syscall read
```

PR #7781, opened 2016-03-28
feat(navBar): add mdNavBar component for page navigation

### LINKED ISSUE #6676: md-tabs can't be used for page navigation
The tab component does not work well as top-navigation (header links, with no md-tab-body). The root of the problem is that md-tabs fires a synthetic click event on init, which causes an infinite redirect loop if clicking on a tab causes a page navigation (which re-initializes the tabs which causes a click which triggers a page navigation...).

tab init triggers the select() function:
https://github.com/angular/material/blob/a4ea9deacd42933f9a4cefaecea23e904bfa1d05/src/components/tabs/js/tabsController.js#L81

select() triggers a click:
https://github.com/angular/material/blob/a4ea9deacd42933f9a4cefaecea23e904bfa1d05/src/components/tabs/js/tabsController.js#L293

gmoothart:
Tabs component specifically for top-level page navigation. The UI follows md-tabs closely, but there is no concept of md-tab-body and the tabs just serve as links.

The navigation/routing method used is somewhat hard-coded but can use `ng-href`, `ui-sref` (ui-router), or `ng-click`. Because it is tied in with routing, it is hard to accurately show usage in the demo page (but I tried anyway).

This resolves #6676 , and it provides a much simpler component than md-tabs to cover a simpler use-case.

This is just a preliminary PR to get feedback on the general approach; in a final version I will add better documentation and testing

Basic API (with ui-router):
```
<md-nav-bar selected-tab-name="currentLinkName">
    <md-nav-link nav-sref="app.home" link-name="home">Home</md-nav-link>
    <md-nav-link nav-sref="app.reports" link-name="reports">Reports</md-nav-link>
    <md-nav-link nav-sref="app.admin" link-name="admin">Admin</md-nav-link>
</md-nav-bar>
...
$rootScope.$on('$stateChangeSuccess', function() {
  // currentLinkName must match one of the link-name attributes in the md-nav-link directives.
  // applications can do the route-to-linkName mapping in any convenient way
  $scope.currentLinkName = getLinkNameFromCurrentRoute();
});

```

### DIFF
--- a/src/components/navBar/demoBasicUsage/index.html
+++ b/src/components/navBar/demoBasicUsage/index.html
@@ -1,0 +1,11 @@
+<div ng-controller="AppCtrl" ng-cloak>
+  <md-content class="md-padding">
+    <md-nav-bar selected-link-name="currentLink" nav-bar-aria-label="navigation links">
+      <md-nav-link nav-click="goto('page1')" link-name="page1">Page One</md-nav-link>
+      <md-nav-link nav-click="goto('page2')" link-name="page2">Page Two</md-nav-link>
+      <md-nav-link nav-click="goto('page3')" link-name="page3">Page Three</md-nav-link>
+      <!-- these require actual routing with ui-router or ng-route, so they won't work in the demo -->
+      <md-nav-link nav-sref="app.page4" link-name="page4">Page Four</md-nav-link>
+      <md-nav-link nav-href="#page5" link-name="page5">Page Five</md-nav-link>
+    </md-nav-bar>
+  </md-content>
--- a/src/components/navBar/demoBasicUsage/index.html
+++ b/src/components/navBar/demoBasicUsage/index.html
@@ -1,11 +1,13 @@
 <div ng-controller="AppCtrl" ng-cloak>
   <md-content class="md-padding">
-    <md-nav-bar selected-link-name="currentLink" nav-bar-aria-label="navigation links">
-      <md-nav-link nav-click="goto('page1')" link-name="page1">Page One</md-nav-link>
-      <md-nav-link nav-click="goto('page2')" link-name="page2">Page Two</md-nav-link>
-      <md-nav-link nav-click="goto('page3')" link-name="page3">Page Three</md-nav-link>
-      <!-- these require actual routing with ui-router or ng-route, so they won't work in the demo -->
-      <md-nav-link nav-sref="app.page4" link-name="page4">Page Four</md-nav-link>
-      <md-nav-link nav-href="#page5" link-name="page5">Page Five</md-nav-link>
+    <md-nav-bar md-selected-nav-item="currentNavItem" nav-bar-aria-label="navigation links">
+      <md-nav-item md-nav-click="goto('page1')" name="page1">Page One</md-nav-item>
+      <md-nav-item md-nav-click="goto('page2')" name="page2">Page Two</md-nav-item>
+      <md-nav-item md-nav-click="goto('page3')" name="page3">Page Three</md-nav-item>
+      <!-- these require actual routing with ui-router or ng-route, so they won't work in the demo
+      <md-nav-item md-nav-sref="app.page4" name="page4">Page Four</md-nav-item>
+      <md-nav-item md-nav-href="#page5" name="page5">Page Five</md-nav-item>
+      -->
     </md-nav-bar>
+    <span>{{currentNavItem}}</span>
   </md-content>
--- a/src/components/navBar/demoBasicUsage/script.js
+++ b/src/components/navBar/demoBasicUsage/script.js
@@ -5,7 +5,6 @@
       .controller('AppCtrl', AppCtrl);
 
   function AppCtrl($scope) {
-    $scope.currentLink = 'page1';
-    $scope.goto = function(hash) { $scope.currentLink = hash; };
+    $scope.currentNavItem = 'page1';
   }
 })();
--- a/src/components/navBar/demoBasicUsage/script.js
+++ b/src/components/navBar/demoBasicUsage/script.js
@@ -1,0 +1,11 @@
+(function() {
+  'use strict';
+
+  angular.module('navBarDemoBasicUsage', ['ngMaterial'])
+      .controller('AppCtrl', AppCtrl);
+
+  function AppCtrl($scope) {
+    $scope.currentLink = 'page1';
+    $scope.goto = function(hash) { $scope.currentLink = hash; };
+  }
+})();
--- a/src/components/navBar/navBar-theme.scss
+++ b/src/components/navBar/navBar-theme.scss
@@ -5,7 +5,7 @@ md-nav-bar.md-THEME_NAME-theme {
     border-color: '{{foreground-4}}';
   }
 
-  .md-button.md-nav-button.md-unselected {
+  .md-button._md-nav-button.md-unselected {
     color: '{{foreground-2}}';
   }
 
--- a/src/components/navBar/navBar-theme.scss
+++ b/src/components/navBar/navBar-theme.scss
@@ -1,0 +1,16 @@
+md-nav-bar.md-THEME_NAME-theme {
+
+  .md-nav-bar {
+    background-color: transparent;
+    border-color: '{{foreground-4}}';
+  }
+
+  .md-button.md-nav-button.md-unselected {
+    color: '{{foreground-2}}';
+  }
+
+  md-nav-ink-bar {
+    color: '{{accent-color}}';
+    background: '{{accent-color}}';
+  }
+}
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -1,0 +1,509 @@
+/**
+ * @ngdoc module
+ * @name material.components.navBar
+ */
+
+
+angular.module('material.components.navBar', ['material.core'])
+    .controller('MdNavBarController', MdNavBarController)
+    .directive('mdNavBar', MdNavBar)
+    .controller('MdNavLinkController', MdNavLinkController)
+    .directive('mdNavLink', MdNavLink);
+
+
+/*****************************************************************************
+ *                            PUBLIC DOCUMENTATION                           *
+ *****************************************************************************/
+/**
+ * @ngdoc directive
+ * @name mdNavBar
+ * @module material.components.navBar
+ *
+ * @restrict E
+ *
+ * @description
+ * The `<md-nav-bar>` directive renders a list of material tabs that can be used
+ * for top-level page navigation. Unlike `<md-tabs>`, it has no concept of a tab
+ * body.
+ *
+ * Because it deals with page navigation, certain routing concepts are built-in.
+ * Route changes via via ng-href, ui-sref, or ng-click events are supported.
+ *
+ * Accessibility functionality is implemented as a site navigator with a
+ * listbox, according to
+ * https://www.w3.org/TR/wai-aria-practices/#Site_Navigator_Tabbed_Style
+ *
+ * @param {string=} selectedLinkName The name of the current tab; this must
+ * match the link-name attribute of `<md-nav-link>`
+ * @param {string=} navBarAriaLabel An aria-label for the nav-bar
+ *
+ * @usage
+ * <hljs lang="html">
+ *  <md-nav-bar selected-link-name="currentLink">
+ *    <md-nav-link nav-click="goto('page1')" link-name="page1">Page One</md-nav-link>
+ *    <md-nav-link nav-sref="app.page2" link-name="page2">Page Two</md-nav-link>
+ *    <md-nav-link nav-href="#page3" link-name="page3">Page Three</md-nav-link>
+ *  </md-nav-bar>
+ *</hljs>
+ * <hljs lang="js">
+ * (function() {
+ *   ‘use strict’;
+ *
+ *    $rootScope.$on('$routeChangeSuccess', function(event, current) {
+ *      $scope.currentLink = getCurrentLinkFromRoute(current);
+ *    });
+ * });
+ * </hljs>
+ */
+
+/*****************************************************************************
+ *                            mdNavLink                                      *
+ *****************************************************************************/
+/**
+ * @ngdoc directive
+ * @name mdNavLink
+ * @module material.components.navBar
+ *
+ * @restrict E
+ *
+ * @description
+ * `<md-nav-link>` describes a page navigation link within the `<md-nav-bar>`
+ * component. It renders an md-button as the actual link.
+ *
+ * Exactly one of the navClick, navHref, navSref attributes are required to be
+ * specified.
+ *
+ * @param {Function=} navClick Function which will be called when the
+ * link is clicked to change the page. Renders as an `ng-click`.
+ * @param {string=} navHref url to transition to when this link is clicked.
+ * Renders as an `ng-href`.
+ * @param {string=} navSref Ui-router state to transition to when this link is
+ * clicked. Renders as a `ui-sref`.
+ * @param {string=} linkName The name of this link. Used by the nav bar to know
+ * which link is currently selected.
+ *
+ * @usage
+ * See `<md-nav-bar>` for usage.
+ */
+
+
+/*****************************************************************************
+ *                                IMPLEMENTATION                             *
+ *****************************************************************************/
+
+function MdNavBar($mdAria) {
+  return {
+    restrict: 'E',
+    transclude: true,
+    controller: MdNavBarController,
+    controllerAs: 'ctrl',
+    bindToController: true,
+    scope: {
+      'selectedLinkName': '=?',
+      'navBarAriaLabel': '@?',
+    },
+    template:
+      '<div class="md-nav-bar">' +
+        '<nav role="navigation">' +
+          '<ul class="md-nav-bar-list" layout="row" ng-transclude role="listbox"' +
+            'tabindex="0"' +
+            'ng-focus="ctrl.onFocus()"' +
+            'ng-blur="ctrl.onBlur()"' +
+            'ng-keydown="ctrl.onKeydown($event)"' +
+            'aria-label="{{ctrl.navBarAriaLabel}}">' +
+          '</ul>' +
+        '</nav>' +
+        '<md-nav-ink-bar></md-nav-ink-bar>' +
+      '</div>',
+    link: function(scope, element, attrs, ctrl) {
+      if (!ctrl.navBarAriaLabel) {
+        $mdAria.expectAsync(element, 'aria-label', angular.noop);
+      }
+    }
+  };
+}
+
+/**
+ * Controller for the nav-bar component.
+ *
+ * Accessibility functionality is implemented as a site navigator with a
+ * listbox, according to
+ * https://www.w3.org/TR/wai-aria-practices/#Site_Navigator_Tabbed_Style
+ * @param {!angular.JQLite} $element
+ * @param {!angular.Scope} $scope
+ * @param {!angular.Timeout} $timeout
+ * @param {!Object} $mdConstant
+ * @constructor
+ * @final
+ * @ngInject
+ */
+function MdNavBarController($element, $scope, $timeout, $mdConstant) {
+  // Injected variables
+  /** @private @const {!angular.Timeout} */
+  this._$timeout = $timeout;
+
+  /** @private @const {!angular.Scope} */
+  this._$scope = $scope;
+
+  /** @private @const {!Object} */
+  this._$mdConstant = $mdConstant;
+
+  // Data-bound variables.
+  /** @type {string} */
+  this.selectedLinkName;
+
+  /** @type {string} */
+  this.navBarAriaLabel;
+
+  // State variables.
+
+  /** @type {?angular.JQLite} */
+  this._navBarEl = $element[0];
+
+  /** @type {?angular.JQLite} */
+  this._inkbar;
+
+  var self = this;
+  // need to wait for transcluded content to be available
+  var deregisterTabWatch = this._$scope.$watch(function() {
+    return self._navBarEl.querySelectorAll('.md-nav-button').length;
+  },
+  function(newLength) {
+    if (newLength > 0) {
+      self._initTabs();
+      deregisterTabWatch();
+    }
+  });
+}
+
+
+
+/**
+ * Initializes the tab components once they exist.
+ * @private
+ */
+MdNavBarController.prototype._initTabs = function() {
+  this._inkbar = angular.element(this._navBarEl.getElementsByTagName('md-nav-ink-bar')[0]);
+
+  var self = this;
+  this._$timeout(function() {
+    self._updateTabs(self.selectedLinkName, undefined);
+  });
+
+  this._$scope.$watch('ctrl.selectedLinkName', function(newValue, oldValue) {
+    // Wait a digest before update tabs for products doing
+    // anything dynamic in the template.
+    self._$timeout(function() {
+      self._updateTabs(newValue, oldValue);
+    });
+  });
+};
+
+/**
+ * Set the current tab to be selected.
+ * @param {string|undefined} newValue New current tab name.
+ * @param {string|undefined} oldValue Previous tab name.
+ * @private
+ */
+MdNavBarController.prototype._updateTabs = function(newValue, oldValue) {
+  var tabs = this._getTabs();
+
+  var oldIndex;
+  if (oldValue) {
+    var oldTab = this._getTabByName(oldValue);
+    if (oldTab) {
+      oldTab.setSelected(false);
+      oldIndex = tabs.indexOf(oldTab);
+    }
+  }
+
+  if (newValue) {
+    var tab = this._getTabByName(newValue);
+    if (tab) {
+      tab.setSelected(true);
+      var newIndex = tabs.indexOf(tab);
+      var self = this;
+      this._$timeout(function() {
+        self._updateInkBarStyles(tab, newIndex, oldIndex);
+      });
+    }
+  }
+};
+
+/**
+ * Repositions the ink bar to the selected tab.
+ * @private
+ */
+MdNavBarController.prototype._updateInkBarStyles = function(tab, newIndex, oldIndex) {
+  var tabEl = tab.getButtonEl();
+  var left = tabEl.offsetLeft;
+
+  this._inkbar.toggleClass('md-left', newIndex < oldIndex)
+      .toggleClass('md-right', newIndex > oldIndex);
+  this._inkbar.css({left: left + 'px', width: tabEl.offsetWidth + 'px'});
+};
+
+/**
+ * Returns an array of the current tabs.
+ * @return {!Array<!NavLinkController>}
+ * @private
+ */
+MdNavBarController.prototype._getTabs = function() {
+  var linkArray = Array.prototype.slice.call(
+      this._navBarEl.querySelectorAll('.md-nav-link'));
+  return linkArray.map(function(el) {
+    return angular.element(el).controller('mdNavLink')
+  });
+};
+
+/**
+ * Returns the tab with the specified name.
+ * @param {string} name The name of the tab, found in its link-name attribute.
+ * @return {!NavLinkController|undefined}
+ * @private
+ */
+MdNavBarController.prototype._getTabByName = function(name) {
+  return this._findTab(function(tab) {
+    return tab.getName() == name;
+  });
+};
+
+/**
+ * Returns the selected tab.
+ * @return {!NavLinkController|undefined}
+ * @private
+ */
+MdNavBarController.prototype._getSelectedTab = function() {
+  return this._findTab(function(tab) {
+    return tab.isSelected()
+  });
+};
+
+/**
+ * Returns the focused tab.
+ * @return {!NavLinkController|undefined}
+ */
+MdNavBarController.prototype.getFocusedTab = function() {
+  return this._findTab(function(tab) {
+    return tab.hasFocus()
+  });
+};
+
+/**
+ * Find a tab that matches the specified function.
+ * @private
+ */
+MdNavBarController.prototype._findTab = function(fn) {
+  var tabs = this._getTabs();
+  for (var i = 0; i < tabs.length; i++) {
+    if (fn(tabs[i])) {
+      return tabs[i];
+    }
+  }
+
+  return null;
+};
+
+/**
+ * Direct focus to the selected tab when focus enters the nav bar.
+ */
+MdNavBarController.prototype.onFocus = function() {
+  var tab = this._getSelectedTab();
+  if (tab) {
+    tab.setFocused(true);
+  }
+};
+
+/**
+ * Clear tab focus when focus leaves the nav bar.
+ */
+MdNavBarController.prototype.onBlur = function() {
+  var tab = this.getFocusedTab();
+  if (tab) {
+    tab.setFocused(false);
+  }
+};
+
+/**
+ * Move focus from oldTab to newTab.
+ * @param {!NavLinkController} oldTab
+ * @param {!NavLinkController} newTab
+ * @private
+ */
+MdNavBarController.prototype._moveFocus = function(oldTab, newTab) {
+  oldTab.setFocused(false);
+  newTab.setFocused(true);
+};
+
+/**
+ * Responds to keypress events.
+ * @param {!Event} e
+ */
+MdNavBarController.prototype.onKeydown = function(e) {
+  var keyCodes = this._$mdConstant.KEY_CODE;
+  var tabs = this._getTabs();
+  var focusedTab = this.getFocusedTab();
+  if (!focusedTab) return;
+
+  var focusedTabIndex = tabs.indexOf(focusedTab);
+
+  // use arrow keys to navigate between tabs
+  switch (e.keyCode) {
+    case keyCodes.UP_ARROW:
+    case keyCodes.LEFT_ARROW:
+      if (focusedTabIndex > 0) {
+        this._moveFocus(focusedTab, tabs[focusedTabIndex - 1]);
+      }
+      break;
+    case keyCodes.DOWN_ARROW:
+    case keyCodes.RIGHT_ARROW:
+      if (focusedTabIndex < tabs.length - 1) {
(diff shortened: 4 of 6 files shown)


gmoothart:
@ThomasBurleson @jelbourn @ErinCoughlan @KarenParker @DerekLouie Please take a look and let me know what you think!

jelbourn:
+ @robertmesserle

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+        'ng-href="{{ctrl.navHref}}">' +
+        '<span ng-transclude></span>' +
+        '</md-button>' +
+        '<md-button ng-if="ctrl.navClick" class="md-nav-button md-accent"' +
+        'ng-class="ctrl.getNgClassMap()"' +
+        'tabindex="-1"' +
+        'ng-click="ctrl.navClick()">' +
+        '<span ng-transclude></span>' +
+        '</md-button>' +
+        '</li>',
```
jelbourn: Needs indenting

src/components/navBar/demoBasicUsage/index.html
```diff
@@ -0,0 +1,11 @@
+<div ng-controller="AppCtrl" ng-cloak>
+  <md-content class="md-padding">
+    <md-nav-bar selected-tab-name="currentLink">
+      <md-nav-link nav-click="goto('page1')" link-name="page1">Page One</md-nav-link>
```
jelbourn: Does `md-nav-link` render as an anchor element at the end of the day?

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+        'ng-focus="ctrl.onFocus()"' +
+        'ng-blur="ctrl.onBlur()"' +
+        'ng-keydown="ctrl.onKeydown($event)"' +
+        'aria-label="[[Top-level navigation links|' +
+        'Description of navigation links, which provide quick access to' +
+        'main sections of the site.]]">' +
+        '</ul>' +
+        '</nav>' +
+        '<div class="md-ink-bar"></div>' +
+        '</div>'
```
jelbourn: Needs indent

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+
+/**
+ * Controller for the nav-bar component.
+ *
+ * Accessibility functionality is implemented as a site navigator with a
+ * listbox, according to
+ * https://www.w3.org/TR/wai-aria-practices/#Site_Navigator_Tabbed_Style
+ * @param {!suiteTypes.jQuery} $element
+ * @param {!angular.Scope} $scope
+ * @ngInject
```
jelbourn: Add `@constructor`
(WebStorm understands JsDoc annotations like `@constructor` and has some intelligence around them)

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+  // Data-bound variables.
+  /** @export {string} */
+  // ctrl.selectedTabName = this.selectedTabName;
+
+  // State variables.
+
+  /** @private {?Element} */
+  ctrl.domEl_;
+
+  /** @private {?suiteTypes.jQuery} */
```
jelbourn: This type doesn't make sense here. I think you want `angular.JQLite`

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+  ctrl.inkbar_;
+
+  // need to wait for transcluded content to be available
+  var deregisterTabWatch = ctrl.scope_.$watch(
+      function() {
```
jelbourn: Too much indent, but also you can move the `function() {` part to the previous line
(assume ng-material has a 100-character line length limit)

src/components/navBar/demoBasicUsage/script.js
```diff
@@ -0,0 +1,11 @@
+(function() {
+  'use strict';
+
+  angular.module('navBarDemoBasicUsage', ['ngMaterial'])
+      .controller('AppCtrl', AppCtrl);
+
+  function AppCtrl($scope) {
+    $scope.currentLink = 'page1';
+    $scope.goto = function(hash) { $scope.currentLink = hash; };
```
robertmesserle: Just wanted to confirm: this is only to side-step the actual routing problem mentioned in the index.html comment?

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+  }
+
+  /**
+   * Returns the tab with the specified name.
+   * @param {string} name The name of the tab, found in its data-name attribute.
+   * @return {!NavLinkController|undefined}
+   * @private
+   */
+  function getTabByName_(name) {
+    return findTab_(function(tab) { return tab.getName() == name; });
```
jelbourn: optional: personally I'm not a fan of the single-lined functions, but this project doesn't officially have a rule against it.

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+   * @param {!NavLinkController} newTab
+   */
+  function moveFocus(oldTab, newTab) {
+    oldTab.setFocused(false);
+    newTab.setFocused(true);
+  }
+
+  /**
+   * Responds to keypress events.
+   * @param {!jQuery.Event} e
```
jelbourn: You probably just want `Event`

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+   * @export
+   */
+  function onKeydown(e) {
+    var tabs = getTabs_();
+    var focusedTab = googAsserts.assert(getFocusedTab_());
```
jelbourn: Where is this assert function coming from? Wherever it is, it shouldn't be called `googAsserts`

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+    var focusedTabIndex = tabs.indexOf(focusedTab);
+
+    // use arrow keys to navigate between tabs
+    switch (e.which) {
+      case KeyCodes.UP:
```
jelbourn: Where are these KeyCodes defined?

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+
+function MdNavBar() {
+  return {
+    restrict: 'E',
+    transclude: true,
+    controller: MdNavBarController,
+    controllerAs: 'ctrl',
+    bindToController: true,
+    scope: {'selectedTabName': '=?'},
+    template: '<div class="suite md-nav-bar">' +
```
robertmesserle: What is the class `suite`?  Should it be prefixed?

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+  /** @private @const */
+  ctrl.element_ = $element;
+
+  // Data-bound variables
+  /** @export @const {?Function} */
+  // this.navClick;
+  /** @export @const {?string} */
+  // this.navHref;
+  /** @export @const {?string} */
+  // this.navSref;
```
jelbourn: Commented-out code

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+
+  /** @private {boolean} */
+  ctrl.focused_ = false;
+
+
+  var hasNavClick = ctrl.navClick != null;
+  var hasNavHref = ctrl.navHref != null;
+  var hasNavSref = ctrl.navSref != null;
+
+  if (!(hasNavClick ^ hasNavHref ^ hasNavSref)) {
```
jelbourn: Why are you using bitwise xor? Using bitwise operators on non-integer types is generally considered an anti-pattern.

src/components/navBar/navBar.scss
```diff
@@ -0,0 +1,71 @@
+/** Matches "md-tabs md-tabs-wrapper" style. */
+$nav-bar-height: 48px;
```
jelbourn: Prefix variables with `md-`

src/components/navBar/navBar.scss
```diff
@@ -0,0 +1,71 @@
...
+$gms-color-grey-50: #fafafa;
```
jelbourn: Colors should generally be defined in terms of the theme.

src/components/navBar/navBar.scss
```diff
@@ -0,0 +1,71 @@
...
+
+.md-nav-bar {
+  background-color: $gms-color-grey-50;
+  border-color: $gms-color-divider;
+  border-style: solid;
+  border-width: 0 0 1px;
+  height: $nav-bar-height;
+  position: relative;
+
+  ul {
```
jelbourn: Instead of nesting, introduce individual class that apply to each type of item, e.g.,:
```
.md-nav-bar {
 ...
}

.md-nav-bar-list {
  ...
}

.md-nav-bar-listitem {
  ...
}
```

This keeps the specificity down and can, in some cases, improve performance.

src/components/navBar/navBar.js
ErinCoughlan: Probably also add `@final`.

src/components/navBar/navBar.scss
```diff
@@ -0,0 +1,71 @@
...
+
+    &:focus {
+      outline: none;
+    }
+
+    &:hover {
+      background-color: inherit;
+    }
+
+    &.unselected {
```
jelbourn: Add CSS classes must be prefixed with `md-`

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+}
+
+/**
+ * Controller for the nav-bar component.
+ *
+ * Accessibility functionality is implemented as a site navigator with a
+ * listbox, according to
+ * https://www.w3.org/TR/wai-aria-practices/#Site_Navigator_Tabbed_Style
+ * @param {!suiteTypes.jQuery} $element
+ * @param {!angular.Scope} $scope
```
ErinCoughlan: Missing `$timeout` as a param.

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+
+  /** @private @const {Element} */
+  ctrl.element_ = $element[0];
+
+  /** @private @const */
+  ctrl.timeout_ = $timeout;
+
+  // Data-bound variables.
+  /** @export {string} */
+  // ctrl.selectedTabName = this.selectedTabName;
```
ErinCoughlan: Should this be commented out? Please remove if it's not necessary.

src/components/navBar/navBar.scss
```diff
@@ -0,0 +1,71 @@
...
+    bottom: 0;
+    height: 2px;
+    left: auto;
+    position: absolute;
+    right: auto;
+    background-color: black;
+
+    &.md-left {
+      transition: left 0.125s cubic-bezier(0.35, 0, 0.25, 1),
+        right 0.25s cubic-bezier(0.35, 0, 0.25, 1);
```
jelbourn: Can these transitions use a predefined variable from `variables.scss`?

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+
+  // State variables.
+
+  /** @private {?Element} */
+  ctrl.domEl_;
```
ErinCoughlan: Maybe navBar is a better name here?

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+ * https://www.w3.org/TR/wai-aria-practices/#Site_Navigator_Tabbed_Style
+ * @param {!suiteTypes.jQuery} $element
+ * @param {!angular.Scope} $scope
+ * @ngInject
+ */
+function MdNavBarController($element, $scope, $timeout) {
+  var ctrl = this;
+
+  /** @private @const */
+  ctrl.scope_ = $scope;
```
ErinCoughlan: The underscores go first in Material (at least that's what we've been doing).

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+        ctrl.timeout_(function() {
+          updateInkBarStyles(tab, newIndex, oldIndex);
+        });
+      }
+    }
+  }
+
+  /**
+   * Repositions the ink bar to the selected tab.
+   * "borrowed" from ng-material
```
ErinCoughlan: You're in ngMaterial now. :) So you're "borrowing" from yourself.

src/components/navBar/navBar.js
ErinCoughlan: You can get KeyCodes from $mdConstant.

`var keyCodes = this.$mdConstant.KEY_CODE;`

Then you can use `DOWN_ARROW`, `LEFT_ARROW`, etc.

src/components/navBar/demoBasicUsage/index.html
gmoothart: It is an md-button so it renders as an anchor, or a button if you use a click event

src/components/navBar/demoBasicUsage/script.js
gmoothart: Yes. in "real" code this function would do a page navigation somehow.

src/components/navBar/navBar.scss
```diff
@@ -0,0 +1,71 @@
+/** Matches "md-tabs md-tabs-wrapper" style. */
+$nav-bar-height: 48px;
+$gms-color-grey-50: #fafafa;
+$gms-color-divider: rgba(0, 0, 0, 0.12);
```
ErinCoughlan: gms?

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+    var linkArray = Array.prototype.slice.call(
+        ctrl.element_.querySelectorAll('.md-nav-link'));
+    return linkArray.map(function(el) {
+      return angular.element(el).controller('mdNavLink')
+    });
+  }
+
+  /**
+   * Returns the tab with the specified name.
+   * @param {string} name The name of the tab, found in its data-name attribute.
```
robertmesserle: Isn't it `link-name`?

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,400 @@
...
+        'tabindex="-1"' +
+        'ng-click="ctrl.navClick()">' +
+        '<span ng-transclude></span>' +
+        '</md-button>' +
+        '</li>',
+    scope: {
+      'navClick': '&?',
+      'navHref': '@?',
+      'navSref': '@?',
+      'linkName': '@',
```
robertmesserle: Does `link-name` serve any purpose other than identifying the tab?  Could we identify tabs another way to remove this burden from the user?  (I don't have an definite answer, but it seems like we could probably find a way to remove this.)

ErinCoughlan: @robertmesserle I'm not sure we can without hard-coding an ordering, which would prevent the ability to use ng-if with element. (For instance, to hide an Admin tab from people with Admin permissions.)

The difficult part is when nav events take place outside of the component, such as by a user clicking a link on a page. The tab needs to update, which is done through updating the `selected-tab` attribute. That value needs to correspond to a specific tab, so without the `name` attribute, we can't match and update them.

robertmesserle: Yea, that makes sense.  If we were only supporting href, we could use route to figure that out - forgot that it also supports function calls.

src/components/navBar/navBar.js
gmoothart: fixed; removed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: Yes, fixed.

src/components/navBar/navBar.js
gmoothart: I don't like them either; it was just clang. fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.js
gmoothart: fixed.

src/components/navBar/navBar.scss
gmoothart: fixed.

src/components/navBar/navBar.scss
gmoothart: fixed.

src/components/navBar/navBar.scss
gmoothart: fixed.

src/components/navBar/navBar.scss
gmoothart: fixed.

src/components/navBar/navBar.scss
gmoothart: fixed.

src/components/navBar/navBar.scss
gmoothart: fixed.

gmoothart:
I believe I've addressed the comments, PTAL. I will work on proper documentation and tests next.

src/components/navBar/navBar.js
gmoothart: @robertmesserle I would like to get rid of it too, but I don't think we can. In addition to Erin's comments, there is not necessarily a 1:1 correlation between selected-tab and route. For example an "admin" tab might need to be selected for a large number of sub-pages within the admin section.

robertmesserle: It's not a major concern, but I am still not completely sure how that scenario would work with or without the label.  An external route change would have to somehow specify which tab is now selected, wouldn't it?

But judging solely by the code that references this, it only really seems to be used for DOM element tracking.  I think that could be solved without requiring the user to add this.

src/components/navBar/navBar.js
```diff
@@ -27,21 +27,22 @@ function MdNavBar() {
...
-        'aria-label="[[Top-level navigation links|' +
-        'Description of navigation links, which provide quick access to' +
-        'main sections of the site.]]">' +
-        '</ul>' +
+          '<ul class="md-nav-bar-list" layout="row" ng-transclude role="listbox"' +
+            'tabindex="0"' +
+            'ng-focus="ctrl.onFocus()"' +
+            'ng-blur="ctrl.onBlur()"' +
+            'ng-keydown="ctrl.onKeydown($event)"' +
+            'aria-label="[[Top-level navigation links|' +
```
ErinCoughlan: I don't think there are translations like this in Material. I haven't seen them anywhere else.

@jelbourn How do aria-labels and other text content get translated? (Since as a library, you don't know what languages are supported.)

src/components/navBar/navBar.js
```diff
@@ -51,60 +52,60 @@ function MdNavBar() {
  * Accessibility functionality is implemented as a site navigator with a
  * listbox, according to
  * https://www.w3.org/TR/wai-aria-practices/#Site_Navigator_Tabbed_Style
- * @param {!suiteTypes.jQuery} $element
+ * @param {!angular.JQLite} $element
  * @param {!angular.Scope} $scope
+ * @param {!angular.Timeout} $timeout
+ * @param {!Object} $mdConstant
+ * @constructor
```
ErinCoughlan: optional: You can put these all on the same line:

`@final @constructor @ngInject`

src/components/navBar/navBar.js
```diff
@@ -51,60 +52,60 @@ function MdNavBar() {
...
   // Data-bound variables.
   /** @export {string} */
-  // ctrl.selectedTabName = this.selectedTabName;
+  ctrl.selectedLinkName;
 
   // State variables.
 
-  /** @private {?Element} */
-  ctrl.domEl_;
+  /** @type {?angular.JQLite} */
```
ErinCoughlan: Should this be `?Element`?

gmoothart:
@jelbourn @robertmesserle @ErinCoughlan @DerekLouie @KarenParker PTAL, I've added tests and documentation.

src/components/navBar/navBar.js
gmoothart: Yes, an external route change would have to specify which tab is now selected. Something like this:
```
$rootScope.$on('$routeChangeSuccess', function(event, current) {
    $scope.currentLink = getCurrentLinkFromRoute(current);
});
```
Where the user is responsible for the details of `getCurrentLinkFromRoute`, but it could include logic like "return 'admin' if current route is a child of /admin".

I think the main issue is that somehow the user has to tell us what tab should be selected. If we remove link-name, we need to replace it with another method for saying "select this tab".

I suppose we could use index, but that does not seem better - individual tabs could be ng-if'd (I have seen some examples of this) and so the tab indices could change over time.

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,496 @@
...
+ */
+function MdNavBarController($element, $scope, $timeout, $mdConstant) {
+  var ctrl = this;
+
+  // Data-bound variables.
+  /** @export {string} */
+  ctrl.selectedLinkName;
+
+  /** @export {string} */
+  ctrl.navBarAriaLabel;
```
ErinCoughlan: Since this is optional, should it be defaulted to something to keep the nav-bar accessible?

src/components/navBar/navBar.spec.js
```diff
@@ -0,0 +1,247 @@
+describe('<md-nav-bar>', function() {
+  var el, compile, scope, timeout, ctrl, tabContainer, $material, $mdConstant;
+
+  /** @ngInject */
+  var injectLocals = function(
+      $compile, $rootScope, $timeout, _$material_, _$mdConstant_) {
+    compile = $compile;
```
ErinCoughlan: If you want, you can keep the $ in the variables names, like $scope, or $timeout.

src/components/navBar/navBar.spec.js
```diff
@@ -0,0 +1,247 @@
+describe('<md-nav-bar>', function() {
```
ErinCoughlan: This should be `mdNavBar`

src/components/navBar/navBar.spec.js
```diff
@@ -0,0 +1,247 @@
...
+        '    tab2' +
+        '  </md-nav-link>' +
+        '  <md-nav-link nav-href="#3" link-name="tab3">' +
+        '    tab3' +
+        '  </md-nav-link>' +
+        '</md-nav-bar>');
+  }
+
+  describe('tabs', function() {
+    it('shows current tab as selected', function() {
```
ErinCoughlan: specs are traditionally of the form: `it('should ...'`

ErinCoughlan:
LGTM. You should make sure to squash the commits and change the commit message to something of the form: `update/fix/feat(componentName): Description.`

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,496 @@
...
+        if (newLength > 0) {
+          initTabs();
+          deregisterTabWatch();
+        }
+      });
+
+  /**
+   * Initializes the tab components once they exist.
+   */
+  function initTabs() {
```
jelbourn: optional: it would be better if these functions were on the `prototype`

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,496 @@
...
+  ctrl.navClick;
+  /** @export @const {?string} */
+  ctrl.navHref;
+  /** @export @const {?string} */
+  ctrl.navSref;
+
+
+  // State variables
+  /** @private {boolean} */
+  var selected_ = false;
```
jelbourn: Inconsistent use of leading vs. training underscore in this file.

Also, these aren't _properties_, so technically they can't be `@private`

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,496 @@
...
+        'Must specify exactly one of nav-click, nav-href, ' +
+        'nav-sref for nav-link directive');
+  }
+
+  /**
+   * Returns a map of class names and values for use by ng-class.
+   * @return {!Object<string,boolean>}
+   * @export
+   */
+  function getNgClassMap() {
```
jelbourn: optional: same thing about the `prototype`

jelbourn:
LGTM aside from minor nits. Commits do need to be squashed to something like
```
feat(navbar): add new navbar component
```

src/components/navBar/navBar.js
gmoothart: No, because we can't do any translation here - there is no way to localize the text.

jelbourn: It's okay to provide a default value that users can customize to internationalize.

gmoothart:
Ok, comments addressed.

src/components/navBar/demoBasicUsage/index.html
```diff
@@ -0,0 +1,11 @@
+<div ng-controller="AppCtrl" ng-cloak>
+  <md-content class="md-padding">
+    <md-nav-bar selected-link-name="currentLink" nav-bar-aria-label="navigation links">
+      <md-nav-link nav-click="goto('page1')" link-name="page1">Page One</md-nav-link>
```
ThomasBurleson: namesapce `link-name` --> change to `md-link-name`.

src/components/navBar/demoBasicUsage/index.html
```diff
@@ -0,0 +1,11 @@
...
+      <md-nav-link nav-click="goto('page2')" link-name="page2">Page Two</md-nav-link>
+      <md-nav-link nav-click="goto('page3')" link-name="page3">Page Three</md-nav-link>
```
ThomasBurleson: *  why `nav-click`... ? 
*  why not `ng-click`? 
*  If do extra things why not override `ng-click` ?

src/components/navBar/navBar-theme.scss
```diff
@@ -0,0 +1,16 @@
+md-nav-bar.md-THEME_NAME-theme {
+
+  .md-nav-bar {
+    background-color: transparent;
+    border-color: '{{foreground-4}}';
+  }
+
+  .md-button.md-nav-button.md-unselected {
```
ThomasBurleson: Any of these private ?

src/components/navBar/demoBasicUsage/index.html
ThomasBurleson: * Why not `md-nav-item`... which is similar to the APIs in other components. 
* Why no use `name` instead of `link-name`?

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,509 @@
...
+ * @ngdoc directive
+ * @name mdNavBar
+ * @module material.components.navBar
+ *
+ * @restrict E
+ *
+ * @description
+ * The `<md-nav-bar>` directive renders a list of material tabs that can be used
+ * for top-level page navigation. Unlike `<md-tabs>`, it has no concept of a tab
+ * body.
```
ThomasBurleson: no body and no bar pagination.

src/components/navBar/demoBasicUsage/index.html
gmoothart: not `ng-click` because it needs to be applied on a child element rendered by the directive, not on the directive itself.

Overriding is an intesting idea, I can probably do that.

src/components/navBar/demoBasicUsage/index.html
gmoothart: is `md-name` what you are suggesting?

src/components/navBar/demoBasicUsage/index.html
gmoothart: Actually I don't think I can override, since what I need is not to add additional functionality to ng-click/ui-sref/etc, but to apply the actual directives at a lower level in the dom.

gmoothart:
Addressed Thomas' suggestions (in a separate commit for ease of review).

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,533 @@
...
+        '<nav role="navigation">' +
+          '<ul class="_md-nav-bar-list" layout="row" ng-transclude role="listbox"' +
+            'tabindex="0"' +
+            'ng-focus="ctrl.onFocus()"' +
+            'ng-blur="ctrl.onBlur()"' +
+            'ng-keydown="ctrl.onKeydown($event)"' +
+            'aria-label="{{ctrl.navBarAriaLabel}}">' +
+          '</ul>' +
+        '</nav>' +
+        '<md-nav-ink-bar></md-nav-ink-bar>' +
```
ThomasBurleson: Do we really need a `md-nav-ink-bar`? Would `md-ink-bar` not work ?

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,533 @@
...
+  /** @type {?angular.JQLite} */
+  this._navBarEl = $element[0];
+
+  /** @type {?angular.JQLite} */
+  this._inkbar;
+
+  var self = this;
+  // need to wait for transcluded content to be available
+  var deregisterTabWatch = this._$scope.$watch(function() {
+    return self._navBarEl.querySelectorAll('._md-nav-button').length;
```
ThomasBurleson: does `querySelectorAll()` trigger a **reflow** in the browser ?

src/components/navBar/navBar.js
```diff
@@ -0,0 +1,533 @@
...
+        '</md-button>' +
+        '<md-button ng-if="ctrl.mdNavClick" class="_md-nav-button md-accent"' +
+          'ng-class="ctrl.getNgClassMap()"' +
+          'tabindex="-1"' +
+          'ng-click="ctrl.mdNavClick()">' +
+          '<span ng-transclude class="_md-nav-button-text"></span>' +
+        '</md-button>' +
+      '</li>',
+    scope: {
+      'mdNavClick': '&?',
```
ThomasBurleson: What about ui-sref, etc... these have caused us *issues* on other components.

src/components/navBar/navBar.js
gmoothart: I do not think so. It is not mentioned in this list by Paul Irish:
https://gist.github.com/paulirish/5d52fb081b3570c81e3a

src/components/navBar/navBar.js
gmoothart: @ThomasBurleson sorry I don't understand, what are you suggesting?

src/components/navBar/navBar.js
gmoothart: I wanted the styles to not be dependent on the md-tabs component.

src/components/navBar/navBar.js
ThomasBurleson: Understood. FYI - https://books.google.com/books?id=ED6ph4WEIoQC&pg=PA50&lpg=PA50&dq=does+querySelectorAll+reflow&source=bl&ots=2Vtc5E0adM&sig=1gXc7D8Pxnjx4Kk497z8bfznoKw&hl=en&sa=X&ved=0ahUKEwir_Ja19qrMAhUkvIMKHe0UB28Q6AEIQzAG#v=onepage&q=does%20querySelectorAll%20reflow&f=false

src/components/navBar/navBar.js
ThomasBurleson: Later, we should refactor `md-ink-bar` to support both agnostically.

src/components/navBar/navBar.js
ThomasBurleson: @gmoothart - plz ignore. My comment is valid for a different issue; not ^^.

ThomasBurleson:
@gmoothart, @ErinCoughlan - lgtm.

> Nice work @gmoothart.

epelc:
@gmoothart have you tried using the `md-no-select-click` option on `md-tabs`? Enabling that stops the automatic click events from firing when a tab is initially selected by `md-tabs`.  I use this option to prevent multiple tabs from bouncing you back to previous ui-router states when the page is reloaded.

I'm a little late to the party, but I don't think we need another set of components so closely related to `md-tabs` if it already works with `md-no-select-click`.

gmoothart:
@epelc I did not know about `md-no-select-click`, I think it wasn't available in the version I was using when I had that problem.

We do have plans, though, to make md-nav-bar also implement the bottom-navigation spec: #8421

epelc:
@gmoothart Thanks for the info. I can see how tabs are different from nav items after looking at the bottom-navigation spec.

I had to implement something similar to bottom-navigation and it was a pain to get the icons and text in a tab how I wanted them. I ended up going horizontal [icon text] [icon text] instead of stacking the icon on top of the text. I wasn't aware of that spec though. 

Looking forward to your work on #8421.

ThomasBurleson:
@epelc, @gmoothart - mdTabs has performance consequences due to tab body switching, dynamic tabs, and other advanced features. The mdNavBar is a tab-button only implementation that does not suffer similar performance issues.

Issue #7385, opened 2016-03-02
FR: Create a new Material component for searching within mdSelect.


ErinCoughlan:
Blocked by #7383 

Using the md-select-header, add an input box with default search behavior (case insensitive contains).

Create a new directive like `md-menu-filter` that people can drop into the menu header.

EladBezalel:
@ErinCoughlan isn't that autocomplete?

ErinCoughlan:
@EladBezalel It's a similar component. If we used autocomplete, we would need to add multi-select options along with using a custom template to get the rest of the select behavior.

Here's a screenshot of what I'm looking for (but a Material version):
![image](https://cloud.githubusercontent.com/assets/2660421/13479492/20397050-e08c-11e5-91e4-99000d04161a.png)

I'm hoping there is a way to combine the work for autocomplete, make it more generic to support multiple and single select, and then use it within both autocomplete and select since they are so similar.

kathkroll:
The mdChips component when used with autocomplete seems to accomplish some of this functionality, but there are a fair amount of bugs with this combination. One notable one being that you can't remove used selections from the available dropdown without introducing another issue, as detailed [here](https://github.com/angular/material/issues/6674).

ErinCoughlan:
@kathroll For a select component, you actually want selected items to appear in the search results. Consider a list of 10,000 items. You search for a value and select all of the items that appeared. You then clear the search term, notice you selected an incorrect item, and then have to scroll through 10,000 items to find and unselect it. That's a broken user experience.

ThomasBurleson:
This issue is closed as part of our ‘Surge Focus on Material 2' efforts.
      For details, see our forum posting @ http://bit.ly/1UhZyWs.

PR #7782, opened 2016-03-28
feat(select): Adding md-select-header directive to md-select.

### LINKED ISSUE #7383: mdSelect: (feat) Add md-select-header as a template to the existing select component.

DerekLouie:
@jelbourn @ErinCoughlan @gmoothart @KarenParker Please review.

Fixes #7383

mdselectheader1

------ NO LONGER AN ISSUE: KEPT FOR POSTERITY ------ 

Currently facing an issue where the style.css in the demoSelectHeader folder is not getting compiled into the demo site and subsequently not being applied. Screenshot above only work because I put the css in select.scss to get it to look correct..

This is what it looks like currently without any of the demo css:

mdselectheadernocss

### DIFF
--- a/src/components/select/demoSelectHeader/index.html
+++ b/src/components/select/demoSelectHeader/index.html
@@ -1,0 +1,25 @@
+<div ng-controller="SelectHeaderController" class="md-padding" ng-cloak>
+  <div>
+    <h1 class="md-title">Pick a vegetable below</h1>
+    <div layout="row">
+      <md-input-container>
+        <label>Vegetables</label>
+        <md-select ng-model="selectedVegetables"
+                   md-on-close="clearSearchTerm()"
+                   data-md-container-class="selectdemoSelectHeader"
+                   multiple>
+          <md-select-header class="demo-select-header">
+            <input ng-model="searchTerm"
+                   type="search"
+                   placeholder="Search for a vegetable.."
+                   class="demo-header-searchbox _md-text">
+          </md-select-header>
+          <md-optgroup label="vegetables">
+            <md-option ng-value="vegetable" ng-repeat="vegetable in vegetables |
+              filter:searchTerm">{{vegetable}}</md-option>
+          </md-optgroup>
+        </md-select>
+      </md-input-container>
+    </div>
+ </div>
+</div>
--- a/src/components/select/demoSelectHeader/style.css
+++ b/src/components/select/demoSelectHeader/style.css
@@ -1,0 +1,25 @@
+/* Please note: All these selectors are only applied to children of elements with the 'selectdemoSelectHeader' class */
+
+.demo-header-searchbox {
+  border: none;
+  outline: none;
+  height: 100%;
+  width: 100%;
+  padding: 0;
+}
+
+.demo-select-header {
+  box-shadow: 0 1px 0 0 rgba(0, 0, 0, 0.1), 0 0 0 0 rgba(0, 0, 0,
+  0.14), 0 0 0 0 rgba(0, 0, 0, 0.12);
+  padding-left: 10.667px;
+  height: 48px;
+  cursor: pointer;
+  position: relative;
+  display: flex;
+  align-items: center;
+  width: auto;
+}
+
+md-content._md {
+  max-height: 240px;
+}
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -23,6 +23,7 @@ angular.module('material.components.select', [
   .directive('mdSelectMenu', SelectMenuDirective)
   .directive('mdOption', OptionDirective)
   .directive('mdOptgroup', OptgroupDirective)
+  .directive('mdSelectHeader', SelectHeaderDirective)
   .provider('$mdSelect', SelectProvider);
 
 /**
@@ -64,6 +65,27 @@ angular.module('material.components.select', [
  *     <label>State</label>
  *     <md-select
  *       ng-model="someModel">
+ *       <md-option ng-value="opt" ng-repeat="opt in neighborhoods2">{{ opt }}</md-option>
+ *     </md-select>
+ *   </md-input-container>
+ * </hljs>
+ *
+ * With a select-header
+ *
+ * When a developer needs to put more than just a text label in the
+ * md-select-menu, they should use the md-select-header.
+ * The user can put custom HTML inside of the header and style it to their liking.
+ * One common use case of this would be a sticky search bar.
+ *
+ * When using the md-select-header the labels that would previously be added to the
+ * OptGroupDirective are ignored.
+ *
+ * <hljs lang="html">
+ *   <md-input-container>
+ *     <md-select ng-model="someModel">
+ *       <md-select-header>
+ *         <span> Neighborhoods - </span>
+ *       </md-select-header>
  *       <md-option ng-value="opt" ng-repeat="opt in neighborhoods2">{{ opt }}</md-option>
  *     </md-select>
  *   </md-input-container>
@@ -212,7 +234,11 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
           };
 
         if (containerCtrl.input) {
-          throw new Error("<md-input-container> can only have *one* child <input>, <textarea> or <select> element!");
+          // We ignore inputs that are in the md-select-header (one
+          // case where this might be useful would be adding as searchbox)
+          if (element.find('md-select-header').find('input')[0] !== containerCtrl.input[0]) {
+            throw new Error("<md-input-container> can only have *one* child <input>, <textarea> or <select> element!");
+          }
         }
 
         containerCtrl.input = element;
@@ -507,11 +533,13 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
 }
 
 function SelectMenuDirective($parse, $mdUtil, $mdTheming) {
-
+  // We want the scope to be set to 'false' so an isolated scope is not created
+  // which would interfere with the md-select-header's access to the
+  // parent scope.
   return {
     restrict: 'E',
     require: ['mdSelectMenu'],
-    scope: true,
+    scope: false,
     controller: SelectMenuController,
     link: {pre: preLink}
   };
@@ -897,14 +925,32 @@ function OptgroupDirective() {
     compile: compile
   };
   function compile(el, attrs) {
-    var labelElement = el.find('label');
-    if (!labelElement.length) {
-      labelElement = angular.element('<label>');
-      el.prepend(labelElement);
+    // If we have a select header element, we don't want to add the normal label
+    // header.
+    if (!hasSelectHeader()) {
+      setupLabelElement();
     }
-    labelElement.addClass('_md-container-ignore');
-    if (attrs.label) labelElement.text(attrs.label);
+
+    function hasSelectHeader() {
+      return el.parent().find('md-select-header').length;
+    }
+
+    function setupLabelElement() {
+      var labelElement = el.find('label');
+      if (!labelElement.length) {
+        labelElement = angular.element('<label>');
+        el.prepend(labelElement);
+      }
+      labelElement.addClass('_md-container-ignore');
+      if (attrs.label) labelElement.text(attrs.label);
+    }
   }
+}
+
+function SelectHeaderDirective() {
+  return {
+    restrict: 'E',
+  };
 }
 
 function SelectProvider($$interimElementProvider) {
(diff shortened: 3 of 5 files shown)


src/components/select/select.js
```diff
@@ -1122,12 +1140,12 @@ function SelectProvider($$interimElementProvider) {
       function watchAsyncLoad() {
         if (opts.loadingAsync && !opts.isRemoved) {
           scope.$$loadingAsyncDone = false;
-          scope.progressMode = 'indeterminate';
+          scope.$mdProgressMode = 'indeterminate';
```
DevVersion: Why are you prefixing that with `$md`?

src/components/select/select.js
```diff
@@ -202,13 +203,19 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
       var valueEl = element.find('md-select-value');
       var isReadonly = angular.isDefined(attr.readonly);
 
+      mdSelectCtrl.mdSelectHeaderInput = element.find('md-select-header').find('input')[0];
```
DevVersion: So the select header is designed to be an input? Wouldn't it make more sense to extract the complete `md-select-header` element?

src/components/select/select.js
DerekLouie: In the `SelectMenuDirective` I changed the scope from `true` to `false`. Adding the `$md` insures the variable does not get clobbered by angular I believe? @jelbourn for clarification?

src/components/select/select.js
DerekLouie: I believe the intention of this directive (the header) is to let the user put in any custom content they desire. The multiple input check I do not believe makes sense in the context of a component header, but does make sense for the rest of the component.

[This upcoming feature](https://github.com/angular/material/issues/7385) is one use case where we would want a second input in the `md-select` directive that this check would clobber.

src/components/select/select.js
ErinCoughlan: It's to make sure that we don't overwrite anything within the user scope (because we are now using the same one). `$md-` variables should not be used in anyone apps.

src/components/select/demoSelectHeader/style.css
```diff
@@ -0,0 +1,20 @@
+.demo-header-searchbox {
+  border: none;
+  outline: none;
+  height: 100%;
+  width: 100%;
+  padding: 0;
+}
+
+.demo-select-header {
+  box-shadow: 0px 1px 0px 0px rgba(0, 0, 0, 0.1), 0px 0px 0px 0px rgba(0, 0, 0,
```
ErinCoughlan: nit: You can just use `0` instead of `0px`

src/components/select/demoSelectHeader/style.css
```diff
@@ -0,0 +1,20 @@
...
+  0.14), 0px 0px 0px 0px rgba(0, 0, 0, 0.12);
+  padding-left:  40px;
+  height: 48px;
+
```
ErinCoughlan: nit: extra newline here

src/components/select/select.js
ErinCoughlan: It's really unclear that this line is _only_ used in the check below. Perhaps move this until right before it's used (where the comment about ignoring the header inputs is).

Alternatively, can we change the selector that is used to find inputs to ignore the select-header? (I haven't looked how input is determined, but it seems like there should be an easier way.)

src/components/select/select.js
```diff
@@ -1176,8 +1194,14 @@ function SelectProvider($$interimElementProvider) {
 
         function onMenuKeyDown(ev) {
           var keyCodes = $mdConstant.KEY_CODE;
-          ev.preventDefault();
-          ev.stopPropagation();
+          // If we have an input the md-select-header we want to allow
+          // users to type in it.
+          if (opts.selectCtrl.mdSelectHeaderInput === ev.target) {
```
ErinCoughlan: I'm unsure you want to leave this early. For example, if I were typing in a searchbox, I'd still want to be able to use the arrow keys to select items. This is a little tricky with maintaining focus, so it's fine if that is a future improvement or comes as part of your search component.

src/components/select/select.js
DerekLouie: Ahh yes, this is reasonable. 

I moved the check to be closer to the usage of it. Previously it was used in another check (for handling `keydown` stuff. I put it on the `mdSelectCtrl` because that's the only controller that was present both here and in the `onMenuKeyDown` function. 

But that code is gone now, so all good!

src/components/select/select.js
DerekLouie: Good point, I'm removed this code and put it in the demo javascript instead.

DerekLouie:
Hello,

I have addressed the aforementioned comments and also taken the keydown logic out of the `md-select` directive code, as to make the header more generic.

Please Review.

src/components/select/select.js
DevVersion: :+1: Yes, I agree. Good point.

src/components/select/demoSelectHeader/index.html
```diff
@@ -0,0 +1,22 @@
...
+        <label>Vegetables</label>
+        <md-select ng-model="selectedVegetables" md-on-close="clearSearchTerm()" multiple>
+          <md-select-header class="demo-select-header">
+            <input ng-model="searchTerm"
+                   type="search"
+                   placeholder="Search for a vegetable.."
+                   class="demo-header-searchbox _md-text">
+          </md-select-header>
+          <md-optgroup label="vegetables">
+            <md-option ng-value="vegetable" ng-repeat="vegetable in vegetables |
```
EricHeath-AI: If I want to perform the search on the server side rather than in memory, then I just do:
```ng-repeat="item in getMatches(searchTerm)"```
similar to https://material.angularjs.org/latest/demo/autocomplete ?
An example with an xhr, or fake xhr might be nice in the docs.

**Comment by ThomasBurleson** [comment]

File: src/components/select/demoSelectHeader/index.html (hunk 6 above)
Comment (DerekLouie) [inline src/components/select/demoSelectHeader/index.html]: Hello Eric,

Solid point, I am actually working on #7385 which will mimic the `autocomplete` directive more closely and will also replace this code.

My intention for this demo was really just to give an example of how the `md-select-header` could be used. I hope it, in no way, serves as an example of what a dedicated search bar component would look like (but that is coming soon :+1: ).

This demo will change after the search component PR is finished.

File: src/components/select/demoSelectHeader/style.css [hunk 7]
```diff
@@ -0,0 +1,19 @@
...
+  box-shadow: 0 1px 0 0 rgba(0, 0, 0, 0.1), 0 0 0 0 rgba(0, 0, 0,
+  0.14), 0 0 0 0 rgba(0, 0, 0, 0.12);
+  padding-left:  40px;
+  height: 48px;
+  cursor: pointer;
+  position: relative;
+  display: flex;
+  align-items: center;
+  width: auto;
+}
```
topherfangio: @DerekLouie The reason this CSS is not applied is because we prepend the `.selectdemoBasicUsage` class to the front of every selector so that it is "scoped" to only this demo.

The problem with this approach is that the menu container which houses the header is actually appended to the body, which is outside of our scoped demo. As far as I know, there is not currently a way to achieve what you want without moving your CSS somewhere more global like the site's `app.css` which I would recommend against.

If you want to do this, perhaps we can update the build system to recognize a `global-styles.css` file in the demos and append that to the generated `styles.css` without the demo scoping. We would just have to be very careful to always add some manual CSS classes (like you've done above) to ensure we don't affect other parts of the app.

src/components/select/select.spec.js
```diff
@@ -333,6 +333,78 @@ describe('<md-select>', function() {
...
+                                '</md-select>' +
+                              '</md-input-container>')($rootScope);
+
+        var optgroupLabel = select[0].querySelector('._md-container-ignore');
+
+        expect(optgroupLabel).toBe(null);
+      }));
+    });
+
+    it('does not allow keydown events to propagate from inside the md-select-menu', inject(function($rootScope, $compile) {
```
topherfangio: I don't see any changes in the code above that would require this new test, so I assume it would pass even without your code changes? If so, can you just briefly explain why it was added?

Not that I'm against better test coverage, I just like to know why it's there :smile:

topherfangio:
If you're adding a new directive, I think we should document it's use. Since it is so simple, it probably does not warrant it's own directive docs, but if you could add something to the Select's docs, I think it would help.

Other than that and my two comments, it looks good to me :+1:

src/components/select/select.js
```diff
@@ -505,11 +510,10 @@ function SelectDirective($mdSelect, $mdUtil, $mdTheming, $mdAria, $compile, $par
 }
 
 function SelectMenuDirective($parse, $mdUtil, $mdTheming) {
-
   return {
     restrict: 'E',
     require: ['mdSelectMenu'],
-    scope: true,
+    scope: false,
```
topherfangio: I lied...one more thought :flushed:

Can you add a quick comment about why this is being changed? If I recall from the Slack conversation, we don't expect it to affect anyone, but it would be good to have a comment in case another dev comes back later and wants to swap it back to `true` to fix something else.

src/components/select/select.spec.js
DerekLouie: Ohh aha, for the sake of the new component (where inputs can be put in the `md-select-header`) I had to override this logic and wrote a test to ensure that.
 
I then noticed there was no test to cover the original keydown logic (which was a larger bit of code) and thought I might add this as a sort of parallel 'negative' to the test I added.

I would be okay with removing this, it was essentially just a copy of the other test I wrote with `not` added to the expects. I will leave it to your discretion :)

src/components/select/demoSelectHeader/style.css
DerekLouie: Hmmm, the build system change sounds great, but a lot of work for the material team so for the sake of everyone, let's table that.

I discovered that I can use a `data-` attribute to add a class to the `menu-container` element. So I think I'll just go that route. Good call adding that hook!

DerekLouie:
Hello I have fixed all the aforementioned comments, please review.

Thank you @topherfangio for debugging the styling issue, it's all good now.

src/components/select/select.js
```diff
@@ -69,6 +70,18 @@ angular.module('material.components.select', [
  *   </md-input-container>
  * </hljs>
  *
+ * With a select-header
```
topherfangio: Can you add a small paragraph here describing potential use-cases and what happens to the original label?

src/components/select/select.js
```diff
@@ -893,6 +912,12 @@ function OptgroupDirective() {
     compile: compile
   };
   function compile(el, attrs) {
+    // If we have a select header element, we don't want to add the normal label
```
topherfangio: Can we refactor this function to separate out the concerns a bit and look more like the following?

```js
function compile(el, attrs) {
  if (!hasSelectHeader()) {
    setupLabelElement();
  }
}
```

I'm hesitant to use the `if (...) return` paradigm unless the condition truly excludes the _purpose_ of the function being run. The concern being that another developer may add something after the label setup not realizing that the header condition's `return` statement would let it be ignored.

src/components/select/select.spec.js
```diff
@@ -333,6 +333,78 @@ describe('<md-select>', function() {
 
     }));
 
+    describe('md-select-header behavior', function() {
+      it('supports rendering a md-select-header', inject(function($rootScope, $compile) {
+        $rootScope.val = [1];
+        var select = $compile('<md-input-container>' +
```
topherfangio: Random note for future reference (you totally don't need to change it here): most code formatters will ignore  the lovely spacing you applied below and align all of the quotation marks up with one another. I've started putting my spacing _inside_ the quotes to avoid this.

**Examples for clarification:**

Instead of

```js
'<div>' +
  'some text' +
'</div>'
```

I generally do this


```js
'<div>' +
'  some text' +
'</div>'
```

It helps me not get frustrated when I do a `Ctrl-Shift-L` later and reformat all that pretty code :smile:

DerekLouie:
Hello,

@topherfangio I have fixed the aforementioned comments, please take another look over at your earliest convenience.

topherfangio:
@DerekLouie Thanks so much for all the updates and putting up with the back-and-forth!

@ThomasBurleson LGTM :+1:

ThomasBurleson:
@topherfangio - thx.

DerekLouie:
Hello @ThomasBurleson 

I changed some css to align the search input's placeholder text and tightened up the height of the select menu.

newheaderstuff

Please merge at your earliest convenience!

Best,

Derek

PR #7790, opened 2016-03-29
fix(sidenav): fix is-locked-open backdrop and incorrect backdrop tests

### LINKED ISSUE #7537: sideNav: auto-disable backdrop when is-locked-open == true.

DevVersion:
* The md-is-locked-open class-selector wasn't valid, because it wasn't probably privatized
* The disabled backdrop tests, weren't valid, because the the sidenav wasn't actually opening.

KUDOS to @EladBezalel for his unbeatable help!

Fixes #7537


ErinCoughlan:
LGTM. Nice catch.

PR #7791, opened 2016-03-29
feat(colors): directive to use any color from any palette everywhere

### LINKED ISSUE #1269: colors and hues should be available everywhere.

EladBezalel:
- added directive and documentation
- added tests
- added smart watch option, when the directive recognize at least one of it's values needs interpolation it starts to watching it

fixes #1269

### DIFF
--- a/src/components/colors/colors.js
+++ b/src/components/colors/colors.js
@@ -1,0 +1,247 @@
+(function () {
+  "use strict";
+
+  /**
+   *  Use a RegExp to check if the `md-colors="<expression>"` is static string
+   *  or one that should be observed and dynamically interpolated.
+   */
+  var STATIC_COLOR_EXPRESSION = /^{((\s|,)*?["'a-zA-Z-]+?\s*?:\s*?('|")[a-zA-Z0-9-.]*('|"))+\s*}$/;
+  var COLOR_PALETTES = undefined;
+
+  /**
+   * @ngdoc module
+   * @name material.components.colors
+   *
+   * @description
+   * Define $mdColors service and a `md-colors=""` attribute directive
+   */
+  angular
+    .module('material.components.colors', ['material.core'])
+    .directive('mdColors', MdColorsDirective)
+    .service('$mdColors', MdColorsService);
+
+  /**
+   * @ngdoc service
+   * @name $mdColors
+   * @module material.core.theming.colors
+   *
+   * @description
+   * With only defining themes, one couldn't get non ngMaterial elements colored with Material colors,
+   * `$mdColors` service is used by the md-color directive to convert the 1..n color expressions to RGBA values and will apply
+   * those values to element as CSS property values.
+   *
+   *  @usage
+   *  <hljs lang="html">
+   *    <div md-colors="{background: 'myTheme-accent-900-0.43'}">
+   *      <div md-colors="{color: 'red-A100', border-color: 'primary-600'}">
+   *        <span>Color demo</span>
+   *      </div>
+   *    </div>
+   *  </hljs>
+   *
+   */
+  function MdColorsService($mdTheming, $mdColorPalette, $mdUtil, $parse) {
+    COLOR_PALETTES = COLOR_PALETTES || Object.keys($mdColorPalette);
+
+    // Publish service instance
+    return {
+      applyThemeColors: applyThemeColors,
+      getThemeColor: getThemeColor
+    };
+
+    // ********************************************
+    // Internal Methods
+    // ********************************************
+
+    /**
+     * Convert the color expression into an object with scope-interpolated values
+     * Then calculate the rgba() values based on the theme color parts
+     */
+    function applyThemeColors(element, scope, colorExpression) {
+      // Json.parse() does not work because the keys are not quoted;
+      // use $parse to convert to a hash map
+      var themeColors = $parse(colorExpression)(scope);
+
+      // Assign the calculate RGBA color values directly as inline CSS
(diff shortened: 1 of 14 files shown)


src/components/colors/colors.js
```diff
@@ -0,0 +1,88 @@
...
+ * </hljs>
+ *
+ */
+
+function mdColorsDirective($mdColorPalette, $mdTheming) {
+  return {
+    restrict: 'A',
+    link: function (scope, elem, attrs) {
+      scope.$watch(attrs.mdColors, function (val) {
+        Object.keys(val).forEach(function (key) {
```
crisbeto: You could spare calling `Object.keys` and iterating it by using `angular.forEach(function(value, key){ ... });`

src/components/colors/colors.js
```diff
@@ -0,0 +1,88 @@
...
+          if (Object.keys($mdColorPalette).indexOf(split[1]) !== -1) {
+            palette += '-' + split.splice(1, 1);
+          }
+          else {
+            // Two word palette can be also written in camelCase, forming camelCase to dash-case
+            palette = palette.replace(/([a-z])([A-Z])/g, '$1-$2').toLowerCase();
+          }
+
+          // If the palette is not in the palette list it's one of primary/accent/warn/background
+          if (Object.keys($mdColorPalette).indexOf(palette) === -1) {
```
crisbeto: You're calling `Object.keys` on `$mdColorPalette` twice in the same function, you could call it once and save it in a variable.

src/components/colors/colors.js
```diff
@@ -0,0 +1,88 @@
...
+            var scheme = $mdTheming.THEMES[theme].colors[palette];
+
+            if (!scheme) {
+              return console.error('mdColors: couldn\'t find \'' + palette + '\' in the palettes.');
```
crisbeto: Not really a big deal, but you could use `$log.error` in here.

src/components/colors/colors.js
```diff
@@ -0,0 +1,88 @@
...
+ *   </div>
+ * </hljs>
+ *
+ */
+
+function mdColorsDirective($mdColorPalette, $mdTheming) {
+  return {
+    restrict: 'A',
+    link: function (scope, elem, attrs) {
+      scope.$watch(attrs.mdColors, function (val) {
```
DevVersion: Just a question, why are you not using `$observe`, are there problems with the interpolating?

src/components/colors/colors.js
EladBezalel: I encountered today something even more beautiful!
``` js
for (key in val) {
 //...
}
```

src/components/colors/colors.js
ThomasBurleson: We should **never** use `console.error()` or the `console.<fn>` methods... always use `$log.<fn>`.

src/components/colors/colors.js
ThomasBurleson: Also `.$watch( ... )` should be consider an performance **expensive** solution.

src/components/colors/demoBasicUsage/script.js
```diff
@@ -0,0 +1,19 @@
+(function () {
```
crisbeto: I believe the IIFE isn't necessary in here, you can inline the `DemoCtrl`.

ErinCoughlan:
@gmoothart Is anything missing here that we could have used for tabs?

ThomasBurleson:
@EladBezalel - The demo needs some text to describe the demo, the intent, and the power of mdColors.

ThomasBurleson:
In a right-aligned column (in the color boxe), please show the hex color values. e.g. `#2196F3`.

screen shot 2016-04-03 at 6 00 27 pm

ThomasBurleson:
Please refactor the `colors.spec.js` to reduce the non-DRY code.

topherfangio:
@EladBezalel If we introduce this API, I can see developers wanting to use it in many places/components. Can you test the following components to make sure that it is compatible:

- Toast (particularly, can you easily change the background color)
- Icon
- Divider
- Subheader
- and Tooltip

If it doesn't work out of the box, we may need to update some of the components to make this easier.

ThomasBurleson:
@EladBezalel - see refactored [`colors.spec.js`](https://gist.github.com/ThomasBurleson/979777137ddebe755c8ba0a7ee71cacd)

EladBezalel:
- [ ] Toast (particularly, can you easily change the background color)
- [x] Icon
- [x] Divider
- [x] Subheader
- [ ] Tooltip

toast and tooltip have elements containing the styles

src/core/services/theming/theming.js
```diff
@@ -61,8 +61,9 @@ var GENERATED = { };
 
 // In memory storage of defined themes and color palettes (both loaded by CSS, and user specified)
 var PALETTES;
-var THEMES;
 
+// Text Colors on light and dakr backgrounds
```
EladBezalel: Dark

PR #7798, opened 2016-03-30
update(docs): fix mdButtonGroup style for layout-sm

### LINKED ISSUE #2907: Feature request: md-button-group
I was looking for Angular Material implementation of buttons that are grouped together and can be toggled exclusively or not.
Here's an example of how bootstrap implements the feature:
http://getbootstrap.com/components/#btn-groups
I could not find such Angular Material nor find a spec for that behavior in the Google Material Design guidelines.
Is there a better way of implementing such a functionality or is it a missing feature?

Thanks.

leocaseiro:
Remove `layout-sm="column"` for layout-sm as border radius left/right don't look good as `column`.

Changed from
```html
<section layout="row" layout-sm="column" layout-align="center center">
```

To this:

```html
<section layout="row" layout-align="center center">
```

Closes #2907

### DIFF
--- a/docs/content/CSS/button.md
+++ b/docs/content/CSS/button.md
@@ -232,8 +232,8 @@ Here is another example of a button with font icons:
 Using the customization approaches documented above, we can easily create a mdButtonGroup:
 
 <hljs lang="html">
-<section layout="row" layout-sm="column" 
-         layout-align="center center" >
+<section layout="row"
+         layout-align="center center">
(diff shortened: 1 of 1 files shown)


leocaseiro:
Hi @ThomasBurleson, I would like to implement it as a directive, like `mdButtonGroup`.
I have some ideas using probably CSS only and I'd follow trully the [specs](https://www.google.com/design/spec/components/buttons.html#buttons-toggle-buttons)

PR #7804, opened 2016-03-30
perf(md-menu): remove unneeded watcher

clshortfuse:
isOpen is an internal value that doesn't need to be checked every digest by using angular watchers.

I tried to fix the old PR, but I gave up because I'm not too familiar with rebasing with git.
https://github.com/angular/material/pull/4463

### DIFF
--- a/src/components/menu/js/menuController.js
+++ b/src/components/menu/js/menuController.js
@@ -31,7 +31,9 @@ function MenuController($mdMenu, $attrs, $element, $scope, $mdUtil, $timeout, $r
 
     menuContainer.on('$mdInterimElementRemove', function() {
       self.isOpen = false;
+      $mdUtil.nextTick(function(){ self.onIsOpenChanged(self.isOpen);});
     });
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
Solution is too complicated without justification. What are the specific performance gains?

clshortfuse:
Okay, here are the code pens:

Here's 1.0.7 which is the standard way of using a watchers per md-menu: http://codepen.io/shortfuse/pen/LNOdvW

In my table of 1000 bind-once rows, it increments the watcher count from 13 to 1000 (1 per row).

On my fork which doesn't use watchers, the watcher count stays the same at 13. 
http://codepen.io/shortfuse/pen/RajMOx

I set it to a high number to illustrate the performance effects. Please disregard any CSS changes in my fork.

ThomasBurleson:
@topherfangio - Seems like a reasonable perf improvement for scenarios where menus [as row options]  are used in tables. Please review.

topherfangio:
@ThomasBurleson LGTM; I tested the demos and ran through the code. If it causes a breaking change somehow, I can't see it.

PR #7808, opened 2016-03-30
feat(autocomplete): add the possibility to specify extra classes for the input

### LINKED ISSUE #7370: Angular Material md-input-class
I think it would be great if we can add or manipulate the class on the input produced by the md-autocomplete

crisbeto:
* Adds the ability to specify classes on the `input` element via the `md-input-class` attribute.
* Avoids some redundancy in the `getInputElement` function by abstracting away attributes that are repeated twice.

Closes #7370.


src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -228,66 +230,54 @@ function MdAutocomplete () {
...
-                  aria-owns="ul-{{$mdAutocompleteCtrl.id}}"\
-                  ' + (attr.mdSelectOnFocus != null ? 'md-select-on-focus=""' : '') + '\
-                  aria-label="{{floatingLabel}}"\
-                  aria-autocomplete="list"\
-                  aria-haspopup="true"\
-                  aria-activedescendant=""\
-                  aria-expanded="{{!$mdAutocompleteCtrl.hidden}}"/>\
-              <div md-autocomplete-parent-scope md-autocomplete-replace>' + leftover + '</div>\
-            </md-input-container>';
+          return '' +
```
DevVersion: This is unnecessary. Even for formatting.

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -228,66 +230,54 @@ function MdAutocomplete () {
...
+            '<md-input-container flex ng-if="floatingLabel"> \n' +
```
DevVersion: Since when, is a line break added here?

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -228,66 +230,54 @@ function MdAutocomplete () {
       }
 
       function getInputElement () {
+        var commonInputAttrs =
```
DevVersion: This seems to be a good idea, it's more clear :+1:

src/components/autocomplete/js/autocompleteDirective.js
crisbeto: Are you referring to the newline?

src/components/autocomplete/js/autocompleteDirective.js
crisbeto: I added it so it's consistent with the rest.

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -228,66 +230,54 @@ function MdAutocomplete () {
...
+          'name="{{inputName}}" \n' +
+          'autocomplete="off" \n' +
+          'ng-required="$mdAutocompleteCtrl.isRequired" \n' +
+          'ng-disabled="$mdAutocompleteCtrl.isDisabled" \n' +
+          'ng-model="$mdAutocompleteCtrl.scope.searchText" \n' +
+          'ng-readonly="$mdAutocompleteCtrl.isReadonly" \n' +
+          'ng-keydown="$mdAutocompleteCtrl.keydown($event)" \n' +
+          'ng-blur="$mdAutocompleteCtrl.blur()" \n' +
+          'ng-focus="$mdAutocompleteCtrl.focus()" \n' +
+          'ng-class="::inputClass" \n' +
```
DevVersion: Wouldn't it make more sense, to remove the one-time binding?

src/components/autocomplete/js/autocompleteDirective.js
DevVersion: No, I mean, this empty string at the start of the concatenation is unnecessary. Just remove it, and concat as before. But caution: the string needs to start at the same line as the `return` statement.

src/components/autocomplete/js/autocompleteDirective.js
crisbeto: I'd say so. I added it mainly because the extra menu class did the same, but it would probably be more useful if it was a live binding.

src/components/autocomplete/js/autocompleteDirective.js
DevVersion: Oh, this is weird. I've never seen this before. It's also not unnecessary, because its just a template.

crisbeto: I meant so it's consistent with the way I did it. I replaced those escaped strings because they cause jshint warnings.

DevVersion: Sorry, are we talking about the same thing? :smile: I mean, those `\n` line-breaks are not necessary in an template and are never used in other components.

crisbeto: Nah I was referring to concatenating instead of escaping stuff. Otherwise you're right about the newlines, they just make it a little more convenient if uou need to log something.

Issue #7703, opened 2016-03-22
mdPanel: Add default fade animation.


ErinCoughlan:
This task involves:

* Adding default CSS class for fade animation (md-panel-animate-fade).
* Setting the CSS class for transitions.
* Using the animator from $mdUtil.dom.animator.

Issue #7702, opened 2016-03-22
mdPanel: Add default scale animation.


ErinCoughlan:
This task involves:

* Adding default CSS class for scale animation (md-panel-animate-scale).
* Setting the CSS class for transitions.
* Using the animator from $mdUtil.dom.animator.

PR #7809, opened 2016-03-30
feat(panel): Add default slide/scale/fade/none animations and custom animations and withAnimation API method.

### LINKED ISSUE #7701: mdPanel: Add default translate animation and withAnimation API method.
This task involves:

* Implementing the following API method: `withAnimation(string): MdPanelAnimation`
* Add default CSS class for translate animation (`md-panel-animate-translate`).
* Setting the CSS class for transitions.
* Use the animator from `$mdUtil.dom.animator`.

Note: translation is the animation for mdDialog, so we can reference code [here](https://github.com/angular/material/blob/32951a7d1b17bb62938f3740d71504a70cf01f8d/src/components/dialog/dialog.js#L1004).

KarenParker:

Please review.

Fixes #7701, #7702, and #7703

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -1689,7 +1689,11 @@ MdPanelAnimation.prototype.animateOpen = function(panelEl, animator) {
   var closeTo = animator.toTransformCss("");
 
   switch (this._animationClass) {
+<<<<<<< 268079b85185013f1683932a5814d9207a0d0fd2
     case 'md-panel-animate-slide':
+=======
+    case MdPanelAnimation.animation.SLIDE:
+>>>>>>> feat(panel): Add default animations and withAnimation API method.
       animationOptions = {
         transitionInClass: '_md-panel-animate-slide-in _md-panel-shown',
         transitionOutClass: '_md-panel-animate-slide-out'
@@ -1703,7 +1707,11 @@ MdPanelAnimation.prototype.animateOpen = function(panelEl, animator) {
       closeTo = animator.toTransformCss(animator.calculateSlideToOrigin(
           panelEl, this._closeTo));
       break;
+<<<<<<< 268079b85185013f1683932a5814d9207a0d0fd2
     case 'md-panel-animate-scale':
+=======
+    case MdPanelAnimation.animation.SCALE:
+>>>>>>> feat(panel): Add default animations and withAnimation API method.
       animationOptions = {
         transitionInClass: '_md-panel-animate-scale-in _md-panel-shown',
         transitionOutClass: '_md-panel-animate-scale-out'
@@ -1717,7 +1725,11 @@ MdPanelAnimation.prototype.animateOpen = function(panelEl, animator) {
       closeTo = animator.toTransformCss(animator.calculateZoomToOrigin(
           panelEl, this._closeTo));
       break;
+<<<<<<< 268079b85185013f1683932a5814d9207a0d0fd2
     case 'md-panel-animate-fade':
+=======
+    case MdPanelAnimation.animation.FADE:
+>>>>>>> feat(panel): Add default animations and withAnimation API method.
       animationOptions = {
         transitionInClass: '_md-panel-animate-fade-in _md-panel-shown',
         transitionOutClass: '_md-panel-animate-fade-out'


src/components/panel/panel.js
```diff
@@ -1010,6 +1028,15 @@ MdPanelRef.prototype._addStyles = function() {
     return; // Don't setup positioning.
   }
 
+  this._$rootScope.$applyAsync(angular.bind(this, this._configurePosition));
+}
+
+
+/**
+ *
```
jelbourn: Add description?

src/components/panel/panel.js
```diff
@@ -1563,22 +1624,161 @@ MdPanelAnimation.prototype.closeTo = function(closeTo) {
...
+ *   `md-panel-animate-slide`: The panel slides in and out from the specified
+ *        elements.
+ *   `md-panel-animate-scale`: The panel scales in and out.
+ *   `md-panel-animate-fade`: The panel fades in and out.
+ *
+ * @param {string|!{open: string, close: string}} cssClass
+ * @returns {MdPanelAnimation}
+ */
+
+MdPanelAnimation.prototype.withAnimation = function(cssClass) {
```
jelbourn: What do you think about the built-in animations being available as something like
`.withAnimation($mdPanel.animation.Slide)`

This way the API isn't tightly coupled to always being just one css class; I might want to later add, for example, an animation that flies a few loops around the page before landing.

src/components/panel/panel.js
```diff
@@ -1563,22 +1624,161 @@ MdPanelAnimation.prototype.closeTo = function(closeTo) {
...
+/**
+ * Set the height and width to match the panel if not provided.
+ * @param {!angular.JQLite} panelEl
+ * @private
+ */
+MdPanelAnimation.prototype._fixBounds = function(panelEl) {
+  var panelWidth = panelEl[0].offsetWidth;
+  var panelHeight = panelEl[0].offsetHeight;
+
+  if (angular.isUndefined(this._openFrom.bounds.height)) {
```
jelbourn: Maybe just use `== null`, which is shorter and captures both `null` and `undefined`?

src/components/panel/panel.spec.js
```diff
@@ -1,5 +1,5 @@
-describe('$mdPanel', function() {
-  var $mdPanel, $rootScope, $rootEl, $templateCache, $q;
+fdescribe('$mdPanel', function() {
```
jelbourn: `fdescribe`

KarenParker:
Comments addressed.

KarenParker:
Any more comments?

src/components/panel/panel.js
```diff
@@ -1563,36 +1636,161 @@ MdPanelAnimation.prototype.closeTo = function(closeTo) {
...
+
+
+/**
+ * Animate the panel open.
+ * @param {!angular.JQLite} panelEl
+ * @param animator
+ * @returns {!angular.$q.Promise}
+ */
+MdPanelAnimation.prototype.animateOpen = function(panelEl, animator) {
+  this._fixBounds(panelEl);
```
jelbourn: optional: this function is pretty long and could benefit from being broken into smaller functions

src/components/panel/panel.scss
```diff
@@ -22,3 +18,51 @@
...
+
+  &._md-panel-shown {
+    // Only used when custom animations are present.
+    // Overridden by the default animations.
+    opacity: 1;
+    transition: none;
+
+    &._md-panel-animate-slide-in {
+      opacity: 1;
+      transition: $swift-ease-out, opacity 1ms;
```
jelbourn: Why aren't any of these using the material design easing curves? Might want to check with the material implementation list if they have recommended easing curves / durations for these use case. (can do in a follow-up)

jelbourn:
LGTM

src/components/panel/panel.js
```diff
@@ -518,6 +518,28 @@ angular
...
+ *   SLIDE: The panel slides in and out from the specified
+ *       elements. It will not fade in or out.
+ *   SCALE: The panel scales in and out. Slide and fade are
+ *       included in this animation.
+ *   FADE: The panel fades in and out.
+ *
+ * Custom classes will by default fade in and out unless
+ * "transition: opacity 1ms" is added to the to custom class.
+ *
+ * @param {string|!{open: string, close: string}} cssClass
```
ErinCoughlan: nit: You don't need the `!` here, you defined the entire object.

src/components/panel/panel.js
```diff
@@ -1127,7 +1155,7 @@ MdPanelRef.prototype._configureClickOutsideToClose = function() {
 MdPanelRef.prototype._configureTrapFocus = function() {
   // Focus doesn't remain instead of the panel without this.
   this._panelEl.attr('tabIndex', '-1');
-  if (this._config.trapFocus) {
+  if (this._config['trapFocus']) {
```
ErinCoughlan: I wonder if we want to define a `config` object somewhere in this file. It would certainly make it easier to know types and create externs later.

src/components/panel/panel.scss
KarenParker: Isn't the $swift-ease-out the material design ones? They are the ones used in dialog.

src/components/panel/panel.js
```diff
@@ -1497,29 +1565,34 @@ MdPanelPosition.prototype._calculatePanelPosition = function(panelEl) {
...
-  /** @private {!{element: !angular.JQLite|undefined, bounds: !DOMRect}} */
-  this._closeTo = {};
-}
+  /**
+   * @private {!{element: !angular.JQLite|undefined, bounds: !DOMRect}|
+   *    undefined}
+   */
+  this._closeTo;
 
+  /** @private {string|!{open: string, close: string} */
```
ErinCoughlan: nit: You're missing a closing `}` at the end.

src/components/panel/demoPanelAnimations/script.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+  switch(this.animationType) {
+    case 'custom':
+      animation.withAnimation({
+        open: 'demo-dialog-custom-animation-open',
+        close: 'demo-dialog-custom-animation-close'
+      });
+      break;
+    case 'slide':
+      animation.withAnimation(MdPanelAnimation.animation.SLIDE);
```
ErinCoughlan: Should these be available on `$mdPanel` since that's the users actually have access to?

src/components/panel/panel.js
```diff
@@ -1563,36 +1636,161 @@ MdPanelAnimation.prototype.closeTo = function(closeTo) {
...
+  this._fixBounds(panelEl);
+  var animationOptions = {};
+  var reverseAnimationOptions = {};
+  var openFrom = animator.toTransformCss("");
+  var openTo = animator.toTransformCss("");
+  var closeFrom = animator.toTransformCss("");
+  var closeTo = animator.toTransformCss("");
+
+  switch (this._animationClass) {
+    case 'md-panel-animate-slide':
```
ErinCoughlan: Use the enum you defined above for these strings.

src/components/panel/demoPanelAnimations/script.js
KarenParker: It would cause a circular dependency.

PR #7811, opened 2016-03-31
test(panel): Updated positioning tests with a custom matcher.

ErinCoughlan:
This reduces some flakiness around different viewports and rounding values with getBoundingClientRect.

@KarenParker @jelbourn @gmoothart @DerekLouie Please review.

### DIFF
--- a/src/components/panel/panel.spec.js
+++ b/src/components/panel/panel.spec.js
@@ -31,6 +31,38 @@ describe('$mdPanel', function() {
     attachToBody($rootEl);
   });
 
+  beforeEach(function() {
+    jasmine.addMatchers({
+      /**
+       * Asserts that two values are within a range of each other. This is
+       * used for testing relative positioning. If no range is set, defaults
+       * to 1.
+       *
+       * Example Use:
+       * expect(panelTop).toBeApproximately(buttonTop, 1);
+       */
+      toBeApproximately: function() {
+        return {
+          compare: function(actual, expected, opt_epsilon) {
+            var epsilon = opt_epsilon || 1;
+
+            var actualNumber = parseFloat(actual);
+            var expectedNumber = parseFloat(expected);
+
+            var pass = Math.abs(expectedNumber - actualNumber) < epsilon;
+            var not = pass ? 'not ' : '';
+
+            return {
+              pass: pass,
+              message: 'Expected ' + expected + not + 'to be within ' +
+                  epsilon + ' of ' + actual
+            }
+          }
+        }
+      }
+    });
+  });
+
   afterEach(function() {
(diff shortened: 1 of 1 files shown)


src/components/panel/panel.spec.js
```diff
@@ -31,6 +31,41 @@ describe('$mdPanel', function() {
...
+       * used for testing relative positioning. If no range is set, defaults
+       * to 1.
+       *
+       * Example Use:
+       * expect(panelTop).toBeApproximately(buttonTop, 1);
+       */
+      toBeApproximately: function() {
+        return {
+          compare: function(actual, expected, opt_offset) {
+            var offset = opt_offset || 1;
```
jelbourn: optional: the traditional term for this is `epsilon`

src/components/panel/panel.spec.js
```diff
@@ -31,6 +31,41 @@ describe('$mdPanel', function() {
...
+    jasmine.addMatchers({
+      /**
+       * Asserts that two values are within a range of each other. This is
+       * used for testing relative positioning. If no range is set, defaults
+       * to 1.
+       *
+       * Example Use:
+       * expect(panelTop).toBeApproximately(buttonTop, 1);
+       */
+      toBeApproximately: function() {
```
jelbourn: How is this different from `toBeCloseTo` ?

KarenParker:
LGTM

src/components/panel/panel.spec.js
ErinCoughlan: `toBeCloseTo` compares floats with a certain amount of decimal precision. I want to check that two numbers are within `x` of each other.

For instance, `expect(54.99).toBeCloseTo(55.01)` fails. With `toBeApproximately` this passes.

src/components/panel/panel.spec.js
```diff
@@ -31,6 +31,41 @@ describe('$mdPanel', function() {
...
+          compare: function(actual, expected, opt_offset) {
+            var offset = opt_offset || 1;
+
+            var actualNumber = parseFloat(actual);
+            var expectedNumber = parseFloat(expected);
+
+            var rangeStart = expectedNumber - offset;
+            var rangeEnd = expectedNumber + offset;
+
+            var pass = actualNumber >= rangeStart && actualNumber <= rangeEnd;
```
jelbourn: The typical way to do this is `Math.abs(num1 - num2) < epsilon`

ErinCoughlan:
Thanks @jelbourn I was trying to figure out why this looked so strange yesterday.

I've fixed all the comments. PTAL.

ErinCoughlan:
Merged.

PR #7818, opened 2016-03-31
update(panel): Update positioning to be calculated after the panel is rendered.

KarenParker:
@jelbourn @gmoothart @ErinCoughlan @DerekLouie 

Please review.

Position was being calculated base on panel size before it was being rendered.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -1010,6 +1010,15 @@ MdPanelRef.prototype._addStyles = function() {
     return; // Don't setup positioning.
   }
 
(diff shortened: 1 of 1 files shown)


ErinCoughlan:
LGTM. Nice catch.

src/components/panel/panel.js
```diff
@@ -1010,6 +1010,15 @@ MdPanelRef.prototype._addStyles = function() {
     return; // Don't setup positioning.
   }
 
+  this._$rootScope.$applyAsync(angular.bind(this, this._configurePosition));
+}
```
ErinCoughlan: Missing semicolon.

PR #7822, opened 2016-03-31
perf(md-tooltip) use MutationObserver instead of watcher if available

### LINKED ISSUE #4033: (perf)md-tooltip uses angular watcher per item

clshortfuse:
use watcher on mdVisible only if being used

fixes #4033

improved version of #4441

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -92,6 +92,11 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
       content.css('transform-origin', origin);
     }
 
+    function onVisibleChanged (isVisible) {
+      if (isVisible) showTooltip();
+      else hideTooltip();
+    }
+
     function configureWatchers () {
       scope.$on('$destroy', function() {
         scope.visible = false;
@@ -99,12 +104,30 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
         angular.element($window).off('resize', debouncedOnResize);
       });
 
-      scope.$watch('visible', function (isVisible) {
-        if (isVisible) showTooltip();
-        else hideTooltip();
-      });
+      if (element[0] && 'MutationObserver' in $window) {
+        var attributeObserver = new MutationObserver(function(mutations) {
+          mutations
+            .forEach(function (mutation) {              
+              if (mutation.attributeName === 'md-visible') {
+                if (!scope.visibleWatcher)
+                  scope.visibleWatcher = scope.$watch('visible', onVisibleChanged );
+              }
+              if (mutation.attributeName === 'md-direction') {
+                updatePosition(scope.direction);
+              }
+            });
+        });
 
-      scope.$watch('direction', updatePosition );
+        attributeObserver.observe(element[0], { attributes: true});
+
+        if (attr.hasOwnProperty('mdVisible')) // build watcher only if mdVisible is being used
+          scope.visibleWatcher = scope.$watch('visible', onVisibleChanged );
+
+      }
+      else { // MutationObserver not supported
+        scope.visibleWatcher = scope.$watch('visible', onVisibleChanged );
+        scope.$watch('direction', updatePosition );
+      }
     }
 
     function addAriaLabel () {
@@ -194,9 +217,15 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
           $timeout(function() {
             scope.visible = setVisible.value;
             setVisible.queued = false;
+            if (!scope.visibleWatcher)
+              onVisibleChanged(scope.visible);
           }, scope.delay);
         } else {
-          $mdUtil.nextTick(function() { scope.visible = false; });
+          $mdUtil.nextTick(function() { 
+            scope.visible = false; 
+            if (!scope.visibleWatcher)
+              onVisibleChanged(false);
+          });
         }
       }
     }


src/components/tooltip/tooltip.js
```diff
@@ -92,19 +92,45 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
...
 
-    function configureWatchers () {
+    function onVisibleChanged(isVisible)
+    {
+      if (isVisible) showTooltip();
+      else hideTooltip();
+    }
+
+    function configureWatchers ()
+    {
```
DevVersion: Braces should be in the same line.

clshortfuse:
I believe I don't have to call scope.visibleWatcher() in onDestroy, but somebody correct me if I'm wrong.

src/components/tooltip/tooltip.js
```diff
@@ -99,12 +104,30 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
         angular.element($window).off('resize', debouncedOnResize);
       });
 
-      scope.$watch('visible', function (isVisible) {
-        if (isVisible) showTooltip();
-        else hideTooltip();
-      });
+      if (element[0] && 'MutationObserver' in $window) {
```
ThomasBurleson: Why use `MutationObserver` instead of `$attrs.$observe( )` ?

ThomasBurleson:
Solution appears to be complex. What are the justifications and perf gains?

clshortfuse:
I'll write a codepen to show the gains of this and the md-menu change.

I use this in a production environment. With the solution I built, virtual repeat is not possible, and I have a table with 100s of rows and an md-tooltip on icons and an md-menu item on each row. Essentially, a 500 row table will have 1/4x the amount of watchers, going from 2000 down to 500 with no features lost. Watchers are processed every digest or apply and 2000 watchers will bring even an i7 to its knees.

The important thing is to make sure that no features are lost, which I think the code is fine in this regard.

clshortfuse:
A short explanation of the code is as follows:

md-direction is a raw value tied to an attribute in the md-tooltip element. Since this variable is raw text, we can use MutationObserver to get a "push notification" from the browser when it changes instead of looping every digest to check if the attribute changed (what Angular does)

md-visible is a reference value that needs to be checked by angular. Because tying md-visible to a value is optional in md-tooltip, the code will only use a watcher if the attribute is being used. If the attribute is not set at runtime, a MutationObserver is used to receive a "push notification" if the user decides to start using the attribute, which then creates the watcher.

src/components/tooltip/tooltip.js
clshortfuse: While $observe is technically faster than $watch since it still since it's a string comparison vs an object/expression evaluation, both are checked synchronously every $digest. 

MutationObserver is handled by the browser and signaled asynchronously.

clshortfuse:
Okay, here are the code pens:

Here's 1.0.7 which is the standard way of using two watchers per tooltip: http://codepen.io/shortfuse/pen/LNOdvW

In my table of 1000 bind-once rows, it increments the watcher count from 13 to 4000 (2 per row).

On my fork which doesn't use watchers, the watcher count stays the same at 13. 
http://codepen.io/shortfuse/pen/RajMOx

I set it to a high number to illustrate the performance effects. Please disregard any CSS changes in my fork since.

ThomasBurleson:
@jelbourn - please review. Seems like a reasonable perf improvement for tooltips in large tables.

src/components/tooltip/tooltip.js
```diff
@@ -92,19 +92,42 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
       content.css('transform-origin', origin);
     }
 
+    function onVisibleChanged (isVisible) {
+      if (isVisible) showTooltip();
```
jelbourn: Missing braces

src/components/tooltip/tooltip.js
```diff
@@ -92,19 +92,42 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
...
-        if (isVisible) showTooltip();
-        else hideTooltip();
-      });
+      if (element[0] && 'MutationObserver' in $window) {
+        var attributeObserver = new MutationObserver(function(mutations) {
+          mutations
+            .forEach(function (mutation) {              
+              if (mutation.attributeName === 'md-visible') {
+                if (!scope.visibleWatcher)
+                  scope.visibleWatcher = scope.$watch('visible', onVisibleChanged );
```
jelbourn: Can you end up adding multiple watchers for the same thing here?

clshortfuse: I use the variable 'visibleWatcher' for this reason. If the value is undefined/null, then there is no watcher. If it isn't null, then it means the object already has a watcher and a second shouldn't be created.

Issue #6581, opened 2016-01-07
Pressing esc while in autocomplete loses focus


calebegg:
If you press 'esc' while typing in an autocomplete field, focus is lost (goes to the top of the page).

This is on Chrome/Ubuntu

smurugavel:
Yes. I see this problem as well

clshortfuse:
Related https://github.com/angular/material/issues/5196

Issue #6534, opened 2016-01-04
md-chips/autocomplete: pressing Esc to close autocomplete list removes focus from input


csvan:
For example, consider the "vegetables" example on the Demo (last under Custom Inputs):

https://material.angularjs.org/latest/demo/chips

To reproduce here:

1. select the input element
2. start typing so that the autocomplete list appears
3. press Esc

**Expected**: the autocomplete list should close, and I should be able to keep typing or otherwise interact with the input element.

**Actual**: focus is removed from the element entirely - user has to manually select it again in order to interact with it.

I am not sure if this is actually by design. However, the expected behavior above is most likely what most users will be expecting as well.

tjlav5:
@csvan would you rather have to press ESC twice to trigger the blur? Once for the autocomplete form, and a second time for the input field?

clshortfuse:
Related https://github.com/angular/material/issues/5196

csvan:
@tjlav5 I am sorry for the late reply. https://github.com/angular/material/issues/5196 perfectly sums up what I would expect from it.

PR #7825, opened 2016-03-31
feat(md-autocomplete) make clear and blur on escape keydown optional

clshortfuse:
fixes:

https://github.com/angular/material/issues/5196
https://github.com/angular/material/issues/6534
https://github.com/angular/material/issues/6581

improves PR: 

https://github.com/angular/material/pull/6707

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -466,12 +466,20 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         select(ctrl.index);
         break;
       case $mdConstant.KEY_CODE.ESCAPE:
+        if (!shouldProcessEscape()) return;
         event.stopPropagation();
         event.preventDefault();
-        if ($scope.searchText) clearValue();
 
-        // Force the component to blur if they hit escape
-        doBlur(true);
+        clearValue(shouldClearOnEscape() && $scope.searchText);
+
+        if (shouldBlurOnEscape()) {
+          // Force the component to blur if they hit escape
+          doBlur(true);
+        }
+        else {
+          // Manually hide (needed for mdNotFound support)
+          ctrl.hidden = true;
+        }
 
         break;
       default:
@@ -549,6 +557,38 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     else if (hasSelection()) return true;           // Hide if there is already a selection
     else if (!hasFocus) return true;                // Hide if the input does not have focus
     else return !shouldShow();                      // Defer to standard show logic
+  }
+
+  /**
+   * Determines if the escape keydown should be processed
+   * @returns {boolean}
+   */
+  function shouldProcessEscape() {
+    if (shouldBlurOnEscape()) return true;                       // Needs blur
+    if (!ctrl.hidden) return true;                               // Hide dropdown
+    if (ctrl.loading) return true;                               // Cancel loading
+    if (shouldClearOnEscape() && $scope.searchText) return true; // Clear Text
+    return false;
+  }
+
+  /**
+   * Determines if the escape key should blur focus
+   * @returns {boolean}
+   */
+  function shouldBlurOnEscape() {
+    if (!$scope.escapeOptions) return true;                     // default true
+    if ($scope.escapeOptions.toLowerCase().indexOf('blur') >= 0) return true; // blur is requested
+    return false;
+  }
+
+  /**
+   * Determines if the escape key should clear input
+   * @returns {boolean}
+   */
+  function shouldClearOnEscape() {
+    if (!$scope.escapeOptions) return true;                      // default true
+    if ($scope.escapeOptions.toLowerCase().indexOf('clear') >= 0) return true; // blur is requested
+    return false;
   }
 
   /**
@@ -639,7 +679,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   /**
    * Clears the searchText value and selected item.
    */
-  function clearValue () {
+  function clearValue (shouldClearText) {
     // Set the loading to true so we don't see flashes of content.
     // The flashing will only occour when an async request is running.
     // So the loading process will stop when the results had been retrieved.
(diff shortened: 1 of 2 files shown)


clshortfuse:
I need to fix something in the code. Escape should close the drop down if it's open first and then pass up the event.

ThomasBurleson:

clshortfuse:
ready for review

clshortfuse:
ready for review

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -441,12 +441,21 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         select(ctrl.index);
         break;
       case $mdConstant.KEY_CODE.ESCAPE:
+        // Only return if blur is disabled
```
topherfangio: Can we refactor this into a helper function? Lots of logic to try to understand within a switch statement 😄

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -441,12 +441,21 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+
         event.stopPropagation();
         event.preventDefault();
-        if ($scope.searchText) clearValue();
 
-        // Force the component to blur if they hit escape
-        doBlur(true);
+        clearValue(!$scope.noClearOnEscape && $scope.searchText);
+
+        if (!$scope.noBlurOnEscape)
```
topherfangio: Please use braces here and with the `else` to follow our [Style Guide](https://google.github.io/styleguide/javascriptguide.xml?showone=Code_formatting#Code_formatting).

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -614,7 +623,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   /**
    * Clears the searchText value and selected item.
    */
-  function clearValue () {
+  function clearValue (clearText) {
```
topherfangio: `clearText` here appears to be a boolean. Can we rename it to `shouldClearText` to be more specific? I originally thought it was a variable of what text to use as a replacement when clearing the old text.

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -149,7 +151,9 @@ function MdAutocomplete () {
       floatingLabel:    '@?mdFloatingLabel',
       autoselect:       '=?mdAutoselect',
       menuClass:        '@?mdMenuClass',
-      inputId:          '@?mdInputId'
+      inputId:          '@?mdInputId',
+      noBlurOnEscape:   '=?mdNoBlurOnEscape',
```
topherfangio: If possible, I like to invert the logic here so that the the option is `md-blur-on-escape` with a default of `true` instead of `md-no-blur-on-escape` with a default of `false`. Same for the `md-no-clear-on-escape` below.

Is it easy/possible to update these here?

clshortfuse: Surely it's possible, but I couldn't find an example of another boolean attribute that defaults to true. 

I could rewrite it as mdEscapeOptions which defaults to "blur clear" which could be clearer to the user, but then it could be confusing in terms of what md-escape-options="" would produce. If it's blank, would it default to "blur clear" or would it not do anything?

I think the current syntax isn't too confusing from end-user perspective, but I can understand how it could look confusing from coding perspective.

clshortfuse:
just tested with string options instead of boolean.

@topherfangio ready for review (again)

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -527,6 +535,38 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+   * Determines if the escape keydown should be processed
+   * @returns {boolean}
+   */
+  function shouldProcessEscape() {
+    if (shouldBlurOnEscape()) return true;                            // Needs blur
+    else if (!ctrl.hidden) return true;                               // Hide dropdown
+    else if (ctrl.loading) return true;                               // Cancel loading
+    else if (shouldClearOnEscape() && $scope.searchText) return true; // Clear Text
+    else return false;
+  }
```
topherfangio: Since all of these end with a `return` statement. You don't actually need all of the `else if` statements; you can just do `if...return`. Same with the final else; it can just be `return false`.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -527,6 +535,38 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+
+  /**
+   * Determines if the escape key should blur focus
+   * @returns {boolean}
+   */
+  function shouldBlurOnEscape() {
+    if (!$scope.escapeOptions) return true;                     // default true
+    if ($scope.escapeOptions.indexOf('blur') >= 0) return true; // blur is requested
```
topherfangio: Does this allow for different capitalizations?

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -149,7 +151,8 @@ function MdAutocomplete () {
       floatingLabel:    '@?mdFloatingLabel',
       autoselect:       '=?mdAutoselect',
       menuClass:        '@?mdMenuClass',
-      inputId:          '@?mdInputId'
+      inputId:          '@?mdInputId',
+      escapeOptions:    '@?mdEscapeOptions'
```
topherfangio: I like this option a lot better 👍

src/components/autocomplete/js/autocompleteController.js
clshortfuse: Oh, I know. I was just using `shouldHide` function as a model in case these it was part of the Style Guide

src/components/autocomplete/js/autocompleteController.js
clshortfuse: I saw using another piece of code for reference but I can easily add `toLowerCase`

clshortfuse:
@topherfangio ready for review

clshortfuse:
i've rebased this. i've been using this code in production for about a month without issue.

ThomasBurleson:

ThomasBurleson:

PR #7826, opened 2016-03-31
fix(menu): resolve an error when going from a nested menu item to a regular one

### LINKED ISSUE #7819: mdMenu: Cannot read property 'open' of null
On mdMenu, if you have a sub menu and another item that doesn't have a submenu in the same menu. If you open the subMenu and then hover over the non-subMenu item it throws the error "Cannot read property 'open' of null" and remains highlighted even after hovering over a different menu item

angular material version: 1.1.0-rc1
browser: chrome 49.0.2623.87

I also tested this on the demo page for the latest version on the angular material website and I'm getting the same error

crisbeto:
Fixes a JS error that was being thrown if the user goes from a menu item with a sub menu to an item that doesn't have one.
Also fixes the menu item without a sub menu staying highlighted afterwards.

Closes #7819.

@topherfangio Could you give me some feedback on this one? I believe it got introduced in https://github.com/angular/material/commit/b58343c20ac4bd3418629c29abddfe3ac3840eb5#diff-7d56e0ee8b5aa2a3434f4f53a8a66168


ThomasBurleson:
lgtm

PR #7831, opened 2016-04-01
fix(list): add target to copiedAttrs

code-tree:
Replacement for  #5877 due to significant changes upstream.


DevVersion:
Looks good to me. Additionally, I'm searching for a more elegant solution for the attribute copy. Manually adding all those attributes is not that elegant.

ThomasBurleson:
use a whitelist ?

DevVersion:
@ThomasBurleson Sorry for the late reply. 

Yes, currently we're using a whitelist, but those attributes in the whitelist, should be also prefixed with the correct attribute prefixes. 

In this case, the PR is fine, I just thought loud, because it's weird to prefix all attributes manually. Tree times.

PR #7843, opened 2016-04-01
update(panel): Adds absolute positioning with center.

### LINKED ISSUE #7794: mdPanel: Add centering for absolute positioning.
I have the perfect screensize, so this has always worked for me!

ErinCoughlan:
Fixes #7794, #7877 

@jelbourn @KarenParker @gmoothart @DerekLouie Please review.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -336,14 +336,14 @@ angular
  * Overlapping the panel with an element:
  * `new MdPanelPosition()
  *     .relativeTo(someElement)
- *     .withPanelXPosition('align-left')
- *     .withPanelYPosition('align-tops');`
+ *     .withPanelXPosition($mdPanel.xPosition.ALIGN_START)
+ *     .withPanelYPosition($mdPanel.yPosition.ALIGN_TOPS);`
  *
  * Aligning the panel with the bottom of an element:
  * `new MdPanelPosition()
  *     .relativeTo(someElement)
- *     .withPanelXPosition('center')
- *     .withPanelYPosition('below');`
+ *     .withPanelXPosition($mdPanel.xPosition.CENTER)
+ *     .withPanelYPosition($mdPanel.yPosition.BELOW);`
  */
 
 /**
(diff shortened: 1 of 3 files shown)


src/components/panel/panel.js
```diff
@@ -335,7 +335,7 @@ angular
  * Overlapping the panel with an element:
  * `new MdPanelPosition()
  *     .relativeTo(someElement)
- *     .withPanelXPosition('align-left')
+ *     .withPanelXPosition('align-start')
```
KarenParker: Consider having these options defined and accessible for the user, similar to the $mdPanel.animation values.

KarenParker:
LGTM

src/components/panel/panel.js
jelbourn: I agree that they should be available "enum-style"

src/components/panel/panel.js
```diff
@@ -1487,58 +1550,95 @@ MdPanelPosition.prototype._calculatePanelPosition = function(panelEl) {
...
-      this._left = targetRight + 'px';
-      break;
-  }
+/**
+ * Calculates the panel position for absolute based on the created
+ * panel element.
+ * @param {number} panelWidth
+ * @param {number} panelHeight
+ */
+MdPanelPosition.prototype._calculateAbsolutePosition =
```
jelbourn: Line limit should be 100, so I think this will fit.

src/components/panel/panel.js
```diff
@@ -1487,58 +1550,95 @@ MdPanelPosition.prototype._calculatePanelPosition = function(panelEl) {
...
+    function(panelWidth, panelHeight) {
+      var windowWidth = window.innerWidth;
```
jelbourn: You can center something completely with CSS:
```css
.globally-centered {
  top: 50%;
  left: 50%;
  transform: translateX(-50%) translateY(-50%);
}
```

jelbourn: Here's what I currently have for the Angular 2 version (subject to change as we alpha):
https://github.com/angular/material2/blob/master/src/core/overlay/position/global-position-strategy.ts

src/components/panel/panel.js
ErinCoughlan: I think this file has been using 80, so I'd rather stay consistent in the file. I'm fine if we want to reflow everything to 100, but not in this PR.

ErinCoughlan:
PTAL. I switched to using the transform css instead of calculating position (which is great because now I don't need to recalculate on page resize!)

src/components/panel/panel.js
```diff
@@ -1051,6 +1094,7 @@ MdPanelRef.prototype._configurePosition = function() {
   this._panelEl.css('bottom', positionConfig.getBottom(this._panelEl));
   this._panelEl.css('left', positionConfig.getLeft(this._panelEl));
   this._panelEl.css('right', positionConfig.getRight(this._panelEl));
+  this._panelEl.css('transform', positionConfig.getTransform());
```
jelbourn: One thing to keep in mind is that an element can only have one `transform` value, so setting it directly will override any transforms used for animations.

jelbourn:
LGTM

KarenParker:
LGTM

PR #7844, opened 2016-04-01
update(panel): Extends attachTo config accepted types.

### LINKED ISSUE #7678: mdPanel: Support string value and DOMElement for attachTo config.
Currently only allows a JQLite element.

Similar version to [this](https://github.com/angular/material/blob/master/src/components/dialog/dialog.js#L751)

ErinCoughlan:
Allows attachTo to accept an Element, JQLite object, or a query selector.

@jelbourn @gmoothart @KarenParker @DerekLouie Please review.

Fixes #7678


KarenParker:
LGTM after you sync.

PR #7848, opened 2016-04-01
fix(autocomplete): don't apply a background if the autocomplete has a floating label

### LINKED ISSUE #7841: Grey background for disabled md-autocomplete
Disabled md-autocomplete instances with the default theme and with md-floating-label have a weird gray background:

http://codepen.io/calebegg/pen/yOzXYw

image

crisbeto:
The grey background should only be applied if the autocomplete doesn't have a floating label.

Closes #7841.

### DIFF
--- a/src/components/autocomplete/autocomplete-theme.scss
+++ b/src/components/autocomplete/autocomplete-theme.scss
@@ -1,6 +1,6 @@
 md-autocomplete.md-THEME_NAME-theme {
   background: '{{background-A100}}';
-  &[disabled] {
+  &[disabled]:not([md-floating-label]) {
     background: '{{background-100}}';
   }
   button {


ThomasBurleson:
@crisbeto - attribute selectors cause perf issues in IE; at least the Layout selectors did. 
Can you confirm (a) this does **not** degrade IE perf or (b) provide a classname approach to solve this instead?
> I suspect (a)... but would you appreciate your investigation.

crisbeto:
@ThomasBurleson There isn't any noticeable difference, but these things tend to stack up over time. I went with the attribute selector, because that's what [was being used to style the floating label mode](https://github.com/angular/material/blob/master/src/components/autocomplete/autocomplete.scss#L46) in the first place. I can switch it over to a class selector but that might be considered a breaking change?

PR #7849, opened 2016-04-01
chore(): bower-material should update style modules and remove stale files

### LINKED ISSUE #7835: remove attribute layout css from modules/js/core/core.css
Please remove attribute layout css from "modules/js/core/core.css" or at least replace them with class layout css "angular-material.layout-attributes.css".

Also please update "modules/css/angular-material-layouts.css" to latest version. Build process does not update files under modules/css.

DevVersion:
* `angular-material.scss` copied into `modules/scss`
* `core/layout/ie_fixes.css` copied into `modules/css`
* `layouts/*.{css,scss}` copied into `modules/{css,scss}`
*  Removed stale `./demos/` folder.
* Core Modules (default + closure) should not contain layout attributes. (only classes)

Fixes #7835

### DIFF
--- a/gulp/util.js
+++ b/gulp/util.js
@@ -85,19 +85,33 @@ function buildModule(module, opts) {
   var name = module.split('.').pop();
   utils.copyDemoAssets(name, 'src/components/', 'dist/demos/');
 
-  var stream = utils.filesForModule(module)
-      .pipe(filterNonCodeFiles())
-      .pipe(gulpif('*.scss', buildModuleStyles(name)))
-      .pipe(gulpif('*.js', buildModuleJs(name)));
+  var stream = utils.filesForModule(module);
+
   if (module === 'material.core') {
+    stream = stream.pipe(filter(function(x) {
+      // We are not injecting the layout-attributes selectors into the core module css,
+      // otherwise we would have the layout-classes and layout-attributes in there.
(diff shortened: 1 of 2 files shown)


gulp/util.js
```diff
@@ -85,13 +85,20 @@ function buildModule(module, opts) {
...
-      .pipe(filterNonCodeFiles())
-      .pipe(gulpif('*.scss', buildModuleStyles(name)))
-      .pipe(gulpif('*.js', buildModuleJs(name)));
+  var stream = utils.filesForModule(module);
+
   if (module === 'material.core') {
+    stream = stream.pipe(filter(function(x) {
+      // We are not injecting the layout-attributes selectors into the core module css,
+      // otherwise we would have the layout-classes and layout-attributes in there.
+      return !/.*layout-attributes\.scss/g.test(x.path);
```
ThomasBurleson: So where is the `layout-attributes.scss` file include with the ^^ changes?

DevVersion: The actual issue was, that the layout attributes were also added to the `modules/core/core.css`, in addition to the class layout attributes. That means that both layout opportunities were added into the core stylesheet.

Filtering in `filesForModule` won't be that elegant, because we only want to filter it for the `material.core` module. Adding this check to `filesForModule` would be really unclear.

Issue #3793, opened 2015-07-17
md-checkbox doesn't center properly when md-label is blank


clshortfuse:
Because md-checkbox has padding to fix alignment with the its md-label child, it will this will add an additional space. md-label has a margin-left:18px value which means it will, at minimum have 18px width. This will forcelly "pad" the parent md-checkbox element by 18px regardless of content inside.  

<edit> removed my fix since it didn't work </edit>
 
I'm creating test code right now. This may need to be done to radiobutton as well (I haven't checked, but will with my test code).

Test: http://codepen.io/anon/pen/oXMpXP

clshortfuse:
I wrote a fix and it works, but feels really hacky:

http://codepen.io/anon/pen/rVrpmp

.fixed md-checkbox {
  padding-left: 8px;  
}

.fixed md-checkbox .md-container {
  margin-left: 5px
}

.fixed md-checkbox .md-label {
  margin-left: 20px
}

PR #7851, opened 2016-04-01
fix(md-checkbox) support horizontal centering of blank md-checkbox

clshortfuse:
A blank md-checkbox will improperly be padded to 30px because md-label's margin-left attribute

With this CSS change, If md-label is empty, then it won't appear and thus, won't pad the checkbox to the right. This works together with md-checkbox:last-of-type so users might have to wrap md-checkbox in a div if it's the only element.

Here's a code pen showing how the change works:

http://codepen.io/shortfuse/pen/MyEvVm

This fixes https://github.com/angular/material/issues/3793

This follows the material design guidelines as shown here under "Column Padding"

https://www.google.com/design/spec/components/data-tables.html#data-tables-specs

Right now there is still an issue with the 16px vertical margin, but that could be reserved for another pull request/issue.

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -66,5 +66,9 @@ md-checkbox {
     @include rtl(margin-left, $checkbox-text-margin + $checkbox-width, 0);
     @include rtl(margin-right, 0, $checkbox-text-margin + $checkbox-width);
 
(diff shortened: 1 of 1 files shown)


clshortfuse:
closing for deprecation and this fix is included in

https://github.com/angular/material/pull/8209

PR #7852, opened 2016-04-01
fix(release): specifies upstream URL when pushing to material repo

robertmesserle:
@ThomasBurleson @jelbourn This should fix the issue where tags and/or CHANGELOG.md do not get pushed during release.


ThomasBurleson:
@robertmesserle  - awesome!

PR #7862, opened 2016-04-03
feat(progress): add the ability to disable the progressbars

crisbeto:
* Adds support for `ngDisabled` to the `progressCircular` and `progressLinear` components. Hides the element and stops all rendering if the element is disabled.
* Fixes a potential weird appearance in `progressCircular` if the animation time goes below 0.
* Adds some missing coverage for the `$mdProgressCircularProvider`.

@ThomasBurleson this is as discussed on Friday.

### DIFF
--- a/src/components/progressCircular/js/progressCircularDirective.js
+++ b/src/components/progressCircular/js/progressCircularDirective.js
@@ -29,6 +29,8 @@
  * should be a pixel-size value (eg '100'). If this attribute is
  * not present then a default value of '50px' is assumed.
  *
+ * @param {boolean=} ng-disabled Determines whether to disable the progress element.
+ *
  * @usage
  * <hljs lang="html">
  * <md-progress-circular md-mode="determinate" value="..."></md-progress-circular>
@@ -52,6 +54,7 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdThe
   var DEGREE_IN_RADIANS = $window.Math.PI / 180;
   var MODE_DETERMINATE = 'determinate';
   var MODE_INDETERMINATE = 'indeterminate';
+  var DISABLED_CLASS = '_md-progress-circular-disabled';
 
   return {
     restrict: 'E',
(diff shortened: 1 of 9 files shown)


src/components/progressCircular/progress-circular.scss
```diff
@@ -2,6 +2,10 @@
 md-progress-circular {
     position: relative;
 
+    &._md-progress-circular-disabled {
```
ThomasBurleson: :+1:

crisbeto:
Ops, CI didn't pass, I'll take a better look.

ThomasBurleson:
lgtm

ThomasBurleson:
@crisbeto - can you add this to `md-progress-linear` also ?

crisbeto:
Sure!

src/components/progressCircular/js/progressCircularDirective.js
```diff
@@ -103,22 +108,34 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdThe
...
+      return node.disabled;
+    }], function(newValues, oldValues) {
       var mode = newValues[1];
+      var isDisabled = newValues[2];
+      var wasDisabled = oldValues[2];
 
-      if (mode !== MODE_DETERMINATE && mode !== MODE_INDETERMINATE) {
-        mode = MODE_INDETERMINATE;
-        attrs.$set('mdMode', mode);
+      if (isDisabled !== wasDisabled) {
```
DevVersion: `wasDisabled` will be always different from the new `isDisabled`. So this check seems to be unnecessary?

crisbeto: This watcher also fires when the value and mode change. In those cases the disabled might not have changed.

DevVersion: @crisbeto Indeed, didn't saw the `watchGroup`. It's a bit weird, using a watchGroup here but I'm fine :)

crisbeto:
@ThomasBurleson Added the same to the linear progress and fixed the CI issues.

src/components/progressCircular/progress-circular.scss
ThomasBurleson: Why `visibility: hidden` instead of `display:none`?

crisbeto: Exactly.

PR #7864, opened 2016-04-03
feat(mdThemeProvider): register custom theme styles

### LINKED ISSUE #7708: allow registering custom style sheet (mdss ?) into `$MD_THEME_CSS`
Would be very handy to be able to register your own component and hook into dynamic theming... This may require changing setting `$MD_THEME_CSS` as a constant to a provider or decorator.

Since these style sheets have angular material specific variables, perhaps store the data in `mdss` files.

```
var fs = require('fs');

// require mdss files as a string
require.extensions['.mdss'] = function(module, filename) {
    module.exports = fs.readFileSync(filename, 'utf8');
};

angular.module('myApp').config(function($mdThemingProvider) {
  $mdThemingProvider.registerStyles(require('./my-component-theme.mdss'));
})
```

my-component-theme.mdss
```
tr[md-row].md-THEME_NAME-theme {
  ...
  background: {{ background-200 }}
}
```

I'm not sure what the logic is for the `md-theme` directive to add the class but that may need to be changed as well.

Emeegeemee:
Closes #7708

### DIFF
--- a/src/core/services/theming/theming.js
+++ b/src/core/services/theming/theming.js
@@ -141,11 +141,12 @@ var generateOnDemand = false;
 var nonce = null;
 
 // Custom styles registered to be used in the theming of custom components.
-var registeredStyles = [];
+var registeredStyles;
 
 function ThemingProvider($mdColorPalette) {
   PALETTES = { };
   THEMES = { };
+  registeredStyles = [];
 
   var themingProvider;
   var defaultTheme = 'default';
--- a/src/core/services/theming/theming.js
+++ b/src/core/services/theming/theming.js
@@ -14,6 +14,11 @@ angular.module('material.core.theming', ['material.core.theming.palette'])
 
 /**
  * @ngdoc method
+ * @name $mdThemingProvider#registerStyles
+ * @param {string} styles The styles to be appended after angular material's built in theme css.
+ */
+/**
+ * @ngdoc method
  * @name $mdThemingProvider#setNonce
  * @param {string} nonceValue The nonce to be added as an attribute to the theme style tags.
  * Setting a value allows the use CSP policy without using the unsafe-inline directive.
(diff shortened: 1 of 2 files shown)


Emeegeemee:
I signed it!

src/core/services/theming/theming.js
```diff
@@ -153,6 +161,9 @@ function ThemingProvider($mdColorPalette) {
     extendPalette: extendPalette,
     theme: registerTheme,
 
+    registerStyles: function(styles) {
```
DevVersion: I initially thought, that there will be a function, which compiles the provided style on demand as well? Thoughts?

Emeegeemee: Could you clarify what you mean by compile? I would expect the provided styles are valid CSS just like $MD_THEME_CSS and will just be appended to the constant before the themes are generated.

DevVersion: @Emeegeemee Sorry, I got some different thoughts, which are not valid anymore, since `$mdColors` will come. 

Just thought again about your method, and it looks good.

ThomasBurleson:
*  Please provide a CodePen or Doc Theme Demo that justifies this feature addition. 
*  Also your test does not test anything. Please provide better tests.

> I am preparing to reject this PR as "too risky".

Emeegeemee:
I've updated the tests to be more specific and added a few missing checks, please let me know if you want something more specific tested.

As for a demo I'm not really sure where to start, from my perspective all of the currently implemented components serve as great demo's to the power of using the theming system. Would a document explaining how to create and register custom styles be better? I'm thinking of updating and moving [this guide](https://github.com/angular/material/blob/master/docs/guides/THEMES_IMPL_NOTES.md) into the theme section of the documentation if that make sense. 

As for your "too risky" comment could you give some examples that I could help reduce or mitigate. I definitely see the potential for developers making a mistake with registering styles. Doing so would cause the rest of the theming to break, but I believe that you would just get components without theming similar to if the themes were disabled.

ThomasBurleson:
@DevVersion - can you rebase this please and force push to this PR?

ThomasBurleson:
@DevVersion - ping!

DevVersion:
@Emeegeemee Thanks for the great PR again - We have a new PR, which includes your changes and gives deservedly credit to you :)

PR #7870, opened 2016-04-04
fix(select): properly disable the progressbar

crisbeto:
The progressbar inside the select with async results was using the old way of disabling progressbars that doesn't disable the element properly.

Related to #7862.

### DIFF
--- a/src/components/progressCircular/js/progressCircularDirective.js
+++ b/src/components/progressCircular/js/progressCircularDirective.js
@@ -108,6 +108,8 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdThe
       startIndeterminateAnimation();
     }
 
+    scope.$on('$destroy', cleanupIndeterminateAnimation);
+
     scope.$watchGroup(['value', 'mdMode', function() {
       var isDisabled = node.disabled;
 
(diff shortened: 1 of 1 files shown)


robertmesserle:
Can you add the test that you used to confirm this fix?

ThomasBurleson:
@robertmesserle - This was closed. Accident ?

robertmesserle:
@ThomasBurleson I was in our team meeting, so I didn't get a chance to look over the PR.  The fix appears to work, but so far, nobody can quite understand WHY it works - without knowing the internals of flush and the cancel method, it looks like it should not work.

I wanted to hold off on merging this until we can understand what is happening and whether or not it is a stable fix, or a fluke based on internal implementation details.

robertmesserle:
To give a bit more detail, the code as it is written is attempting to cancel the callback that is already running.  This should be a noop, but somehow resolves the infinite loop issue.

At the very least, we should add a comment noting that this is a temporary fix until we come up with a proper fix - @matsko is also working on adding a call limit to `$animate.flush` to prevent infinite loops such as this.

crisbeto:
I closed this one because it was supposed to be only for that select fix that became a part of the other PR, but then I also pushed some code here so @robertmesserle could try it against his tests.

robertmesserle:
Yea, the `$destroy` part is fine, but I wanted more time to review the `rAF` related fix.

crisbeto:
I believe I got a way of avoiding doing that rotation in JS altogether. I initially resorted to it because IE made it wobble.

PR #7872, opened 2016-04-04
feat(panel): Add hasBackdrop config to the panel.

### LINKED ISSUE #7682: mdPanel: Add hasBackdrop config option to the panel.
Whether there should be an opaque backdrop behind the panel. Defaults to false.
 
Consider following the mdDialog implementation [here](https://github.com/angular/material/blob/master/src/components/dialog/dialog.js#L867).

-----------
This task involves:

- Updating public documentation
- Adding hasBackdrop to the panel config.

KarenParker:
Fixes #7682.

@gmoothart @ErinCoughlan @DerekLouie @jelbourn Please review.

### DIFF
--- a/src/components/backdrop/backdrop.scss
+++ b/src/components/backdrop/backdrop.scss
@@ -1,5 +1,5 @@
 // !!Important - Theme-based Background-color can be configured in backdrop-theme.scss
-//             - Animate background-color opacityy only for `.md-opaque` styles
+//             - Animate background-color opacity only for `.md-opaque` styles
 
 md-backdrop {
   transition: opacity 450ms;
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -4,7 +4,8 @@
  */
 angular
     .module('material.components.panel', [
-      'material.core'
+      'material.core',
+      'material.components.backdrop'
     ])
     .service('$mdPanel', MdPanelService);
 
@@ -126,14 +127,14 @@ angular
  *   - `animation` - `{MdPanelAnimation=}`: An MdPanelAnimation object that
  *     specifies the animation of the panel. For more information, see
  *     `MdPanelAnimation`.
+ *   - `hasBackdrop` - `{boolean=}`: Whether there should be an opaque backdrop
+ *     behind the panel. Defaults to false.
  *
  * TODO(ErinCoughlan): Add the following config options.
  *   - `groupName` - `{string=}`: Name of panel groups. This group name is
  *     used for configuring the number of open panels and identifying specific
  *     behaviors for groups. For instance, all tooltips will be identified
  *     using the same groupName.
- *   - `hasBackdrop` - `{boolean=}`: Whether there should be an opaque backdrop
- *     behind the panel. Defaults to false.
  *   - `disableParentScroll` - `{boolean=}`: Whether the user can scroll the
  *     page behind the panel. Defaults to false.
  *
@@ -610,6 +611,7 @@ function MdPanelService($rootElement, $rootScope, $injector) {
     escapeToClose: false,
     focusOnOpen: true,
     fullscreen: false,
+    hasBackdrop: false,
     transformTemplate: angular.bind(this, this.wrapTemplate_),
     trapFocus: false,
     zIndex: defaultZIndex
@@ -761,6 +763,11 @@ function MdPanelRef(config, $injector) {
   /** @private @const {!angular.$animate} */
   this._$animate = $injector.get('$animate');
 
+  /** @private @const {!MdPanelRef} */
+  this._$mdPanel = $injector.get('$mdPanel');
+
+  /** @private @const {!angular.$log} */
+  this._$log = $injector.get('$log');
 
   // Public variables.
   /**
@@ -811,6 +818,9 @@ function MdPanelRef(config, $injector) {
 
   /** @private {!angular.JQLite|undefined} */
   this._bottomFocusTrap;
+
+  /** @private {!$mdPanel|undefined} */
+  this._backdropRef;
 }
 
 
@@ -831,13 +841,10 @@ MdPanelRef.prototype.open = function() {
   // TODO(ErinCoughlan) - Cancel any in-progress actions.
 
   var self = this;
-  this._openPromise = this._$q(function(resolve, reject) {
-    self.attachOnly().then(function() {
-      self.show().then(function() {
-        resolve(self);
-      }, reject);
-    }, reject);
-  });
+  this._openPromise = this.attachOnly()
+      .then(function() {
+        return self.show();
+      }, this._$q.reject);
 
   return this._openPromise;
 };
@@ -853,15 +860,17 @@ MdPanelRef.prototype.close = function() {
   // TODO(ErinCoughlan) - Cancel any in-progress actions.
 
   var self = this;
-  return this._$q(function(resolve, reject) {
-    self.hide().then(function () {
-      self.detach().then(function() {
+  var closePromise = this.hide()
+      .then(function() {
+        return self.detach();
+      }, this._$q.reject)
+      .then(function() {
         // TODO(ErinCoughlan) - Add destroy. This will make the code here
         // different than just calling this.detach().
-        resolve(self);
-      }, reject);
-    }, reject);
-  });
+        return self._$q.resolve(self);
+      }, this._$q.reject);
+
+  return closePromise;
 };
 
 
@@ -879,13 +888,15 @@ MdPanelRef.prototype.attachOnly = function() {
   // TODO(ErinCoughlan) - Cancel any in-progress actions.
 
   var self = this;
-  this._attachPromise = this._$q(function(resolve, reject) {
-    self._createPanel().then(function() {
+
+  this._attachPromise = this._$q.all([
+    this._createBackdrop(),
+    this._createPanel().then(function() {
       self.isAttached = true;
       self._addEventListeners();
-      resolve(self);
-    }, reject);
-  });
+      return self._$q.resolve(self);
+    }, this._$q.reject)
+  ]);
 
   return this._attachPromise;
 };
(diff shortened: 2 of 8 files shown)


KarenParker:
Any comments?

src/components/panel/panel.js
```diff
@@ -988,6 +1007,48 @@ MdPanelRef.prototype.focusOnOpen = function() {
 
 
 /**
+ * Shows the backdrop.
+ */
+MdPanelRef.prototype._showBackdrop = function() {
+  if (this._config.hasBackdrop) {
+    this._backdrop = this._$mdUtil.createBackdrop(this._$rootScope,
```
jelbourn: Why isn't the backdrop just another pane instance? If that were the case, we could remove `$mdUtil.createBackdrop`, since that logic doesn't really have any business being there.

src/components/panel/panel.js
```diff
@@ -988,6 +1007,48 @@ MdPanelRef.prototype.focusOnOpen = function() {
...
+    destroy ? this._backdrop.remove() : this._$animate.leave(this._backdrop);
+  }
+};
+
+
+/**
+ * Ensure the dialog container fill-stretches to the viewport
+ */
+MdPanelRef.prototype._stretchPanelContainerToViewport = function() {
+  var isFixed = this._$window.getComputedStyle(this._$document[0].body)
```
jelbourn: This function could definitely use some comments to explain what/why it's doing.

src/components/backdrop/backdrop.scss
```diff
@@ -23,6 +23,9 @@ md-backdrop {
   &._md-dialog-backdrop {
     z-index: $z-index-dialog - 1;
   }
+  &._md-panel-backdrop {
+    z-index: $z-index-dialog - 1;
```
ErinCoughlan: We can't assume this is the z-index. (A tooltip within a dialog is an example.) Yes, the default is $z-index-dialog, but we allow the user to customize that to any value they want. You could set this in the panel instead where you have access to the panel z-index.

src/components/panel/panel.js
```diff
@@ -988,6 +1007,48 @@ MdPanelRef.prototype.focusOnOpen = function() {
 
 
 /**
+ * Shows the backdrop.
```
ErinCoughlan: `@private`? here and below

src/components/panel/panel.spec.js
```diff
@@ -441,7 +442,21 @@ describe('$mdPanel', function() {
 
       openPanel(config);
       expect(PANEL_EL).toHaveClass(FULLSCREEN_CLASS);
+    });
+
+    it('should not show backdrop on open', function() {
```
ErinCoughlan: perhaps, `it('should default hasBackdrop to false', `

since you're testing the default behavior, not setting `hasBackdrop` to false.

src/components/panel/panel.js

jelbourn:
LGTM, be sure to squash the commits

KarenParker:
Tests are all working and added a few to cover the new promise structure.

src/components/panel/panel.spec.js
```diff
@@ -189,6 +190,40 @@ describe('$mdPanel', function() {
       expect(panelRef.isAttached).toEqual(false);
     });
 
+    fit('should resolve on animate failure when opening', function () {
```
DerekLouie: oops I think you left a `fit` in here.

src/components/panel/panel.spec.js
```diff
@@ -211,6 +246,29 @@ describe('$mdPanel', function() {
       expect(panelRef.isAttached).toEqual(true);
     });
 
+    fit('should resolve on animate failure when closing', function () {
```
DerekLouie: also `fit` here

src/components/panel/panel.js
```diff
@@ -945,13 +960,24 @@ MdPanelRef.prototype.show = function() {
...
-      resolve(self);
-    }, reject);
-  });
+    return self._animateOpen();
+  }
+
+  this._showPromise = this._$q.all([
+      this._backdropRef ? this._backdropRef.show() : this._$q.resolve(self),
+      animatePromise().then(function() {
+        return self._$q.resolve(self);
```
ErinCoughlan: Does this need to be wrapped in a function? Can you just have the resolve part?

src/components/panel/panel.js
```diff
@@ -945,13 +960,24 @@ MdPanelRef.prototype.show = function() {
...
+      }, function() {
+        console.warn('MdPanel Animations failed. Showing panel without animating.');
```
ErinCoughlan: I think we should use $log for consistency within Material:
https://docs.angularjs.org/api/ng/service/$log

src/components/panel/panel.js
```diff
@@ -1131,6 +1144,51 @@ MdPanelRef.prototype._configurePosition = function() {
...
 /**
+ * Focuses on the panel or the first focus target.
+ * @private
+ */
+MdPanelRef.prototype._focusOnOpen = function() {
+  if (this._config['focusOnOpen']) {
+    // Wait a digest to guarantee md-autofocus has finished adding the class
+    // _md-autofocus, otherwise the focusable element isn't available to focus.
+    var self = this;
+    this._$rootScope.$applyAsync(function() {
```
ErinCoughlan: Maybe add as a TODO, but use $$postDigest, which is when angular is done rendering everything and is stable, but doesn't have to wait a full digest where other things can happen.

src/components/panel/panel.scss
```diff
@@ -28,41 +28,43 @@
 .md-panel {
   opacity: 0;
   overflow: auto;
+  position: fixed;
```
ErinCoughlan: Did you remove position: fixed from addStyles?

src/components/panel/panel.spec.js
ErinCoughlan: fit

src/components/panel/panel.js
KarenParker: Not if you want to pass self.

KarenParker: If I just used "self._$q.resolve" I wouldn't be able to specify the self param. I don't know if that strictly matters though or if the "then" passes it anyways.

src/components/panel/panel.js
```diff
@@ -808,6 +813,9 @@ function MdPanelRef(config, $injector) {
 
   /** @private {!angular.JQLite|undefined} */
   this._bottomFocusTrap;
+
+  /** @private {!angular.JQLite|undefined} */
```
gmoothart: the `_backdropRef` is an mdPanel

src/components/panel/panel.scss
KarenParker: Yes

src/components/panel/panel.js
```diff
@@ -1293,7 +1351,7 @@ MdPanelRef.prototype._animateClose = function() {
   return this._$q(function(resolve, reject) {
     animationConfig.animateClose(self._$q).then(function(){
       self.removeClass('md-panel-is-showing');
-      resolve(self);
+      return resolve(self);
```
gmoothart: when using this style I think you can omit the `return`; it's only when chaining promises that you have to explicitly return.

src/components/panel/panel.scss
```diff
@@ -28,41 +28,43 @@
...
-      transition: $swift-ease-out;
-    }
-
-    &._md-panel-animate-fade-in {
-      opacity: 1;
-      transition: all 0.4s linear;
-    }
+  &._md-panel-animate-enter {
+    opacity: 1;
+    transition: all 0.5s cubic-bezier(0.0, 0.0, 0.2, 1);
```
gmoothart: Can you add variables for these transitions? It looks like you use 2 different ones in this file.

gmoothart:
LGTM

src/components/panel/panel.scss
```diff
@@ -28,41 +28,43 @@
...
+  &._md-panel-animate-scale-out,
+  &._md-panel-animate-fade-out {
     opacity: 0;
-    transition: $swift-ease-out;
   }
 
-  &._md-panel-animate-fade-out {
+  &.md-opaque-enter {
+    height: 100%;
+    opacity: .48;
```
ErinCoughlan: Perhaps make a constant for this opacity. It seems random, and likely to change.

src/components/panel/panel.spec.js
```diff
@@ -189,6 +190,40 @@ describe('$mdPanel', function() {
...
     });
 
+    fit('should resolve on animate failure when opening', function () {
+      var openRejected = false;
+
+      panelRef._animateOpen = function() {
+        return panelRef._$q.reject();
+      };
+
+      panelRef.open().then(function() {}, function () {
```
ErinCoughlan: Can you make sure that the openResolve is called? Implicitly this is what you are testing right now, but having both makes it more explicit.

src/components/panel/panel.spec.js
```diff
@@ -499,8 +557,51 @@ describe('$mdPanel', function() {
...
     });
+
+    it('should show backdrop when hasBackdrop=true', function() {
+      var config = { template: DEFAULT_TEMPLATE, hasBackdrop: true };
+
+      openPanel(config);
+      expect(BACKDROP_CLASS).toExist();
+
+      closePanel();
+      expect(BACKDROP_CLASS).not.toExist();
```
ErinCoughlan: Should you add a real class to the backdrop, which would also help debugging in the future? Perhaps, `md-panel-backdrop`? You're checking for `'.md-opaque-enter'`, which is an animation class.

src/components/panel/panel.spec.js
```diff
@@ -788,7 +889,6 @@ describe('$mdPanel', function() {
         openPanel(config);
 
         var panelCss = document.querySelector(PANEL_EL).style;
-        expect(panelCss.position).toEqual('fixed');
```
ErinCoughlan: I think this needs to be in at least 1 test, since it actually does verify that positioning is correct.

src/components/panel/panel.js
KarenParker: Does it matter if I am chaining this function above in different spots?

src/components/panel/panel.scss
KarenParker: Only used here and this is the value used by all backdrops so not likely to change.

src/components/panel/panel.spec.js
KarenParker: Can't access the value. It is now set in a css class and the style sheets aren't loaded in tests.

KarenParker:
Comments addressed.

src/components/panel/panel.js
```diff
@@ -811,6 +818,9 @@ function MdPanelRef(config, $injector) {
 
   /** @private {!angular.JQLite|undefined} */
   this._bottomFocusTrap;
+
+  /** @private {!$mdPanel|undefined} */
```
ErinCoughlan: This is MdPanelRef, not $mdPanel.

src/components/panel/panel.scss
```diff
@@ -28,41 +28,43 @@
...
-      opacity: 1;
-      transition: all 0.4s linear;
-    }
   }
 
-  &._md-panel-animate-slide-out {
+  &._md-panel-animate-enter {
     opacity: 1;
-    transition: $swift-ease-out, opacity 1ms;
+    transition: all 0.5s cubic-bezier(0.0, 0.0, 0.2, 1);
```
ErinCoughlan: Should there be variables for these two transitions (enter and leave)?

src/components/panel/panel.js
```diff
@@ -1137,6 +1152,53 @@ MdPanelRef.prototype._configurePosition = function() {
...
+      .openFrom(this._config.attachTo)
+      .withAnimation({
+        open: '_md-opaque-enter',
+        close: '_md-opaque-leave'
+      });
+    var backdropConfig = {
+      animation: backdropAnimation,
+      attachTo: this._config.attachTo,
+      focusOnOpen: false,
+      panelClass: 'md-panel-backdrop',
```
ErinCoughlan: I think this should be a private class. We don't want people modifying the backdrop.

ErinCoughlan:
LGTM, aside from small comments.

Make sure to rebase in addition to the squash before submitting.

DerekLouie:
LGTM

PR #7873, opened 2016-04-04
perf(progress-circular): proper cleanup when the element gets destroyed or the animations are flushed

crisbeto:
* Fixes the `$interval` that does the indeterminate animation not being cancelled when the element is destroyed.
* Fixes the `$$rAF` that does the indeterminate rotation being kicked into an infinite loop when calling `$animate.flush`.
* Fixes the select progressbar continuing to render while it's hidden.

@ThomasBurleson this fixes the unit test issues that @robertmesserle and @ErinCoughlan were having.

### DIFF
--- a/src/components/progressCircular/js/progressCircularDirective.js
+++ b/src/components/progressCircular/js/progressCircularDirective.js
@@ -79,7 +79,7 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdThe
         var mode = hasValue ? MODE_DETERMINATE : MODE_INDETERMINATE;
         var info = "Auto-adding the missing md-mode='{0}' to the ProgressCircular element";
 
-          // $log.debug( $mdUtil.supplant(info, [mode]) );
+        $log.debug( $mdUtil.supplant(info, [mode]) );
         attrs.$set('mdMode', mode);
       } else {
         attrs.$set('mdMode', attrs.mdMode.trim());
@@ -97,6 +97,8 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdThe
     var endIndeterminate = $mdProgressCircular.endIndeterminate;
     var rotationIndeterminate = 0;
     var lastAnimationId = 0;
+    var lastDrawFrame;
+    var lastRotationFrame;
     var interval;
 
     $mdTheming(element);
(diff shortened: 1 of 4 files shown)


ErinCoughlan:
Do you want a test that tries `$animate.flush` somewhere and makes sure it's not an infinite loop?

I know you tested this with our code and it works, but I don't want this to break with future releases.

crisbeto:
Sure, I wasn't quite sure how to test for an infinite loop.

ErinCoughlan:
What about rendering an `md-progress-circular` with `md-mode="indeterminate"` in the DOM with an ng-if for some variable that you can change with a `$scope.$apply`. After the apply, you'd have to call `$animate.flush()`, to make sure the element is removed from the DOM. If the test never finishes, then you've reproduced the problem locally.

crisbeto:
Well me and Rob managed to reproduce it like this:

```
it('test', inject(function($animate) {
    var progress = buildIndicator('<md-progress-circular md-mode=" indeterminate"></md-progress-circular>');
    $animate.flush();
  }));
```

I was talking more about being able to check with an `expect` whether it went into an infinite loop, but I suppose that's impossible.

ErinCoughlan:
Ahh, I understand now. Yeah, you can't check if its an infinite loop, but if the tests don't finish running, they'll fail, which is all I'm looking for here.

crisbeto:
Done.

src/components/progressCircular/js/progressCircularDirective.js
```diff
@@ -79,7 +79,7 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdThe
         var mode = hasValue ? MODE_DETERMINATE : MODE_INDETERMINATE;
         var info = "Auto-adding the missing md-mode='{0}' to the ProgressCircular element";
 
-          // $log.debug( $mdUtil.supplant(info, [mode]) );
+        $log.debug( $mdUtil.supplant(info, [mode]) );
```
ErinCoughlan: I remember there being a problem with $log not being present sometimes. @jelbourn Do you remember what this was?

crisbeto: I accidentally introduced this one in a PR yesterday together with that bump in the animation time for the determinate animation.

ErinCoughlan:
LGTM. Thanks for investigating.

PR #7893, opened 2016-04-05
fix(select): md-checkbox inside md-list-item using incorrect text color.

### LINKED ISSUE #7885: md-checkbox inside md-list-item have wrong colour text in 1.1.0-rc2
You can see this bug in the list demos.

Under 1.1.0-rc2 labels for checkboxes in lists are a lighter shade of grey: http://codepen.io/anon/pen/reYGVp

Which doesn't happen in the demos for 1.0.7: http://codepen.io/anon/pen/GZOMpK

DerekLouie:
Hello,

@ThomasBurleson @DevVersion @EladBezalel.

I used an attribute selector which is not picked up by theming service which made some text miss-colored. Here is the fix + screenshot.

Fixes #7885.

Please Review.

fixedselectcheckboxcss

### DIFF
--- a/src/components/select/select-theme.scss
+++ b/src/components/select/select-theme.scss
@@ -81,7 +81,7 @@ md-select-menu.md-THEME_NAME-theme {
   }
 }
 
-[md-checkbox-enabled].md-THEME_NAME-theme {
+._md-checkbox-enabled.md-THEME_NAME-theme {
   @include checkbox-primary('[selected]');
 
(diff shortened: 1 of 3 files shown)


DevVersion:
LGTM

EladBezalel:
LGTM @DevVersion please test manually

DevVersion:
@EladBezalel Yep, this works as expected and fix the ripple text color side-effect.

PR #7900, opened 2016-04-05
fix(sidenav): $mdSideNav component lookups should null if not found

ThomasBurleson:
The current `$mdSidenav` API does not support any way to determine if the component exists (has been instantiated) since the response will give a "fake" instance reference. 

> Typical usages of `<md-sidenav ng-if="$ctrl.allowSideNav" md-component-id="left"> ... </md-sidenav>` can easily create a situation where the sideNav `left` cannot be found / (is not registered).

If the component cannot be found (by its `md-component-id` reference), then logically the lookup *should fail* and return `null`; and the following code usages would be supported:

Previous API does not support any way to determine if the component exists (has been instantiated) since the response will give a "fake" instance reference. The current API supports `$mdSideNav().find(id)`, `$mdSideNav().waitFor(id)`, `$mdSideNav(id)` usages.

```js
// Same functionality as previous API:
// If the component has already been instantiated
// otherwise a NPE will occur

$scope.isOpenRight = function(){
      return $mdSidenav('right').isOpen();
};

var right;

// If the component will be created later then
// use the new `.waitFor(<id>)` api

$scope.isOpenRight = function(){
  right = $mdSidenav('right');  // sync lookup

  if ( !right) {
    $mdSidenav()
      .waitFor('right')         // async lookup
      .then(function(instance){
        right = instance;
      });
  }
  return right ? right.isOpen() : false;
};
```
#breaking

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -57,58 +57,52 @@ angular
  * $mdSidenav(componentId).isLockedOpen();
  * </hljs>
  */
-function SidenavService($mdComponentRegistry, $q) {
-  return function(handle) {
+function SidenavService($mdComponentRegistry, $mdUtil, $q, $log) {
+  var errorMsg = "SideNav '{0}' is not available! Did you use md-component-id='{0}'?";
+  var service = {
+        find    : findInstance,     //  sync  - returns proxy API
+        waitFor : waitForInstance   //  async - returns promise
+      };
 
-    // Lookup the controller instance for the specified sidNav instance
-    var self;
-    var errorMsg = "SideNav '" + handle + "' is not available!";
-    var instance = $mdComponentRegistry.get(handle);
+  /**
+   * Service API that supports three (3) usages:
+   *   $mdSidenav().find("left")                       // sync (must already exist) or returns null
+   *   $mdSidenav("left").toggle();                    // sync (must already exist) or returns null; deprecated
+   *   $mdSidenav("left",true).then( function(left){   // async returns instance when available
+   *    left.toggle();
+   *   });
+   */
+  return function(handle, enableWait) {
+    if ( angular.isUndefined(handle) ) return service;
 
-    if(!instance) {
-      $mdComponentRegistry.notFoundError(handle);
-    }
-
-    return self = {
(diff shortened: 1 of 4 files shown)


ThomasBurleson:
@ErinCoughlan, @topherfangio, @EladBezalel  - please review.

ThomasBurleson:
Question: should `null` or `undefined` be returned. 

> I suggest that **`undefined`** may be a superior response.

ThomasBurleson:
@ErinCoughlan, @topherfangio, @EladBezalel  - please review.

src/components/sidenav/sidenav.js
```diff
@@ -57,58 +57,52 @@ angular
...
+     * registered with the markup `md-component-id`
+     */
+    function findInstance(handle) {
+      var instance = $mdComponentRegistry.get(handle);
+      if(!instance) {
+        // Report missing instance
+        $log.error( $mdUtil.supplant(errorMsg, [handle || ""]) );
+
+        // The component has not registered itself... most like NOT yet created
+        // return null to indicate that the Sidenav is not in the DOM
```
ErinCoughlan: Update comment `null` -> `undefined`

src/components/sidenav/sidenav.spec.js
```diff
@@ -85,11 +84,11 @@ describe('mdSidenav', function() {
     it('should focus child with md-sidenav-focus', inject(function($rootScope, $material, $document, $compile) {
       jasmine.mockElementFocus(this);
       var parent = angular.element('<div>');
-      var markup = '<md-sidenav md-is-open="show">'+
-                      '<md-input-container><label>Label</label>' +
-                        '<input type="text" md-sidenav-focus>' +
-                      '</md-input-container>' +
-                    '<md-sidenav>';
+      var markup = '<md-sidenav md-is-open="show">' +
+        '<md-input-container><label>Label</label>' +
```
ErinCoughlan: Can you keep the indents here? Or add the spaces within the quotes so that your formatter won't mess it up?

ErinCoughlan:
LGTM. (with the few comments above) Now we don't need a silly one-time `$scope.$watch()`.

PR #7901, opened 2016-04-05
perf(progress-circular): switch back to keyframe animation for the rotation

crisbeto:
In #7402 I had to fall back to using JS to do the indeterminate rotation, because of IE and Edge which weren't accounting for the element's stroke and caused a wobble in the animation. I managed to work around this issue by specifying the `transform-origin` in all dimensions (x, y and z), as well as using the element's radius as the origin, instead of `center`.

@ThomasBurleson this also spares us one of the `$$rAF` loops that was causing problems with `$animate.flush`.


ThomasBurleson:
@crisbeto  - Very nicely done. :+1:

PR #7906, opened 2016-04-06
feat(menuFilter): Adding a menu filter directive.

### LINKED ISSUE #6643: menu: support sync filtering with input field
Support searchbox in the select menu, which filters the list of items.

DerekLouie:
@jelbourn @ErinCoughlan @gmoothart @KarenParker @ThomasBurleson 

Hello,

This is a directive which allows a developer to easily add a search bar to a component (it does substring based filtering). It is particularly keen for the select / menu components but is really a general filtering directive that can be used anywhere. This directive has an clear button that appears when there is a search term.

I tried to design it in such a way that no assumptions are made about how the developer is going to use it.

The demo has keyboard navigation and a sticky header baked in. Please note that this is not apart of the component (as no assumptions are made about where the developer is going to use this directive) but simply serves as an possible example use case.

Fixes #6643 and #7385.

This PR depends on #7782.

Please review. 

Screenshots:

md-menu-filter1

md-menu-filter2

### DIFF
--- a/src/components/menuFilter/demoBasicUsage/index.html
+++ b/src/components/menuFilter/demoBasicUsage/index.html
@@ -1,0 +1,26 @@
+<div ng-controller="menuFilterController as ctrl" class="md-padding" ng-cloak>
+  <div>
+    <h1 class="md-title">Pick a drink below</h1>
+    <div layout="row">
+      <md-input-container>
+        <label>Drinks</label>
+        <md-select ng-model="ctrl.selectedDrinks" multiple
+                   md-on-close="ctrl.clearSearch()"
+                   data-md-container-class="menuFilterdemoBasicUsage">
+          <md-select-header class="demo-select-header">
+            <md-menu-filter items="ctrl.drinks"
+                            placeholder="Select a drink.."
+                            filtered-items="ctrl.filteredDrinks"
+                            ng-model="ctrl.searchTerm"
+                            tabindex="1"
+                            class="demo-header-searchbox">
+            </md-menu-filter>
+          </md-select-header>
+          <md-optgroup label="drinks">
+            <md-option ng-value="drink" ng-repeat="drink in ctrl.filteredDrinks">{{drink}}</md-option>
+          </md-optgroup>
+        </md-select>
+      </md-input-container>
+    </div>
+ </div>
+</div>
--- a/src/components/menuFilter/menu-filter.js
+++ b/src/components/menuFilter/menu-filter.js
@@ -1,0 +1,146 @@
+/**
+ * @ngdoc module
+ * @name material.components.menuFilter
+ */
+
+angular.module('material.components.menuFilter', [
+  'material.core',
+])
+  .directive('mdMenuFilter', MdMenuFilterDirective);
+
+/**
+ * @ngdoc directive
+ * @name mdMenuFilter
+ * @module material.components.menuFilter
+ * @restrict E
+ *
+ * @description
+ * The menu filter directive is used to make adding a searchbar to
+ * the top of a component quick and painless.
+ *
+ * @param {string=} aria-label Optional label for accessibility. Only necessary if no placeholder or
+ *   explicit label is present.
+ * @param {Array=} items The items to filter. This is an input, it will act as a
+ *   list of items to display when there is no search term and to filter through
+ *   when a search term is present.
+ * @param {Array=} filtered-items The variable the filtered results will be stored in.
+ *  This will be overwritten with the values in the 'items' array that substring
+ *  match the 'ng-model' every time the 'ng-model' is updated.
+ *  every time the ng-model is changed.
+ * @param {string} ng-model Assignable angular expression to data-bind the
+ *   search term to, this is the variable the directive watches and triggers
+ *   filtering on.
+ * @param {string=} placeholder Placeholder hint text.
+ * @param {string=} tabindex Optional tab-index for accessibility.
+ *
+ * @usage
+ * <hljs lang="html">
+ *  <md-menu-filter items="ctrl.items"
+ *                  placeholder="Select an item.."
+ *                  filtered-items="ctrl.filteredItems"
+ *                  ng-model="ctrl.searchTerm"
+ *                  tabindex="1"
+ *                  aria-label="search box">
+ *  </md-menu-filter>
+ * </hljs>
+ *
+ * Please Note: When using this in the context of of an md-select directive w/
+ * the 'multiple' attribute enabled - this component cannot detect when that
+ * select menu closes. If the developer does not clear the ng-model on close,
+ * the closed menu will only display the filtered results (although the actual
+ * options selected may range beyond what is currently being filtered). An
+ * example of how one might clear the ng-model on close is provided in the basic
+ * demo.
+ */
+function MdMenuFilterDirective($mdTheming, $mdAria) {
+  return {
+    restrict: 'E',
+    template: '<input ng-model="searchTerm" type="search"' +
+        'class="md-menu-filter _md-text">' +
+            '<button' +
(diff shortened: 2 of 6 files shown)


ThomasBurleson:
Based on talks regarding menu vs select... these features should only be added to select; not menu.

ErinCoughlan:
@ThomasBurleson This is a generic component that _can_ be used in select. It's never explicitly added to it. I could also use the menu-filter in a custom component I'm building without it being inside another material one, which would keep the right styling and behavior without needing extra work. Do you have a better name that we could use? I also thought `md-list-filter` could work, but that's just as confusing.

src/components/menuFilter/menu-filter.js
```diff
@@ -0,0 +1,146 @@
...
+ *
+ * @description
+ * The menu filter directive is used to make adding a searchbar to
+ * the top of a component quick and painless.
+ *
+ * @param {string=} aria-label Optional label for accessibility. Only necessary if no placeholder or
+ *   explicit label is present.
+ * @param {Array=} filtered-items The variable the filtered
+ *   results will be stored in.
+ * @param {Array=} items The items to filter.
```
ErinCoughlan: Maybe using input vs output would help explain the difference between filteredItems and items. You should also be explict about which you are watching for updates. For instance, if I change filteredItems, what will happen in the UI?

src/components/menuFilter/menu-filter.js
```diff
@@ -0,0 +1,146 @@
...
+                ' type="button"' +
+                ' tabindex="-1"' +
+                ' ng-if="searchTerm"' +
+                ' class="search-clear-button"' +
+                ' ng-click="clear()">' +
+              '<md-icon md-svg-icon="md-close"></md-icon>' +
+              '<span class="_md-visually-hidden">Clear</span>' +
+            '</button>',
+    scope: {
+      filteredItems: '=',
```
ErinCoughlan: What happened to `aria-label`, `placeholder`, and `tabindex`?

src/components/menuFilter/menu-filter.js
```diff
@@ -0,0 +1,146 @@
...
+
+    function init() {
+      // This is for the clear button that appears when there is a search term.
+      scope.clear = function() {
+        scope.searchTerm = '';
+      };
+
+      // Set our tabindex of the md-menu-filter directive to -1 by default,
+      // because the select/menu input will hold the actual tabindex.
+      element.find('input').attr('tabindex', attr.tabindex || '-1');
```
ErinCoughlan: Hmmm, this means that this component _must_ be used within a menu or select (really only select since menus shouldn't allow search). I had thought that we wanted to make something generic that I could even use it within my custom component outside of material.

src/components/menuFilter/menu-filter.js
```diff
@@ -0,0 +1,146 @@
...
+      element.on('keydown', onMenuKeyDown);
+
+      scope.$on('$destroy', function() {
+        element.off('keydown', onMenuKeyDown);
+      });
+
+      // The md-select eats keydown events which would otherwise disallow users to
+      // type in this directive's input unless this is present.
+      function onMenuKeyDown(ev) {
+        // We still want to close the dialog on esc
```
ErinCoughlan: If we're in a dialog with that functionality set...

I think all key handlers should be controller by the user since we don't know what functionality they want.

src/components/menuFilter/menu-filter.scss
```diff
@@ -0,0 +1,21 @@
+md-menu-filter {
+  &:focus {
+    outline: none;
+  }
+
+  .search-clear-button {
```
ErinCoughlan: Should this be a private class?

src/components/menuFilter/menu-filter.js
ErinCoughlan: I think this prevents me from setting tabindex to 0. Please add a test for that case as well.

src/components/menuFilter/menu-filter.spec.js
```diff
@@ -0,0 +1,128 @@
...
+    expect(filterMenuClearButton).not.toBe(null);
+  });
+
+  it('should clear search term with clear button', function() {
+    scopeWithController.ctrl.searchTerm = 'cool beans';
+
+    var filterMenuClearButton = setupFilterMenu().find('button');
+
+    filterMenuClearButton.triggerHandler('click');
+    scopeWithController.$digest();
```
ErinCoughlan: $apply instead of $digest?

src/components/menuFilter/menu-filter.spec.js
```diff
@@ -0,0 +1,128 @@
...
+    };
+
+    filterMenu.triggerHandler(keydownEvent);
+
+    expect(keydownEvent.stopPropagation).not.toHaveBeenCalled();
+  });
+
+
+  function setupFilterMenu(attrs, scope) {
+    inject(function($compile, $rootScope) {
```
ErinCoughlan: Instead of injecting here, inject these in the beforeEach, save them to variables and use them here.

DerekLouie:
Hello,

I have addressed all the aforementioned comments. Now that the `md-select-header` PR is in, I have also attached screenshots.

Feel free to play with the demo, it is pretty fully featured (keyboard navigation and all).

Please review at your earliest convenience.

src/components/menuFilter/demoBasicUsage/index.html
```diff
@@ -0,0 +1,26 @@
...
+  <div>
+    <h1 class="md-title">Pick a drink below</h1>
+    <div layout="row">
+      <md-input-container>
+        <label>Drinks</label>
+        <md-select ng-model="ctrl.selectedDrinks" multiple
+                   md-on-close="ctrl.clearSearch()"
+                   data-md-container-class="menuFilterdemoBasicUsage">
+          <md-select-header class="demo-select-header">
+            <md-menu-filter items="ctrl.drinks"
```
ThomasBurleson: @DerekLouie - I keep returning to this name `md-menu-filter` which I find VERY confusing.

* What is the intent of this directive?
* IMO, I think a better name should be used.

DerekLouie: @ThomasBurleson 

The hope for this directive is essentially to have a directive that a user can drop in and give a list of items (an array of strings) and that will give them a list of matching items (based on what the user types in the search input).

In essence we wanted a directive users could just drop anywhere that takes a list of strings and gives them an input box to filter those items.

It will be up to the developer on how they want to display those filtered items.

I agree the name could be better, but I am still trying to think of a better description for what is going on here, any suggestions off the top of your head?

topherfangio:
@DerekLouie I just did another quick review of this PR and it looks like there might be some styling issues after my latest changes to the `<md-select>`.

Additionally, there is an interesting issue with the floating label. In the first picture, the input had focus (so the label floats), but in the second picture, it was not (so it doesn't float).

screen shot 2016-06-20 at 3 02 42 pm
screen shot 2016-06-20 at 3 02 49 pm

Think you could rebase and take a look at these issues?

ThomasBurleson:
@ErinCoughlan - can you get Derek to fix this. 
> It will not be v1.1.0 since the fix has been so delayed.

ErinCoughlan:
@ThomasBurleson Yes, this feature was prioritized by my team around launch, which is why it was never finished. I think there have been enough changes to mdSelect in the last three months that it may make sense to abandon this and start a new component if it comes up again. The select can already hold search boxes without it thanks to the md-select-header changes. This would just make the styling more consistent across all uses and make it slightly easier for teams to implement (although there is a demo of a searchbox in select already). What do you think?

topherfangio:
Closing per @ErinCoughlan's feedback.

PR #7915, opened 2016-04-06
feat(panel): Inject mdPanelRef when instantiating the panel controller.

### LINKED ISSUE #7177: mdPanel: Inject mdPanelRef when instantiating the panel controller.
Figure out how to do this:

1. mdPanelRef doesn't exist until the panel is created.
2. mdPanelRef needs to be injected during creation.

KarenParker:
Fixes #7177.

@ErinCoughlan @gmoothart @DerekLouie @jelbourn Please review.

### DIFF
--- a/src/components/panel/demoBasicUsage/script.js
+++ b/src/components/panel/demoBasicUsage/script.js
@@ -1,6 +1,6 @@
 angular.module('panelDemo', ['ngMaterial'])
     .controller('BasicDemoCtrl', BasicDemoCtrl)
-    .controller('DialogCtrl', DialogCtrl);
+    .controller('PanelDialogCtrl', PanelDialogCtrl);
 
 
 function BasicDemoCtrl($mdPanel) {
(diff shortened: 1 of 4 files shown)


jelbourn:
LGTM

src/components/panel/panel.spec.js
```diff
@@ -597,12 +598,13 @@ describe('$mdPanel', function() {
 
     it('should allow locals to be injected in the controller', function() {
```
ErinCoughlan: optional: Add a test that a panel can close itself. That's the main benefit (I think) of injecting MdPanelRef.

ErinCoughlan:
LGTM

src/components/panel/demoBasicUsage/script.js
```diff
@@ -32,10 +29,12 @@ BasicDemoCtrl.prototype.showDialog = function() {
 };
 
 
-BasicDemoCtrl.prototype.closeDialog = function() {
-  this._panelRef && this._panelRef.close();
-};
+// Necessary to pass locals to the dialog template.
+function PanelDialogCtrl(mdPanelRef) {
```
ThomasBurleson: I strongly suggest a different name for `mdPanelRef`.

*  This is a reference to some instance ?
*  What type of instance does this reference ?
*  `Ref` seems very cludgy

jelbourn:
@ThomasBurleson `Ref` is the term Angular 2 uses for this sort of thing (`ComponentRef`, `ApplicationRef`, `ElementRef`, `TemplateRef`, etc.). The naming simply says "This is a reference to some existing panel." I think it does a good job of conveying what you're injecting.

PR #7919, opened 2016-04-06
fix(input): prevent the input from jumping when it becomes disabled

### LINKED ISSUE #7640: Disabled fieldset modifies md-input-container height
May be related to https://github.com/angular/material/issues/7639

Click on enable/disable button. There is a small height glitch.
http://jsbin.com/lafoqureru/1/edit?html,js,output

crisbeto:
* The inputs used to offset their border with a -1px margin, however this caused all surrounding content to jump. This change switches to using `background-position` instead.
* Fixes inputs not getting their disabled styles when they're in a disabled fieldset.

Closes #7640.

### DIFF
--- a/src/components/input/input-theme.scss
+++ b/src/components/input/input-theme.scss
@@ -76,7 +76,7 @@ md-input-container.md-THEME_NAME-theme {
   }
   .md-input {
     &[disabled],
-    [disabled] {
+    [disabled] & {
       border-bottom-color: transparent;
       color: '{{foreground-3}}';
       background-image: linear-gradient(to right, '{{foreground-3}}' 0%, '{{foreground-3}}' 33%, transparent 0%);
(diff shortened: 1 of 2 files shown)


src/components/input/input-theme.scss
```diff
@@ -76,7 +76,7 @@ md-input-container.md-THEME_NAME-theme {
   }
   .md-input {
     &[disabled],
-    [disabled] {
+    [disabled] &{
```
DevVersion: Nitpicky: Space bracket from the `&`.

src/components/input/input.scss
```diff
@@ -307,12 +307,13 @@ md-input-container {
   .md-input {
     &[disabled],
     [disabled] & {
-      background-position: 0 bottom;
+      // The negative border width offsets the dotted "border" so
+      // it's placed in the same place as the solid one before it.
+      background-position: bottom $input-border-width-default*-1 left 0;
```
DevVersion: Also the arithmetic operator should have space around.

PR #7920, opened 2016-04-06
fix(tabs): not selected tab text color as spec

EladBezalel:
- In da48b6c9a6247c9175d69f7f256a6e4fd452921f the color changed from 100 to A100, after checking the [spec](https://www.google.com/design/spec/components/tabs.html#tabs-specs), the color should be:

> Unfocused tab color: `#fff 70%`

### DIFF
--- a/src/components/tabs/tabs-theme.scss
+++ b/src/components/tabs/tabs-theme.scss
@@ -1,67 +1,21 @@
-@mixin md-tab-primary {
+@mixin md-tab-theme($theme) {
   > md-tabs-wrapper {
-    background-color: '{{primary-color}}';
+    background-color: '{{#{$theme}-color}}';
     > md-tabs-canvas {
       > md-pagination-wrapper {
         > md-tab-item:not([disabled]) {
-          color: '{{primary-100}}';
           &.md-active, &.md-focused {
             &, md-icon {
-              color: '{{primary-contrast}}';
+              color: '{{#{$theme}-contrast}}';
             }
           }
           &.md-focused {
-            background: '{{primary-contrast-0.1}}';
+            background: '{{#{$theme}-contrast-0.1}}';
           }
         }
       }
     }
   }
-}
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@DevVersion - please review.

DevVersion:
@EladBezalel I'm running your changes locally and taking a look at some different style configurations. (Grey Primary) 

But I'm wondering, why you applied those changes only to the `accent` styling? 
`primary` and `warn` are using the `100` hue.

EladBezalel:
@DevVersion thank you i didn't think of it

EladBezalel:
@DevVersion please review

src/components/tabs/tabs-theme.scss
```diff
@@ -97,25 +51,37 @@ md-tabs.md-THEME_NAME-theme {
     }
   }
 
+  &.md-accent, &.md-primary, &.md-warn{
```
DevVersion: Nitpicky: Space curly brace.

src/components/tabs/tabs-theme.scss
```diff
@@ -97,25 +51,37 @@ md-tabs.md-THEME_NAME-theme {
...
+    > md-tabs-wrapper {
+      > md-tabs-canvas {
+        > md-pagination-wrapper {
+          > md-tab-item:not([disabled]) {
+            color: rgba(255, 255, 255, 0.7);
```
DevVersion: I'm still thinking about using a hard-coded color. Is the Black Color also valid for dark themes? I know the specs doesn't mention that, I'm just curious. What do you think? :)

EladBezalel: This is not theme related, the background of tabs will look the same in dark and light themes,
this is only the not selected tabs labels and it should be 70% white according the spec

ThomasBurleson:
Closing. A different solution is needed.

PR #7934, opened 2016-04-07
fix(autocomplete): don't show the loading bar when hitting escape on an empty input

### LINKED ISSUE #7927: mdAutocomplete: Progress bar is visible after pressing ESC
Since Angular Material v1.0.7 a progress bar is visbile after pressing **ESC** inside an **empty** autocomplete. 

Example from the demo page:
image

crisbeto:
The loading bar was being shown if the user pressed escape on an empty autocomplete input.

Closes #7927.

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -443,7 +443,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       case $mdConstant.KEY_CODE.ESCAPE:
         event.stopPropagation();
         event.preventDefault();
-        clearValue();
+        if ($scope.searchText) clearValue();
 
         // Force the component to blur if they hit escape
         doBlur(true);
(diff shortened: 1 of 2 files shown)


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -618,7 +618,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     // Set the loading to true so we don't see flashes of content.
     // The flashing will only occour when an async request is running.
     // So the loading process will stop when the results had been retrieved.
-    setLoading(true);
+    if ($scope.searchText) setLoading(true);
```
DevVersion: I think, since the clear button is hidden, when the search text is empty, the autocomplete should not call `clearValue()` when pressing <kbd>ESC</kdb> as well.

crisbeto: That works too, thanks for the input.

DevVersion:
LGTM now.

PR #7936, opened 2016-04-07
fix(progress-circular): use a non-flushable requestAnimationFrame

crisbeto:
Reverts to using the plain `requestAnimationFrame`, instead of `$$rAF`, in order to avoid infinite loops when calling `$animate.flush`.

@ThomasBurleson this was still causing infinite loops in some of @robertmesserle's tests, so we decided that using `$$rAF` doesn't provide any benefits in this case and we can use the plain `requestAnimationFrame`, which isn't flushable by Angular.

### DIFF
--- a/src/components/progressCircular/js/progressCircularDirective.js
+++ b/src/components/progressCircular/js/progressCircularDirective.js
@@ -48,9 +48,13 @@ angular
   .directive('mdProgressCircular', MdProgressCircularDirective);
 
 /* @ngInject */
-function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdTheming,
+function MdProgressCircularDirective($window, $mdProgressCircular, $mdTheming,
                                      $mdUtil, $interval, $log) {
 
+  // Note that this shouldn't use use $$rAF, because it can cause an infinite loop
+  // in any tests that call $animate.flush.
+  var rAF = $window.requestAnimationFrame || angular.noop;
+  var cAF = $window.cancelAnimationFrame || angular.noop;
   var DEGREE_IN_RADIANS = $window.Math.PI / 180;
   var MODE_DETERMINATE = 'determinate';
   var MODE_INDETERMINATE = 'indeterminate';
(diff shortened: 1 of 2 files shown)


robertmesserle:
LGTM

robertmesserle:
But still want @ThomasBurleson to weigh in

src/components/progressCircular/js/progressCircularDirective.js
```diff
@@ -200,7 +197,9 @@ function MdProgressCircularDirective($$rAF, $window, $mdProgressCircular, $mdThe
       if (animateTo === animateFrom) {
         path.attr('d', getSvgArc(animateTo, diameter, pathDiameter, rotation));
       } else {
-        lastDrawFrame = $$rAF(function animation(now) {
+        // Note that this shouldn't use use $$rAF, because it can cause an infinite loop
+        // in any tests that call $animate.flush.
+        $window.requestAnimationFrame(function animation(now) {
```
ThomasBurleson: Note that the ng1 `RAFProvider` does this:

```js
function $$RAFProvider() { //rAF
  this.$get = ['$window', '$timeout', function($window, $timeout) {
    var requestAnimationFrame = $window.requestAnimationFrame ||
                                $window.webkitRequestAnimationFrame;

    var cancelAnimationFrame = $window.cancelAnimationFrame ||
                               $window.webkitCancelAnimationFrame ||
                               $window.webkitCancelRequestAnimationFrame;

    var rafSupported = !!requestAnimationFrame;
```

With the can on L202, we lose all of that! My first thought is that this is the wrong solution.

ThomasBurleson: Why not put a loop counter in our `$$rAF` decorator in `/test/angular-material-mocks.js`#L70 ?

crisbeto: We do lose all that, however we don't support any browsers that either don't have it, or use the prefixed version.

robertmesserle: We do lose all of that, but we could very easily replicated it if we needed.  The real problem is that we are trying to flush something that is by definition not meant to be flushed.  This is strictly animation code as far as I can tell.  It becomes (a) a lot more work and (b) a lot more fragile to try to keep it using $$rAF

robertmesserle: I'd be curious if we get any concrete advantages to using `$$rAF` which is why I wanted a discussion before merging this in.  It's a good point that we don't actually support any browsers that don't support `requestAnimationFrame` out of the box.

ThomasBurleson: Let's:

* use `var raf = ($window.requestAnimationFrame || angular.noop);` to account for possible undefined API (in node situations). 
* add support for `cancelAnimationFrame` on `$destroy`

ThomasBurleson:
:+1:

crisbeto:
Haha that was fast. I was still waiting on Travis.

Issue #6042, opened 2015-12-02
mdAutocomplete not passing Accessibility Test


ByzanTine:
![image](https://cloud.githubusercontent.com/assets/5234867/11545654/4c9f7c1a-98fd-11e5-954b-b2aab5942d3b.png)
There are two errors.
Elements with ARIA roles must ensure required owned elements are present.
This element has an unsupported ARIA attribute.
Run the Chrome Accessibility Tool inside Audit in Chrome Developer Tool will replay this issue.

marcysutton:
For the owned elements error:
You could try moving `aria-owns="md-autocomplete-suggestions"` from `input` to mdAutocomplete (which has `role="listbox"`) and changing `role="button"` to `role="option"` on each virtual repeat list item.

For the unsupported ARIA attribute error:
Try moving `aria-autocomplete="listbox"` to mdAutocomplete, which has the `listbox` role. http://www.w3.org/TR/wai-aria/states_and_properties#aria-autocomplete

marcysutton:
Here is a combobox demo: http://oaa-accessibility.org/example/10/

jelbourn:
@marcysutton @ByzanTine: does the autocomplete actually have an usability problems in our supported browser + screen reader combinations? The audit from the tool is useful, but it's possible everything is working fine as is, in which case I would put a lower priority on this.

marcysutton:
In Voiceover and Safari, the field is announced as "search text field", but not listbox...so it's hard to tell that it's anything other than a plain text field. But I could see this kind of issue being a lower priority, since it is currently operable.

In JAWS 16 and IE, we have a bigger problem: although the widget is announced correctly as a listbox, the options and screen reader messages are all announced literally as "left brace left brace message right brace right brace", so there something wrong with the interpolation. This makes the widget completely unusable. If someone can check it in JAWS 17, that would be good.

NVDA and Firefox seems to work okay, but I'd be curious to hear from a proficient screen reader user. On touch devices with a screen reader running, the autocomplete options are pretty unusable...but I suspect most autocompletes would suffer from the same problem. 

In short: with the exception of JAWS incorrectly reading messages, any changes could come after actual user testing.

biomassives:
This message was created automatically by mail delivery software.

A message that you sent could not be delivered to one or more of its
recipients. This is a temporary error. The following address(es) deferred:

  acmeideal@gmail.com
[...]

PR #7941, opened 2016-04-07
[accessibility] Autocomplete should use role=combobox

tirana:
from https://www.w3.org/TR/wai-aria/roles#combobox
"Typically editable combo boxes are used for autocomplete behavior, and authors SHOULD set aria-autocomplete attribute on the textfield."

### DIFF
--- a/src/components/autocomplete/js/autocompleteDirective.js
+++ b/src/components/autocomplete/js/autocompleteDirective.js
@@ -174,8 +174,7 @@ function MdAutocomplete () {
       return '\
         <md-autocomplete-wrap\
             layout="row"\
-            ng-class="{ \'md-whiteframe-z1\': !floatingLabel, \'md-menu-showing\': !$mdAutocompleteCtrl.hidden }"\
-            role="listbox">\
+            ng-class="{ \'md-whiteframe-z1\': !floatingLabel, \'md-menu-showing\': !$mdAutocompleteCtrl.hidden }">\
           ' + getInputElement() + '\
           <md-progress-linear\
               class="' + (attr.mdFloatingLabel ? 'md-inline' : '') + '"\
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@marcysutton  - can you review ?

marcysutton:
Is a fix for an issue with the usability of the autocomplete, or have you tested it? I know of one issue that is open related to this, but it worked correctly in a variety of browser and AT combos when we chose the current semantics. https://github.com/angular/material/issues/6042

tirana:
1. a user who is blind reported an issue with a search box. the our search box is md-autocomplete.  
the element has annotation for screen reader yet it's incorrect.
our accessibility QA said that role was wrong. 
2. I looked at google.com search box which had similar logic and found that role=combobox was used.
3. I looked at the doc and found that combobox is recommend role for autocomplete input fields. https://www.w3.org/TR/wai-aria/roles#combobox

testing: I uses http://www.chromevox.com/ for testing.
using dev tool I manually changed role and compared description
edit text vs combobox autocompletion

tirana:
BTW, #6042 the first point is exactly the same what I did
_"Remove role="listbox" from <md-autocomplete-wrap>, add role="combobox" to <input>, "_

tirana:
@marcysutton Does it make sense?

marcysutton:
Makes sense to me, I just don't have time to test it at the moment. If you can confirm it still works with the Material-supported browsers and a few screen readers, that would help a lot. Some good targets for that would be Firefox/NVDA, Safari/Voiceover, and optionally IE11 and JAWS (which has a free 40 minute mode). Chromevox nor Chrome/Voiceover are widely used, so we didn't focus on those in the past.

tirana:
confirmed. combobox sounds better

marcysutton:
Yay, that's great! LGTM, ship it!

PR #7935, opened 2016-04-07
Added option to disable backdrop for md-menu directive;

tahq69:
<md-menu use-backdrop="false"> - will disable backdrop and allow use menu in contextmenu event;
If backdrop is disabled, you should manually close menu on own events - same as md-prevent-menu-close;

PR #7953, opened 2016-04-08
(fixed) Added option to disable backdrop for md-menu

tahq69:
Correct implementation of #7935

< md-menu use-backdrop="false" > - will disable backdrop and allow use menu in contextmenu event;
If backdrop is disabled, you should manually close menu on own events - same as md-prevent-menu-close;

### DIFF
--- a/src/components/menu/js/menuController.js
+++ b/src/components/menu/js/menuController.js
@@ -13,7 +13,7 @@ function MenuController($mdMenu, $attrs, $element, $scope, $mdUtil, $timeout, $r
   var self = this;
   var triggerElement;
 
-  this.nestLevel = parseInt($attrs.mdNestLevel, 10) || 0;
+  this.nestLevel = getAttr('mdNestLevel', 0);
 
   /**
    * Called by our linking fn to provide access to the menu-content
@@ -118,6 +118,7 @@ function MenuController($mdMenu, $attrs, $element, $scope, $mdUtil, $timeout, $r
     triggerElement = triggerElement || (ev ? ev.target : $element[0]);
     triggerElement.setAttribute('aria-expanded', 'true');
     $scope.$emit('$mdMenuOpen', $element);
+    var hasBackdrop = getAttr('useBackdrop', true);
     $mdMenu.show({
       scope: $scope,
       mdMenuCtrl: self,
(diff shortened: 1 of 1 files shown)


src/components/menu/js/menuController.js
```diff
@@ -222,4 +225,42 @@ function MenuController($mdMenu, $attrs, $element, $scope, $mdUtil, $timeout, $r
...
+  }
+
+  /**
+   * Convert value to type
+   *
+   * @param {String} type Object type
+   * @param {object} value Value to be converted
+   * @returns {*} Converted value to type
+   */
+  function parseTypeOf(type, value) {
```
DevVersion: I don't like those *two* new methods. They're in my opinion not really necessary.

src/components/menu/js/menuController.js
```diff
@@ -118,14 +118,17 @@ function MenuController($mdMenu, $attrs, $element, $scope, $mdUtil, $timeout, $r
     triggerElement = triggerElement || (ev ? ev.target : $element[0]);
     triggerElement.setAttribute('aria-expanded', 'true');
     $scope.$emit('$mdMenuOpen', $element);
+    var hasBackdrop = getAttr('useBackdrop', true);
```
DevVersion: Same functionality, without the new method 
```js
angular.isDefined($attrs.useBackdrop) ? $attrs.useBackdrop : true;
```

But in case of using it as an attribute, it would better call it `md-disable-backdrop`. Because `use-backdrop="true"` is actually not making a sense.

src/components/menu/js/menuController.js
```diff
@@ -118,14 +118,17 @@ function MenuController($mdMenu, $attrs, $element, $scope, $mdUtil, $timeout, $r
...
     $mdMenu.show({
       scope: $scope,
       mdMenuCtrl: self,
       nestLevel: self.nestLevel,
       element: menuContainer,
       target: triggerElement,
       preserveElement: true,
-      parent: 'body'
+      parent: 'body',
+      hasBackdrop: hasBackdrop,
```
DevVersion: Please add a test for that. 

Additionally, I can't see the actual implementation of the `hasBackdrop` option? Is this not implemented yet?

ThomasBurleson:
Will not merge. Implementation is not DRY.

PR #7957, opened 2016-04-08
fix(aria): $mdAria should not use texts from aria-hidden nodes

### LINKED ISSUE #7376: [a11y] aria-hidden/aria-label on <md-icon> inside of button is ignored with ligature icon
See demo:
http://codepen.io/anon/pen/mPJOMX?editors=1000

I want to hide the gear icon entirely from screen readers, as it doesn't add anything. But `aria-hidden="true"` and `aria-label=""` don't work. The only thing that works is a non-empty, non-whitespace aria-label. But I don't want the icon to have any label; I just want it ignored, and for the label for the button to just be "Options"

DevVersion:
* $mdAria is currently just using `textContent` for retrieving the `aria-label`.
  This is not valid, since some child-texts can be hidden in aria (`aria-hidden="true")

* Using a TreeWalker is super elegant and performant.
  The current use of the TreeWalker is supported by all browser we support.

Fixes #7376

### DIFF
--- a/src/core/services/aria/aria.js
+++ b/src/core/services/aria/aria.js
@@ -50,7 +50,7 @@ function AriaService($$rAF, $log, $window, $interpolate) {
 
   function expectWithText(element, attrName) {
     var content = getText(element) || "";
-    var hasBinding = content.indexOf($interpolate.startSymbol())>-1;
+    var hasBinding = content.indexOf($interpolate.startSymbol()) > -1;
 
     if ( hasBinding ) {
       expectAsync(element, attrName, function() {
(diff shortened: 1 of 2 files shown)


DevVersion:
FYI: Waiting for @topherfangio's fix for the current CI issues :) Thx.

ThomasBurleson:
@DevVersion - please rebase.

src/core/services/aria/aria.js
```diff
@@ -62,7 +62,26 @@ function AriaService($$rAF, $log, $window, $interpolate) {
   }
 
   function getText(element) {
-    return (element.text() || "").trim();
+    element = element[0] || element;
+    var walker = document.createTreeWalker(element, NodeFilter.SHOW_TEXT, null, false);
+    var text = '';
+
+    var node;
+    while (node = walker.nextNode()) {
```
topherfangio: Rather than walking through all of the nodes here, can you not just use the third parameter `filter` of the `createTreeWalker()` method? I would assume you could just pass the `isAriaHiddenNoe()` method to it (wrapped in an object I think) and it would filter the nodes for you.

I guess you would still have to traverse the remaining nodes to add the text content, but it might shorten the code for you a bit.

DevVersion: @topherfangio I intentionally didn't use the filter for the `TreeWalker`, because it's not supported by IE browsers and this will break ngMaterial on IE browsers. 

Additionally, the filter is actually doing the same as me, just a shorter code + The amount of nodes is really minimized, because we're only iterating through the text nodes.

DevVersion:
@ThomasBurleson Done, thanks for the fix :+1:

PR #7961, opened 2016-04-09
feat(floating labels) use expressions

daniel-nagy:
Use expressions to conditionally disable floating labels. Floating labels will be disabled if

1. The `md-no-float` attribute exists and is empty, or
2. The `md-no-float` attribute evaluates to true.

In the docs, `@param md-no-float {boolean=}`.

### DIFF
--- a/src/components/input/input.spec.js
+++ b/src/components/input/input.spec.js
@@ -346,6 +346,49 @@ describe('md-input-container directive', function() {
     expect(label.textContent).toEqual('some placeholder');
   });
 
+  it('should not create a floating label from a placeholder if md-no-float is empty', function () {
+    var el = compile(
+      '<md-input-container md-no-float>' +
+      '  <input placeholder="Foo" ng-model="foo">' +
+      '</md-input-container>'
+    );
+
+    expect(el.find('label').length).toBe(0);
+  });
+
+  it('should not create a floating label from a placeholder if md-no-float is truthy', function () {
+    pageScope.inputs = [{
+      placeholder: 'Name',
+      model: ''
+    }, {
+      placeholder: 'Email',
+      model: ''
+    }];
+
+    var el = compile(
+      '<div>' +
+      '  <md-input-container ng-repeat="input in inputs" md-no-float="$index !== 0">' +
+      '    <input placeholder="{{input.placeholder}}" ng-model="input.model">' +
+      '  </md-input-container>' +
+      '</div>'
+    );
+
+    var labels = el.find('label');
+
+    expect(labels.length).toBe(1);
+    expect(labels[0].textContent).toEqual('Name');
+  });
+
+  it('should create a floating label from a placeholder if md-no-float is falsey', function () {
+    var el = compile(
+      '<md-input-container md-no-float="false">' +
+      '  <input placeholder="Foo" ng-model="foo">' +
+      '</md-input-container>'
+    );
+
+    expect(el.find('label').length).toBe(1);
+  });
+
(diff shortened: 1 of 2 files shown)


DevVersion:
@daniel-nagy Now if we're using `$eval`, then it's rather an expression than a boolean. Boolean Attributes can be easily parsed with our util function. 
```js
$mdUtil.parseAttributeBoolean(attrs.mdNoFloat)
```

daniel-nagy:
Either way works, I suppose it comes down two which syntax you believe is more appropriate,

```html
<md-input-container md-no-float="expression"></md-input-container>
<!-- or -->
<md-input-container md-no-float="{{expression}}"></md-input-container>
```

src/components/input/input.spec.js
```diff
@@ -346,6 +346,29 @@ describe('md-input-container directive', function() {
     expect(label.textContent).toEqual('some placeholder');
   });
 
+  it('should not create a floating label from a placeholder if md-no-float exists and is empty, or evaluates to true', function () {
```
topherfangio: Perhaps I'm not seeing the corresponding code, but I think the description here is slightly misleading since this doesn't seem to test the case of `md-no-float exists and is empty`.

I think we may want to separate this into two tests if that's possible:

1. Test the empty case
2. Test the true case

daniel-nagy: You are correct, I am not testing for when the attribute exist but is empty.

topherfangio: Could you add that test real fast, or perhaps update the existing test's description to just say `md-no-float evaluates to true`? :smile:

daniel-nagy: I added a second test f6cf66d 👍

topherfangio:
Awesome; thanks for adding that extra test.

Just to confirm, this doesn't watch the expression, correct? It just evaluates it the first time?

daniel-nagy:
Correct, it does not watch the expression. In addition, this will have no effect if the `<label>` element is explicitly used.

I don't know if this is already in the works but it would be really nice to standardize all inputs. By that I mean ensure each input is the same height, can have errors or hint text, and configurable to enable/disable the floating label. [Here](http://codepen.io/anon/pen/KzQRVb) you can see what I mean by ensure each has the same height.

src/components/input/input.spec.js
```diff
@@ -346,6 +346,39 @@ describe('md-input-container directive', function() {
...
+    var el = compile(
+      '  <md-input-container md-no-float>' +
+      '    <input placeholder="Foo" ng-model="foo">' +
+      '  </md-input-container>'
+    );
+
+    expect(el.find('label').length).toBe(0);
+  });
+
+  it('should not create a floating label from a placeholder if md-no-float evaluates to true', function () {
```
ThomasBurleson: Please add 1 more test for `md-no-float` expression evaluating to 'false'

daniel-nagy:
Added another test for when `md-no-float` evaluates to `false`.

```javascript
scope.$eval('false'); // -> false
```

topherfangio:
@daniel-nagy Thanks for adding the extra test! Is this ready for re-review or is there more work that needs to be done?

daniel-nagy:
@topherfangio this is ready for re-review.

ThomasBurleson:

DevVersion:
@daniel-nagy This LGTM. 

One little thing - can you reference #8605 in the commit message?

daniel-nagy:
@DevVersion I added a reference to the commit message.

topherfangio:
@ThomasBurleson LGTM

Issue #7869, opened 2016-04-04
mdTooltip: incorrectly reappears when sidenav closes


sirbarrence:
As noted in #3525, when closing, an mdSidenav intentionally refocuses whatever element triggered it to open for accessibility reasons. 

However, if there is an mdTooltip on the triggering element, this programmatic refocus of that element causes the tooltip to reappear. This doesn't seem like expected behavior for mdTooltip. Here it is in action: 

![output](https://cloud.githubusercontent.com/assets/816158/14258126/f93f50b2-fa6e-11e5-8090-549c89dbb300.gif)

That's a click on the chat bubble button, then a click outside the sidenav to close it, unexpected tooltip reappears, then another click to make the unexpected tooltip disappear.

This happens in Material v1.1.0-rc2 and v1.0.7 with Angular 1.5.3.
I've reproduced this in Chrome for Mac and IE 11 for Windows.

This happens with or without `md-autohide` on the `md-tooltip`.

I would suggest one of the following:

- Provide an attribute on `md-tooltip` to opt-out of focus events triggering the tooltip to appear. (Is `md-autohide` supposed to do this? It's not entirely clear from the docs. But it does not solve the problem currently.)
- Provide an attribute on `md-sidenav` to opt-out of the autofocus-trigger-elment-on-close behavior.
- Change `md-tooltip` so that it magically knows that programmatically-triggered focus events should not cause it to appear. No idea how you'd accomplish that.

DevVersion:
Duplicate of #5563 - Fix is already pending #7965

FreakTheMighty:
I've found a related behavior and want to make sure its duplicate of this ticket. It can be reproduced on the Speed Dial Demo here: https://material.angularjs.org/latest/demo/fabSpeedDial

On the second demo showing the use of tooltips

* On Chrome, change to mobile emulation
* Click on the hamburger fab
* Click on the "Facebook" button

Notice that the tooltip disappears then briefly reappears.

Even `md-autohide=false` is not respected because the element isn't currently in focus. https://github.com/angular/material/blob/afeb736d5a3e2f93f5e3960c25469f48e8c13c22/src/components/tooltip/tooltip.js#L197

I'm not sure if that's the correct behavior, but it appears that the button doesn't take focus when tapped on a mobile device. 
https://github.com/angular/material/blob/afeb736d5a3e2f93f5e3960c25469f48e8c13c22/src/components/button/button.js#L150

DevVersion:
[...]

PR #7965, opened 2016-04-09
feat(interaction): added service to detect last interaction

### LINKED ISSUE #5434: Ripple 'sticks' if using UI Router in the app under certain conditions

### LINKED ISSUE #5563: md-toolbar and sidenav focus problems

DevVersion:
This is a re-submit / rework of my old PR (#5589), should start from clean again. 

Sample use-case:
- Sidenav want's to restore focus after the pane gets closed, this should only happen for keyboard interactions.

Fixes #5563 Fixes #5434 Closes #5583

### DIFF
--- a/src/core/services/interaction/interaction.js
+++ b/src/core/services/interaction/interaction.js
@@ -1,0 +1,132 @@
+/**
+ * @ngdoc module
+ * @name material.core.interaction
+ * @description
+ * User interaction detection to provide proper accessibility.
+ */
+angular
+  .module('material.core.interaction', [])
+  .service('$mdInteraction', MdInteractionService);
+
+
+/**
+ * @ngdoc service
+ * @name $mdInteraction
+ * @module material.core.interaction
+ *
+ * @description
+ *
+ * Service which keeps track of the last interaction type and validates them for several browsers.
+ * The service hooks into the document's body and listens for touch, mouse and keyboard events.
+ *
+ * The most recent interaction type can be retrieved by calling the `getLastInteractionType` method.
+ *
+ * Here is an example markup for using the interaction service.
+ *
+ * <hljs lang="js">
+ *   var lastType = $mdInteraction.getLastInteractionType();
+ *
+ *   if (lastType === 'keyboard') {
+ *     // We only restore the focus for keyboard users.
+ *     restoreFocus();
+ *   }
+ * </hljs>
+ *
+ */
+function MdInteractionService($timeout) {
+  this.$timeout = $timeout;
+
+  this.bodyElement = angular.element(document.body);
+  this.isBuffering = false;
+  this.bufferTimeout = null;
+  this.lastInteractionType = null;
+
+  // Type Mappings for the different events
+  // There will be three three interaction types
+  // `keyboard`, `mouse` and `touch`
+  // type `pointer` will be evaluated in `pointerMap` for IE Browser events
+  this.inputEventMap = {
+    'keydown': 'keyboard',
+    'mousedown': 'mouse',
+    'mouseenter': 'mouse',
+    'touchstart': 'touch',
+    'pointerdown': 'pointer',
+    'MSPointerDown': 'pointer'
+  };
+
+  // IE PointerDown events will be validated in `touch` or `mouse`
+  // Index numbers referenced here: https://msdn.microsoft.com/library/windows/apps/hh466130.aspx
+  this.iePointerMap = {
+    2: 'touch',
+    3: 'touch',
+    4: 'mouse'
+  };
+
+  this.initializeEvents();
+}
+
+/**
+ * Initializes the interaction service, by registering all interaction events to the
+ * body element.
+ */
+MdInteractionService.prototype.initializeEvents = function() {
+  // IE browsers can also trigger pointer events, which also leads to an interaction.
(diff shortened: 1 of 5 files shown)


src/core/services/interaction/interaction.js
```diff
@@ -0,0 +1,95 @@
...
+    $timeout.cancel(timer);
+
+    onInput(event);
+    buffer = true;
+
+    // The timeout of 650ms is needed to delay the touchstart, because otherwise the touch will call
+    // the `onInput` function multiple times.
+    timer = $timeout(function() {
+      buffer = false;
+    }, 650);
```
EladBezalel: two things:
1. is this timeout need to happen only if touchstart is being listened
2. have we already talked about the 650 timeout? why 650? there's no way to avoid that?

DevVersion: 1. Yes this is only needed for the touch event.
2. Yes, we already talked about that. 
    
The `650ms` are coming from Patrick H. Lauke (W3C)
> 300ms is causing iOS (which has default delay of ~350ms) to trigger touch, then mouse (once mouse compatibility events fire). through trial and error (on iPhone 6S) a value of around 650ms seems more appropriate (and still snappier than the original 1s from previous version)

crisbeto: I don't think this timeout needs to trigger a digest.

src/core/services/interaction/interaction.js
```diff
@@ -0,0 +1,95 @@
...
+  }
+
+  body.on('keydown', onInput);
```
crisbeto: This is pretty nitpicky and probably down to preference, but you can combine the calls like `body.on('keydown mouseenter ' + mouseEvent, onInput);`

DevVersion: That works too, but listing those events in order is more clear here.

src/core/services/interaction/interaction.js
DevVersion: That is a good catch @crisbeto Thanks!

crisbeto:
Apart from that digest on the `$timeout`, LGTM.

DevVersion:
@crisbeto Thanks for the review. @ThomasBurleson @EladBezalel PTAL.

FreakTheMighty:
Do ya'll expect this will resolve https://github.com/angular/material/issues/7869#issuecomment-218598387

DevVersion:
@FreakTheMighty Yes it will.

topherfangio:
@bradrich and @topherfangio to review.

DevVersion:
Ping @bradrich and @topherfangio for review.

topherfangio:
This looks good to me. Quick question though (and this may be totally irrelevant), but your input map only handles `keydown`.

Does this mean that if someone holds down a key, then presses a button with the mouse, then lets up the key `getLastInteractionType()` will incorrectly return the mouse event?

Like I said, this may be totally irrelevant/probably not a case we need to worry about; just curious more than anything :-)

DevVersion:
@topherfangio Yeah this is valid. 

> For example if you hold the mouse and quickly press a button while holding the mouse, it will incorrectly display use recognize a keyboard.

But I don't think it's worth addressing this kind of edge-case, because the intention is just to provide proper accessibility -> If he doesn't trigger the element with a keyboard, the focus will be not restored.

The only thing which could happen is, that the focus will be restored incorrectly, due to a click while holding a keyboard press.

I think the reason why we don't listen for the up events, is that we don't want to register too many listeners on the body element, because this is potentially performance damaging.

Here is a package, which does the same
> https://github.com/ten1seven/what-input/blob/master/src/what-input.js#L197

topherfangio:
Cool; that's what I thought.

LGTM 👍

ThomasBurleson:
Fixes #8749

PR #7966, opened 2016-04-09
fix(switch, slider): click, mouse events should not propagate

### LINKED ISSUE #7937: mdSwitch: within mdListItem, mdSwitch toggles on mdSlider
I've got a md-list-item which contains both a md-slider as well as a md-switch. Whenever the user clicks/drags the md-slider, this event is also propagated to the md-switch which then toggles. Whenever other controls are present within the same list-item, md-switch should only toggle by clicks on the control itself. 

Codepen: http://codepen.io/jsiegmund/pen/repwyR

Angular version: 1.5.3
Browser: using Chrome 49.0.2623.110 on Windows 10
Possibly related: https://github.com/angular/material/issues/2594

ThomasBurleson:
Switch events should not trigger with Slider listeners.
Slider events should not trigger Switch listeners.

*  [mdList Demo](http://codepen.io/team/AngularMaterial/pen/aNEeBZ)
*  [div Demo](http://codepen.io/team/AngularMaterial/pen/xVpvLR)

> note Codepen uses http://localhost:8080  with `angular-material.js` build from this PR

Fixes #7937

### DIFF
--- a/src/components/slider/slider.js
+++ b/src/components/slider/slider.js
@@ -369,7 +369,7 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
       }
     }
 
-    function mouseDownListener() {
+    function mouseDownListener(ev) {
       redrawTicks();
 
       scope.mouseActive = true;
@@ -378,6 +378,8 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
       $timeout(function() {
         scope.mouseActive = false;
       }, 100);
+
+     if (ev.currentTarget == wrapper) ev.stopPropagation();
     }
 
     function focusListener() {
--- a/src/components/switch/switch.js
+++ b/src/components/switch/switch.js
@@ -50,6 +50,7 @@ angular.module('material.components.switch', [
  */
 function MdSwitch(mdCheckboxDirective, $mdUtil, $mdConstant, $parse, $$rAF, $mdGesture) {
   var checkboxDirective = mdCheckboxDirective[0];
+  var notDisabled = function() { return false; };
 
   return {
     restrict: 'E',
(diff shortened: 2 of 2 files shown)


ThomasBurleson:
@DevVersion - please review

DevVersion:
@ThomasBurleson I understand what your point is.
If we disable the bubbling of click events for `slider` and `switch`, then we should disable bubbling for all other components as well, because they can be also placed inside a list.

In my opinion, we should not disable event bubbling and capturing on all components. Bubbling for controls should be just disabled for cases, where the component is placed inside of a list-item.

A elegant solution is, checking the bubbling / capturing path for a component and cancel the event is the path contains a blocked component.

This means, that we need to define a list of controls, which are not able to be bubbled / captured inside a list.


```js
        function isEventFromControl(e) {
          var maxPath = e.path.indexOf($element.children()[0]);

          for (var i = 0; i < maxPath; i++) {
            if (e.path[i].tagName === 'MD-SLIDER') return true;
          }
        }
```

I've tested that locally and it works like a charm.

ThomasBurleson:
@DevVersion - interesting... Not sure I understand your `isEventFromControl(e)` can you elaborate ?

DevVersion:
@ThomasBurleson Sure :) 

`isEventFromControl(e)` is checking the current event and iterates though the bubbling / capturing path.

```
               / \
---------------| |-----------------
|     WINDOW   | |                |
|   -----------| |-----------     |
| md-list-item | |                |
|   -----------| |-----------     |
|   md-slider  | |                |
|   -----------| |-----------     |
|   _md-thumb  | |                |
|   -----------| |-----------     |
|        Event BUBBLING           |
-----------------------------------
```

I will provide an example, because then it's easier to explain it.
For example, if you click on the `md-slider`, then you mostly hit any child of the slider, like the `_md-thumb`. 

In this case, we're not able to associate it with the `md-slider` control at first glance.
That's why we take advantage of the events path. With contains the bubbling order.

From the clicked element, up to the `window` element. (All elements which will retrieve the bubbled event as well - see the little sketch).

So now we're iterating from the lowest element (`_md-thumb`) up to the `list-item` (not `window` - because that's unnecessary).

So when any item of the iteration is a control, like `md-slider`, then we can conclude that this shouldn't trigger proxied element.

If there any further questions, please let me know on Slack :)

ThomasBurleson:
@DevVersion - close this implementation. Please submit a new PR with the `isEventFromControl(e)` ideas.

PR #7967, opened 2016-04-10
feat(md-ink-explode): New directive for ink circle outside the element

NickBolles:
Ink explode directive that I was using in a project and thought it might go well with the current md-ink directive. 

Example with ng-animate-ref for some material transitions.
http://nickbolles.github.io/Javascript-Sandbox/Angular%20App/dist/index.html#/materialtransitions

### DIFF
--- a/src/core/services/ripple/explode_ripple.js
+++ b/src/core/services/ripple/explode_ripple.js
@@ -12,7 +12,7 @@ angular.module('material.core')
  * element that this directive is on is clicked. This can be used for transitions to another state.
  *
  * @param {string} md-ink-explode A DOM selector for the container to be passed to QuerySelector()
- * @param {number} md-ink-explode-dur The duration of the explode from start to finish
+ * @param {number} md-ink-explode-duration The duration of the explode from start to finish
  * @param {string} md-ink-explode-color A hex, rgb or rgba color string
  *
  * @usage
@@ -45,20 +45,18 @@ angular.module('material.core')
  *     Ripples if 'canRipple' function return value is not 'false' or '0'
  *   </ANY>
  * </hljs>
- *  Note: The container and duration will not observe changes to the Interpolated values via
- *  `md-ink-explode={{ContainerSelector}}` The initial values will be the final values.
  */
-function InkExplodeRippleDirective($mdInkRipple) {
-    return{
-        controller: angular.noop,
-        link: function(scope,element,attr){
-            $mdInkRipple.attach(scope, angular.element(element), {
-                center: false,
-                dimBackground: false,
-                fitRipple: false,
-                duration: parseInt(attr.mdInkExplodeDur) || 1000,
-                inkExplode: true
-            });
-        }
+function InkExplodeRippleDirective ($mdInkRipple) {
+  return {
+    controller: angular.noop,
+    link:       function (scope, element, attr) {
+      $mdInkRipple.attach(scope, angular.element(element), {
+        center:        false,
+        dimBackground: false,
+        fitRipple:     false,
+        duration:      parseInt(attr.mdInkExplodeDuration) || 1000,
+        inkExplode:    true
+      });
     }
+  }
 }
--- a/src/core/services/ripple/explode_ripple.js
+++ b/src/core/services/ripple/explode_ripple.js
@@ -1,0 +1,64 @@
+angular.module('material.core')
+    .directive('mdInkExplode', InkExplodeRippleDirective);
+
+
+/**
+ * @ngdoc directive
+ * @name mdInkExplode
+ * @module material.core
+ *
+ * @description
+ * The `md-ink-explode` directive allows you to specify a different container for the ripple to take place in when the
+ * element that this directive is on is clicked. This can be used for transitions to another state.
+ *
+ * @param {string} md-ink-explode A DOM selector for the container to be passed to QuerySelector()
+ * @param {number} md-ink-explode-dur The duration of the explode from start to finish
+ * @param {string} md-ink-explode-color A hex, rgb or rgba color string
+ *
+ * @usage
+ * ### String values
+ * <hljs lang="html">
+ *   <ANY id="rippleContainer">
+ *       <ANY md-ink-ripple="#FF0000" md-ink-explode="#rippleContainer">
+ *         Ripples in red
+ *       </ANY>
+ *   </ANY>
+ *
+ *   <ANY id="rippleContainer">
+ *       <ANY md-ink-ripple='#0000FF' md-ink-explode-color="#FF0000" md-ink-explode="#rippleContainer">
+ *         Both types of ripple
+ *       </ANY>
+ *   </ANY>
+ *
+ *   <ANY md-ink-explode="false">
+ *     Not rippling
+ *   </ANY>
+ * </hljs>
+ *
+ * ### Interpolated values
+ * <hljs lang="html">
+ *   <ANY md-ink-explode-color="{{ randomColor() }}">
+ *     Ripples with the return value of 'randomColor' function
+ *   </ANY>
+ *
+ *   <ANY md-ink-ripple="{{ canRipple() }}">
+ *     Ripples if 'canRipple' function return value is not 'false' or '0'
+ *   </ANY>
+ * </hljs>
+ *  Note: The container and duration will not observe changes to the Interpolated values via
+ *  `md-ink-explode={{ContainerSelector}}` The initial values will be the final values.
+ */
+function InkExplodeRippleDirective($mdInkRipple) {
+    return{
+        controller: angular.noop,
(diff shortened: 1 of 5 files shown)


src/core/services/ripple/explode_ripple.js
```diff
@@ -0,0 +1,64 @@
...
+ * @ngdoc directive
+ * @name mdInkExplode
+ * @module material.core
+ *
+ * @description
+ * The `md-ink-explode` directive allows you to specify a different container for the ripple to take place in when the
+ * element that this directive is on is clicked. This can be used for transitions to another state.
+ *
+ * @param {string} md-ink-explode A DOM selector for the container to be passed to QuerySelector()
+ * @param {number} md-ink-explode-dur The duration of the explode from start to finish
```
DevVersion: When we provide this feature, then this should be IMO, `md-ink-explode-duration`.

src/core/services/ripple/ripple.js
```diff
@@ -117,25 +117,34 @@ function InkRippleService ($injector) {
...
+  this.timeout          = null; // Stores a reference to the most-recent ripple timeout
+  this.lastRipple       = null;
+  this.duration         = null;
+  this.updateDuration(); //initialize duration
+
 
   $mdUtil.valueOnUse(this, 'container', this.createContainer);
 
-  this.$element.addClass('md-ink-ripple');
+  this.$$element.addClass('md-ink-ripple');
```
DevVersion: Double dollar is weird for the element. I would either create a new variable, just for the explode target.

src/core/services/ripple/ripple.js
```diff
@@ -167,6 +176,10 @@ InkRippleCtrl.prototype.color = function (value) {
     self._color = self._parseColor(value);
   }
 
+  if (self.explode){
```
DevVersion: Space brace from if expression.

src/core/services/ripple/ripple.js
```diff
@@ -252,32 +265,57 @@ InkRippleCtrl.prototype.bindEvents = function () {
...
+ * Either remove or unlock any remaining ripples when the user mouses off of the element (either by
+ * mouseup, touchend or mouseleave event)
+ */
+InkRippleCtrl.prototype.handleMouseup = function (event) {
+  //If this is an explode ripple and it is not a right click create the ripple
+  if (this.explode && event.which !== 3){
+    this.mousedown = false;
+    // When jQuery is loaded, we have to get the original event
+    createRippleFromEvent(this,event);
+  }else{
```
DevVersion: Space braces.

src/core/services/ripple/ripple.js
```diff
@@ -252,32 +265,57 @@ InkRippleCtrl.prototype.bindEvents = function () {
...
+};
+
+/**
+ * If this instance is an explode create the ripple, otherwise,
+ * Either remove or unlock any remaining ripples when the user mouses off of the element (either by
+ * mouseup, touchend or mouseleave event)
+ */
+InkRippleCtrl.prototype.handleMouseup = function (event) {
+  //If this is an explode ripple and it is not a right click create the ripple
+  if (this.explode && event.which !== 3){
```
DevVersion: Space brace here as well. And about the `3` here, IMO we should provide this in the `$mdConstant` service. 

https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent/which

src/core/services/ripple/ripple.js
```diff
@@ -307,17 +345,71 @@ InkRippleCtrl.prototype.clearRipples = function () {
...
   }
 };
-
+/**
+ * Check the md-ink-explode-dur attribute for modification and update the duration
+ * @returns {*}
+ */
+InkRippleCtrl.prototype.updateDuration = function() {
+  this.duration = parseInt(this.$$element.attr('md-ink-explode-dur')) || DURATION;
+  this.fadeInDuration = (this.explode) ? this.duration : this.duration*.35;
```
DevVersion: Arithmetic operator should be spaces as well.

src/core/services/ripple/ripple.js
```diff
@@ -307,17 +345,71 @@ InkRippleCtrl.prototype.clearRipples = function () {
...
+
+/**
+ * Look for an existing container in the $element
+ * @returns {*}
+ */
+InkRippleCtrl.prototype.getContainer = function(){
+  //Check the attribute for modification
+  this.updateContainer();
+  var container;
+  if (!this.$element){return;}
```
DevVersion: This is a really weird line.
```js
if (!this.$element) return;
```

src/core/services/ripple/ripple.js
```diff
@@ -307,17 +345,71 @@ InkRippleCtrl.prototype.clearRipples = function () {
...
+  return this.duration;
+};
 /**
  * Creates the ripple container element
  * @returns {*}
  */
 InkRippleCtrl.prototype.createContainer = function () {
-  var container = angular.element('<div class="md-ripple-container"></div>');
-  this.$element.append(container);
+  var container = this.getContainer();
```
DevVersion: Since when, do we need to check for an existing container? Please explain.

src/core/services/ripple/ripple.js
```diff
@@ -326,13 +418,14 @@ InkRippleCtrl.prototype.clearTimeout = function () {
 };
 
 InkRippleCtrl.prototype.isRippleAllowed = function () {
-  var element = this.$element[0];
+  var element = this.$$element[0];
```
DevVersion: Just need to repeat this again. This is **very** confusing. `$$element` and `$element` are mostly the same name.

src/core/services/ripple/ripple.js
```diff
@@ -363,6 +463,7 @@ InkRippleCtrl.prototype.createRipple = function (left, top) {
   var y           = Math.max(Math.abs(height - top), top) * 2;
   var size        = getSize(this.options.fitRipple, x, y);
   var color       = this.calculateColor();
+  this.explode    ? ripple.addClass('md-ripple-explode'):null;
```
DevVersion: Better use:
```js
this.explode && ripple.addClass('md-ripple-explode');
```

src/core/services/ripple/ripple.js
```diff
@@ -449,7 +552,9 @@ InkRippleCtrl.prototype.removeRipple = function (ripple) {
  */
 InkRippleCtrl.prototype.fadeOutComplete = function (ripple) {
   ripple.remove();
-  this.lastRipple = null;
+  if (this.lastRipple === ripple){
```
DevVersion: Nitpicky again: Please space the brace here, just to be consistent with the Code Conventions.

src/core/style/structure.scss
```diff
@@ -136,6 +136,20 @@ input {
     opacity: 0.20;
   }
 }
+.md-ripple-explode{
```
DevVersion: Space brace here.

DevVersion:
@NickBolles Thanks for the PR and your effort. But in cases, where PR's are that big, you should file an issue first and discuss with the Team about the relevance.

@ThomasBurleson Do we wan't to support this? I like the feature and it's also used in Pages like Google Design etc.

src/core/services/ripple/ripple.js
NickBolles: Shoot, My code style settings in IntelliJ got messed up, ill fix these up.

NickBolles:
@DevVersion As for the issue before the PR, I was going to use this in my own project. So I was doing the work anyways. This is just an idea and starting point. Feel free to close this if its not relevant, or suggest any different methods (Like you have in some ways already).

DevVersion:
@NickBolles Cool to hear that. If you are interessted I will work with you (commenting / reviewing) on this PR. But I can't promise yet, that this PR will be accepted, but I will confirm with the Team.

src/core/services/ripple/ripple.js
NickBolles: Nice! That is way cleaner!

src/core/services/ripple/ripple.js
NickBolles: The reasoning behind this is that one element with many different `md-ink-explode` references to it can have many `<div class="md-ripple-container"></div>` elements. Which I guess is fine, just seems unnecessary. The other reason is to observe changes in the container attribute. 
So here (with the old createContainer function) there are three containers, which correspond to each of the three list items. 
multiple containers

This works fine either way, it might be less expensive to keep the old function then to look for an existing container.

NickBolles:
@DevVersion Cool. There are some changes.
It is also failing the one test. I am not understanding why the test doesn't see the ripple added to the new container. 

Let me know what your thoughts are.:)

ThomasBurleson:
Will not merge. Not part of core functionality.

PR #7989, opened 2016-04-11
fix(select): Fail validation if selection is not a valid option.

### LINKED ISSUE #7954: Required validation is not working for md-select dropdown if the ng-model value is not null
In my case I am using integer type for ng-model value which by default contains value as zero.
The ng-options doesn't contain any item with the value zero. But still validation is not firing.

If I set model value as null then it is working.

please check the below link.

https://plnkr.co/edit/X8zJ56?p=preview

topherfangio:
The md-select currently passes validation even if the model is
not one of the provided options.

Fixes #7954.


topherfangio:
@ThomasBurleson Just a quick ping to see if this can be merged soon 😄

PR #7991, opened 2016-04-11
feat(textarea): textarea shrinking and resizing

### LINKED ISSUE #5919: textarea with min_rows grows slightly too small, scrolls when it shouldn't
I have a texarea control inside an md-input-container that wants the autogrow behavior, but if the text gets larger than a few rows, the textarea is slightly smaller.

More specifically: we have a `<textarea rows="5">`. If I start entering text into it, with 1-4 lines of text, all the text is visible and there's empty space at the bottom. With 5 lines of text, the textarea is full and all 5 lines of text are visible. But when I add a 6th line of text, the textarea remains having rows=5 and so the top line is truncated. As I then add more rows of text, the textarea grows again adding 1 row of actual size for each line of text I add, so it always remains 1 row too small (i.e. I can see all the text except the first line), until I hit 16 lines of text, at which point the textarea again fails to grow and now the top 2 lines are missing (scrolled out of view).

My codebase is still on AM 0.11 but looking at the code on master (currently: https://github.com/angular/material/blob/7fe6f87bb3ecd1d3bc913c19b35eba3273a2ea03/src/components/input/input.js#L320) it still looks the same.

What's going on is the path where the textarea supplies an initial `rows` attribute calculates `lineHeight` by looking at the initial `clientHeight` of the element, and thereafter it uses that `lineHeight` as a divisor against the `node.scrollHeight` to set the `rows` attribute.

In my case (my box, browser, styles, etc), `lineHeight` is 29 but each subsequent line of text I add to the textarea only adds 26 to its `scrollHeight`. So, with some integer-floor stair-steppiness, dividing by 29 instead of 26 means we end up truncating by 1 and then 2 and then more rows as the content grows.

I'm not a CSS/DOM expert but it seems to me the initial clientHeight is interpreted as the height of just a row of text, but also includes some padding (3px in my case) that is not repeated on additional rows. I don't know the right way to calculate the height of this padding. Since the lineHeight is calculated only once when the textarea is ostensibly empty: instead of taking the element height of an empty textarea as the `lineHeight`, perhaps we should calculate `lineHeight` as the difference in heights between a textarea containing no text and one line of text, that is, look at node.scrollHeight when the textarea is empty and get `oneLineHeight`, then temporarily set the textarea value to `'\n'` and look again at node.scrollHeight and get `twoLineHeight`, then set `lineHeight = twoLineHeight - oneLineHeight`?

Note that this behavior only affects the textarea if you supply an initial rows=, and not the default auto-sizing path.

### LINKED ISSUE #7649: Textarea improvements and additions
TL;DR from the [Google Doc](https://docs.google.com/document/d/1NDtBURsRIkswgrjdDEcB1wzt-asBLnezXiZ0M_ryYPs/edit?usp=sharing):

We should look into some improvements to the `textarea` directive:

* We should support growing and shrinking according to the content. The current implementation doesn't shrink down if the `rows` are specified.
* We should support a minimum amount of rows. The current solution doesn't respect the `rows` unless there's enough content to force the size beyond the amount of rows.
* We should have the ability to resize the textarea.

CC @EladBezalel, @DevVersion, @ThomasBurleson

### LINKED ISSUE #8135: <md-input-container> Text area does not shrink when all text is cleared
See [codepen](http://codepen.io/cmacdonnacha/pen/bpKvJe) example.

Here's the initial **md-input-container** component with a text area:

          <label>Enter some text and hit enter a few times. Then delete all the text.</label>
          <textarea rows="5"></textarea>
      </md-input-container>

image

I then enter some text and hit enter a few times:

image

I've set **rows="5"** on the textarea so it only grows to show those 5 rows.

However, when I clear all the text it doesn't shrink again:
image

crisbeto:
* Changes the textarea behavior to consider the "rows" attribute as the minimum, instead of the maximum, when auto-expanding the element.
* Adds a handle for vertically resizing the textarea element.
* Simplifies the logic for determining the textarea height.
* Makes the textarea sizing more accurate by using scrollHeight directly, instead of depending on the line height and amount of rows.
* Avoids potential issues where the textarea wouldn't resize when adding a newline.
* Adds the option to specify a maximum number of rows for a textarea via the `max-rows` attribute.

Fixes #7649. 
Fixes #5919. 
Fixes #8135.

BREAKING CHANGE
This changes the behavior from considering the "rows" attribute as the maximum to considering it as the minimum.

@ThomasBurleson @EladBezalel @DevVersion These are the textarea changes that we talked about and specced in the Google Doc.

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -174,6 +174,8 @@ function labelDirective() {
  *   PRESENT. The placeholder text is copied to the aria-label attribute.
  * @param md-no-autogrow {boolean=} When present, textareas will not grow automatically.
  * @param md-no-asterisk {boolean=} When present, an asterisk will not be appended to the inputs floating label
+ * @param md-no-resize {boolean=} Disables the textarea resize handle.
+ * @param {number=} max-rows The maximum amount of rows for a textarea.
  * @param md-detect-hidden {boolean=} When present, textareas will be sized properly when they are
  *   revealed after being hidden. This is off by default for performance reasons because it
  *   guarantees a reflow every digest cycle.
@@ -259,9 +261,21 @@ function labelDirective() {
  * error animation effects. Therefore, it is *not* advised to use the Layout system inside of the
  * `<md-input-container>` tags. Instead, use relative or absolute positioning.
  *
+ *
+ * <h3>Textarea directive</h3>
+ * The `textarea` element within a `md-input-container` has the following specific behavior:
+ * - By default the `textarea` grows as the user types. This can be disabled via the `md-no-autogrow`
+ * attribute.
+ * - If a `textarea` has the `rows` attribute, it will treat the `rows` as the minimum height and will
+ * continue growing as the user types. For example a textarea with `rows="3"` will be 3 lines of text
+ * high initially. If no rows are specified, the directive defaults to 1.
+ * - If you wan't a `textarea` to stop growing at a certain point, you can specify the `max-rows` attribute.
+ * - The textarea's bottom border acts as a handle which users can drag, in order to resize the element vertically.
+ * Once the user has resized a `textarea`, the autogrowing functionality becomes disabled. If you don't want a
+ * `textarea` to be resizeable by the user, you can add the `md-no-resize` attribute.
  */
 
-function inputTextareaDirective($mdUtil, $window, $mdAria, $timeout) {
+function inputTextareaDirective($mdUtil, $window, $mdAria, $timeout, $mdGesture) {
   return {
     restrict: 'E',
     require: ['^?mdInputContainer', '?ngModel'],
@@ -379,13 +393,16 @@ function inputTextareaDirective($mdUtil, $window, $mdAria, $timeout) {
     }
 
     function setupTextarea() {
-      if (attr.hasOwnProperty('mdNoAutogrow')) {
-        return;
-      }
+      var isAutogrowing = !attr.hasOwnProperty('mdNoAutogrow');
+
+      attachResizeHandle();
+
+      if (!isAutogrowing) return;
 
       // Can't check if height was or not explicity set,
       // so rows attribute will take precedence if present
       var minRows = attr.hasOwnProperty('rows') ? parseInt(attr.rows) : NaN;
+      var maxRows = attr.hasOwnProperty('maxRows') ? parseInt(attr.maxRows) : NaN;
       var lineHeight = null;
       var node = element[0];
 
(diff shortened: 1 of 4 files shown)


EladBezalel:
@crisbeto is it auto growing/shrinkning according the content?

crisbeto:
It is.

EladBezalel:
@topherfangio please take a look as well

topherfangio:
@crisbeto This should be noted as a breaking change in the commit message (and it should have a label).

Just to confirm, this means that there is no longer a maximum size for the text area? Will this break any apps that depend on it having a maximum size?

src/components/input/input.js
```diff
@@ -391,78 +394,130 @@ function inputTextareaDirective($mdUtil, $window, $mdAria, $timeout) {
...
+        element.after(handle);
+        handle.on('mousedown', onMouseDown);
+        $mdGesture.register(handle, 'drag', { horizontal: false });
+
+        container
+          .on('$md.dragstart', onDragStart)
+          .on('$md.drag', onDrag)
+          .on('$md.dragend', onDragEnd);
+
+        scope.$on('$destroy', function() {
```
topherfangio: Haven't used it much; do we need to deregister the drag gesture above (i.e. undo the `$mdGesture.register`?

src/components/input/input.spec.js
```diff
@@ -515,33 +530,60 @@ describe('md-input-container directive', function() {
...
       scope.$apply();
 
-      $timeout.flush();
       var newHeight = textarea.offsetHeight;
       expect(textarea.offsetHeight).toBeGreaterThan(oldHeight);
     });
 
-    it('should make the textarea scrollable once it has reached the row limit', function() {
-      var scrollableClass = '_md-textarea-scrollable';
+    it('should set the rows attribute as the user types', function() {
```
topherfangio: Really like the simplicity of these tests :+1:

src/components/input/input.js
crisbeto: I'd say so. I remember trying to find if there was a `deregister` method, but I couldn't find one.

crisbeto:
Yes @topherfangio, this is a breaking change. Mainly because the old behavior was: start at 1 row or fit any of the text and go until the `rows` are reached. The new behaviour treats the `rows` as a minimum.

topherfangio:
@crisbeto Okay, awesome. Can you make sure to do a `git commit --amend` to add the BREAKING CHANGE notice to the commit message? It's important because we automatically generate a list of breaking changes when we build a new release, and this is what that script looks for :-)

crisbeto:
Done, hopefully it's not too picky with the placement.

src/components/input/input.js
DevVersion: The `deregister` function will be returned from the `register` call.

- https://github.com/angular/material/blob/master/src/core/services/gesture/gesture.js#L401-L414

topherfangio:
I just tested this locally and it looks pretty good, but I think there might be a small issue with the height calculation.

Particularly, when you have 5 rows of text (some might be newlines), there is about `20px` of extra whitespace at the bottom. As soon as you start typing the 6th row, it does indeed grow, but only by a tiny amount so that there's only `8px` of whitespace at the bottom. This was on both Chrome and Firefox, so I don't think it's browser-specific.

**5 Rows**
text-5-rows

**6 Rows**
text-6-rows

Also, is it true that once a user resizes the textarea, it no longer does any automatic resizing? If so, I think that's fine, but we may want to mention this in the docs. In fact, we may want to just add a Textarea Usage section to the docs with some additional info about how users should expect it to work and what options are available.

crisbeto:
Regarding the height, I believe it's because 5 rows in the textarea don't necessarily translate to 5 lines of text. We use `offsetHeight` to measure the height of 1 row of text, but that measurement includes padding, which means that the more padding we have, the more the measurement will be offset. E.g. here's an example of 5 rows, but with 20px of padding:

angular_material_-_demos__input_-_google_chrome_2016-04-14_18-28-11

I wouldn't say it's too much of an issue in our case, because the design doesn't have a lot of padding by default. I've tried using `getComputedStyle` to remove the padding, but that was also problematic because:

1. It didn't behave consistently between IE and other browsers.
2. We'd have to either call it on each keypress or assume that it would't change.

Otherwise I've also tried resizing the textarea purely with the `rows` attribute, but it wasn't reliable on IE either. And the other alternative is having a whole separate DOM element that mirrors the text and is being used for measurements, but that can turn into a big mess.

---
As for the resizing, I'll add some info in the docs.

crisbeto:
I managed to work around those height issues. Also updated the PR to include a `max-rows` attribute.

topherfangio:
LGTM 👍

cmacdonnacha:
Just wondering if this will be merged for the next release?

PR #8001, opened 2016-04-12
feat(tabs): add rtl support for md-tabs

### LINKED ISSUE #1078: md-tabs rtl support
When i add md-tabs on rtl page the tabs ink bar always underlines the opposite side

### LINKED ISSUE #3538: No rtl support for mdTabs
Hi guys
I've noticed that there's no rtl support for mdTabs

As you can see in [this codepen](http://codepen.io/dani3l/pen/PqQpoj) the tab-items are aligned to the left. This was [easily fixed with some css](http://codepen.io/dani3l/pen/eNVgab)

Another issue is when you have alot of tabs and [the pagination don't know how to handle rtl](http://codepen.io/dani3l/pen/eNVvOx)

And last issue is that the tab-content transition should be reversed when page is rtl. Meaning that when user clicks tab "TWO" the content should come from the left-to-right and when going back to tab "ONE"  content should come from right-to-left

I found that pull request: https://github.com/angular/material/pull/1711 but I guess its related to a really early version so I don't know if it can help.

### LINKED ISSUE #4768: md-tabs transition - when direction is switched from ltr to rtl, tab transition should switch
Hello,

This might be minor on your list, but it would be great if the tabs slide transitions switched with the direction. 

For example, when the document direction is from left-to-right, the tab transition is slide-left when the tabs-index goes from 0 to 1+ and slide right when it goes from index 1+ to 0. (this is correct)
Now when the document direction is from right-to-left, the tab transition should be slide-right when the tabs-index goes from 0 to 1+ and slide left when it goes from 1+ to 0. 

Thank you !

HosseinMoradiDavijani:
add rtl support for md-tabs using "md-rtl" attribute

```html
<md-tabs md-rtl>

</md-tabs>
```

Closes #1078

Closes #3538

Closes #4768

### DIFF
--- a/src/components/tabs/tabs.scss
+++ b/src/components/tabs/tabs.scss
@@ -286,3 +286,30 @@ md-toolbar + md-tabs {
   border-top-left-radius: 0;
   border-top-right-radius: 0;
 }
+
+md-tabs[md-rtl] {
+  .md-tab {
+    float: right;
+  }
+  md-tab-content {
+    &.md-left:not(.md-active) {
+      -webkit-transform: translateX(100%);
+      transform: translateX(100%);
+    }
+    &.md-right:not(.md-active) {
+      -webkit-transform: translateX(-100%);
+      transform: translateX(-100%);
+    }
+  }
+  md-tabs-canvas {
+    .md-dummy-wrapper {
+      left: auto;
+      right: 0;
+    }
+
+    md-pagination-wrapper {
+      left: auto;
+      right: 0;
+    }
+  }
+}


src/components/tabs/tabs.scss
```diff
@@ -286,3 +286,30 @@ md-toolbar + md-tabs {
   border-top-left-radius: 0;
   border-top-right-radius: 0;
 }
+
+md-tabs[md-rtl] {
```
DevVersion: I don't think we should introduce a new attribute just for RTL support.

We normally apply stylings dynamically, by detecting the current language direction.

Check the `rtl-prop`, `rtl-value` mixin.

EladBezalel:
Please have a look how we handle RTL, as @DevVeraion said, the tabs should be smart and auto adjust

HosseinMoradiDavijani:
Which of these approaches is better in this case?
```css
html[dir=rtl] md-tabs { ... }
```
or
```css
*[dir=rtl] md-tabs { ... }
```

DevVersion:
I would go with the second variant. But supporting RTL for the tabs should be also possible by using the `rtl` mixins we provide.

- https://github.com/angular/material/blob/master/src/core/style/mixins.scss#L111-L129

HosseinMoradiDavijani:
I checked everything in code and it seems everything is OK. Also, I updated angular-material to latest (1.1.0-rc4) version and everything works fine. So, I think related issues such as #1078 and #3538  should be closed.

Thanks a lot for your great works guys.

DevVersion:
@hosseinkhaledi Thanks for your efforts! Also thanks for alerting us on that.

PR #7056, opened 2016-02-06
md-menu: add handling for 'data-' prefix of HTML attributes 

StritziK:
- md-menu caused error when using 'data-' prefixed ng-click directive.

Fixes: #6659

Version:
```javascript
/*!
 * Angular Material Design
 * https://github.com/angular/material
 * @license MIT
 * v1.0.5
 */
```

PR #8002, opened 2016-04-12
feat(chips): allow mdChips component to be used with 'data-'-prefixes

janeisklar:
Compare to [pull request #7056](https://github.com/angular/material/pull/7056) for a similar issue with md-menu
Changes:
* `<md-chips>`, `<md-chip-template>`, etc. can now also be used as `<data-md-chips>`, `<data-md-chip-template>`
* added a basic usage demo that features those `<data-*>` tags
* made the corresponding tests parametrized so that they are executed both with and without the data-prefix

Background:
'data-'-prefixing is f.e. done by angular-htmlify and allows for the W3C validation of angular templates

### DIFF
--- a/src/components/chips/chips-theme.scss
+++ b/src/components/chips/chips-theme.scss
@@ -1,4 +1,4 @@
-md-chips.md-THEME_NAME-theme {
+md-chips.md-THEME_NAME-theme, data-md-chips.md-THEME_NAME-theme {
 
   .md-chips {
     box-shadow: 0 1px '{{foreground-4}}';
@@ -15,11 +15,11 @@ md-chips.md-THEME_NAME-theme {
     }
   }
 
-  md-chip {
+  md-chip, data-md-chip {
     background: '{{background-300}}';
     color: '{{background-800}}';
 
-    md-icon {
+    md-icon, data-md-icon {
       color: '{{background-700}}';
     }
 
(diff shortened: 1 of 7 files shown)


janeisklar:
I signed it!

src/components/chips/chips-theme.scss
```diff
@@ -1,4 +1,4 @@
-md-chips.md-THEME_NAME-theme {
+md-chips.md-THEME_NAME-theme, data-md-chips.md-THEME_NAME-theme {
```
DevVersion: I don't like that. This would require a selector for `x-` as well.

We should either apply a class, to identify the chips directive.

src/components/chips/demoBasicUsageDataPrefixed/index.html
```diff
@@ -0,0 +1,62 @@
+<div data-ng-controller="BasicDemoCtrl as ctrl" layout="column" data-ng-cloak>
```
DevVersion: An extra demo, just for the `data-` prefix is unnecessary. Please remove that.

ThomasBurleson:
A generic solution to `data-` prefixing is in progress. Will not merge this PR.

janeisklar:
Is there an issue that you could reference allowing us to track the progress of the mentioned generic solution?

DevVersion:
@janeisklar Yes it is.
- Issue: #8121
- PR: #8163

PR #8013, opened 2016-04-12
update(panel): Support multiple panels and clean up some code.

KarenParker:
Fixes the problem of not resetting the defaults when there are multiple panels.

@ErinCoughlan @jelbourn @gmoothart @DerekLouie @ThomasBurleson  Please review.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -598,26 +598,34 @@ var FOCUS_TRAP_TEMPLATE = angular.element(
 function MdPanelService($rootElement, $rootScope, $injector) {
   /**
    * Default config options for the panel.
+   * Anything angular related needs to be done later. Therefore
+   *     scope: $rootScope.$new(true),
+   *     attachTo: $rootElement,
+   * are added later.
    * @private {!Object}
    */
   this._defaultConfigOptions = {
-    attachTo: $rootElement,
     bindToController: true,
     clickOutsideToClose: false,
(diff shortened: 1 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -649,6 +652,8 @@ function MdPanelService($rootElement, $rootScope, $injector) {
 MdPanelService.prototype.create = function(opt_config) {
   var configSettings = opt_config || {};
 
+  angular.copy(this._defaultConfigOptions, this._config);
```
ErinCoughlan: nit: What about using extend for both? As written, it harder to remember that one of these copies left to right and the other right to left.

KarenParker: Extend won't copy the values, just point to the other object's values. This is what I am trying to avoid.

ErinCoughlan:
test to make sure multiple panels is possible?

src/components/panel/panel.js
jelbourn: Do you need a full deep-copy? The performance of `angular.copy` is not the best.

KarenParker:
Test added.

src/components/panel/panel.js
KarenParker: The deep-copy isn't needed anymore. Code has been updated.

ErinCoughlan:
LGTM

jelbourn:
LGTM

Issue #8011, opened 2016-04-12
mdTooltip: $digest already in progress when there are a number of disabled elements


Frank3K:
In Firefox 45.0.1 on Linux I encounter the following issue when loading a page with a number of disabled `md-button` elements, each with its own `md-tooltip`.

```
Error: [$rootScope:inprog] $digest already in progress
http://errors.angularjs.org/1.5.3/$rootScope/inprog?p0=%24digest
```

I have tracked down the issue to the following [line 157 in tooltip.js](https://github.com/angular/material/blob/e7f866d9e5f0239fe1b3b93f598bc1185966823d/src/components/tooltip/tooltip.js#L157):

``` js
      // add an mutationObserver when there is support for it
      // and the need for it in the form of viable host(parent[0])
      if (parent[0] && 'MutationObserver' in $window) {
        // use an mutationObserver to tackle #2602
        var attributeObserver = new MutationObserver(function(mutations) {
          mutations.forEach(function (mutation) {
            if (mutation.attributeName === 'disabled' && parent[0].disabled) {
              setVisible(false);
              scope.$digest(); // make sure the elements gets updated
            }
          });
        });

        attributeObserver.observe(parent[0], { attributes: true});
      }
```

When there are mutations for multiple elements (I enable most elements on a certain loaded event), a number of digest are executed if I read the code correctly. This is what (I think) triggers the error.

I'm no expert in `$digest()`, so I do not know what the best solution is. My first approach was to move the `scope.$digest()` outside the `forEach` but that did not fix the issue. Note that commenting the line altogether does remove the error, so I'm quite sure this is the offending call.

clshortfuse:
I believe we can change scope.$digest() to angular.noop(). Do you have a codepen we can test with?

Frank3K:
I don't have a codepen at the moment since the scenario seems quite complex (i.e. using a number of disabled elements that are enabled at some point in time). I will test your PR on my current project tonight.

Frank3K:
Unfortunately the current fix from PR #8015 does not work. I'll try to construct a minimal example in a codepen.

Frank3K:
I havent been able to reproduce the scenario in a codepen as of yet. Is it someway possible to get more information about the digest that is already in progress?

When looking at stack traces the conflicting digest is always the one I have pointed out in the first comment.

Frank3K:
[...]

PR #8015, opened 2016-04-12
fix(md-tooltip) prevent multiple digests when checking for disabled parent

clshortfuse:
fixes https://github.com/angular/material/issues/8011


Frank3K:
Confirmed by manual test in https://github.com/angular/material/issues/8011#issuecomment-210057786.

PR #8022, opened 2016-04-13
fix(theming): theming should probably parse background hue names.

### LINKED ISSUE #7510: 1.1.0-rc1 seems to break .dark themes
You can see it in the first "Basic Usage" input demo
https://material.angularjs.org/latest/demo/input
The top is supposed to be grey and the 2 first inputs cant be seen because they are white.

DevVersion:
* The background palettes are using `default` as a hue value, which is not covered by the prefixed with `hue-` check.

@EladBezalel Please review.

Fixes #7510.


ThomasBurleson:
@EladBezalel  - Plz review.

EladBezalel:
LGTM

PR #8024, opened 2016-04-13
feat(panel): Add onDomAdded, onOpenComplete, and onRemoving animation hooks.

### LINKED ISSUE #7875: mdPanel: Add onDomAdded, onOpenComplete, and onRemoving animation hooks.
These will be similar to the parameters used in mdDialog.

onDomAdded should trigger after attach is done
onOpenComplete should trigger after show is done
onRemoving should trigger after hide is done

This task include updating documentation.

DerekLouie:
Fixes #7875.

@gmoothart @ErinCoughlan @KarenParker @ThomasBurleson @jelbourn Please review.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -131,6 +131,14 @@ angular
  *     behind the panel. Defaults to false.
  *   - `disableParentScroll` - `{boolean=}`: Whether the user can scroll the
  *     page behind the panel. Defaults to false.
+ *   - `onDomAdded` - `{function=}`: Callback function used to announce when
+ *     the panel is added to the DOM.
+ *   - `onOpenComplete` - `{function=}`: Callback function used to announce
+ *     when the open() action is finished.
+ *   - `onRemoving` - `{function=}`: Callback function used to announce the
+ *     close/hide() action is starting.
+ *   - `onDomRemoved` - `{function=}`: Callback function used to announce when the
+ *     panel is removed from the DOM.
  *
  * TODO(ErinCoughlan): Add the following config options.
  *   - `groupName` - `{string=}`: Name of panel groups. This group name is
@@ -212,15 +220,6 @@ angular
  *     create.
  *   - `isAttached` - `{boolean}`: Whether the panel is attached to the DOM.
  *     Visibility to the user does not factor into isAttached.
- *
- * TODO(ErinCoughlan): Add the following properties.
- *   - `onDomAdded` - `{function=}`: Callback function used to announce when
- *     the panel is added to the DOM.
- *   - `onOpenComplete` - `{function=}`: Callback function used to announce
- *     when the open() action is finished.
- *   - `onRemoving` - `{function=}`: Callback function used to announce the
- *     close/hide() action is starting. This allows developers to run custom
- *     animations in parallel the close animations.
  */
 
 /**
@@ -821,8 +820,9 @@ MdPanelRef.prototype.open = function() {
     var show = self._simpleBind(self.show, self);
 
     self.attach()
-        .then(show, reject)
-        .then(done, reject);
+        .then(show)
+        .then(done)
+        .catch(reject);
   });
 };
 
@@ -835,13 +835,15 @@ MdPanelRef.prototype.open = function() {
  */
 MdPanelRef.prototype.close = function() {
   var self = this;
+
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
     var detach = self._simpleBind(self.detach, self);
 
     self.hide()
-        .then(detach, reject)
-        .then(done, reject);
+        .then(detach)
+        .then(done)
+        .catch(reject);
   });
 };
 
@@ -860,14 +862,21 @@ MdPanelRef.prototype.attach = function() {
   var self = this;
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
-
-    self._$q.all([
-      self._createBackdrop(),
-      self._createPanel().then(function() {
+    var onDomAdded = self._config['onDomAdded'] || angular.noop;
+    var addListeners = function(response) {
         self.isAttached = true;
         self._addEventListeners();
-      }, reject)
-    ]).then(done, reject);
+        return response;
+    };
+
+    self._$q.all([
+        self._createBackdrop(),
+        self._createPanel()
+            .then(addListeners)
+            .catch(reject)
+    ]).then(onDomAdded)
+      .then(done)
+      .catch(reject);
   });
 };
 
@@ -884,6 +893,8 @@ MdPanelRef.prototype.detach = function() {
   }
 
   var self = this;
+  var onDomRemoved = self._config['onDomRemoved'] || angular.noop;
+
   var detachFn = function() {
     self._removeEventListener();
 
@@ -913,7 +924,9 @@ MdPanelRef.prototype.detach = function() {
     self._$q.all([
       detachFn(),
       self._backdropRef ? self._backdropRef.detach() : true
-    ]).then(done, reject);
+    ]).then(onDomRemoved)
+      .then(done)
+      .catch(reject);
   });
 };
 
@@ -943,11 +956,14 @@ MdPanelRef.prototype.show = function() {
 
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
+    var onOpenComplete = self._config['onOpenComplete'] || angular.noop;
 
     self._$q.all([
       self._backdropRef ? self._backdropRef.show() : self,
       animatePromise().then(function() { self._focusOnOpen(); }, reject)
-    ]).then(done, reject);
+    ]).then(onOpenComplete)
+      .then(done)
+      .catch(reject);
   });
 };
 
@@ -972,11 +988,15 @@ MdPanelRef.prototype.hide = function() {
   var self = this;
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
+    var onRemoving = self._config['onRemoving'] || angular.noop;
+
 
     self._$q.all([
       self._backdropRef ? self._backdropRef.hide() : self,
-      self._animateClose().then(function() { self.addClass(MD_PANEL_HIDDEN); },
-          reject)
+      self._animateClose()
+          .then(onRemoving)
+          .then(function() { self.addClass(MD_PANEL_HIDDEN); })
+          .catch(reject)
     ]).then(done, reject);
   });
 };
@@ -1037,6 +1057,7 @@ MdPanelRef.prototype.toggleClass = function(toggleClass) {
  */
 MdPanelRef.prototype._createPanel = function() {
   var self = this;
+
   return this._$q(function(resolve, reject) {
     if (!self._config.locals) {
       self._config.locals = {};
--- a/src/components/panel/panel.spec.js
+++ b/src/components/panel/panel.spec.js
@@ -66,9 +66,9 @@ describe('$mdPanel', function() {
               pass: pass,
               message: 'Expected ' + expected + not + ' to be within ' +
               epsilon + ' of ' + actual
-            }
+            };
           }
-        }
+        };
       }
     });
   });
@@ -432,16 +432,7 @@ describe('$mdPanel', function() {
     it('should not close when clickOutsideToClose set to false', function() {
       openPanel();
 
-      var container = panelRef._panelContainer;
-      container.triggerHandler({
-        type: 'mousedown',
-        target: container[0]
-      });
-      container.triggerHandler({
-        type: 'mouseup',
-        target: container[0]
-      });
-      flushPanel();
+      clickPanelContainer();
 
       expect(PANEL_EL).toExist();
     });
@@ -453,16 +444,7 @@ describe('$mdPanel', function() {
 
       openPanel(config);
 
-      var container = panelRef._panelContainer;
-      container.triggerHandler({
-        type: 'mousedown',
-        target: container[0]
-      });
-      container.triggerHandler({
-        type: 'mouseup',
-        target: container[0]
-      });
-      flushPanel();
+      clickPanelContainer();
 
       // TODO(ErinCoughlan) - Add this when destroy is added.
       // expect(panelRef).toBeUndefined();
@@ -472,12 +454,7 @@ describe('$mdPanel', function() {
     it('should not close when escapeToClose set to false', function() {
       openPanel();
 
-      var container = panelRef._panelContainer;
-      container.triggerHandler({
-        type: 'keydown',
-        keyCode: $mdConstant.KEY_CODE.ESCAPE
-      });
-      flushPanel();
+      pressEscape();
 
       expect(PANEL_EL).toExist();
     });
@@ -489,12 +466,7 @@ describe('$mdPanel', function() {
 
       openPanel(config);
 
-      var container = panelRef._panelContainer;
-      container.triggerHandler({
-        type: 'keydown',
-        keyCode: $mdConstant.KEY_CODE.ESCAPE
-      });
-      flushPanel();
+      pressEscape();
 
       // TODO(ErinCoughlan) - Add this when destroy is added.
       // expect(panelRef).toBeUndefined();
@@ -611,16 +583,7 @@ describe('$mdPanel', function() {
             return panelRef._$q.resolve(self);
           };
 
-          var container = panelRef._panelContainer;
-          container.triggerHandler({
-            type: 'mousedown',
-            target: container[0]
-          });
-          container.triggerHandler({
-            type: 'mouseup',
-            target: container[0]
-          });
-          flushPanel();
+          clickPanelContainer();
 
           expect(closeCalled).toBe(true);
         });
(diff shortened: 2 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -126,6 +126,16 @@ angular
  *   - `animation` - `{MdPanelAnimation=}`: An MdPanelAnimation object that
  *     specifies the animation of the panel. For more information, see
  *     `MdPanelAnimation`.
+ *   - `onDomAdded` - `{function=}`: Callback function used to announce when
+ *     the panel is added to the DOM. This function needs to return a promise
```
ErinCoughlan: We shouldn't require the user to return a promise. We can wrap their function response in one so chaining will still work on our end.

src/components/panel/panel.js
```diff
@@ -126,6 +126,16 @@ angular
...
+ *     for chaining purposes.
+ *   - `onOpenComplete` - `{function=}`: Callback function used to announce
+ *     when the open() action is finished. This function needs to return a promise
+ *     for chaining purposes.
+ *   - `onRemoving` - `{function=}`: Callback function used to announce the
+ *     close/hide() action is starting. This allows developers to run custom
+ *     animations in parallel the close animations. This function needs to return
```
ErinCoughlan: Maybe remove the comment about custom animations. The close animation can be customized already.

src/components/panel/panel.js
```diff
@@ -819,11 +820,16 @@ MdPanelRef.prototype.open = function() {
   // TODO(ErinCoughlan) - Cancel any in-progress actions.
 
   var self = this;
+
+  var onOpenComplete = self._config['onOpenComplete'];
+  var onOpenCompleteScope = self._config['scope'] || self._$rootScope;
```
ErinCoughlan: `self._config['scope']` will always be defined. We default it to a new scope off of `$rootScope`.

src/components/panel/panel.js
```diff
@@ -1272,18 +1284,26 @@ MdPanelRef.prototype._animateOpen = function() {
...
 MdPanelRef.prototype._animateClose = function() {
+  var onRemoving = this._config['onRemoving'];
+  var onRemovingScope = this._config['scope'] || this._$rootScope;
+
   var animationConfig = this._config['animation'];
   if (!animationConfig) {
     this.removeClass('md-panel-is-showing');
     this.removeClass('_md-panel-shown');
+    var onRemoving = onRemovingScope.$eval(onRemoving);
+    onRemoving && onRemoving.resolve && onRemoving.resolve();
```
ErinCoughlan: Why not just return `this._$q(onRemovingScope.$eval(onRemoving)`? That's still a promise, which resolves or rejects to whatever the user specifies.

src/components/panel/panel.spec.js
```diff
@@ -104,6 +104,43 @@ describe('$mdPanel', function() {
...
+  it('should call onDomAdded if provided when adding the panel to the DOM', function() {
+    var onDomAdded = jasmine.createSpy();
+    var animationHookConfig = angular.extend(DEFAULT_CONFIG, {'onDomAdded': onDomAdded});
+
+    panelRef = $mdPanel.create(animationHookConfig);
+
+    panelRef.attachOnly();
+    expect(onDomAdded).toHaveBeenCalled();
+
+    angular.element(document.querySelector(PANEL_WRAPPER_CLASS)).remove();
```
KarenParker: This should be removed in the afterEach already and if not it should be there for all tests.

src/components/panel/panel.spec.js
```diff
@@ -1201,6 +1238,34 @@ describe('$mdPanel', function() {
...
+      var deferred = $q.defer();
+      var resolvedPromise = deferred.promise;
+
+      var onRemovingAnimation = {
+        'animateClose' : function() {
+          return resolvedPromise;
+        }
+      };
+
+      var animationHookConfig = angular.extend(DEFAULT_CONFIG, {'onRemoving': onRemoving});
```
KarenParker: No animation configuration is being added here. You set it below, but animations (close and open) are determined when the panel is created.

DerekLouie: Ahh yes, so the issue here is that the `_animateClose` function doesn't execute [this line](https://github.com/angular/material/blob/feat/panel/src/components/panel/panel.js#L1294) of code unless `animation` is defined on the `config` object. 

If I add this from the beginning, a bunch of `open` code gets called (wherever this `if (!animationConfig)` check gets called) and the animation I have to construct becomes a little complicated. 

If I add the animation to the `config` object after the open, I can still get the code path in the link I sent above to trigger, and I don't need to construct a huge `animation` object.

Thinking more about it though, this is hardly black box testing. So I am okay with constructing a proper `animation` object, I will leave it up to your discretion @KarenParker.

DerekLouie:
Hello All,

I have addressed all the aforementioned comments. 

Please review at your earliest convenience!

src/components/panel/panel.spec.js
KarenParker: It isn't hard to setup an Animation just use the default fade and you don't need to set the openFrom and closeTo values.

src/components/panel/panel.js
```diff
@@ -1282,18 +1288,24 @@ MdPanelRef.prototype._animateOpen = function() {
...
     return this._$q.resolve();
   }
 
   var self = this;
   return this._$q(function(resolve, reject) {
-    animationConfig.animateClose(self._$q).then(function(){
+    animationConfig.animateClose(self._$q).then(function() {
       self.removeClass('md-panel-is-showing');
-      resolve(self);
+      onRemoving = self._$q.when(this._config['scope'].$eval(onRemoving))
```
KarenParker: Why are you setting the promise result back to onRemoving? Also this is the wrong spot to be calling onRemoving. The panel will still be visible here depending on the animation. This should be in the hide section after the HIDDEN class has been added.

src/components/panel/panel.js
```diff
@@ -1087,6 +1090,9 @@ MdPanelRef.prototype._createPanel = function() {
 
           self._addStyles();
           self._configureTrapFocus();
+        }, reject)
+        .then(self._$q.when(self._config['scope'].$eval(onDomAdded)), reject)
```
KarenParker: Optional: Consider moving this to the attach function after _createPanel is called. It will be called in the same place but if someone is looking in the code it would make more sense in attach.

src/components/panel/panel.js
```diff
@@ -828,11 +825,15 @@ MdPanelRef.prototype.open = function() {
...
   var self = this;
+
+  var onOpenComplete = self._config['onOpenComplete'];
   this._openPromise = this._$q(function(resolve, reject) {
     self.attachOnly().then(function() {
-      self.show().then(function() {
-        resolve(self);
-      }, reject);
+      self.show()
+          .then(self._$q.when(self._config['scope'].$eval(onOpenComplete)), reject)
```
KarenParker: Please move this to the end of the show() section. If the user calls attach and show on their own we still want this callback to trigger.

src/components/panel/panel.js
KarenParker: The 'this' would be undefined at this point. Use 'self'. Also please verify (in a test or demo) that the callbacks are in the proper scope (using the caller's 'this' value).

DerekLouie:
Hello All,

I have fixed the aforementioned comments please review again at your earliest convenience!

Best,

Derek

src/components/panel/panel.js
```diff
@@ -865,6 +865,15 @@ MdPanelRef.prototype.close = function() {
         return self.detach();
       }, this._$q.reject)
       .then(function() {
+        var onDomRemoved = self._config['onDomRemoved'];
+
+        if (onDomRemoved) {
+          return self._$q.when(onDomRemoved());
+        } else {
+          return true;
```
ErinCoughlan: Shouldn't you return a promise in both cases so that chaining won't be broken.

I think this should be: `self._$q.resolve(self)`

src/components/panel/panel.js
```diff
@@ -896,6 +905,18 @@ MdPanelRef.prototype.attachOnly = function() {
...
+
+      if (onDomAdded) {
+        return self._$q.when(onDomAdded());
+      } else {
+        return true;
+      }
+    }, self._$q.reject)
+    .then(function() {
+      return self._$q.resolve(self);
+    }, this._$q.reject)
```
ErinCoughlan: Should this be `self._$q.reject`?

src/components/panel/panel.spec.js
```diff
@@ -77,6 +77,9 @@ describe('$mdPanel', function() {
 
     // TODO(ErinCoughlan) - Remove when close destroys.
     panelRef = null;
+
+    // If just attachPanel is called w/o close, this cleans things up.
+    angular.element(document.querySelector(PANEL_WRAPPER_CLASS)).remove();
```
ErinCoughlan: We call close a few lines above this to make sure it's cleaned up. If any element is added to the DOM, it gets added to the attachedElements list, which are all removed here as well.

src/components/panel/panel.spec.js
```diff
@@ -105,6 +108,62 @@ describe('$mdPanel', function() {
...
+
+    panelRef = $mdPanel.create(animationHookConfig);
+
+    panelRef.attachOnly().then(function() {
+      expect(onDomAdded).toHaveBeenCalled();
+      expect(PANEL_EL).toExist();
+    });
+  });
+
+  it('should call onOpenComplete if provided after adding the panel to the DOM and animating', function() {
```
ErinCoughlan: We've been using 80 char line limits in this file, please wrap these where necessary.

src/components/panel/panel.spec.js
```diff
@@ -105,6 +108,62 @@ describe('$mdPanel', function() {
     expect(PANEL_EL).not.toExist();
   });
 
+  it('should call onDomAdded if provided when adding the panel to the DOM', function() {
+    var onDomAdded = jasmine.createSpy();
+    var animationHookConfig = angular.extend({'onDomAdded': onDomAdded}, DEFAULT_CONFIG);
+
+    panelRef = $mdPanel.create(animationHookConfig);
+
+    panelRef.attachOnly().then(function() {
```
ErinCoughlan: These thens never get called since this isn't an async test. You can test that by adding a `fail` before the expects -- the test will still pass. The close function already clears the promise queue before continuing (via $scope.$apply and flush). For attach, you should add a function below, like `closePanel`, `openPanel`, etc.

src/components/panel/panel.spec.js
```diff
@@ -105,6 +108,62 @@ describe('$mdPanel', function() {
...
+    });
+  });
+
+  it('should call onOpenComplete if provided after adding the panel to the DOM and animating', function() {
+    var onOpenComplete = jasmine.createSpy();
+    var animationHookConfig = angular.extend({'onOpenComplete': onOpenComplete}, DEFAULT_CONFIG);
+
+    panelRef = $mdPanel.create(animationHookConfig);
+
+    openPanel();
```
ErinCoughlan: These lines can be combined, `openPanel` takes an optional config and if the panelRef isn't created yet, creates one for you. This should be:

`openPanel(animationHookConfig);`

src/components/panel/panel.spec.js
```diff
@@ -105,6 +108,62 @@ describe('$mdPanel', function() {
...
+
+    expect(onOpenComplete).toHaveBeenCalled();
+    expect(PANEL_EL).toExist();
```
ErinCoughlan: The existence of the panel is irrelevant to this test, it can be removed.

src/components/panel/panel.spec.js
```diff
@@ -105,6 +108,62 @@ describe('$mdPanel', function() {
...
+
+    closePanel().then(function() {
+      expect(onDomRemoved).not.toHaveBeenCalled();
+      expect(onRemoving).toHaveBeenCalled();
+      expect(PANEL_EL).toExist();
+    });
+
+  });
+
+  it('should call onDomRemoved if provided when removing the panel from the DOM', function() {
```
ErinCoughlan: For each of these callbacks, there should be 3 tests:

1. That the user can provide a function and it gets called. (Which is the group you wrote here.)
2. That the user can return a promise that resolves and everything continues appropiately.
3. That the user can return a promise that rejects and stops the panel. (For instance, the dev wants to show an "Are you sure you want to delete this?" dialog, and if the user responds "no", the panel does not close.)

Please add the other two types of tests. Also, since there will be 12 tests, that seems like enough for a general describe block.

ErinCoughlan: You also needs tests to make sure you callbacks are called when clicking outside the panel to close or when hitting escape to close. We want to make sure the user is able to call functions in those cases, for instance, to redirect focus somewhere on the page. (You can copy the setup of the `clickOutsideToClose` and `escapeToClose` tests.)

src/components/panel/panel.spec.js
```diff
@@ -105,6 +108,62 @@ describe('$mdPanel', function() {
     expect(PANEL_EL).not.toExist();
   });
 
+  it('should call onDomAdded if provided when adding the panel to the DOM', function() {
```
ErinCoughlan: These tests should be in the `config options:` describe block, not before the basic functionality tests for showing and hiding a panel.

src/components/panel/panel.spec.js
```diff
@@ -105,6 +108,62 @@ describe('$mdPanel', function() {
...
+    var onDomAdded = jasmine.createSpy();
+    var animationHookConfig = angular.extend({'onDomAdded': onDomAdded}, DEFAULT_CONFIG);
```
ErinCoughlan: This variable name does not explain what is going in it. I think just plain `config` is better in this case.

src/components/panel/panel.spec.js
```diff
@@ -1320,6 +1379,31 @@ describe('$mdPanel', function() {
       myButton = angular.element(document.querySelector('button'));
     });
 
+    it('should call onRemoving if provided when removing the panel from the DOM with animation config provided', function() {
+      var onRemoving = jasmine.createSpy();
+      var deferred = $q.defer();
+      var resolvedPromise = deferred.promise;
+      deferred.resolve('Animation Closing Promise Resolved');
+
+      var onRemovingAnimation = {
```
ErinCoughlan: Is this variable used?

src/components/panel/panel.spec.js
```diff
@@ -643,6 +643,227 @@ describe('$mdPanel', function() {
...
+
+        panelRef = $mdPanel.create(config);
+
+        var openRejected = false;
+
+        panelRef.show = function() {
+          return panelRef._$q.reject();
+        };
+
+        panelRef.open().then(function() {}, function () {
```
ErinCoughlan: .catch is a shortcut for only having the reject case.

src/components/panel/panel.spec.js
```diff
@@ -643,6 +643,227 @@ describe('$mdPanel', function() {
...
+          openRejected = true;
+        });
+
+        $rootScope.$apply();
```
ErinCoughlan: Please use `flushPanel()` instead.

src/components/panel/panel.spec.js
```diff
@@ -643,6 +643,227 @@ describe('$mdPanel', function() {
...
+            var onOpenComplete = jasmine.createSpy();
+            var config = angular.extend(
+                {'onOpenComplete': onOpenComplete}, DEFAULT_CONFIG);
+
+            openPanel(config);
+
+            expect(onOpenComplete).toHaveBeenCalled();
+          });
+
+      it('should call onRemoving if provided when removing the panel from the DOM',
```
ErinCoughlan: hidePanel is not removing it from the DOM, just hiding it. Please update this test description.

src/components/panel/panel.spec.js
```diff
@@ -643,6 +643,227 @@ describe('$mdPanel', function() {
...
+          });
+
+      it('should onRemoving reject, the panel should not be hidden', function() {
+        var closeRejected = false;
+        var onRemoving = jasmine.createSpy();
+        var config = angular.extend({'onRemoving': onRemoving},
+            DEFAULT_CONFIG);
+
+        openPanel(config);
+        $rootScope.$apply();
```
ErinCoughlan: This is already incorporated into `openPanel`. You can remove it.

src/components/panel/panel.spec.js
```diff
@@ -643,6 +643,227 @@ describe('$mdPanel', function() {
...
+
+        var config = angular.extend(
+            {'onDomAdded': onDomAdded}, DEFAULT_CONFIG);
+
+        panelRef = $mdPanel.create(config);
+
+        var openRejected = false;
+
+        panelRef.show = function() {
+          return panelRef._$q.reject();
```
ErinCoughlan: You aren't testing what the description says you are. You want to test what happens when `onDomAdded` is rejected. As written, you are testing what happens when `show` rejects. This also applies to the other tests in this section.

For the returning a promise tests, onDomAdded needs to be a promise that resolves (`$q.when(true)`), not a spy. For the rejecting a promise tests, onDomAdded needs to be a promise that rejects (`$q.reject()`).

src/components/panel/panel.spec.js
```diff
@@ -1475,6 +1696,29 @@ describe('$mdPanel', function() {
...
   /**
+   * Attaches the panel. If a config value is passed, creates a new panelRef
+   * using $mdPanel.open(config); Otherwise, called open on the panelRef,
+   * assuming one has already been created.
+   * @param {!Object=} opt_config
+   */
+  function attachPanel(opt_config) {
+    var attachPromise; 
+    if (opt_config) {
+      panelRef = $mdPanel.create(opt_config);
```
ErinCoughlan: This doesn't work. `$mdPanel.create` returns a promise that resolves to panelRef, not panelRef.

src/components/panel/panel.spec.js
```diff
@@ -1475,6 +1696,29 @@ describe('$mdPanel', function() {
...
+    }
+
+    if (panelRef) {
+      attachPromise = panelRef.attach();
+    }
+
+    $rootScope.$apply();
```
ErinCoughlan: You should use `flushPanel` here, like the other tests.

src/components/panel/panel.js
```diff
@@ -838,12 +837,18 @@ MdPanelRef.prototype.open = function() {
  */
 MdPanelRef.prototype.close = function() {
   var self = this;
+  
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
     var detach = self._simpleBind(self.detach, self);
 
+    var onDomRemoved = function() {
```
ThomasBurleson: try this:

```
  var self = this;

    var done = self._done(resolve, self),
        detach = self._simpleBind(self.detach, self),
        announceRemoved = self._config['onDomRemoved'] || angular.noop;

    self.hide()
        .then( detach )
        .then( announceRemoved )
        .then( done )
        .catch( reject );
  });
};
```

src/components/panel/panel.spec.js
```diff
@@ -1475,6 +1696,29 @@ describe('$mdPanel', function() {
   }
 
   /**
+   * Attaches the panel. If a config value is passed, creates a new panelRef
+   * using $mdPanel.open(config); Otherwise, called open on the panelRef,
```
ErinCoughlan: Please update this description. I also lean away from creating a panel here. You can do that in a test that uses this function, since you don't need a config to create a panel.

If you are allowing the create panel here, you'll need to support both having and not having a config. (Which means you'll need to keep track somewhere in this test whether a panel's been created or not. Since `open` also creates, you don't have an easy way to do that check.)

src/components/panel/panel.js
```diff
@@ -864,13 +869,18 @@ MdPanelRef.prototype.attach = function() {
   return this._$q(function(resolve, reject) {
```
ThomasBurleson: try this:

```js
  var done = self._done(resolve, self),
      announceAdded = self._config['onDomAdded'] || angular.noop,
      addListeners = function (response) {
          self.isAttached = true;
          self._addEventListeners();
          return response;
      };

  self._$q.all([ self._createBackdrop(),  self._createPanel().then( addListeners ) ])
    .then( announceAdded )
    .then( done )
    .catch( reject );
});
```

src/components/panel/panel.js
```diff
@@ -946,11 +956,15 @@ MdPanelRef.prototype.show = function() {
 
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
+    var onOpenComplete = function() {
```
ThomasBurleson: See above ^^ comment.

src/components/panel/panel.js
```diff
@@ -975,12 +989,16 @@ MdPanelRef.prototype.hide = function() {
   var self = this;
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
+    var onRemoving = function() {
```
ThomasBurleson: See above ^^ comment.

src/components/panel/panel.spec.js
```diff
@@ -1475,6 +1696,29 @@ describe('$mdPanel', function() {
...
+
+    if (panelRef) {
+      attachPromise = panelRef.attach();
+    }
+
+    $rootScope.$apply();
+    $rootScope.$apply();
+    $material.flushOutstandingAnimations();
+
+    return attachPromise;
```
ErinCoughlan: You shouldn't return the promise here. If you need a promise in the tests, you can use the real call in that test. (The `$rootScope.$apply()` resolves the attachPromise, so even if it's returned, you can't _do_ anything with it.)

src/components/panel/panel.spec.js
DerekLouie: Ahh, I agree with that call actually. 

Previously there were stand alone functions for the show, hide, close, functions etc that did this. But now they are simplified and I think this code would indeed be better removed.

src/components/panel/panel.spec.js
```diff
@@ -643,6 +643,227 @@ describe('$mdPanel', function() {
...
+            var config = angular.extend(
+                {'onDomAdded': onDomAdded}, DEFAULT_CONFIG);
+
+            attachPanel(config);
+
+            expect(onDomAdded).toHaveBeenCalled();
+            expect(PANEL_EL).toExist();
+          });
+
+      it('onDomAdded should return a promise and continue resolving the ' +
```
KarenParker: Please combine these two tests since they do the same thing.

src/components/panel/panel.spec.js
```diff
@@ -643,6 +643,227 @@ describe('$mdPanel', function() {
...
+
+            expect(onDomAdded).toHaveBeenCalled();
+            expect(PANEL_EL).toExist();
+            expect(PANEL_WRAPPER_CLASS).not.toHaveClass(HIDDEN_CLASS);
+          });
+
+      it('should onDomAdded reject, the panel should not be shown', function() {
+
+        var onDomAdded = function() {
+          return panelRef._$q.reject();
```
KarenParker: Don't use panelRef, inject $q if it isn't already injected above.

src/components/panel/panel.js
```diff
@@ -838,13 +838,17 @@ MdPanelRef.prototype.open = function() {
...
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
     var detach = self._simpleBind(self.detach, self);
+    var onDomRemoved = self._config['onDomRemoved'] || angular.noop;
 
     self.hide()
-        .then(detach, reject)
-        .then(done, reject);
+        .then(detach)
+        .then(onDomRemoved)
```
KarenParker: Please move this to the end of detach instead of close.

src/components/panel/panel.js
```diff
@@ -975,12 +986,16 @@ MdPanelRef.prototype.hide = function() {
...
     var done = self._done(resolve, self);
+    var onRemoving = self._config['onRemoving'] || angular.noop;
+
 
     self._$q.all([
       self._backdropRef ? self._backdropRef.hide() : self,
       self._animateClose().then(function() { self.addClass(MD_PANEL_HIDDEN); },
           reject)
-    ]).then(done, reject);
+    ]).then(onRemoving)
```
KarenParker: This line should be moved up to: 
self._animateClose()
    .then(onRemoving)
    .then(function() { self.addClass(MD_PANEL_HIDDEN); })
    .catch(reject)

src/components/panel/panel.spec.js
```diff
@@ -643,6 +606,226 @@ describe('$mdPanel', function() {
...
+            panelRef.attach();
+
+            flushPanel();
+
+            expect(onDomAdded).toHaveBeenCalled();
+            expect(PANEL_EL).toExist();
+            expect(PANEL_WRAPPER_CLASS).toHaveClass(HIDDEN_CLASS);
+          });
+
+      it('when onDomAdded resolves continue resolving the ' +
```
ErinCoughlan: This needs to be of the form: "should continue resolving when onDomAdded resolves"

src/components/panel/panel.spec.js
```diff
@@ -643,6 +606,226 @@ describe('$mdPanel', function() {
...
+            expect(attachResolved).toBe(true);
+            expect(panelRef.isAttached).toEqual(true);
+            expect(panelRef._panelContainer).not.toHaveClass(HIDDEN_CLASS);
+          });
+
+      it('when onDomAdded rejects, open should also reject', function() {
+        var openRejected = false;
+
+        var onDomAdded = function() {
+          return $q.reject();
```
ErinCoughlan: Stick with the same structure as above. In both the above tests, you also verify that onDomAdded is called.

src/components/panel/panel.spec.js
```diff
@@ -1450,6 +1633,41 @@ describe('$mdPanel', function() {
...
+
+    container.triggerHandler({
+      type: 'mouseup',
+      target: container[0]
+    });
+
+    flushPanel();
+  }
+
+  function escapeToClosePanel() {
```
ErinCoughlan: nit: this function name should indicate that an event will take place. It really has nothing to do with closing the panel. Perhaps, `pressEscape`?

src/components/panel/panel.spec.js
```diff
@@ -1450,6 +1633,41 @@ describe('$mdPanel', function() {
     attachedElements.push(element);
   }
 
+  function clickOutsideToClosePanel() {
```
ErinCoughlan: nit: This function performs a click. It doesn't close the panel. Rename this to something like: 'clickPanelContainer`

src/components/panel/panel.spec.js
```diff
@@ -643,6 +606,226 @@ describe('$mdPanel', function() {
...
+
+      it('should call onDomRemoved on escapeToClose', function() {
+        var onDomRemoved = jasmine.createSpy();
+        var config = angular.extend({
+          'onDomRemoved': onDomRemoved, escapeToClose: true},
+          DEFAULT_CONFIG);
+
+        openPanel(config);
+
+        var container = panelRef._panelContainer;
```
ErinCoughlan: This line is part of `escapeToClose`. It doesn't need to be here.

src/components/panel/panel.spec.js
```diff
@@ -643,6 +606,226 @@ describe('$mdPanel', function() {
...
+      });
+
+      it('should call onOpenComplete if provided after adding the panel to the ' +
+          'DOM and animating', function() {
+            var onOpenComplete = jasmine.createSpy();
+            var config = angular.extend(
+                {'onOpenComplete': onOpenComplete}, DEFAULT_CONFIG);
+
+            panelRef = $mdPanel.create(config);
+            openPanel(config);
```
ErinCoughlan: Since you don't care about the panelRef in this test, you can just call `openPanel(config)`

ThomasBurleson:
@DerekLouie  - the JS lgtm

DerekLouie:
Alrighty! Just finished the last round of comments, please take another look at your earliest convenience!

KarenParker:
👍

PR #8034, opened 2016-04-14
feat(chips): Implement ngModelController.$isEmpty to return true when…

Emeegeemee:
… the model has no elements

This lets you use the built-in required validator to check when the user has added at least a single element to the chips array.

Added a demo for the md-max-chips and required validators.

### DIFF
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -659,6 +659,59 @@ describe('<md-chips>', function() {
 
       });
 
+      describe('required', function() {
+        beforeEach(function() {
+          // Clear default items to test the required chips functionality
+          scope.items = [];
+        });
+
+        it('should set the required error when chips is compiled with an empty array', function() {
+          var template =
+              '<form name="form">' +
+              '<md-chips name="chips" required ng-model="items"></md-chips>' +
+              '</form>';
+
+          var element = buildChips(template);
+          element.scope().$apply();
+
+          expect(scope.form.chips.$error['required']).toBe(true);
+        });
+
+        it('should unset the required error when the first chip is added', function() {
+          var template =
+              '<form name="form">' +
+              '<md-chips name="chips" required ng-model="items"></md-chips>' +
+              '</form>';
+
+          var element = buildChips(template);
+          var ctrl = element.find('md-chips').controller('mdChips');
+
+          element.scope().$apply(function() {
+            ctrl.chipBuffer = 'Test';
+            simulateInputEnterKey(ctrl);
+          });
+
+          expect(scope.form.chips.$error['required']).toBeUndefined();
+        });
+
+        it('should set the required when the last chip is removed', function() {
+          scope.items = ['test'];
+          var template =
+              '<form name="form">' +
+              '<md-chips name="chips" required ng-model="items"></md-chips>' +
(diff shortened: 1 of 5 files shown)


src/components/chips/js/chipsController.js
```diff
@@ -397,6 +397,7 @@ MdChipsCtrl.prototype.hasMaxChipsReached = function() {
  */
 MdChipsCtrl.prototype.validateModel = function() {
   this.ngModelCtrl.$setValidity('md-max-chips', !this.hasMaxChipsReached());
+  this.ngModelCtrl.$validate(); // rerun any registered validators
```
crisbeto: It might be a good idea to check if the `$validate` method exists here. If I remember correctly, it was introduced in Angular 1.4.

src/components/chips/chips.spec.js
```diff
@@ -659,6 +659,59 @@ describe('<md-chips>', function() {
...
+        it('should set the required error when chips is compiled with an empty array', function() {
+          var template =
+              '<form name="form">' +
+              '<md-chips name="chips" required ng-model="items"></md-chips>' +
+              '</form>';
+
+          var element = buildChips(template);
+          element.scope().$apply();
+
+          expect(scope.form.chips.$error['required']).toBe(true);
```
crisbeto: You can use `scope.form.chips.$error.required`, instead of the bracket notation here and the other tests.

DevVersion: I think the bracket-notation is fine here, because this shows that we're not sure about the existence of that variable.

crisbeto: I'm not sure how it implies that we don't know whether `required` exists, considering that both notations behave exactly the same.

DevVersion: @crisbeto Tue, the bracket-notation is mostly used for variables which contain special characters, but in my opinion it also represents the existence.

src/components/chips/js/chipsController.js
Emeegeemee: Looking at the angular documentation it appears to have been added in [1.3.0](https://code.angularjs.org/1.3.0/docs/api/ng/type/ngModel.NgModelController), which is the minimum version angular material supports.

src/components/chips/chips.spec.js
Emeegeemee: Let me know if you want me to switch to the dot notation, I was just copying how the md-max-chips validator test did it, but realize now that it used bracket notation since has special characters in the name.

src/components/chips/js/chipsController.js
crisbeto: I'd say it's alright in that case.

Emeegeemee:
@crisbeto @DevVersion Friendly ping.

Please let me know if I should switch to the dot notation just didn't want to make a whole commit with just switching the syntax. Otherwise it would be great if this PR could make it into the 1.1.0 release, but understand if that isn't possible.

Thanks

ThomasBurleson:
This does not address a known issue and presents risk to current stability.
Closing and will not merge.

Emeegeemee:
I've filed #8126 which shows the current behavior this PR was fixing that is needed for a project I'm working on.

Issue #7683, opened 2016-03-21
mdPanel: Add disableParentScroll config option to the panel.


KarenParker:
Whether the user can scroll the page behind the panel. Defaults to false.

When the option is set to true, handle the panel scrolling with the page. This means the panel may be partially hidden off the screen.

Consider following the mdDialog implementation [here](https://github.com/angular/material/blob/master/src/components/dialog/dialog.js#L861).

-----------
This task involves:

- Updating public documentation
- Adding disableParentScroll to the panel config

ErinCoughlan:
[This resize code](https://github.com/angular/material/blob/3524aa9573fe29e53341c60ce6127fc7ad323446/src/components/select/select.js#L1089) from select may be helpful. The relativePositioning code will likely be based off of select as well.

gmoothart:
I think we should rename this just `disableScroll`. In md-dialog, there is a `parent` option which says where to append the dialog, so it makes sense to use that word. But we renamed that to `attachTo`. I think just dropping the 'parent' from this config option is cleanest and least confusing. Thoughts?

ErinCoughlan:
@gmoothart That sounds good to me. I think it still conveys the meaning, plus its shorter to type.

LuisRevillaM:
I am building a panel that contains a box of pictures. This box has an explicit height, and the overflow is set to scroll.

But when I set disableParentScroll to true (so that the body behind the panel is fixed), the scrolling in the panel's inner box of pictures does not work. 

When I set disableParentScroll to false, the scrolling in the panel's box works well, but the body behind the panel scrolls.

PR #8048, opened 2016-04-14
update(panel): Adding disableParentScroll

gmoothart:
This adds a config to the panel which disables scrolling of the panel's
parent (attachTo) element, if set.

Part of #7683

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -129,14 +129,14 @@ angular
  *     `MdPanelAnimation`.
  *   - `hasBackdrop` - `{boolean=}`: Whether there should be an opaque backdrop
  *     behind the panel. Defaults to false.
+ *   - `disableParentScroll` - `{boolean=}`: Whether the user can scroll the
+ *     page behind the panel. Defaults to false.
  *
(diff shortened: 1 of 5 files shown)


gmoothart:
PTAL @DerekLouie @ErinCoughlan @KarenParker @ThomasBurleson @jelbourn

KarenParker:
Please add a test for this (dialog has an example of what you can test).

gmoothart:
Added tests and fixed a z-index issue, PTAL @DerekLouie @ErinCoughlan @KarenParker @ThomasBurleson @jelbourn

src/components/panel/panel.js
```diff
@@ -935,6 +939,11 @@ MdPanelRef.prototype.detach = function() {
     return self._$q.resolve(self);
   };
 
+  if (this._restoreScroll) {
+    this._restoreScroll();
+    delete this._restoreScroll;
```
jelbourn: Just set to `null`; it accomplishes the same thing and is generally always faster.

jelbourn:
LGTM

PR #8051, opened 2016-04-14
fix(colors): coverts COLOR_PALETTES to colorPalettes

robertmesserle:
Value is reassigned, so it is not a constant.


robertmesserle:
@ThomasBurleson This was breaking internal builds that treat all-caps variables as constants.

Issue #7988, opened 2016-04-11
bug(select): select menu should not be underlined


robertmesserle:
In RC2, `md-select` menus are now underlined.  This is breaking a number of layouts and does not appear to be according to spec.

clshortfuse:
Well, sometimes it's underlined, sometimes it's not. I haven't figured out exactly when, but I believe it should be optional.

[material design spec example](https://material-design.storage.googleapis.com/publish/material_v_4/material_ext_publish/0Bzhp5Z4wHba3dUVPRTdfUXlZWnM/components_textfields_single1.png)

The top select isn't underlined, but the bottom one is.

I worked around this by making it optional. See http://codepen.io/shortfuse/pen/oxGowZ

    md-select.md-no-border .md-select-value {
      border-bottom-color: transparent !important;
    }

Koslun:
Got a quick-fix until the RC version with a merged PR #8672 gets merged and the next RC version with it, for those of us not working directly with the master branch:

```CSS
md-select ._md-select-value {
  border-bottom-color: transparent !important;
}
```
Obviously not ideal since it removes the option of showing the underline in some cases while it also depends on a private CSS class that could be changed at any point.

topherfangio:
@Koslun Couldn't you add your own class to the select so that you have more control over when it hides the underline?

```css
md-select.my-custom-class ._md-select-value {
  border-bottom-color: transparent !important;
}
```

Koslun:
@topherfangio Yup, you're right, that fixes it nicely :+1: . 

Tried your solution previously but with a space between `md-select` and `.my-custom-class`, which does not select the correct element.

topherfangio:
@Koslun Awesome; glad it worked! Hopefully the PR will be merged soon, but it looks like it breaks pretty badly on IE, so I have to fix that first.

PR #8052, opened 2016-04-14
fix(input, select, autocomplete, checkbox) correct all margins, paddings and offsets to match spec

clshortfuse:
This is a WIP but I believe completed my changes for md-input-container

I still need to write in the rest, but I wanted to share the changes I've done with input.scss

You can monitor and compare progress of this here:

specFix: http://codepen.io/shortfuse/pen/qZojGW
master: http://codepen.io/shortfuse/pen/ZWxYpN

This is my first time playing with scss so I'll need somebody else to double my work.

You will notice there are some awkward raw numbers being used for padding and margins. The reason why we can't use the exact pixels as represented by the spec is because with webkit, the raw height of the <input> includes leading metrics (space above and below font glyphs). You can't set a line-height less than 100%, so I had to manually set the padding and margins.

The other way to tackle this problem would require JS implementation. In this case, you would be able set the height attribute of <input> equal to the font size. The way I did it, while the numbers are awkward would require no JS. I'm not sure if it'll play nice with IE, but if somebody wants to write a JS component to change the height attribute of the <input>, that could work. My only concern is if, by changing the height, font glyphs could be clipped.

There are no forced height options and most items resize properly to spec, with a margin of error of a pixel or two. In my codepen, all items will align properly to baseline, as long as layout-align is used properly.

I'm using last-of-type, so if you want items to fill width properly, wrap it in a div if it's alone.

The main reason why everything flows properly is because I added ~~use the label+ CSS syntax~~ md-input-has-label attribute which will only apply certain metrics if a label present.

Right now I'm working on adding on the errors component of md-input-container to the scss, but I already have that in my codepen.

### DIFF
--- a/src/components/autocomplete/autocomplete.scss
+++ b/src/components/autocomplete/autocomplete.scss
@@ -71,16 +71,19 @@ md-autocomplete {
     &.md-menu-showing {
       z-index: $z-index-backdrop + 1;
     }
+
+    md-input-container.md-input-has-messages md-progress-linear.md-inline {
+      bottom: 16px;
+    }
+
     md-progress-linear {
       position: absolute;
       bottom: -2px;
       left: 0;
       // When `md-inline` is present, we adjust the offset to go over the `ng-message` space
       &.md-inline {
-        bottom: 40px;
-        right: 2px;
-        left: 2px;
-        width: auto;
+        bottom: 8px;
+        width: 100%;
       }
       ._md-mode-indeterminate {
         position: absolute;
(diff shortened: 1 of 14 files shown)


clshortfuse:
fixes https://github.com/angular/material/issues/7988

ThomasBurleson:
Will not merge.

clshortfuse:
@ThomasBurleson Can I get a more detailed response as to why? 

Is it because it fails build? It's a WIP as I said in the first sentence.

PR #8053, opened 2016-04-14
fix(list): Correct avatar/icon size/spacing.

topherfangio:
Avatars and icons within the lists had incorrect sizing/spacing on iOS due to some flex issues and incorrect `height: 100%` calculation on the SVG.

Tested on Desktop Chrome/Firefox/Safari/Edge/IE and iOS Safari/Chrome. Could use some manual testing on Android versions.

### DIFF
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -82,6 +82,7 @@ md-list {
           @include rtl-prop(margin-right, margin-left, $list-item-primary-width - $list-item-dense-primary-avatar-width);
         }
         .md-avatar {
+          flex: none;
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@topherfangio LGTM

src/components/list/list.scss
```diff
@@ -267,11 +268,19 @@ md-list-item {
       box-sizing: content-box;
     }
     & .md-avatar {
+      flex: none;
```
ThomasBurleson: Truly odd requirement.

Issue #8044, opened 2016-04-14
md-colors not working


sdesmond:
Let me know if I'm just doing this wrong.  I'm not seeing my newly defined palettes in $mdColorPalette which is causing an error at line 6908.

Pen here: http://codepen.io/anon/pen/aNYPNe?editors=1010

Angular Material 1.1.0-rc3

Issue #8043, opened 2016-04-14
md-colors not picking from default theme


sdesmond:
See Pen here: http://codepen.io/anon/pen/dMmQBP?editors=1010

I set a new theme with $mdThemingProvider.setDefaultTheme.  I would expect md-colors="{background: 'primary'}" to pick the primary color from the new default theme, but it doesn't.  It's actually using the default blue/pink theme.

Angular Material 1.1.0-rc3

PR #8061, opened 2016-04-15
fix(colors): using default palette and defined palettes from $mdTheming

### LINKED ISSUE #8036: docs(colors): document $mdColors service
As `mdColors` is documented, i can't seem to find any reason `$mdColors` service shouldn't be documented as well.

EladBezalel:
- replaced `'default'` to use the actual defined palette from `$mdTheming`
- `$mdColorPalette` didn't reflected the registered palettes from `$mdTheming`, made `$mdTheming` expose it's registered palettes
- added documentation to `$mdColors` service

fixes #8036 #8043 #8044

### DIFF
--- a/src/components/colors/colors.js
+++ b/src/components/colors/colors.js
@@ -23,7 +23,7 @@
   /**
    * @ngdoc service
    * @name $mdColors
-   * @module material.core.theming.colors
+   * @module material.components.colors
    *
    * @description
    * With only defining themes, one couldn't get non ngMaterial elements colored with Material colors,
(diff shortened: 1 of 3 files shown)


src/core/services/theming/theming.js
```diff
@@ -353,6 +353,7 @@ function ThemingProvider($mdColorPalette) {
         };
 
     applyTheme.THEMES = angular.extend({}, THEMES);
+    applyTheme.PALETTES = angular.extend({}, PALETTES);
```
ThomasBurleson: Why are you duplicating the Palette structures ?

src/components/colors/colors.js
```diff
@@ -123,9 +158,9 @@
      * For the evaluated expression, extract the color parts into a hash map
      */
     function extractColorOptions(expression) {
-      var parts = expression.split('-'),
-        hasTheme = angular.isDefined($mdTheming.THEMES[parts[0]]),
-        theme = hasTheme ? parts.splice(0, 1)[0] : 'default';
+      var parts = expression.split('-');
+      var hasTheme = angular.isDefined($mdTheming.THEMES[parts[0]]);
+      var theme = hasTheme ? parts.splice(0, 1)[0] : $mdTheming.defaultTheme();
```
crisbeto: Isn't `parts.splice(0, 1)[0]` the same as `parts[0]`?

crisbeto: I believe `parts.splice(0, 1)[0]` is the same as `parts.shift`.

ThomasBurleson: @crisbeto - is that trivia or important? Can you elaborate ?

crisbeto: It's not particularly important, but `splice` produces a new array with one element, from which we take the first element, whereas `shift` just removes the first element and returns it.

PR #8067, opened 2016-04-15
fix(md-content): Allow md-content to use md-hue-x for background colors

### LINKED ISSUE #7886: md-content no longer allows md-hue-1/2/3 in 1.1.0-rc2
md-hue-1 / 2 / 3 all return the same background colour for md-content

looking at the specs https://www.google.com/design/spec/style/color.html#color-text-background-colors

it suggests that I should be able to use hues 50/100/300 on background elements.

hodeyp:
re-introduce ability for md-content to use md-hue-x colors

closes #7886

### DIFF
--- a/src/components/content/content-theme.scss
+++ b/src/components/content/content-theme.scss
@@ -1,6 +1,4 @@
 md-content.md-THEME_NAME-theme {
   color: '{{foreground-1}}';
-  background-color: '{{background-default}}';
+  background-color: '{{background-color}}';
 }
-
-
--- a/src/components/sidenav/sidenav-theme.scss
+++ b/src/components/sidenav/sidenav-theme.scss
@@ -1,5 +1,5 @@
 md-sidenav.md-THEME_NAME-theme {
   &, md-content {
-    background-color: '{{background-hue-1}}';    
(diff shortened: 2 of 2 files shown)


ThomasBurleson:
@EladBezalel - Can you investigate/research this?

src/components/content/content-theme.scss
```diff
@@ -1,6 +1,4 @@
 md-content.md-THEME_NAME-theme {
   color: '{{foreground-1}}';
-  background-color: '{{background-default}}';
+  background-color: '{{background-color}}';
```
DevVersion: Not sure, where this property exists? I think you're using our theming service wrong.

hodeyp: this is simply adding back functionality that was introduced in commit 0df4b16eca2c8af9805615da646d38db04a33d7f on March 18, 2015. The md-content component has always allowed for md-hue-x since v0.9.0.

DevVersion: Indeed, I was just wondering where the `color` suffix is validated in the theming service.

ThomasBurleson:
Will not merge.

PR #8073, opened 2016-04-16
docs(tools-bottom): remove undefined class

chrisconover:
Remove undefined "md-toolbar-tools-bottom" class from the example code in the documentation to avoid confusion (the class is not defined anywhere in the repo).


DevVersion:
LGTM

PR #8083, opened 2016-04-17
feat(datepicker): add support for md-input-container

### LINKED ISSUE #4233: md-datepicker should support md-input-container and floating label
Currently the datepicker doesn't align well with other input fields. It would be nice if we can do something like

```html
<md-input-container>
    <label>Date</label>
    <md-datepicker ng-model="myDate"></md-datepicker>
</md-input-container>
```
and have everything well aligned with other form fields.
Also the icon should be optional.

crisbeto:
Makes the datepicker compatible with the md-input-container directive.

Closes #4233.

@ThomasBurleson, @DevVersion, @EladBezalel, @topherfangio Could you guys give me some feedback on this one? I tried to test out most scenarios to make sure it behaves properly (with other input elements, with errors etc.), but it's always possible that I missed something.

@topherfangio Can you take a look at [this change in particular](https://github.com/angular/material/pull/8083/files#diff-da06475ed96460fbcb944d6f1f31748aL780), it was preventing consecutive errors from animating in (e.g. going from a `required` error to some other type). It seems that it [got introduced in this change](https://github.com/angular/material/commit/cf7f21aa6c633d9f0bd22013e9a79f99b82b16af#diff-da06475ed96460fbcb944d6f1f31748aR635). Removing it doesn't seem to break any tests and I couldn't see anything wrong with the `ngMessages` in other demos.

### DIFF
--- a/src/components/datepicker/datePicker.scss
+++ b/src/components/datepicker/datePicker.scss
@@ -1,15 +1,18 @@
 /** Styles for mdDatepicker. */
 $md-datepicker-button-gap: 12px !default;  // Space between the text input and the calendar-icon button.
 $md-datepicker-border-bottom-gap: 5px !default;  // Space between input and the grey underline.
+$md-date-arrow-size: 5px !default; // Size of the triangle on the right side of the input.
 $md-datepicker-open-animation-duration: 0.2s !default;
 $md-datepicker-triangle-button-width: 36px !default;
+$md-datepicker-input-mask-height: 40px !default;
 
 md-datepicker {
   // Don't let linebreaks happen between the open icon-button and the input.
   white-space: nowrap;
   overflow: hidden;
 
-  // Leave room for the down-triangle button to "overflow" it's parent without modifying scrollLeft
+  // Leave room for the down-triangle button to "overflow" it's parent without modifying scrollLeft.
+  // This prevents the element from shifting right when opening via the triangle button.
   @include rtl-prop(padding-right, padding-left, $md-datepicker-triangle-button-width / 2);
   @include rtl-prop(margin-right, margin-left, -$md-datepicker-triangle-button-width / 2);
 
@@ -31,11 +34,36 @@ md-datepicker {
 }
 
 // The input into which the user can type the date.
-.md-datepicker-input {
+.md-datepicker-input, label:not(.md-no-float):not(._md-container-ignore) {
   @include md-flat-input();
   min-width: 120px;
   max-width: $md-calendar-width - $md-datepicker-button-gap;
 }
+
+// If the datepicker is inside of a md-input-container
+._md-datepicker-floating-label {
+  > label:not(.md-no-float):not(._md-container-ignore) {
+    $offset: $md-datepicker-triangle-button-width + $md-datepicker-button-gap + $icon-button-margin + $baseline-grid;
+    @include rtl-prop(left, right, $offset);
+    width: calc(100% - #{$offset + $md-datepicker-triangle-button-width});
+  }
+
+  > md-datepicker {
+    // Prevents the ripple on the triangle from being clipped.
+    overflow: visible;
+
+    .md-datepicker-input-container {
+      border: none;
+    }
+
+    .md-datepicker-button {
+      // Prevents the button from wrapping around.
+      @include rtl(float, left, right);
+      margin-top: -$md-datepicker-border-bottom-gap / 2;
+    }
+  }
+}
+
 
 // Container for the datepicker input.
 .md-datepicker-input-container {
@@ -61,11 +89,15 @@ md-datepicker {
 
 // Floating pane that contains the calendar at the bottom of the input.
 .md-datepicker-calendar-pane {
+  // On most browsers the `scale(0)` below prevents this element from
+  // overflowing it's parent, however IE and Edge seem to disregard it.
+  // The `left: -100%` pulls the element back in order to ensure that
+  // it doesn't cause an overflow.
+  left: -100%;
   position: absolute;
   top: 0;
   left: 0;
   z-index: $z-index-calendar-pane;
-
   border-width: 1px;
   border-style: solid;
   background: transparent;
@@ -81,7 +113,9 @@ md-datepicker {
 
 // Portion of the floating panel that sits, invisibly, on top of the input.
 .md-datepicker-input-mask {
-  height: 40px;
+  // It needs to be 1px shorter because of the datepicker-input-container's bottom border,
+  // which can cause a slight hole in the mask when it's inside a md-input-container.
+  height: $md-datepicker-input-mask-height - 1px;
   width: $md-calendar-width;
   position: relative;
 
@@ -122,7 +156,6 @@ md-datepicker {
 // Down triangle/arrow indicating that the datepicker can be opened.
 // We can do this entirely with CSS without needing to load an icon.
 // See https://css-tricks.com/snippets/css/css-triangle/
-$md-date-arrow-size: 5px !default;
 .md-datepicker-expand-triangle {
   // Center the triangle inside of the button so that the
   // ink ripple origin looks correct.
@@ -142,7 +175,7 @@ $md-date-arrow-size: 5px !default;
 .md-datepicker-triangle-button {
   position: absolute;
   @include rtl-prop(right, left, 0);
-  top: 0;
+  top: $md-date-arrow-size;
 
   // TODO(jelbourn): This position isn't great on all platforms.
   @include rtl(transform, translateY(-25%) translateX(45%), translateY(-25%) translateX(-45%));
@@ -151,7 +184,7 @@ $md-date-arrow-size: 5px !default;
 // Need crazy specificity to override .md-button.md-icon-button.
 // Only apply this high specifiy to the property we need to override.
 .md-datepicker-triangle-button.md-button.md-icon-button {
-  height: 100%;
+  height: $md-datepicker-triangle-button-width;
   width: $md-datepicker-triangle-button-width;
   position: absolute;
 }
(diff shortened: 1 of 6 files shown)


src/components/datepicker/datePicker.scss
```diff
@@ -165,21 +197,31 @@ md-datepicker[disabled] {
...
+  overflow: hidden;
+
   .md-datepicker-input-container {
     @include rtl-prop(margin-left, margin-right, -$md-datepicker-button-gap);
 
     // The negative bottom margin prevents the content around the datepicker
     // from jumping when it gets opened.
     margin-bottom: -$md-datepicker-border-bottom-gap;
-    border: none;
+    border-color: #fff;
```
DevVersion: Why is that explicit `#fff`?

src/components/datepicker/datePicker.scss
```diff
@@ -30,12 +32,38 @@ md-datepicker {
...
+
+    .md-datepicker-input-container {
+      border: none;
+      padding-bottom: 0;
+    }
+
+    .md-datepicker-button {
+      // Prevents the button from wrapping around.
+      @include rtl(float, left, right);
+      margin-top: -$md-datepicker-border-bottom-gap/2;
```
DevVersion: Nit: Space the arithmetic operator.

src/components/datepicker/datePicker.scss
crisbeto: Ah, it should be `transparent`, must've missed it while trying out stuff. I was trying to avoid it jumping by removing the border altogether.

src/components/datepicker/datePicker.js
```diff
@@ -221,6 +250,9 @@
     /** @type {boolean} Whether the date-picker's calendar pane is open. */
     this.isCalendarOpen = false;
 
+    /** @final */
```
DevVersion: Why are you using `@final` as an annotation here? This seems to be unnecessary.

crisbeto: I'm not quite sure what types of annotations these were. There were some other properties annotated that way.

src/components/datepicker/datePicker.scss
```diff
@@ -57,9 +85,11 @@ md-datepicker {
 
 // Floating pane that contains the calendar at the bottom of the input.
 .md-datepicker-calendar-pane {
+  // We need to pull the pane backwards, otherwise it
+  // causes an overflow in IE and Edge.
+  left: -100%;
```
DevVersion: I'm confused about this line and the comment is IMO not saying why this is required.

src/components/datepicker/datePicker.js
DevVersion: Oh yes, can see that. Then you can keep that - but it seems weird.

src/components/datepicker/datePicker.scss
```diff
@@ -138,7 +170,7 @@ $md-date-arrow-size: 5px !default;
 .md-datepicker-triangle-button {
   position: absolute;
   @include rtl-prop(right, left, 0);
-  top: 0;
+  top: 5px;
```
DevVersion: Why is that `5px` moved down? Are you trying to center it vertically?

src/components/datepicker/datePicker.scss
crisbeto: [A little bit down](https://github.com/angular/material/pull/8083/files#diff-1be65f67ef78c5d0c4961b3ff6776f45R99) the element's scale is set to 0, which prevents it from overflowing in all browsers, except IE and Edge. The `left: -100%` prevents that overflow. I'll work on the explanation.

src/components/input/input.js
```diff
@@ -777,14 +777,6 @@ function ngMessagesAnimation($q, $animateCss) {
 function ngMessageAnimation($animateCss) {
   return {
     enter: function(element, done) {
-      var messages = getMessagesElement(element);
-
-      // If we have the md-auto-hide class, the md-input-invalid animation will fire, so we can skip
-      if (messages.hasClass('md-auto-hide')) {
-        done();
-        return;
```
DevVersion: Yes, this lines are required, otherwise the input container will fire the enter animations twice. 
 > https://github.com/angular/material/pull/7583/files

src/components/datepicker/datePicker.scss
crisbeto: Yes, but I forgot to use `$md-datepicker-border-bottom-gap` instead.

src/components/datepicker/datePicker.js
```diff
@@ -385,7 +420,7 @@
         var maxDate = this.dateUtil.createDateAtMidnight(this.maxDate);
         this.ngModelCtrl.$setValidity('maxdate', date <= maxDate);
       }
-
```
EladBezalel: i'd revert those empty lines and lines you didn't changed so when looking at things that change you won't show up

src/components/input/input.js
crisbeto: But in that case consecutive errors weren't being animated in. E.g. when the input is empty, the `required` error is shown, but if I enter an invalid date afterwards, it's error message stays at `opacity: 0` and `margin: -100%`. The only workaround that I found were removing that line or using `md-auto-hide`.

src/components/datepicker/datePicker.js
crisbeto: Shouldn't those stray white spaces at the ends of lines be removed in general?

src/components/input/input.js
DevVersion: Weird. Is the the class `md-auto-hide` not added automatically? 

> https://github.com/DevVersion/material/blob/master/src/components/input/input.js#L687

src/components/datepicker/datePicker.scss
```diff
@@ -165,21 +197,31 @@ md-datepicker[disabled] {
...
+  .md-datepicker-input,
+  label:not(.md-no-float):not(._md-container-ignore) {
+    margin-bottom: -$md-datepicker-border-bottom-gap;
   }
 
   .md-datepicker-input {
     @include rtl-prop(margin-left, margin-right, 24px);
-    height: 40px;
+    height: $md-datepicker-input-mask-height;
+    border-bottom-color: transparent !important;
```
EladBezalel: if `.md-datepicker-input` isn't private than we can't apply `!important`

src/components/input/input.js
crisbeto: Sorry, what I meant is if removed the `md-auto-hide`, e.g. by having an `ng-if` or `ng-show`.

src/components/datepicker/datePicker.scss
EladBezalel: BTW i'm not sure @jelbourn did privatizing to the datepicker at all

crisbeto: Doesn't look like it. I can do it now, but it would make this a breaking change.

EladBezalel: I guess you should make another PR for it and rebase this one on top of it

crisbeto: Makes sense, I'll make another issue and wait up until this one is ready and merged.

crisbeto:
I reverted that workaround for the `ngMessages` since it's a known issue.

jelbourn:
@EladBezalel @crisbeto I didn't mark any of the datepicker classes as private since, at the time, none of them stuck me as particularly sensitive.

ThomasBurleson:
@crisbeto - I worry about the scope of these changes.

crisbeto:
How come @ThomasBurleson? Pretty much all other input components seem to support being inside a `md-input-container`, which I think is why a lot of people were asking for the datepicker in #4233. Also I wouldn't say it's much of a breaking change because the old behaviour is still there and you won't need to do any changes.

mramanauskas:
@crisbeto One of the main targets of this issue was:
Icon should be optional.

So I think that md-hide-button attribute is very important, which is implementend here:
https://github.com/angular/material/pull/7675

CSS hack is very ugly
<pre>
.md-datepicker-button {
  display: none;
}
</pre>

crisbeto:
How is that ugly? IMO it's a legitimate way of hiding something with CSS.

mramanauskas:
Just hidding button doesnt provide what is expected, because of wrong paddings and weird empty spaces (on desktop).

In my opinion, legitimate way is to configure component with predefined attributes. Overriding component with custom CSS would be just opportunity.

crisbeto:
I'll ask for some opinions, but I think this is more of a specific use-case, rather than a more general one and it doesn't have to be in the library.

mramanauskas:
This PR issue first comment (with 29 thumbs up) and it's last sentence informs about other opinions.
https://github.com/angular/material/issues/4233#issue-101855018

I we want this icon to be by default, why don't we add this to select or simple input....?

EricWVGG:
If this works and gets merged in your are as like a hero unto me.

RIAstar:
Just tested by cherry-picking this commit onto 1.1.0-rc.5 and noticed one small flaw:
when applying the `required` attribute, an asterisk (*) is not added to the label as it is with any other input in an `md-input-container`.

Might be I'm missing something due to the cherry-picking, but I thought I'd let you know.

crisbeto:
I probably missed it when I was working on this PR. Can you file an issue about it?

RIAstar:
Another one: integration with ng-messages is quirky.
First validation goes well, but if I make another mistake, the new message is in the DOM, but hidden (`md-auto-hide` class applied). It works as expected when the `md-datepicker` is not inside an `md-input-container`.

Would you like a separate issue for that? Or do you prefer to have both in one place?

crisbeto:
That's a known issue that we're tracking already. It's with the input container, not the datepicker.

RIAstar:
Here you go: https://github.com/angular/material/issues/8950

crisbeto:
Thank you.

RIAstar:
Up to you to assess whether this can be considered a bug: 
the ng-messages are aligned to the far left edge of the container; I would expect them to be aligned with the date input field, just like the label.

Looks like this now:
```
        label
  icon  _________
Some error
```

I would expect:
```
        label
  icon  _________
        Some error
```

RIAstar:
Sorry to bombard you like this ;)

This selector (https://github.com/angular/material/pull/8083/commits/4e4de24c63dcd87cc9d051b4809fbcec5bd1cdcf#diff-1be65f67ef78c5d0c4961b3ff6776f45R37) is way to broad. It basically selects all the labels in your app.

crisbeto:
It was fixed in a later PR.

PR #8084, opened 2016-04-17
fix(tooltip): empty tooltips should not show up.

### LINKED ISSUE #8021: Tooltip should not show if content is empty
Pen at http://codepen.io/anon/pen/bpvBJV shows one normal tooltip and one empty one.

If the tooltip content is empty, consider not showing anything instead of showing an empty tooltip.  The Angular Bootstrap team does this (empty strings and undefined/null don't show) and it makes for very easy conditional tooltips.

DevVersion:
Fixes #8021.


ThomasBurleson:
Closed with SHA c513e493c547352ba3b1641e38c2e2ee43da8894

PR #8098, opened 2016-04-18
 update(panel): Updates the detach / close code to focus on close.

### LINKED ISSUE #7902: mdPanel: Add destroy.

DerekLouie:
Hello,

@gmoothart @ErinCoughlan @KarenParker @ThomasBurleson 

Here is a small PR to update the panels close functionality.

The user can also now specify an element / selector to focus on after the panel closes.

Minor bug fixes as well.

Closes #7902

Please Review.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -139,6 +139,9 @@ angular
  *     close/hide() action is starting.
  *   - `onDomRemoved` - `{function=}`: Callback function used to announce when the
  *     panel is removed from the DOM.
+ *   - `origin` - `{(string|!angular.JQLite|!Element)=}`: The element to
+ *     focus on when the panel closes. This is commonly the element which triggered
+ *     the opening of the panel.
  *
  * TODO(ErinCoughlan): Add the following config options.
  *   - `groupName` - `{string=}`: Name of panel groups. This group name is
@@ -236,8 +239,7 @@ angular
  * @ngdoc method
  * @name MdPanelRef#close
  * @description
- * Hides and detaches the panel. This method destroys the reference to the panel.
- * In order to open the panel again, a new one must be created.
+ * Hides and detaches the panel.
  *
  * @returns {!angular.$q.Promise} A promise that is resolved when the panel is
  * closed.
@@ -724,6 +726,7 @@ MdPanelService.prototype._wrapTemplate = function(origTemplate) {
       '</div>';
 };
 
+
 /*****************************************************************************
  *                                 MdPanelRef                                *
  *****************************************************************************/
@@ -777,7 +780,6 @@ function MdPanelRef(config, $injector) {
    * @type {boolean}
    */
   this.isAttached = false;
-
 
   // Private variables.
   /** @private {!Object} */
@@ -896,7 +898,7 @@ MdPanelRef.prototype.detach = function() {
   var onDomRemoved = self._config['onDomRemoved'] || angular.noop;
 
   var detachFn = function() {
-    self._removeEventListener();
+    self._removeEventListeners();
 
     // Remove the focus traps that we added earlier for keeping focus within
     // the panel.
@@ -928,6 +930,14 @@ MdPanelRef.prototype.detach = function() {
       .then(done)
       .catch(reject);
   });
+};
+
+
+/**
+ * Destroys the panel. The Panel cannot be opened again after this.
+ */
+MdPanelRef.prototype.destroy = function() {
+  this._config.locals = null;
 };
 
 
@@ -986,16 +996,28 @@ MdPanelRef.prototype.hide = function() {
   }
 
   var self = this;
+
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
     var onRemoving = self._config['onRemoving'] || angular.noop;
 
+    var focusOnOrigin = function() {
+      var origin = self._config['origin'];
+      if (origin) {
+        getElement(origin).focus();
+      }
+    };
+
+    var hidePanel = function() {
+      self.addClass(MD_PANEL_HIDDEN);
+    };
 
     self._$q.all([
       self._backdropRef ? self._backdropRef.hide() : self,
       self._animateClose()
           .then(onRemoving)
-          .then(function() { self.addClass(MD_PANEL_HIDDEN); })
+          .then(hidePanel)
+          .then(focusOnOrigin)
           .catch(reject)
     ]).then(done, reject);
   });
@@ -1062,6 +1084,7 @@ MdPanelRef.prototype._createPanel = function() {
     if (!self._config.locals) {
       self._config.locals = {};
     }
+
     self._config.locals.mdPanelRef = self;
     self._$mdCompiler.compile(self._config)
         .then(function(compileData) {
@@ -1205,7 +1228,7 @@ MdPanelRef.prototype._addEventListeners = function() {
  * Remove event listeners added in _addEventListeners.
  * @private
  */
-MdPanelRef.prototype._removeEventListener = function() {
+MdPanelRef.prototype._removeEventListeners = function() {
   this._removeListeners && this._removeListeners.forEach(function(removeFn) {
     removeFn();
   });
@@ -1309,6 +1332,12 @@ MdPanelRef.prototype._configureTrapFocus = function() {
     };
     this._topFocusTrap.addEventListener('focus', focusHandler);
     this._bottomFocusTrap.addEventListener('focus', focusHandler);
+
+    // Queue remove listeners function
+    this._removeListeners.push(this._simpleBind(function() {
+      this._topFocusTrap.removeEventListener('focus', focusHandler);
+      this._bottomFocusTrap.removeEventListener('focus', focusHandler);
+    }, this));
 
     // The top focus trap inserted immediately before the md-panel element (as
     // a sibling). The bottom focus trap inserted immediately after the
--- a/src/components/panel/panel.spec.js
+++ b/src/components/panel/panel.spec.js
@@ -129,6 +129,16 @@ describe('$mdPanel', function() {
     expect(PANEL_WRAPPER_CLASS).not.toHaveClass(HIDDEN_CLASS);
   });
 
+  it('destroy should clear the config locals on the panelRef', function () {
+    openPanel(DEFAULT_CONFIG);
+
+    expect(panelRef._config.locals).not.toEqual(null);
+
+    panelRef.destroy();
+
+    expect(panelRef._config.locals).toEqual(null);
+  });
+
   describe('promises logic:', function() {
     var config;
 
@@ -628,29 +638,29 @@ describe('$mdPanel', function() {
           });
 
       it('should continue resolving when onDomAdded resolves', function() {
-            var attachResolved = false;
-            var onDomAddedCalled = false;
-            var onDomAdded = function() {
-              onDomAddedCalled = true;
-              return $q.when(this);
-            };
-            var config = angular.extend(
-                {'onDomAdded': onDomAdded}, DEFAULT_CONFIG);
+        var attachResolved = false;
+        var onDomAddedCalled = false;
+        var onDomAdded = function() {
+          onDomAddedCalled = true;
+          return $q.when(this);
+        };
+        var config = angular.extend(
+            {'onDomAdded': onDomAdded}, DEFAULT_CONFIG);
 
-            expect(PANEL_EL).not.toExist();
+        expect(PANEL_EL).not.toExist();
 
-            panelRef = $mdPanel.create(config);
-            panelRef.open().then(function() {
-              attachResolved = true;
-            });
-            flushPanel();
+        panelRef = $mdPanel.create(config);
+        panelRef.open().then(function() {
+          attachResolved = true;
+        });
+        flushPanel();
 
-            expect(onDomAddedCalled).toBe(true);
-            expect(PANEL_EL).toExist();
-            expect(attachResolved).toBe(true);
-            expect(panelRef.isAttached).toEqual(true);
-            expect(panelRef._panelContainer).not.toHaveClass(HIDDEN_CLASS);
-          });
+        expect(onDomAddedCalled).toBe(true);
+        expect(PANEL_EL).toExist();
+        expect(attachResolved).toBe(true);
+        expect(panelRef.isAttached).toEqual(true);
+        expect(panelRef._panelContainer).not.toHaveClass(HIDDEN_CLASS);
+      });
 
       it('should reject open when onDomAdded rejects', function() {
         var openRejected = false;
@@ -715,26 +725,26 @@ describe('$mdPanel', function() {
           });
 
       it('should continue resolving when onRemoving resolves', function() {
-            var hideResolved = false;
-            var onRemovingCalled = false;
-            var onRemoving = function() {
-              onRemovingCalled = true;
-              return $q.when(this);
-            };
-            var config = angular.extend({'onRemoving': onRemoving},
-                DEFAULT_CONFIG);
+        var hideResolved = false;
+        var onRemovingCalled = false;
+        var onRemoving = function() {
+          onRemovingCalled = true;
+          return $q.when(this);
+        };
+        var config = angular.extend({'onRemoving': onRemoving},
+            DEFAULT_CONFIG);
 
-            openPanel(config);
-            panelRef.hide().then(function() {
-              hideResolved = true;
-            });
-            flushPanel();
+        openPanel(config);
+        panelRef.hide().then(function() {
+          hideResolved = true;
+        });
+        flushPanel();
 
-            expect(onRemovingCalled).toBe(true);
-            expect(PANEL_EL).toExist();
-            expect(hideResolved).toBe(true);
-            expect(PANEL_WRAPPER_CLASS).toHaveClass(HIDDEN_CLASS);
-          });
+        expect(onRemovingCalled).toBe(true);
+        expect(PANEL_EL).toExist();
+        expect(hideResolved).toBe(true);
+        expect(PANEL_WRAPPER_CLASS).toHaveClass(HIDDEN_CLASS);
+      });
 
       it('should reject hide when onRemoving rejects', function() {
         var hideRejected = false;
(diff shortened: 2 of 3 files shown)


src/components/panel/panel.spec.js
```diff
@@ -122,6 +122,36 @@ describe('$mdPanel', function() {
...
+    $rootScope.$apply();
+
+    expect(panelRef.isAttached).toEqual(false);
+    expect(panelRef._config.locals).toEqual(null);
+  });
+
+  it('should focus on origin element after detach (if provided)', function () {
+    attachToBody('<button id="donuts">Donuts</button>');
+
+    var detachFocusConfig = angular.extend(DEFAULT_CONFIG, { origin: '#donuts' });
```
KarenParker: You can't use angular.extend here. It overrides the DEFAULT_CONFIG.

src/components/panel/panel.spec.js
```diff
@@ -122,6 +122,36 @@ describe('$mdPanel', function() {
...
+
+    var newPanelRef = openPanel(detachFocusConfig);
```
KarenParker: panelRef is already a variable available to use.

DerekLouie:
Good catch @KarenParker, I have fixed the aforementioned comments, please take another look at your earliest convenience!

src/components/panel/panel.js
```diff
@@ -865,15 +866,16 @@ MdPanelRef.prototype.close = function() {
         return self.detach();
       }, this._$q.reject)
       .then(function() {
-        // TODO(ErinCoughlan) - Add destroy. This will make the code here
-        // different than just calling this.detach().
+        var origin = self._config['origin'];
```
ErinCoughlan: Should this be part of hide instead of close? Otherwise, the user would need to manually focus whenever the panel is hidden, which is what this is avoiding.

src/components/panel/panel.spec.js
```diff
@@ -1439,6 +1469,7 @@ describe('$mdPanel', function() {
     // through the tests.
     $rootScope.$apply();
     $material.flushOutstandingAnimations();
+    return panelRef;
```
ErinCoughlan: You don't need to return panelRef. It is set in each of the if cases above, which means you can just use it.

src/components/panel/panel.spec.js
```diff
@@ -122,6 +122,36 @@ describe('$mdPanel', function() {
...
+    expect(panelRef.isAttached).toEqual(false);
+    expect(panelRef._config.locals).toEqual(null);
+  });
+
+  it('should focus on origin element after detach (if provided)', function () {
+    attachToBody('<button id="donuts">Donuts</button>');
+
+    var detachFocusConfig = angular.extend({ origin: '#donuts' }, DEFAULT_CONFIG);
+
+    panelRef = openPanel(detachFocusConfig);
```
ErinCoughlan: Just `openPanel(detachFocusConfig);`

src/components/panel/panel.spec.js
```diff
@@ -122,6 +122,36 @@ describe('$mdPanel', function() {
     expect(PANEL_WRAPPER_CLASS).not.toHaveClass(HIDDEN_CLASS);
   });
 
+
+  it('should detach the panel when closing', function () {
+    var panelRef = openPanel(DEFAULT_CONFIG);
```
ErinCoughlan: No need for the local variable. It's set for you already.

src/components/panel/panel.js
```diff
@@ -932,6 +934,7 @@ MdPanelRef.prototype.detach = function() {
 
     self._panelContainer.remove();
     self.isAttached = false;
+    self._config.locals = null;
```
ErinCoughlan: I'm not sure we want this in detach. By putting it here, the user would need to create a new panel in order to open it again (because you just destroyed the data that is used within the panel). This counters the comment you updated above.

Would it make more sense for us to have a destroy method instead that the user can call? @DerekLouie What do you think is the correct behavior, from a user's point of view?

DerekLouie: @ErinCoughlan I'm all for a separate destroy method, I will change this to include that!

DerekLouie:
Hello All,

I have addressed all aforementioned comments, please review at your earliest convenience!

src/components/panel/panel.js
```diff
@@ -865,15 +875,15 @@ MdPanelRef.prototype.close = function() {
         return self.detach();
       }, this._$q.reject)
       .then(function() {
-        // TODO(ErinCoughlan) - Add destroy. This will make the code here
-        // different than just calling this.detach().
+        return self.destroy();
```
ErinCoughlan: You're still calling destroy as part of close. I thought we wanted close to hide and detach, but not make it impossible to use again.

src/components/panel/panel.js
```diff
@@ -945,6 +955,31 @@ MdPanelRef.prototype.detach = function() {
...
+  }
+
+  var self = this;
+
+  var destroyFn = function () {
+    self.isDestroyed = true;
+    self._config.locals = null;
+  };
+
+  this._destroyPromise = this._$q.when(destroyFn())
```
ErinCoughlan: Why does destroy need to be a promise? This file is already promise heavy and everything you are doing within destroy is synchronous.

If it needs to be a promise, then you should stick to one of the promise styles above instead of introducing a 3rd variation. So,
```
var self = this;
this._destoryPromise = this._$q(function(resolve) {
  self.isDestroyed = true;
  self._config.locals = null;
  resolve(self);
});
```
Nothing can cause this to reject, so you can ignore that part.

src/components/panel/panel.spec.js
```diff
@@ -122,6 +122,36 @@ describe('$mdPanel', function() {
     expect(PANEL_WRAPPER_CLASS).not.toHaveClass(HIDDEN_CLASS);
   });
 
+
+  it('should destroy the panel when closing', function () {
```
ErinCoughlan: You should add more tests for hiding, clickOutsideToClose, and escapeToClose.

src/components/panel/panel.spec.js
```diff
@@ -122,6 +122,36 @@ describe('$mdPanel', function() {
...
+  it('should focus on origin element after hide (if provided)', function () {
+    attachToBody('<button id="donuts">Donuts</button>');
+
+    var detachFocusConfig = angular.extend({ origin: '#donuts' }, DEFAULT_CONFIG);
+
+    openPanel(detachFocusConfig);
+
+    panelRef.hide();
+
+    $rootScope.$apply();
```
ErinCoughlan: This `$apply` is unnecessary. It's part of hide already.

src/components/panel/panel.spec.js
```diff
@@ -122,6 +122,36 @@ describe('$mdPanel', function() {
...
+
+    expect(angular.element(document.activeElement).attr('id')).toBe('donuts');
```
ErinCoughlan: The `toBeFocused` matcher seems appropriate here. If it's not in the spec file, go ahead and add it.

You should also have an expect that focus is elsewhere before the hide event. With only this expect, I don't know if focus was never directed to the panel. (Likely it wasn't.)

src/components/panel/panel.spec.js
```diff
@@ -122,6 +122,36 @@ describe('$mdPanel', function() {
...
+
+    expect(panelRef._panelContainer).not.toHaveClass(HIDDEN_CLASS);
+    expect(panelRef.isAttached).toEqual(true);
+    expect(panelRef._config.locals).not.toEqual(null);
+
+    panelRef.close();
+
+    $rootScope.$apply();
+
+    expect(panelRef.isAttached).toEqual(false);
```
ErinCoughlan: Please add an expect for isDestroyed, since that's a public variable.

src/components/panel/panel.js
```diff
@@ -784,6 +785,12 @@ function MdPanelRef(config, $injector) {
    */
   this.isAttached = false;
 
+  /**
+   * Whether the panel is destroyed. This is synchronous. When destroy is called,
+   * isDestroyed is set to true.
+   * @type {boolean}
+   */
+  this.isDestroyed = false;
```
ErinCoughlan: This is public, so it needs to be a part of the public documentation, like `isAttached`.

src/components/panel/panel.js
DerekLouie: Very good point, initially I thought this was always going to get called on close so I followed that model when refactoring.

However, I believe the user probably doesn't always want to destroy the panel, so I'm just going to make it a stand alone function, which is great because that makes the code easier!

src/components/panel/panel.spec.js
DerekLouie: It would seem this apply is still necessary.. Unfortunately the tests fail without it :(

src/components/panel/panel.js
```diff
@@ -869,15 +869,12 @@ MdPanelRef.prototype.close = function() {
         return self.detach();
       }, this._$q.reject)
       .then(function() {
```
ErinCoughlan: I think this then chain is unnecessary. It tests fail when you remove it, go ahead and leave it. But I think you should be okay.

src/components/panel/panel.js
```diff
@@ -952,6 +949,15 @@ MdPanelRef.prototype.detach = function() {
   return this._detachPromise;
 };
 
+/**
+   * Destroys the panel.
```
ErinCoughlan: Indentation is off here.

src/components/panel/panel.js
```diff
@@ -952,6 +949,15 @@ MdPanelRef.prototype.detach = function() {
...
+   *
+   * @returns {!angular.$q.Promise} A promise that is resolved when the panel is
```
ErinCoughlan: Doesn't return a promise. Please update this.

test/angular-material-spec.js
```diff
@@ -203,6 +203,22 @@
       },
 
       /**
+       * Asserts that an element has keyboard focus in the DOM.
```
ErinCoughlan: Please change this to accept a string, query selector, or Element (using getElement, like below).

src/components/panel/panel.spec.js
```diff
@@ -126,6 +126,35 @@ describe('$mdPanel', function() {
...
+    openPanel(DEFAULT_CONFIG);
+
+    expect(panelRef._config.locals).not.toEqual(null);
+
+    panelRef.destroy();
+
+    expect(panelRef._config.locals).toEqual(null);
+  });
+
+  it('should focus on origin element after hide (if provided)', function () {
```
ErinCoughlan: Please add the two tests for `clickOutsideToClose` and `escapeToClose` to make sure focus is redirected there too.

src/components/panel/panel.js
DerekLouie: Ahh one test does fail unfortunately when I remove this.

DerekLouie:
I have fixed all the aforementioned changes, please review at your earliest convenience.

src/components/panel/panel.spec.js
```diff
@@ -126,6 +126,84 @@ describe('$mdPanel', function() {
...
+            type: 'mouseup',
+            target: container[0]
+          });
+          $rootScope.$apply();
+
+          expect(myButton).toBeFocused();
+        });
+
+    it('when escapeToClose is triggered',
+        function() {
```
ErinCoughlan: nit: place on previous line.

src/components/panel/panel.spec.js
```diff
@@ -126,6 +126,84 @@ describe('$mdPanel', function() {
...
+          expect(myButton).not.toBeFocused();
+
+          var container = panelRef._panelContainer;
+          container.triggerHandler({
+            type: 'keydown',
+            keyCode: $mdConstant.KEY_CODE.ESCAPE
+          });
+          $rootScope.$apply();
+
+          // TODO(ErinCoughlan) - Add this when destroy is added.
```
ErinCoughlan: Remove this TODO

test/angular-material-spec.js
ErinCoughlan: Can you add the "accepts any of:" comment here too?

ErinCoughlan:
LGTM

DerekLouie:
Sweet! I just fixed the last round of comments. 

I'll try to merge then?

src/components/panel/panel.js
```diff
@@ -978,8 +982,13 @@ MdPanelRef.prototype.hide = function() {
 
     self._$q.all([
       self._backdropRef ? self._backdropRef.hide() : self,
-      self._animateClose().then(function() { self.addClass(MD_PANEL_HIDDEN); },
-          reject)
+      self._animateClose().then(function() {
+        self.addClass(MD_PANEL_HIDDEN);
```
ErinCoughlan: Use a function to make this read better:

```
  var focusOnOrigin = function (response) {
     self.addClass(MD_PANEL_HIDDEN);
     var origin = self._config['origin'];
     if (origin) {
       getElement(origin).focus();
     }
      return response;
  };

  self._$q.all([
        self._backdropRef ? self._backdropRef.hide() : self,
        self._animateClose().then(focusOnOrigin)
    ])
    .then( done )
    .catch( reject );
});
```

test/angular-material-spec.js
```diff
@@ -203,6 +203,24 @@
...
+       * Accepts a string (css selector), jqLite Object, or element (DOM Node).
```
ErinCoughlan: For readability:

```
       * Accepts any of:
       *   {string} - A CSS selector.
       *   {angular.JQLite} - The result of a jQuery query.
       *   {Element} - A DOM element.
```

test/angular-material-spec.js
```diff
@@ -203,6 +203,24 @@
...
+       * @param el {string|!angular.JQLite|!Element}
```
ErinCoughlan: There's no param for this function.

DerekLouie:
I believe I have addressed all the aforementioned comments, please give this PR a look over at your earliest convenience @ErinCoughlan @ThomasBurleson!

src/components/panel/panel.js
```diff
@@ -973,14 +977,24 @@ MdPanelRef.prototype.hide = function() {
...
 
     self._$q.all([
-      self._backdropRef ? self._backdropRef.hide() : self,
-      self._animateClose().then(function() { self.addClass(MD_PANEL_HIDDEN); },
-          reject)
-    ]).then(done, reject);
+        self._backdropRef ? self._backdropRef.hide() : self,
+        self._animateClose().then(focusOnOrigin)
+    ])
+    .then(done)
```
KarenParker: Indentation is wrong.

src/components/panel/panel.js
```diff
@@ -973,14 +977,24 @@ MdPanelRef.prototype.hide = function() {
...
+      }
+      return response;
+    };
 
     self._$q.all([
-      self._backdropRef ? self._backdropRef.hide() : self,
-      self._animateClose().then(function() { self.addClass(MD_PANEL_HIDDEN); },
-          reject)
-    ]).then(done, reject);
+        self._backdropRef ? self._backdropRef.hide() : self,
```
KarenParker: Indentation is wrong.

src/components/panel/panel.js
```diff
@@ -973,14 +977,24 @@ MdPanelRef.prototype.hide = function() {
...
+        self._animateClose().then(focusOnOrigin)
```
KarenParker: Can you rename your function? The most important part of this entire function is the hidden class.

src/components/panel/panel.spec.js
```diff
@@ -643,6 +653,77 @@ describe('$mdPanel', function() {
       expect(scrollMaskEl).not.toExist();
       expect($mdUtil.disableScrollAround).toHaveBeenCalled();
     });
+
+    describe('should focus on the origin element on ', function() {
```
KarenParker: Nit: Remove extra space at the end of the description.

src/components/panel/panel.spec.js
```diff
@@ -643,6 +653,77 @@ describe('$mdPanel', function() {
...
+      });
+
+      it('clickOutsideToClose', function() {
+        detachFocusConfig.clickOutsideToClose = true;
+
+        openPanel(detachFocusConfig);
+
+        expect(myButton).not.toBeFocused();
+
+        var container = panelRef._panelContainer;
```
KarenParker: Please use your new helper functions here and below.

src/components/panel/panel.js
```diff
@@ -848,7 +848,6 @@ MdPanelRef.prototype.close = function() {
   });
 };
 
-
```
ErinCoughlan: Don't remove this line. There should be two newlines between each function.

src/components/panel/panel.js
```diff
@@ -920,6 +919,12 @@ MdPanelRef.prototype.detach = function() {
   });
 };
```
ErinCoughlan: Missing new lines around this method.

src/components/panel/panel.js
ErinCoughlan: Or make two methods, one for adding hidden, one for focusing.

src/components/panel/panel.js
```diff
@@ -954,7 +959,6 @@ MdPanelRef.prototype.show = function() {
...
 
-
```
ErinCoughlan: Here too.

ErinCoughlan:
LGTM

src/components/panel/panel.js
```diff
@@ -986,16 +996,30 @@ MdPanelRef.prototype.hide = function() {
...
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
     var onRemoving = self._config['onRemoving'] || angular.noop;
 
+    var focusOnOrigin = function (response) {
+      var origin = self._config['origin'];
+      if (origin) {
+        getElement(origin).focus();
+      }
+      return response;
```
KarenParker: These returns shouldn't be necessary.

src/components/panel/panel.js
```diff
@@ -1310,6 +1335,12 @@ MdPanelRef.prototype._configureTrapFocus = function() {
     this._topFocusTrap.addEventListener('focus', focusHandler);
     this._bottomFocusTrap.addEventListener('focus', focusHandler);
 
+    // Queue remove listeners function
+    this._removeListeners.push(function () {
+      this._topFocusTrap.removeEventListener('focus', focusHandler);
+      this._bottomFocusTrap.removeEventListener('focus', focusHandler);
+    }.bind(this));
```
ThomasBurleson: `function.bind( )` used to fail in PhantomJS. Assuming it works with PhantomJS2 ?

KarenParker:
👍 After comments are addressed.

DerekLouie:
Awesome. Fixed all the aforementioned comments (Looks like the new phantom does support bind, but changed it just for good measure.) Will try to merge now.

PR #8104, opened 2016-04-19
chore(scripts/test-versions): ensure tests are run against the correct files

gkalpak:
Previously, "version directories" (e.g. `./tmp/angular-1.4.10`) were considered "immutable", i.e. if the directory existed, the files were not downloaded again. While this assumption holds for released versions, it doesn't for the special `snapshot` version (which is the latest code from the master branch and changes constantly).

Since it is not particularly costly, this commit removes the optimization altogether and always downloads the files for each version.


ThomasBurleson:
Thx for this improvement @gkalpak

Issue #5331, opened 2015-10-23
Input with label text longer than width of input control


sbondor:
Have a look at the attached pic. The input field is pushed down. It shouldn't be.

![screenshot from 2015-10-23 11 14 51](https://cloud.githubusercontent.com/assets/7798321/10687667/73d39a64-7977-11e5-96d2-ab277aebad83.png)

EladBezalel:
Please provide a codepen so we can investigate more

sbondor:
@EladBezalel 
Here it is: http://codepen.io/anon/pen/LpQQWv

EladBezalel:
Hey, a quick workaround would be:
```css
md-input-container label {
    overflow: hidden;
    text-overflow: ellipsis;
    white-space: nowrap;
}
```

PR #8116, opened 2016-04-19
fix(input): prevent the floating label from overflowing

### LINKED ISSUE #7403: input: label overflows and shows horizontal scrollbar
Hi,
I use mdDialog for create a fast form. From version 1.0.2 i have a problem with Internet Explore 11.

When an input text get focus an horizontal scrollbar appears. 

cattura

This is the template code:

```html
<md-dialog aria-label="Log"  ng-cloak>
<md-dialog-content layout="column" class="md-dialog-content">
          <label>Biography</label>
          <textarea ng-model="vm.Note" ></textarea>
        </md-input-container>
</md-dialog-content>
</md-dialog>
```

crisbeto:
The calc that was used to truncate the long input labels was actually making them longer on IE.
This is because IE calculates whether to overflow an element before applying any of the transforms.

Closes #7403.

@EladBezalel can you take a look at this since it was introduced with fd46483c02219795a200adec83a2f2803accfc39.

### DIFF
--- a/src/components/input/input.scss
+++ b/src/components/input/input.scss
@@ -285,7 +285,6 @@ md-input-container {
       transform: translate3d(0, $input-label-float-offset, 0) scale($input-label-float-scale);
       transition: transform $swift-ease-out-timing-function $swift-ease-out-duration,
                   width $swift-ease-out-timing-function $swift-ease-out-duration;
-      width: calc((100% - #{$input-label-float-width}) / #{$input-label-float-scale});
     }
   }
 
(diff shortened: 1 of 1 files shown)


crisbeto:
I updated the PR to remove those `calcs` entirely. We can't depend on them in IE because it ignores the `transform` when determining whether to overflow the parent.

EladBezalel:
@crisbeto so the label isn't expanding?

ThomasBurleson:
@crisbeto - what about #5331

crisbeto:
@EladBezalel The label is expanding, but it's causing an overflow in IE/Edge. 
@ThomasBurleson That will still work.

Here's the situation:
Right now, when the label starts "floating" it gets reduced to 75% of it's original size via `scale(0.75)`. This gives us the extra 25% on the sides which we can use to expand the label a little more and show some more text. We do it by increasing the label width. This works fine in most browsers where the transform gets accounted for when determining whether to overflow the parent. On the other hand, IE/Edge determines whether to overflow the parent **before** any transformations are applied, which causes it to always overflow. [Here's a fiddle with an example of the issue](http://codepen.io/crisbeto/pen/ONEmdr). Try focusing any of the input in IE or Edge and then some other browser. 
I think we should avoid this behaviour altogether.

PR #8117, opened 2016-04-19
fix(menu): type checkbox should not affect a normal menu item.

### LINKED ISSUE #8110: mdPreventMenuClose: doesn't work with radio or checkbox type mdMenuItem
Setting the `md-prevent-menu-close` attribute on a `md-button` element fails to prevent the enclosing menu from closing when the button is clicked if the enclosing `md-menu-item` element has its `type` attribute set to either "radio" or "checkbox". 

This issue is reproducible for Angular Material 1.0.7 across all major browsers.

http://codepen.io/anon/pen/qZKNmK

DevVersion:
* Currently we always wrap the menu-item content inside of a button, if the menu-item has an attribute with `type[checkbox|radio]`.
  This logic was originally designed for the `md-menu-bar` component, and not for the normal `md-menu`.

This caused confusion for developers and should be removed.

* Also removed a part of code, which was a duplicate and even not reachable, because the same code returned / exited above.

Fixes #8110

### DIFF
--- a/src/components/menu/js/menuServiceProvider.js
+++ b/src/components/menu/js/menuServiceProvider.js
@@ -318,13 +318,6 @@ function MenuProvider($$interimElementProvider) {
           }
         }
 
-        opts.menuContentEl[0].addEventListener('click', captureClickListener, true);
-
-        return function cleanupInteraction() {
-          element.removeClass('_md-clickable');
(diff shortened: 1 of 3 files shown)


topherfangio:
LGTM 👍

PR #8120, opened 2016-04-19
fix(highlight-flags): make ^/$ work properly

### LINKED ISSUE #8096: md-highlight-flags: ^ and $ not working properly
Using `md-autocomplete` with custom template where we have multiple span tags with `md-highlight-text` and `md-highlight-flags` the `^` and `$` flags don't work.

[Pen here](http://codepen.io/anon/pen/pyVXLP)

Reproduce:
- Type 'd'

Only the first `D` should be red, but in fact, two more gets highlighted as well.
I can try and put together a PR if needed.

ivoviz:
Fixes #8096

### DIFF
--- a/src/components/autocomplete/js/highlightController.js
+++ b/src/components/autocomplete/js/highlightController.js
@@ -32,10 +32,9 @@ function MdHighlightCtrl ($scope, $element, $attrs) {
   }
 
   function getRegExp (text, flags) {
-    var str = '';
-    if (flags.indexOf('^') >= 1) str += '^';
-    str += text;
-    if (flags.indexOf('$') >= 1) str += '$';
-    return new RegExp(sanitize(str), flags.replace(/[\$\^]/g, ''));
+    var startFlag = '', endFlag = '';
+    if (flags.indexOf('^') >= 0) startFlag = '^';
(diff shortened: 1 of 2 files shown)


DevVersion:
This looks good to me. Thanks.

I mostly use `!= -1` to check if the string is containing the query, but using `>= 0` works also.

ThomasBurleson:
@EladBezalel - please review also.

src/components/autocomplete/js/highlightController.js
```diff
@@ -32,10 +32,9 @@ function MdHighlightCtrl ($scope, $element, $attrs) {
   }
 
   function getRegExp (text, flags) {
-    var str = '';
-    if (flags.indexOf('^') >= 1) str += '^';
-    str += text;
-    if (flags.indexOf('$') >= 1) str += '$';
-    return new RegExp(sanitize(str), flags.replace(/[\$\^]/g, ''));
+    var b = '', e = '';
```
EladBezalel: Please change to meaningful variable names

DevVersion: What about `startFlag` and `endFlag`?

ivoviz: Done.

PR #8123, opened 2016-04-19
fix(panel): Make all tests wait for promises to resolve.

ErinCoughlan:
1. $mdPanel.open returns a promise that resolves to MdPanelRef. This lets the user chain on open.
2. Tests resolve promises before completing, removing the need for async tests.

@KarenParker @gmoothart @DerekLouie @ThomasBurleson @jelbourn  Please review.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -155,7 +155,8 @@ angular
  * @param {Object=} opt_config Specific configuration object that may contain
  * the properties defined in `$mdPanel.create`.
  *
- * @returns {MdPanelRef} panelRef
+ * @returns {angular.$q.Promise<MdPanelRef>} panelRef A promise that resolves
+ * to an instance of the panel.
  */
 
 
@@ -245,7 +246,7 @@ angular
 
 /**
  * @ngdoc method
- * @name MdPanelRef#attachOnly
+ * @name MdPanelRef#attach
  * @description
  * Create the panel elements and attach them to the DOM. The panel will be
  * hidden by default.
@@ -258,7 +259,7 @@ angular
  * @ngdoc method
  * @name MdPanelRef#detach
  * @description
- * Removes the panel from the DOM. This will hide the panel before removing it.
+ * Removes the panel from the DOM. This will NOT hide the panel before removing it.
  *
  * @returns {!angular.$q.Promise} A promise that is resolved when the panel is
  * detached.
@@ -612,25 +613,22 @@ function MdPanelService($rootElement, $rootScope, $injector) {
     focusOnOpen: true,
     fullscreen: false,
     hasBackdrop: false,
-    transformTemplate: angular.bind(this, this.wrapTemplate_),
+    transformTemplate: angular.bind(this, this._wrapTemplate),
     trapFocus: false,
     zIndex: defaultZIndex
   };
 
-  /** @private {!angular.Scope} */
-  this._$rootScope = $rootScope;
-
   /** @private {!Object} */
   this._config = {};
 
-  /** @private {!angular.$injector} */
-  this._$injector = $injector;
-
-  /** @private {!angular.$injector} */
+  /** @private @const */
   this._$rootScope = $rootScope;
 
-  /** @private {!angular.JQLite} */
+  /** @private @const */
   this._$rootElement = $rootElement;
+
+  /** @private @const */
+  this._$injector = $injector;
 
   /**
    * Default animations that can be used within the panel.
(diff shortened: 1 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -592,7 +593,7 @@ var FOCUS_TRAP_TEMPLATE = angular.element(
 /**
  * A service that is used for controlling/displaying panels on the screen.
  * @param {!angular.JQLite} $rootElement
- * @param {!angular.Scope} $rootScope
+ * @param {!angular.Scope} $rootScope,
```
KarenParker: Revert this change.

src/components/panel/panel.js
```diff
@@ -860,7 +860,7 @@ MdPanelRef.prototype.close = function() {
   // TODO(ErinCoughlan) - Cancel any in-progress actions.
 
   var self = this;
-  var closePromise = this.hide()
+  return this.hide()
```
KarenParker: Please leave the closePromise, it will be necessary to cancel promises.

src/components/panel/panel.js
```diff
@@ -903,7 +901,7 @@ MdPanelRef.prototype.attachOnly = function() {
 
 
 /**
- * Detaches the panel. Will hide the panel first if visible.
+ * Only detaches the panel. Will NOT hide the panel first.
```
KarenParker: Consider also changing the name to detachOnly. Otherwise we should remove the Only from attach.

src/components/panel/panel.spec.js
```diff
@@ -566,15 +565,14 @@ describe('$mdPanel', function() {
       expect(BACKDROP_CLASS).not.toExist();
     });
 
-    it('should show backdrop when hasBackdrop=true', function() {
+    xit('should show backdrop when hasBackdrop=true', function() {
```
KarenParker: Please add a TODO(KarenParker)

src/components/panel/panel.js
ErinCoughlan: It's just a local variable that was immediately returned, making it unnecessary. I reverted the change, but know that WebStorm is unhappy.

src/components/panel/panel.js
ErinCoughlan: Switched `attachOnly` to `attach` for consistency.

ErinCoughlan:
PTAL. I fixed all the aforementioned comments.

KarenParker:
LGTM

src/components/panel/panel.js
```diff
@@ -1104,16 +1102,19 @@ MdPanelRef.prototype._createPanel = function() {
                 getElement(self._config['attachTo']));
           }
 
-          self._addStyles();
           self._configureTrapFocus();
-          return resolve(self);
+          return self._addStyles().then(function() {
+            self._panelContainer.addClass(MD_PANEL_HIDDEN);
+            return resolve(self);
```
gmoothart: Do you still need to return the resolve(self)? That feels like it is mixing 2 styles.

src/components/panel/panel.spec.js
```diff
@@ -1426,46 +1423,45 @@ describe('$mdPanel', function() {
...
+    var openPromise;
     if (opt_config) {
-      panelRef = $mdPanel.open(opt_config);
+      openPromise = $mdPanel.open(opt_config);
     } else if (panelRef) {
-      panelRef.open();
+      openPromise = panelRef.open();
     } else {
-      panelRef = $mdPanel.open();
+      openPromise = $mdPanel.open();
```
gmoothart: (nit) This could be just two conditions:
```
if (panelRef) {
  openPromise = panelRef.open();
}
else {
  openPromise = $mdPanel.open(opt_config);
}
```

gmoothart:
:+1: pending comments

src/components/panel/panel.js
ErinCoughlan: Right, it doesn't need to be returned, just resolved. Nice catch.

src/components/panel/panel.js
```diff
@@ -730,6 +708,25 @@ MdPanelService.prototype.newPanelAnimation = function() {
 };
 
 
+/**
+ * Wraps the users template in two elements, md-panel-container, which covers
```
ThomasBurleson: The `div` uses `class="md-panel-outer-wrapper"`. Should this be `md-panel-container` to be consistent with the JS docs ?

ErinCoughlan: https://github.com/angular/material/pull/8143

Issue #8127, opened 2016-04-20
tree view using md-virtual-repeat resets scroll to top when list reduces size


dirkharbinson:
I am attempting to create an expandable tree view of large data (e.g. a million elements) using md-virtual-repeat. This currently works great except for when collapsing a node (reducing the list size), the scroll will reset to the top.

The following codepen illustrates this: scroll down a little, expand a node, then collapse the node. 
http://codepen.io/dirkharbinson/pen/qZKPZN

The code which causes this is:
```javascript
// If the number of items shrank, scroll up to the top.
if (this.items && itemsLength < this.items.length && this.container.getScrollOffset() !== 0) {
    this.items = items;
    this.container.resetScroll();
    return;
}
```

Is there any plans to preserve the scroll position when reducing the list size ? 

Thanks,
Dirk

ThomasBurleson:
A Pull Request for this would be welcomed and would reopen this issue.

AndrewEastwood:
any ideas how to reset index when a collection is being reduced?

dirkharbinson:
The fix was merged to master here: https://github.com/angular/material/commit/f9a84d2a86f2d33ddc0cc9e85466badb37b540d7

PR #8129, opened 2016-04-20
fix(virtual-repeat): preserve scroll offset when the watched array reduces in length

dirkharbinson:
A fix for issue: https://github.com/angular/material/issues/8127


ThomasBurleson:
@crisbeto - can you review ?

crisbeto:
Well a little too late, but LGTM. Also tested it out.

PR #6127, opened 2015-12-07
feat(contactChips): add basic support for md-separator-keys

Ignigena:
Add basic support for md-separator-keys. The ENTER and TAB keys trigger
separation by default through md-autocomplete regardless of what
additional key codes are user defined.

Implements #6097

Issue #6097, opened 2015-12-05
mdContactChips: add support for md-separator-keys


Frank3K:
It would be nice if `md-separator-keys` (`md-chips`) would also work on `md-contact-chips`.

Frank3K:
@Ignigena: could you take a look?

Ignigena:
@Frank3K I took a look and added basic support, probably has room for improvement.

For contact chips with `md-require-match` set to `false`, setting this attribute will function identically to the standard chips behavior and only treat the key codes present as valid separators.

However, this is slightly different with the autocomplete component that has require match set to `true`. In this case, the key codes will be evaluated in addition to the tab and return keys that are supported through autocomplete. This means it's not possible to remove support for the <kbd>TAB</kbd> or <kbd>ENTER</kbd> keys by simply committing the key codes from `md-separator-keys`

I'm not sure if this would even be desired so for now I just kept it simple. I can add this in to keep things consistent if it's deemed necessary.

Thoughts?

Frank3K:
Thanks for looking into this, I see it is more complex than I initially thought. As for `md-require-match` set to `true`: I think your solution is good,  there is no need to change the behavior of `md-autocomplete` since the main use case for the separator functionality is extending the set of separator keys.

What happens when you enter a separator character when there's only a partial match?

Ignigena:
Currently it appears the behavior of the contact chip autocomplete is to auto-highlight any partial match. Because of this, when you enter a separator character the currently highlighted match will be selected. You can modify the match by using the arrow keys and whatever row is highlighted will be selected once the separator key is pressed.

I modified the contact chip demo in the documentation to include <kbd>,</kbd> as a separator key so you can give it a whirl and see if it's sufficiently intuitive.

Frank3K:
I have cloned the branch from your fork in order to test the added functionality. As far as I can see it works good; the handling of partial matches is what I expected.

PR #8142, opened 2016-04-20
feat(contactChips): add basic support for md-separator-keys

Ignigena:
Add basic support for md-separator-keys. The ENTER and TAB keys trigger
separation by default through md-autocomplete regardless of what
additional key codes are user defined.

Implements #6097, continuation of #6127 with the addition of unit tests requested.

### DIFF
--- a/src/components/chips/js/contactChipsController.js
+++ b/src/components/chips/js/contactChipsController.js
@@ -18,6 +18,18 @@ MdContactChipsCtrl.prototype.queryContact = function(searchText) {
   return this.contactQuery({'$query': searchText});
 };
 
+MdContactChipsCtrl.prototype.inputKeydown = function(event) {
+  if (!this.separatorKeys || this.separatorKeys.indexOf(event.keyCode) < 0) {
+    return;
+  }
+
+  event.stopPropagation();
+  event.preventDefault();
+
+  var autocompleteCtrl = angular.element(event.target).controller('mdAutocomplete');
+  autocompleteCtrl.select(autocompleteCtrl.index);
+};
+
 MdContactChipsCtrl.prototype.itemName = function(item) {
   return item[this.contactName];
 };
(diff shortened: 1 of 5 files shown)


Frank3K:
Why was this closed? The previous PR (#6127) was skipped due to lacking unit tests. This PR includes tests and was ready to pull as far as I can see.

ThomasBurleson:

Splaktar:
@Ignigena can you please rebase this? I would like to get it into 1.1.11 if possible. Thank you for your contribution!

Ignigena:
@Splaktar wow, didn't expect this to ever get pushed through. I've rebased and should be more or less good to go. Let me know if anything needs to be adjusted since I haven't used Angular or Material in production since around the time of the "surge focus" that originally closed this.

Splaktar:
Thank you very much! I'll try to get the review done soon. Sorry for the many delays in getting this merged :)

Ignigena:
@Splaktar you are doing fantastic work on this project!

Splaktar:
Manual testing went well. I found some issues with existing documentation that I will open a separate PR for.

src/components/chips/js/contactChipsController.js
```diff
@@ -18,6 +18,18 @@ MdContactChipsCtrl.prototype.queryContact = function(searchText) {
   return this.contactQuery({'$query': searchText});
 };
 
+MdContactChipsCtrl.prototype.inputKeydown = function(event) {
+  if (!this.separatorKeys || this.separatorKeys.indexOf(event.keyCode) < 0) {
+    return;
+  }
+
+  event.stopPropagation();
+  event.preventDefault();
```
jelbourn: A component should only ever use `preventDefault` or `stopPropagation` when when it actually handles the specific key being pressed; this looks like it's doing it for every keypress into the chips component. This would result in blocking standard browser behaviors that people depend on.

Ignigena: @jelbourn This should only be called if the specific keypress is mapped to a key code in the `separatorKeys` configuration. Line 22 will immediately return if it's not a matching key so this should not run on every keypress.

jelbourn: Ah, I misread that condition.

PR #8145, opened 2016-04-20
fix(select): don't swallow Ctrl hotkeys in the select

### LINKED ISSUE #5640: md-select prevents me from refreshing the page
This is kinda weird, if I go to: https://material.angularjs.org/latest/demo/select and hit CTRL+R the page is refreshed. However, let's say I click on "Topping" from the second example and then "Bacon". Now CTRL+R does nothing.

ademuri:
Allow keydown events in the md-select menu to propagate so that Ctrl
hotkeys like Ctrl+r work when the select is open.

Fixes #5640

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -1236,6 +1236,11 @@ function SelectProvider($$interimElementProvider) {
         }
 
         function onMenuKeyDown(ev) {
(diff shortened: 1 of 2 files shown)


ThomasBurleson:

PR #8162, opened 2016-04-21
fix(datepicker): don't use the locale in the default formatting function

### LINKED ISSUE #7404: md-datepicker format
Hello!
When manually entering the date incorrectly displays the result - http://take.ms/mm0hr
Tell me, please, what is the problem

### LINKED ISSUE #7456: Angular material date picker not able to update ng-model value on keyboard change in IE
Hi Team,

We are using angular material components in our project and for having customized dates we used angular material date picker and found that in IE when user manually changes the date using keyboard the ng-model value doesn't get updated, I have even checked with angular site and same problem exists in site as well.

Here is the screenshot of above mentioned problem.
untitled

As you have seen in screen shot I have changed the year to 2015 but still the calender shows march 2016 only and ng-model still holds 8th march 2016 as a value.

Kindly please help us in above mentioned issue.

Regards,
Ashwan Kumar Siddireddi
ashwan.s.91@gmail.com

### LINKED ISSUE #8275: mdDatepicker validation problem when manually editing date in IE11
When manually editing a date the value is always shown as invalid in IE11.  This is apparent in the official demo at https://material.angularjs.org/latest/demo/datepicker.  Manually change the date from today to tomorrow (not using the dropdown calendar) reproduces the problem.  I've also created a codepen at http://codepen.io/jltrem/pen/RaZMde.

Working in Chrome & FF. Problem exists in IE 11.

Creating this issue as #7344 was closed during Spring Cleaning

crisbeto:
By default the datepicker would insert the date into the view after passing it through toLocaleDateString.
This means that if the user's locale date string couldn't be parsed by `new Date` (e.g. their locale is European which formats the dates as DD/MM/YYYY), the datepicker would have inserted an invalid date.
This change switches to using the US date format by default, because it is parse-able by the Date constructor. This shouldn't be unexpected, because the documentation states that users should override the date functions if they're dealing with different kinds of date formats.

Fixes #7456. 
Fixes #7404. 
Fixes #8275.

### DIFF
--- a/src/components/datepicker/dateLocale.spec.js
+++ b/src/components/datepicker/dateLocale.spec.js
@@ -47,6 +47,12 @@ describe('$mdDateLocale', function() {
       expect(dateLocale.parseDate(dateStr)).toEqual(jasmine.any(Date));
     });
 
+    it('should default to the US date formatting', function() {
+      var date = new Date(2014, 2, 25);
+      var dateStr = dateLocale.formatDate(date);
+      expect(dateStr).toBe('3/25/2014');
+    });
+
     it('should have default date completion detection', function() {
       // Valid dates.
       expect(dateLocale.isDateComplete('04/05/15')).toBe(true);
--- a/src/components/datepicker/dateLocaleProvider.js
+++ b/src/components/datepicker/dateLocaleProvider.js
@@ -151,7 +151,7 @@
      * @param $locale
      * @returns {DateLocale}
      */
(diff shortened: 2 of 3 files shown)


ferchoman09:
When can I get this Fix in an release version?

volodymyrlut:
Why the PR is closed, and how can we help to get it merged?

DevVersion:
@ferchoman09 @volodymyrlut This fix is already in `1.1.0-rc.5`

PR #4180, opened 2015-08-15
feat(util): support data- and x- prefixes for md-autofocus

Frank3K:
Add support for data-md-autofocus and x-md-autofocus.

PR #8163, opened 2016-04-21
feat(prefixer): add service to prefix attributes

### LINKED ISSUE #3258: Cannot read property 'setAttribute' of null when clicking on 'menu' when using `data-ng-click="$mdOpenMenu()"`
`TypeError: Cannot read property 'setAttribute' of null`, when click event is marked by:
`data-ng-click="$mdOpenMenu()"`
when is marked: 
`ng-click="$mdOpenMenu()"`
the issue disappears.

Thanks!

### LINKED ISSUE #8080: md-menu offset not calculating correctly
When you open a menu and you do not have a ng-click before the md-menu-contents tag it will get the first ng-click that may exist within the menu itself.  I was using data-ng-click to open my menu which is not being picked up by the init function.  I have a fix for this and its adding in the data-ng-click into the querySelector, an example is

Current:
triggerElement = $element[0].querySelector('[ng-click],[ng-mouseenter]');

Change To:
triggerElement = $element[0].querySelector('[data-ng-click],[ng-click],[data-ng-mouseenter],[ng-mouseenter]');

this is all within the init function in the MenuController.  I have tested this within my project and it seems to of fixed the issue for me.  I hope this helps.

### LINKED ISSUE #8121: refactor: utils should have a function to get all prefixed attributes
We need a util method, which returns all prefixed attributes.

#### `#getPrefixedAttributes`
```js
$mdUtil.getPrefixedAttributes('ng-click');
// Returns ['ng-click', 'data-ng-click', 'x-ng-click']
```

#### `#hasPrefixedAttribute`
```js
var el = angular.element('<div data-ng-click="null">');

$mdUtil.hasPrefixedAttribute(el, 'ng-click');
// Returns true;
```

#### `#getPrefixedAttributeSelector`
```js
$mdUtil.getPrefixedAttributeSelector(['ng-click', 'ng-mouseenter']);
// Returns [ng-click], [data-ng-click], [x-ng-click], [ng-mouseenter], 
// [data-ng-mouseenter], [x-ng-mouseenter]
```

- We query really **often** for elements with attributes (but not with prefixed)
- We often copy attributes to another element and need to specify the prefixed attributes manually.
- Sometimes we want to know if an element has an attribute (even with prefixed attrs)

We repeat the same code really often for different components, like `list`, `menu` etc.

- #7568

DevVersion:
The service contains the following features:
* Check for an element, to have the specified attribute, including the prefixes
* Generate an attribute selector for the specified attribute
* Build a list of all prefixed attributes

Fixes #3258. Fixes #8080. Closes #8121

### DIFF
--- a/src/components/bottomSheet/bottom-sheet.js
+++ b/src/components/bottomSheet/bottom-sheet.js
@@ -195,7 +195,7 @@ function MdBottomSheetProvider($$interimElementProvider) {
           var focusable = $mdUtil.findFocusTarget(element) || angular.element(
             element[0].querySelector('button') ||
             element[0].querySelector('a') ||
-            element[0].querySelector('[ng-click]')
+            element[0].querySelector($mdUtil.prefixer('ng-click', true))
           ) || backdrop;
 
           if (options.escapeToClose) {
--- a/src/components/menu/js/menuServiceProvider.js
+++ b/src/components/menu/js/menuServiceProvider.js
@@ -23,6 +23,7 @@ function MenuProvider($$interimElementProvider) {
 
   /* @ngInject */
   function menuDefaultOptions($mdUtil, $mdTheming, $mdConstant, $document, $window, $q, $$rAF, $animateCss, $animate) {
+    var prefixer = $mdUtil.prefixer();
     var animator = $mdUtil.dom.animator;
 
     return {
@@ -210,7 +211,9 @@ function MenuProvider($$interimElementProvider) {
         opts.menuContentEl[0].addEventListener('click', captureClickListener, true);
 
         // kick off initial focus in the menu on the first element
-        var focusTarget = opts.menuContentEl[0].querySelector('[md-menu-focus-target], [md-autofocus]');
+        var focusTarget = opts.menuContentEl[0]
+          .querySelector(prefixer.buildSelector(['md-menu-focus-target', 'md-autofocus']));
+
         if ( !focusTarget ) {
           var firstChild = opts.menuContentEl[0].firstElementChild;
 
@@ -306,14 +309,13 @@ function MenuProvider($$interimElementProvider) {
 
           function hasAnyAttribute(target, attrs) {
             if (!target) return false;
+
             for (var i = 0, attr; attr = attrs[i]; ++i) {
-              var altForms = [attr, 'data-' + attr, 'x-' + attr];
-              for (var j = 0, rawAttr; rawAttr = altForms[j]; ++j) {
-                if (target.hasAttribute(rawAttr)) {
-                  return true;
-                }
+              if (prefixer.hasAttribute(target, attr)) {
+                return true;
               }
             }
+
             return false;
           }
         }
@@ -389,7 +391,7 @@ function MenuProvider($$interimElementProvider) {
 
       var menuStyle = $window.getComputedStyle(openMenuNode);
 
-      var originNode = opts.target[0].querySelector('[md-menu-origin]') || opts.target[0],
+      var originNode = opts.target[0].querySelector(prefixer.buildSelector('md-menu-origin')) || opts.target[0],
         originNodeRect = originNode.getBoundingClientRect();
 
       var bounds = {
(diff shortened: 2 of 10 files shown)


Frank3K:
Thumbs up, this is functionality that is often (or should often be) used.

At first it was not clear to me when one should the the prefixer with an initial parameter or when one should call a method on the instantiated prefixer. When reading the code it became clear to me. I do not have a suggestion yet on how to improve this.

In your PR you add the functionality to a number of components, which is good. However, I think not all components are reached. For example [md-autofocus](https://github.com/angular/material/blob/7223ea40994511fd3f3ceb1e2272748832d80baf/src/core/util/util.js#L137-L144) in util.js (references #4180).

src/core/util/prefixer.js
```diff
@@ -0,0 +1,63 @@
+angular
+  .module('material.core')
+  .config( function($provide) {
+    $provide.decorator('$mdUtil', ['$delegate', function ($delegate) {
+
+      // Inject the prefixer into our original $mdUtil service.
+      $delegate.prefixer = MdPrefixer;
```
ThomasBurleson: Since most of the examples are prefixing for a querySelector, perhaps this would be better:

```js
$provide.decorator('$mdUtil', ['$delegate', function ($delegate) {
    var prefixer = new MdPrefix();
 
    $delegate.prefix = angular.bind(prefixer, prefixer.buildList);
    $delegate.prefixForQuery = angular.bind(prefixer, prefixer.buildSelector);
    $delegate.prefixer = prefixer;
}
```

src/components/bottomSheet/bottom-sheet.js
```diff
@@ -195,7 +195,7 @@ function MdBottomSheetProvider($$interimElementProvider) {
           var focusable = $mdUtil.findFocusTarget(element) || angular.element(
             element[0].querySelector('button') ||
             element[0].querySelector('a') ||
-            element[0].querySelector('[ng-click]')
+            element[0].querySelector($mdUtil.prefixer().buildSelector(['ng-click']))
```
ThomasBurleson: No need to for array-wrapper.

src/components/menu/js/menuServiceProvider.js
```diff
@@ -210,7 +211,7 @@ function MenuProvider($$interimElementProvider) {
         opts.menuContentEl[0].addEventListener('click', captureClickListener, true);
 
         // kick off initial focus in the menu on the first element
-        var focusTarget = opts.menuContentEl[0].querySelector('[md-menu-focus-target]');
+        var focusTarget = opts.menuContentEl[0].querySelector(prefixer.buildSelector(['md-menu-focus-target']));
```
ThomasBurleson: No need from array-wrapper.

src/core/util/prefixer.js
```diff
@@ -0,0 +1,63 @@
...
+    });
+
+    return attributes;
+  }
+
+  function _buildSelector(attributes) {
+    attributes = angular.isArray(attributes) ? attributes : [attributes];
+
+    return _buildList(attributes)
+      .map(function (item) {
```
ThomasBurleson: Nice!

src/components/bottomSheet/bottom-sheet.js
DevVersion: Good eye for detail :smile: Will change thx

src/core/util/prefixer.js
DevVersion: Hm, I don't really like it.

Then the actual `prefixer()` will be only used for `hasAttribute`. 
This will make the whole constructor function unnecessary.

I would rather change the constructor fn to something like that
```js
$mdUtil.prefixer('ng-click', true)
```

DevVersion:
@Frank3K Thanks for the feedback.

Awesome, that the code was enough the clarify the intention. When the code is clear enough to show, how it works, then it's good code (no self-praise :smile:)

Yea, I went through the most components and searched for queries, which can be accessed by developers.

Thanks for the alert of the auto-focus directive. I will add some more today, maybe you can help and list some other queries you found ;)

Frank3K:
I have done a quick search but haven't found more.

DevVersion:
@Frank3K Great, thanks for your effort. @ThomasBurleson PTAL.

src/core/util/prefixer.js
DevVersion: FYI: This comment is already outdated, since @ThomasBurleson had a discussion about it and figured out a better solution.

ThomasBurleson:
@DevVersion  - merge confict with master. Can you rebase please.

PR #8170, opened 2016-04-22
fix(tooltip): interpolate the tooltip text when adding it to the aria-label

### LINKED ISSUE #6855: mdTooltip: Aria-label added for md-tooltip but not interpolated.
md-tooltip adds an aria-label to the element ([source](https://github.com/angular/material/blob/master/src/components/tooltip/tooltip.js#L111)), but if the text in the md-tooltip is `{{ text }}`, the aria-label is also set to `{{ text }}`. When the text is displayed in the tooltip, it's correctly interpolated, but the aria-label is still wrong.

crisbeto:
Fixes the raw tooltip text text being added to the parent, if a tooltip contained a data binding, instead of the interpolated value.

Fixes #6855.

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -33,7 +33,7 @@ angular
  * @param {string=} md-direction Which direction would you like the tooltip to go?  Supports left, right, top, and bottom.  Defaults to bottom.
  */
 function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdTheming, $rootElement,
-                            $animate, $q) {
+                            $animate, $q, $interpolate) {
 
   var TOOLTIP_SHOW_DELAY = 0;
   var TOOLTIP_WINDOW_EDGE_SPACE = 8;
@@ -150,11 +150,22 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
         element.remove();
         attributeObserver && attributeObserver.disconnect();
       });
+
+      // Updates the aria-label when the element text changes. This watch
+      // doesn't need to be set up if the element doesn't have any data
+      // bindings.
+      if (element.text().indexOf($interpolate.startSymbol()) > -1) {
+        scope.$watch(function() {
+          return element.text().trim();
+        }, addAriaLabel);
+      }
     }
 
-    function addAriaLabel () {
-      if (!parent.attr('aria-label') && !parent.text().trim()) {
-        parent.attr('aria-label', element.text().trim());
+    function addAriaLabel (override) {
+      if ((override || !parent.attr('aria-label')) && !parent.text().trim()) {
+        var rawText = override || element.text().trim();
+        var interpolatedText = $interpolate(rawText)(parent.scope());
+        parent.attr('aria-label', interpolatedText);
       }
     }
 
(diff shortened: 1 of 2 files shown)


src/components/tooltip/tooltip.js
```diff
@@ -128,11 +128,22 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
         scope.visibleWatcher = scope.$watch('visible', onVisibleChanged );
         scope.$watch('direction', updatePosition );
       }
+
+      // Updates the aria-label when the element text changes. This watch
+      // doesn't need to be set up if the element doesn't have any data
+      // bindings.
+      if (element.text().indexOf($interpolate.startSymbol()) > -1) {
+        scope.$watch(function() {
```
ThomasBurleson: Huge performance impact with large lists where each item as a tooltip.
This is why we did the mutation observer... this `$watch( )` here my re-introduce perf issues.

crisbeto: Alright, I'll try to rework it and keep this one as a fallback for the case where a mutation observer isn't supported.

src/components/tooltip/tooltip.js
```diff
@@ -128,11 +128,22 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
         scope.visibleWatcher = scope.$watch('visible', onVisibleChanged );
         scope.$watch('direction', updatePosition );
       }
+
+      // Updates the aria-label when the element text changes. This watch
+      // doesn't need to be set up if the element doesn't have any data
+      // bindings.
+      if (element.text().indexOf($interpolate.startSymbol()) > -1) {
```
ThomasBurleson: What about ngMaterial's `aria.js` features?

src/components/tooltip/tooltip.js
```diff
@@ -128,11 +128,22 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
...
+        scope.$watch(function() {
+          return element.text().trim();
+        }, addAriaLabel);
+      }
     }
 
-    function addAriaLabel () {
-      if (!parent.attr('aria-label') && !parent.text().trim()) {
-        parent.attr('aria-label', element.text().trim());
+    function addAriaLabel (override) {
```
ThomasBurleson: See ^^ regarding `aria.js`

src/components/tooltip/tooltip.js
crisbeto: I followed the way it was done previously, but I'll at the Material aria provider.

src/components/tooltip/tooltip.js
crisbeto: A question about the mutation observer: In order to get it to fire for text changes, the `characterData` and `subtree` option need to be enabled. This fires the observer roughly twice as often as before. I think this might end up being slower than the watcher because we'll be calling the function that sets `aria-label` about 5 times on mouseover and about 5 times again on mouseleave.

src/components/tooltip/tooltip.js
crisbeto: `$mdAria` doesn't seem to be useful in this case, because it doesn't allow for the `aria-label` to be overwritten once it is set. I believe the expected behaviour here is that the `aria-label` changes together with the tooltip's text.

crisbeto:
Relabelling as per our discussion.

PR #8171, opened 2016-04-22
refactor(tests): switch back to karma-phantomjs-launcher

bvahdat:
which now supports PhantomJS2 as the default. For the details see [here](https://github.com/gskachkov/karma-phantomjs2-launcher/blob/master/README.md#karma-phantomjs2-launcher-is-depricated-use-karma-phantomjs-launcher100-instead) and [here] (https://github.com/karma-runner/karma-phantomjs-launcher/issues/97).


bvahdat:
The tests run automatically on `PhantomJS 2.1.1`as it's now the default by `karma-phantomjs-launcher`. Please note that I've not tested this PR on Windows but only on OS X

ThomasBurleson:
@bvahdat - this looks good. Thank you.

PR #8179, opened 2016-04-22
Added mdPreventRefocus in sideNav

jaxoncreed:
The code here https://github.com/angular/material/blob/7c6ff3653eb3d006dfcd5733da084c0d01c8b7ea/src/components/sidenav/sidenav.js#L377 refocuses on the element that opened the sidenav. This functionality isn't desirable in every context (for example I have an element with a md-tooltip that opens a sidenav. When the sidenav is closed, the tooltip with show up because the opener has been focused on)

Now adding md-prevent-refocus="true" will prevent this behaviour.


jaxoncreed:
I've signed it!

DevVersion:
@jaxoncreed Thanks for the contribution - I will close this PR, because a more elegant solution is already pending. See #7965

PR #8180, opened 2016-04-22
fix(list): proxy elements should be not triggered by other controls

### LINKED ISSUE #7937: mdSwitch: within mdListItem, mdSwitch toggles on mdSlider

DevVersion:
* The list item should not trigger the proxied element, when clicking on some specified components

The current forbidden controls are now:
- `md-slider`

Fixes #7937.

### DIFF
--- a/src/components/list/list.js
+++ b/src/components/list/list.js
@@ -286,7 +286,7 @@ function mdListItemDirective($mdAria, $mdConstant, $mdUtil, $timeout) {
           // Append the button wrap before our list-item content, because it will overlay in relative.
           itemContainer.prepend(buttonWrap);
           itemContainer.children().eq(1).append(tEl.contents());
-          
+
           tEl.addClass('_md-button-wrap');
         }
 
@@ -360,7 +360,7 @@ function mdListItemDirective($mdAria, $mdConstant, $mdUtil, $timeout) {
 
       function postLink($scope, $element, $attr, ctrl) {
         $element.addClass('_md');     // private md component indicator for styling
-        
+
         var proxies       = [],
             firstElement  = $element[0].firstElementChild,
             isButtonWrap  = $element.hasClass('_md-button-wrap'),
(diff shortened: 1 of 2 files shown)


topherfangio:
@DevVersion We already have some code that takes care of this. You can simply add `md-slider` to the list on line 220 of [list.js](https://github.com/angular/material/blob/master/src/components/list/list.js#L220).

This actually handles it at a higher level by changing the structure. Although I'm not terribly sure I like the current approach (which I think I wrote/refactored), I do think we should do it the same way.

If you feel like taking a stab at making the code cleaner, definitely feel free; but if not, you can just alter the code to add the `md-slider` to the existing list 😄

DevVersion:
@topherfangio This is quite a different list.

We don't want to have `md-slider` as a proxied element. Some components like the `md-slider` should just not bubble their click events up the DOM.

This is necessary, because otherwise users are not able to use the `md-slider` component for example.

See @ThomasBurleson's approach and the issue
- https://github.com/angular/material/issues/7937
- https://github.com/angular/material/pull/7966

topherfangio:
@DevVersion Gotcha; sorry, I didn't see the Codepen in the original issue and my suggestion made the test pass locally so I thought I knew what I was saying 😄 

That said, it seems like our solution of specifying which child elements we should ignore may not be the best approach.

Perhaps we could make a new directive like `md-no-bubble="click"` or something that stops the requested events from bubbling? This way, users could add it to any element that was causing problems and we wouldn't have to have a hard-coded list.

Thoughts?

DevVersion:
@topherfangio Hm, I can see the point of your opinion, but I don't think it's the most elegant approach.

The list should actually handle that automatically, because we know what custom components from ngMaterial should not bubble their events inside of a list.

Having an extra attribute for that, will make the list even more complicated IMO. 

Having a "hard-coded" list, which is actually a whitelist, is pretty the same as the array for the attributes, which should be copied over. 

I think this is something, were no solution is the "perfect" one.

To be honest, I also hate it, to have several "hard-coded" arrays inside of a component.

PR #8182, opened 2016-04-23
feat(panel): Clean up promises

KarenParker:
@ErinCoughlan @gmoothart @jelbourn @DerekLouie @ThomasBurleson  Please review.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -668,7 +668,7 @@ MdPanelService.prototype.create = function(opt_config) {
   angular.extend(this._config, this._defaultConfigOptions, configSettings);
 
   var instanceId = 'panel_' + this._$injector.get('$mdUtil').nextUid();
-  var instanceConfig = angular.extend({id : instanceId}, this._config);
+  var instanceConfig = angular.extend({ id: instanceId }, this._config);
 
   return new MdPanelRef(instanceConfig, this._$injector);
 };
@@ -787,21 +787,6 @@ function MdPanelRef(config, $injector) {
   /** @private {!Object} */
   this._config = config;
 
-  /** @private {!angular.$q.Promise|undefined} */
-  this._openPromise;
-
-  /** @private {!angular.$q.Promise|undefined} */
-  this._attachPromise;
-
-  /** @private {!angular.$q.Promise|undefined} */
-  this._detachPromise;
-
-  /** @private {!angular.$q.Promise|undefined} */
-  this._showPromise;
-
-  /** @private {!angular.$q.Promise|undefined} */
-  this._hidePromise;
-
   /** @private {!angular.JQLite|undefined} */
   this._panelContainer;
 
@@ -833,24 +818,15 @@ function MdPanelRef(config, $injector) {
  * the panel is opened and animations finish.
  */
 MdPanelRef.prototype.open = function() {
-  if (this._openPromise) {
-    // Panel is already shown so re-use (already resolved) promise from when
-    // it was shown.
-    return this._openPromise;
-  }
-
-  // TODO(ErinCoughlan) - Cancel any in-progress actions.
-
   var self = this;
-  this._openPromise = this.attach()
-      .then(function() {
-        return self.show();
-      }, this._$q.reject)
-      .then(function() {
-        return self; // Return a reference to the MdPanelRef.
-      }, this._$q.reject);
+  return this._$q(function(resolve, reject) {
+    var done = self._done(resolve, self);
+    var show = self._simpleBind(self.show, self);
 
-  return this._openPromise;
+    self.attach()
+        .then(show, reject)
+        .then(done, reject);
+  });
 };
 
 
@@ -861,20 +837,15 @@ MdPanelRef.prototype.open = function() {
  * closed and animations finish.
  */
 MdPanelRef.prototype.close = function() {
-  // TODO(ErinCoughlan) - Cancel any in-progress actions.
-
   var self = this;
-  var closePromise = this.hide()
-      .then(function() {
-        return self.detach();
-      }, this._$q.reject)
-      .then(function() {
-        // TODO(ErinCoughlan) - Add destroy. This will make the code here
-        // different than just calling this.detach().
-        return self._$q.resolve(self);
-      }, this._$q.reject);
+  return this._$q(function(resolve, reject) {
+    var done = self._done(resolve, self);
+    var detach = self._simpleBind(self.detach, self);
 
-  return closePromise;
+    self.hide()
+        .then(detach, reject)
+        .then(done, reject);
+  });
 };
 
 
@@ -885,42 +856,37 @@ MdPanelRef.prototype.close = function() {
  * the panel is attached.
  */
 MdPanelRef.prototype.attach = function() {
-  if (this.isAttached) {
-    return this._attachPromise;
+  if (this.isAttached && this._panelEl) {
+    return this._$q.when(this);
   }
 
-  // TODO(ErinCoughlan) - Cancel any in-progress actions.
-
   var self = this;
+  return this._$q(function(resolve, reject) {
+    var done = self._done(resolve, self);
 
-  this._attachPromise = this._$q.all([
-    this._createBackdrop(),
-    this._createPanel().then(function() {
-      self.isAttached = true;
-      self._addEventListeners();
-      return self._$q.resolve(self);
-    }, this._$q.reject)
-  ]);
-
-  return this._attachPromise;
+    self._$q.all([
+      self._createBackdrop(),
+      self._createPanel().then(function() {
+        self.isAttached = true;
+        self._addEventListeners();
+      }, reject)
+    ]).then(done, reject);
+  });
 };
 
 
 /**
  * Only detaches the panel. Will NOT hide the panel first.
  *
- * @returns {!angular.$q.Promise} A promise that is resolved when the panel is
+ * @returns {!angular.$q.Promise<MdPanelRef>} A promise that is resolved when the panel is
  * detached.
  */
 MdPanelRef.prototype.detach = function() {
   if (!this.isAttached) {
-    return this._detachPromise;
+    return this._$q.when(this);
   }
 
-  // TODO(ErinCoughlan) - Cancel any in-progress actions.
-
   var self = this;
-
   var detachFn = function() {
     self._removeEventListener();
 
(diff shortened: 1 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -861,20 +837,15 @@ MdPanelRef.prototype.open = function() {
...
-        return self.detach();
-      }, this._$q.reject)
-      .then(function() {
-        // TODO(ErinCoughlan) - Add destroy. This will make the code here
-        // different than just calling this.detach().
-        return self._$q.resolve(self);
-      }, this._$q.reject);
-
-  return closePromise;
+  return this._$q(function(resolve, reject) {
```
ThomasBurleson: Careful with the cluttering of rejects. I contend that this is a better structure:

```js
 var self = this; 
 
    var done = angular.bind(self,resolve, self),
        detach = angular.bind(self, self.detach);
        
    self.hide()
        .then( detach )
        .then( done )
        .catch( reject );
 });
};
```

> ^ is untested but I believe the logic is valid and more clear.

src/components/panel/panel.js
```diff
@@ -1003,31 +968,35 @@ MdPanelRef.prototype.show = function() {
...
+    return this._$q(function(resolve, reject) {
+      reject('Panel does not exist yet. Call open() or attach().');
+    });
   }
 
-  if (this._panelContainer.hasClass(MD_PANEL_HIDDEN) && this._hidePromise) {
-    return this._hidePromise;
+  var self = this;
+  if (this._panelContainer.hasClass(MD_PANEL_HIDDEN)) {
+    return this._$q(function(resolve) {
```
ThomasBurleson: Needless verbose... use `return this._$q.when(self)`;

src/components/panel/panel.js
```diff
@@ -1361,22 +1342,25 @@ MdPanelRef.prototype._animateOpen = function() {
  * @private
  */
 MdPanelRef.prototype._animateClose = function() {
+  var self = this;
   var animationConfig = this._config['animation'];
   if (!animationConfig) {
     this.removeClass('md-panel-is-showing');
     this.removeClass('_md-panel-shown');
-    return this._$q.resolve();
+    return this._$q(function(resolve) {
```
ThomasBurleson: `return this._$q.when(self);`

src/components/panel/panel.js
```diff
@@ -2044,11 +2028,22 @@ MdPanelAnimation.prototype.animateOpen = function(panelEl, animator) {
...
-    return this._reverseAnimation();
-  }
-  return $q.reject('No panel close animation. ' +
-      'Have you called MdPanelAnimation.animateOpen()?');
+  var self = this;
+  return $q(function(resolve, reject) {
+    debugger;
+    if (self._reverseAnimation) {
+      self._reverseAnimation().then(function() {
+        debugger;
```
ThomasBurleson: Remove `debugger` statements

src/components/panel/panel.js
```diff
@@ -2044,11 +2028,22 @@ MdPanelAnimation.prototype.animateOpen = function(panelEl, animator) {
...
+        resolve(self);
+      }, function() {
```
ThomasBurleson: Usually you specify the error argument and reject with that:

```js
}, function(error) { 
    reject(errror); 
});

File: src/components/panel/panel.js [hunk 6]
```diff
@@ -1343,15 +1316,23 @@ MdPanelRef.prototype._configureTrapFocus = function() {
...
+    return this._$q(function(resolve) {
+      resolve(self);
+    });
   }
 
-  return animationConfig.animateOpen(this._panelEl,
-      this._$mdUtil.dom.animator);
+  return this._$q(function(resolve, reject) {
+    animationConfig.animateOpen(self._panelEl, self._$mdUtil.dom.animator)
+        .then(function() {
```
ThomasBurleson:
```js
.then(function() {
   resolve(self);
}, 
``` 
is used in many places. Recommend you do this:

````
var respondWith = function(callback, value) {
   return angular.bind(null, callback, value);
}
```
then you can reuse this:

```js
var animator = self._$mdUtil.dom.animator;

   animationConfig
        .animateOpen(self._panelEl, animator)
        .then( respondWith(resolve, self) )
        .catch( respondWith(reject, self) );
```

src/components/panel/panel.spec.js
```diff
@@ -136,67 +138,69 @@ describe('$mdPanel', function() {
...
-    it('should reject on create when opening', function () {
+    it('should reject on create when opening', function() {
       var openRejected = false;
 
       panelRef._createPanel = function() {
         return panelRef._$q.reject();
       };
 
-      panelRef.open().then(function() {}, function () {
+      panelRef.open().then(function() {
```
DevVersion: Using those `noop` functions is not good here.

You're able to use `panelRef.open().catch(errorCallback)`, which is a shorthand for 
```js
promise.then(null, errorCallback)
```

This can be used in the same lines below.

src/components/panel/panel.js
KarenParker: This will work in some places, but others will need to change the thens differently. According to Erin, we can't guarantee the order the thens are executed if there are more than one.

ThomasBurleson: Not true. Each `then( )` returns a new promise; so the chaining sequence is guaranteed.

ThomasBurleson:
@KarenParker  - can you rebase with `master` and squash the 48 commits ?

KarenParker:
Erin broke all of feat/panel. I have it rebased and squashed correctly. All comments addressed and all tests passing.

src/components/panel/panel.js
```diff
@@ -885,42 +858,38 @@ MdPanelRef.prototype.close = function() {
...
+    var done = self._done(resolve, self);
+
+    self._$q.all([
+      self._createBackdrop(),
+      self._createPanel().then(function() {
+        self.isAttached = true;
+        self._addEventListeners();
+      }, reject)
+    ]).then(done)
+        .catch(reject);
```
ThomasBurleson: Do you expect `done()` to reject or throw an exception? 
* If yes, the `.catch(reject)` is good. 
* If no, then you can do `.then( done, reject )`

src/components/panel/panel.js
```diff
@@ -1369,18 +1341,53 @@ MdPanelRef.prototype._animateClose = function() {
...
+ */
+MdPanelRef.prototype._simpleBind = function(callback, self) {
+  return function(value) {
+    return callback.apply(self, value);
+  };
+};
+
+
+/**
+ * //Faster, more basic than angular.bind
```
ErinCoughlan: nit: Strange comment format

KarenParker:
Comments addressed.

src/components/panel/panel.spec.js
```diff
@@ -243,43 +235,39 @@ describe('$mdPanel', function() {
...
       expect(panelRef.isAttached).toEqual(true);
     });
 
-    it('should resolve on animate failure when closing', function () {
+    it('should resolve on animate failure when closing', function() {
       var closeResolved = false;
 
       panelRef.open();
-      $rootScope.$apply();
+      flushPanel();
```
ErinCoughlan: How about using `openPanel()` instead, which is these two lines?

src/components/panel/panel.js
```diff
@@ -1369,18 +1341,53 @@ MdPanelRef.prototype._animateClose = function() {
   if (!animationConfig) {
     this.removeClass('md-panel-is-showing');
     this.removeClass('_md-panel-shown');
-    return this._$q.resolve();
+    return this._$q.when(this);
   }
 
   var self = this;
```
ThomasBurleson: I am a huge advocate of code clarity; try this:

```js
var self = this,
    animator = self._$mdUtil.dom.animator;

return this._$q(function(resolve) {
    var done = function(response) {
            resolve(self, response);
            return self;
        },
        warnAndClose = function(error) {
            self._$log.warn('MdPanel Animations failed. Hiding panel without animating.');
            return done(error);
        };

    animationConfig
        .animateClose(self._panelEl, animator)
        .then( done )
        .catch( warnAndClose );
});
```

KarenParker:
Done.

ThomasBurleson:
Nice work @KarenParker. Thanks for being so responsive. 👍

src/components/panel/panel.js
```diff
@@ -961,37 +929,29 @@ MdPanelRef.prototype.detach = function() {
...
-      }, self._$q.reject)
-  ]);
-
-  return this._showPromise;
+  return this._$q(function(resolve, reject) {
+    var done = self._done(resolve, self);
+
+    self._$q.all([
+      self._backdropRef ? self._backdropRef.show() : self,
+      animatePromise().then(function() { self._focusOnOpen(); })
```
ErinCoughlan: Doesn't this `then` also need a reject? (I think this will break in Angular 1.6)

src/components/panel/panel.js
```diff
@@ -1003,31 +963,24 @@ MdPanelRef.prototype.show = function() {
...
-    }, self._$q.reject)
-  ]);
-
-  return this._hidePromise;
+  return this._$q(function(resolve, reject) {
+    var done = self._done(resolve, self);
+
+    self._$q.all([
+      self._backdropRef ? self._backdropRef.hide() : self,
+      self._animateClose().then(function() { self.addClass(MD_PANEL_HIDDEN); })
```
ErinCoughlan: Ditto here with the reject

src/components/panel/panel.js
```diff
@@ -1175,10 +1130,9 @@ MdPanelRef.prototype._focusOnOpen = function() {
     // Wait a digest to guarantee md-autofocus has finished adding the class
```
ErinCoughlan: Update this comment? We are waiting for angular to finish rendering the template now rather than a digest.

src/components/panel/panel.js
```diff
@@ -1369,18 +1337,53 @@ MdPanelRef.prototype._animateClose = function() {
...
+        .then(done, warnAndClose);
   });
 };
 
+
+/**
+ * Faster, more basic than angular.bind
+ * http://jsperf.com/angular-bind-vs-custom-vs-native
+ * @param {function} callback
+ * @param {Object} self
```
ErinCoughlan: nit: `!Object`

src/components/panel/panel.spec.js
```diff
@@ -27,12 +27,14 @@ describe('$mdPanel', function() {
...
     $mdConstant = $injector.get('$mdConstant');
     $mdUtil = $injector.get('$mdUtil');
+    $animate = $injector.get('$animate');
   };
 
   beforeEach(function() {
     module('material.components.panel', 'ngSanitize');
 
     inject(injectLocals);
+    $animate.enabled(false);
```
ErinCoughlan: Re-enable in an afterEach so we don't break others.

ErinCoughlan:
LGTM after comments are addressed.

Issue #7692, opened 2016-03-22
Cards do not show images in Safari


Tommatheussen:
The cards images are not being shown on safari. If I change some properties through the in browser editor I can sometimes see the images. I know they are some specific rules for flex attributes in safari.

The major problems is that the issue also exists on the demo site: https://material.angularjs.org/latest/demo/card
If might be good if it gets fixed there, are is mentioned somewhere so ppl are aware that they need specific attributes for this to work.

Safari Version 9.0.3 (11601.4.4) for reference

DevVersion:
I can reproduce that.

![image](https://cloud.githubusercontent.com/assets/4987015/13948498/b808840e-f020-11e5-9489-a6fcd5bc3e15.png)

topherfangio:
@DevVersion Can you look into this and find the cause?

DevVersion:
@topherfangio I've found the issue, it's caused by the width, which is set to `100%`, it seems like the computed height of 100% is zero. That means, when switching that to `auto`, everything works.

But I've noticed, that the `100%` were recently and intentionally  added, so I'm trying to understand those thoughts.

Tommatheussen:
Btw, when testing my app cross browser, I noticed that the images are not quite correct in FF either. So I went to check if they are correct in the demo page, and it seems the issue exists there as well, probably is fixed by the auto height as well, just wanted to let you know.

![screen shot 2016-03-24 at 11 17 11](https://cloud.githubusercontent.com/assets/13683094/14014306/1200f2ec-f1b2-11e5-91c0-cdf86ed4d130.png)

topherfangio:
@DevVersion Can you point me to the commit where it was added so I can take a look?

TitanNano:
@topherfangio The image height was set from `auto` to `100% !important` in df902381664bc568ad4011f1747ccc44a856995f

topherfangio:
@DevVersion It might be helpful for you to pair with @EladBezalel on this one as he did a major upgrade to the cards for 1.0.

DevVersion:
@EladBezalel I just assigned the issue to you, because you worked on those image rations.

jeremypokeefe:
I found a workaround for now if anyone is interested. I wrapped the img element in a div with the layout set to row and the alignment to center, center.

`<div layout="row" layout-align="center center">
   <img src="{{feature.imagePath}}" class="md-card-image" alt="{{feature.title}}" >
 </div>`

ThomasBurleson:
[...]

PR #8183, opened 2016-04-23
feat(card): add a class for overriding the image stretching

### LINKED ISSUE #7447: Images in md-card are being blown up to 100% width
Report from Googler on this line: https://github.com/angular/material/blob/master/src/components/card/card.scss#L63

> This is causing images to suddenly blow out our layouts.... I'm not an expert, but it seems that this rule might be incorrect it's applying to all image tags that happen to NOT be within an md-card-content.
> 
> Seems like this rule is for "full bleed" images
> 
> context: https://github.com/angular/material/issues/6273
> 
> We don't use md-card-content since our spec calls for 24px margins and we'd prefer to not use overrides.
> 
> Perhaps this could/should be a more semantically named rule rather than applying to all images? (e.g. <md-card-fullbleed-image src="blah.gif">

Reproduction: http://codepen.io/anon/pen/graaLZ

crisbeto:
* Adds a class that can be used if users don't want their images to be stretched inside `<md-card>`.
* Avoids some repetition in the `md-card` unit tests.

Fixes #7447.

@ThomasBurleson, @topherfangio this is as discussed in the issue. I tried to keep it as simple as possible.

### DIFF
--- a/src/components/card/card.js
+++ b/src/components/card/card.js
@@ -21,7 +21,8 @@ angular.module('material.components.card', [
  * @description
  * The `<md-card>` directive is a container element used within `<md-content>` containers.
  *
- * An image included as a direct descendant will fill the card's width, while the `<md-card-content>`
+ * An image included as a direct descendant will fill the card's width. If you want to avoid this,
+ * you can add the `md-plain-image` class to the parent element. The `<md-card-content>`
  * container will wrap text content and provide padding. An `<md-card-footer>` element can be
  * optionally included to put content flush against the bottom edge of the card.
  *
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -211,3 +211,10 @@ md-card {
   }
 }
 
+.md-plain-image {
+  > img {
+    width: auto;
+    height: auto;
+  }
+}
+
--- a/src/components/card/card.spec.js
+++ b/src/components/card/card.spec.js
@@ -1,21 +1,27 @@
 describe('mdCard directive', function() {
 
   var $mdThemingMock = function() { $mdThemingMock.called = true; };
+  var $compile;
+  var $rootScope;
 
-  beforeEach(module(function($provide) {
-    $provide.value('$mdTheming', $mdThemingMock);
-  }));
+  beforeEach(function() {
+    module('material.components.card');
+    module(function($provide) {
+      $provide.value('$mdTheming', $mdThemingMock);
+    });
+    inject(function(_$compile_, _$rootScope_) {
+      $compile = _$compile_;
+      $rootScope = _$rootScope_;
+    });
+  });
 
-  beforeEach(module('material.components.card'));
-
-  it('should be themable', inject(function($compile, $rootScope) {
+  it('should be themable', function() {
     $compile('<md-card></md-card>')($rootScope.$new());
     expect($mdThemingMock.called).toBe(true);
-  }));
+  });
 
-  it('should have `._md` class indicator', inject(function($compile, $rootScope) {
+  it('should have `._md` class indicator', function() {
     var element = $compile('<md-card></md-card>')($rootScope.$new());
     expect(element.hasClass('_md')).toBe(true);
-  }));
-  
+  });
 });


src/components/card/card.js
```diff
@@ -128,3 +129,41 @@ function mdCardDirective($mdTheming) {
...
+ * <hljs lang="html">
+ * <md-card>
+ *  <img src="card-image.png" alt="image caption" md-plain-image>
+ * </md-card>
+ * </hljs>
+ */
+function mdPlainImageDirective($mdUtil) {
+  return {
+    restrict: 'A',
+    link: function ($scope, $element) {
```
DevVersion: Can you not do this in the `compile` function?

crisbeto: I wasn't sure whether it wouldn't mess with other directives.

crisbeto: Just tried it out, it throws the browser into an infinite loop if it gets used within a `ngRepeat`.

DevVersion: Oh well, yeah it's even logic, since it tries to compile the wrapped element again.

src/components/card/card.js
```diff
@@ -128,3 +129,41 @@ function mdCardDirective($mdTheming) {
...
+ * particular image.
+ *
+ * @usage
+ * ### Card with a plain image
+ * <hljs lang="html">
+ * <md-card>
+ *  <img src="card-image.png" alt="image caption" md-plain-image>
+ * </md-card>
+ * </hljs>
+ */
```
DevVersion: I think it would be good to mention, the new directive in the `md-card` directive description as well.

src/components/card/card.js
crisbeto: Yeah, it works if we remove the directive attribute before wrapping it.

crisbeto: Too bad we don't your prefixing util yet :D

DevVersion: haha :smile:

EladBezalel:
Why this whole solution is made as css class that the user adds?

src/components/card/card.js
DevVersion: @crisbeto Just saw your change - very nice!

crisbeto:
Because the `flex-direction: column;` on `md-card` also stretches the image (in addition to the `width: 100%`). By wrapping it a `div`, the `div` gets stretched instead.

EladBezalel:
This is a weird approach :confused: are we sure that there's no simple css solution here?

BTW this will fix #7692 as well?

crisbeto:
The only CSS solution I found was setting `flex-direction: row` and `flex-wrap: wrap` on the `md-card`, in addition to the CSS changes that I've done with this commit. Setting those flex properties changes the alignment in other components which is why I decided to avoid it.

As for #7692, I'm not sure. People could use this to avoid the resizing altogether, though.

src/components/card/card.js
```diff
@@ -8,7 +8,8 @@
 angular.module('material.components.card', [
     'material.core'
   ])
-  .directive('mdCard', mdCardDirective);
+  .directive('mdCard', mdCardDirective)
+  .directive('mdPlainImage', mdPlainImageDirective);
```
ThomasBurleson: I think this should be `md-no-stretch`

ThomasBurleson: or perhaps we should just create a directive:

`<md-card-image ng-href=""></md-card-image>`

topherfangio: Correct me if I am wrong, but that would not be backward's compatible right?

Is it better to just make a breaking change?

crisbeto: Yeah, that sounds like a breaking change. I tried to keep it as conflict-free as possible.

ThomasBurleson:
@crisbeto, @topherfangio  - I do not like this solution

ThomasBurleson:
Is this PR still needed after SHA d381619 (committed just now).

crisbeto:
I'd say so, d3816196bd4ae3f265c4d4d6d8341a9623bd76d0 just fixes the image always being stretched to 100% height, this PR adds the ability to disable stretching completely (usually it gets stretched to 100% width as well).

src/components/card/card.js
ThomasBurleson: @topherfangio - what are you suggesting code wise ?

topherfangio: If we don't mind introducing a breaking change, then I think your `<md-card-image>` solution is great. I just wanted to make sure we knew it was a breaking change to do something like that 😄

ThomasBurleson: @crisbeto, @topherfangio - we already have 

```html
<md-card-header>
  <md-card-avatar>
    <img class="md-user-avatar" src="avatar.png"/>
  </md-card-avatar>
</md-card-header> 
```
And we also have a `.md-card-image` class:

```html
<md-card>
 <img ng-src="card-image.png" class="md-card-image" alt="image caption">
</md-card>
```

So I think we should not use .md-plain-image and instead improve the class:

```css
.md-card-image {
// We need to use !important, because md-card also uses it.
height: auto !important;
width: auto;
max-width: 100%;
}
```

ThomasBurleson: @crisbeto, @topherfangio - we already have 

```html
<md-card-header>
  <md-card-avatar>
    <img class="md-user-avatar" src="avatar.png"/>
  </md-card-avatar>
</md-card-header> 
```
And we also have a `.md-card-image` class:

```html
<md-card>
 <img ng-src="card-image.png" class="md-card-image" alt="image caption">
</md-card>
```

So I think we should **not** use `.md-plain-image` and instead improve the **`md-card-image`** type style:

```css
md-card-image {
  display:block;
  position:relative;

  width: auto;
  max-width: 100%;
  height: auto !important; // !important because md-card also uses it.
}
```

```html
<md-card-image>
  <img ng-href="xxx" ></img>
</md-card-image>
```

crisbeto: I didn't know about `md-card-image`, I'll check it out.

crisbeto:
@ThomasBurleson I replaced the directive with a class that you can apply to a parent of an image, in order to prevent the stretching (note that it can't be on the image itself). I've kept the improvements to the testing setup.

PR #8184, opened 2016-04-23
fix(sidenav): update position of sidenav and backdrop if scrolled.

### LINKED ISSUE #5850: md-sidenav was delayed

DevVersion:
Currently the sidenav is not animating when the parent is scrolled, this
was caused by the absolute position.

This commit, updates the position of the sidenav and backdrop when the
parent is scrolled.

It sets the position to the current scroll position and sets the height to
the parents `clientHeight`. Using a `100%` height, will impact the
performance heavily.

Fixes #5850.


stephengardner:
This is merged in 1.1.0-rc.5? I'm still having the issue, scrolling the backdrop, clicking the menu button and the page scrolls to the top

PR #8187, opened 2016-04-23
fix(constants): prefixes for properties should be correct.

### LINKED ISSUE #8186: sniffer: vendorProperty needs to be checked
`$mdConstant` uses a function, which prefixes the given property with the correct prefix for several browsers.

| Browser | Supported Properties |
| ----------- | ----------------------------- |
| Chrome | `transform | webkitTransform`  |
| Firefox | `transform | MozTransform`  |
| IE11| `transform | msTransform`  |

Also some Android Browsers are requiring a prefix, `Webkit` or `webkit`.

DevVersion:
* Some Android Browsers are requiring different prefixes
* IE requires in some versions a lower-case prefix.
* Old Firefox Versions are requiring a `Moz` prefix.
* Chrome and latest Firefox are now supporting the un-prefixed property.

This commit, checks the properties on the browser and validates the correct property.
This makes sure, that the properties are working properly.

Closes #8186.

### DIFF
--- a/src/core/util/constant.js
+++ b/src/core/util/constant.js
@@ -7,9 +7,30 @@ angular.module('material.core')
  */
 function MdConstantFactory($sniffer) {
 
-  var webkit = /webkit/i.test($sniffer.vendorPrefix);
+  var vendorPrefix = $sniffer.vendorPrefix;
+  var isWebkit = /webkit/i.test(vendorPrefix);
+  var SPECIAL_CHARS_REGEXP = /([:\-_]+(.))/g;
+  var prefixTestEl = document.createElement('div');
+
   function vendorProperty(name) {
-    return webkit ?  ('webkit' + name.charAt(0).toUpperCase() + name.substring(1)) : name;
+    // Add a dash between the prefix and name, to be able to transform the string into camelcase.
+    var prefixedName = vendorPrefix + '-' + name;
+    var ucPrefix = camelCase(prefixedName);
+    var lcPrefix = ucPrefix.charAt(0).toLowerCase() + ucPrefix.substring(1);
+
+    return hasStyleProperty(name)     ? name     :       // The current browser supports the un-prefixed property
+           hasStyleProperty(ucPrefix) ? ucPrefix :       // The current browser only supports the prefixed property.
+           hasStyleProperty(lcPrefix) ? lcPrefix : name; // Some browsers are only supporting the prefix in lowercase.
+  }
+
+  function hasStyleProperty(property) {
+    return angular.isDefined(prefixTestEl.style[property]);
+  }
+
+  function camelCase(input) {
+    return input.replace(SPECIAL_CHARS_REGEXP, function(matches, separator, letter, offset) {
+      return offset ? letter.toUpperCase() : letter;
+    });
   }
 
   return {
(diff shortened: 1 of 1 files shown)


src/core/util/constant.js
```diff
@@ -7,9 +7,22 @@ angular.module('material.core')
  */
 function MdConstantFactory($sniffer) {
 
-  var webkit = /webkit/i.test($sniffer.vendorPrefix);
+  var isWebkit = /webkit/i.test($sniffer.vendorPrefix);
+  var prefixTestEl = document.createElement('div');
```
ThomasBurleson: This code is not DRY. Here is a better solution:

```js
var prefixTestEl = document.createElement('div');
var SPECIAL_CHARS_REGEXP = /([\:\-\_]+(.))/g;
var MOZ_HACK_REGEXP = /^moz([A-Z])/;


function vendorProperty(name) {
    var prefix = $sniffer.vendorPrefix;
    var prefixedName = prefix + name ;
    var ucName = camelCase( prefixedName );
    var lcName = camelCase( prefixedName, true );

    return  hasStyle( name )    ? name   :          // The current browser supports the un-prefixed property
            hasStyle( ucName  ) ? ucName :          // The current browser only supports the prefixed property.
            hasStyle( lcName )  ? lcName : name;    // Some browsers are only supporting the prefix in lowercase.
}


/**
 * Util to test for valid css property name/key
 */
function hasStyle (key) {
    return angular.isDefined( prefixTestEl.style[key] );
}

/**
 * Converts snake_case to camelCase.
 * Also there is special case for Moz prefix starting with upper case letter.
 * @param name Name to normalize
 */
function camelCase(name, lowerCamelCase) {
  return name.
    replace(SPECIAL_CHARS_REGEXP, function(_, separator, letter, offset) {
      return offset || !lowerCamelCase ? letter.toUpperCase() : letter.toLowerCase();
    }).
    replace(MOZ_HACK_REGEXP, 'Moz$1');
}
```

DevVersion: To be honest, this code is much more complicated as the code before. 

- It uses a RegEx specially for the `Moz` prefix.
- Using a RegEx and iterate through all groups is unnecessary, because we only need to modify the first letter.

I like the `hasStyle` function, this is very clean.
I also like the idea of the `camelCase` function, but it can be improved, just for our cases.

````js
function camelCase(input, isLowerCase) {
 return (isLowerCase() ? name.charAt(0).toLowerCase() : 
                         name.charAt(0).toUpperCase()) + name.substring(1);
}
```

DevVersion: This code is not working, because the `camelCase` function is converting snake case to camel case, but the properties are not in snake case, which means, that we need to modify this function.

ThomasBurleson:
@DevVersion  - unit tests please.

DevVersion:
@ThomasBurleson As written on Slack. Tests are not possible. 

I tried
- Property Descriptors, 
- Overwriting CSSStyleDeclaration prototype, 
- Intercepting `document.createElement`
- Mocking the Sniffer

src/core/util/constant.js
```diff
@@ -7,9 +7,30 @@ angular.module('material.core')
...
+
   function vendorProperty(name) {
-    return webkit ?  ('webkit' + name.charAt(0).toUpperCase() + name.substring(1)) : name;
+    // Add a dash between the prefix and name, to be able to transform the string into camelcase.
+    var prefixedName = vendorPrefix + '-' + name;
+    var ucPrefix = camelCase(prefixedName);
+    var lcPrefix = ucPrefix.charAt(0).toLowerCase() + ucPrefix.substring(1);
+
+    if (hasStyleProperty(name))   return name;       // The current browser supports the un-prefixed property
```
ThomasBurleson: Don't like it. I recommend this:

```

return hasStyleProperty(name)     ? name     : 
       hasStyleProperty(ucPrefix) ? ucPrefix :
       hasStyleProperty(lcPrefix) ? lcPrefix : name;
```

DevVersion: Sorry for the delay. In my opinion, both code-styles were good / clear. Just updated it to yours.

PR #8190, opened 2016-04-24
fix(toolbar): remove transition duration for $ngAnimate 

### LINKED ISSUE #7929: mdToolbar: Hide button via ng-show/ng-hide takes a while
In Angular Material v1.1.0-rc2 hiding buttons inside `.md-toolbar-tools` via ng-show/ng-hide takes a while. Seems to be caused by this css transition:

```css 
.md-toolbar-tools .md-button, .md-toolbar-tools .md-button.md-icon-button md-icon {
    transition: all 0.5s cubic-bezier(0.35, 0, 0.25, 1);
    transition-property: background-color, fill, color;
}
```

Demo: http://codepen.io/anon/pen/YqYNzX
Browsers: Chrome 49, Firefox 45 and Edge

DevVersion:
* Other components, like `md-toolbar` use a custom transition on the background-color, color etc, but the related CSS selector has a high specificity.

* So for example, when ng-hide is applied, then we set the transition to `none`.
But that's not working sometimes, because the both selectors have the same specificity.

To fix this, we can apply `!important` or increase the selector specificity by using a CSS hack.

https://www.w3.org/TR/CSS2/cascade.html#specificity

The toolbar selector has in our case a specificity of 20.
* .md-button.ng-hide // 0,0,2,0
* .md-toolbar-tools .md-button // 0,0,2,0

------
So this is actually a thing, which is caused by $ngAnimate, which is not able to solve this issue.

We need to reset the transition duration, when $ngAnimate looks for it, manually.

$ngAnimate Issue: angular/angular.js#8224 (comment)

Fixes #7929

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -71,6 +71,9 @@ button.md-button::-moz-focus-inner {
     text-decoration: none;
   }
 
+  // By default $ngAnimate looks for transition durations on the element, when using ng-hide, ng-if, ng-show.
+  // The .md-button has a transition duration applied, which means, that $ngAnimate delays the hide process.
+  // To avoid this, we need to reset the transition, when $ngAnimate looks for the duration.
   &.ng-hide, &.ng-leave {
     transition: none;
   }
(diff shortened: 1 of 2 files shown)


crisbeto:
LGTM @ThomasBurleson

ThomasBurleson:
@DevVersion Nice 👍

PR #8191, opened 2016-04-24
feat(ripple): add the ability to disable ripples globally

### LINKED ISSUE #5669: Disable ripple ink effect globally
please add config attribute to disable ripple ink effect on all components (button, checkboxes, list items etc.)

crisbeto:
Exposes the `$mdInkRippleProvider` during app config and allows the user to disable ripples globally via the `disableInkRipple` method.

Fixes #5669.


ThomasBurleson:
Nice 👍

Issue #7257, opened 2016-02-23
md-button: md-icon-button vertically misaligned, text misaligned when used with icon


ivoviz:
## Icon Button Alignment

Vertical alignment for the icon within a `md-icon-button` is not correct. 

``` scss

.md-button.md-icon-button {
  width: 39px !important;
  height: 39px !important;
  // Setting line-height to 0px fixes the issue
  line-height: 0px !important;
}
```

_Kudos Workaround_ @bluestella
## Text Alignment

``` html
<md-button>
  <md-icon md-svg-src="your/icon.svg"></md-icon>
  Register Now
</md-button>
```

[Codepen here](http://codepen.io/anon/pen/NxQywr)

Probably a simple `margin-top: -2px` on `md-icon` would solve the problem. (Also tested on the said browsers.)

Using the above snippet the icon is misaligned vertically by 2px.
 Tested in latest Chrome, FF and Edge (31.14267.1000).
ngMaterial Version: 1.0.5

outofrange:
Had the same problem, and used the workaround from here: http://stackoverflow.com/a/33950707/1436981 (uses `height: auto` instead of px margins)

DevVersion:
Yes, can reproduce that. I will take a look at that.

![](https://i.gyazo.com/60360f41a69e40770e6ff4d33464b7d8.png)

ivoviz:
Well I must be blind then, but even on your second image the icon seems to be a bit off for me (probably the text as well), but please correct me if I'm wrong.

![pxs](https://cloud.githubusercontent.com/assets/1143746/13296894/fe18216c-db2f-11e5-8249-5f554b8b794a.png)

DevVersion:
Indeed, can be true :+1: Can you tell me, is the text normally (without an icon - plain text) correctly aligned?

btw: Nice specification, you created your image manually?

ivoviz:
Yeah I did. :)

Text seems to be inconsistent. On Win10 (14267) with latest Chrome it's misaligned:

![basic](https://cloud.githubusercontent.com/assets/1143746/13298479/30415364-db37-11e5-8f61-cdd503cc24e5.png)

On Ubuntu though, it seems to be correct (same version of Chrome):

![basic_ubuntu](https://cloud.githubusercontent.com/assets/1143746/13298508/40058a0e-db37-11e5-810c-cb400df00f7a.png)

If needed I can check other browsers too.

DevVersion:
@ivoviz Awesome detailed images!

[...]

Issue #3383, opened 2015-06-21
md-button should use wrapper div for flexbox support on firefox


rschmukler:
Right now `md-menu` is broken on firefox. This is because firefox does not allow applying flexbox to elements that defer to OS for rendering, specifically the `button`. See [this stackoverflow](https://stackoverflow.com/questions/24289251/flexbox-not-working-in-buttons-in-firefox) and this [bug tracker](https://bugzilla.mozilla.org/show_bug.cgi?id=984869#c2) for commentary on it. The point is, Mozilla has marked this as "intended" behavior and the solution is to put a `div` inside the button that uses flexbox. Wanted to ping people to see if there is any breakage that comes to immediate mind with this solution...

@jelbourn @robertmesserle @ThomasBurleson. Any objections?

This will also help anyone using ngMaterial who expects the button to behave like a flexbox (which, given our CSS, should be everyone).

jelbourn:
I'm actually of the opinion that none of our components should have `display: flex` or `display: inline-flex` on their outermost element. The problem is that the baseline computation for flex elements is different from `block` / `inline-block`, and this makes it very difficult to do something like putting an `md-button` next to an `input`.

ThomasBurleson:
Concensus with @jelbourn and @rschmukler - outside container for component should be `display:inline-block;`

PR #8194, opened 2016-04-24
fix(mdButton): icon vertical alignment

### LINKED ISSUE #8066: md-button-icon: Icon not centered
Icons in `md-button`s are no longer centered vertically. There is more space on the top then to bottom.

See picture 
image
You can see it easily if you set a background color on the icon so the full square is shown. But it's still noticeable without. See the button demo for an example https://material.angularjs.org/latest/demo/button

Adding `layout="row" layout-align="center center"` seems to fix but seems a bit overkill.

OSX 10.11.4
Chrome Version 50.0.2661.75 (64-bit)

epelc:
I applied the display change to `.md-button md-icon` because regular buttons with icons and fab buttons also require this fix.

This fixes `md-icon` vertical alignment within `md-button`.  Closes #8066

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -145,6 +145,11 @@ button.md-button::-moz-focus-inner {
     &.ng-hide, &.ng-leave {
       transition: none;
     }
+
+    md-icon {
+      // Fixes icon vertical alignment.
+      display: block;
+    }
   }
 
   &:not([disabled]) {
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -147,7 +147,8 @@ button.md-button::-moz-focus-inner {
     }
 
     md-icon {
-      // Fixes icon vertical alignment.
+      // Using `display:block;` is required for correct vertical alignment
+      // because `.md-button` is using `display: inline-block;`.
       display: block;
     }
   }
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -145,12 +145,6 @@ button.md-button::-moz-focus-inner {
     &.ng-hide, &.ng-leave {
       transition: none;
     }
-
-    md-icon {
-      // Using `display:block;` is required for correct vertical alignment
-      // because `.md-button` is using `display: inline-block;`.
-      display: block;
-    }
   }
 
   &:not([disabled]) {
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@DevVersion - can you review/test ?

epelc:
See the notes on the issue for how to test it easily.

On Sunday, April 24, 2016, Thomas Burleson <notifications@github.com> wrote:

> @DevVersion <https://github.com/DevVersion> - can you review/test ?
>
> —
> You are receiving this because you authored the thread.
> Reply to this email directly or view it on GitHub
> <https://github.com/angular/material/pull/8194#issuecomment-214044570>
>


-- 
Best regards,
Ed Pelc
*President*


Website: freightchick.com
Phone: 941-735-4047

src/components/button/button.scss
```diff
@@ -145,6 +145,11 @@ button.md-button::-moz-focus-inner {
     &.ng-hide, &.ng-leave {
       transition: none;
     }
+
+    md-icon {
+      // Fixes icon vertical alignment.
```
DevVersion: I think this comment is actually not saying really much. 

Can we explain more, why this is needed and for example, which components are affected by the vertical alignment problem?

epelc: @DevVersion Agreed. I'll see if I can figure out why it actually fixes it. iirc `layout="row"` has a rule to have all children use `display: block;`.

epelc:
So I did some digging, and the icon alignment issue was actually introduced way back in `0.10.1` by the fix for #3383. That fix changed `.md-button`'s `display` property from `flex` to `inline-block`. Which made `.md-button`'s `align-items:  center;` a no op because it requires a flex container. This causes the icon to fallback to default browser alignment for buttons.

The default browser alignment for buttons doesn't appear to use any css properties and is just something specific to buttons. I couldn't find any info about this around the web or in the spec for buttons. It's using some hidden builtin method for vertical alignment, instead of a regular css property.

I'll update my comment shortly.

epelc:
Note I also just tested this with `a href` buttons and the fix works as expected.

ThomasBurleson:
👍

ivoviz:
I was curious since I've already opened an issue about this (#7257), so I've tested. It only works for fab buttons. Didn't you misplace the rule @epelc? Also, if you add this to `.md-button md-icon` as you mentioned in the first comment, it would break buttons with icon and text together.

epelc:
@ivoviz In the original issue I was thinking to add it for `.md-icon-button md-icon` but then I switched to `.md-button md-icon` so that it also fixes fabs which do not use `md-icon-button`.

It does in fact break the layout in a couple cases. I original checked that `a href` works with an `md-icon-button` but it is broken with a `.md-fab` using an `a href`.

Here is a [demo](http://codepen.io/epelc/pen/qZMNeY?editors=1100) with all the combos of button/a href and the different classes.

epelc:
Here is an updated [demo](http://codepen.io/epelc/pen/ZWMpEK?editors=1100) that's working across all cases.

Using this rule
```
.md-button.md-icon-button,
button.md-button.md-fab {
  md-icon {
    display: block;
  }
}
```
I'll update my pr to include this.

ivoviz:
@epelc Yup, this new one seems to work, thanks!

DevVersion:
@epelc Just wanted to let you know, that I'll test the PR tomorrow.

DevVersion:
I'm sorry that I didn't test it. Issues are sometimes getting lost. 

I just checked your PR out and compared with the latest demos.
And the PR seems to be correct / fine.

Some calculations:

- Button Height: `56px`
- Icon Height: `24px`

So the correct top offset (`offsetTop`) should be `(56px - 24px) / 2`.

- Head: `17px`
- PR: `16px` (**correct**)

Since this is actually testable, we should add this as a unit test, because we're able to execute those calculations on Headless browsers as well.

Once the Test is there, LGTM.

PR #8206, opened 2016-04-25
fix(util): restore scrolling after test executed.

DevVersion:
* The current test, was not restoring the scroll mask, which means, that we're not able to remove the scrollbars on the karma site.

This broke some positioning tests for `feat/panel`.

cc. @ErinCoughlan


ErinCoughlan:
LGTM

Issue #7477, opened 2016-03-09
Autocomplete ng-messages wrong font-size


gpopovic:
In master, ng-messages font size for autocomplete is wrong.

In codepen, you can compare input ng-message  with autocomplete ng-message
http://codepen.io/anon/pen/ONNgMQ

DevVersion:
Great demo, I'll take a look at this soon.

gpopovic:
Any updates on this one?

DevVersion:
Just investigated a bit and it seems to relate to the scope transclusion.

I tried to create a scope transclusion myself to take a look at its most elegant solution.

http://codepen.io/DevVersion/pen/WwRqaO?editors=1010

So just want to notify you, that this issue is in progress.

EladBezalel:
This is related to the styles applied, 
Take a look at [input messages style](https://github.com/angular/material/blob/dbb75a64e76f5ac1c22b75e48318c6ce4d4095cb/src/components/input/input.scss#L233)

DevVersion:
@EladBezalel Yes actually it is related to the styes. But updating the style selectors won't be the right solution, because the ngMessage isn't retrieving the animation class correctly. 

This is caused by the parent transclusion from the autocomplete. 

(https://github.com/angular/material/blob/master/src/components/input/input.js#L711)

EladBezalel:
:+1:

cairomax:
I am not sure if this may helps:

Pull request https://github.com/angular/material/pull/5637 removed the code the replaces a wrapper div generated by mdAutocomplete. See this line:
https://github.com/angular/material/pull/5637/commits/05d7f24d6929b034036daa43167335d5b32e2b8d#diff-f700fe0b273e851967e22440f2c464eeL25
Also, see the occurences of the string `md-autocomplete-replace` in the whole repository. It was used as a marker class on the div to remove, but it is never taken into account.

Since `mdInputContainer` requires `ngMessages` to be a direct child, this might be the cause of the problem.

PR #8209, opened 2016-04-25
feat(input) add md-dense and match css to spec

clshortfuse:
I've rewritten many of the scss and js files to conform to Material Design spec. There are many changes, but it wasn't possible to separate much of these into smaller pull requests because of the way they depend on each other. I have changes for md-toolbar that can be used later on another PR.

Here are codepens I've made to show the issues with the current implementations and the new changes. Be sure to set your browser with to less than 960px for toolbar height to match the examples.

Layout Tests
Master: http://codepen.io/shortfuse/full/ZWxYpN/
SpecFix: http://codepen.io/shortfuse/full/qZojGW/

Horizontal Tests
Master: http://codepen.io/shortfuse/full/JKGdQv/
SpecFix: http://codepen.io/shortfuse/full/gMPYjE/

In summary, over JS, input-containers will have the classes `md-input-has-label` and `md-input-has-messages` so certain CSS rules can apply dependent on if there is a label and/or messages.

* Add md-dense to `md-input`, `md-autocomplete`, and `md-select`
* Password spacing set to 2px
* Variable top and bottom margins
* Intelligently hide md-error-spacer when blank and messages present
* Fix md-select to align properly and match spec
* Fix checkbox to 18px instead of 20px
* Add `md-no-border` option for md-select
* Ensure `md-select-menu` `font-size` matches `md-select` `font-size`
* Fix `md-autocomplete` spacing
* Fix text-less `md-checkbox` width
* Align `md-icon` to top of input box (support all font-sizes)
* Support **any** font size for `md-input`
* Added missing spaces to counter text (ie: `26 / 30` instead of `26/30`)

Material Design Sources:

https://www.google.com/design/spec/components/text-fields.html
https://www.google.com/design/spec/patterns/errors.html

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -140,6 +140,8 @@ function labelDirective() {
       if (!containerCtrl || attr.mdNoFloat || element.hasClass('md-container-ignore')) return;
 
       containerCtrl.label = element;
+      containerCtrl.element.addClass('md-input-has-label');
+      
       scope.$on('$destroy', function() {
         containerCtrl.label = null;
       });
@@ -179,6 +181,8 @@ function labelDirective() {
  * @param md-detect-hidden {boolean=} When present, textareas will be sized properly when they are
  *   revealed after being hidden. This is off by default for performance reasons because it
  *   guarantees a reflow every digest cycle.
+ * @param md-spacing {string=} Override spacing between icon and text. Default is 36px. Use `wide` for
+ *   48px. Use `extra-wide` for 56px.
  *
  * @usage
  * <hljs lang="html">
@@ -704,7 +708,7 @@ function placeholderDirective($compile) {
       // it later. This is necessary, because if we compile the element beforehand,
       // it won't be able to find the `mdInputContainer` controller.
       inputContainer.element
-        .addClass('md-icon-float')
+        .addClass('md-icon-float md-input-has-label')
         .prepend(newLabel);
 
       $compile(newLabel)(scope);
@@ -808,16 +812,18 @@ function ngMessagesDirective() {
     // If we are not a child of an input container, don't do anything
     if (!inputContainer) return;
 
-    // Add our animation class
-    element.toggleClass('md-input-messages-animation', true);
+    // Ensure inputContainer has class `md-input-has-messages`
+    inputContainer.element.addClass('md-input-has-messages');
 
+    // Add our animation class
     // Add our md-auto-hide class to automatically hide/show messages when container is invalid
-    element.toggleClass('md-auto-hide', true);
+    element.addClass('md-input-messages-animation md-auto-hide');
 
     // If we see some known visibility directives, remove the md-auto-hide class
     if (attrs.mdAutoHide == 'false' || hasVisibiltyDirective(attrs)) {
       element.toggleClass('md-auto-hide', false);
     }
+
   }
 
   function hasVisibiltyDirective(attrs) {
--- a/src/components/input/input.scss
+++ b/src/components/input/input.scss
@@ -1,37 +1,87 @@
-$input-container-padding: 2px !default;
+$input-container-margin-end: 16px !default;
+$input-container-margin: 0 !default;
+$input-container-padding-start: 1px !default;
+$input-container-min-height: 48px !default;
+$input-container-min-height-dense: 40px !default;
 
+
+
+$default-top-margin: 10px !default;
+$default-bottom-margin: 11px !default;
+$default-border-width: 1px !default;
+$default-border-width-focused: 2px !default;
+
+$default-top-margin-dense: $default-top-margin - 2px !default;
+$default-bottom-margin-dense: $default-bottom-margin - 2px !default;
+
+
+$input-label-font-size: 12px !default;
+$input-label-margin-top: $default-top-margin !default;
+
+// Material Design spec says it should be an extra 4px,
+// but that would break horizontal alignment, so use 0px instead
+$input-label-with-messages-margin-top: $default-top-margin + 0px !default;
+
+$input-label-margin-top-dense: $default-top-margin-dense !default;
+$input-label-margin-start: 0 !default;
 $input-label-default-offset: 24px !default;
-$input-label-default-scale: 1.0 !default;
-$input-label-float-offset: 6px !default;
-$input-label-float-scale: 0.75 !default;
-$input-label-float-width: $input-container-padding + 16px !default;
+$input-label-default-offset-dense: $input-label-default-offset - 4px !default;
+$input-label-float-width: 16px !default;
 
-$input-placeholder-offset: $input-label-default-offset !default;
+$input-margin-top: $default-top-margin !default;
+$input-margin-top-dense: $default-top-margin-dense !default;
+$input-margin-bottom: $default-bottom-margin !default;
+$input-margin-bottom-dense: $default-bottom-margin-dense !default;
+$input-font-size: inherit !default;
+$input-font-size-dense: 13px !default;
+$input-padding-top: 0 !default;
+$input-padding-start: 0 !default;
+$input-padding-bottom: 8px !default;
+$input-border-width-default: $default-border-width !default;
+$input-border-width-focused: $default-border-width-focused !default;
 
-$input-border-width-default: 1px !default;
-$input-border-width-focused: 2px !default;
-$input-line-height: 26px !default;
-$input-padding-top: 2px !default;
+$textarea-margin-bottom: 9px !default;
+
 
 $input-error-font-size: 12px !default;
 $input-error-height: 24px !default;
 $input-error-line-height: $input-error-font-size + 2px !default;
 $error-padding-top: ($input-error-height - $input-error-line-height) / 2 !default;
 
-$icon-offset: 36px !default;
+$icon-internal-padding-ratio: (2 / 24) !default; // Material system icon default padding
 
-$icon-top-offset: ($icon-offset - $input-padding-top - $input-border-width-focused) / 4 !default;
+
+$icon-height: 24px !default;
+$icon-offset: ($icon-height * 2) + 8px !default;  //56px
+$icon-internal-padding: $icon-internal-padding-ratio * $icon-height !default;
+
+$icon-height-dense: 20px !default;
+$icon-offset-dense: ($icon-height-dense * 2) + 8px !default;  //48px
+$icon-internal-padding-dense: $icon-internal-padding-ratio * $icon-height-dense !default;
+
+$icon-top-offset: $default-top-margin !default;
 
 $icon-float-focused-top: -8px !default;
 
 $input-resize-handle-height: 10px !default;
 
+// md-spacing values (horizontal spacing with icons)
+// normal     = 36px (Angular Material default)
+// wide       = 48px (Material Design Spec dense)
+// extra-wide = 56px (Material Design Spec default)
+
+$input-with-icon-padding-start: 36px !default;
+$input-with-icon-padding-start-wide: $icon-offset-dense !default;
+$input-with-icon-padding-start-extra-wide: $icon-offset !default;
+
 md-input-container {
   @include pie-clearfix();
   display: inline-block;
   position: relative;
-  padding: $input-container-padding;
-  margin: 18px 0;
+  @include rtl(padding-left, $input-container-padding-start, 0);
+  @include rtl(padding-right, 0, $input-container-padding-start);
+  min-height: $input-container-min-height;
+  margin: $input-container-margin;
   vertical-align: middle;
 
   &.md-block {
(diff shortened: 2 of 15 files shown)


clshortfuse:
Because of the way labels work now, this fixes https://github.com/angular/material/issues/7403 on IE

ThomasBurleson:
@clshortfuse  - Really EXCELLENT work here. @DevVersion and I really like the clickable cross-hair to compare alignments.

clshortfuse:
@ThomasBurleson Thanks!

You can probably tell this took a bit of work. This is really my first time playing with SCSS. From a maintainability standpoint, maybe somebody could suggest a better `md-dense` structure. There's a lot of duplicate code, but I wasn't sure how to make it use one variable for default and another if md-dense was present.

I also didn't know how to support md-dense being anywhere in the parent. Right now, you'd have to add `.md-dense` to each `input-container`, but ideally, I would like to just add `md-dense` to something like `md-content` and it'll apply to every child input-container.

Some things that might cause issues are

* `md-autocomplete`: moves `md-progress-linear` to be inside input-container, so it can have a `relative` `position` under the input box, overlapping the bottom border.
* `animate`: I added an all-encompassing default to the switch statement. This may cause issues elsewhere on other piece of code, though I personally couldn't find any. I didn't treat `font-size` like the other attributes (`height`, `width`, etc) isn't always represented in `px`. It's sometimes `pt` or `rem`.
* `md-dense`: poor code maintainability
* `md-label`: removed `scale` transformed and replaced it with `font-size` transforms. AFAIK, this is the first time this is done in angular-material. I'm not sure if browsers would have an issue with this, but I know it works fine with IE10. IE9 transforms the label, but without any animation.

You may see strange pixel numbers (for example `10px` `margin-top` instead of `8px`) but that's because the Material Design spec has 0px font leading for text fields. I couldn't set line-height to anything under 100%, so I had to bake in an approximation. The alternative is to change `box-sizing` to `border-box` and manually override `height` forcefully match `font-size`. The problem with this is the need of JS needed to implement this and possible rendering issues with browsers. The margin of error is rather small. For example, `md-display-1` will have have a height of 93px instead 92px.

clshortfuse:
Just tested on Firefox.

It needs

    md-input-container label {
      left: 0px;
    }

IE10,11,Edge work fine. Go figure.

clshortfuse:
squash, rebased and tested on Firefox, IE10, IE11, Safari and Chrome

I also added RTL and Ellipsis examples. No regressions spotted.

DevVersion:
@clshortfuse Just FYI: It is really difficult to test and review a PR, which contains so many sub-fixes. Separated PR's are more appropriated in my opinion.

> Fixes font-size issue with #7477

How does that fix #7477?

clshortfuse:
@DevVersion  You're right. It doesn't. That's my mistake.

I know it's a huge batch of changes, but changing one thing normally breaks another. There are a few things I can separate. I could separate 3 ways

* md-dense
* counter space
* md-no-border on md-select

clshortfuse:
rebased again, I also added blank checkbox checks to the codepen. It should now center properly in a table.

ThomasBurleson:
@clshortfuse  - I am excited to review this next week. Thanks for your efforts with this.

ThomasBurleson:
@crisbeto - can you review and test ?

crisbeto:
Sure, but it won't be until tomorrow at least. @clshortfuse can you rebase meanwhile?

clshortfuse:
Rebased. it took me a bit because I had to readjust for `.md-resize-spacer` but it's all working now.

I did find a bug with `.md-resize-spacer`, but that's both in master and my PR. If the parent height doesn't have a max-height, there are some issues. You'll have to check `textarea` resizing with this codepen.

http://codepen.io/shortfuse/pen/aZOvVg 

On another note, node-sass fails on node v6.20 even with `master`. I had to go back to v5.10.1 (what I used to build it). At least now I can run karma tests on ChromeOS without issues, so I can check things much faster now.

ThomasBurleson:
@clshortfuse - 

* we are aware of the Node6 issue... but have not investigated that yet.
* rebasing is always a pain. Thank you for this effort
* @crisbeto - fyi on the the textarea resizing issue ^.

crisbeto:
Ah it's not accounting for the scrolling position. I'll file a separate issue for it. Shouldn't stop this from going through, though.

ThomasBurleson:
@crisbeto - one major concern with this PR... does it affect any public CSS that we currently have?
*  If not (and new CSS is used and/or private CSS is modified), then a merge can happen quickly. 
*  If public CSS is modified, we must identify impacts and possible alternate solutions.

src/components/input/input.scss
```diff
@@ -215,43 +267,47 @@ md-input-container {
   }
 
   .md-char-counter {
-    @include rtl(text-align, right, left);
-    @include rtl(padding-right, $input-container-padding, 0);
-    @include rtl(padding-left, 0, $input-container-padding);
+    //@include rtl(text-align, right, left);
+    //@include rtl(padding-right, $input-container-padding, 0);
```
crisbeto: Is this supposed to stay commented?

src/components/input/input.scss
```diff
@@ -336,9 +407,9 @@ md-input-container {
     }
 
     > md-icon {
-      top: $icon-top-offset;
-      @include rtl(left, 2px, auto);
-      @include rtl(right, auto, 2px);
+      //top: $icon-top-offset;
+      //@include rtl(left, 2px, auto);
+      //@include rtl(right, auto, 2px);
```
crisbeto: Should this be commented out as well?

src/components/input/input.js
```diff
@@ -685,6 +687,7 @@ function placeholderDirective($log) {
 
       inputContainer.element.addClass('md-icon-float');
       inputContainer.element.prepend(placeholder);
+      inputContainer.element.toggleClass('md-input-has-label', true);
```
crisbeto: This is the same as `inputContainer.element.addClass('md-input-has-label');`. You can also combine it with the `addClass` above into something like:

```js
inputContainer.element
    .addClass('md-icon-float md-input-has-label')
    .prepend(placeholder);
```

src/components/input/input.js
```diff
@@ -795,6 +798,8 @@ function ngMessagesDirective() {
     if (attrs.mdAutoHide == 'false' || hasVisibiltyDirective(attrs)) {
       element.toggleClass('md-auto-hide', false);
     }
+
+    inputContainer.element.toggleClass('md-input-has-messages', true);
```
crisbeto: This is the same as `inputContainer.element.addClass('md-input-has-messages');`.

src/components/input/input.js
```diff
@@ -140,6 +140,8 @@ function labelDirective() {
       if (!containerCtrl || attr.mdNoFloat || element.hasClass('_md-container-ignore')) return;
 
       containerCtrl.label = element;
+      containerCtrl.element.toggleClass('md-input-has-label', true);
```
crisbeto: This one's also the equivalent of `containerCtrl.element.addClass('md-input-has-label');`

crisbeto:
@clshortfuse Here's some notes after going through the UI:

1. There seems to be quite a lot of space between checkboxes and text. This isn't the case for radio buttons.
angular_material_-_demos__checkbox_-_google_chrom_2016-06-03_18-33-51

2. This might be an issue with the demo, rather than the input, but these 2 are basically touching in your fork, but not in master
angular_material_-_demos__autocomplete_-_google_c_2016-06-03_18-32-30

3. There seem to be some alignment issues in the input demo, like the datepicker not aligning with the other inputs and the full-width ones not extending to the edge of the screen.
untitled-1_ _100 _ layer_1 _rgb8 __2016-06-03_18-3_2016-06-03_18-43-53

4. The input that says "Mountain View" doesn't extend all the way to the edge, it does it in master.
angular_material_-_demos__input_-_google_chrome_2016-06-03_18-41-07

5. There's some jumpiness when the errors get toggled in the input demo.
jump

clshortfuse:
@crisbeto The spacing is actually according to spec. Text fields with icons, checkboxes, and radio buttons should have the same spacing. I can double check, but I think it's radio buttons that don't follow spec.

2 and 4 are related and I think it has to do with last-of-item or last-child.

The jumpiness is because things are relatively positioned now, making height dynamic instead of fixed. Let me see what I can come up with in the demo. IIRC, the demo is an example of a customizing md-input-container element.

crisbeto:
Still that seems like an awful lot of space, though. For example in the toast demo, you can't tell which label belongs to which checkbox:

angular_material_-_demos__toast_-_google_chrome_2016-06-03_20-51-02

Did you find any references to checkboxes + text in the Material spec, because I didn't?

clshortfuse:
Yeah, I can understand. I would say padding-right needs to be reworked, but the padding between the checkbox and text item is visible here:

https://material-design.storage.googleapis.com/publish/material_v_8/material_ext_publish/0Bzhp5Z4wHba3dUVPRTdfUXlZWnM/components_textfields_single1.png

https://material-design.storage.googleapis.com/publish/material_v_8/material_ext_publish/0B0iWdZjcphyMQ1ctVF92SlM2d1E/components_fp_17-.png

This isn't in a list circumstance, but just general usage.  We can write special CSS something for md-checkbox inside md-input-container. I think the idea is when using touch interaction, your finger shouldn't obscure any text. I honestly can't find any usage of md-checkbox in a horizontal configuration in the spec, so I'm not sure there's any guidelines with horizontal padding. 

I've reworked md-checkbox to not take up horizontal space when you don't assign text to it, so technically, any user can create a secondary text item for md-checkbox and pad it as far as they want.

crisbeto:
Thing is that spacing might make sense on mobile (which also seems to be the main focus of the spec), but Material is also being used on larger screens. For example, we're not following the spec for the datepicker, precisely because it doesn't work very well on PC.

clshortfuse:
I don't want to dwell on the point too much, but the way it's set up, it matches horizontal padding for text fields with icons. Text fields don't spacing on mobile or desktop. My thoughts on it being like that because may not be accurate since you don't really touch icons on input fields. I think Google just wanted them to match up. Even so, my use of material-design is for both mobile phones and touchscreen desktops (24" Chromebases), so you can't rule out touch on PCs.

Regardless, as I said, I could now rework any md-checkbox to any padding I want by putting it blank, which I couldn't do before. The value is $checkbox-text-margin which I set it to 56px, which is what I measured and matches text fields. For dense layouts it would be have to be 48px to match text fields. I think 36px would be like it is in `master` branch. It's a single line and not that big of an issue.

I don't have an issue changing it back personally. I would like users to have the ability to match Google spec 1:1 without workarounds/hacks though. Maybe an attribute called md-spacing="narrow" to force spacing to 36px?

crisbeto:
I'd say something like this should be opt-in, rather than opt-out, otherwise we'd be breaking a lot of layouts.

clshortfuse:
Got it. Any suggestions on naming scheme? `very-dense` (36px)  `dense` (48px) `wide` (56px) ?

crisbeto:
Well I'd say the default stays the way it is in master, then something either like `wide` for 48 and `extra-wide` for 56, or going for a scheme lake the layout attributes, e.g. `md`, `xl` etc. I'm not too good at naming things either.

clshortfuse:
@crisbeto, just need to fix the demo and take a look at md-datepicker

clshortfuse:
@crisbeto `master` branch just broke RTL support. I'm trying to pinpoint where.

AdriVanHoudt:
does this include the missing errors-spacer for md-select?

clshortfuse:
@crisbeto Fixed the positioning. Technically, `bottom: 7px` works fine as in the original demo, but top: 67px works even better since md-input dynamically resizes.

crisbeto:
@clshortfuse I'll look at it when I have time. Also we talked about it with @ThomasBurleson and it might be safer if we split these changes up into a few smaller PRs. Pushing something like this through all at once can be unpredictable.

clshortfuse:
@crisbeto I think I hit a wall with `padding-right`. This is an issue that occurs with `master` but I want to get it right here as well. Essentially, I'm trying to write some sort of detection in CSS that if the md-input-container is in a column, don't add horizontal padding. If the md-input-container is in a row configuration and it's not the last child, add padding. There's a problem that I can't figure out what's going on with breakpoints. That seems like the last issue. I think I could manage it, but the SCSS would be very wordy since I would have to start attaching @media checks for every type of layout-column.

crisbeto:
What about the other way around: always adding it, but overriding it from the column class?

clshortfuse:
I think I misunderstood something in the `layout` spec. I think it will only have one class at time. I think I can do something like this: https://github.com/angular/material/blob/d86efaf81bdc1f85e290f202005f56570ad5dea3/src/components/radioButton/radio-button.scss#L112

clshortfuse:
I think I'm done now. I found a few bugs in master but I'm leaving this as is for now. Here's are some horizontal tests.

Horizontal Tests
Master: http://codepen.io/shortfuse/full/JKGdQv/
SpecFix: http://codepen.io/shortfuse/full/gMPYjE/

You don't need `.md-inline-form` unless you unless you have an object without a `label` or if you change `layout-align`. If all (or none) of the elements use labels and you leave layout-align alone, `.md-inline-form` isn't needed.

Hopefully I can start chopping this up into smaller PRs.

crisbeto:
@clshortfuse I went through it again and it looks good. The only issue I found was the alignment on multi-selects:

angular_material_-_demos__select_-_google_chrome_2016-06-09_21-57-52

clshortfuse:
@crisbeto Are you sure it's not like that in master? It looks like it's lining up with `Corn`.

I also split `md-select` only to #8716 

How do I test with the Angular Material home page? I've been testing with 

    gulp watch-demo -c select

clshortfuse:
@crisbeto oh, you mean the checkbox alignment. Yeah, that's from me changing checkbox from 50%. Let me play around with it.

clshortfuse:
@crisbeto I reverted it back and pushed the change to #8713 

I structured the PRs to work independently.

Once they're all good, I can work on a PR for just `md-input` based on the other ones. The only change I'd have to make is readjust `.md-inline-form`. Then, once `md-input` is fine, I'll make the PR for `md-dense`.

clshortfuse:
I didn't want to redo `md-datepicker` but I had to in order to rebase with new changes. I also had to do a lot to rebase `md-select`. Hopefully it's all good now. Will test on IE later.

clshortfuse:
Rebase in #9344

Issue #7878, opened 2016-04-04
mdPanel: Add intelligent positioning to the panel


KarenParker:
This task involves:

- determining if it is offscreen and reposition if possible by flipping it
- accepting a priority list of positions
   - try each position and find the first one that is positioned onscreen
   - last is a default position

ErinCoughlan:
Update: Multiple positions are now accepted. The remaining task in this issue to force position the panel if none of the provided options work.

PR #8211, opened 2016-04-25
update(panel): refactor positioning to allow multiple.

gmoothart:
API changes needed for #7878 

@ErinCoughlan @jelbourn @DerekLouie @KarenParker @ThomasBurleson

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -338,14 +338,12 @@ angular
  * Overlapping the panel with an element:
  * `new MdPanelPosition()
  *     .relativeTo(someElement)
- *     .withPanelXPosition($mdPanel.xPosition.ALIGN_START)
- *     .withPanelYPosition($mdPanel.yPosition.ALIGN_TOPS);`
+ *     .addPanelPosition($mdPanel.xPosition.ALIGN_START, $mdPanel.yPosition.ALIGN_TOPS);`
  *
  * Aligning the panel with the bottom of an element:
  * `new MdPanelPosition()
  *     .relativeTo(someElement)
- *     .withPanelXPosition($mdPanel.xPosition.CENTER)
- *     .withPanelYPosition($mdPanel.yPosition.BELOW);`
+ *     .addPanelPosition($mdPanel.xPosition.CENTER, $mdPanel.yPosition.BELOW);
  */
 
 /**
@@ -438,10 +436,15 @@ angular
 
 /**
  * @ngdoc method
- * @name MdPanelPosition#withPanelXPosition
+ * @name MdPanelPosition#addPanelPosition
  * @param {string} xPosition
+ * @param {string} yPosition
  * @description
- * Sets the x position for the panel relative to another element.
+ * Sets the x and y position for the panel relative to another element. Can be
+ * called multiple times to specify an ordered list of panel positions. The
+ * first position which allows the panel to be completely on-screen will be
+ * chosen; the last position will be chose whether it is on-screen or not.
+ *
  * xPosition must be one of the following values available on
  * $mdPanel.xPosition:
  *
@@ -459,14 +462,7 @@ angular
  * C: CENTER
  * D: ALIGN_END (for LTR displays)
  * E: OFFSET_END (for LTR displays)
- */
-
-/**
- * @ngdoc method
- * @name MdPanelPosition#withPanelYPosition
- * @param {string} yPosition
- * @description
- * Sets the y position for the panel relative to another element.
+ *
  * yPosition must be one of the following values available on
  * $mdPanel.yPosition:
  *
@@ -485,6 +481,7 @@ angular
  * H: CENTER
  * I: ALIGN_BOTTOMS
  * J: ABOVE
+ * @returns {MdPanelPosition}
  */
 
 /**
@@ -1090,7 +1087,7 @@ MdPanelRef.prototype._addStyles = function() {
   return this._$q(function(resolve) {
     self._panelContainer.css('z-index', self._config['zIndex']);
     self._panelEl.css('z-index', self._config['zIndex'] + 1);
-    
+
     if (self._config['fullscreen']) {
       self._panelEl.addClass('_md-panel-fullscreen');
       resolve(self);
@@ -1436,11 +1433,8 @@ function MdPanelPosition() {
   /** @private {!Array<string>} */
   this._translateY = [];
 
-  /** @private {string} */
-  this._xPosition = '';
-
-  /** @private {string} */
-  this._yPosition = '';
+  /** @private {!Array<{x:string, y:string}>} */
+  this._positions = [];
 }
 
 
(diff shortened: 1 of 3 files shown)


src/components/panel/panel.js
```diff
@@ -438,10 +437,11 @@ angular
 
 /**
  * @ngdoc method
- * @name MdPanelPosition#withPanelXPosition
- * @param {string} xPosition
+ * @name MdPanelPosition#withPanelPosition
+ * @param {string} xpos
```
ErinCoughlan: nit: Can you keep these as `xPosition` and `yPosition` as it directs the user towards the constants with the same name?

src/components/panel/panel.js
```diff
@@ -489,6 +482,16 @@ angular
 
 /**
  * @ngdoc method
+ * @name MdPanelPosition#tryPanelPositions
+ * @description
+ * Specify an ordered list of panel positions. The first position which
+ * allows the panel to be completely on-screen will be chosen; the last position
+ * will be chose whether it is on-screen or not.
+ * @param {!Array<{y: string, x:string}>} posArray
```
ErinCoughlan: Missing @return.

src/components/panel/panel.js
```diff
@@ -489,6 +482,16 @@ angular
 
 /**
  * @ngdoc method
+ * @name MdPanelPosition#tryPanelPositions
+ * @description
+ * Specify an ordered list of panel positions. The first position which
+ * allows the panel to be completely on-screen will be chosen; the last position
+ * will be chose whether it is on-screen or not.
```
ErinCoughlan: Can you show an example of how this is used? It will be helpful in explaining the functionality of this method.

src/components/panel/panel.js
```diff
@@ -1578,60 +1578,92 @@ MdPanelPosition.prototype.relativeTo = function(element) {
 
 
 /**
- * Sets the x position for the panel relative to another element.
- * xPosition must be one of the MdPanelPosition.xPosition values.
- *
- * @param {string} xPosition
- * @returns {MdPanelPosition}
+ * Sets the x and y postion for the panel relative to another element.
+ * @param {string} x must be one of the MdPanelPosition.xPosition values.
```
ErinCoughlan: nit: Param names should match public documentation.

src/components/panel/panel.js
```diff
@@ -1578,60 +1578,92 @@ MdPanelPosition.prototype.relativeTo = function(element) {
 
 
 /**
- * Sets the x position for the panel relative to another element.
- * xPosition must be one of the MdPanelPosition.xPosition values.
- *
- * @param {string} xPosition
- * @returns {MdPanelPosition}
+ * Sets the x and y postion for the panel relative to another element.
```
ErinCoughlan: typo: postion -> positions

src/components/panel/panel.js
```diff
@@ -338,14 +338,13 @@ angular
...
- *     .withPanelYPosition($mdPanel.yPosition.ALIGN_TOPS);`
+ *     .withPanelPosition($mdPanel.xPosition.ALIGN_START,
+ *        $mdPanel.yPosition.ALIGN_TOPS);`
  *
  * Aligning the panel with the bottom of an element:
  * `new MdPanelPosition()
  *     .relativeTo(someElement)
- *     .withPanelXPosition($mdPanel.xPosition.CENTER)
- *     .withPanelYPosition($mdPanel.yPosition.BELOW);`
+ *     .withPanelPosition($mdPanel.xPosition.CENTER, $mdPanel.yPosition.BELOW);
```
ErinCoughlan: Please format this the same as the above example.

Also add an example here that demonstrates the tryPanelPositions function.

src/components/panel/panel.js
```diff
@@ -489,6 +482,16 @@ angular
 
 /**
  * @ngdoc method
+ * @name MdPanelPosition#tryPanelPositions
```
ErinCoughlan: Can we make the array for the users so that they can follow the same format as `withPanelPosition`?

```
new MdPanelPosition()
    .relativeTo('myButton')
    .tryPanelPosition(xPosition.ALIGN_END, yPosition.ALIGN_TOPS)
    .tryPanelPosition(xPosition.OFFSET_START, yPosition.ALIGN_TOPS)
    .tryPanelPosition(xPosition.CENTER, yPosition.CENTER);
```

I think that `withPanelPosition` is really an alias for `tryPanelPosition` but with only one option.

I'm also afraid the user will end up writing the same xPosition and try the vertical ones, and then only trying the horizontal ones if something fails, which makes a lot of duplication here. Basically, if a user "just wants the panel centered and on screen", they end up trying a lot of options before they find something that works. Essentially, I don't know if there's enough feedback to say "The panel fits in the xPosition, so I will only try to change the vertical ones."

src/components/panel/panel.js
```diff
@@ -1578,60 +1578,92 @@ MdPanelPosition.prototype.relativeTo = function(element) {
...
 
 /**
- * Sets the x position for the panel relative to another element.
- * xPosition must be one of the MdPanelPosition.xPosition values.
- *
- * @param {string} xPosition
- * @returns {MdPanelPosition}
+ * Sets the x and y postion for the panel relative to another element.
+ * @param {string} x must be one of the MdPanelPosition.xPosition values.
+ * @param {string} y must be one of the MdPanelPosition.yPosition values.
```
ErinCoughlan: Missing @returns. This should still chain to allow for offsets. Same below.

src/components/panel/panel.js
```diff
@@ -1578,60 +1578,92 @@ MdPanelPosition.prototype.relativeTo = function(element) {
...
- *
- * @param {string} yPosition
- * @returns {MdPanelPosition}
+ * Specify an ordered list of panel positions. The first position which
+ * allows the panel to be completely on-screen will be chosen; the last position
+ * will be chose whether it is on-screen or not.
+ * @param {!Array<{x:string, y:string}>} pos
  */
-MdPanelPosition.prototype.withPanelYPosition = function(yPosition) {
+MdPanelPosition.prototype.tryPanelPositions = function(posArray) {
```
ErinCoughlan: Please keep using position for consistency and clarity.

src/components/panel/panel.js
```diff
@@ -1578,60 +1578,92 @@ MdPanelPosition.prototype.relativeTo = function(element) {
...
+    this._validateXPosition(pos.x);
+    this._validateYPosition(pos.y);
+  };
+
+  this._positions = this._positions.concat(posArray);
+  return this;
+};
+
+
+MdPanelPosition.prototype._validateYPosition = function(ypos) {
```
ErinCoughlan: Missing jsDoc.

src/components/panel/panel.js
```diff
@@ -1578,60 +1578,92 @@ MdPanelPosition.prototype.relativeTo = function(element) {
...
   var positionValues = [];
   for (var key, i = 0; key = positionKeys[i]; i++) {
     var position = MdPanelPosition.yPosition[key];
     positionValues.push(position);
-    if (position === yPosition) {
-      this._yPosition = yPosition;
-      return this;
+
+    if (position === ypos) {
+      return;
```
ErinCoughlan: `return true;`? Or does the above line not need to return a boolean.

src/components/panel/panel.js
gmoothart: We can, I just thought it would be confusing to have two different names for the same functionality.

The duplication doesn't bother me, since this is a manual fallback. Another part of #7878 is to try to do the axis-flipping automatically

gmoothart: Addressed for now by using just the one function `withPanelPosition`, which can be called multiple times.

gmoothart:
PTAL. Comments addressed and API simplified to just one method `addPanelPosition` that can be called multiple times.

src/components/panel/panel.js
```diff
@@ -1765,7 +1776,10 @@ MdPanelPosition.prototype._calculatePanelPosition = function(panelEl) {
   var targetRight = targetBounds.right;
   var targetWidth = targetBounds.width;
 
-  switch (this._xPosition) {
+  // TODO(gmoothart): intelligently pick the first on-screen position.
+  var pos = this._positions[0] || {};
```
ErinCoughlan: Watch variable names pos -? position

ErinCoughlan:
Please add a test for multiple positions, verifying that they are added to an array.

Then LGTM.

PR #8205, opened 2016-04-25
feat(switch): add attribute to invert the switch

Lordnoname:
- Adds an attribute which invertes the switch when set to true

Fixes #7889

PR #8215, opened 2016-04-26
feat(switch-refactor): add classes md-full-width and md-invert

ReneNeubert:
- refactor css styling for better support alignment with text (also with empty label), tested on IE11, Edge, Chrome, Safari, Firefox
- add class md-full-width to change inline-flex to flex
- add class md-invert to change the places of _md-container and _md-label


DevVersion:
The `md-invert` enhancement is a duplicate of #8205.

ThomasBurleson:
Duplicate.

ReneNeubert:
pure accident... but only in css and with text correct alignment

PR #8227, opened 2016-04-26
fix(panel): Fixes custom animations in the demo to account for opacity.

ErinCoughlan:
@gmoothart Please review.


gmoothart:
:+1:

PR #8228, opened 2016-04-27
docs(sidenav): leftNav text isn't hiding when it should

Splaktar:


src/components/sidenav/demoBasicUsage/index.html
```diff
@@ -4,11 +4,10 @@
   <section layout="row" flex>
 
     <md-sidenav
-        class="md-sidenav-left"
+        class="md-sidenav-left md-whiteframe-4dp"
```
DevVersion: Not sure, why you've changed the directive to the class.

`md-whiteframe="4"` should work fine and also shows the API in use.

Splaktar: That seems to be covered in the md-whiteframe docs. Do we need to cover both usages of md-whiteframe in the md-sidenav docs? It seems like it only adds to the confusion and complexity of learning about sidenavs.

Splaktar: @DevVersion updated only to fix text hiding. Both forms of `md-whiteframe` API usage (attribute and class) preserved.

DevVersion: Yeah, looks fine now :+1: thx.

PR #8238, opened 2016-04-27
fix(panel): Remove useless button in dialog demos.

ErinCoughlan:
@KarenParker @gmoothart @DerekLouie Please review.

There is no action associated with the button, so the panel seems broken when it's clicked.


gmoothart:
:+1:

PR #8240, opened 2016-04-27
update(panel): Small refactor to make updating position with scroll easier.

ErinCoughlan:
Instead of keeping track if the position was calculated already, move the calculation before all the getters.

@gmoothart @KarenParker @DerekLouie  Please review.


gmoothart:
:+1:

KarenParker:
:+1:

PR #8244, opened 2016-04-27
fix(dialog): don't clobber md-dialog id

jelbourn:


src/components/dialog/dialog.spec.js
```diff
@@ -168,6 +168,27 @@ describe('$mdDialog', function() {
       expect(content.id).toBe('dialogContent_' + dialog[0].id);
     }));
 
+    it('should not clobber the id from `md-dialog` when there is no content', inject(function ($mdDialog, $rootScope, $document) {
```
topherfangio: LGTM except the end of this description "when there is no content". Shouldn't it say something like "when the user provides an ID"?

jelbourn: The bug is only when the dialog doesn't have an `md-dialog-content`. The fact that the `md-dialog` has an ID is implied by "clobbers" (i.e., overwrites).

topherfangio: Gotcha; I didn't catch that the test didn't have the `<md-dialog-content>`. Thanks for clarifying :wink:

PR #8265, opened 2016-04-28
feat(panel): Update the panel position on scroll.

### LINKED ISSUE #7897: mdPanel: Reposition on scroll.

ErinCoughlan:
Fixes #7897

@KarenParker @gmoothart @DerekLouie @ThomasBurleson @jelbourn Please review.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -766,6 +766,12 @@ function MdPanelRef(config, $injector) {
   /** @private @const {!angular.$log} */
   this._$log = $injector.get('$log');
 
+  /** @private @const {!angular.$window} */
+  this._$window = $injector.get('$window');
+
+  /** @private @const {!Function} */
+  this._$$rAF = $injector.get('$$rAF');
+
   // Public variables.
   /**
    * Unique id for the panelRef.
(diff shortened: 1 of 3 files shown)


KarenParker:
👍

DerekLouie:
LGTM

src/components/panel/panel.js
```diff
@@ -1288,6 +1305,31 @@ MdPanelRef.prototype._configureClickOutsideToClose = function() {
 
 
 /**
+ * Configures the listeners for updating the panel position on scroll.
+ * @private
+*/
+MdPanelRef.prototype._configureScrollListener = function() {
+  var updatePosition = angular.bind(this, this._updatePosition);
+  var debouncedUpdatePosition = this._$$rAF.throttle(updatePosition);
```
jelbourn: What's the timer on this? And is `throttle` actually performing a debounce?

ErinCoughlan: `throttle` is doing the debounce, once per animation frame. It uses the last event that happened before the frame.

ThomasBurleson: Yes - we forget to rename it. (sigh)

src/components/panel/panel.js
```diff
@@ -1311,6 +1328,31 @@ MdPanelRef.prototype._configureClickOutsideToClose = function() {
...
+      debouncedUpdatePosition();
+    }
+  };
+
+  // Add listeners.
+  this._$window.addEventListener('scroll', onScroll, true);
+
+  // Queue remove listeners function.
+  this._removeListeners.push(function() {
+    self._$window.removeEventListener('scroll', onScroll);
```
ThomasBurleson: I think you are missing an argument. If you use the *capture* phase in the `addEventListener( )`, then you must match when removing:

```js
self._$window.removeEventListener('scroll', onScroll, true);
```

ErinCoughlan:
Fixed the removeListener. PTAL.

ErinCoughlan:
LGTM from @ThomasBurleson via Slack.

PR #8266, opened 2016-04-28
perf(tooltip): clean up improvements

### LINKED ISSUE #8198: md tooltip, memory leak
Hi,

there is a demo: http://codepen.io/djFD/full/LNJEaw/

how to reproduce.
by default all initial settings of the demo are ready for test (clean leftovers: ON, tooltips-kind: MD, add md-visible attribute: ON, global_visibility: ON)

Strarting from this initial state just press the 'Create fresh buttons' a few times. When the button is clicked: 1) md-buttons get removed using ```element.remove()```; 2) their lost tooltips (see #8033) are being removed too (thus we need global visibility to be able find then and remove); 3) new buttons get created and inserted into document

So after any click we expect to have exactly two buttons and two tooltips. Now toggle global visibility flag OFF, then ON. How many tooltips do you see? Every button click adds 2 tooltips, so we get 2 x N leaked tooltips, not just lost but exactly leaked, as we removed them entirely from DOM using ```.remove()``` op and nobody else now is keeping the references to them except browser's JS engine

Thanks

crisbeto:
* Fixes the toolip not being removed if the element to which it is attached gets removed from the DOM.
* Fixes the tooltip element sometimes not being removed when it's scope gets destroyed.
* Fixes some event handlers that weren't being removed when the tooltip is destroyed.

Fixes #8198.

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -98,16 +98,10 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
     }
 
     function configureWatchers () {
-      scope.$on('$destroy', function() {
-        scope.visible = false;
-        element.remove();
-        angular.element($window).off('resize', debouncedOnResize);
-      });
-
       if (element[0] && 'MutationObserver' in $window) {
         var attributeObserver = new MutationObserver(function(mutations) {
           mutations
-            .forEach(function (mutation) {              
+            .forEach(function (mutation) {
               if (mutation.attributeName === 'md-visible') {
                 if (!scope.visibleWatcher)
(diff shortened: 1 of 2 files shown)


clshortfuse:
This isn't part of your pull request, but it looks like the MutationObservers aren't disconnected on destroy.

crisbeto:
I think they are, [take a look here](https://github.com/angular/material/pull/8266/files#diff-0732597cd96cb9d201490dfa8fe61b4bR206), there's actually 2 scope `$destroy` listeners.

clshortfuse:
I saw that, but both variables have the same name and initiliazed inside child functions.

crisbeto:
Oh damn, I didn't notice that there's two observers. Thanks for the tip, I'll disconnect the other one as well.

PR #8278, opened 2016-04-29
update(tooltip): Remove will-change css property.

ErinCoughlan:
This property makes the panel disappear when hovering over the tooltip in Chrome on Linux. By removing it, the panel remains on screen.

@ThomasBurleson PTAL.


ThomasBurleson:
yes. definitely remove. http://caniuse.com/#search=will-change
@ErinCoughlan  - lgtm.

PR #8285, opened 2016-04-30
fix(backdrop): adjust the backdrop height when the viewport resizes

### LINKED ISSUE #8155: Window resize doesn't resize backdrop when document body is position fixed.
When the window is a smaller size and I bring up a dialog (with a backdrop), when we maximize the window the backdrop is cut to the viewport size.

capture

crisbeto:
The backdrop's height was being set only on initialization, but this meant that it would be clipped if the viewport got resized.

Fixes #8155.

### DIFF
--- a/src/components/backdrop/backdrop.js
+++ b/src/components/backdrop/backdrop.js
@@ -19,7 +19,7 @@
 
 angular
   .module('material.components.backdrop', ['material.core'])
-  .directive('mdBackdrop', function BackdropDirective($mdTheming, $animate, $rootElement, $window, $log, $$rAF, $document) {
+  .directive('mdBackdrop', function BackdropDirective($mdTheming, $mdUtil, $animate, $rootElement, $window, $log, $$rAF, $document) {
     var ERROR_CSS_POSITION = "<md-backdrop> may not work properly in a scrolled, static-positioned parent container.";
 
     return {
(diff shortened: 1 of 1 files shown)


src/components/backdrop/backdrop.js
```diff
@@ -66,6 +71,13 @@ angular
         }
       });
 
+      function resize() {
```
ThomasBurleson: *  should we not also check and adjust the `width` ? 
*  what about *margins* ?

crisbeto: * The width is always 100%. 
* I'm not sure about margins, I just moved the logic to a separate function and attached it to the viewport resize. I don't see a situation where the backdrop might have a margin, though.

ThomasBurleson: @crisbeto - my apologies. I am thinking about the dialog not the backdrop. Above is for Backdrop. So disregard my comments plz.

ThomasBurleson:
@crisbeto lgtm

src/components/backdrop/backdrop.js
FDIM: Out of curiosity, why simply setting height to 100vh is not an option?

crisbeto: Viewport units seem to be problematic on iOS.

FDIM: Yes, but on old version. It is fixed on iOS 8 and above as far as I know. Lower versions if still important can be targeted with media queries: https://gist.github.com/BenMorel/e9e34c08360ebbbd0634 (link taken from caniuse)

crisbeto: 1. That seems like a lot of CSS for something so simple.
2. My intention here was just to fix the issue, not refactor the backdrop. If it works like this, it probably has a good reason to. 
3. I can't really see viewport units being used anywhere else in the project.

FDIM: Yes, a lot of CSS - but for support which I don't think is needed anymore. 
Point 3 is interesting, but maybe for legacy reasons. Anyway, I won't bother you anymore :)

crisbeto: No prob, feedback is always welcome.

PR #8291, opened 2016-05-01
fix(input): duplicate placeholders and data bindings aria-label

### LINKED ISSUE #8251: mdInput duplicates the label while bind placeholder
I realized that if we want to bind the placeholder, the automatic label generated get's duplicated on Chrome and Safari on OSX.

My code example:

```javascript
$scope.label = 'Date of Birth';
```

```html
<md-input-container>
    <input type="text" placeholder="{{label}}">
</md-input-container>
````

Here is a GIF of this bug:
material-placeholder-duplicated

Angular 1.5.3
Angular Material 1.1.0-rc4
https://codepen.io/leocaseiro/pen/GZYZxr

### LINKED ISSUE #8377: Input field labels appear fuzzy when using md-input container + ng-repeat + placeholder + template variable in placeholder
**Actual behavior**:
-  What is the issue? The combination seems to produce multiple placeholder values superimposed on one another producing a fuzzy look. It seems to happen when:

1. Using a placeholder value,

2. ng-repeat the fields, and

3. there is a template variable in the placeholder (in the demo using {{ $index }} in placeholder)

Removing any of these things causes the behavior to return to normal. This doesn't happen when using a label above the field and no placeholder or when using both a label and placeholder. I came across this behavior when using a button to dynamically add fields with ng-repeat. 

-  CodePen Demo which shows your issue : http://codepen.io/a12k/pen/rePGMm
-  Details: As you can see, all the field labels look fuzzy. When the field is clicked, one label stays on the field and another which appears to have been superimposed on top animates to above the field. 

Looks to be similar to this older issue which was solved: https://github.com/angular/material/issues/4462 

**Additional Information**:
-  Browser Type: Chrome, Firefox, Internet Explorer 11
-  OS: Windows 10
-  Stack Traces:

----

  Add more lines if needed.

### PR DESCRIPTION
**Pull request by crisbeto**
* Fixes the element label being duplicated, if it has a data binding in the placeholder attribute. This was caused by Angular re-adding the placeholder attribute, even though it was removed by the Material placeholder directive.
* Fixes the aria-label getting the raw data binding (e.g. `aria-label="{{expression}}"`), when it's gets taken from a dynamic placeholder.

Fixes #8251.
Fixes #8377.

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -308,7 +308,7 @@ function inputTextareaDirective($mdUtil, $window, $mdAria, $timeout, $mdGesture)
     element.after(errorsSpacer);
 
     if (!containerCtrl.label) {
-      $mdAria.expect(element, 'aria-label', element.attr('placeholder'));
+      $mdAria.expect(element, 'aria-label', attr.placeholder);
     }
 
     element.addClass('md-input');
@@ -660,15 +660,22 @@ function mdMaxlengthDirective($animate, $mdUtil) {
   }
 }
 
-function placeholderDirective($log) {
+function placeholderDirective($compile) {
   return {
     restrict: 'A',
     require: '^^?mdInputContainer',
     priority: 200,
-    link: postLink
+    link: {
+      // Note that we need to do this in the pre-link, as opposed to the post link, if we want to
+      // support data bindings in the placeholder. This is necessary, because we have a case where
+      // we transfer the placeholder value to the `<label>` and we remove it from the original `<input>`.
+      // If we did this in the post-link, Angular would have set up the observers already and would be
+      // re-adding the attribute, even though we removed it from the element.
+      pre: preLink
+    }
   };
 
-  function postLink(scope, element, attr, inputContainer) {
+  function preLink(scope, element, attr, inputContainer) {
     // If there is no input container, just return
     if (!inputContainer) return;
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: src/components/input/input.js [hunk 1]
```diff
@@ -289,7 +289,11 @@ function inputTextareaDirective($mdUtil, $window, $mdAria, $timeout) {
     element.after(errorsSpacer);
 
     if (!containerCtrl.label) {
-      $mdAria.expect(element, 'aria-label', element.attr('placeholder'));
+      // It needs to wait for the next tick, in order to allow any
+      // expressions in the placeholder to be evaluated.
+      $mdUtil.nextTick(function() {
+        $mdAria.expect(element, 'aria-label', element.attr('placeholder'));
+      }, false);
```
DevVersion: The next tick timeout is not necessary.

```js
    if (!containerCtrl.label) {
      $mdAria.expect(element, 'aria-label', attr.placeholder);
    }
```

Using the interpolated attribute is working as well.

crisbeto: I thought so too, but try looking at the `aria-label` on the fiddle from the issue: https://codepen.io/leocaseiro/pen/GZYZxr

DevVersion: Oh it works fine for me, when I checked out your changes and modified it to `attr.placeholder`.

Patched: http://codepen.io/DevVersion/pen/reQepZ
HEAD: http://codepen.io/DevVersion/pen/BKGKYp

crisbeto: Ah, I didn't notice that you used the `attr` object, I'll give it a shot.

crisbeto: Thanks, works the same. I'll update the PR.

DevVersion: Great, no problem, thanks for giving a shot ;)

ThomasBurleson:
@crisbeto - please rebase.

crisbeto:
Done @ThomasBurleson

Issue #4940, opened 2015-09-30
Performance Issues with <md-tabs> and Internet Explorer


happschinson:
Hi! 

When I have md-tabs with md-dynamic-height everything works smooth in chrome and almost smooth in firefox.

When i try it on Internet Explorer, the performance gets really bad. It gets even worse, when I add more complex content to the tabs like inputfields, images, switches etc.

See my fiddle here: https://jsfiddle.net/72fxcg1m/1/
I also asked about the issue on stackoverflow with possible solutions: 
http://stackoverflow.com/questions/32872905/angular-material-tabs-performance-with-ie

happschinson:
are there any ideas about how to solve this problem?

TimKlein87:
I just experienced the same problem. After some searching I only found one solution that turned off the animation using CSS, but tab switching was still incredibly slow in IE (11). So after some thinking I came up with the following. It looks very good and works very fast.

```html
		<section class="md-whiteframe-z2">
			<md-toolbar>
				<div class="md-toolbar-tools">
					<h3>Window</h3>
				</div>
			</md-toolbar>
			<div class="">
				<div ng-cloak>
					<md-content>
						<md-tabs md-dynamic-height>
							<md-tab label="Tab 1" ng-click="tab = 'tab1'">
								<md-content></md-content>
							</md-tab>
							<md-tab label="Tab 2" ng-click="tab = 'tab2'">
								<md-content></md-content>
							</md-tab>
							<md-tab label="Tab 3" ng-click="tab = 'tab3'">
								<md-content></md-content>
							</md-tab>
						</md-tabs>
						<ng-include ng-if="tab == 'tab1'" src="'..'"></ng-include>
						<ng-include ng-if="tab == 'tab2'" src="'..'"></ng-include>
						<ng-include ng-if="tab == 'tab3'" src="'..'"></ng-include>
					</md-content>
				</div>
			</div>
		</section>
```

ThomasBurleson:
Please provide a CodePen demo.

To expedite resolution, please always test with the latest versions:
*  Angular Material HEAD
  *  Bower Install - https://github.com/angular/bower-material/blob/master/angular-material.js
  *  Bower Install - https://github.com/angular/bower-material/blob/master/angular-material.css 
  *  RAWGIT - https://rawgit.com/angular/bower-material/master/angular-material.js
  *  RAWGIT - https://rawgit.com/angular/bower-material/master/angular-material.css
* AngularJS Latest Release versions:
[...]

PR #8293, opened 2016-05-01
fix(tabs): avoid width calculation deviations.

### LINKED ISSUE #8289: md-tabs performance (updateInkBarStyles gets executed in never ending loop)
**Actual behavior**:

**updateInkBarStyles** function inside **tabs.js** constantly gets executed in background, causing performance problems in mobile devices.

-  What is the issue? 

inside updateInkBarStyles totalWidth is always larger by exactly 2px than tabWidth, so 
` if (totalWidth > tabWidth) $mdUtil.nextTick(updateInkBarStyles, false);` always pass.

Here is a link to CodePen  that I created by modifying the existing angular material demo and just adding "md-center-tabs" to tabs options.

http://codepen.io/madani/pen/ZWmbXg 

Try to set a breakpoint inside updateInkBarStyles to see the problem for yourself.

DevVersion:
* Currently when using centered tabs, the `updateInkBarStyles` method is getting executed infinitely (even without $$rAF, sync)
 This is impacting the performance significant, because we're calculating the styles on each function call.

The problem is, that the pagination wrapper uses another rounding process than the `updateInkBarStyle` method.
Now both calculations are using the same rounding process / logic and the width's are now correct.

Fixes #8289.

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -494,7 +494,11 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
   function shouldPaginate () {
     if (ctrl.noPagination || !loaded) return false;
     var canvasWidth = $element.prop('clientWidth');
-    angular.forEach(getElements().dummies, function (tab) { canvasWidth -= tab.offsetWidth; });
+
+    angular.forEach(getElements().dummies, function (tab) {
+      canvasWidth -= tab.offsetWidth;
+    });
+
     return canvasWidth < 0;
   }
 
@@ -558,17 +562,22 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
   }
 
   /**
+   * Calculates the width of the pagination wrapper by summing the widths of the dummy tabs.
    * @returns {number}
    */
   function calcPagingWidth () {
-    var width = 1;
+    return calcTabsWidth(getElements().dummies);
+  }
 
-    angular.forEach(getElements().dummies, function (element) {
+  function calcTabsWidth(tabs) {
+    var width = 0;
+
+    angular.forEach(tabs, function (tab) {
       //-- Uses the larger value between `getBoundingClientRect().width` and `offsetWidth`.  This
       //   prevents `offsetWidth` value from being rounded down and causing wrapping issues, but
       //   also handles scenarios where `getBoundingClientRect()` is inaccurate (ie. tabs inside
       //   of a dialog)
-      width += Math.max(element.offsetWidth, element.getBoundingClientRect().width);
+      width += Math.max(tab.offsetWidth, tab.getBoundingClientRect().width);
     });
 
     return Math.ceil(width);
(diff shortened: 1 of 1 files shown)


src/components/tabs/js/tabsController.js
```diff
@@ -558,17 +562,22 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+   * Calculates the width of the pagination wrapper by summing the widths of the dummy tabs.
    * @returns {number}
    */
   function calcPagingWidth () {
-    var width = 1;
+    return calcTabsWidth(getElements().dummies);
+  }
+
+  function calcTabsWidth(tabs) {
+    var width = 0;
```
DevVersion: @robertmesserle Not sure, why the `width` was initially at `1px`?

crisbeto:
Have you tried this in IE/Edge? Maybe it would help with https://github.com/angular/material/issues/4940/ as well.

DevVersion:
@crisbeto Thanks for the idea - but unfortunately it's not related to that issue, because this bug is only appearing when using `md-center-tabs`

src/components/tabs/js/tabsController.js
ThomasBurleson: AFAIR, there was a specific reasoning for this.

ThomasBurleson:
@robertmesserle - can you review this please.

src/components/tabs/js/tabsController.js
robertmesserle: I do not recall the reason at all, unfortunately.

robertmesserle: I believe this was to account for a wrapping bug regarding pagination where the last item was wrapping to a new line without it; however, this looks like a band-aid that was hastily thrown in rather than diagnosing the root cause.

robertmesserle: I'll also note that this is added in response to issues with tabs living inside of dialogs, so you might try creating that sort of test case and see if any bugs show up.

robertmesserle:
@ThomasBurleson @DevVersion LGTM, but I'd like to confirm with a demo of tabs inside of a dialog before merging - just to confirm that `width = 1` is not necessary.

ThomasBurleson:
@DevVersion - can you create a codePen with a dialog that contains tabs ?

DevVersion:
@ThomasBurleson Sure. Here is the demo
- http://codepen.io/DevVersion/pen/VaNRYG
But unfortunately with `localhost`, since I have no possibility to host the PR changes anywhere.

DevVersion:
@topherfangio Can you take a look at this PR as well (since you recently made some calculation changes)

topherfangio:
@DevVersion Code looks good, but it seems this introduces a small issue in Safari.

In the "Dynamic Tabs" demo, if you add a new tab (I named mine "woot"), it does not properly scroll it into view. It seems to work fine in master on Safari, and it also works correctly in Chrome/Firefox with and without your PR, so I'm not sure why it's different in Safari.

If we can figure this out and get it fixed, I think it's good to go.

topherfangio:
@ThomasBurleson Should be ready for review/merge.

I chatted with @DevVersion and realized my version of Safari is older and he can't reproduce the issue I mentioned above on the latest version, so I think it's good to go.

PR #8307, opened 2016-05-02
Improve md-prevent-menu-close to wrap child Elements

PapyElGringo:
Allow the use of "md-prevent-menu-close" on parent element inside a md-menu to prevent close behavior on click.

Uses cases:
-Allow to use md-prevent-menu-close on a directive who has a md-button inside the template.

### DIFF
--- a/src/components/menu/js/menuServiceProvider.js
+++ b/src/components/menu/js/menuServiceProvider.js
@@ -289,7 +289,7 @@ function MenuProvider($$interimElementProvider) {
           do {
             if (target == opts.menuContentEl[0]) return;
             if ((hasAnyAttribute(target, ['ng-click', 'ng-href', 'ui-sref']) ||
-                target.nodeName == 'BUTTON' || target.nodeName == 'MD-BUTTON') && !hasAnyAttribute(target, ['md-prevent-menu-close'])) {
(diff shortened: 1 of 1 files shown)


PapyElGringo:
I signed it!

src/components/menu/js/menuServiceProvider.js
```diff
@@ -316,6 +316,18 @@ function MenuProvider($$interimElementProvider) {
             }
             return false;
           }
+          
+          function hasOrParentHasAnyAttribute(target, attrs){
+            if (!target) return false;
+            var parent = target.parentElement;
+            while(parent && parent.nodeName != 'MD-MENU'){
```
DevVersion: Add space between the brace and `while(`.

src/components/menu/js/menuServiceProvider.js
```diff
@@ -316,6 +316,18 @@ function MenuProvider($$interimElementProvider) {
...
+              if(hasAnyAttribute(parent,attrs)){
```
DevVersion: To be consistent, here as well please.

DevVersion:
I like the idea of inheritances for the `md-prevent-menu-close` attribute. 

But there are other approaches we should consider as well:
- Making it a directive - allows us to check for the controller (Angulars Controller inheritance)
- Walking the Event Path Up

PapyElGringo:
The changes are dones.

You are free to consider alternatives but this is actually working. And I use it in complex menu templates.

ThomasBurleson:

PR #8312, opened 2016-05-02
fix(ng-maxLength): rerender mdMaxLength when ngModel is updated in the controller

josefromsanjose:
This issue comes up when sharing one form for several models. For example a list. If you were to click on one item that displays the form and it has a value for the scope variable being counted, it works fine. But if you change to a different list item that does not have a value, mdMaxLength keeps the count from the previous model, but the input is empty.

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -527,9 +527,7 @@ function mdMaxlengthDirective($animate, $mdUtil) {
       
       //render count when model changes as well
       //rerenders when the model changes in the controller and not an event
-      scope.$watch(attr.ngModel, function(value) {
-         renderCharCount();
-      });
+      scope.$watch(attr.ngModel, renderCharCount);
       
       scope.$watch(attr.mdMaxlength, function(value) {
         maxlength = value;
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -524,7 +524,13 @@ function mdMaxlengthDirective($animate, $mdUtil) {
       element.on('input keydown keyup', function() {
         renderCharCount(); //make sure it's called with no args
       });
-
+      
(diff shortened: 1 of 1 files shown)


src/components/input/input.js
```diff
@@ -524,7 +524,13 @@ function mdMaxlengthDirective($animate, $mdUtil) {
       element.on('input keydown keyup', function() {
```
crisbeto: I believe you could achieve the same without using a watcher by doing `ngModelCtrl.$parsers.push(renderCharCount);` here. Also it would be nice if you could add a unit test for this so it doesn't get broken in the future.

josefromsanjose:
@crisbeto that was my first approach but it did not work. The problem still persists.

crisbeto:
Alright, you should still be able to reduce it to `scope.$watch(attr.ngModel, renderCharCount)`. Also can you rename your commit and PR so they match the format of the other PRs?

josefromsanjose:
@crisbeto  I renamed to PR but could not rename the commits. Moving forward I will make sure I check what commit titles look like. This is my first open source contribution.

crisbeto:
Apart from combining and renaming the commits, LGTM @ThomasBurleson

DevVersion:
I'm sorry, that I create something like a second approach, but your PR wasn't referencing the issues properly. So there is a more generic / universal solution now.

See #8351

crisbeto:
Yeah that definitely looks better.

crisbeto:
@josefromsanjose Thank you for the PR, but I'll have to close it, because @DevVersion's solution solves the problem in a better way, removes a bunch of event listeners and has unit tests.

PR #8315, opened 2016-05-03
fix(switch-drag): drag to left doesn't update switch view

ReneNeubert:
md-switch drag to left generates a second click event in checkbox directive and the view was reset. This fix resolves this.

### DIFF
--- a/src/components/checkbox/checkbox.js
+++ b/src/components/checkbox/checkbox.js
@@ -169,7 +169,8 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
       }
 
       function listener(ev) {
-        if (element[0].hasAttribute('disabled')) {
+        if (element[0].hasAttribute('disabled') || scope.mouseActive == true) {
+          scope.mouseActive = false;
           return;
         }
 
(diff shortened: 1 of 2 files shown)


DevVersion:
Duplicate of #6715

ReneNeubert:
@DevVersion 
Sorry, but thats not true! Please check out the switch demo to see the issue!

DevVersion:
@ReneNeubert I can't see a difference.

- https://github.com/angular/material/pull/6715/files#diff-2aa62f0f8092a8589a0769a12c270030R155 <br>is doing the same as your `mouseActive`, just with a shorter timeout.

- https://github.com/angular/material/pull/6715/files#diff47cd4e7edc0ec7d28ed39fef4df9f831R144<br/>is doing the same as in your PR.

Can't find any differences.

ReneNeubert:
@DevVersion You are right! Sorry! But your PR is not merged yet!?

DevVersion:
@ReneNeubert No problem ;) No it's still pending. As it's not a major bug, it has a lower priority.

ReneNeubert:
@DevVersion nice to see that we had the same approach ;-)

PR #8321, opened 2016-05-03
fix(chips): keypress handler for chips separator keys

### LINKED ISSUE #8319: md-separator-keys ignores shift, other modifiers
**Actual behavior**:
-  What is the issue? md-separator-keys ignores shift, other modifiers
-  What is the expected behavior? Only the given separator keys should be used to separate chips, not combinations with key modifiers.

-  CodePen Demo which shows your issue : https://material.angularjs.org/HEAD/demo/chips
-  Details: The official "Custom Separator Keys" demo sets up comma as separator, but also shift+comma separates. On an English/Dvorak keyboard, this is the "<" key. On a Danish keyboard, this is the "æ" key. This makes it impossible to enter chips containing these characters.

**Angular Versions**: *
-  Angular Version: 1.5.5
-  Angular Material Version: HEAD, 1.0.7

**Additional Information**:
-  Browser Type: Chrome, Firefox
-  Browser Version: Chrome 50, Firefox 45 
-  OS: Linux, OSX
-  Stack Traces:

Spring cleaned issue: #6505

vbraun:
Chips separator keys can now either be numeric (and matched in the
keydown handler) or string (and matched in the keypress handler).

Closes #8319

### DIFF
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -116,6 +116,7 @@ function MdChipsCtrl ($scope, $mdConstant, $log, $element, $timeout, $mdUtil) {
  * @param event
  */
 MdChipsCtrl.prototype.inputKeydown = function(event) {
+  console.log('keydown', event);
   var chipBuffer = this.getChipBuffer();
 
   // If we have an autocomplete, and it handled the event, we have nothing to do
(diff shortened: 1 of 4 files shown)


vbraun:
It seems that it is necessary to have both a keydown handler (to catch backspace, enter) and a keypress handler (to catch actual keys like comma).

src/components/chips/js/chipsController.js
```diff
@@ -131,19 +131,43 @@ MdChipsCtrl.prototype.inputKeydown = function(event) {
...
+  if (this.separatorKeys.indexOf(key) !== -1) {
     if ((this.hasAutocomplete && this.requireMatch) || !chipBuffer) return;
     event.preventDefault();
 
     // Only append the chip and reset the chip buffer if the max chips limit isn't reached.
     if (this.hasMaxChipsReached()) return;
 
+    // If keypress: remove the already-added separator
+    if (angular.isString(key) && chipBuffer.endsWith(key))
+      chipBuffer = chipBuffer.slice(0, chipBuffer.length - 1)
```
DevVersion: At first, thanks for the great Issue and PR.

---
What about preventing the event here? This should not add the separator?

ThomasBurleson:
Nice work!

vbraun:
Thanks. Indeed, preventing the event is enough. I've added doc + test.

ThomasBurleson:

PR #8322, opened 2016-05-03
feat(): add fastclick handler.

DevVersion:
This adds a better support for the delay of touchevents (aka `Fastclick`) 

The current implementation is not very flexible / dynamic. We're intercepting **all** events and only allow our `$material` events. 

This caused a lot of issues:
- Bad Support inside of Webviews
- Custom Events of Developers are always prevented.
- Blur / Focus. (See `md-select` is not blurring on touch devices)
- Scrolling is not working properly on list (clicks getting executed on the items etc.)

The current implementation, is a very generic fastclick handler, it filters for different conditions and accelerates the touch events, which means, that the UI feels super fast.

We are no longer preventing all extraneous events. 
- We are no longer preventing all extraneous events.<br/>
  Only the most necessary events get prevented.
- We don't have to redirect all events from `$material`.
- Better Performance on Scrolling
- Detects movements in the touch click
- Separated from the gesture service (clean; modular) 
- No longer affects browsers without the touch delay
- And lots of other conditions

**Notice**: This is still a Work in Progress, because it needs to be tested more thoroughly. (Code may change; few debug conditions are included)

### DIFF
--- a/src/core/services/gesture/fastclick.js
+++ b/src/core/services/gesture/fastclick.js
@@ -1,0 +1,125 @@
+angular
+  .module('material.core')
+  .run(attachFastclick);
+
+var isRegistered = false;
+
+function attachFastclick() {
+  if (isRegistered) return;
+  isRegistered = true;
+
+  var START_EVENTS = 'touchstart pointerdown';
+  var MOVE_EVENTS = 'touchmove pointermove';
+  var END_EVENTS = 'touchend touchcancel pointerup pointercancel';
+  var MAX_TAP_DELAY = 700;
+  var MIN_TAP_DELAY = 100;
+
+  // Register all required events on the document.
+  angular
+    .element(document)
+    .on(START_EVENTS, onTouchStart)
+    .on(MOVE_EVENTS, onTouchMove)
+    .on(END_EVENTS, onTouchEnd)
+    .on('click', onClick);
+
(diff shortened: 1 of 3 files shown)


DevVersion:
@ThomasBurleson I think this is something we should consider anytime later. It not means, we should go with that, but the gesture click hijack should be revistited IMO. It breaks scrolling etc. and makes the Sidenav Gesture unuseable.

Please take a look again (just at the description) and let me know if we should keep the PR for a while.

DevVersion:
FYI: Marking this back as deprecated, since I consider this as too risky. The bugs, which are caused by the Click Hijack are edge-cases, which means that this PR has no high priority.

We're able to revisit / reopen this PR later, once we approved this concept.

ThomasBurleson:

src/core/services/gesture/fastclick.js
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    // When the last click was quicker than the minimum tap delay, then we can skip the next click to avoid
+    // a ghost click.
+    if (event.timeStamp - lastClickTime <= MIN_TAP_DELAY) {
+      preventNextClick = true;
+      return;
+    }
+
+    if (trackingElement.tabIndex !== -1) trackingElement.focus();
```
DevVersion: This needs to support `contentEditable` and other focus indicators.

PR #8326, opened 2016-05-03
fix(sidenav): don't log an error when waiting for an instance.

### LINKED ISSUE #8308: Async lookup for sidenav instance logs an error in the console (when it should just wait until the instance is available).
From the docs:
https://material.angularjs.org/latest/api/service/$mdSidenav

// Async lookup for sidenav instance; will resolve when the instance is available
$mdSidenav(componentId).then(function(instance) {
  $log.debug( componentId + "is now ready" );
});

However, when I do that, I get an error in the logs:

```console
angular.js:13550 SideNav 'left' is not available! Did you use md-component-id='left'?
(anonymous function) @ angular.js:13550findInstance @ angular-material.js:15747(anonymous function) @ angular-material.js:15712(anonymous function) @ index.dev.html:520invoke @ angular.js:4665(anonymous function) @ angular.js:4473forEach @ angular.js:322createInjector @ angular.js:4473doBootstrap @ angular.js:1746bootstrap @ angular.js:1767gamingPlatformInitFinished @ index.dev.html:530(anonymous function) @ app-gameinvite-common-code.ts:258trigger @ angular.js:3166defaultHandlerWrapper @ angular.js:3456eventHandler @ angular.js:3444
```

Even when passing true, it still logs the error:
$mdSidenav('left', true);

Except the error logs, things work correctly: the promise is resolved correctly :)
The problem is that it spams the error logs (and i have protractor tests that make sure the error log is empty).

The problem is in this code:
```js
/**
   * Service API that supports three (3) usages:
   *   $mdSidenav().find("left")                       // sync (must already exist) or returns undefined
   *   $mdSidenav("left").toggle();                    // sync (must already exist) or returns reject promise;
   *   $mdSidenav("left",true).then( function(left){   // async returns instance when available
   *    left.toggle();
   *   });
   */
  return function(handle, enableWait) {
    if ( angular.isUndefined(handle) ) return service;

    var instance = service.find(handle); // THIS IS THE BUG: if enableWait is true, then it should use waitForInstance (and not findInstance).


Because findInstance logs an error if the instance is not ready:
function findInstance(handle) {
      var instance = $mdComponentRegistry.get(handle);
      if(!instance) {
        // Report missing instance
        $log.error( $mdUtil.supplant(errorMsg, [handle || ""]) );
```

See:

```js
   var service = {
        find    : findInstance,     //  sync  - returns proxy API
        waitFor : waitForInstance   //  async - returns promise
      };
```

crisbeto:
* The `$mdSidenav` service was logging errors, even if the option to wait for an asynchronous instance was passed. This changes it to only log if the lookup for the instance is synchronous.
* Fixes the error for an invalid sidenav instance name not being logged if the `enableWait` argument is passed.
* Adds info about the async argument of `$mdSidenav` in the docs.

Fixes #8308.

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -23,11 +23,17 @@ angular
  *
  * @description
  * `$mdSidenav` makes it easy to interact with multiple sidenavs
- * in an app.
+ * in an app. When looking up a sidenav instance, you can either look
+ * it up synchronously or wait for it to be initializied asynchronously.
+ * This is done by passing the second argument to `$mdSidenav`.
  *
  * @usage
  * <hljs lang="js">
  * // Async lookup for sidenav instance; will resolve when the instance is available
+ * $mdSidenav(componentId, true).then(function(instance) {
+ *   $log.debug( componentId + "is now ready" );
+ * });
+ * // Sync lookup for sidenav instance; this will resolve immediately.
  * $mdSidenav(componentId).then(function(instance) {
  *   $log.debug( componentId + "is now ready" );
  * });
@@ -75,8 +81,9 @@ function SidenavService($mdComponentRegistry, $mdUtil, $q, $log) {
   return function(handle, enableWait) {
     if ( angular.isUndefined(handle) ) return service;
 
-    var instance = service.find(handle);
-    return  !instance && (enableWait === true) ? service.waitFor(handle) :
+    var shouldWait = enableWait === true;
+    var instance = service.find(handle, shouldWait);
+    return  !instance && shouldWait ? service.waitFor(handle) :
             !instance && angular.isUndefined(enableWait) ? addLegacyAPI(service, handle) : instance;
   };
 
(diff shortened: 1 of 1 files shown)


src/components/sidenav/sidenav.js
```diff
@@ -106,9 +113,11 @@ function SidenavService($mdComponentRegistry, $mdUtil, $q, $log) {
      * Synchronously lookup the controller instance for the specified sidNav instance which has been
      * registered with the markup `md-component-id`
      */
-    function findInstance(handle) {
+    function findInstance(handle, shouldWait) {
       var instance = $mdComponentRegistry.get(handle);
-      if(!instance) {
+
+      if(!instance && !shouldWait) {
```
DevVersion: Nitpicky as always :) - Please add space for consistency.

crisbeto: It wasn't my changes this time! Thanks for the tip, though :)

Issue #8195, opened 2016-04-24
mdAutocomplete: Progress bar is visible after clearing input when caching is turned on


edbentinck:
Since `v1.0.7` a bug was introduced whereby the loading bar is incorrectly visible after clearing the search input when caching is turned on.

NB: This is similar, but not identical, to [issue #7927](https://github.com/angular/material/issues/7927)

[Here is a codepen](http://codepen.io/edbentinck/pen/wGEwqy) based on the demo code from `v1.0.7`.

Steps to recreate the bug:
- Type 'a' into the input (notice that caching is *turned on*)
- Click the close icon ('x') inside the input

You'll see that the loading bar is activated, but never goes away.

PR #8341, opened 2016-05-04
fix(autocomplete): always hide the progressbar when clearing the input

### LINKED ISSUE #8301: md-autocomplete: infinite progress bar after clearing with cached queries
**Actual behavior**:
-  The loading bar under the input is appearing and stay for infinite amount of time after clearing the md-autocomplete.
-  The loading bar should not appear.

**Steps to reproduce the issue**: 
You can reproduce it on the md-autocomplete basic usage demo:
- Select *Simulate query* for results
- Uncheck *Disable caching of queries*
- Select one item
- Clear

**CodePen**:
[CodePen](http://codepen.io/achaussende/pen/eZQGJy)

**Angular Versions**:
I spotted this issue on my own app:
-  Angular Version: 1.4.10
-  Angular Material Version: 1.0.8 but the issue is still on master version. (cf demo)

**Additional Information**:
-  Browser Type: Chrome
-  Browser Version: 50.0.2661.94 m
-  OS: Windows 8.1

NB: The issue does not happen in release 1.0.0 of Angular Material.
It does not happen with md-no-cache="true" either.

crisbeto:
Fixes #8301, #8195.

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -775,6 +775,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     if (!$scope.noCache && cache[ term ]) {
       ctrl.matches = cache[ term ];
       updateMessages();
+      setLoading(false);
     } else {
       fetchResults(searchText);
     }


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -640,15 +640,11 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    * Clears the searchText value and selected item.
    */
   function clearValue () {
-    // Set the loading to true so we don't see flashes of content.
-    // The flashing will only occour when an async request is running.
-    // So the loading process will stop when the results had been retrieved.
-    setLoading(true);
```
DevVersion: @crisbeto Those flashes are appearing, when asynchronously querying. I actually wasn't the guy who introduced this fix, but it seems reasonable for me. 

We should keep that, and set the `loading` to false, once the results are fetched successfully. 
Everything works fine, when caching is disabled, so the issue is, that we're resetting the `loading`, when the results were already cached.

To fix the issue, we simply need to reset the loading [here](https://github.com/crisbeto/material/blob/fix/autocomplete-progress-again/src/components/autocomplete/js/autocompleteController.js#L772)

crisbeto: Thank you, I'll rework it.

PR #8349, opened 2016-05-05
perf(dialog): improve the opening time on IE

crisbeto:
* Speeds up the `$mdUtil.disableScrollAround` method by removing a couple of redundant reads of `body.clientWidth`, which cause reflows.
* Moves the `getComputedStyle` of the backdrop to the `$$rAF` call.
* A couple of minor improvements to some DOM lookups in the dialog.

Note that the dialog start-up time is down about 120ms (from 350ms before, to around 230ms with this change), but it's still not ideal. There's still some potential to reduce this further, but it might be considered a breaking change (for example replacing the backdrop resizing logic with CSS).

### DIFF
--- a/src/components/backdrop/backdrop.js
+++ b/src/components/backdrop/backdrop.js
@@ -28,21 +28,20 @@ angular
     };
 
     function postLink(scope, element, attrs) {
-
-      // If body scrolling has been disabled using mdUtil.disableBodyScroll(),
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
Closed with SHA 4803b49793c93e03d422770ab524fa4b1019c51b

PR #8351, opened 2016-05-05
fix(input): md-maxlength not properly updates on model changes.

### LINKED ISSUE #1870: md-maxlength doesn't update immediately if model content is updated in javascript
Steps to reproduce: 
1. create a textarea with md-maxlength="500" ng-model="modelitem.description"
2. create a button with a function that updates modelitem.description to be null

issue - the couter of characters used next to the text area will not update, it will stay as 32/500 for example. It will only update once the user starts typing again. 

This is in version 0.7.1

DevVersion:
* Currently the md-maxlength char counter was not updated properly and listening on the different events is unnecessary.

  ngModelController will also update the model, when the `input` event gets called.

  Additonally we can cover all updates in the $validator itself, because it will always update, when changes get applied to the model.

Fixes #1870. Referencing #8351

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -620,12 +620,6 @@ function mdMaxlengthDirective($animate, $mdUtil) {
       // over the maxlength still counts as invalid.
       attr.$set('ngTrim', 'false');
 
-      ngModelCtrl.$formatters.push(renderCharCount);
-      ngModelCtrl.$viewChangeListeners.push(renderCharCount);
-      element.on('input keydown keyup', function() {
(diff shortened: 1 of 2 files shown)


crisbeto:
LGTM @ThomasBurleson

corentin-gautier:
:+1:

ThomasBurleson:
@DevVersion  - please rebase and test. Thx.

DevVersion:
@ThomasBurleson Done.

DevVersion:
@topherfangio Ping for review :)

PR #8354, opened 2016-05-05
fix(tabs): fix object reference-related errors in IE

### LINKED ISSUE #8276: The right arrow icon for tabs on IE in a dialog is always inactive and the selected tab is not underlined
The issue is reproducible in this codepen: http://codepen.io/cfung76/pen/PNyxZE

Steps to reproduce:
1.  Decrease the size of the browser to about 400 px.
2.  Click on the "TAB DIALOG" button.
3.  The dialog box contains 4 tabs where tab # 4 is hidden.
4.  On IE, the right arrow that allows the user to go to tab # 4 is disabled.  As a result, the user will not be able to go to tab # 4.  Also, the underline is missing for the selected tab.

# On IE
image

# On Chrome
image

crisbeto:
The tabs component usually only queries the DOM once for all the elements, however when the template is compiled later on, those references get lost. This fix addresses it by updating the references on the next digest.

Also improves some DOM lookups and removes some that weren't being used anymore.

Referencing #3953.
Fixes #8276.

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -76,6 +76,10 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
     bindEvents();
     $mdTheming($element);
     $mdUtil.nextTick(function () {
+      // Note that the element references need to be updated, because certain "browsers"
+      // (IE/Edge) lose them and start throwing "Invalid calling object" errors, when we
+      // compile the element contents down in `compileElement`.
+      elements = getElements();
       updateHeightFromContent();
       adjustOffset();
       updateInkBarStyles();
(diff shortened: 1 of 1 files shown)


src/components/tabs/js/tabsController.js
```diff
@@ -428,19 +433,17 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
    */
   function getElements () {
     var elements = {};
+    var node = $element[0];
 
     // gather tab bar elements
-    elements.wrapper = $element[ 0 ].getElementsByTagName('md-tabs-wrapper')[ 0 ];
-    elements.data    = $element[ 0 ].getElementsByTagName('md-tab-data')[ 0 ];
```
Splaktar: `md-tab-data` is just used in the `compileTemplate()` function and never by anything that uses the `elements` returned by `getElements()`?

crisbeto: Nope, searching through the whole thing didn't bring up anything.

Splaktar:
Tested locally on IE11 and Edge for Windows 10 and this does seem to work. Though I don't see any exceptions without this change.

crisbeto:
It seems to be throwing mainly when it paginates. For example try [opening the "Tab dialog" in the dialog demo](https://material.angularjs.org/HEAD/demo/dialog) in HEAD and resizing your browser waaay down. If it doesn't throw try resizing the window up/down.

PR #8359, opened 2016-05-06
mdSelect:fix name attribute issue for mdSelect

abdvl:
* if `mdSelect` has `name` attribute, it will create a hidden select component, but this is a `.` before the attribute name


DevVersion:
@abdvl I agree, I'm wondering why the name is prefixed with a dot. (See commit https://github.com/angular/material/commit/aa440efcb381a61abb49ea846d1fe0b458cc3a50)

ThomasBurleson:

PR #8362, opened 2016-05-06
fix(panel): Fix type annotations.

ErinCoughlan:
Closure compiler was angry. This makes it happy again.

@KarenParker @gmoothart Please review.


KarenParker:
👍

PR #8368, opened 2016-05-06
fix(autocomplete): properly parse unsafe strings to prevent advanced XSS attacks

### LINKED ISSUE #8356: md-highlight-text escaping ampersands (&)
When using md-highlight-text where the result has an & and you type up to that & it will start to display as `&amp;` on the page.

Codepen: http://codepen.io/edgar131/pen/bpOjPj
Type: 'test&' and see what happens.

DevVersion:
* The highlight controller no longer sanitizes the term or content, because this would require complex libraries like `ngSanitize`
* Also there is no need to sanitize those strings, because after this change we never insert them as HTML nodes anymore.

* We decompose the content string into different tokens and then compose them together into different elements.
   This allows us to be 100% sure that we only insert trusted HTML code and no unsafe HTML code, which could include XSS attacks.

* This approach also supports now HTML identifiers and special text characters in the highlight text and content.

Fixes #8356

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -1214,9 +1214,11 @@ describe('<md-autocomplete>', function() {
   });
 
   describe('xss prevention', function() {
+
     it('should not allow html to slip through', inject(function($timeout, $material) {
       var html = 'foo <img src="img" onerror="alert(1)" />';
-      var scope = createScope([{display: html}]);
+      var scope = createScope([{ display: html }]);
+
       var template = '\
           <md-autocomplete\
               md-selected-item="selectedItem"\
@@ -1227,6 +1229,7 @@ describe('<md-autocomplete>', function() {
               placeholder="placeholder">\
             <span md-highlight-text="searchText">{{item.display}}</span>\
           </md-autocomplete>';
+
       var element = compile(template, scope);
       var ctrl = element.controller('mdAutocomplete');
       var ul = element.find('ul');
@@ -1250,6 +1253,7 @@ describe('<md-autocomplete>', function() {
 
       element.remove();
     }));
+
   });
 
   describe('Async matching', function() {
@@ -1774,6 +1778,7 @@ describe('<md-autocomplete>', function() {
   });
 
   describe('md-highlight-text', function() {
+
     it('updates when content is modified', inject(function() {
       var template = '<div md-highlight-text="query">{{message}}</div>';
       var scope = createScope(null, {message: 'some text', query: 'some'});
@@ -1794,7 +1799,7 @@ describe('<md-autocomplete>', function() {
       element.remove();
     }));
 
-    it('should properly apply highlight flags', inject(function() {
+    it('should properly apply highlight flags', function() {
       var template = '<div md-highlight-text="query" md-highlight-flags="{{flags}}">{{message}}</div>';
       var scope = createScope(null, {message: 'Some text', query: 'some', flags: '^i'});
       var element = compile(template, scope);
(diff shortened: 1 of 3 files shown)


robertmesserle:
@ThomasBurleson LGTM, closes #8388 as well.

jammed343:
This doesn't work properly with html escaping. Also Is there a reason why you are using html escaping instead of regex escaping.

I changed:
`function parseUnsafeString(text) {
    return angular.element('<div>').text(text).html();
 }`
To:
`function parseUnsafeString(text) {
    // http://stackoverflow.com/questions/3446170/escape-string-for-use-in-javascript-regex
    return text.replace(/[\-\[\]\/\{\}\(\)\*\+\?\.\\\^\$\|]/g, "\\$&");
 }`

Which makes it work well for me.

Otherwise i can search for "amp" and all the ampersands come up in search results with the current solution.

I suggest we add regex escape and then merge this into angular-material

DevVersion:
@jammed343 I cannot see that issue. The way we currently parse the unsafe string, seems to work as expected. It just depends on the native browser behavior, which makes sure that the displayed results and the search text match.

Can you explain what issues you are experiencing? Did you run that PR already?

jammed343:
@DevVersion
Using your current html escaping gives me this:
screen shot 2016-08-28 at 11 24 11 pm
screen shot 2016-08-28 at 11 24 42 pm

DevVersion:
@jammed343 Everything works for me as expected. And the tests are also passing.

How did you pull this PR to run on your codebase?

| Before | After |
| ------- | ------|
| image | image |

Also the way this Regex parses the string is not valid. We want to retrieve the HTML encoded way, instead of the UTF-8 text.

> Here is a demo: http://codepen.io/DevVersion/pen/PzrQGy

jammed343:
@DevVersion I used the DevVersion:fix/parse-unsafe-regex  branch which didn't work for me at first. then i modified the `parseUnsafeString` function you added to use regex escape and that worked for me.

DevVersion:
@jammed343 Hmm weird. The tests pass, the demos are working as expected. It may be an issue on the PR pulling. Not sure how you served your application with the new changes etc.

jammed343:
@DevVersion 

I added some console output to your original pull:
```
if (text === null || state.unsafeText !== prevState.unsafeText) {
5        text = parseUnsafeString(state.unsafeText);
5      }
-      if (regex === null || state.term !== prevState.term) {
5        regex = getRegExp(parseUnsafeString(state.term), flags);
5      }
Added these console.logs
4      console.log("STATE CHANGE");
4      console.log("state.unsafeText: ", state.unsafeText);
4      console.log("state.term: ", state.term);
4      console.log("text: ", parseUnsafeString(state.unsafeText));
4      console.log("regex: ", getRegExp(parseUnsafeString(state.term), flags));
4      console.log("$element.html", text.replace(regex, '<span class="highlight">$&</span>'));
4
4      $element.html(text.replace(regex, '<span class="highlight">$&</span>'));
```

```
STATE CHANGE
angular-material.js:23902 state.unsafeText:  Animals & Pet Supplies > Pet Supplies > Bird Supplies > Bird Cages & Stands
angular-material.js:23903 state.term:  a
angular-material.js:23904 text:  Animals &amp; Pet Supplies &gt; Pet Supplies &gt; Bird Supplies &gt; Bird Cages &amp; Stands
angular-material.js:23905 regex:  /a/gi
angular-material.js:23906 $element.html <span class="highlight">A</span>nim<span class="highlight">a</span>ls &<span class="highlight">a</span>mp; Pet Supplies &gt; Pet Supplies &gt; Bird Supplies &gt; Bird C<span class="highlight">a</span>ges &<span class="highlight">a</span>mp; St<span class="highlight">a</span>nds
```

jammed343:
@DevVersion 
Your latest commit:
https://github.com/angular/material/pull/8368/commits/776d82a158659a11da6087291ecbf51e509ab10b
is now working.

DevVersion:
@ThomasBurleson @jammed343 I made some changes regarding the way we display the highlighted text. 

The current approach was very unstable and really dangerous for XSS attacks, since we are setting the `HTML` of the element.

ThomasBurleson:
@jammed343 - Thx for your persistence on improving this PR. 👍 
@DevVersion  -  lgtm.

ThomasBurleson:
@DevVersion - plz squash and (if needed) rebase.

PR #8371, opened 2016-05-07
fix($mdUtil) nextTick now keeps reference to correct scope.

### LINKED ISSUE #8358: md-autocomplete freezes UI when removed while loading items.
There are a number of other issues which seem to be related, but they're all closed:

#7634
#3287
#3334
#3371  <--- closest that I can figure...

**Actual behavior**:
-  What is the issue?*
  * It seems that if the autocomplete gets removed from the UI while it is still loading items, the associated `.md-scroll-mask` persists and eats mouse/keyboard events making the UI seem unresponsive.
-  What is the expected behavior?*
  * The `.md-scroll-mask` should go away when the autocomplete does...

-  [CodePen Demo](http://codepen.io/anon/pen/xVmMzy)
-  Details:
  1. Open the pen.
  2. Start typing (slowly) in the autocomplete.  I usually use "alaska".  I usually get to about "al" before the autocomplete disappears.
  3. click the switch to bring the autocomplete back.
  4. refocus the autocomplete and keep adding to your state (slowly).
  5. Continue with steps 3 and 4 until the switch can no longer be clicked due to the presence of the `.md-scroll-mask`

**Angular Versions**: *
-  Angular Version:*
  * __1.4.8__
-  Angular Material Version:*
  * __1.0.8__

**Additional Information**:
-  Browser Type:*
  * __Chrome__
-  Browser Version:*
  * __49.0.2623.112__
-  OS:*
  * __OS-X 10.11.2__
-  Stack Traces:*
  * __None in the console.__

----

  Add more lines if needed.

mgilson:
Previously, `nextTick` would decide whether to run the functions based
on the first `scope` argument passed during a given digest cycle.  This
could fail in multiple scenarios ...

1) The first `nextTick` call didn't supply a scope, but subsequent calls did.
2) Multiple components (with different scopes) registered callbacks during the same digest cycle.

This could lead to callbacks being executed when they shouldn't be (if
the first scope registered wasn't $$destroyed, this _this_ scope was),
or callbacks not being executed when they should be (the first scope was
$$destroyed, but this scope wasn't).

I believe that this fixes #8358

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -587,7 +587,7 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
       var queue = nextTick.queue || [];
 
       //-- add callback to the queue
-      queue.push(callback);
+      queue.push({scope: scope, callback: callback});
 
       //-- set default value for digest
       if (digest == null) digest = true;
@@ -606,16 +606,18 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
        * Trigger digest if necessary
        */
       function processQueue() {
-        var skip = scope && scope.$$destroyed;
-        var queue = !skip ? nextTick.queue : [];
-        var digest = !skip ? nextTick.digest : null;
+        var queue = nextTick.queue;
+        var digest = nextTick.digest;
 
         nextTick.queue = [];
         nextTick.timeout = null;
         nextTick.digest = false;
 
-        queue.forEach(function(callback) {
-          callback();
+        queue.forEach(function(queueItem) {
+          var skip = queueItem.scope && queueItem.scope.$$destroyed;
+          if (!skip) {
+            queueItem.callback();
+          }
         });
 
         if (digest) $rootScope.$digest();
--- a/src/core/util/util.spec.js
+++ b/src/core/util/util.spec.js
@@ -247,15 +247,76 @@ describe('util', function() {
         //-- but callback is still called more
         expect(callback.calls.count()).toBe(10);
       }));
+
       it('should return a timeout', inject(function($mdUtil) {
         var timeout = $mdUtil.nextTick(angular.noop);
         expect(timeout.$$timeoutId).toBeOfType('number');
       }));
+
       it('should return the same timeout for multiple calls', inject(function($mdUtil) {
         var a = $mdUtil.nextTick(angular.noop),
           b = $mdUtil.nextTick(angular.noop);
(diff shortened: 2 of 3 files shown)


mgilson:
I don't know how you guys feel about tiny cosmetic fixups being incorporated with a real change... If you'd rather, I can drop 0ea79f2 from this PR and submit it in a later PR (or not).

src/core/util/util.spec.js
```diff
@@ -443,6 +443,22 @@ describe('util', function() {
       flush(function(){ expect( callBackUsed ).toBeUndefined(); });
     }));
 
+    it('should only skip callbacks of scopes which were destroyed', inject(function($mdUtil) {
+      var callBackUsed1, callback1 = function(){ callBackUsed1 = true};
```
mgilson: FWIW, I like the way the callbacks are tracked via a jasmine spy (see line 232), but I used the same idiom used in the previous test since these tests are related.  I'm happy to update both tests to use the spy idiom if you'd like.

src/core/util/util.spec.js
```diff
@@ -443,6 +443,22 @@ describe('util', function() {
...
+      var callBackUsed2, callback2 = function(){ callBackUsed2 = true};
+      var scope1 = $rootScope.$new(true);
+      var scope2 = $rootScope.$new(true);
+
+      $mdUtil.nextTick(callback1, false, scope1);
```
mgilson: The old bug also had a problem with:

```javascript
$mdUtil.nextTick(callback1, false);  // no scope
$mdUtil.nextTick(callback2, false, scope);
$scope.$destroy();

flush(function() {
  expect(callback1).toHaveBeenCalled();
  expect(callback2).not.toHaveBeenCalled();
});
```

Would have failed previously.  Should I add that as a test too?

DevVersion: Yea, if it failed previously and it's a case which is not covered yet, then it would make definitely sense.

mgilson: Done.

src/core/util/util.js
```diff
@@ -587,7 +587,7 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
       var queue = nextTick.queue || [];
 
       //-- add callback to the queue
-      queue.push(callback);
+      queue.push({scope: scope, callback: callback});
```
ThomasBurleson: just use `queue.push( angular.bind(scope, callback) );`

PR #8375, opened 2016-05-07
fix(datepicker): align input border colors with inputs.

### LINKED ISSUE #8148: md-datepicker border-bottom not the same as md-input border-bottom
md-input border-bottom seems to use RGB(#, #, #, 0.12) on unactivated inputs while md-datepicker's div used a different hue color RBG(#, #, #) with no opacity adjustment. If you inspect the demo here you can see the difference: https://material.angularjs.org/1.0.7/demo/input

This is causing problems for forms and different colors. The borders don't match. md-datepicker should use the same hue and opacity as md-input.

DevVersion:
* Currently the datepicker had some different theme colors for the border.
  Those colors should be the same as in the input component.

Fixes #8148.

### DIFF
--- a/src/components/datepicker/datePicker-theme.scss
+++ b/src/components/datepicker/datePicker-theme.scss
@@ -8,10 +8,10 @@
   }
 
   .md-datepicker-input-container {
-    border-bottom-color: '{{background-300}}';
+    border-bottom-color: '{{foreground-4}}';
 
     &.md-datepicker-focused {
-      border-bottom-color: '{{primary-500}}'
+      border-bottom-color: '{{primary-color}}'
     }
 
     &.md-datepicker-invalid {


src/components/datepicker/datePicker-theme.scss
```diff
@@ -8,10 +8,10 @@
   }
 
   .md-datepicker-input-container {
-    border-bottom-color: '{{background-300}}';
+    border-bottom-color: '{{foreground-4}}';
```
BevanR: 👍  This matches https://github.com/angular/material/blob/master/src/components/input/input-theme.scss#L5

src/components/datepicker/datePicker-theme.scss
```diff
@@ -8,10 +8,10 @@
...
 
     &.md-datepicker-focused {
-      border-bottom-color: '{{primary-500}}'
+      border-bottom-color: '{{primary-color}}'
```
BevanR: Was this reported in the original bug report?

BevanR:
LGTM!

src/components/datepicker/datePicker-theme.scss
DevVersion: No. It's just a revisited line, which is now aligned with the input component

BevanR: 👍

PR #8387, opened 2016-05-09
fix(input): add support for dynamically loaded ngMessage directives.

### LINKED ISSUE #6810: Form validation issues with ng-messages-include
I'm trying to start using `ng-messages-include` as described in [ngMessages](https://docs.angularjs.org/api/ngMessages) under "Reusing and Overriding Messages", and I'm running into issues.  I'm primarily seeing it not working with `md-auto-hide`, showing the error messages when the form is displayed.  I like hiding validation errors until the user has had an opportunity to fill out the form and would like to use `md-auto-hide` with `ng-messages-include`. 

See [this CodePen](http://codepen.io/jvmccarthy/pen/EPoywL) for this issue.  The pen is showing other issues including weirdness with adjacent input containers.

Thanks again for all the hard work on angular-material!

### LINKED ISSUE #7477: Autocomplete ng-messages wrong font-size

### LINKED ISSUE #7596: Different ng-message size with ng-show and ng-if
As you can see in this [codepen](http://codepen.io/anon/pen/xVgwdm), the ng-message font size varies depending on whether the input is defined with a `ng-show` or `ng-if` attribute.

It seems that the ng-message div is only generated with the `md-input-message-animation` class when `ng-show` is used.

### LINKED ISSUE #7823: ngMessagesInclude not animated
`ng-message` which are within `ng-messages-include` don't get the class 'md-input-messages-animation'.

ie this does not work well :
```html
    <input md-maxlength="30" required name="description" g-model="project.description">
    <div ng-messages="projectForm.description.$error">
        <div ng-messages-include="errors-message.html"></div>
    </div>
</md-input-container>
```

where errors-message.html contains:
```html
<div ng-message="required">This is required.</div>
<div ng-message="md-maxlength">The name has to be less than 30 characters long.</div>
```

DevVersion:
* Currently ngMessage directives inside of an input-container, which are loaded / transcluded dynamically, didn't get initialized properly.
  The animation was not working. As well as the auto-hide didn't work properly.

cc. @topherfangio 

Fixes #7477. Fixes #7596. Fixes #6810. Fixes #7823

### DIFF
--- a/src/components/input/input.spec.js
+++ b/src/components/input/input.spec.js
@@ -436,6 +436,59 @@ describe('md-input-container directive', function() {
     expect(el[0].querySelector("[ng-messages]").classList.contains('md-auto-hide')).toBe(false);
   }));
 
+  it('should set the animation class on the ngMessage properly', inject(function() {
+    var element = compile(
+      '<md-input-container>' +
+        '<input ng-model="inputVal">' +
+        '<div ng-messages>' +
+          '<ng-message id="requiredMessage" when="required">Field required</ng-message>' +
+        '</div>' +
+      '</md-input-container>'
+    );
+
+    var ngMessage = element.find('ng-message');
+    expect(ngMessage).toHaveClass('md-input-message-animation');
+  }));
+
+  it('should set the animation class on a transcluded ngMessage', function() {
+    // We can emulate the transclusion, by wrapping the ngMessage inside of a document fragment.
+    // It is not necessary to add a *extra* component / directive for that, since we just
+    // want to the test the DocumentFragment detection.
+    var fragment = document.createDocumentFragment();
+
+    var inputContainer = compile(
+      '<md-input-container>' +
+        '<input ng-model="inputVal">' +
+        '<div ng-messages id="messageInsertion">' +
+        '</div>' +
+      '</md-input-container>'
+    );
+
+    // We build our element, without compiling and linking it.
+    // Because we invoke those steps manually during the tests.
+    var messageElement = angular.element(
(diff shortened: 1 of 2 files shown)


DevVersion:
@topherfangio Added the tests as discussed :)

src/components/input/input.spec.js
```diff
@@ -436,6 +436,59 @@ describe('md-input-container directive', function() {
...
+          '<ng-message id="requiredMessage" when="required">Field required</ng-message>' +
+        '</div>' +
+      '</md-input-container>'
+    );
+
+    var ngMessage = element.find('ng-message');
+    expect(ngMessage).toHaveClass('md-input-message-animation');
+  }));
+
+  it('should set the animation class on a transcluded ngMessage', function() {
```
topherfangio: I'm curious if we could do a different/shorter test here.

Based on the issues this is supposed to fix, if we put it inside an autocomplete, or use the `ng-messages-include` directive, it would fail correct (I mean, without this PR)?

Is there a reason to do it as a document fragment test rather than one of the standard transclusion methods listed in the issues?

DevVersion: Actually, as I commented on the test, using the document fragment is testing the new added code in the input component.

Using a tranclusion here and testing it, would be more complicated. 

I will give it a try with the `ng-messages-include`.

topherfangio: Okay, if that's the case, then go ahead and add the `ng-messages-include` as a separate test if you can 😄 

Thanks for the clarification!

PR #8394, opened 2016-05-10
fix(tooltip): cancel show timeouts when leaving before delay.

### LINKED ISSUE #8363: md-tooltip: md-delay with md-autohide not working as expected
If you have md-autohide enabled on a tooltip that has a md-delay set, and you mouse over and off the object in less time then the delay period, then the tooltip will still show up (which it shouldn't) and will remain on (which it also shouldn't do). 

**Actual behavior**:
-  What is the issue? Tooltip displays and remains after the mouse leaves the object.
-  What is the expected behavior? Tooltip should not display because the mouse left the object before the delay period.

**Steps to reproduce the issue**:
-  Add a md-tooltip to a md-fab. Set md-delay and md-autohide (set delay to about 1000 to make this easier to demonstrate)
- Move your mouse over the fab and then off the fab in less time then the delay (ie <1000ms)
- The tooltip will show after the delay, and will remain until you mouse over again. This is new in 1.1. This worked on the previous version that I used (material 1.0.5).

**Angular Versions**: 
-  Angular Version: 1.5.5
-  Angular Material Version: 1.1.0-rc4

**Additional Information**:
-  Browser Type: Chromuim
-  Browser Version: 49

DevVersion:
* When a user hovers over a tooltip, which has a delay, and we're leaving the tooltips parent before the delay is finished, then we're detaching all leave handlers, which is not valid.<br/>
  We should cancel the show timeout, when the leave occurs before the delay.

Fixes #8363.

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -59,6 +59,7 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
     var parent        = $mdUtil.getParentWithPointerEvents(element),
         content       = angular.element(element[0].getElementsByClassName('_md-content')[0]),
         tooltipParent = angular.element(document.body),
+        showTimeout   = null,
         debouncedOnResize = $$rAF.throttle(function () { updatePosition(); });
 
     if ($animate.pin) $animate.pin(element, parent);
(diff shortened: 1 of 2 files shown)


BevanR:
A [codepen](http://codepen.io/team/AngularMaterial/pen/bEGJdd) would be really useful here.

DevVersion:
@BevanR Why is that? It's not necessary, because this the actual fix for the issue #8363.

BevanR:
@DevVersion I wanted to review the PR, but I was not sure I would be able to understand the bug correctly from the textual description.  A codepen example would have made it clear and unambiguous what the bug was and how to reproduce it.  No matter no more...

PR #8396, opened 2016-05-10
feat(panel): add mdPanel service

ErinCoughlan:


docs/app/js/app.js
```diff
@@ -289,6 +289,10 @@ function(SERVICES, COMPONENTS, DEMOS, PAGES, $location, $rootScope, $http, $wind
       pages: apiDocs.service.sort(sortByName),
       type: 'toggle'
     },{
+      name: 'Types',
+      pages: apiDocs.type.sort(sortByName),
+      type: 'toggle'
+    },{
```
DevVersion: Not sure, about reviewing here, but this should have a space. Just to be consistent

Issue #8393, opened 2016-05-10
autocomplete: add ability to set ID on md-virtual-repeat-container for custom CSS


annnguyen:
Found a closed ticket that wasn't actually resolved: [#4772](https://github.com/angular/material/issues/4772)

We need the ability to style our global search input field using `md-autocomplete` that is persistent throughout the site to be styled uniquely from any other autocomplete field we're using in various forms on various pages. The width of the element `<md-virtual-repeat-container>` is what needs to be overriden but there's no option available to do so.

**Actual behavior**:
-  What is the issue? 
  - There's no ability to set a unique ID to `<md-virtual-repeat-container>` of an `<md-autocomplete>` to do styling changes to the width of the dropdown list. 
-  What is the expected behavior?
  - Would like for there to be directive attribute `md-menu-container-id` to set ID of `<md-virtual-repeat-container>` element.

**CodePen** or Steps to reproduce the issue: *
-  CodePen Demo: [http://codepen.io/annnguyen/pen/NNomLo](http://codepen.io/annnguyen/pen/NNomLo)
-  Details: Set `md-virtual-repeat-container` width to 80%, which is applied to all. There is no ability to set unique selectors to multiple `md-autocomplete` elements.

**Angular Versions**: *
-  Angular Version: 1.5.0
-  Angular Material Version: 1.0.8

**Additional Information**:
-  Browser Type: Any
-  Browser Version: Any
-  OS: Any

BevanR:
Do you want to specify the width of an `<md-virtual-repeat-container>` element that corresponds to some `<md-autocomplete>` elements, but not when it corresponds to other `<md-autocomplete>` elements?

My first thought was that you could specify the class in your `<md-autocomplete>` element and use that class in your selector.  E.g. `<md-autocomplete class="wide-options">` and `md-autocomplete.wide-options md-virtual-repeat-container { width: 80%; }`.

However I see now that is not possible because the `<md-virtual-repeat-container>` is not a descendant of its corresponding `<md-autocomplete>` but a direct child of the root element of the app, that gets absolutely positioned when it is visible.

There is a lot of unrelated code in your codepen.  Could you reduce it to two `<md-autocomplete>` elements, one CSS property, and zero JavaScript (or perhaps just enough to populate the autocomplete options).

You might also consider making them different so that it is clear which one you want to change and which one you don't.  E.g. "The blue one should be default width, the red one should be wider".

[...]

PR #8397, opened 2016-05-10
feat(autocomplete): add ability to style md-virtual-repeat-container

annnguyen:
…er generated by `md-autocomplete`. 

This allows control over css styling for specific instances of `<md-virtual-repeat-container>` created by `md-autocomplete`.

Fixes [#8393](https://github.com/angular/material/issues/8393)

### DIFF
--- a/src/components/autocomplete/js/autocompleteDirective.js
+++ b/src/components/autocomplete/js/autocompleteDirective.js
@@ -55,6 +55,8 @@ angular
  * @param {boolean=} md-no-asterisk When present, asterisk will not be appended to the floating label
  * @param {boolean=} md-autoselect If true, the first item will be selected by default
  * @param {string=} md-menu-class This will be applied to the dropdown menu for styling
+ * @param {string=} md-menu-container-id This will be applied to the parent container of the dropdown
+ *     menu for styling
  * @param {string=} md-floating-label This will add a floating label to autocomplete and wrap it in
  *     `md-input-container`
  * @param {string=} md-input-name The name attribute given to the input element to be used with
@@ -149,6 +151,7 @@ function MdAutocomplete () {
       floatingLabel:    '@?mdFloatingLabel',
       autoselect:       '=?mdAutoselect',
       menuClass:        '@?mdMenuClass',
+      menuContainerId:  '@?mdMenuContainerId',
       inputId:          '@?mdInputId'
     },
     link: function(scope, element, attrs, controller) {
@@ -181,6 +184,7 @@ function MdAutocomplete () {
               ng-if="$mdAutocompleteCtrl.loadingIsVisible()"\
               md-mode="indeterminate"></md-progress-linear>\
           <md-virtual-repeat-container\
+              id="{{menuContainerId}}"\
               md-auto-shrink\
               md-auto-shrink-min="1"\
               ng-mouseenter="$mdAutocompleteCtrl.listEnter()"\
--- a/src/components/autocomplete/js/autocompleteDirective.js
+++ b/src/components/autocomplete/js/autocompleteDirective.js
@@ -55,8 +55,8 @@ angular
  * @param {boolean=} md-no-asterisk When present, asterisk will not be appended to the floating label
  * @param {boolean=} md-autoselect If true, the first item will be selected by default
  * @param {string=} md-menu-class This will be applied to the dropdown menu for styling
- * @param {string=} md-menu-container-id This will be applied to the parent container of the dropdown
- *     menu for styling
+ * @param {string=} md-menu-container-class This will add a class to `md-virtual-repeat-container`,
+ *     the parent container of the dropdown list
  * @param {string=} md-floating-label This will add a floating label to autocomplete and wrap it in
  *     `md-input-container`
  * @param {string=} md-input-name The name attribute given to the input element to be used with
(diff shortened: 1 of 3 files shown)


annnguyen:
I signed it!

BevanR:
Thanks for your contribution.  I responded on #8393 with some ideas to make this easier to review.

BevanR:
A quick look over the code looks good.  It has a test, which is great!

I think it would be more useful if it allowed class(es) to be set instead of IDs.  IDs must be unique in a document.  However classes can be re-used.  If only IDs can be set, then a component producing many `<md-autocomplete>` elements (such as a directive that expands some element into multiple `<md-autocomplete>` elements) will need to ensure each one has unique ID, and the consumer's CSS will not be able to efficiently modify any elements coming from the factory-component.

BevanR:
This appears to duplicate #8403.  I like the approach in #8403 better because it allows classes to be set, instead of IDs.  However this PR has tests.  Perhaps @vassagus and you can merge efforts?

annnguyen:
Thank @BevanR. I'll make updates to the ticket and pull request code. @vassagus if you're okay, i can update my code to set class to md-virtual-repeat-container instead of id if that's the preferred method. Lmk!

vassagus:
@annnguyen sounds good, you can go ahead and merge my commit. Thanks guys.

annnguyen:
@DevVersion code is ready for review whenever you guys have bandwidth!

ThomasBurleson:

ausminternet:
:rage: I can't count anymore the bugs I need to be fixed and which are closed in favour of a not-yet-released version 2... Even if there are PRs...

jensbodal:
Using information from this pull request I made a decorator to add to my app's config;  the decorator allows an `md-menu-container-class` attribute to be added to the `md-autocomplete` directive which can then be used for styling.  [Plunker](https://plnkr.co/edit/0tzee8RpoD2WeFSJX1z1?p=preview)

ThomasBurleson:
@jensbodal - This is a nice technique to extend/override an existing ngM1 component DDO!  :-). Perhaps you should blog this for the community ;-).

>

vassagus:
@jensbodal thank you, that works for me as well.

dohomi:
@jensbodal I think this is exactly what the current `md-autocomplete` should do. Would be great if this issue would be opened to add the class to the outer container.

jensbodal:
@ThomasBurleson thanks for the comment and the suggestion.  It took me a bit to decide on a proper blog framework but decided to use codepen as it seems to fit.  Blog post made available here: http://codepen.io/jensbodal/post/extending-an-angular-ddo-directive-definition-object

tiagomsmagalhaes:
@jensbodal thanks for the decorator. I had to tweak it because i had `md-autocomplete` in another directive so i had to pass it as a variable, so I tweaked it to allow that. Is there a way that could allow both options so we could merge it? my codepen: http://codepen.io/tiagomagalhaes/pen/dOXPXj

jensbodal:
@tiagomsmagalhaes you can just check to see if the element resolves to include an expression, and if not read from the attribute as a string directly.

http://codepen.io/jensbodal/pen/oYLWxr

*(relevant part)*
```js
return function(e, a) {
  var menuContainer = a.find('md-virtual-repeat-container');
  var menuContainerClass = '';
  
  if (typeof e.menuContainerClass !== 'boolean' && e.menuContainerClass !== 0) {
    menuContainerClass = e.menuContainerClass;      
  }

  else {
    menuContainerClass = attr.mdMenuContainerClass || '';
  }
  
  menuContainer.addClass(menuContainerClass);

  template.apply(this, arguments);
};
```

I'm looking for `!== 'boolean' && !== 0` as that seemed to fit all the cases where the attribute didn't exist (`false`), the attribute referenced a controller variable which didn't exist (`true`), the attribute was a string and contained dashes (`0`), and the attribute was a string (`true`).  The `!== 0` is related to how angular interpolates attributes I believe.

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -181,13 +184,15 @@ function MdAutocomplete () {
...
+              id="{{menuContainerId}}"\
               md-auto-shrink\
               md-auto-shrink-min="1"\
               ng-mouseenter="$mdAutocompleteCtrl.listEnter()"\
               ng-mouseleave="$mdAutocompleteCtrl.listLeave()"\
               ng-mouseup="$mdAutocompleteCtrl.mouseUp()"\
               ng-hide="$mdAutocompleteCtrl.hidden"\
-              class="md-autocomplete-suggestions-container md-whiteframe-z1"\
+              class="md-autocomplete-suggestions-container md-whiteframe-z1' +
+                (attr.mdMenuContainerClass != null ? ' ' + attr.mdMenuContainerClass : '') + '"\
```
Splaktar: Can this use `ng-class` like `menuClass` uses below?

Splaktar: Looks like that isn't needed, but the logic here can be pulled out of the template to make things cleaner.

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -181,13 +184,15 @@ function MdAutocomplete () {
               ng-if="$mdAutocompleteCtrl.loadingIsVisible()"\
               md-mode="indeterminate"></md-progress-linear>\
           <md-virtual-repeat-container\
+              id="{{menuContainerId}}"\
```
Splaktar: This looks to be left over from the previous implementation and should be removed.

Splaktar:
Moved this over to a new PR https://github.com/angular/material/pull/11194

PR #8403, opened 2016-05-10
feat(autocomplete): Adding md-container-class

vassagus:
This attribute enables to set a class to the autocomplete dropdown wrapper, so the caller can have full control over css. Currently you can set a class for the menu ul inside the md-virtual-repeat-container, but this doesn't allow to style the entire container.


vassagus:
I signed it!

vassagus:
I signed it!

BevanR:
This appears to duplicate #8397.  I like this approach better because it allows classes to be set, instead of IDs.  However #8397 has tests.  Perhaps @annnguyen and you can merge efforts?

vassagus:
Thanks @BevanR I agree, this efforts will be merged to  [https://github.com/angular/material/pull/8397](8397)

PR #8419, opened 2016-05-11
fix(navbar): fixes navbar validation logic in angular 1.3

gmoothart:
@ThomasBurleson @ErinCoughlan @EladBezalel

### DIFF
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -461,9 +461,9 @@ function MdNavItemController($element) {
   /** @private {boolean} */
   this._focused = false;
 
-  var hasNavClick = this.mdNavClick != null;
-  var hasNavHref = this.mdNavHref != null;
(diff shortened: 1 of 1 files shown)


src/components/navBar/navBar.js
```diff
@@ -461,9 +461,9 @@ function MdNavItemController($element) {
   /** @private {boolean} */
   this._focused = false;
 
-  var hasNavClick = this.mdNavClick != null;
-  var hasNavHref = this.mdNavHref != null;
-  var hasNavSref = this.mdNavSref != null;
+  var hasNavClick = !!($element.attr('md-nav-click'));
```
EladBezalel: There's `$mdUtil.parseAttributeBoolean`

gmoothart: the actual values aren't boolean; I just need to know if they have been provided at all - so just checking for truthiness is ok.

PR #8423, opened 2016-05-12
fix(all): Use $templateRequest instead of $http for security.

### LINKED ISSUE #8413: Change all templates to be requests through $templateRequest
Currently the library goes through `$http` for template requests, which is a security problem. The templates should instead be requested through `$templateRequest`

topherfangio:
For security purposes, we have switched from using `$http` for our template and icon requests to using `$templateRequest` since it provides some automatic caching, and more importantly, security checks about the URL/data.

See https://docs.angularjs.org/api/ng/service/$templateRequest for more information.

Fixes #8413.


topherfangio:
@jelbourn Ping for review when you have a chance.

jelbourn:
LGTM

PR #8438, opened 2016-05-12
feat(compiler): add support for sequential resolve.

### LINKED ISSUE #8417: mdDialog resolve as uiRoute resolve
**Actual behavior**:
-  What is the issue? *
This is working exactly as uibModal. It would be great that mdDialog resolve proeprties could be used in furthrer properties (as in uiRoute resolve).

```javascript
resolve: {
  aProperty : function(AService){
    return AService.getA();
  },
  bProperty : function(BService, aProperty){
    return BService.getB('/b-url/' + aProperty.id)
  }
}
```
-  Details:
This wont work, aPropertyProvider error, of course.

**Angular Versions**: *
-  Angular Version: 1.4
-  Angular Material Version: 1.0.8

**Additional Information**:
-  Browser Type: * Any
-  Browser Version: * Any
-  OS: * Any

DevVersion:
* Currently all promises are executed in order, but they don't take advantage of each promises resolve.
  All promises are getting executed in order, and at the end we wait for all promises to resolve.
image

This commit changes the invoking behavior.
* All promises getting executed in order, but we only execute the next promise, when the current promise was resolved.
  This allows us to inject the resolved value into the next resolve property / key.

image

FYI: We didn't discuss about that enhancement with the team yet. I just put that PR together because I'd like to do something fun :)

Closes #8417.

### DIFF
--- a/src/core/services/compiler/compiler.js
+++ b/src/core/services/compiler/compiler.js
@@ -77,28 +77,58 @@ function mdCompilerService($q, $http, $injector, $compile, $controller, $templat
     // Take resolve values and invoke them.
     // Resolves can either be a string (value: 'MyRegisteredAngularConst'),
     // or an invokable 'factory' of sorts: (value: function ValueGetter($dependency) {})
-    angular.forEach(resolve, function(value, key) {
+    function invokeResolve(key, value) {
       if (angular.isString(value)) {
         resolve[key] = $injector.get(value);
+        return $q.when();
       } else {
-        resolve[key] = $injector.invoke(value);
+        return $q.when(resolve[key] = $injector.invoke(value, null, resolve));
       }
-    });
-    //Add the locals, which are just straight values to inject
-    //eg locals: { three: 3 }, will inject three into the controller
-    angular.extend(resolve, locals);
+    }
 
-    if (templateUrl) {
-      resolve.$template = $http.get(templateUrl, {cache: $templateCache})
-        .then(function(response) {
-          return response.data;
+    function resolveLocals() {
+      var defer = $q.defer();
+
+      function runResolve(index) {
+        var key = Object.keys(resolve)[index];
+
+        if (!key) {
+          return resolveComplete();
+        }
+
+        invokeResolve(key, resolve[key]).then(function(val) {
+          // When the promise returned a value, then we can immediately update the property to
+          // be able to inject it into later resolve invokes.
+          if (val) resolve[key] = val;
+          runResolve(index + 1);
         });
-    } else {
-      resolve.$template = $q.when(template);
+      }
+
+      // Start the resolve at index 0
+      runResolve(0);
+
+      function resolveComplete() {
+        // Add the locals, which are just straight values to inject
+        // eg locals: { three: 3 }, will inject three into the controller
+        angular.extend(resolve, locals);
+
+        if (templateUrl) {
+          resolve.$template = $http.get(templateUrl, {cache: $templateCache})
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@DevVersion - interesting and seemingly worthwhile.

src/core/services/compiler/compiler.js
```diff
@@ -77,28 +77,58 @@ function mdCompilerService($q, $http, $injector, $compile, $controller, $templat
     // Take resolve values and invoke them.
     // Resolves can either be a string (value: 'MyRegisteredAngularConst'),
     // or an invokable 'factory' of sorts: (value: function ValueGetter($dependency) {})
-    angular.forEach(resolve, function(value, key) {
+    function invokeResolve(key, value) {
       if (angular.isString(value)) {
         resolve[key] = $injector.get(value);
+        return $q.when();
```
ThomasBurleson:
```js
return $q.when( resolve[key] = $injector.get(value) );
```

src/core/services/compiler/compiler.js
```diff
@@ -77,28 +77,58 @@ function mdCompilerService($q, $http, $injector, $compile, $controller, $templat
...
-    });
-    //Add the locals, which are just straight values to inject
-    //eg locals: { three: 3 }, will inject three into the controller
-    angular.extend(resolve, locals);
-
-    if (templateUrl) {
-      resolve.$template = $http.get(templateUrl, {cache: $templateCache})
-        .then(function(response) {
-          return response.data;
+    }
```
ThomasBurleson: Let's use this:

```js
var response = angular.isString(value) ? $injector.get(value) : $injector.invoke(value, null, resolve);

return $q.when( resolve[key] = response );
```

ThomasBurleson:
```js
/**
       * Resolves can either be a string service name (value: 'MyRegisteredAngularConst'),
       * or an invokable 'factory' of sorts: (value: function ValueGetter($dependency) {})
       */
    function invokeResolve(key, value) {
      var isFunction = function(target)  { return typeof target === 'function'; };
      var response = isFunction(value) ? $injector.invoke(value, null, resolve) : $injector.get(value);
      return $q.when(resolve[key] = response);
    }

    function resolveLocals(map) {
        var defer = $q.defer(),
           runResolveAt = function(index) {
              var key = Object.keys(map)[index];

              if ( map[key] ) {
                  var origVal = map[key];
                  invokeResolve(key, origVal).then(function(newVal) {
                    // When the promise returned a value, then we can immediately update the property to
                    // be able to inject it into later resolve invokes.
                    map[key] = newVal || origVal;
                    runResolveAt(index + 1);
                  });
              } else {
                 announceComplete()
              }
           };

      // Start the resolve at index 0
      runResolveAt(0);

      function announceComplete() {
        var loadRemoteTemplate = function() {
            return  $http.get(templateUrl, {cache: $templateCache})
                         .then(function(response) {
                           return response.data;
                         });
        };

        // Add the locals, which are just straight values to inject
        // eg locals: { three: 3 }, will inject three into the controller

        $q.all( angular.extend( map, locals, {
                '$template' : templateUrl ? loadRemoteTemplate() : $q.when(template)
           }))
           .then(defer.resolve, defer.reject);
      }

      return defer.promise;
    }

    // Wait for all the resolves to finish if they are promises
    return resolveLocals(resolve).then(function(locals) { ... };
```

DevVersion:
Closing this, and probably resubmitting with planned changes to the compiler.

PR #8468, opened 2016-05-16
fix($mdIcon): prevent "Possibly unhandled rejection" errors

gkalpak:
The `announceNotFound()` function was creating and returning a rejected promise that was neither
handled nor exposed to the user, causing "Possibly unhandled rejection" errors.

This should fix the failing tests against Angular snapshot on CI.

<sub>
This PR also includes a `refactor` commit (removing some unnecessary module loading and making whitespace a little more consistent :stuck_out_tongue:) and a `test` commit that more thoroughly covers the actual behavior when icons/groups are not found (i.e. logging + rejecting).
</sub>

### DIFF
--- a/src/components/icon/icon.spec.js
+++ b/src/components/icon/icon.spec.js
@@ -4,22 +4,20 @@ describe('MdIcon directive', function() {
   var $compile;
   var $mdIconProvider;
 
-  beforeEach(module('material.core'));
-  beforeEach(module('material.components.icon'));
-  beforeEach(module('material.components.icon',function(_$mdIconProvider_){
-     $mdIconProvider = _$mdIconProvider_;
-   }));
-   afterEach( function() {
-     $mdIconProvider.defaultFontSet('material-icons');
-     $mdIconProvider.fontSet('fa', 'fa');
-   });
+  beforeEach(module('material.components.icon', function(_$mdIconProvider_) {
+    $mdIconProvider = _$mdIconProvider_;
+  }));
+  afterEach(function() {
+    $mdIconProvider.defaultFontSet('material-icons');
+    $mdIconProvider.fontSet('fa', 'fa');
+  });
 
 
   describe('for font-icons:', function () {
 
-    beforeEach( inject(function($rootScope, _$compile_){
-        $scope = $rootScope;
-        $compile = _$compile_;
+    beforeEach(inject(function($rootScope, _$compile_) {
+      $scope = $rootScope;
+      $compile = _$compile_;
     }));
 
 
@@ -339,19 +337,18 @@ describe('MdIcon service', function() {
   var $scope;
   var $mdIconProvider;
 
-  beforeEach(module('material.core'));
-  beforeEach(module('material.components.icon',function(_$mdIconProvider_){
+  beforeEach(module('material.components.icon', function(_$mdIconProvider_) {
     $mdIconProvider = _$mdIconProvider_;
     $mdIconProvider
-      .icon('android'     , 'android.svg')
-      .icon('c2'          , 'c2.svg')
-      .icon('notfound'    ,'notfoundicon.svg')
-      .iconSet('social'   , 'social.svg' )
-      .iconSet('notfound' , 'notfoundgroup.svg' )
+      .icon('android'    , 'android.svg')
+      .icon('c2'         , 'c2.svg')
+      .icon('notfound'   , 'notfoundicon.svg')
+      .iconSet('social'  , 'social.svg' )
+      .iconSet('notfound', 'notfoundgroup.svg' )
       .defaultIconSet('core.svg');
   }));
 
-  beforeEach(inject(function($templateCache, _$httpBackend_, _$mdIcon_, $rootScope){
+  beforeEach(inject(function($templateCache, _$httpBackend_, _$mdIcon_, $rootScope) {
     $mdIcon = _$mdIcon_;
     $httpBackend = _$httpBackend_;
     $scope = $rootScope;
@@ -362,10 +359,9 @@ describe('MdIcon service', function() {
 
     $httpBackend.whenGET('notfoundgroup.svg').respond(404, 'Cannot GET notfoundgroup.svg');
     $httpBackend.whenGET('notfoundicon.svg').respond(404, 'Cannot GET notfoundicon.svg');
-
   }));
 
-  describe('should configure fontSets',function() {
+  describe('should configure fontSets', function() {
 
     it('with Material Icons by default', function () {
       expect($mdIcon.fontSet()).toBe('material-icons');
--- a/src/components/icon/icon.spec.js
+++ b/src/components/icon/icon.spec.js
@@ -510,29 +510,39 @@ describe('MdIcon service', function() {
     });
 
     describe('icon group is not found', function() {
-      it('should log Error', function() {
-        var msg;
-        try {
-          $mdIcon('notfound:someIcon')
-            .catch(function(error) {
-              msg = error.data;
-            });
+      it('should log Error and reject', inject(function($log) {
+        var errorMessage = 'Cannot GET notfoundgroup.svg';
+        var caughtRejection = false;
 
-          $httpBackend.flush();
-        } finally {
-          expect(msg).toEqual('Cannot GET notfoundgroup.svg');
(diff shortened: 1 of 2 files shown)


src/components/icon/icon.spec.js
```diff
@@ -514,29 +510,39 @@ describe('MdIcon service', function() {
...
 
-          $httpBackend.flush();
-        } finally {
-          expect(msg).toEqual('Cannot GET notfoundgroup.svg');
-        }
-      });
+        $httpBackend.flush();
+
+        expect(caughtRejection).toBe(true);
+        expect($log.warn.logs[0]).toEqual([errorMessage]);
```
ThomasBurleson: Is this a new feature? I did not know that that a log stack was maintained ?

gkalpak: No it's not new !
But even some tests in Angular itself seem to ignore this feature :)

gkalpak: (BTW, this is only for tests obviously.)

src/components/icon/icon.spec.js
```diff
@@ -514,29 +510,39 @@ describe('MdIcon service', function() {
...
+      }));
     });
 
     describe('icon is not found', function() {
-      it('should not throw Error', function() {
-        expect(function(){
-          $mdIcon('notfound');
+      it('should log Error and reject', inject(function($log) {
```
topherfangio: I believe we have a separate error below which tests this exact case. The test you overwrote was designed to test the case of "I don't want to catch the error, so I expect one not to be thrown".

In effect, isn't this a breaking change from the previous API?

Don't get me wrong, we went back and forth on whether to change/throw out this test because it seemed a bit odd! Just want to hear your opinion as well so we have a better idea of what the expected functionality is for the user 😄

gkalpak: > I believe we have a separate test below which tests this exact case.

I don't see any test below :stuck_out_tongue: 

That said, I don't think you need to explicitly test that something does not throw - if it throws the test will fail anyway.
The changes in this test only test the underlying behavior, I didn't change the behavior (in the corresponding commit), so there was no breaking change.

The question is what _is_ the desired behavior ? Previously, if the icon/group was not found, there was a "Possibly unhandled rejection" error (with v1.6) which the user would not be able to silence no matter what. With this PR, this error is will only be thrown if the user doesn't handle the rejection (i.e. not chain a `.catch()` handler to `$mdIcon('nofoundicon.svg')`).
If the desired behavior is that it should never raise a "Possibly unhandled rejection" error (even if the user never caught the rejection), then more modification is necessary.

If you ask me, I can't think of a good reason why I would want to silence a "Possibly unhandled rejection" for a non-existing icon, but there might be a usecase I am missing (related to icons).
If there are such usecases, you might want to make this configurable (either globally or per request - depending on what makes most sense to you). E.g. throw the error by default, but let the user specify that the `$mdIcon`-related unhandled rejection are OK and should be silenced.

topherfangio: The original behavior in 1.3, 1.4 and 1.5 was to not throw any error if the user called `$mdIcon()` with an unknown icon. Since 1.6 was going to throw a PUR, we decided to update the test to just check for the specific error (to account for 1.6) and instruct users to use `.catch()` or disable the new PUR checks if they wanted the old pre 1.6 behavior.

@ThomasBurleson Any thoughts on whether we should keep the "don't throw an error" behavior, or if we should make a breaking change to say that it will now throw an error/reject which should always be caught?

gkalpak: Just to be clear: This will reject in every version. The difference with 1.6 is that there will also be a PUR (Possibly Unhandled Rejection) error passed to the `$exceptionHandler`. In production, it will just log the error in the console (by default). In tests, it will throw an exception (by default).

--
BTW, the return value of `$mdIcon` is documented as:

> @returns {obj} Clone of the initial SVG DOM element; which was created from the SVG markup in the SVG data file.

This probably needs fixing. It actually returns a promise that resolves to the SVG clone, right ?

ThomasBurleson: @gkalpak  - Nice Point
@topherfangio  - ping

topherfangio: @gkalpak Just to confirm, you are saying that *this PR* will reject in all versions; however, in versions prior to 1.6, it would *not* reject, correct? Or am I missing something?

Totally agree regarding the documentation change. Have you already created an issue for it or should I?

src/components/icon/icon.spec.js
petebacondarwin: I.E. `ngMock` provides a mock version of `$log` that keeps a collection of the calls that are made to each of its logging methods.

src/components/icon/icon.spec.js
gkalpak: Prior to this PR, there would be **two** rejections in all versions. The one could be caught by the user, the other couldn't. So, in 1.6 there would be at least one PUR (which the user could not prevent in any way), which would log in roduction and throw in tests. 

All this PR changes in terms of functionality (i.e. apart from style and test changes), is that it drops the unnecessary, unexposed second rejection, leaving only one rejection that can be caught by the user if they want to.

gkalpak: Submitted #8489 for the doc fix.

topherfangio: Thanks for the docs fix!

So, I guess my question is: what happens if they just type `$mdIcon('notfound')`? You're saying that pre 1.6, it will work identically as before? It's just that in 1.6 they will get a PUR?

If so, then I think this is good.

gkalpak: tl;dr: Yes

Longer answer:

**Before this PR**
- pre-1.6:
  - `$mdIcon('notfound')`: Nothing happens
  - `$mdIcon('notfound').catch(callback)`: `callback` gets called
- post-1.6:
  - `$mdIcon('notfound')`: 2 PURs (thrown in tests, logged in production)
  - `$mdIcon('notfound').catch(callback)`: `callback` gets called + 1 PUR (thrown in tests, logged in production)

**After this PR**
- pre-1.6:
  - `$mdIcon('notfound')`: Nothing happens
  - `$mdIcon('notfound').catch(callback)`: `callback` gets called
- post-1.6:
  - `$mdIcon('notfound')`: 1 PUR (thrown in tests, logged in production)
  - `$mdIcon('notfound').catch(callback)`: `callback` gets called

topherfangio: Awesome, thanks for taking the time to write that all down! I think this is good to go!

topherfangio: Ha! Okay, I guess it was "good to go" two days ago when it got merged :-D

Sorry for the long discussion, but thanks for walking me through it!

src/components/icon/icon.spec.js
ThomasBurleson: Love this learning tips!

topherfangio: @gkalpak @petebacondarwin Is this different than the `$exceptionHandler` stack that is maintained?

gkalpak: @topherfangio: It isn't related to `$exceptionHandler`. 

`ngMock`'s [`$log` service][1] will keep track of the arguments passed when calling any of its methods. (It also has additional, handy methods for resetting and asserting emptyness).

Although the core (production) `$exceptionHandler` basically delegates to `$log.error()`, `ngMock`'s `$exceptionHandler` doesn't use `$log` at all. It just stores errors in an array (exposed as `$exceptionHandler.errors`) and might also throw if the mode is set to `rethrow` (the default).

[1]: https://docs.angularjs.org/api/ngMock/service/$log

petebacondarwin: In tests `$exceptionHandler` is overridden by `ngMock`. This test version has two modes `log` and `rethrow`. In both cases the exception is written to the `$exceptionHandler.errors` array.

Also in tests the `$log` is overridden by `ngMock`. This test version does not write to the console but instead writes to arrays, one for each of the possible logging methods, `log`, `warn`, `info`, `error`, `debug`.

In this particular test, I believe that we are not hitting `$exceptionHandler` at all as we are handling the error and calling `$log.warn`.

Issue #8030, opened 2016-04-14
md-calendar not showing inside md-sidenav


flick36:
The md-calendar only renders the top row of day names but not the calendar itself when inside a md-sidenav
 
http://codepen.io/flick36/pen/jqzGqN

topherfangio:
Indeed, the `md-calendar-scroll-mask` appears to have `width: 0px` which makes it invisible.

Issue #4650, opened 2015-09-15
md-datepicker should open on focus by default


jelbourn:
The datepicker should open its calendar pane as soon as it is focused by default, with an attribute option to go back to the explicit opening behavior. Focus will remain in the text input.

opadillaintegrify:
For what it's worth, consider a disabled user that can't use a mouse and has to tab through all of the inputs.  The most common case is that the user simply wants to tab past the date without changing it. 

Now, by opening on focus by default, you're making the disabled user have to escape out of the calendar every time.

jelbourn:
My plan is for nothing to change for screen reader users; the calendar will be visually present, but hidden to screen readers until "opened".

opadillaintegrify:
Oh OK.  I thought that users with mobility impairments would use adaptive keyboards/alternative mice but no screen reader, and that screen readers would be used mostly by those with visual/audio impairments.

jelbourn:
I meant for keyboard interaction as well. Even now, the calendar is not in the tab order; you have to use a specific keyboard shortcut (Alt + Down) to move focus into the calendar.

mattdanna:
@jelbourn Is there a workaround to get this working until it officially lands in master?

jelbourn:
@mattdanna You could set up a focus handler on the datepicker's input and then programatically click the "open calendar" button, but it will probably mess up the keyboard flow.

wizzzet:
+1

daniel-nagy:
Example of how you can achieve this in the meantime.

```javascript
angular.module('myApp').directive('autoOpen', function () {
  
  function compile(tElement) {
    tElement.find('input').attr('ng-focus', 'ctrl.onFocus($event)');
    
    return function (scope, element, attrs, datePicker) {
      let focused = false;
      
      datePicker.onFocus = (event) => {
        focused = !focused;
        
        // The datepicker is going to return focus to the input when
        // the calendar pane is closed. We don't want to reopen the
        // the calendar pane after it is closed.
        if(focused) {
          datePicker.openCalendarPane(event);
        }
      };
    };
  }
  
  return {
    compile: compile,
    priority: -1,
    require: 'mdDatepicker',
    restrict: 'A'
  };
});
```

```html
<md-datepicker auto-open></md-datepicker>
```

Personally, I believe the user should hit enter before the calendar pane opens when the date picker is focused via tabbing. In addition, the arrow button becomes obsolete.

```javascript
[...]

PR #8472, opened 2016-05-16
feat(datepicker): year view, open on focus, bug fixes

### LINKED ISSUE #4251: md-datepicker should have option to select year and month
If we have to select future of past dates, (date of birth of something), it is difficult to scroll up and down. Should have selection of year and month.

crisbeto:
* Adds the ability for users to switch years quicker by clicking on the month header. Cheers to @mhchen for helping me get started.
* Adds the ability to open a datepicker by focusing the input.
* Fixes alignment issues, if the user has set the `box-sizing: border-box` on everything.
* Fixes cases where the calendar's width could be set to 0.

Fixes #4251, #4650, #8547, #8030, #8557.


crisbeto:
@ThomasBurleson how should we proceed with reviewing this one? I had to shuffle around a lot of code in order to avoid duplication.

src/components/datepicker/calendar.js
```diff
@@ -182,350 +104,173 @@
...
+      var value = this.$viewValue;
+
+      // Notify the child scopes of any changes.
+      self.$scope.$broadcast('md-calendar-parent-changed', value);
+
+      // Set up the selectedDate if it hasn't been already.
+      if (!self.selectedDate) self.selectedDate = value;
+
+      // Also set up the displayDate.
+      if (!self.displayDate) self.displayDate = self.selectedDate || self.today;
```
jelbourn: Add braces

src/components/datepicker/calendar.spec.js
```diff
@@ -66,16 +66,30 @@ describe('md-calendar', function() {
     return null;
   }
 
+  /** Find the `tbody` for a year in the calendar. */
+  function findYearElement(year) {
+    var years = element.querySelectorAll('[md-calendar-year-body]');
+    var yearHeader = year.toString();
```
jelbourn: Can we not stamp the year element with an attribute to more directly query it, similar to how the date cells operate?

jelbourn: Nevermind.

src/components/datepicker/calendarYear.js
```diff
@@ -0,0 +1,258 @@
...
+  };
+
+  /**
+   * Hides the vertical scrollbar on the calendar scroller by setting the width on the
+   * calendar scroller and the `overflow: hidden` wrapper around the scroller, and then setting
+   * a padding-right on the scroller equal to the width of the browser's scrollbar.
+   *
+   * This will cause a reflow.
+   */
+  CalendarYearCtrl.prototype.hideVerticalScrollbar = function() {
```
jelbourn: Can we refactor so that the year and month share code?

jelbourn:
High level LGTM from me, thanks!

crisbeto:
@ThomasBurleson @topherfangio @EladBezalel @DevVersion I've worked through the feedback from Jeremy. Can you give me some feedback on this?

src/components/datepicker/calendar.js
```diff
@@ -182,350 +104,206 @@
...
     var self = this;
+
+    self.ngModelCtrl = ngModelCtrl;
+
     ngModelCtrl.$render = function() {
-      self.changeSelectedDate(self.ngModelCtrl.$viewValue);
+      var value = this.$viewValue;
+
+      // Notify the child scopes of any changes.
+      self.$scope.$broadcast('md-calendar-parent-changed', value);
```
EladBezalel: I think we should avoid broadcasting, but if it's inevitable  I think it should be `md-calendar:parent-changed`

EladBezalel: I think we should avoid broadcasting, but if it's inevitable  I think it should be `md-calendar:parent-changed`

You can avoid broadcasting by providing a `registerChanges` that you provide on the controller that gets a callback and saves it in a callback array and when the change happened go through the callback array and call it

src/components/datepicker/calendar.js
```diff
@@ -182,350 +104,206 @@
...
    */
-  CalendarCtrl.prototype.buildInitialCalendarDisplay = function() {
-    this.buildWeekHeader();
-    this.hideVerticalScrollbar();
-
-    this.displayDate = this.selectedDate || this.today;
-    this.isInitialized = true;
+  CalendarCtrl.prototype.setNgModelValue = function(date) {
+    var value = this.dateUtil.createDateAtMidnight(date);
+    this.$scope.$emit('md-calendar-change', value);
```
EladBezalel: Where we listen for this event?

src/components/datepicker/calendar.js
```diff
@@ -182,350 +104,206 @@
...
+      case keyCode.END: return 'end';
 
-    var cellId = this.getDateId(date);
-    var cell = document.getElementById(cellId);
-    if (cell) {
-      cell.classList.add(FOCUSED_DATE_CLASS);
-      cell.focus();
-    } else {
-      this.focusDate = date;
+      default: return null;
```
EladBezalel: You can just use return

src/components/datepicker/calendar.js
crisbeto: Doing a `registerChanges` method is basically the same as broadcasting it on the scope, except that we'd need to handle keeping track of the callbacks and cleaning up afterwards ourselves.

src/components/datepicker/calendar.js
crisbeto: It's used on parent scopes to know that the calendar changed. In this case it's only used in the datepicker, in order to know when to close the calendar.

crisbeto:
It should be good to go now @ThomasBurleson.

topherfangio:
Overall, I think this is really great and helpful!

A couple of thoughts from a UX perspective:

1. I think the year should have some additional feedback that it is clickable. Not sure the best way to do this, but maybe just changing the background on hover is enough (like we do with the days).

2. Without an animation between the two, it feels very jumpy and "un-material" IMO. Would it be fairly easy to add a slide or simple fade transition between the two?

3. Would it be possible to re-use the "day of the week" header area to provide a back button/icon? Right now, because the header disappears, it changes the size of the picker which feels odd, and we also don't give users an easy way to get back to the standard view if they decide they don't want to change anything after all.

Just some of my own personal thoughts 😄

crisbeto:
@topherfangio 

1. Definitely, but I didn't find a nice way of doing it. [I've been trying stuff like this](http://i.imgur.com/mhTInNc.png), but I wouldn't say that it looks good. [I was leaning more towards something like this.](http://i.imgur.com/Z5IKIVA.png)

2. That sounds interesting! I couldn't get it to work nicely on the first try (for some reason the one with the datepicker didn't transition at all and the one outside a datepicker wasn't rendering the month content immediately), but I might be able to get it working.

3. I'm not sure about this one, that's a would be a pretty big back button.

ThomasBurleson:
@crisbeto - Should i merge this now ?

crisbeto:
I'd say so @ThomasBurleson. Once we have this set up, making improvements should be a bit easier.

PR #8496, opened 2016-05-18
fix/perf(tabs): Fix performance and visual issues.

### LINKED ISSUE #5681: Listening for DOMSubtreeModified during compilation dramatically impacts performance
The [tabsTemplateDirective](https://github.com/angular/material/blob/386e2acb6badcfd086d04cb8a876cb49468cd12a/src/components/tabs/js/tabsTemplateDirective.js) attaches an event listener for the DOMSubtreeModified event.  Tabs with complex hierarchies of nested directives tend to modify the DOM a lot during compilation, so this event gets triggered a lot.  Despite the fact that most of the time the handler isn't actually doing much due to the debounce, just the act of dispatching the event has a dramatic effect on performance (we saw an initial compilation go from 2-3 seconds down to around a second just by removing this handler, with no appreciable effects on functionality).

Is the attachment of this event handler something that can be delayed?  It seems to only impact the initial compilation, so perhaps attaching the handler later on could result in the same behavior without the performance overhead?

topherfangio:
Initially, the tabs used DOMSubtreeModified to listen for changes and update the height, pagination and inkbar styles. At some point, we used a different method to update the height, but we were still watching all DOM changes and updating the pagination and inkbar styles whenever the contents changed.

Fix this with the following two changes:

1. Move the DOM watching to a new md-tabs-dummy-wrapper directive which only watches for changes to the tab's labels (not the content too) so that the updates are called less frequently.

2. Replace DOMSubtreeModified watching with new MutationObservers which are more drastically more performant.

Additionally, some recent changes caused some bugs in the tabs by using a cached version of the internal elements instead of calling `getElements()` which searches the DOM and refreshes the cached elements.

Fix by adding more calls to `getElements()` where appropriate to properly update the cache after tabs may have changed.

Fixes #5681. References #4940, and might cause PR #8293 to need a rebase if this goes in first.

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -159,6 +159,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
    * @param stretchTabs
    */
   function handleStretchTabs (stretchTabs) {
+    var elements = getElements();
     angular.element(elements.wrapper).toggleClass('md-stretch-tabs', shouldStretchTabs());
     updateInkBarStyles();
   }
@@ -169,6 +170,8 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
 
   function handleMaxTabWidth (newWidth, oldWidth) {
     if (newWidth !== oldWidth) {
+      var elements = getElements();
+
       angular.forEach(elements.tabs, function(tab) {
         tab.style.maxWidth = newWidth + 'px';
       });
@@ -200,7 +203,9 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
    * @param left
    */
   function handleOffsetChange (left) {
+    var elements = getElements();
     var newValue = ctrl.shouldCenterTabs ? '' : '-' + left + 'px';
+
     angular.element(elements.paging).css($mdConstant.CSS.TRANSFORM, 'translate3d(' + newValue + ', 0, 0)');
     $scope.$broadcast('$mdTabsPaginationChanged');
   }
@@ -212,7 +217,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
    */
   function handleFocusIndexChange (newIndex, oldIndex) {
     if (newIndex === oldIndex) return;
-    if (!elements.tabs[ newIndex ]) return;
+    if (!getElements().tabs[ newIndex ]) return;
     adjustOffset();
     redirectFocus();
   }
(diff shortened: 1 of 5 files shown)


topherfangio:
Shoutout to @crisbeto for helping me track down the fact that the MutationObservers are async and require async tests!

Also, @crisbeto - Can you review when you get a chance? 😄

src/components/tabs/js/tabsController.js
```diff
@@ -169,6 +170,8 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
 
   function handleMaxTabWidth (newWidth, oldWidth) {
     if (newWidth !== oldWidth) {
+      var elements = getElements();
```
crisbeto: `getElements` queries the DOM, right? I _think_ (not really sure) it could be avoided by using the data from the mutation observer, which should return a list of the added/removed nodes, which can be used to update the references in the list.

topherfangio: You are correct, it does query the DOM, and we might be able to use a MutationObserver instead, but this function doesn't always get called from a MutationObserver. For example, it is called when the component is initialized.

If we have some hard data that suggests rewriting this to use the MutationObserver makes a drastic difference (verses querying the DOM on updates), then I'm all for it 😄

crisbeto: Alright, makes sense. It starts querying from the element anyway, so it shouldn't be too much of a difference.

crisbeto:
LGTM @ThomasBurleson

src/components/tabs/js/tabsDirective.js
```diff
@@ -154,7 +154,7 @@ function MdTabs () {
                   'md-scope="::tab.parent"></md-tab-item> ' +
               '<md-ink-bar></md-ink-bar> ' +
             '</md-pagination-wrapper> ' +
-            '<div class="_md-visually-hidden md-dummy-wrapper"> ' +
+            '<md-tabs-dummy-wrapper class="_md-visually-hidden md-dummy-wrapper"> ' +
```
ThomasBurleson: why is this a `dummy` wrapper?

topherfangio: I was going based off of the original class name of `md-dummy-wrapper` which is still applied to the class.

This wrapper is visually hidden from the screen and is used to calculate the widths/heights so that it can properly position the real tabs. It's basically a way of doing off-screen rendering/double-buffering (in a way).

I'm happy to change the name if you have something better, but I think we should leave the classname the same in case someone is doing something with it (since we didn't mark it as private).

ThomasBurleson:
lgtm

PR #8523, opened 2016-05-20
update(docs): Updated README CSS CDN version

cpamp21:


cpamp21:
I signed CLA

ThomasBurleson:
That is simply an example. We will update when v1.1.0 is released.

cpamp21:
@ThomasBurleson I still think the comment or the link should be changed. Example or not, it's confusing and wrong.

PR #8526, opened 2016-05-22
feat(datepicker): add the ability to open the calendar when the input is focused

### LINKED ISSUE #4650: md-datepicker should open on focus by default

crisbeto:
Adds the "md-open-on-focus" attribute, which when specified, opens the datepicker's calendar when the input is focused.

Fixes #4650.

### DIFF
--- a/src/components/datepicker/datePicker.js
+++ b/src/components/datepicker/datePicker.js
@@ -26,6 +26,7 @@
    * @param {Date=} md-max-date Expression representing a max date (inclusive).
    * @param {(function(Date): boolean)=} md-date-filter Function expecting a date and returning a boolean whether it can be selected or not.
    * @param {String=} md-placeholder The date input placeholder value.
+   * @param {String=} md-open-on-focus When present, the calendar will be opened when the input is focused.
    * @param {boolean=} ng-disabled Whether the datepicker is disabled.
    * @param {boolean=} ng-required Whether a value is required for the datepicker.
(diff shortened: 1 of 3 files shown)


src/components/datepicker/datePicker.js
```diff
@@ -574,15 +579,24 @@
...
+        self.calendarPaneOpenedFrom.focus();
+        self.calendarPaneOpenedFrom = null;
+      }
 
-      this.documentElement.off('click touchstart', this.bodyClickHandler);
-      window.removeEventListener('resize', this.windowResizeHandler);
+      // Ensures that all focus events have fired before detaching
+      // the calendar. Prevents the calendar from reopening immediately
+      // in IE when md-open-on-focus is set.
+      this.$mdUtil.nextTick(function() {
```
DevVersion: Wouldn't it make sense to only run the detach-block in the next tick, if the `md-open-on-focus` attribute is set to true? Otherwise you should be able to run this immediately.

crisbeto: Yeah, good point.

ThomasBurleson:
@crisbeto  - please rebase from master!

crisbeto:
Closing since this was included with #8472 as discussed.

PR #8558, opened 2016-05-26
Add required attribute for accessibility

atscott:
The aria-label attribute is required for accessibility on the datepicker button.

### DIFF
--- a/src/components/datepicker/datePicker.js
+++ b/src/components/datepicker/datePicker.js
@@ -54,6 +54,7 @@
           // may be confusing.
           '<md-button class="md-datepicker-button md-icon-button" type="button" ' +
               'tabindex="-1" aria-hidden="true" ' +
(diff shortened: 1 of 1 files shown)


atscott:
I signed it!

src/components/datepicker/datePicker.js
```diff
@@ -54,6 +54,7 @@
           // may be confusing.
           '<md-button class="md-datepicker-button md-icon-button" type="button" ' +
               'tabindex="-1" aria-hidden="true" ' +
+              'aria-label="{{::ctrl.dateLocale.msgOpenCalendar}}" ' +
```
ThomasBurleson: Why is this needed, when aria-hidden == true ?

atscott:
Ah, sorry - didn't notice that. The problem I thought I was fixing was actually a bug in mdAria and was fixed in https://github.com/angular/material/commit/b3cb84d3fa9408fb3f7192020eaef0d2c82bd929

PR #8578, opened 2016-05-27
fix(theming): only insert default theme if its usable.

### LINKED ISSUE #7425: default theme is injected even when we use setDefaultTheme with another value

DevVersion:
* Currently we always generate the default theme, even when the defaultTheme has changed.
  This is not necessary and impacts the performance significant.

We should only inject the default theme, when it's usable.

@EladBezalel and I went through the logic again, and this one should be fine.

Fixes #7425.

### DIFF
--- a/src/core/services/theming/theming.js
+++ b/src/core/services/theming/theming.js
@@ -526,7 +526,7 @@ function parseRules(theme, colorType, rules) {
 var rulesByType = {};
 
 // Generate our themes at run time given the state of THEMES and PALETTES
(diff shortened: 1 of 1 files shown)


EladBezalel:
SGTM

Issue #8609, opened 2016-06-01
$mdMedia 'landscape' and 'portrait'


davidenke:
The `$mdMedia` service should return boolean for `landscape` and `portrait`. Changes in the layout may can be reflected by the corresponding css media queries, but it would be more consistent to assign e.g. class names with ng-styles in the templates using this service.

ThomasBurleson:
@davidenke - great idea. Please provide a pull request; at which time this issue will be reopened.

> This issue is closed as part of our ‘Surge Focus on Material 2' efforts.
For details, see our [forum posting](http://bit.ly/1UhZyWs).

PR #8617, opened 2016-06-02
added landscape and portrait queries

davidenke:
As mentioned by @ThomasBurleson in #8609 the landscape and portrait media queries have been added.

### DIFF
--- a/src/core/util/constant.js
+++ b/src/core/util/constant.js
@@ -56,15 +56,17 @@ function MdConstantFactory($sniffer) {
      *
      */
     MEDIA: {
-      'xs'    : '(max-width: 599px)'                         ,
-      'gt-xs' : '(min-width: 600px)'                         ,
-      'sm'    : '(min-width: 600px) and (max-width: 959px)'  ,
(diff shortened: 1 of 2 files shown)


davidenke:
I signed it!

DevVersion:
We should probably support these new breakpoints in the layout engine as well.
> https://github.com/angular/material/blob/master/src/core/services/layout/layout.js#L80

So you can use for example:
```html
<div layout-landscape="row">
```

ThomasBurleson:
@daniel-nagy My only concern here is the potential large increase in generated Layout CSS. 

> I will investigate.

ThomasBurleson:
@davidenke - So my worry was misplaced. Your change is only for the `$mdMedia` service and does not impact the Layout API nor its generated Layout CSS.

ThomasBurleson:
@DevVersion - perhaps we could just support the `layout-landscape` and `layout-portrait` features without any other enhancements ?

DevVersion:
@ThomasBurleson I agree, that would keep the CSS size small and would be still super useful enhancement.

PR #8627, opened 2016-06-02
fix(datepicker): disabled dates blending in with background on a dark theme

### LINKED ISSUE #8550: mdDatepicker: difficult to read text with dark theme
**Actual behavior**:
-  What is the issue? 

The datepicker is not playing nice with dark themes.  Disabled days do not show up as greyed out, but an unreadable white.  The date picker dark styling wasn't quite what I expected either -- considering most of it is light.  

Unreadable days are visible, perhaps a retuning of the dark styling for the datepicker. 

**CodePen**  
* http://codepen.io/mckenzielong/pen/NrKpRP
* Click on the 5th and 6th date picker and note the unreadable dates
* Bonus potentially unrelated issue (opening near an edge seems to produce all unreadable dates?)

**Angular Versions**: *
-  Angular Version: 1.5.5
-  Angular Material Version: 1.1.0 rc4

**Additional Information**:
-  Browser Type: Opera 37/ Chrome 50
-  OS: Windows 7 64bit

crisbeto:
The disabled dates were blending in with the calendar when used together with a dark theme.

Fixes #8550.

### DIFF
--- a/src/components/datepicker/calendar-theme.scss
+++ b/src/components/datepicker/calendar-theme.scss
@@ -44,6 +44,10 @@
 
   .md-calendar-date-disabled,
   .md-calendar-month-label-disabled {
-    color: '{{foreground-3}}';
+    // Note that this uses half the opacity of the default text color,
+    // because the calendar is white, even on the dark theme, otherwise
+    // the default disabled color `foreground-3` blends in with the
+    // background.
+    color: '{{background-A200-0.435}}';
   }
 }


src/components/datepicker/calendar-theme.scss
```diff
@@ -44,6 +44,6 @@
 
   .md-calendar-date-disabled,
   .md-calendar-month-label-disabled {
-    color: '{{foreground-3}}';
+    color: '{{background-A200-0.435}}';
```
DevVersion: Maybe you can add a comment why you use this hue and opacity.

crisbeto: Will do.

PR #8637, opened 2016-06-03
minor doc fix: changed erroneous `flex="22"` to `flex="20"`

areologist:
Example code had `flex="22"` while MD spec and angular material docs say multiples of 5. Changed to `flex="20"`


areologist:
I signed it!

PR #8644, opened 2016-06-03
feat(util): add the ability to pass in a predicate function

crisbeto:
Note that this is a nice-to-have that I found myself needing while working on the datepicker. I eventually worked around it, but it could be good for the future.

* It could be useful to be able to match node in `getClosest` by more than their tag name. This change adds the ability to pass in a function instead.
* Adds a few unit tests for `getClosest`, since it didn't have any.

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -472,19 +472,29 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
      * getClosest replicates jQuery.closest() to walk up the DOM tree until it finds a matching nodeName
      *
      * @param el Element to start walking the DOM from
-     * @param tagName Tag name to find closest to el, such as 'form'
+     * @param check Either a string or a function. If a string is passed, it will be evaluated against
+     * each of the parent nodes' tag name. If a function is passed, the loop will call it with each of
+     * the parents and will use the return value to determine whether the node is a match.
      * @param onlyParent Only start checking from the parent element, not `el`.
      */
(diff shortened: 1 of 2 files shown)


src/core/util/util.js
```diff
@@ -472,16 +472,28 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
      *
      * @param el Element to start walking the DOM from
-     * @param tagName Tag name to find closest to el, such as 'form'
+     * @param check Either a string or a function. If a string is passed, it will be evaluated against
+     * each of the parent nodes' tag name. If a function is passed, the loop will call it with each of
+     * the parents and will use the return value to determine whether the node is a match.
      * @param onlyParent Only start checking from the parent element, not `el`.
      */
-    getClosest: function getClosest(el, tagName, onlyParent) {
+    getClosest: function getClosest(el, check, onlyParent) {
```
ThomasBurleson: Refactored: 

```js
getClosest: function getClosest(el, check, onlyParent) {
  if (el instanceof angular.element) el = el[0];
  var tagName = angular.isString(check) ? check : "", 
      predicate = angular.isFunction(check) ? check : function(el) {
        return el.nodeName === tagName;
      }; 

  if (onlyParent) el = el.parentNode;
  if (!el) return null;
  
  do {
    if (predicate(el)) {
      return el;
    }
  } while (el = el.parentNode);
  
  return null;
}
```

crisbeto: Ah even better. Updated.

PR #8645, opened 2016-06-03
fix(datepicker): icon alignment

crisbeto:
Some of the recent changes was throwing the datepicker's alignment off. This gets it back to normal.


ThomasBurleson:
@crisbeto  - Darn it. I wish this was in rc.5

PR #8660, opened 2016-06-05
chore(): enable chrome testing for karma

DevVersion:
* Adding Chrome to the Karma by using SauceLabs, which runs parallel to the normal Tests, which means we won't lose any time
* Accelerating Travis CI build time


ThomasBurleson:
@DevVersion - FYI - Travis build time was 6min... now 12min.

DevVersion:
@ThomasBurleson 

You are counting the total time, (both jobs calculated together with idle).
- The elapsed time, was 7 min.

 Also it was slower on both jobs, because the Cache was rebuilding
- 
image


So actually, the next commits should be faster ;)

PR #8672, opened 2016-06-06
fix(select): Fix pristine/dirty error, scope conflict, and many styles.

### LINKED ISSUE #7988: bug(select): select menu should not be underlined

### LINKED ISSUE #8527: 'md-select' causes 'md-fab-speed-dial' to trigger
**Actual behavior**:
-  What is the issue? I have md-select and md-fab-speed-dial in my code, and when I click on the md-select element it also opens up the md-fab-speed-dial element.
-  What is the expected behavior? These are seperate elements, md-fab-speed-dial shouldn't open.

-  CodePen Demo which shows your issue : http://codepen.io/drorsu/pen/XdvddG
-  Details: Just click on the select menu.

**Angular Versions**: *
-  Angular Version: 1.5.3
-  Angular Material Version: 1.1.0-rc2

**Additional Information**:
-  Browser Type: Chrome
-  Browser Version: 50.0.2661.102
-  OS: 
-  Stack Traces:

----

  Add more lines if needed.

### LINKED ISSUE #8529: md-select required not working on pristine input
**Actual behavior**:
-  What is the issue? 
User submit form but md-select required attribute is ignored
md-select should show required message and red borders around the `md-input-container` directly if the user hits `SAVE`

-  CodePen Demo which shows your issue : http://codepen.io/anon/pen/jqoEOG
-  Details: 
User hits `SAVE` button and there is no indicator that the `md-select` is required. This is misleading and it is currently impossible to show `ng-messages` to required attribute on `md-select` elements.
If the User selects `FAVORITE NUMBER` and hits `CLEAR` the `ng-messages` are shown correctly

**Angular Versions**: *
-  Angular Version: 1.5.5
-  Angular Material Version: 1.0.8

**Additional Information**:
-  Browser Type: every
-  Browser Version: every
-  OS: every

topherfangio:
Fix styles and code to follow pristine/dirty styling of other input elements and provide CSS class for stand-alone usage.

 - Select now behaves like a normal input, appearing as invalid
   if the user focuses/blurs the element, or submits the form,
   without selecting an option.
 - Fix issues with floating labels not working on focus.
 - Add new `md-no-underline` CSS class to allow for stand-alone
   usage (non-form).
 - Update demos to show new stand-alone usage with required
   example.
 - Standardize asterisk visibility when required when standalone
   or inside of a `<md-input-container>`

Additionally, the select component currently sets the `isOpen` variable on the `$scope`. This can cause conflicts if the user has their own `isOpen` variable on the scope.

Fix by privatizing our own variable to `_mdSelectIsOpen` to reduce chances of a conflict.

Fixes #8529. Fixes #7988. Fixes #8527.

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -715,7 +715,21 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
       return value === '' || !!value && (negatedCheck === false || value !== 'false' && value !== '0');
     },
 
-    hasComputedStyle: hasComputedStyle
+    hasComputedStyle: hasComputedStyle,
+
+    /**
+     * Returns true if the parent form of the element has been submitted.
+     * 
+     * @param element An Angular or HTML5 element.
+     * 
+     * @returns {boolean}
+     */
+    isParentFormSubmitted: function(element) {
+      var parent = $mdUtil.getClosest(element, 'form');
+      var form = parent ? angular.element(parent).controller('form') : null;
+
+      return form ? form.$submitted : false;
+    }
   };
 
 
--- a/src/components/input/demoErrors/index.html
+++ b/src/components/input/demoErrors/index.html
@@ -12,13 +12,23 @@
         </div>
       </md-input-container>
 
-      <md-input-container class="md-block">
-        <label>Client Name</label>
-        <input required name="clientName" ng-model="project.clientName">
-        <div ng-messages="projectForm.clientName.$error">
-          <div ng-message="required">This is required.</div>
-        </div>
-      </md-input-container>
+      <div layout="row">
+        <md-input-container flex="50">
+          <label>Client Name</label>
+          <input required name="clientName" ng-model="project.clientName">
+          <div ng-messages="projectForm.clientName.$error">
+            <div ng-message="required">This is required.</div>
+          </div>
+        </md-input-container>
+
+        <md-input-container flex="50">
+          <label>Project Type</label>
+          <md-select name="type" ng-model="project.type" required>
+            <md-option value="app">Application</md-option>
+            <md-option value="web">Website</md-option>
+          </md-select>
+        </md-input-container>
+      </div>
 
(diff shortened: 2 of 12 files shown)


topherfangio:
@ThomasBurleson @EladBezalel @DevVersion @crisbeto Can you guys all review this and actually test out the new select demos? Ideally, this would be on a mix of Windows/Mac/iOS/Android/etc.

crisbeto:
@topherfangio In general LGTM on Chrome, Firefox and iOS Safari, however it seems kind of trippy on IE11. Here's what I mean:

a

Note the dropdown in the lower right. It doesn't seem to happen in HEAD.

src/core/util/util.js
```diff
@@ -677,7 +677,21 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
+     * 
+     * @param element An Angular or HTML5 element.
+     * 
+     * @returns {boolean}
+     */
+    isParentFormSubmitted: function(element) {
+      var parent = $mdUtil.getClosest(element, 'form');
+      var form = parent ? angular.element(parent).controller('form') : null;
+
+      return form ? form.$submitted : false;
```
EladBezalel: i would get rid of form and just do:
```js
return parent ? angular.element(parent).controller('form').$submitted : false;
```

topherfangio: I actually started with that, but you may not always have a from controller (since this can be used outside of a form) in which case that would throw a NPE. Thus, why I also have a test for this situation 😉

topherfangio:
@crisbeto Thanks! I'll see if I can snag an IE11 machine around here somewhere (I think my father-in-law has one) and do some additional testing. Glad you caught that before it went into master 😄

topherfangio:
@ThomasBurleson I believe this is ready for review/merge now. Let me know if it needs any changes.

PR #8675, opened 2016-06-06
fix(menu): Fix icon/text alignment in Firefox.

### LINKED ISSUE #8464: bug(md-menu): icons misaligned on md-menu
Icons are not aligned.
Works on Safari and Chrome, not on Firefox

CodePen Demo which shows your issue : https://material.angularjs.org/HEAD/demo/menu

**Angular Versions**: *
-  Angular Version: 1.5.5
-  Angular Material Version: 1.1.0-rc4/head

**Additional Information**:
-  Browser Type: Firefox

topherfangio:
By default, Firefox provides focus border styling which was modifying the layout of the button elements within a menu item. The original fix was to use `display: flex` to fix this, however, Firefox currently
disregards the `display` property on the buttons and was thus causing the text to be on a separate line from the text.

Fix by using `display: inline-block` for both `<a>` and `<button>` elements and adding some special padding for the `<a>` tags to fix the top alignment. Also apply special Firefox-specific styling to remove the focus border/styling.

Fixes #8464.

### DIFF
--- a/src/components/menu/menu.scss
+++ b/src/components/menu/menu.scss
@@ -81,7 +81,25 @@ md-menu-item {
     padding-right: 2*$baseline-grid;
   }
 
+  /*
+   * We cannot use flex on <button> elements due to a bug in Firefox, so we also can't use it on
+   * <a> elements. Add some top padding to fix alignment since buttons automatically align their
+   * text vertically.
+   */
+  > a.md-button {
+    padding-top: 5px;
+  }
+
   > .md-button {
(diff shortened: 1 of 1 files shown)


topherfangio:
@DevVersion Can you please test this on Windows Edge/IE/Chrome/Firefox when you get a chance. I have done OS X Chrome/Safari/Firefox and it looks good.

topherfangio:
@ThomasBurleson Fix verified on Windows by @crisbeto.

PR #8678, opened 2016-06-06
docs(sidenav): whitespace in 'servic  e'[sic]

jsr6720:


jayant-yadav:
The whitespace in 'servic e' is still persistant.

PR #8702, opened 2016-06-09
fix:(md-select) option with id 0 not selected when there is a debounce definition

YonatanKra:
On select an option with id 0, with a debounce definition, the view did not change.


topherfangio:
This needs some unit tests and some more comments on why it is necessary.

Splaktar:
@YonatanKra is there any chance that you can squash your commits and add the unit test as requested? Thank you.

PR #8703, opened 2016-06-09
fix(md-nav-bar): clear the selected nav-item if it is null

### LINKED ISSUE #8694: md-nav-ink-bar stays visible when nav-item is no longer selected.
**Actual behavior**:
When you have a link that deselects the navbar item the `md-nav-ink-bar` stays selected.
I would expect that clicking a link located in the navbar which does not correspond to a page (for example dropdown or a link to a page which is not in the nav bar) that the ink-bar would also disappear.

 http://codepen.io/anon/pen/GqpByo
-  Click a nav-item see that it is selected.
-  Click the link at the end of the nav-bar, see that the previously selected nav-item is no longer selected but the `md-nav-ink-bar` is.

**Angular Versions**: *
-  Angular Version: ~1.5.3
-  master and 1.1.0-rc.5

**Additional Information**:
-  Browser Type: *
-  Browser Version: *
-  OS: *
-  Stack Traces:

fhernandezn:
toggle the class md-button if there is no a new value for the selected tab

Fixes #8694

### DIFF
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -209,9 +209,9 @@ MdNavBarController.prototype._initTabs = function() {
 MdNavBarController.prototype._updateTabs = function(newValue, oldValue) {
   var tabs = this._getTabs();
 
-  var oldIndex;
+  var oldIndex, oldTab;
   if (oldValue) {
-    var oldTab = this._getTabByName(oldValue);
+    oldTab = this._getTabByName(oldValue);
     if (oldTab) {
       oldTab.setSelected(false);
       oldIndex = tabs.indexOf(oldTab);
@@ -228,6 +228,11 @@ MdNavBarController.prototype._updateTabs = function(newValue, oldValue) {
         self._updateInkBarStyles(tab, newIndex, oldIndex);
       });
     }
+  } else {
+    var self = this;
+    this._$timeout(function(){
+      self._updateInkBarStyles(oldTab, -1, oldIndex);
+    });
   }
 };
 
@@ -240,7 +245,9 @@ MdNavBarController.prototype._updateInkBarStyles = function(tab, newIndex, oldIn
   var left = tabEl.offsetLeft;
 
   this._inkbar.toggleClass('_md-left', newIndex < oldIndex)
-      .toggleClass('_md-right', newIndex > oldIndex);
+      .toggleClass('_md-right', newIndex > oldIndex)
+      .toggleClass('md-button', newIndex === -1);
+
   this._inkbar.css({left: left + 'px', width: tabEl.offsetWidth + 'px'});
 };
 
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -207,33 +207,26 @@ MdNavBarController.prototype._initTabs = function() {
  * @private
  */
 MdNavBarController.prototype._updateTabs = function(newValue, oldValue) {
+  var self = this;
   var tabs = this._getTabs();
+  var oldIndex;
+  var newIndex = -1;
+  var newTab = this._getTabByName(newValue);
+  var oldTab = this._getTabByName(oldValue);
 
-  var oldIndex, oldTab;
-  if (oldValue) {
-    oldTab = this._getTabByName(oldValue);
-    if (oldTab) {
-      oldTab.setSelected(false);
-      oldIndex = tabs.indexOf(oldTab);
-    }
+  if (oldTab) {
+    oldTab.setSelected(false);
+    oldIndex = tabs.indexOf(oldTab);
   }
 
-  if (newValue) {
-    var tab = this._getTabByName(newValue);
-    if (tab) {
-      tab.setSelected(true);
-      var newIndex = tabs.indexOf(tab);
-      var self = this;
-      this._$timeout(function() {
-        self._updateInkBarStyles(tab, newIndex, oldIndex);
-      });
-    }
-  } else {
-    var self = this;
-    this._$timeout(function(){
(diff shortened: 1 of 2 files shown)


src/components/navBar/navBar.js
```diff
@@ -240,7 +245,9 @@ MdNavBarController.prototype._updateInkBarStyles = function(tab, newIndex, oldIn
   var left = tabEl.offsetLeft;
 
   this._inkbar.toggleClass('_md-left', newIndex < oldIndex)
-      .toggleClass('_md-right', newIndex > oldIndex);
+      .toggleClass('_md-right', newIndex > oldIndex)
+      .toggleClass('md-button', newIndex === -1);
```
gmoothart: I do not understand this part; it is toggling a 'md-button' class on the inkbar, but there shouldn't be any such class.

src/components/navBar/navBar.js
```diff
@@ -228,6 +228,11 @@ MdNavBarController.prototype._updateTabs = function(newValue, oldValue) {
         self._updateInkBarStyles(tab, newIndex, oldIndex);
       });
     }
+  } else {
+    var self = this;
+    this._$timeout(function(){
+      self._updateInkBarStyles(oldTab, -1, oldIndex);
+    });
```
gmoothart: I think we can refactor to remove the need for a duplicated call to $timeout/updateInkbarStyler. How about (slightly more compact):

- get rid of the `if (newValue)` check (`tab` will be null)
- move the `$timeout` out of the `if(tab)` block
- change updateInkbarStyles to be able to handle a null tab

gmoothart:
Please also add a test.

ThomasBurleson:
@fhernandezn - will you make the changes requested... or should @gmoothart ?

src/components/navBar/navBar.js
fhernandezn: For some reason there is a md-button class that is adding the border-bottom color

screen shot 2016-06-11 at 10 09 21 am

If I don't toggle that class the ink from the text disappear but the border bottom one persists.

screen shot 2016-06-11 at 10 10 45 am

If I do toggle the class, the border-bottom color also disappears

screen shot 2016-06-11 at 10 11 41 am

src/components/navBar/navBar.js
```diff
@@ -209,9 +209,9 @@ MdNavBarController.prototype._initTabs = function() {
 MdNavBarController.prototype._updateTabs = function(newValue, oldValue) {
   var tabs = this._getTabs();
 
-  var oldIndex;
+  var oldIndex, oldTab;
   if (oldValue) {
```
fhernandezn: We can also delete this `if(oldValue)` because two lines below there is a validation `if(oldTab)` and `oldTab` will be never used again.

@gmoothart, what do you think?

src/components/navBar/navBar.js
```diff
@@ -207,41 +207,43 @@ MdNavBarController.prototype._initTabs = function() {
...
  */
 MdNavBarController.prototype._updateInkBarStyles = function(tab, newIndex, oldIndex) {
-  var tabEl = tab.getButtonEl();
-  var left = tabEl.offsetLeft;
-
   this._inkbar.toggleClass('_md-left', newIndex < oldIndex)
-      .toggleClass('_md-right', newIndex > oldIndex);
-  this._inkbar.css({left: left + 'px', width: tabEl.offsetWidth + 'px'});
+      .toggleClass('_md-right', newIndex > oldIndex)
+      .toggleClass('md-button', newIndex === -1);
```
gmoothart: The colors are defined on md-nav-bar directly:
https://github.com/angular/material/blob/master/src/components/navBar/navBar-theme.scss#L12

What this is doing is _adding_ the md-button class, which changes the layout such that the ink-bar is no longer visible.

Can you just set display:none if newIndex is < 0 ?

src/components/navBar/navBar.spec.js
```diff
@@ -77,6 +77,20 @@ describe('mdNavBar', function() {
...
+      createTabs();
+
+      updateSelectedTabRoute('');
+
+      expect(getTab('tab3')).not.toHaveClass('md-active');
+      expect(getTab('tab3')).not.toHaveClass('md-primary');
+      expect(getTab('tab2')).not.toHaveClass('md-active');
+      expect(getTab('tab2')).not.toHaveClass('md-primary');
+      expect(getTab('tab1')).not.toHaveClass('md-active');
+      expect(getTab('tab1')).not.toHaveClass('md-primary');
```
gmoothart: also add an expect for the ink-bar not being shown

src/components/navBar/navBar.js
```diff
@@ -207,41 +207,43 @@ MdNavBarController.prototype._initTabs = function() {
  * @private
  */
 MdNavBarController.prototype._updateTabs = function(newValue, oldValue) {
+  var self = this;
   var tabs = this._getTabs();
-
   var oldIndex;
```
gmoothart: can you init this to `-1` too?

fhernandezn: sure :)

src/components/navBar/navBar.js
fhernandezn: Yes, I can

What style do you prefer?

```javascript
if(newIndex < 0) {
  this._inkbar.css({display: 'none'});
} else {
  this._inkbar.css({display: ''}); // default value
}
```

or 

```javascript
this._inkbar.css({display: newIndex < 0 ? 'none' : ''});
```

gmoothart: Thanks, the ternary one is fine.

PR #8709, opened 2016-06-09
feat(aria): add provider to disable console warnings.

### LINKED ISSUE #3507: Disable aria-label warnings in a video game
I'm working on making a video game using the standalone `node-webkit`/[`nwjs`](https://github.com/nwjs/nw.js) and thought the clean look of Material Design might fit well for the UI. I saw the reasoning #600 was closed but @marcysutton stated:

> I don't think this is a good idea anymore. If it comes up again and we can make a good case for it, we'll reopen.

I was curious if this is a good case for disabling warnings? Apologies if I come across as entitled, I'm just trying to wrap my head around this. For a web site/app, I totally get the need for ARIA. I'd like to make my game(s) accessible but some genres just ... aren't? So, having an accessible menu system/UI on top of an inaccessible game just doesn't make sense to me.

DevVersion:
Fixes #3507.

### DIFF
--- a/src/core/services/aria/aria.js
+++ b/src/core/services/aria/aria.js
@@ -1,11 +1,66 @@
+/**
+ * @ngdoc module
+ * @name material.core.aria
+ * @description
+ * Aria Expectations for ngMaterial components.
+ */
+angular
+  .module('material.core')
+  .provider('$mdAria', MdAriaProvider);
 
-angular.module('material.core')
-  .service('$mdAria', AriaService);
+/**
+ * @ngdoc service
+ * @name $mdAriaProvider
+ * @module material.core.aria
+ *
+ * @description
+ *
+ * Modify options of the `$mdAria` service, which will be used by most of the Angular Material components.
+ **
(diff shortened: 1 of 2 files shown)


src/core/services/aria/aria.js
```diff
@@ -1,11 +1,60 @@
...
+    $get: MdAriaService
+  };
+
+  /**
+   * @ngdoc method
+   * @name $mdAriaProvider#disableWarnings
+   * @param {boolean} disable Whether to show `$mdAria` console warnings. Default value is `false`.
+   */
+  function disableWarnings(disable) {
+    showWarnings = !disable || false;
```
ThomasBurleson: Why do we need to pass a value to a function that is only available during the config phase? Why not `disableWarnings()` ?

DevVersion: @ThomasBurleson I actually added that to be able to reset the provider in the tests (we should not make changes for the tests to pass)

I will try to find a better solution.

PR #8713, opened 2016-06-09
fix(checkbox): update CSS to match spec 

### LINKED ISSUE #9351: checkbox. Release 1.1.0-rc.5. Chrome v.52. Label is not centered
I use your latest Angular Material Release 1.1.0-rc.5.
And your official codepen. 
**Windows 10. Chrome Version 52.0.2743.116 m**

When I resize a window or simply checkbox parent container. Label position is calculated wrong.
It seems like the trouble is in fractional `numbers`. And Chrome has issues with it.
Because if I increase width of checkbox container in 1px alignment is perfect.

**I have no issues with latest Firefox.**

```
md-checkbox.md-checked ._md-icon:after {
    box-sizing: border-box;
    -webkit-transform: rotate(45deg);
    transform: rotate(45deg);
    position: absolute;
    left: 6.67px;
    top: 2.22px;
    display: table;
    width: 6.67px;
    height: 13.33px;
    border-width: 2px;
    border-style: solid;
    border-top: 0;
    border-left: 0;
    content: '';
}
```

capture

clshortfuse:
Checkbox did not support a11y minimum touch target and the box was
larger than spec

 - Change checkbox size from 20px to 18px
 - Keep horizontal size flush when text is blank
 - Add md-dense support for checkbox
 - Add inline alignment math to SCSS
 - Increase size of md-checkbox to properly support touch

Fixes #9351, #9927

### DIFF
--- a/src/components/checkbox/checkbox.scss
+++ b/src/components/checkbox/checkbox.scss
@@ -1,39 +1,53 @@
-//$checkbox-width: 20px !default;
+//$checkbox-width: 18px !default;
 //$checkbox-height: $checkbox-width !default;
 //$checkbox-border-radius: 2px !default;
 //$checkbox-border-width: 2px !default;
 //
 // ^^ defined in variables.scss
 //
-$checkbox-margin: 16px !default;
-$checkbox-text-margin: 10px !default;
-$checkbox-top: 12px !default;
+$checkbox-margin-end: 16px !default;
+$checkbox-text-margin-top: 10px !default;
+$container-checkbox-margin: 3px !default;
+
+$checkbox-min-height: 48px !default;
+$checkbox-min-height-dense: 36px !default;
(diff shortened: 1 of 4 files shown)


clshortfuse:
part or #8209

Splaktar:
My guess is that the Google presubmit tests will break due to finicky screenshot tests. But let's give it a shot as see how it goes.

Splaktar:
I really like the changes here, but unfortunately this failed the presubmit process due to the following “won’t be able to be merged since it breaks a lot of screendiffs (lots of changes in padding)“.

So this is currently on hold until we can determine if it is possible to move forward without requiring all of the screenshot tests within Google to pass.

jelbourn:
Updating: this causes around 80 different test failures for Google apps; it's unlikely we can merge it

PR #8715, opened 2016-06-09
feat(select): allow support for any font-size

clshortfuse:
dropdown arrow font-size is now forced to 16px
md-option now matches md-select font-size


clshortfuse:
part of #8209

PR #8717, opened 2016-06-09
fix(input) match char counter spacing to spec

clshortfuse:
matches official spec

https://material.google.com/components/text-fields.html#text-fields-character-counter


clshortfuse:
part of #8209

DevVersion:
This one seems to be valid :+1:

PR #8723, opened 2016-06-10
fix(dialog): apply foreground color and automatically detect theme

### LINKED ISSUE #8719: Dialogs are not rendered with the correct theme
**Actual behavior**:
-  What is the issue? 
*Dialogs are not respecting the current theme.*
*I would expect dialogs to render according to the current or provided theme.*

-  CodePen Demo which shows your issue : *http://codepen.io/anon/pen/gMPxbL?editors=1010*
-  Details:
*Click on "Change Theme" and pop a dialog.  They always show in the original default theme.  I set .theme('defaultDark') on the confirm dialog and it's half right.  The background is fine, but the text color is wrong.*

**Angular Versions**: *
-  Angular Version: 1.5.5
-  Angular Material Version: 1.1.0-rc5

**Additional Information**:
-  Browser Type: Chrome
-  Browser Version: 51.0.2704.84 m
-  OS: Win
-  Stack Traces: Nope

----

  Add more lines if needed.

DevVersion:
* Apply the missing foreground color on the md-dialog element
* Automatically detect the theme from the `targetEvent` property.
  - Developers are still able to overwrite the theme (as before)
  - Theme will fallback to the default theme (as before)
  - If `targetEvent` is specified, it will use the `mdTheme` controller to lookup for a changed theme

Fixes #8719.

### DIFF
--- a/src/components/dialog/dialog-theme.scss
+++ b/src/components/dialog/dialog-theme.scss
@@ -2,7 +2,10 @@ $dialog-border-radius: 4px !default;
 
 md-dialog.md-THEME_NAME-theme {
   border-radius: $dialog-border-radius;
+
   background-color: '{{background-hue-1}}';
+  color: '{{foreground-1}}';
+
 
   &.md-content-overflow {
     .md-actions, md-dialog-actions {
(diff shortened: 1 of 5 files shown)


ThomasBurleson:
@ErinCoughlan - can you review and test ?

PR #8730, opened 2016-06-10
fix(tooltip): fix regression on touch devices with a mouse

### LINKED ISSUE #8710: tooltips will no longer fire with mouse on touchscreen devices
**Actual behavior**:
If you have a touchscreen device that uses a mouse/touchpad, tooltips will not fire on mouseover. Surface Pro devices, touchscreen Chromebooks and touchscreen Chromebases will not fire without touching the screen.

https://material.angularjs.org/HEAD/demo/tooltip

**Angular Versions**: *
Master

**Additional Information**:
Tested on Chrome

This is because of this commit: https://github.com/angular/material/commit/15402f851e2c7726234670b97fa30eec9626473a

It's not proper to say a device "IS_TOUCH" but rather "HAS_TOUCH". Touchscreen is a feature, not an exclusive identifier. I would actually argue that touch devices rarely have hover support. On Android, you would need to hold down touch for 500ms to trigger longClickPress. Hover events in web browsers Chrome on Android are supported by triggering longClickPress. 

You can actually disable touch events on Chrome and the following events occur after holding down the touchscreen.

* touchstart

(after a long delay)
* mouseover
* mouseenter
* mousemove
* mousedown (right click event)
* focus
* touchcancel (because chrome's context menu appeared)

I then tap outside to close the context menu:

* mouseout
* mouseleave
* blur

As you can see, on Chrome, you don't need to map any touch events and allow Chrome to internally to allow touch operation. 

Alternatively, if `onTouchStart` was triggered, you can ignore any MouseEvents until `onTouchEnd`/`onTouchCancel`. For browser compatibility, we can delay for a few milliseconds to see is `mouseover` fires before `touchstart`, though I'm not sure any browser does that. I do have a touchscreen monitor I can connect to my Windows PC and Mac to test any functionality on other browsers.

crisbeto:
Reverts most of the changes from #8700 and switches to a simpler approach that uses a repeated `touchstart` to hide the tooltip, instead of `touchend`, which fires too early.
Also re-enables mouse events for all kinds of devices, in order to cover cases where the devices has both a touchscreen and a mouse.

Fixes #8710.

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -34,14 +34,11 @@ angular
  * @param {string=} md-direction Which direction would you like the tooltip to go?  Supports left, right, top, and bottom.  Defaults to bottom.
  */
 function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdTheming, $rootElement,
-                            $animate, $q, $interpolate, $mdConstant) {
+                            $animate, $q, $interpolate) {
 
-  // Note that touch devices still fire mouse events, even though they don't have a
-  // mouse. We shouldn't bind them in those cases, because it causes the callbacks
-  // to fire too often.
-  var ENTER_EVENTS = 'focus ' + ($mdConstant.IS_TOUCH ? 'touchstart' : 'mouseenter');
-  var LEAVE_EVENTS = 'blur ' + ($mdConstant.IS_TOUCH ? 'touchend touchcancel' : 'mouseleave');
-  var TOOLTIP_SHOW_DELAY = $mdConstant.IS_TOUCH ? 75 : 0;
+  var ENTER_EVENTS = 'focus touchstart mouseenter';
+  var LEAVE_EVENTS = 'blur touchcancel mouseleave';
+  var TOOLTIP_SHOW_DELAY = 0;
   var TOOLTIP_WINDOW_EDGE_SPACE = 8;
 
   return {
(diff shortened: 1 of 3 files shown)


clshortfuse:
I believe you need to watch for ‘touchend‘ as well in LEAVE_EVENTS or the tooltip will get stuck. You can test this in Chrome with device emulation. Chrome will change your mouse events to touch events.

crisbeto:
I removed `touchend` on purpose, because it fires right after you lift your finger. This isn't particularly good UX, because it'll show the tooltip for a second and then remove it when you lift your finger. Additionally, this caused it to be a bit glitchy, because when the element isn't focused, the tooltip would get stuck, even though you lifted you finger. 

I switched it to open/close when you tap the tooltip target, in addition to the blur and on scroll from before.

clshortfuse:
Okay. Maybe there should be an autoclose timer for touch operations if it doesn't close automatically? It doesn't have to be part of this PR.

crisbeto:
I don't think that's an issue on a touch device, because you have 3 options:

1. You tapped the same target again, in which case it would close.
2. You tapped outside the target, in which case it fires the blur event and closes.
3. You scrolled away, in which case it closes.

clshortfuse:
Okay. Hopefully this doesn't break anything. I'll test it later. My tooltips don't have any delay set and I have plenty of md-icon-button buttons with tooltips. I don't want a tooltip to fire and not close every time a button a pressed.

crisbeto:
Well I retested on a proper touch device (I was mostly testing in the Chrome dev tools) and the blur event isn't quite as predictable. I'll figure something out for tapping out of it.

PR #8734, opened 2016-06-11
feat(list): support md-no-focus class

### LINKED ISSUE #8691: feat/bug: md-button mdNoFocusStyle not working for md-list-item / md-card
**Actual behavior**:
-  What is the issue? * The mdNoFocusStyle can be attach on normal md-buttons but cant be attached on md-list-items aswell as md-cards with an ng-click
The md-list-item aswell as the md-card generate an internal button without the mdNoFocusStyle
-  What is the expected behavior? The mdNoFocusStyle should be attached on md-list-items / md-cards

https://material.angularjs.org/latest/demo/list

Thoughts:
After this would be a bigger task, the question is, if it would be possible/better to add the posibility to disable the mdNoFocusStyle globally for all buttons.

I wrote a simple code for this:
```
if (!angular.isDefined(attr.mdNoFocusStyle) && !$injector.has('$MD_DISABLE_FOCUS_STYLE')) {
  // restrict focus styles to the keyboard
  scope.mouseActive = false;
  element.on('mousedown', function() {
    scope.mouseActive = true;
    $timeout(function(){
      scope.mouseActive = false;
    }, 100);
  })
  .on('focus', function() {
    if (scope.mouseActive === false) {
    }
  })
  .on('blur', function(ev) {
    element.removeClass('md-focused');
  });
}
```
in .config -> `$provide.constant('$MD_DISABLE_FOCUS_STYLE', '');`

Also this would enable a performance improvement for Internet Explorer (specially on mobile versions aswell).
The problem is:
1. $timeout - Triggers a $q.promise chain for angular. (maybe here use `$timeout(func,timeout,false)`)
2. The addClass/removeClass triggers a complete repaint of the page.
3. Because of the md-list-item/md-card: Many event-listeners cant be disabled (specially on ng-repeat scenarios)

**Angular Versions**: *
-  Angular Version: 1.5.5
-  Angular Material Version: RC-5

**Additional Information**:
-  Browser Type: All browsers
-  Browser Version: All versions
-  OS: All OS
-  Stack Traces:

DevVersion:
* Refactors `md-no-focus-style` attribute into the `md-no-focus` class.
* Adds support for `md-no-focus` class forwarding for the list item.
* Add documentation for `md-menu` as proxy item
* Enhance tests for `md-menu` inside of `md-list-item`

Closes #8691.

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -53,8 +53,30 @@ function MdAnchorDirective($mdTheming) {
  * the FAB button background is filled with the accent color [by default]. The primary color palette may be used with
  * the `md-primary` class.
  *
+ * Developers can also change the color palette of the button, by using the following classes
+ * - `md-primary`
+ * - `md-accent`
+ * - `md-warn`
+ *
+ * See for example
+ *
+ * <hljs lang="html">
+ *   <md-button class="md-primary">Primary Button</md-button>
+ * </hljs>
+ *
+ * Button can be also raised, which means that they will use the current color palette to fill the button.
+ *
+ * <hljs lang="html">
+ *   <md-button class="md-accent md-raised">Raised and Accent Button</md-button>
+ * </hljs>
+ *
+ * It is also possible to disable the focus effect on the button, by using the following markup.
+ *
+ * <hljs lang="html">
+ *   <md-button class="md-no-focus">No Focus Style</md-button>
+ * </hljs>
+ *
  * @param {boolean=} md-no-ink If present, disable ripple ink effects.
- * @param {boolean=} md-no-focus-style If present, disable focus style on button
  * @param {expression=} ng-disabled En/Disable based on the expression
  * @param {string=} md-ripple-size Overrides the default ripple size logic. Options: `full`, `partial`, `auto`
  * @param {string=} aria-label Adds alternative text to button for accessibility, useful for icon buttons.
@@ -138,7 +160,7 @@ function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
       }
     });
 
-    if (!angular.isDefined(attr.mdNoFocusStyle)) {
+    if (!element.hasClass('md-no-focus')) {
       // restrict focus styles to the keyboard
       scope.mouseActive = false;
       element.on('mousedown', function() {
@@ -156,6 +178,7 @@ function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
         element.removeClass('md-focused');
       });
     }
+
   }
 
 }
--- a/src/components/list/list.js
+++ b/src/components/list/list.js
@@ -113,6 +113,7 @@ function mdListDirective($mdTheming) {
  * Currently supported proxy items are
  * - `md-checkbox` (Toggle)
  * - `md-switch` (Toggle)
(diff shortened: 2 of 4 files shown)


src/components/list/list.js
```diff
@@ -327,7 +363,9 @@ function mdListItemDirective($mdAria, $mdConstant, $mdUtil, $timeout) {
       function copyAttributes(item, wrapper) {
         var copiedAttrs = $mdUtil.prefixer([
           'ng-if', 'ng-click', 'ng-dblclick', 'aria-label', 'ng-disabled', 'ui-sref',
-          'href', 'ng-href', 'target', 'ng-attr-ui-sref', 'ui-sref-opts'
+          'href', 'ng-href', 'target', 'ng-attr-ui-sref', 'ui-sref-opts',
+          // Custom attributes for the md-button component.
+          'md-no-focus-style'
```
ThomasBurleson: This seems REALLY weird; why not "class="md-no-focus" instead of `md-no-focus-style`?

DevVersion: @ThomasBurleson This is just forwarding the attribute, because it's not supported by the `md-list-item`, but it should be supported for the button executor.

Actually, I agree with you, it would be \*cooler\* to use a class here, but this will be for now more complicated and will require some additional checks, to only forward this specific class to the button executor.

The attribute itself isn't introduced in this PR, it's just supporting it for the button executor (See #8691)

ThomasBurleson:
@DevVersion - I propose we use `class="md-no-focus"` here AND for the mdButton. Let's fix it right for all uses.

DevVersion:
@ThomasBurleson As discussed with @EladBezalel, I've updated the PR :)

src/components/button/button-theme.scss
```diff
@@ -4,7 +4,7 @@
     &:hover {
       background-color: '{{background-500-0.2}}';
     }
-    &.md-focused {
+    &.md-focused:not(.md-no-focus) {
```
ThomasBurleson: In what scenario will we have `class="md-focus md-no-focus"` ?

src/components/button/button.js
```diff
@@ -138,24 +160,22 @@ function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
...
+    scope.mouseActive = false;
+    element.on('mousedown', function() {
+      scope.mouseActive = true;
+      $timeout(function(){
+        scope.mouseActive = false;
+      }, 100);
+    })
+    .on('focus', function() {
+      if (scope.mouseActive === false) {
+        element.addClass('md-focused');
```
ThomasBurleson: No. Only add the `md-focused` class if the `md-no-focus` is **not** present. This will then let the CSS be simpler. 

> Fix the CSS above.

DevVersion: That's what I've had before, just with the attribute.

I'm a bit confused, because you explicitly said _class_. I'll change that

src/components/button/button-theme.scss
DevVersion: This line actually shouldn't mind anymore. The scenario itself will be present, if the button is focused and the user did disable the focus.

The Class won't just match with any selectors because of the `:not`

ThomasBurleson:
@DevVersion  - lgtm. Nice 👍

PR #8736, opened 2016-06-11
fix(datePicker): remove unresolved variable

DevVersion:
* Currently the datePicker scss contains a variable which is not defined.
  This causes issues with the module compilation for closure and default.

The CI can't currently build a new snapshot to `bower-material`, because the variable is undefined.
I guess the variable has been removed, because the icon doesn't have any margin (yet)

image


crisbeto:
That one is actually necessary, otherwise the label isn't aligned properly (http://i.imgur.com/wXJViJb.png). Seems like it's being taken from the button.scss.

DevVersion:
@crisbeto Indeed, the `build-all-modules` task for `default` and `closures` does not allow Sass to import variables from other components, because they should run independently. Please review again.

crisbeto:
Yep, that LGTM.

PR #8738, opened 2016-06-11
fix(toolbar): transitions applied too soon before css loaded

### LINKED ISSUE #7986: bug(theme): md-toolbar flashes with different theme color when page loads

EladBezalel:
- Moved the transitions to a class, had to use nextTick to make sure the app was loaded

fixes #7986

### DIFF
--- a/src/components/toolbar/toolbar.js
+++ b/src/components/toolbar/toolbar.js
@@ -72,6 +72,10 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
       element.addClass('_md');     // private md component indicator for styling
       $mdTheming(element);
 
+      $mdUtil.nextTick(function () {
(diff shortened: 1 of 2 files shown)


src/components/toolbar/toolbar.js
```diff
@@ -72,6 +72,10 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
       element.addClass('_md');     // private md component indicator for styling
       $mdTheming(element);
 
+      $mdUtil.nextTick(function () {
```
crisbeto: I don't think this should trigger a digest, try passing in false as the second argument.

EladBezalel: fixed

crisbeto:
LGTM @ThomasBurleson

PR #8743, opened 2016-06-12
feat(datepicker): is-open attribute, ability to hide icons, demo layout

### LINKED ISSUE #8481: Make datepicker less opinionated
**Actual behavior**:
-  What is the issue? 
md-datepicker forces the calendar icon, and the drop arrow with no way to style it otherwise.  Also, the only way to open the popup is by clicking one of those icons.

-  What is the _desired_ behavior?
make md-datepicker less opinionated.  Allow the possibility to remove the default icons.  Allow a way to open the popup programmatically.  I should be able to open the datepicker popup any way I choose.  Bootstrap has an is-open attribute on the directive to accomplish this, maybe something like that.

crisbeto:
* Adds is-open attribute that allows for the datepicker calendar to be triggered from the outside.
* Adds the ability to disable the datepicker icons.
* Changes the datepicker demo to a 2-column layout, making it more compact.

Fixes #8481.

### DIFF
--- a/src/components/datepicker/js/datepickerDirective.js
+++ b/src/components/datepicker/js/datepickerDirective.js
@@ -27,6 +27,12 @@
    * @param {(function(Date): boolean)=} md-date-filter Function expecting a date and returning a boolean whether it can be selected or not.
    * @param {String=} md-placeholder The date input placeholder value.
    * @param {String=} md-open-on-focus When present, the calendar will be opened when the input is focused.
+   * @param {Boolean=} md-is-open Expression that can be used to open the datepicker's calendar on-demand.
+   * @param {String=} md-hide-icons Determines which datepicker icons should be hidden. Note that this may cause the
+   * datepicker to not align properly with other components. **Use at your own risk.** Possible values are:
+   * * `"all"` - Hides all icons.
+   * * `"calendar"` - Only hides the calendar icon.
+   * * `"triangle"` - Only hides the triangle icon.
    * @param {boolean=} ng-disabled Whether the datepicker is disabled.
    * @param {boolean=} ng-required Whether a value is required for the datepicker.
    *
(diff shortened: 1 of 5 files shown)


src/components/datepicker/js/datepickerDirective.js
```diff
@@ -27,6 +27,12 @@
    * @param {(function(Date): boolean)=} md-date-filter Function expecting a date and returning a boolean whether it can be selected or not.
    * @param {String=} md-placeholder The date input placeholder value.
    * @param {String=} md-open-on-focus When present, the calendar will be opened when the input is focused.
+   * @param {Boolean=} md-is-open Expression that can be used to open the datepicker's calendar on-demand.
+   * @param {String=} md-hide-icons Determines which datepicker icons should be hidden. Note that this may cause the
+   * datepicker to not align properly with other components. **Use at your own risk.** Possible values are:
+   * * `"all"` - Hides all icons.
+   * * `"calendar"` - Only hides the calendar icon.
+   * * `"triangle"` - Only hides the triangle icon.
```
DevVersion: Very nice idea :+1:

src/components/datepicker/js/datepickerDirective.js
```diff
@@ -646,6 +683,10 @@
       self.isCalendarOpen = false;
       self.ngModelCtrl.$setTouched();
 
+      if (self.$attrs.mdIsOpen) {
```
DevVersion: You can probably make the `mdIsOpen` scope variable to optional and just check here for it being defined.

```js
{
  isOpen: '=?mdIsOpen' 
}

angular.isDefined(this.isOpen)
```

crisbeto: Ah that's perfect, I had forgotten about those optional bindings.

PR #8744, opened 2016-06-12
fix(tooltip): properly handle tapping away

crisbeto:
This is a follow-up to #8700 and #8730. It helps with the fact that we can't rely on the focus/blur events in touch devices, in order to hide the tooltip.

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -232,11 +232,19 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
         // Prevent the tooltip from showing when the window is receiving focus.
         if (e.type === 'focus' && elementFocusedOnWindowBlur) {
           elementFocusedOnWindowBlur = false;
-        } else if (e.type === 'touchstart' && scope.visible) {
-          leaveHandler();
         } else if (!scope.visible) {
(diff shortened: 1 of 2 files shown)


crisbeto:
@ThomasBurleson @EladBezalel @topherfangio @DevVersion Can you guys give take a look at this? I'm not sure if I'm overly complicating this, but it seems like the only to properly handle tapping away from a tooltip on touch.

src/components/tooltip/tooltip.js
```diff
@@ -232,11 +232,19 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdThe
...
-        } else if (e.type === 'touchstart' && scope.visible) {
-          leaveHandler();
         } else if (!scope.visible) {
           parent.on(LEAVE_EVENTS, leaveHandler);
           setVisible(true);
+
+          // If the user is on a touch device, we should bind the tap away after
+          // the `touched` in order to prevent the tooltip being removed immediately.
+          if (e.type === 'touchstart') {
+            parent.one('touchend', function() {
```
ThomasBurleson: 👍  🎉

DevVersion: Nice approach. Didn't know that the `one` method is existing

Issue #7987, opened 2016-04-11
bug(input): placeholder text inside toolbar is difficult to read


robertmesserle:
![image](https://cloud.githubusercontent.com/assets/571363/14438514/58ecd2d4-ffdb-11e5-9175-e693af3829d2.png)

HTML:

``` html
<md-toolbar>
  <header class="md-toolbar-tools">
    <div>
      <md-input-container md-no-float>
        <input placeholder="placeholder" />
      </md-input-container>
    </div>
  </header>
</md-toolbar>
```

DevVersion:
Just writing it here. I'm currently looking for a good / the right solution.

The placeholder currently uses the `foreground-1` hue, which is actually for disabled texts and also looking as same as in the specs. (Not chosen by me) 

The foreground palettes are supporting Dark and Light themes, which means that those palettes are not caring about the primary / accent / warn color.

```js
      self.foregroundPalette = self.isDark ? LIGHT_FOREGROUND : DARK_FOREGROUND;
      self.foregroundShadow = self.isDark ? DARK_SHADOW : LIGHT_SHADOW;
```

So, as shown in your screenshot, you're using Indigo, which is likely dark.
But as the foreground palette is not caring about the primary color, it will be still dark and not good to read.

I think we need to find another palette / solution.

EladBezalel:
@DevVersion any update on this?

epelc:
@DevVersion I think the real problem here is that the `forground-*`'s only support light or dark. There needs to be a `foreground-opposite` or something similar for when you are ontop of a background that is opposite of the normal isDark.

Would be nice to get this fixed. Hopefully it'll also fix the same issue with `md-select` in dark toolbars.

clshortfuse:
I have somewhat a solution for this, but I can't find a neat way to get it all working as I want. Right now I was thinking of doing adding support for 

`foreground-1-contrast`
`foregroundLight-1-contrast`
`foregroundDark-2-contrast-0.50`

The problem is, though you can force light and dark, that's not really much of the solution. The reality is that `contrast` and `foreground` are essentially the same thing. 

Right now, `contrast` == `foreground-1` but no ability to select any other levels of `foreground`. I have a solution ready, but I'm just trying to get it all working without breaking other changes. The idea of using `foreground-1` syntax instead of `foreground1` is the root of my issue.

ThomasBurleson:
@EladBezalel  - ping!

EladBezalel:
[...]

Issue #8518, opened 2016-05-20
switch: focus indicator is wrong on dark themes


DevVersion:
Seems like the focus indicator is not correct on dark themes.
- It should be white on dark themes with an opacity of 12%
- It should be black on light themes with an opacity of 12% as well.

FYI: Actually the ripple indicator is correct on dark themes. But not the focus indicator for keyboard interaction.

![image](https://cloud.githubusercontent.com/assets/4987015/15434473/50d64e40-1eb7-11e6-91af-b01c56421c1e.png)

EladBezalel:
Actually the colors of the switch aren't matching the specs in dark/light themes

'' | Light | Dark
--- | --- | ---
Thumb On | Swatch 500, Opacity  100% | Swatch 200, Opacity  100%
Track On | Swatch 500, Opacity  50% | Swatch 200, Opacity  50%
Thumb Off | Grey 50, #FAFAFA, Opacity  100% | Grey 400, #BDBDBD, Opacity  100%
Track Off | #000000, Opacity  38% | #FFFFFF, Opacity  30%
Thumb Disabled | Grey 400, #BDBDBD, Opacity  100% | Grey 800, #424242, Opacity 100%
Track Disabled | #000000, Opacity  12% |  #FFFFFF, Opacity  10%

Splaktar:
@rudzikdawid I haven't had time to look into this one yet, but it seemed like one that you may be interested in investigating. If this is still an issue, I'd like to try to get it into 1.1.11.

PR #8757, opened 2016-06-14
feat(md-theme) improve support for contrast/text colors

clshortfuse:
This deprecates `foreground-*` and instead uses `background-default-contrast-*`. In short:

    newMapping = {
      "foreground-1": "background-default-contrast",
      "foreground-2": "background-default-contrast-secondary",
      "foreground-3": "background-default-contrast-hint",
      "foreground-4": "background-default-contrast-divider"
    }

I also added `contrast-disabled` and `contrast-icon` which are the same as `contrast-hint`.

Now you can create css for items with `primary-contrast-secondary` or others. This should help close issue #7987 

For example, you can now do this:

    md-toolbar.md-THEME_NAME-theme:not(.md-menu-toolbar) md-input-container {
      .md-input {
        @include input-placeholder-color('\'{{primary-contrast-hint}}\'');
        border-color: '{{primary-contrast-divider}}';
      }
    }

### DIFF
--- a/src/components/input/input-theme.scss
+++ b/src/components/input/input-theme.scss
@@ -1,17 +1,17 @@
 md-input-container.md-THEME_NAME-theme {
   .md-input {
-    @include input-placeholder-color('\'{{foreground-3}}\'');
-    color: '{{foreground-1}}';
-    border-color: '{{foreground-4}}';
+    @include input-placeholder-color('\'{{background-default-contrast-hint}}\'');
+    color: '{{background-default-contrast}}';
+    border-color: '{{background-default-contrast-divider}}';
   }
 
   > md-icon {
-    color: '{{foreground-1}}';
+    color: '{{background-default-contrast-icon}}';
   }
 
   label,
   ._md-placeholder {
-    color: '{{foreground-3}}';
+    color: '{{background-default-contrast-hint}}';
   }
 
   &.md-input-invalid {
@@ -21,22 +21,34 @@ md-input-container.md-THEME_NAME-theme {
   }
 
   &:not(.md-input-focused):not(.md-input-invalid) label.md-required:after {
-    color: '{{foreground-2}}';
+    color: '{{background-default-contrast-secondary}}';
   }
 
   .md-input-messages-animation, .md-input-message-animation {
     color: '{{warn-A700}}';
     .md-char-counter {
-      color: '{{foreground-1}}';
+      color: '{{background-default-contrast}}';
     }
   }
 
   &:not(.md-input-invalid) {
     &.md-input-has-value {
       label {
-        color: '{{foreground-2}}';
+        color: '{{background-default-secondary}}';
       }
     }
+  }
+  .md-input {
+    &[disabled],
+    [disabled] & {
+      border-bottom-color: transparent;
+      color: '{{background-default-contrast-disabled}}';
+      background-image: linear-gradient(to right, '{{background-default-contrast-disabled}}' 0%, '{{background-default-contrast-disabled}}' 33%, transparent 0%);
+      background-image: -ms-linear-gradient(left, transparent 0%, '{{background-default-contrast-disabled}}' 100%);
+    }
+  }
+
+  &:not(.md-input-invalid) {
     &.md-input-focused,
     &.md-input-resized {
       .md-input {
@@ -78,15 +90,6 @@ md-input-container.md-THEME_NAME-theme {
     }
     .md-input-message-animation, .md-char-counter {
       color: '{{warn-A700}}';
-    }
-  }
-  .md-input {
-    &[disabled],
-    [disabled] & {
-      border-bottom-color: transparent;
-      color: '{{foreground-3}}';
-      background-image: linear-gradient(to right, '{{foreground-3}}' 0%, '{{foreground-3}}' 33%, transparent 0%);
-      background-image: -ms-linear-gradient(left, transparent 0%, '{{foreground-3}}' 100%);
     }
   }
 }
--- a/src/components/switch/switch-theme.scss
+++ b/src/components/switch/switch-theme.scss
@@ -1,57 +1,43 @@
+@mixin md-switch-checked ($group, $shade) {
+  .md-ink-ripple {
+      color: '{{' + $group + '-'+ $shade +'}}';
+  }
+  ._md-thumb {
+    background-color: '{{' + $group + '-' + $shade + '}}';
+  }
+  ._md-bar {
+    background-color: '{{' + $group + '-' + $shade + '-0.5}}';
+  }
+  &.md-focused ._md-thumb:before {
+    background-color: '{{' + $group + '-' + $shade + '-0.26}}';
+  }
+}
+
 md-switch.md-THEME_NAME-theme {
   .md-ink-ripple {
-    color: '{{background-500}}';
+    color: '{{background-contrast-1.0}}';
   }
   ._md-thumb {
     background-color: '{{background-50}}';
   }
   ._md-bar {
-    background-color: '{{background-500}}';
+    background-color: '{{background-contrast-0.38}}';
+  }
+  &.md-focused ._md-thumb:before {
+    background-color: '{{background-contrast-0.26}}';
   }
 
   &.md-checked {
-    .md-ink-ripple {
-      color: '{{accent-color}}';
-    }
-    ._md-thumb {
-      background-color: '{{accent-color}}';
-    }
-    ._md-bar {
-      background-color: '{{accent-color-0.5}}';
-    }
-    &.md-focused ._md-thumb:before {
-      background-color: '{{accent-color-0.26}}';
-    }
+    @include md-switch-checked('accent', '500');
 
     &.md-primary {
-      .md-ink-ripple {
-        color: '{{primary-color}}';
-      }
-      ._md-thumb {
-        background-color: '{{primary-color}}';
-      }
-      ._md-bar {
-        background-color: '{{primary-color-0.5}}';
-      }
-      &.md-focused ._md-thumb:before {
-        background-color: '{{primary-color-0.26}}';
-      }
+      @include md-switch-checked('primary', '500');
     }
 
     &.md-warn {
-      .md-ink-ripple {
-        color: '{{warn-color}}';
-      }
-      ._md-thumb {
-        background-color: '{{warn-color}}';
-      }
-      ._md-bar {
-        background-color: '{{warn-color-0.5}}';
-      }
-      &.md-focused ._md-thumb:before {
-        background-color: '{{warn-color-0.26}}';
-      }
+      @include md-switch-checked('warn', '500');
     }
+
   }
 
   &[disabled] {
@@ -59,7 +45,38 @@ md-switch.md-THEME_NAME-theme {
       background-color: '{{background-400}}';
     }
     ._md-bar {
-      background-color: '{{foreground-4}}';
+      background-color: '{{background-contrast-0.12}}';
+    }
+  }
+
+  &.md-theme-dark {
+    ._md-thumb {
+      background-color: '{{background-400}}';
+    }
+    ._md-bar {
+      background-color: '{{background-contrast-0.30}}';
+    }
+
+    &.md-checked {
+      @include md-switch-checked('accent', '200');
+
+      &.md-primary {
+        @include md-switch-checked('primary', '200');
+      }
+
+      &.md-warn {
+        @include md-switch-checked('warn', '200');
+      }
+
+    }
+
+    &[disabled] {
+      ._md-thumb {
+        background-color: '{{background-800}}';
+      }
+      ._md-bar {
+        background-color: '{{background-contrast-0.10}}';
+      }
     }
   }
 }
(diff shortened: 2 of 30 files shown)


src/core/services/theming/theming.js
```diff
@@ -145,7 +145,7 @@ var DARK_FOREGROUND = {
 };
 var LIGHT_FOREGROUND = {
   name: 'light',
-  '1': 'rgba(255,255,255,1.0)',
+  '1': 'rgb(255,255,255)',
```
DevVersion: Why using `RGB` here? It's actually the same (I know), but it would keep the consistency.

src/core/services/theming/theming.js
```diff
@@ -154,9 +154,32 @@ var LIGHT_FOREGROUND = {
...
-var DARK_CONTRAST_COLOR = colorToRgbaArray('rgba(0,0,0,0.87)');
-var LIGHT_CONTRAST_COLOR = colorToRgbaArray('rgba(255,255,255,0.87)');
-var STRONG_LIGHT_CONTRAST_COLOR = colorToRgbaArray('rgb(255,255,255)');
+var DARK_CONTRAST_COLORS = {
+  primary: colorToRgbaArray('rgba(0,0,0,0.87)'),
+  secondary: colorToRgbaArray('rgba(0,0,0,0.54)'),
+  disabled: colorToRgbaArray('rgba(0,0,0,0.38)'),
+  hint: colorToRgbaArray('rgba(0,0,0,0.38)'),
+  icon: colorToRgbaArray('rgba(0,0,0,0.38)'),
+  divider: colorToRgbaArray('rgba(0,0,0,0.12)')
```
DevVersion: Actually I think this just makes the theming more complicated as it needs to be.

We should just extend the foreground palettes, by creating those in contrast mode. 

We should be able to do this automatically (recommended) or manually.

DevVersion:
@clshortfuse I think the idea is quite good. But IMO it would be way better to start with something like a Design Doc, which talks about

- What are the Advantages
- Current Implementation
- Proposed / Planned Implementation

cc. @ThomasBurleson

src/core/services/theming/theming.js
clshortfuse: You're right, there's no need for this. Might have been me testing.

clshortfuse:
@DevVersion I'm commenting in the thread so it won't disappear on a code change.

foregroundPalette only allows for one contrast color for one background color type. Per palette, there can be upto three. (Light, StrongLight, and Dark). It would make more sense to use allow themes to override theme.darkContrastColors, theme.lightContrastColors, and theme.strongLightContrastColors. Extra complication is inherit in the change because thr current one is too simple.

From there, previous references to foregroundPalette would be mapped to light/darkContrastColor dependent on isDark.

I didn't want to write all that part yet because I would have to rewrite sanitizePalette to not be global, but that's the idea.

EladBezalel:
@clshortfuse that will probably solve #8518 as well, right?

clshortfuse:
@EladBezalel  After checking it out, it does not fully. That's because THEME_NAME does not have any specification if the theme is light or dark. Swatches need primary-500-1.0 when light and primary-200-1.0 when dark.

I can see mdTheming adding `.md-theme-dark` as a class so you can do special .scss for dark.

Let me see what I can come up with.

ThomasBurleson:
@EladBezalel - please review.

clshortfuse:
@EladBezalel I added `md-theme-dark` and included the fix for md-switch with dark theme as an example.

I'm actually rethinking this now. It might better to include different opacity levels like `primary` `secondary`  `divider` `icon` and deprecate foreground-*. The reality is, the contrast colors are always the same base color, but what differs in level is opacity.

clshortfuse:
Ready for review. I added the md-input and md-switch color fixes.

fixes #7987 
fixes #8518

src/components/switch/switch-theme.scss
```diff
@@ -59,7 +59,72 @@ md-switch.md-THEME_NAME-theme {
...
+          background-color: '{{warn-200}}';
+        }
+        ._md-bar {
+          background-color: '{{warn-200-0.5}}';
+        }
+        &.md-focused ._md-thumb:before {
+          background-color: '{{warn-200-0.26}}';
+        }
+      }
+    }
```
DevVersion: This is is actually a repetition (3x). You should be able to create a local `mixin` and just specify the theme.

clshortfuse:
@DevVersion I updated the scss with the `@mixin` now. It took me a bit longer because I wasn't sure why the focus ripple was the wrong color on a dark theme and then realized it was in `switch.scss`

clshortfuse:
Sorry guys, one more change. This literally came to me in my sleep. The selection of using light or dark contrast opacities were based on theme darkness, which is inaccurate. It should be based on the palette's preference. I added two spec tests that would have failed in the older commit. Previously `primary-500-secondary` would change based on theme darkness. Now it's as it should be.

ThomasBurleson:
@clshortfuse - please squash and rebase. Thx

clshortfuse:
@ThomasBurleson done :)

src/components/input/input-theme.scss
```diff
@@ -37,6 +37,66 @@ md-input-container.md-THEME_NAME-theme {
...
+    [disabled] & {
+      border-bottom-color: transparent;
+      color: '{{foreground-3}}';
+      background-image: linear-gradient(to right, '{{foreground-3}}' 0%, '{{foreground-3}}' 33%, transparent 0%);
+      background-image: -ms-linear-gradient(left, transparent 0%, '{{foreground-3}}' 100%);
+    }
+  }
+}
+
+md-toolbar.md-THEME_NAME-theme:not(.md-menu-toolbar) md-input-container {
```
ThomasBurleson: this should not be in the `input-theme.scss`. It should be in the `toolbar-theme.scss` right ?

clshortfuse: IIRC, I had issues with overrides. Let me try it again.

ThomasBurleson:
@clshortfuse - This looks very comprehensive. Nice work.

src/components/input/input-theme.scss
clshortfuse: Yes, the reason it's in input-theme.scss because overrides. Of the three SCSS collections, the third needs to override the second and the third also shares values.

Quick ref:

Second Collection:

    md-toolbar.md-THEME_NAME-theme:not(.md-menu-toolbar) md-input-container {

Third Collection:

    md-toolbar.md-THEME_NAME-theme:not(.md-menu-toolbar) md-input-container,
    md-input-container.md-THEME_NAME-theme

It's a weird order of overrides, but I feel like perhaps there should be a `@mixin` where the backcolor could change and include md-toolbar in there. In fact, off the top of my head, I can see colored `md-card`s with child md-input-containers having the same issue as when it was in a toolbar.

ThomasBurleson: Agreed. Can you explore a common solution Cards + Toolbars ? 
Thx

clshortfuse: No problem!

clshortfuse:
@ThomasBurleson I've made the changes for `md-toolbar` and `md-card` but this is still `In Progress`. I have a lot of code yet to clean up, and I'm manually checking every element to make sure it works. So far, I've realized an interesting issue where children with autogenerated theme won't change when the parent theme changes, even if md-watch-theme is present.

Also, `md-autocomplete` did not display properly with any dark/colored background. I'll add a comment when I'm done reviewing everything, but the principal changes to `mdTheme` seem complete (albeit messy).

ThomasBurleson:
@EladBezalel - can you help @clshortfuse  with ^^ ?

EladBezalel:
@clshortfuse lets pair on this

src/core/services/theming/theming.js
```diff
@@ -474,10 +504,49 @@ function ThemingProvider($mdColorPalette) {
...
+        var processChild = function(childElement) {
+          var oldThemeName = childElement.data('$mdThemeName');
+          if (oldThemeName) {
+            var attrThemeWatchValue = childElement.attr('md-theme-watch');
+            var watchTheme = (alwaysWatchTheme || angular.isDefined(attrThemeWatchValue)) && attrThemeWatchValue != 'false';
+            if (watchTheme)
+              return; // element watches its own theme
+            if (angular.isDefined(childElement.attr('md-theme')))
+              return; // element has programmatic theme
+            if (oldThemeName) childElement.removeClass('md-' + oldThemeName +'-theme');
```
EladBezalel: i'd break line here as well for clarity

src/core/services/theming/theming.js
```diff
@@ -474,10 +504,49 @@ function ThemingProvider($mdColorPalette) {
       }
 
       /**
+       * Called when theme changes and md-watch-theme is set
+       * Changes theme of all themed children
+       */
+      function onWatchThemeChanged(newValue, oldValue)
```
EladBezalel: Does this function traverse down the dom tree to find mdTheme and apply the changes?

can't we use 'require' of directive and listen programmatically to changes??

src/core/services/theming/theming.js
```diff
@@ -597,6 +679,50 @@ function parseRules(theme, colorType, rules) {
     generatedRules.push(newRule);
   });
 
+  var inverseTypes = ['primary', 'accent', 'warn']
```
EladBezalel: what about background?

src/core/services/theming/theming.js
```diff
@@ -597,6 +679,50 @@ function parseRules(theme, colorType, rules) {
...
+        var selectors = [];
+        var colorTypeClass = '.md-' + colorType;
+        var hueNameClass = hueName === 'default' ? '' : '.md-' + hueName;
+        angular.forEach(regexMatch[1].split(/,/g), function(selector) {
+          selectors.push('._md-theme-fill' + colorTypeClass + hueNameClass + ' ' + selector.trim());
+        });
+        relatedRules += selectors.join(',') + ' {' + regexMatch[2] + '}';
+      }
+      var allStylesRegex = /'?"?\{\{\s*([a-zA-Z]+)-?(A?\d+|hue\-[0-3]|shadow|default|color)?-?(contrast)?-?((?:\d\.?\d*)|(?:[a-zA-Z]+))?\s*\}\}'?"?/g;
+      var r = relatedRules.replace(allStylesRegex, function(match, regexColorType, hue, contrast, opacity) {
```
EladBezalel: r?

src/core/services/theming/theming.js
```diff
@@ -597,6 +679,50 @@ function parseRules(theme, colorType, rules) {
...
+      var regexMatch;
+      var relatedRules = '';
+      while ((regexMatch = parseCSSRulesRegex.exec(rules)) !== null) {
+        if (regexMatch.index === parseCSSRulesRegex.lastIndex) {
+          regexMatch.lastIndex++;
+        }
+        var selectors = [];
+        var colorTypeClass = '.md-' + colorType;
+        var hueNameClass = hueName === 'default' ? '' : '.md-' + hueName;
+        angular.forEach(regexMatch[1].split(/,/g), function(selector) {
```
EladBezalel: forEach within while within forEach within forEach

This must be suuuper expensive, we should come up with better idea how to do it

src/core/services/theming/theming.js
```diff
@@ -722,6 +849,13 @@ function generateAllThemes($injector, $mdTheming) {
           }
         }
       }
+      function getOpacityValues() {
+        if (defaultContrast === 'light') {
+          return darkColors.indexOf(hueName) > -1 ? DARK_CONTRAST_OPACITY : LIGHT_CONTRAST_OPACITY;
+        } else {
```
EladBezalel: redundant else

src/core/services/theming/theming.js
ThomasBurleson: `var rules`

src/core/style/mixins.scss
```diff
@@ -284,3 +284,18 @@
     }
   }
 }
+
+@mixin md-theme-fill-support($hueColor: 'color') {
```
EladBezalel: `md-theme-fill-support` doesn't sounds like a proper name

clshortfuse:
@EladBezalel I cleaned up the code in `theming.js` I can confirm it's working as it should with a few quirks. A quick way to test is to add `md-warn md-hue-3` to `md-content` element in the `button` demo. The only issue I have now is failed spec tests. I'm still trying to narrow it down. I'll work on the other fixes now.

src/core/services/theming/theming.js
clshortfuse: The logic is to swap references the `background` palette with these types.

EladBezalel:
@clshortfuse thank you very much, really appreciate your help

ThomasBurleson:
@clshortfuse - nice work so far!

clshortfuse:
Looks good now, just needs review. I'm just a bit worried about performance implications about `md-theme-fill` with too many rules. But essentially, the alternative is to create 3 new themes, each with a new background based on `primary`, `accent`, `warn`.

clshortfuse:
Closing since I'm breaking this up into smaller PRs

PR #8763, opened 2016-06-14
fix(select): ngModel validator should use option hashkeys.

### LINKED ISSUE #8666: mdSelect: validation bug
**Actual behavior**:
-  What is the issue? *
md-select shows validation error (required) even when a value is selected

-  CodePen Demo which shows your issue : http://codepen.io/anon/pen/PzPwZO

EDIT: Here is a revised one:
http://codepen.io/anon/pen/beVdRb
Click the select and then click away without changing the value.

-  Details: The demo works with rc4

**Angular Versions**: *
-  Angular Version: 1.5
-  Angular Material Version: 1.10-rc.5

**Additional Information**:
-  Browser Type: Chrome
-  Browser Version: 51.0.2704.79
-  OS: Fedora 23
-  Stack Traces:

DevVersion:
* Currently when using `required` on the select, is adding the required validator to the ngModelCtrl.
  The required validator is checking the `$viewValue`by using the `$isEmpty` function, which is overwritten
  by the select component.

  `$isEmpty` normally checks just for truthy values, but our custom `$isEmpty` function checks the value for being a registered option.
  This does not work, because we are just trying to find the value in the options object.
  Once the the value is also an object, we are not able to find the option (which is wrong). All options inside of the variable are stored within their
  associated hashkey.

  That's why we have to retrieve the hashkey of the current value. This is also required for ngModel's with a `trackBy` option.

* Once an option is set to the ngModel, before the actual `md-option` directive is registered, the required validator will be set to `false`.
  We have to trigger a new validation, when the option, which is currently the $modelValue, has been registered.

Fixes #8666.

Please review @topherfangio @crisbeto

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -685,7 +685,9 @@ function SelectMenuDirective($parse, $mdUtil, $mdTheming) {
 
       // Setup a more robust version of isEmpty to ensure value is a valid option
       self.ngModel.$isEmpty = function($viewValue) {
-        return !self.options[$viewValue];
+        // We have to transform the viewValue into the hashKey, because otherwise the
+        // OptionCtrl may not exist. Developers may have specified a trackBy function.
+        return !self.options[self.hashGetter($viewValue)];
       };
 
       // Allow users to provide `ng-model="foo" ng-model-options="{trackBy: 'foo.id'}"` so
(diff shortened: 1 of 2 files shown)


crisbeto:
LGTM

PR #8766, opened 2016-06-14
fix(menu): clean up the backdrop if the menu got destroyed mid-animation

### LINKED ISSUE #8727: (md-menu) cursor is trapped if md-menu is destroyed too early
**Actual behavior**:
If md-menu is destroyed too quickly after a menu is fired, the cursor focus is lost completely.

http://codepen.io/shortfuse/pen/XKXPvJ
At about 300-400 timeout the cursor gets trapped

**Angular Versions**: *
master

**Additional Information**:
Chrome v50/v51

crisbeto:
The menu backdrop wasn't being cleaned up, if the menu got destroyed while it was animating.

Fixes #8727.

### DIFF
--- a/src/components/menu/js/menuServiceProvider.js
+++ b/src/components/menu/js/menuServiceProvider.js
@@ -74,7 +74,7 @@ function MenuProvider($$interimElementProvider) {
      * and backdrop
      */
     function onRemove(scope, element, opts) {
-      opts.cleanupInteraction();
+      opts.cleanupInteraction && opts.cleanupInteraction();
       opts.cleanupResizing();
       opts.hideBackdrop();
 
@@ -437,7 +437,7 @@ function MenuProvider($$interimElementProvider) {
       }
 
       var rtl = ($mdUtil.bidi() == 'rtl');
-      
+
       switch (positionMode.left) {
         case 'target':
           position.left = existingOffsets.left + originNodeRect.left - alignTargetRect.left;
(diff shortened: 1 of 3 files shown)


crisbeto:
@DevVersion Could you give me some feedback on this since you were dealing with the interim elements?

src/components/menu/js/menuServiceProvider.js
```diff
@@ -60,13 +60,19 @@ function MenuProvider($$interimElementProvider) {
         $animate.enter(options.backdrop, $document[0].body);
       }
 
+      // Handles cases where the menu was destroyed, while it was animating.
+      var destroyListener = scope.$on('$destroy', hideBackdrop);
```
DevVersion: It is quite unnormal to have a destroy listener inside of the interims `onShow` method.

DevVersion:
@crisbeto Having a `destroy` listener here is not the correct solution.

Once the interim Element will be destroyed, the `onRemove` event will be called as well.
Also with the option `options.$destroy` (you don't need it for this case)

At #8727, an error will be thrown inside of a promise, which leads to a cancellation of the `onRemove` method.

The error will be thrown, because the `cleanupInteraction` method is not defined before the animation completes.

The fix would be
```js
    function onRemove(scope, element, opts) {
      opts.cleanupInteraction && opts.cleanupInteraction();
```

crisbeto:
Thanks, I'll give it another shot. If I remember correctly, the onRemove wasn't firing yesterday.

DevVersion:
Sounds good! Yeah it will be triggered, I guess you checked the execution after the Runtime Exception.

crisbeto:
That null check for the `cleanupInteraction` worked fine for the codepen from #8727 @DevVersion, but it didn't help when destroying the scope programmatically in the unit test. I also had to call the remove method in the directive's destroy listener.

DevVersion:
@crisbeto Yeah, actually you're using the fix in your current version, because you're triggering the interimElements `destroy` function, which will come to the `cleanupInteraction`.

This is definitely the perfect solution. LGTM!

PR #8775, opened 2016-06-15
feat(md-autocomplete) make clear and blur on escape keydown optional

clshortfuse:
Adds to the ability to choose what happens when escape is pressed. This PR allows consistency with standard dropdown autocomplete boxes. 

For example, on Chrome, the default autocomplete browser behavior is to simply close the dropdown, not clear the text and not blur.

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -320,6 +320,67 @@ describe('<md-autocomplete>', function() {
       element.remove();
     }));
 
+    describe('md-escape-options checks', function() {
+      var scope, ctrl, element;
+      var template = '\
+              <md-autocomplete\
+                  md-escape-options="{{escapeOptions}}"\
+                  md-search-text="searchText"\
+                  md-items="item in match(searchText)"\
+                  md-item-text="item.display"\
+                  placeholder="placeholder">\
+                <span md-highlight-text="searchText">{{item.display}}</span>\
+              </md-autocomplete>';
+      beforeEach( inject(function($timeout) {
+        scope = createScope();
+        element = compile(template, scope);
+        ctrl = element.controller('mdAutocomplete');
+
+        // Focus the input
+        ctrl.focus();
+        $timeout.flush();
+        expect(scope.searchText).toBe('');
+
+        scope.$apply('searchText = "test"');
+
+        expect(scope.searchText).toBe('test');
+
+        $timeout.flush();
+      }));
+
+      afterEach(function() { element.remove() });
+      it('does not clear the value nor blur when hitting escape', inject(function($mdConstant, $document) {
+        scope.$apply('escapeOptions = "none"');
+        scope.$apply(function() {
+          ctrl.keydown(keydownEvent($mdConstant.KEY_CODE.ESCAPE));
+        });
+
+        expect(scope.searchText).toBe('test');
+        expect($document.activeElement).toBe(ctrl[0]);
+      }));
+
+      it('does not clear the value but does blur when hitting escape', inject(function($mdConstant, $document) {
+        scope.$apply('escapeOptions = "blur"');
+        scope.$apply(function() {
+          ctrl.keydown(keydownEvent($mdConstant.KEY_CODE.ESCAPE));
+        });
+
+        expect(scope.searchText).toBe('test');
+        expect($document.activeElement).toBe(undefined);
+      }));
+
+      it('clear the value but does not blur when hitting escape', inject(function($mdConstant, $document) {
+        scope.$apply('escapeOptions = "clear"');
+        scope.$apply(function() {
+          ctrl.keydown(keydownEvent($mdConstant.KEY_CODE.ESCAPE));
+        });
+
+        expect(scope.searchText).toBe('');
+        expect($document.activeElement).toBe(ctrl[0]);
+      }));
+
+    });
+
     it('should not show the progressbar when hitting escape on an empty input', inject(function($mdConstant, $timeout) {
       var scope = createScope();
       var template = '\
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -466,12 +466,22 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         select(ctrl.index);
         break;
       case $mdConstant.KEY_CODE.ESCAPE:
+        if (!shouldProcessEscape()) return;
         event.stopPropagation();
         event.preventDefault();
-        if ($scope.searchText) clearValue();
 
-        // Force the component to blur if they hit escape
-        doBlur(true);
+        clearSelectedItem();
+        if ($scope.searchText && hasEscapeOption('clear')) {
+          clearSearchText();
+        }
+
+        if (hasEscapeOption('blur')) {
+          // Force the component to blur if they hit escape
+          doBlur(true);
+        } else {
+          // Manually hide (needed for mdNotFound support)
+          ctrl.hidden = true;
+        }
 
         break;
       default:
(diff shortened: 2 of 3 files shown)


clshortfuse:
fixes:

#5196
#6534
#6581

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -466,12 +466,20 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
 
-        // Force the component to blur if they hit escape
-        doBlur(true);
+        clearValue(shouldClearOnEscape() && $scope.searchText);
+
+        if (shouldBlurOnEscape()) {
+          // Force the component to blur if they hit escape
+          doBlur(true);
+        }
+        else {
```
DevVersion: Nitpicky: This `else` statement should be next to the brace (consistency) to follow the Google Code Style Conventions

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -639,7 +679,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   /**
    * Clears the searchText value and selected item.
    */
-  function clearValue ($event) {
+  function clearValue (shouldClearText) {
```
DevVersion: We should remove any parameters here, because those weren't used before. 

Actually this method should just clear the autocompletes value and not validate anything.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -552,6 +560,38 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+    return false;
+  }
+
+  /**
+   * Determines if the escape key should clear input
+   * @returns {boolean}
+   */
+  function shouldClearOnEscape() {
+    if (!$scope.escapeOptions) return true;                      // default true
+    if ($scope.escapeOptions.toLowerCase().indexOf('clear') >= 0) return true; // blur is requested
```
DevVersion: Can be simplified

```js
return !$scopeOptions || $scope.escapeOptions.toLowerCase().indexOf('clear') !== -1; 
```

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -552,6 +560,38 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+    return false;
+  }
+
+  /**
+   * Determines if the escape key should blur focus
+   * @returns {boolean}
+   */
+  function shouldBlurOnEscape() {
+    if (!$scope.escapeOptions) return true;                     // default true
+    if ($scope.escapeOptions.toLowerCase().indexOf('blur') >= 0) return true; // blur is requested
```
DevVersion: This can be also simplified

```js
return !$scopeOptions || $scope.escapeOptions.toLowerCase().indexOf('blur') !== -1; 
```

src/components/autocomplete/js/autocompleteController.js
clshortfuse: oops, that was me rebasing and didn't noticing the `$event` was added, should I append the variable instead?

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -552,6 +560,38 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
   /**
+   * Determines if the escape keydown should be processed
+   * @returns {boolean}
+   */
+  function shouldProcessEscape() {
+    if (shouldBlurOnEscape()) return true;                       // Needs blur
+    if (!ctrl.hidden) return true;                               // Hide dropdown
+    if (ctrl.loading) return true;                               // Cancel loading
+    if (shouldClearOnEscape() && $scope.searchText) return true; // Clear Text
+    return false;
```
DevVersion:
```js
return shouldBlurOnEscape() || !ctrl.hidden || ctrl.loading || shouldClearOnEscape() && $scope.searchText;
```

src/components/autocomplete/js/autocompleteController.js
DevVersion: This method, should accept no parameter. It should just clear the value.

src/components/autocomplete/js/autocompleteController.js
clshortfuse: I was copying `shouldHide`'s format. Should I still simplify it?

src/components/autocomplete/js/autocompleteController.js
clshortfuse: Maybe I wasn't clear. That's the problem though. `clearValue` will always clear the selected item and the text. I'm trying to allow text clearing to be optional.

clshortfuse:
@DevVersion i made the changes and I added a spec test as well.

src/components/autocomplete/js/autocompleteController.js
DevVersion: Agreed, I can see the point. Thanks for the clarification - Probably rename the boolean to
```js
function clearValue(clearText) {}
```

src/components/autocomplete/js/autocompleteController.js
DevVersion: Yeah, this is definitely better here.

src/components/autocomplete/js/autocompleteController.js
DevVersion: Ah, your new approach, looks even more elegant!

clshortfuse: I made a change that seems to not break anything. I separated `clearSearchText` and `clearSelectedItem`. This allows me to call clearSearchText selectively.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -466,12 +466,21 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
       case $mdConstant.KEY_CODE.ESCAPE:
+        if (!shouldProcessEscape()) return;
         event.stopPropagation();
         event.preventDefault();
-        if ($scope.searchText) clearValue();
 
-        // Force the component to blur if they hit escape
-        doBlur(true);
+        clearSelectedItem();
+        if (shouldClearOnEscape() && $scope.searchText)
```
DevVersion: Please add braces here.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -552,6 +561,30 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+    return !$scope.escapeOptions || $scope.escapeOptions.toLowerCase().indexOf('blur') !== -1;
+  }
+
+  /**
+   * Determines if the escape key should clear input
+   * @returns {boolean}
+   */
+  function shouldClearOnEscape() {
+    return !$scope.escapeOptions || $scope.escapeOptions.toLowerCase().indexOf('clear') !== -1;
+  }
```
DevVersion: Sorry for the nitpicky comments.

I'm just thinking that having those two functions is quite unnecessary.

```js
function hasEscapeOption(option) {
  return !$scope.escapeOptions || $scope.escapeOptions.toLowerCase().indexOf(option) !== -1;
}

// Then you can check the options like this
hasEscapeOption('clear');
```

DevVersion:
@clshortfuse Thanks for the quick changes. The PR now looks pretty good.

Please fix the two missing comments I've added and probably add some tests for the other two options.

clshortfuse:
@DevVersion made the changes and added spec tests

src/components/autocomplete/autocomplete.spec.js
```diff
@@ -320,6 +320,111 @@ describe('<md-autocomplete>', function() {
...
+      var scope = createScope();
+      var template = '\
+          <md-autocomplete\
+              md-escape-options="clear"\
+              md-search-text="searchText"\
+              md-items="item in match(searchText)"\
+              md-item-text="item.display"\
+              placeholder="placeholder">\
+            <span md-highlight-text="searchText">{{item.display}}</span>\
+          </md-autocomplete>';
```
crisbeto: These templates seem pretty repetitive, can you try to do them in the before each or compile them in a function?

clshortfuse:
@DevVersion Okay. I've never used `beforeEach` before, but I think I got it. Let me know if there's more I can do to clean it up.

PR #8786, opened 2016-06-17
fix(card): Image selector should only select immediate children or im…

### LINKED ISSUE #8785: mdCard: img selector should be more specific to avoid breaking custom layouts
**Actual behavior**:
-  What is the issue? * https://github.com/angular/material/commit/df902381 seriously breaks any custom layouts using cards and images. The selector should be changed and possibly remove `!important` but more so the selector changed.
-  What is the expected behavior? The selector should target `md-card-header img` or add a class(maybe `.md-img` to allow for custom layouts. ping @ThomasBurleson 

-  CodePen Demo which shows your issue :
-  Details:

**Angular Versions**: *
-  Angular Version: 1.5.6
-  Angular Material Version: 1.1.0-rc.5

**Additional Information**:
-  Browser Type: * chrome
-  Browser Version: *51.0.2704.84 (64-bit)
-  OS: * os x 10.11.5
-  Stack Traces:

----

  Add more lines if needed.

epelc:
…ages within `md-card-header`

Closes #8785

This allows you to have completely custom card layouts without having the `img` selectors interfere. It should also make it safer to update these selectors as it will no longer break custom layout(only ones uses specific material layouts).

Examples of custom layouts. 

- A table within a card with some fields being images
- A profile type card with your picture in it(think google+ style)
- Any other non standard layout

Note I'm aware you might say use a div with `.md-whiteframe-z1` like the docs. But that means you can't use any other parts that you get for free from cards. You'd also probably end up with something like `<my-custom-card>` as a wrapper which is confusing when it is simple to make `md-card` support more custom layouts.

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -60,7 +60,7 @@ md-card {
   }
 
   > img,
-  > :not(md-card-content) img {
+  > md-card-header img {
     box-sizing: border-box;
     display: flex;
     flex: 0 0 auto;


EladBezalel:
lgtm

PR #8793, opened 2016-06-18
fix(mdAria): apply aria-label to buttons correctly

### LINKED ISSUE #8789: mdButton adds aria-label when element already has text content
**Actual behavior**:
-  What is the issue? *

mdButton is erroneously adding an `aria-label` attribute when it shouldn't because there's already text in the element. Text in a button element (custom or native, doesn't matter) is announced just fine on focus. 

An `aria-label` is not added when the button already has text, and it doesn't warn for a missing attribute.

The problem is in $mdAria, either in the `expectWithText` method or `expect` method called internally. The `expectWithText` method seems more appropriate for `md-radio-button`, which DOES need an `aria-label` even when there is child text content (required to expose an accessible name). The button expect version needs the same interpolation support, though.

Visible on the button demos on https://material.angularjs.org
Inspect the buttons with text content, observe the extra `aria-label`

**Angular Versions**: *
Current

**Additional Information**:
-  Browser Type: * 
All
-  Browser Version: *
All
-  OS: *
All

----

  Add more lines if needed

marcysutton:
I added a method to check for text content before copying or expecting an `aria-label`. It's essentially the inverse of the `expectWithText` method needed for custom radio buttons and checkboxes, which require an `aria-label` to expose an accessible name, even when there's text content. Buttons don't have that same requirement so they shouldn't have text content copied to the `aria-label` attribute. They only need it when there's no text content at all.

Closes #8789

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -142,7 +142,7 @@ function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
     $mdButtonInkRipple.attach(scope, element);
 
     // Use async expect to support possible bindings in the button label
-    $mdAria.expectWithText(element, 'aria-label');
+    $mdAria.expectWithoutText(element, 'aria-label');
 
     // For anchor elements, we have to set tabindex manually when the
     // element is disabled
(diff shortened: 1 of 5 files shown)


ThomasBurleson:
@marcysutton  - can you update the `tooltip.spec.js` failures ?

marcysutton:
Fixed! That was an other instance of unnecessarily adding `aria-label`.

ThomasBurleson:
@marcysutton  - thx. lgtm

PR #8797, opened 2016-06-18
perf(icon): better handling of large icon sets

### LINKED ISSUE #8687: Performance issue with md-icon and md-svg-icon
**Actual behavior**:
It is very slow when i use `md-svg-icon`.

I load the `mdi.svg` from https://materialdesignicons.com with a`$mdIconProvider`, but the same issue exists with other (smaller) iconsets.
Pre-fetching the icons did also not help.

**CodePen**
http://codepen.io/tburgherr/pen/BzodNm

Large delay when adding SVG icons, oke for font-icons and avatar images.

**Angular Versions**: 
-  Angular Version: 1.5.3
-  Angular Material Version: 1.0.9

**Additional Information**:
-  Browser Type: Chrome
-  Browser Version: 47.0.2526.106 (64-bit)
-  OS: Linux

------------------------------------------------------------

Is there something I do wrong, or should I use font icons if many icons are needed?

crisbeto:
When loading SVG icons, the icon service was re-parsing the entire response for each icon. It caused massive performance issues with large SVG icon sets.  This change caches the SVG when it gets parsed for the first time and then returns the cached object on subsequent calls.

Fixes #8687.

### DIFF
--- a/src/components/icon/js/iconService.js
+++ b/src/components/icon/js/iconService.js
@@ -341,7 +341,7 @@ MdIconProvider.prototype = {
     return this;
   },
 
-  $get: ['$templateRequest', '$q', '$log', '$templateCache', '$mdUtil', '$sce', function($templateRequest, $q, $log, $templateCache, $mdUtil, $sce) {
+  $get: ['$templateRequest', '$q', '$log', '$mdUtil', '$sce', function($templateRequest, $q, $log, $mdUtil, $sce) {
     return MdIconService(config, $templateRequest, $q, $log, $mdUtil, $sce);
   }]
 };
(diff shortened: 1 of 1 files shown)


src/components/icon/js/iconService.js
```diff
@@ -547,8 +548,10 @@ function MdIconService(config, $templateRequest, $q, $log, $mdUtil, $sce) {
             reject(err);
           },
           extractSvg = function(response) {
-            var svg = angular.element('<div>').append(response).find('svg')[0];
-            resolve(svg);
+            if (!svgCache[url]) {
```
ThomasBurleson: 👍

Issue #8815, opened 2016-06-21
ripple shows strangely after chrome updated


cyx8808:
After chrome stable channel updated to version 51.0.2704.103, the ripple shows strangely.

JingkaiZhao:
same issue happens with me.
the ripple animation duration is too short (looks like blink) under latest chrome version!

DevVersion:
@ThomasBurleson - @crisbeto and I will probably take a look at this today.

EladBezalel:
I'll try to take a look :)

EladBezalel:
@DevVersion seems like somehow `md-ripple-active` class is being removed early, couldn't figure why `fadeInComplete` is called too early..

cyx8808:
I think the problem is not fadeInComplete is called to early. 

I found that Chrome changed the rule of showing transition. When a ripple is active, we change the opacity from 0 to .2 with a swift-ease-out-timing-function during 0.9s. Let's say after 0.5s mouseup triggered and the opacity is .1 at the moment, then we remove "opacity: .2". Before version 51 it still takes 0.9s to change opacity from .1 to 0, but in the last version it only takes .5s to change opacity from .1 to 0 (I'm not sure if the time is .5s, but I think it's the same as the time it takes to play animation before), it's much shorter than 0.9s, so it seems like the ripple disappear suddenly.

I don't know whether Chrome will fix this in the future, I sent a pull request to fix this issue at present.

crisbeto:
@cyx8808 I believe this will change the way the ripple behaves on other browsers.

cyx8808:
@crisbeto Yes, but I think the difference can hardly perceive, you can test. Or maybe there are better ways.

ThomasBurleson:
Fixed with SHA 15da974d51795eb5e0919946ae165a4093a74799

PR #8824, opened 2016-06-22
fix issue causing ripple shows strangely in Chrome version 51

cyx8808:
to fix issue [#8815](https://github.com/angular/material/issues/8815)

### DIFF
--- a/src/core/services/ripple/ripple.js
+++ b/src/core/services/ripple/ripple.js
@@ -432,6 +432,7 @@ InkRippleCtrl.prototype.removeRipple = function (ripple) {
   if (index < 0) return;
   this.ripples.splice(this.ripples.indexOf(ripple), 1);
   ripple.removeClass('md-ripple-active');
+  ripple.addClass('md-ripple-remove');
   if (this.ripples.length === 0) this.container.css({ backgroundColor: '' });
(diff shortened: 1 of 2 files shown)


src/core/services/ripple/ripple.js
```diff
@@ -432,6 +432,7 @@ InkRippleCtrl.prototype.removeRipple = function (ripple) {
   if (index < 0) return;
   this.ripples.splice(this.ripples.indexOf(ripple), 1);
   ripple.removeClass('md-ripple-active');
+  ripple.addClass('md-ripple-remove');
```
crisbeto: Shouldn't this class be removed at some point? Otherwise it will always stick around after the first click.

cyx8808:
I signed it!

Date: Wed, 22 Jun 2016 06:04:49 -0700

src/core/services/ripple/ripple.js
cyx8808: the ripple itselt will be removed soon after

ThomasBurleson:
@crisbeto - can you validate this?

crisbeto:
Tested it out and LGTM @ThomasBurleson, it seems to help with a similar issue with Edge as well.

PR #8842, opened 2016-06-23
fix(site): Fix navigation toggle to properly open on page refresh.

### LINKED ISSUE #8841: Docs Site: Navigation menu doesn't open properly when refreshing.
**Actual behavior**:
-  What is the issue? *When refreshing the docs page, the navigation menu does not properly toggle the open group; forcing you to click on it twice.
-  What is the expected behavior? It should automatically open the group and scroll the selected item into view.

-  CodePen Demo which shows your issue : https://material.angularjs.org/HEAD/demo/icon
-  Details: Notice how if you just click the link; you cannot access the other demos without clicking on the "Demos" nav item twice (once to close; and then once to re-open).

**Angular Versions**: *
-  Angular Version: N/A
-  Angular Material Version: All

**Additional Information**:
-  Browser Type: *
-  Browser Version: *
-  OS: *
-  Stack Traces:

topherfangio:
Currently, the navigation groups on the left would have their icon appear to be toggled open, but the actual content would be invisible because the element's height was not being properly set.

Fix by setting the height in a `$mdUtil.nextTick()` so that it fires at the appropriate time (and can measure the height).

Additionally, add some code to scroll the selected item into view so that users know which page they are visiting.

Fixes #8841.

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -19,7 +19,7 @@ angular
 /**
  * @ngInject
  */
-function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $interpolate, $log, $rootElement, $window) {
+function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $interpolate, $log, $rootElement, $window, $$rAF) {
   // Setup some core variables for the processTemplate method
   var startSymbol = $interpolate.startSymbol(),
     endSymbol = $interpolate.endSymbol(),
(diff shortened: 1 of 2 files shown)


topherfangio:
@crisbeto @EladBezalel @DevVersion Can you guys run this locally and tell me if you have any issues? It was bugging the crap out of me while I'm testing some other stuff, so I took a bit to fix it :-)

DevVersion:
@topherfangio Just run it locally and it seems to work pretty nicely.

Also the animated scroll to the current `menu-item` is working great.

crisbeto:
I'll give it a try later, but that seems like a lot of timeouts. Did you try with rAF instead?

topherfangio:
@crisbeto I have not tried with rAF yet. I could give that a go and see if I can make it smoother.

crisbeto:
Worked pretty good on all browsers @topherfangio. An improvement could be to have some nicer easing when scrolling down. [You can check the function that we use on progressCircular.](https://github.com/angular/material/blob/master/src/components/progressCircular/js/progressCircularProvider.js#L72) It also has a link where you can generate your own easing functions.

topherfangio:
@crisbeto Thanks for testing it! I'll checkout that easing function and see if I can make things a bit smoother.

topherfangio:
@crisbeto Updated to use the easing function. Think you could give it another quick test?

As a note: I rebased/force-pushed, so you may need to delete/re-checkout the branch.

src/core/util/util.js
```diff
@@ -750,6 +750,51 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
+        element.scrollTop = newPosition;
+        
+        if (scrollingDown ? newPosition < scrollEnd : newPosition > scrollEnd) {
+          $$rAF(scrollChunk);
+        }
+      }
+      
+      function calculateNewPosition() {
+        var duration = 1000;
+        var currentTime = Date.now() - startTime;
```
crisbeto: There's `$mdUtil.now` for getting the timestamp.

topherfangio:
@ThomasBurleson Fixed the few issues @crisbeto mentioned. Ready for your review.

PR #8848, opened 2016-06-24
fix(panel): destroy the scope when the panelRef is destroyed

### LINKED ISSUE #8845: mdPanel doesn't destroy controller
**Actual behavior**:
-  What is the issue? mdPanel doesn't destroy controller
-  What is the expected behavior? mdPanel destroy controller after close

-  CodePen Demo which shows your issue :  http://codepen.io/anon/pen/ezgzEO?editors=0010 (it's official demo modified https://material.angularjs.org/latest/demo/panel)
-  Details:  Just click several times to "dialog" button (or "select menu") and look to console where will displayed $scope.$id. On every click displayed $scope.$id's will grow

**Angular Versions**: *
-  Angular Version: 1.5.5
-  Angular Material Version: 1.1.0-rc.5

**Additional Information**:
-  Browser Type: Chromium
-  Browser Version: 50.0.2661.102 
-  OS: Ubuntu 16.04 (64-bit)
-  Stack Traces:

----

  Add more lines if needed.

crisbeto:
The panel wasn't destroying its' scope within the `destroy` method, causing potential memory leaks.

Fixes #8845.

CC @ErinCoughlan

### DIFF
--- a/src/components/panel/demoBasicUsage/script.js
+++ b/src/components/panel/demoBasicUsage/script.js
@@ -92,8 +92,11 @@ function PanelDialogCtrl(mdPanelRef) {
 
 
 PanelDialogCtrl.prototype.closeDialog = function() {
-  this._mdPanelRef && this._mdPanelRef.close().then(function() {
+  var panelRef = this._mdPanelRef;
+
+  panelRef && panelRef.close().then(function() {
     angular.element(document.querySelector('.demo-dialog-open-button')).focus();
+    panelRef.destroy();
   });
 };
 
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -245,7 +245,9 @@ angular
  * @ngdoc method
  * @name MdPanelRef#close
  * @description
- * Hides and detaches the panel.
+ * Hides and detaches the panel. Note that this will **not** destroy the panel. If you
+ * don't intend on using the panel again, call the {@link MdPanelRef#destroy destroy} method
+ * afterwards.
  *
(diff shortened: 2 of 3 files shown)


ErinCoughlan:
LGTM Thanks @crisbeto

clshortfuse:
This could be an issue if you actually want to close the panel and not destroy it. I'm in the process of porting my custom component to panel. In the interest of reusing and precaching components, I would like a panel to close, but not destroy, and then call it to open again.

But I do think `destroy` should be the default operation, but there should be an option such as autoDestroy.

ErinCoughlan:
@clshortfuse Destroy is still a separate method from `open/close` and `show/hide`. If you want to reuse components, you would call any of those methods and only call `destroy` when you truly want to kill it.

clshortfuse:
Ah, I see it now. Should panels autoDestroy in that case then? I can foresee a lot of users not expecting that and inadvertently making memory leaks. If angular does destroy it when it gets garbage collected, then that's fine too.

ErinCoughlan:
I don't like autoDestroy since it adds additional configuration. I don't think it's any clearer to tell the users to set some field on the config object vs telling them to call a method. The panel is low level enough that anyone using it should create a directive that they can then use throughout their app, so they only need to know once.

I think making destroy behavior clearer in the documentation could help. All of the examples also include destroy so if someone forks it, they will have a working version.

clshortfuse:
+1 for docs. Perhaps even comments in the Basic Usage code.

crisbeto:
I've mentioned in tje docs for the close method that it doesn't destroy the instance.

PR #8852, opened 2016-06-25
fix(menu-item): properly compile when used with ng-repeat

### LINKED ISSUE #8697: md-menu: ng-repeat with md-menu-item
**Actual behavior**:
-  What is the issue? 
When using ngRepeat with mdMenuItem (type 'checkbox'), the icon is null and throws an error

-  CodePen Demo which shows your issue : http://codepen.io/anon/pen/xOwMwq
-  Details:

**Angular Versions**: *
-  Angular Version: 1.5.6
-  Angular Material Version: 1.1.0-rc4 >

**Additional Information**:
-  Browser Type: *
-  Browser Version: *
-  OS: *
-  Stack Traces:
TypeError: Cannot read property 'style' of undefined
    at render [as $render] (http://localhost:58246/Scripts/angular-material.js:27808:11)
    at ngModelWatch (http://localhost:58246/Scripts/angular.js:27102:14)
    at Scope.$digest (http://localhost:58246/Scripts/angular.js:17211:34)
    at Scope.$apply (http://localhost:58246/Scripts/angular.js:17486:24)
    at done (http://localhost:58246/Scripts/angular.js:11637:47)
    at completeRequest (http://localhost:58246/Scripts/angular.js:11843:7)
    at XMLHttpRequest.requestLoaded (http://localhost:58246/Scripts/angular.js:11776:9) undefined

crisbeto:
* Fixes the menuItem directive not compiling properly when used together with ngRepeat. This is due to the fact that ngRepeat passes around a documentFragment, instead of a DOM node, which means that it doesn't have a parent and doesn't allow us to properly determine whether the menuItem is inside of a menuBar. This change switches the behavior by making the menuBar directive mark it's children instead.
* Minor cleanup in the menuItem directive.
* Avoided some repetition in the menuBar unit tests.
* Switches to using the prefixer in the menuItem directive.

Referencing #8117.

Fixes #8697.

### DIFF
--- a/src/components/menuBar/js/menuBarDirective.js
+++ b/src/components/menuBar/js/menuBarDirective.js
@@ -123,6 +123,13 @@ function MenuBarDirective($mdUtil, $mdTheming) {
         }
       });
 
+      // Mark the child menu items that they're inside a menu bar. This is necessary,
+      // because mnMenuItem has special behaviour during compilation, depending on
+      // whether it is inside a mdMenuBar. We can usually figure this out via the DOM,
+      // however if a directive that uses documentFragment is applied to the child (e.g. ngRepeat),
+      // the element won't have a parent and won't compile properly.
+      templateEl.find('md-menu-item').addClass('md-in-menu-bar');
+
       return function postLink(scope, el, attr, ctrl) {
         el.addClass('_md');     // private md component indicator for styling
         $mdTheming(scope, el);
--- a/src/components/menuBar/js/menuItemDirective.js
+++ b/src/components/menuBar/js/menuItemDirective.js
@@ -6,12 +6,16 @@ angular
  /* @ngInject */
 function MenuItemDirective($mdUtil) {
   return {
+    controller: 'MenuItemController',
     require: ['mdMenuItem', '?ngModel'],
     priority: 210, // ensure that our post link runs after ngAria
     compile: function(templateEl, templateAttrs) {
+      var type = templateAttrs.type;
+      var inMenuBarClass = 'md-in-menu-bar';
 
       // Note: This allows us to show the `check` icon for the md-menu-bar items.
-      if (isInsideMenuBar() && (templateAttrs.type == 'checkbox' || templateAttrs.type == 'radio')) {
+      // The `md-in-menu-bar` class is set by the mdMenuBar directive.
+      if ((type == 'checkbox' || type == 'radio') && templateEl.hasClass(inMenuBarClass)) {
         var text = templateEl[0].textContent;
         var buttonEl = angular.element('<md-button type="button"></md-button>');
             buttonEl.html(text);
@@ -20,10 +24,10 @@ function MenuItemDirective($mdUtil) {
         templateEl.html('');
         templateEl.append(angular.element('<md-icon md-svg-icon="check"></md-icon>'));
         templateEl.append(buttonEl);
-        templateEl[0].classList.add('md-indent');
+        templateEl.addClass('md-indent').removeClass(inMenuBarClass);
 
-        setDefault('role', (templateAttrs.type == 'checkbox') ? 'menuitemcheckbox' : 'menuitemradio', buttonEl);
-        angular.forEach(['ng-disabled'], moveAttrToButton);
+        setDefault('role', type == 'checkbox' ? 'menuitemcheckbox' : 'menuitemradio', buttonEl);
+        moveAttrToButton('ng-disabled');
 
       } else {
         setDefault('role', 'menuitem', templateEl[0].querySelector('md-button, button, a'));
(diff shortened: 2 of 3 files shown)


src/components/menuBar/js/menuItemDirective.js
```diff
@@ -20,10 +23,11 @@ function MenuItemDirective($mdUtil) {
...
         templateEl.append(angular.element('<md-icon md-svg-icon="check"></md-icon>'));
         templateEl.append(buttonEl);
-        templateEl[0].classList.add('md-indent');
+        templateEl.addClass('md-indent');
+        templateAttrs.$set('mdInMenuBar', null);
 
-        setDefault('role', (templateAttrs.type == 'checkbox') ? 'menuitemcheckbox' : 'menuitemradio', buttonEl);
-        angular.forEach(['ng-disabled'], moveAttrToButton);
+        setDefault('role', type == 'checkbox' ? 'menuitemcheckbox' : 'menuitemradio', buttonEl);
+        moveAttrToButton('ng-disabled');
```
DevVersion: This function should be updated to use the `prefixer`.

To check if the attribute (with all prefixes) is present on the element.
```js
$mdUtil.prefixer().hasAttribute(element, 'ng-disabled');
```

src/components/menuBar/menu-bar.spec.js
```diff
@@ -417,27 +419,24 @@ describe('material.components.menuBar', function() {
...
-        inject(function($compile, $rootScope) {
-          // We need to have a `md-menu-bar` as a parent of our menu item, because the menu-item
-          // is only wrapping and indenting the content if it's inside of a menu bar.
-          var menuBarMock = angular.element('<md-menu-bar>');
-          var itemEl = angular.element(template);
+    function setupMenuItem(attrs) {
+      // We need to have a `md-menu-bar` as a parent of our menu item, because the menu-item
+      // is only wrapping and indenting the content if it's inside of a menu bar.
+      var menuBar;
+      var template = '' +
```
DevVersion: This empty `''` can be removed.

src/components/menuBar/js/menuBarDirective.js
```diff
@@ -123,6 +123,13 @@ function MenuBarDirective($mdUtil, $mdTheming) {
         }
       });
 
+      // Mark the child menu items that they're inside a menu bar. This is necessary,
+      // because mnMenuItem has special behaviour during compilation, depending on
+      // whether it is inside a mdMenuBar. We can usually figure this out via the DOM,
+      // however if a directive that uses documentFragment is applied to the child (e.g. ngRepeat),
+      // the element won't have a parent and won't compile properly.
+      templateEl.find('md-menu-item').attr('md-in-menu-bar', '');
```
DevVersion: Just curious, why not using a class here? A class does not require any value. It's just single state.

crisbeto: Doesn't matter really, I'll switch it up.

src/components/menuBar/menu-bar.spec.js
crisbeto: It's in case the automatic semicolon insertion catches it.

DevVersion: Interesting. It seems to work for me (Webstorm)

Also we don't do this in other components. But it should be fine, just very nitpicky

- https://github.com/angular/material/blob/master/src/components/autocomplete/autocomplete.spec.js#L590

src/components/menuBar/js/menuItemDirective.js
crisbeto: Done.

src/components/menuBar/js/menuItemDirective.js
```diff
@@ -20,10 +23,11 @@ function MenuItemDirective($mdUtil) {
         templateEl.html('');
         templateEl.append(angular.element('<md-icon md-svg-icon="check"></md-icon>'));
         templateEl.append(buttonEl);
-        templateEl[0].classList.add('md-indent');
+        templateEl.addClass('md-indent');
+        templateAttrs.$set('mdInMenuBar', null);
```
DevVersion: I think you forgot to change that - it should remove the class.

src/components/menuBar/js/menuItemDirective.js
DevVersion: Sorry I was not clear.

I mean, that we should also move prefixed attributes.

```js
function moveAttrToButton(attribute) {
  var attributes = $mdUtil.prefixer(attribute);

  attributes.forEach(function(attr) {
    if (templateEl.hasAttribute(attr) {
      button[0].setAttribute(attr, val);
      templateEl[0].removeAttribute(attr);
    }
  }
```

src/components/menuBar/js/menuItemDirective.js
crisbeto: Ops, thanks. Good catch.

DevVersion:
@ThomasBurleson Looks good to me ;)

PR #8853, opened 2016-06-25
fix(autocomplete): check select-on-match with cached results.

### LINKED ISSUE #8836: mdAutocomplete: auto-select doesn't select if value is re-entered and caching is enabled
**Actual behavior**:
-  What is the issue? * The autocomplete will not auto-select the value if it is re-entered and the built in caching is enabled.  For example 'value' gets selected and the callback is fired with 'value'.  I remove 'value' and undefined will be provided in the callback. Re-entering 'value' will not trigger the autoselect again.  This happens after any value has been auto-selected and removed, and only appears to be happening when the internal cache (aka not using md-no-cache) is used.

probably related to #8768 (item 7 of that issue, though that issue is severely lacking information.)
-  What is the expected behavior? The value will be auto-selected again as it was the first time it was auto-selected.

-  CodePen Demo which shows your issue : http://codepen.io/mckenzielong/pen/yJgByz
-  Details: First enter the value 'Hawaii' -- it will be auto-selected (note the callback).  Tab out to the check box, then shift-tab back and clear the value (note the callback.)  Now re-enter 'Hawaii'.  The suggestions open up and the value is not automatically selected.  

md-no-cache returns the expected behavior.

**Angular Versions**: *
-  Angular Version: 1.5.6
-  Angular Material Version: 1.1.0 rc5

**Additional Information**:
-  Browser Type: * Chrome/Opera
-  Browser Version: * 50/36
-  OS: * Windows 7 64bit
-  Stack Traces: N/A

----

  Add more lines if needed.

DevVersion:
* The autocomplete was not checking for any exact matches if the results were cached.
* Cached results will be now handled the same as they would be treated if they weren't cached.

Fixes #8836.

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -1102,6 +1102,7 @@ describe('<md-autocomplete>', function() {
   });
 
   describe('md-select-on-match', function() {
+
     it('selects matching item on exact match when `md-select-on-match` is toggled', inject(function($timeout) {
       var scope = createScope();
       var template = '\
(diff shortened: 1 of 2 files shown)


DevVersion:
@ThomasBurleson This should probably go into `1.1.0`, since it's kind of a major issue.

PR #8870, opened 2016-06-27
Fix a punctuation error

isaaclyman:
$mdToast documentation uses "it's" (contraction form) where it means "its" (possessive form).


DevVersion:
LGTM @ThomasBurleson

PR #8871, opened 2016-06-27
fix(md-theme) add missing literal slashes to themeNameRegex

clshortfuse:
`.` in regex strings means any character. We want to search for the literal `.`


EladBezalel:
lgtm

PR #8872, opened 2016-06-27
feat(mdTheme) add contrast opacity values for all color types and hues

clshortfuse:
This new change allows one to use different opacity types per color type and hue. This also deprecates `foreground-*` and instead uses `background-contrast-*`. In short:

    newMapping = {
      "foreground-1": "background-contrast",
      "foreground-2": "background-contrast-secondary",
      "foreground-3": "background-contrast-hint",
      "foreground-4": "background-contrast-divider"
    }

I also added `contrast-disabled` and `contrast-icon`.

Now you can create css for items with `primary-contrast-secondary` or others. This should help close issue #7987 

For example, you can now do this:

    md-toolbar.md-THEME_NAME-theme:not(.md-menu-toolbar) md-input-container {
      .md-input {
        @include input-placeholder-color('\'{{primary-contrast-hint}}\'');
        border-color: '{{primary-contrast-divider}}';
      }
    }

There are no color changes in this PR. I will add that previous CSS rules that used `foreground-` with a numerical opacity value were wrong. `foreground-` did not support custom opacities and it was ignored.


src/core/services/theming/theming.js
```diff
@@ -689,6 +717,45 @@ function generateAllThemes($injector, $mdTheming) {
     delete palette.contrastStrongLightColors;
     delete palette.contrastDarkColors;
 
+    function getContrastType(hueName) {
+      if (defaultContrast === 'light') {
+        if (darkColors.indexOf(hueName) > -1) {
```
DevVersion: I prefer the `!== -1` check for `indexOf` calls.

See https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf

clshortfuse: Inequality check is probably faster mathematically than "greater than" check. Updated.

src/core/services/theming/theming.js
```diff
@@ -689,6 +717,45 @@ function generateAllThemes($injector, $mdTheming) {
...
+        }
+        return 'light';
+      }
+      if (lightColors.indexOf(hueName) !== -1) {
+        if (strongLightColors.indexOf(hueName) !== -1) {
+          return 'strongLight';
+        }
+        return 'light';
+      }
+      return 'dark';
```
EladBezalel: Can't all of this be merged into one?

clshortfuse:
I think we need to come up with a better solution for legacy support. Maybe if foregroundPalette isn't user generated set, we should use background-contrast instead

clshortfuse:
Ugh, apparently it pushed everything i wanted to revert over my old code. Now I'm on the hunt to see how I can undo it.

EladBezalel:
Please squash your commits and have a commit message that follows our style guide

clshortfuse:
@EladBezalel okay, I've reworked it and added some tests. I don't think there should any possible behavioral change. Stylistically, we'll be fixing incorrect opacities that are no longer part of spec (87% opacity white).

src/core/services/theming/theme.palette.js
```diff
@@ -16,8 +16,7 @@ angular.module('material.core.theming.palette', [])
     'A400': '#ff1744',
     'A700': '#d50000',
     'contrastDefaultColor': 'light',
-    'contrastDarkColors': '50 100 200 300 A100',
-    'contrastStrongLightColors': '400 500 600 700 A200 A400 A700'
```
EladBezalel: Why `contrastStrongLightColors`? there are colors that the white or black doesn't have 87% opacity.

clshortfuse:
@EladBezalel I have a bad habit of seeing something wrong and slapping it onto the PR I'm working on. 

I've reverted back to white 87%, I'll file and issue and make a PR for that later.

src/core/services/theming/theming.js
```diff
@@ -551,20 +568,43 @@ function parseRules(theme, colorType, rules) {
...
+  var simpleVariableRegex = /'?"?\{\{\s*([a-zA-Z]+)-(A?\d+|hue\-[0-3]|shadow|default)-?(contrast)?-?((?:\d\.?\d*)|(?:[a-zA-Z]+))?\s*\}\}'?"?/g;
   var palette = PALETTES[color.name];
+  var defaultBgHue = theme.colors['background'].hues['default'];
+  var defaultBgContrastType = PALETTES[theme.colors['background'].name][defaultBgHue].contrastType;
 
   // find and replace simple variables where we use a specific hue, not an entire palette
   // eg. "{{primary-100}}"
   //\(' + THEME_COLOR_TYPES.join('\|') + '\)'
-  rules = rules.replace(simpleVariableRegex, function(match, colorType, hue, opacity, contrast) {
+  rules = rules.replace(simpleVariableRegex, function(match, colorType, hue, contrast, opacity) {
```
clshortfuse: While it looks like this would be a breaking change the order groups in the regex, that's not the case. Before, either `opacity` or `contrast` could be used, but not both. Now both can be used.

docs/guides/THEMES_IMPL_NOTES.md
```diff
@@ -24,15 +24,20 @@ mini-DSL, applies the colors for the theme, and appends the resulting CSS into t
...
-color on an accent-colored background.
+* Palettes are `primary`, `accent`, `warn`, `background`
+* The hues for each type use the Material Design hues. When not specified, each palette defaults
+`hue` to `500` with the exception of `background`
+* The `opacity` value can be a decimal between 0 and 1 or one the following values:
+  * `icon`: icon
+  * `secondary`: secondary text,
+  * `disabled`: disabled text or icons
+  * `hint`: hint text,
+  * `divider`: dividers
```
topherfangio: I think it would be helpful to have the numerical values displayed here as well, just so users have an idea of what these names mean.

src/core/services/theming/theming.js
```diff
@@ -185,6 +174,34 @@ var DARK_DEFAULT_HUES = {
     'hue-3': 'A200'
   }
 };
+
+// use inactive-icon opacity from  https://material.google.com/style/color.html#color-text-background-colors
+// do not system icon opacity from https://material.google.com/style/icons.html#icons-system-icons
```
topherfangio: Typo? "do not system"

src/core/services/theming/theming.js
topherfangio: Is there any way that we could extract some of this logic into a service (or something) so that we can write some tests? I'd love some tests to make sure that it parses everything correctly in case we make more changes in the future.

src/core/services/theming/theming.js
```diff
@@ -551,20 +568,43 @@ function parseRules(theme, colorType, rules) {
...
+        return rgba( colorToRgbaArray( theme.foregroundPalette[hue] ) );
+      } else if (theme.foregroundPalette['1']){
+        return rgba( colorToRgbaArray( theme.foregroundPalette['1'] ) );
+      }
+      // Default to background-default-contrast-{opacity}
+      colorType = 'background';
+      contrast = 'contrast';
+      if (!opacity && hue) {
+        // Convert references to legacy hues to opacities (ie: foreground-4 to *-divider)
+        switch(hue) {
```
topherfangio: Curious, why do this not include a `case` for `foreground-1`?

src/core/services/theming/theming.js
```diff
@@ -689,6 +758,40 @@ function generateAllThemes($injector, $mdTheming) {
     delete palette.contrastStrongLightColors;
     delete palette.contrastDarkColors;
 
+    function getContrastType(hueName) {
+      if (defaultContrast === 'light') {
+        if (darkColors.indexOf(hueName) > -1) return 'dark';
```
topherfangio: Although the style guide doesn't specifically state it, we (well maybe just me) usually avoid one-line `if` statements as they can cause confusion if the formatter runs and changes the lines.

I know it's more lines of code, but can you use curly braces here and below and put the `return` statements on their own lines?

src/core/services/theming/theming.js
topherfangio: Ignore that comment; didn't see all of the tests at the bottom 😆

topherfangio:
Just to confirm; we don't need to change any existing code because you have mapped all of the old values to the new ones?

clshortfuse:
@topherfangio Correct. We don't have to change the CSS files because of the mapping. That said, I would like to migrate the CSS to the new syntax, but it's not necessary right now. I was careful to not break any CSS that uses `foreground` so if we do move away from that syntax in the future, users' custom CSS component that use `foreground` doesn't break.

src/core/services/theming/theming.js
clshortfuse: That's funny because I normally use explicit braces and used inline because I saw it being using elsewhere in our code.

src/core/services/theming/theming.js
clshortfuse: The first `if`checks if the user has a custom foreground color that matches the hue specified in the CSS. The second `if` uses the user specified `foreground-1` as a fallback. (ie: CSS wants foreground-4 but user only specified foreground-1) . I'm not seeing a case for a switch statement here without having an if statement inside the case.

topherfangio: Sorry, look down from this comment. You have:

```js
switch (hue) {
  case '2': ...
  case '3': ...
  case '4': ...
}
```

My question was why there was no `case '1'`?

clshortfuse: Ah, because `foreground-1` uses default opacity. I was thinking of adding a `primary` opacity for clarity, but that's redundant. I'll add a comment in the switch case to explain why we don't need a check for `foreground-1`

src/core/services/theming/theming.js
```diff
@@ -689,6 +758,40 @@ function generateAllThemes($injector, $mdTheming) {
...
+    }
+    function getOpacityValues(contrastType) {
+      switch(contrastType) {
+        default:
+        case 'strongLight':
+          return STRONG_LIGHT_CONTRAST_OPACITY;
+        case 'light':
+          return LIGHT_CONTRAST_OPACITY;
+        case 'dark':
+            return DARK_CONTRAST_OPACITY;
```
EladBezalel: nit: extra spaces

src/core/services/theming/theming.js
topherfangio: Sounds good 👍

docs/guides/THEMES_IMPL_NOTES.md
clshortfuse: Added: 
> Using a decimal opacity value as in `accent-contrast-0.25` will apply the contrast color for the accent color at 25% opacity.

src/core/services/theming/theming.js
clshortfuse: I'm not sure I can refactor it much more:

Current:
```
if (a) {
  if (b) return 3;
  if (c) return 2;
  return 1;
}
if (d) {
  if (c) return 2;
  return 1;
}
return 3;
```
I can do this:
```
if (a) {
  if (b) return 3;
  if (c) return 2;
  return 1;
}
if (!d) return 3;
if (c) return 2;
return 1;
```

docs/guides/THEMES_IMPL_NOTES.md
topherfangio: Sorry, I meant above; where you have

`icon`: icon
`secondary`: secondary

etc

I think those should be followed by the opacity value that it sets.

topherfangio:
@ThomasBurleson LGTM

@EladBezalel You want to give it one more look over?

src/core/services/theming/theming.js
clshortfuse: For reference, comment was added below the line notes. That's why GitHub isn't marking these comments as outdated.

topherfangio:
@clshortfuse Correct me if I'm wrong, but this is slated for 1.2.0 right? If so, I'll update the milestone.

src/core/services/theming/theming.js
```diff
@@ -639,13 +680,42 @@ function parseRules(theme, colorType, rules) {
...
     }
 
-    return rgba( (PALETTES[ theme.colors[colorType].name ][hue] || '')[contrast ? 'contrast' : 'value'], opacity );
+    var colorDetails = (PALETTES[ theme.colors[colorType].name ][hue] || '');
+
+    // If user has specified a foreground color, use those
+    if (colorType === 'background' && contrast && regexColorType !== 'foreground' && colorDetails.contrastType == defaultBgContrastType) {
+      // Don't process if colorType was changed
+      switch (opacity) {
+        default:
```
EladBezalel: move default to be the last

src/core/services/theming/theming.js
```diff
@@ -639,13 +680,42 @@ function parseRules(theme, colorType, rules) {
...
+          if (theme.foregroundPalette['1']) return rgba( colorToRgbaArray( theme.foregroundPalette['1'] ) );
+          break;
+        case 'secondary':
+        case 'icon':
+          if (theme.foregroundPalette['2']) return rgba( colorToRgbaArray( theme.foregroundPalette['2'] ) );
```
EladBezalel: the line feels overloaded, put in brackets or at least break return to a separate line

src/core/services/theming/theming.js
```diff
@@ -639,13 +680,42 @@ function parseRules(theme, colorType, rules) {
...
+    }
+
+    return rgba( colorDetails[contrast ? 'contrast' : 'value'], opacity );
   });
 
   // For each type, generate rules for each hue (ie. default, md-hue-1, md-hue-2, md-hue-3)
   angular.forEach(color.hues, function(hueValue, hueName) {
     var newRule = rules
-      .replace(hueRegex, function(match, _, colorType, hueType, opacity) {
+      .replace(hueRegex, function(match, _, colorType, hueType, contrast, opacity) {
```
EladBezalel: why `_` is passed?

ThomasBurleson:
Consider publishing simultaneously with #8994 

> Need VisualDiff testing also.

topherfangio:
@topherfangio Figure out what milestone this needs to be in.

clshortfuse:
Rebased. Would be nice to have this merged to fix some a11y issues

Splaktar:
Fixed the Travis issue. I'd like to run this through presubmit in the next batch.

Splaktar:
Can you please rebase this?

I'll try to do some manual testing and see if we can possibly get this in soon.

clshortfuse:
@Splaktar I'll try to do it this weekend.

PR #8883, opened 2016-06-28
docs: Add mobile-specific styles for better usability.

### LINKED ISSUE #8677: docs: Reduce margin/padding of demos on mobile devices
**Actual behavior**:
-  What is the issue? * Margins on the demo site are very large on mobile devices and squish the content.
-  What is the expected behavior? On mobile devices, the demo padding/margins should be smaller to allow more space for the content.

-  CodePen Demo which shows your issue : N/A (see below)
-  Details: It should look more like the image on the right below:

margin-padding-reduction

**Angular Versions**: *
-  Angular Version: N/A
-  Angular Material Version: 1.1.0-RC5

**Additional Information**:
-  Browser Type: *All
-  Browser Version: *All
-  OS: *All
-  Stack Traces: N/A

topherfangio:
- Reduce the margins/padding around doc content and demos so that
   the content/demos utilize more of the screen.

 - Reduce the size of the sidenav logo/header so more of the actual
   navigation can be seen.

Fixes #8677.

### DIFF
--- a/docs/app/css/style.css
+++ b/docs/app/css/style.css
@@ -934,3 +934,36 @@ table.md-api-table  tr  td:first-child {
   padding-top: 10px;
   padding-bottom: 10px;
 }
+
+/* Provide some layout/padding enhancements for mobile/small devices */
+@media(max-width: 500px) {
+  /* Reduce the size of the nav logo/header */
+  .docs-logo > img {
+    height: 75px;
+    margin-top: 10px;
+  }
+
+  .docs-logo h1 {
+    font-size: 18px;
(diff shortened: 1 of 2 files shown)


DevVersion:
@topherfangio While improving those things for mobile devices. 

It feels kind of weird, that the user can zoom in the Docs on a mobile device.
Normally in SPA's the scaling is disabled.

```html
<meta name="viewport" content="initial-scale=1" />
```

```html
<meta name="viewport" content="width=device-width, initial-scale=1.0, user-scalable=no">
```

What do you think?

Issue #3084, opened 2015-06-02
[md-autocomplete] Unable to reset the input


jgoux:
Hi,
on the master, I'm unable to reset the selected input from md-autocomplete by setting the model to *null*
Here is the demo : http://codepen.io/jgx/pen/XbMmxE

CxRes:
Well, you need to set the md-search-text variable to null as well. See the modified demo: http://codepen.io/anon/pen/oXZmVG

stefan-schweiger:
@CxRes this works but if you have set `md-min-length="0"` it will open the popup again.

Anyways I think it should work the same ways as with `md-select` by just setting the selected item to `null`.

robertmesserle:
This is as designed and would require you to set both the searchText and the selectedItem.  I believe resetting just searchText would also clear selectedItem.

stefan-schweiger:
@robertmesserle 

![why don't we have both?](http://i.imgur.com/sp8v92J.gif?1)

it would be more in line with `md-select` as you also just set the `ng-model` to `null` (and in the context of `autocomplete` `selectedItem` is the model in my opinion)

btw setting searchText to `null` actually results in an exception, setting it to `''` works as far as I can tell in a codepen with the current master (but I had problems with a master from last week, so I will recheck this tomorrow at my actual project)

robertmesserle:
@stefan-schweiger The problem is that we have two models that we need to track for Autocomplete.  The problem is that an empty `searchText` always means that there is no `selectedItem`, but the opposite is not always true.

stefan-schweiger:
@robertmesserle ok I see. Also now I remember what the problem was, when you have set `md-min-length="0"` the popup comes back open when you set `md-search-text` to `''` and as it has no focus it behaves kind of weird.

http://codepen.io/anon/pen/waEzgv

What's even worse in my case is that I hide the `md-autocomplete` after selection to display the item, when you clear it again you have the popup in the upper left corner nowhere near the `md-autocomplete`.

![](http://i.imgur.com/AWhf65B.png)

tolemac:
The same here, I understand that autocomplete can have no model value but searchText has it.
Just need that popup dont back open.

stefan-schweiger:
As a workaround, you can disable the `md-autocomplete` before resetting and enable it again afterwards, which is stupid as hell, but at least it works...

technotomahi:
It looks like if we go by triggering the dom element it works. 

[...]

PR #8884, opened 2016-06-28
docs: Add how to reset md-autocomplete value

DavidFrahm:
I spent a lot of time trying to figure out how to reset the displayed value of an autocomplete that I was (re)using in a drawer. I eventually found the answer in issue https://github.com/angular/material/issues/3084.

So this update attempts to prevent others from struggling with this issue, which to me seems basic enough for it to be in the docs.


DavidFrahm:
I signed it!

PR #8888, opened 2016-06-28
fix(chips): Fix static chips remove padding.

### LINKED ISSUE #8887: chips: Static chips have erroneous "remove" padding.
**Actual Behavior**:
-  `What is the issue? *` The static chips have extra padding for the remove container.
-  `What is the expected behavior?` The static chips should not have the extra padding.

-  `CodePen Demo which shows your issue:` https://material.angularjs.org/HEAD/demo/chips
-  `Details:`

screen shot 2016-06-28 at 11 51 46 am

**Angular Versions**: *
-  `Angular Version:` N/A
-  `Angular Material Version:` HEAD

**Additional Information**:
-  `Browser Type: *` All
-  `Browser Version: *` All
-  `OS: *` All
-  `Stack Traces:`

----

  Add more lines if needed.

topherfangio:
Static chips had some erroneous "remove" padding due to some recent changes with the new `md-removable` feature.

Fix by updating the `isRemovable()` method to also check if the component has an `ng-model` supplied and return false if not.

Also, minor change to demo to provide slightly better feedback to the user.

Fixes #8887.

### DIFF
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -1067,6 +1067,14 @@ describe('<md-chips>', function() {
         expect(chipArray[1].innerHTML).toContain('2');
         expect(chipArray[2].innerHTML).toContain('3');
       });
+
+      it('does not allow removal of chips', function() {
+        scope.chipItem = 'Football';
+        var element = buildChips(STATIC_CHIPS_TEMPLATE);
+        var wrap = element.find('md-chips-wrap');
+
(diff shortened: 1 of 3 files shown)


DevVersion:
Even a better approach as the one we discussed. LGTM!

topherfangio:
@DevVersion Yeah; I started to do it in the template, but realized there were 3-4 places that would need to be updated, and this made more sense 😄

PR #8893, opened 2016-06-28
fix(select): Allow non-english characters for keyboard selection.

### LINKED ISSUE #7730: md-select: search does not work when you type the first letter of any other language
Hello Boys!

The specialist said that the problem could be [here](https://github.com/angular/material/blob/3524aa9573fe29e53341c60ce6127fc7ad323446/src/components/select/select.js):

`var searchStr = '';
var clearSearchTimeout, optNodes, optText;
var CLEAR_SEARCH_AFTER = 300;
self.optNodeForKeyboardSearch = function(e) {
  clearSearchTimeout && clearTimeout(clearSearchTimeout);
  clearSearchTimeout = setTimeout(function() {
    clearSearchTimeout = undefined;
    searchStr = '';
    optText = undefined;
    optNodes = undefined;
  }, CLEAR_SEARCH_AFTER);
  searchStr += String.fromCharCode(e.keyCode);
  var search = new RegExp('^' + searchStr, 'i');
  if (!optNodes) {
    optNodes = $element.find('md-option');
    optText = new Array(optNodes.length);
    angular.forEach(optNodes, function(el, i) {
      optText[i] = el.textContent.trim();
    });
  }
  for (var i = 0; i < optText.length; ++i) {
    if (search.test(optText[i])) {
      return optNodes[i];
    }
  }
};`

Also, experts say that the wrong key code is converted into a symbol.

[Example](http://codepen.io/yeoman/pen/MymmOg?editors=1111)

Version angular is 1.4.8.
Version angular materials is 1.1.0-rc1.

Thank you!

topherfangio:
Our previous method for checking whether or not the typed character matched the first letter of an `<md-option>` was constrained to the letters `[a-zA-Z]`.

Fix by allowing more characters and checking the validity via the `String.fromCharCode()` method.

Fixes #7730.


topherfangio:
@ThomasBurleson Not sure who would be best to review this besides yourself.

clshortfuse:
Your PR improves the interaction, and should probably go through, but I'm going to be honest, I'm not sure how I feel about keyCode.

We're translating keyCode to map to character, but I've seen bad results with different operating systems and browsers.

keyCode looks deprecated as well, unless I'm reading it wrong: https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/keyCode

Also, á is a Latin character, and still part of ASCII. I believe it's 160. You should put a test for Unicode.

topherfangio:
I looked at switching it out from using keyCode, but the replacement is supposed to be [KeyboardEvent.key](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key) which as far as I can tell from the Browser Compatibility list is extremely unsupported at the moment (it has 0 support from Safari and only became available in Chrome 51...the current version).

I'm certainly open to changing it if we can find a better approach, but I'm honestly thinking we might just wait until `.key` becomes more widely supported and replace it then.

clshortfuse:
Darn Safari. We can use keyCode as a fallback. To be honest, keyboard input parsing is very messy. I've had to write a while service/module in the past to deal with getting the right keys. If Unicode works with keyCode, we can backlog moving to ‘key‘. I still would want to do it, since I don't like the idea of swallowing inputs. For example F11 would trigger Fullscreen and Ctrl+ would zoom in. I don't want those actions to be unavailable when somebody triggersa popup

topherfangio:
I added a unicode test. We could add a bit more logic (inside a function) which filters out some of the known ones like the function keys.

I didn't think about that. I'll go ahead an work on that and see if I can add some more tests.

clshortfuse:
You can also just only stopPropagation if a match is found. I can't think of an instance where that would have a negative effect.

src/components/select/select.js
```diff
@@ -1582,4 +1582,14 @@ function SelectProvider($$interimElementProvider) {
...
   }
+
+}
+
+function shouldHandleKey(ev) {
+  var char = String.fromCharCode(ev.keyCode);
+  var isNonUsefulKey = (ev.keyCode <= 31);
+  var isLeftRightSelectKey = (ev.keyCode >= 91 && ev.keyCode <= 93);
+  var isFnLockKey = (ev.keyCode >= 112 && ev.keyCode <= 145);
+
```
clshortfuse: I think it's safe to abort if the user press `Ctrl` or `Meta` or `Alt`.  I say this because `master` locked me out from refreshing the page with Ctrl+Shift+R.

`Shift` alone should be okay. I tried with with `AltGr` and special Latin characters (ñÑéóö) and `Alt` is true while holding `AltGr` but gets set to `false` when I press AltGr+N. `ñ` is detected as it's own key with `Alt` set to `false`, so it looks safe.

src/components/select/select.js
```diff
@@ -1582,4 +1582,14 @@ function SelectProvider($$interimElementProvider) {
     }
     return isScrollable;
   }
+
+}
+
+function shouldHandleKey(ev) {
+  var char = String.fromCharCode(ev.keyCode);
+  var isNonUsefulKey = (ev.keyCode <= 31);
+  var isLeftRightSelectKey = (ev.keyCode >= 91 && ev.keyCode <= 93);
```
ofirmgr: Why not just comparing ev.keyCode to 92 instead?

topherfangio: Because we're checking for 91, 92 & 93. If we did that, we'd have:

```js
(ev.keyCode == 91 || ev.keyCode == 92 || ev.keyCode == 93)
```

which seemed even more verbose.

src/components/select/select.js
topherfangio: I'm really confused by what you mean here 😄 

You're saying we should also test for `Ctrl`, `Meta` or `Alt` and return false too?

clshortfuse: Yes return false. So users can refresh (Ctrl+R) or ctrl/alt tab while it's open

ThomasBurleson:
@topherfangio - please confirm this PR is valid and still needed. Then please rebase. Thank you.

topherfangio:
@ThomasBurleson Rebased and ready for a final review.

I do believe it is relevant because the old functionality was

- Allow only certain keys

whereas the new approach says

- Filter known bad keys, allow everything else

In particular, this means that non-english characters are allowed.

src/components/select/select.js
```diff
@@ -1673,4 +1673,13 @@ function SelectProvider($$interimElementProvider) {
     }
     return isScrollable;
   }
+
+}
+
+function shouldHandleKey(ev, $mdConstant) {
+  var char = String.fromCharCode(ev.keyCode);
+  var isNonUsefulKey = (ev.keyCode <= 31);
+
```
clshortfuse: We need a check to make sure the user isn't performing a keyboard combination. For example, we don't want to stop users from being able to close a window (Ctrl+W) or stop a Refresh (Ctrl+R) simply because one of the combination keys is valid. Basically, we should abort if Alt, Ctrl, or Meta is being held.

This isn't the same as the isMetaKey function, which is only checking for single key presses of meta keys.

PR #8901, opened 2016-06-29
fix(panel): Click outside to close when disable parent scrolling is enabled

### LINKED ISSUE #8891: mdPanel - Click to close when disable parent scrolling is enabled.
**Actual behavior**:
-  `What is the issue? *` When checking the *Disable Parent Scroll* option, the clickOutsideToClose functionality is not working.
-  `What is the expected behavior?` Clicking outside of the open panel should close the panel.

-  `CodePen Demo which shows your issue :` N/A
-  `Details:` This issue is on the current docs site:  https://material.angularjs.org/latest/demo/panel

**Angular Versions**: *
-  `Angular Version:` Latest

**Additional Information**:
-  `Browser Type: *` Chrome
-  `OS: *` OS X 10.11.5
-  `Stack Traces:` N/A

----

  Add more lines if needed.

bradrich:
- Added options to the `$mdUtil.disableScrollAround` function, the only option so far is `disableScrollMask`, a boolean that when passed in as true will force the function to not create the `md-scroll-mask` element and instead use the passed-in parent element.
- Modified the `MdPanelRef.prototype._createPanel` function to use the new `disableScrollMask` option.

Fixes #8891.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -1119,7 +1119,10 @@ MdPanelRef.prototype._createPanel = function() {
 
           if (self._config['disableParentScroll']) {
             self._restoreScroll = self._$mdUtil.disableScrollAround(
-                null, self._panelContainer);
+              null,
+              self._panelContainer,
+              { doNotCreateScrollMaskElement: true }
+            );
           }
 
           self._panelEl = angular.element(
@@ -1197,12 +1200,12 @@ MdPanelRef.prototype._updatePosition = function(opt_init) {
 
   if (positionConfig) {
     positionConfig._setPanelPosition(this._panelEl);
-    
+
     // Hide the panel now that position is known.
     if (opt_init) {
       this._panelContainer.addClass(MD_PANEL_HIDDEN);
     }
-    
+
     this._panelEl.css('top', positionConfig.getTop());
     this._panelEl.css('bottom', positionConfig.getBottom());
     this._panelEl.css('left', positionConfig.getLeft());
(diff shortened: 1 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -1119,7 +1119,10 @@ MdPanelRef.prototype._createPanel = function() {
 
           if (self._config['disableParentScroll']) {
             self._restoreScroll = self._$mdUtil.disableScrollAround(
-                null, self._panelContainer);
+              null,
+              self._panelContainer,
+              { doNotCreateScrollMaskElement: true }
```
DevVersion: I would rather call this
```js
disableScrollMask: true
```

It's possible to name it that way, because the function is actually called `disableScrollAround`.

bradrich: @DevVersion That is a good idea. I will make the change.

src/core/util/util.js
```diff
@@ -218,12 +221,18 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
-          '<div class="md-scroll-mask">' +
-          '  <div class="md-scroll-mask-bar"></div>' +
-          '</div>');
-        element.appendChild(scrollMask[0]);
+        element = angular.element(element || body);
+        var scrollMask;
+        if (options && options.disableScrollMask) {
+          scrollMask = element;
+        } else {
+          element = element[0];
```
crisbeto: Why are you dealing with the raw DOM node here? You can do `element.append(scrollMask)`.

bradrich: @crisbeto I did not change that functionality from the original.

However, I understand that `angular.element` creates a JQLite object of the element and that `[0]` gets the HTML node that can be appended / prepended. Whether or not `element.append` of a JQLite object will work is something that I did not introduce to this functionality.

ErinCoughlan:
Can you add a test to panel that verifies the scrolling is correct now?

bradrich:
@ErinCoughlan I sure can and I will re-push.

Issue #8753, opened 2016-06-13
Autocomplete:  md-input-minlength and md-input-maxlength doesn't translate to input


eddieebeling:
**Actual behavior**:
-  What is the issue? When adding md-input-maxlength on the autocomplete directive the attribute and values are not translated to the input when using the non-floating label autocomplete directive.
-  What is the expected behavior? ng-minlength and ng-maxlength are added to the input.

**CodePen** or Steps to reproduce the issue: *
-  CodePen Demo which shows your issue : http://codepen.io/eddieebeling/pen/KMzZyR
-  Details: Added md-input-maxlength="18" to the directive and the ng-maxlength is not rendered with the input element.

**Angular Versions**: *
-  Angular Version: 1.5.5
-  Angular Material Version: 1.1.0-rc.5

**Additional Information**:
-  Browser Type: Webkit
-  Browser Version: Chrome
-  OS: Mac OS X
-  Stack Traces:

----
Shortcut to create a [new CodePen Demo](http://codepen.io/team/AngularMaterial/pen/bEGJdd).
Note: `*` indicates required information. Without this information, your issue may be auto-closed.

> Do not modify the titles or questions. Simply add your responses to the ends of the questions.
  Add more lines if needed.

clshortfuse:
Going to add on there's another way to see this issue.

https://material.angularjs.org/HEAD/demo/autocomplete

On Floating Label:

* Focus Favorite State autocomplete
* Type `a`
* press Escape
* note that `a` isn't cleared
* Focus autocomplete
* Watch it disappear

DevVersion:
We will address that with #4492. Thanks for the issue!

Issue #4492, opened 2015-09-08
mdAutocomplete: pass ng-* applicable attributes to input


jualvarez:
Attributes applicable to the input tag within an autocomplete should be passed to the control. For example: ng-trim, ng-blur, ng-focus, etc...

spoco2:
Yes. +1 from me for this one. I need ng-trim in particular, having nothing happen when spaces are added to the input is frustrating.

ThomsC:
+1
I need ng-trim and some custom directives too

bayucandra:
+1, absolutely need ng-trim="false", as sometimes space is necessary to do some programmatic things on the input.

DevVersion:
I was investigating at some possible solutions to dynamically forward attribute events to another element.

> For example I was linking the different attributes by connecting the different scopes with internal variables. 
> Those internal variables / functions will be parsed and evaluated in the user scope.

Unfortunately are we not able to access the `$locals` of a parsed expression, so this approach didn't work out.

- https://github.com/angular/angular.js/pull/13454 (>`1.5.x`)

We decided to manually forward the `ngBlur` and `ngFocus` to the input, since those are already used by autocomplete anyway.

`ngTrim` will follow in another change.

PR #8904, opened 2016-06-29
feat(autocomplete): optionally replace input while using up and down keys

clshortfuse:
Creates an option for replacing the input text with the selected dropdown
item when using keyboard navigation

* Add `md-replace-text-on-select` option
* Update docs to show replace text behavior
* Create a new model called `inputText` representing the input box text
* Do not change `searchText` unless user types a change
* Allow users to accept replaced text with Enter or Tab

Note: This creates added complexity to autocomplete, since the user changes
the `inputText` model, and then the `searchText` is updated. Essentially,
this will decouple the text that exists inside the input box from the text
from which queries are run against. This will facilitate the implementation
of an "append text" autocomplete interaction.

Autocomplete UX Research: https://docs.google.com/document/d/1t60A2_Qni3IOSZDRqwwum3r5EVPjANJAkvFuMoqGfQA/

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -22,7 +22,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       lastCount            = 0,
       fetchesInProgress    = 0,
       enableWrapScroll     = null,
-      inputModelCtrl       = null;
+      inputModelCtrl       = null,
+      suggestedInputText   = '';
 
   // Public Exported Variables with handlers
   defineProperty('hidden', handleHiddenChange, true);
@@ -66,7 +67,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    * Initialize the controller, setup watchers, gather elements
    */
   function init () {
-    $mdUtil.initOptionalProperties($scope, $attrs, { searchText: '', selectedItem: null });
+    $mdUtil.initOptionalProperties($scope, $attrs, { inputText: '', searchText: '', selectedItem: null });
     $mdTheming($element);
     configureWatchers();
     $mdUtil.nextTick(function () {
@@ -189,6 +190,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     $attrs.$observe('disabled', function (value) { ctrl.isDisabled = $mdUtil.parseAttributeBoolean(value, false); });
     $attrs.$observe('required', function (value) { ctrl.isRequired = $mdUtil.parseAttributeBoolean(value, false); });
     $attrs.$observe('readonly', function (value) { ctrl.isReadonly = $mdUtil.parseAttributeBoolean(value, false); });
+    $scope.$watch('inputText', handleInputText);
     $scope.$watch('searchText', wait ? $mdUtil.debounce(handleSearchText, wait) : handleSearchText);
     $scope.$watch('selectedItem', selectedItemChange);
     angular.element($window).on('resize', positionDropdown);
@@ -397,7 +399,26 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   }
 
   /**
-   * Handles changes to the searchText property.
+   * Handles changes to the inputText property.
+   * @param inputText
+   * @param previousInputText
+   */
+  function handleInputText (inputText, previousInputText) {
+    // Only continue if display text was changed by user
+    if (suggestedInputText === inputText) return;
+    ctrl.index = getDefaultIndex();
+    suggestedInputText = undefined;
+
+    // do nothing on init
+    if (inputText === previousInputText) return;
+    if (inputText === $scope.searchText) return;
+    $scope.$evalAsync(function() {
+      $scope.searchText = inputText;
+    });
+  }
+
+  /**
+   * Handles calls from the searchText watcher.
    * @param searchText
    * @param previousSearchText
    */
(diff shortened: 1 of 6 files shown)


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -169,6 +170,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     $attrs.$observe('disabled', function (value) { ctrl.isDisabled = $mdUtil.parseAttributeBoolean(value, false); });
     $attrs.$observe('required', function (value) { ctrl.isRequired = $mdUtil.parseAttributeBoolean(value, false); });
     $attrs.$observe('readonly', function (value) { ctrl.isReadonly = $mdUtil.parseAttributeBoolean(value, false); });
+    $scope.$watch('displayText', wait ? $mdUtil.debounce(handleDisplayText, wait) : handleDisplayText);
```
clshortfuse: `wait` isn't need here and will only serve to double the wait time

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -450,6 +483,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         ctrl.index   = ctrl.index < 0 ? ctrl.matches.length - 1 : Math.max(0, ctrl.index - 1);
         updateScroll();
         updateMessages();
+        updateDisplayText();
```
clshortfuse: probably needs an option so it's not default for everyone who upgrade

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -442,6 +474,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         ctrl.index   = Math.min(ctrl.index + 1, ctrl.matches.length - 1);
         updateScroll();
         updateMessages();
+        updateDisplayText();
```
clshortfuse: probably needs an option so it's not default for people who upgrade

gkalpak:
Not directly related to this PR, but I wonder if some watchers could be disabled ay some point (e.g. when the element doesn't have focus) and be re-enabled when necessary.

clshortfuse:
I had thought about that as well while writing this. I think anything references as a ng-model would already have some sort of watcher attached to it and we should ng-change instead. Before this PR, that would have been on `searchText` and `selectedItem`. With this PR, it would be `displayText` and `selectedItem`, but you'd still need the watcher for `searchText`, since that would be an internal value that's no longer directly tired to the UI.

DevVersion:
@clshortfuse At first, before reviewing, you should update your PR / commit description and explain what the intention of that change is.

> This change makes md-autocomplete matches modern browser's built-in autocomplete components.
This is an optional attribute and disabled by default.

This is some **really** unclear description. (I know there has been the title, but that's not enough)

clshortfuse:
I'll update the commit in a second, but I'm realizing that decoupling display text from input text should help fix https://github.com/angular/material/issues/8753

I'm trying to fix that as well.

clshortfuse:
blocked by PRs #8919 #8920

clshortfuse:
This also fixes #8947 without changing validation, but it may take time to review and finish.  I'm in the process of creating a design doc as well. 

I have noticed a behavior difference with master even with the attribute not set. It validates fine, but it doesn't blank `searchText` when it's invalid. This is the very reason why it fixes #8947.

Related #8977

DevVersion:
@clshortfuse As discussed on Slack, the "change" itself in #8977, is not specifically for that bug. It's kind of how the autocomplete should work as Rob said.

clshortfuse:
@DevVersion I am aware, the only issue we haven't addressed is `searchText` loses parity with `md-input-container`. I'm just exploring all options.

DevVersion:
@clshortfuse I was also explaining why that doesn't play any big role. 

For now the `input` of the autocomplete is completely internal, and you're not able to take advantage of specific things for that input. 

To be able to take "advantage" of the `md-input-container`, the issue #4492 will allow users to manually enable validation and other things.

clshortfuse:
The benefits would be explained in the design doc. Validation has it's uses, for example, searching by phone number. It's basically adding complexity and allowing Angular's built in validation to work, letting the user fine-tune at his/her discretion

DevVersion:
@clshortfuse Definitely agreed. That's what I was trying to say:

Developers are able to "use" validation after #4492

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -795,6 +851,19 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     }
   }
 
+  /**
+   * Update displayed text to currently selected item
+   */
+  function updateDisplayText () {
+    getDisplayValue(ctrl.matches[ ctrl.index ]).then(function (val) {
+      if (!val)
```
DevVersion: Add braces here

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -244,7 +246,7 @@ function MdAutocomplete ($$mdSvgRegistry) {
                   ng-minlength="inputMinlength"\
                   ng-maxlength="inputMaxlength"\
                   ng-disabled="$mdAutocompleteCtrl.isDisabled"\
-                  ng-model="$mdAutocompleteCtrl.scope.searchText"\
+                  ng-model="$mdAutocompleteCtrl.scope.displayText"\
```
DevVersion: Instead of having "display" text, it would be better to have "inputText"

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -795,6 +851,19 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     }
   }
 
+  /**
+   * Update displayed text to currently selected item
+   */
+  function updateDisplayText () {
+    getDisplayValue(ctrl.matches[ ctrl.index ]).then(function (val) {
```
DevVersion: Use `getCurrentDisplayValue`

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -681,13 +734,16 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
 
   /**
    * Clears the searchText value
+   * Clears the displayText value
+   * Clears the input value
```
DevVersion: Combine comments into a sentence, as done in other comments.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -566,7 +607,16 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+    return hasEscapeOption('blur') || !ctrl.hidden || ctrl.loading || hasEscapeOption('clear')
+           && ($scope.searchText || $scope.displayText);
+  }
+
+  /**
+   * Determines if there is text that can cleared
+   * @returns {boolean}
+   */
+  function isTextClearable() {
+    return $scope.searchText || $scope.displayText || elements.input.value;
```
DevVersion: Why is #8919 also included here?

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -566,7 +607,16 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    * @returns {boolean}
    */
   function shouldProcessEscape() {
-    return hasEscapeOption('blur') || !ctrl.hidden || ctrl.loading || hasEscapeOption('clear') && $scope.searchText;
+    return hasEscapeOption('blur') || !ctrl.hidden || ctrl.loading || hasEscapeOption('clear')
+           && ($scope.searchText || $scope.displayText);
```
DevVersion: Since you created the `isTextClearable` function, why not using it here?

src/components/chips/chips.spec.js
```diff
@@ -922,7 +922,7 @@ describe('<md-chips>', function() {
           input.val('    ');
           input.triggerHandler('input');
 
-          expect(input.controller('ngModel').$modelValue).toBe('');
+          expect(input.controller('ngModel').$modelValue).toBeFalsy();
```
clshortfuse: IE sets `input` to `undefined` now while Chrome and FireFox still evaluate to `''`

DevVersion: Please add a comment for that then.

clshortfuse:
Rebased, tested, and updated the commit message.

clshortfuse:
Closing until we can milestone this properly, and an issue is filed.

PR #8906, opened 2016-06-29
fix(datePicker): Process input on blur.

### LINKED ISSUE #8905: datePicker: model isn't updated until 500ms has passed since the last edit
**Actual Behavior**:
 -  `What is the issue?` *

My setup: I have a datePicker in a dialog, and when the dialog closes, I try to submit the data from the datePicker to my server.

If a user enters data into the datePicker and closes the dialog in less than 500ms, stale or otherwise invalid data gets sent to the server, because datePicker hasn't updated my model.

The general problem, I think, is that the debouncing in datePicker is too aggressive, and the datePicker component provides no way to force it to sync with the model. You're just stuck waiting for 500ms.

I think it's reasonable to sync to the model on blur, because the blur event is a good indication that the user has finished entering data into the input field.

 -  `Details:`

You can see the impact of the 500ms delay on the Datepicker demo at https://material.angularjs.org/latest/demo/datepicker :

1) Put your cursor in the textbox under "Standard date-picker"
2) Quickly enter a new date and press tab
3) Note that it takes 500ms before the other date pickers are synced with the first one

**Angular Versions**: *
 -  `Angular Version:`
1.5

Noted on 1.0.9

**Additional Information**:
 -  `Browser Type: *`
Chrome

 -  `Browser Version: *`
51.0.2704.106 (64-bit)

 -  `OS: *`
Ubuntu

 -  `Stack Traces:`
N/A

mkowalchuk:
Triggers input processing on the blur event so that the model can immediately be read.
    
Fixes #8905

### DIFF
--- a/src/components/datepicker/js/datepickerDirective.js
+++ b/src/components/datepicker/js/datepickerDirective.js
@@ -369,6 +369,8 @@
     // TODO(chenmike): Add ability for users to specify this interval.
     self.ngInputElement.on('input', self.$mdUtil.debounce(self.handleInputEvent,
         DEFAULT_DEBOUNCE_INTERVAL, self));
+
+    self.ngInputElement.on('blur', angular.bind(self, self.handleInputEvent));
   };
 
   /** Attach event listeners for user interaction. */


crisbeto:
I don't think this is the root of the issue. As you mention in #8905, you want the value to update immediately. We already have a TODO to add this as an option.

mkowalchuk:
The TODO is to make the debounce period configurable, but it seems like the only way to make the component behave reasonably without this change is to set it configure the debounce period to be 0ms every time (otherwise, we're just making race conditions less likely but not impossible, right?).

That said, I'm more than happy to create another pull request for making it configurable. 0ms debouncing will work fine for my project.

crisbeto:
The `blur` approach definitely works as well, but the point I was trying to make is that it might not be expected by some users. Instead if we made the debounce period configurable, we'd both solve cases like yours and keep the module consistent.

mkowalchuk:
Fair enough. Knowing that the project is currently in maintenance mode, I agree that we should avoid unexpected behavioral changes whenever possible. I'll create a patch for making it configurable and close this one out.

PR #8914, opened 2016-06-30
Fix(bottomSheet): md-bottom-sheet grid list has spacing issue within list item

### LINKED ISSUE #8913: fix(bottomSheet): md-bottom-sheet grid list has spacing issue within list item
**Actual Behavior**:
 -  `What is the issue? *` Implementing a bottom sheet with a grid list causes each list item to have a spacing issue at the top caused by a pseudo-before element.
 -  `What is the expected behavior?` There should be no extra space at the top of the list item.

 -  `Details:` This is an internal docs issue that is only visible in a development local build by contributors.

**Angular Versions**: *
 -  `Angular Version:` Latest

**Additional Information**:
 -  `Browser Type: *` Chrome, IE 11
 -  `OS: *` OS X 10.11.5, Windows 10
 -  `Stack Traces:` N/A

----

  Add more lines if needed.

bradrich:
Tested this fix within IE 11 where the addition of the pseudo-before element was meant to fix an issue with min-height and flex containers in IE11, and the pseudo-before element was causing a spacing issue of the exact height as the min-height setting of the pseudo element at the top of each bottom sheet grid list item.

Ignoring the pseudo item fixed the spacing issue and didn't break the flex layout of IE 11 due to the fact that the `md-list` is built with the same configuration as the fix described here: https://codepen.io/chriswrightdesign/pen/emQNGZ.

Fixes #8913.

### DIFF
--- a/src/components/bottomSheet/bottom-sheet.scss
+++ b/src/components/bottomSheet/bottom-sheet.scss
@@ -115,7 +115,7 @@ md-bottom-sheet {
           }
         }
       }
-      
+
       @media (max-width: $layout-breakpoint-sm) {
         @include grid-items-per-row(3, true);
       }
@@ -132,7 +132,10 @@ md-bottom-sheet {
         @include grid-items-per-row(7);
       }
 
-
+      // Override of the IE11 fix from @mixin ie11-min-height-flexbug, line 109 mixins.scss
+      &::before {
+        display: none;
+      }
 
       .md-list-item-content {
         display: flex;
--- a/.gitignore
+++ b/.gitignore
@@ -5,3 +5,4 @@ dist
 /.idea/
 .nvmrc
 tmp/
+.editorconfig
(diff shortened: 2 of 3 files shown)


src/components/bottomSheet/bottom-sheet.scss
```diff
@@ -132,7 +132,10 @@ md-bottom-sheet {
         @include grid-items-per-row(7);
       }
 
-
+      // Override of the IE11 fix from @mixin ie11-min-height-flexbug, line 109 mixins.scss
```
DevVersion: As we discussed on Slack:

```js
// We have to remove the `pseudo` fix for IE on the list-item here, because we're 
// manually changing the `flex` layout of the list-item.
```

Can you maybe add something like that to the comment (to clarify the change - please do not use my weird sample :smile: )

PR #8915, opened 2016-06-30
feat(md-virtual-repeat): Add DOM scope events for removal and addition of elements

bryanerayner:
I am building a UI where elements inside a virtual-repeater list, display pins on a map. These pins change size depending on their visibility.

I required a way to detect when the element was no longer visible in the virtual-repeat list. Hopefully this is a welcome addition to the project.

    The $mdVirtualRepeatItemDetached event is $broadcast with no arguments,
    on the scope of elements which are removed from the DOM.

    The $mdVirtualRepeatItemReattached event is $broadcast with no arguments,
    on the scope of elements which are re-attached to the DOM.

### DIFF
--- a/src/components/virtualRepeat/virtual-repeater.js
+++ b/src/components/virtualRepeat/virtual-repeater.js
@@ -767,6 +767,13 @@ VirtualRepeatController.prototype.virtualRepeatUpdate_ = function(items, oldItem
         this.blocks[maxIndex] && this.blocks[maxIndex].element[0].nextSibling);
   }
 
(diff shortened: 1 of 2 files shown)


DevVersion:
I think the change in general is reasonable. 

---
I have to discuss the necessity of this change with the Team.

Even though, thanks for the PR! @bryanerayner

topherfangio:
@bryanerayner Thanks for the PR. After team discussion, we like the idea, but we're pursuing other design approaches.

bryanerayner:
Will there be a way of knowing whether or not components are removed from the DOM in the other design approach?

PR #8919, opened 2016-06-30
fix(mdAutocomplete): check input value when determining to clear text

### LINKED ISSUE #8916: autocomplete: when min-length is specified, text clears "late" after escape
**Actual Behavior**:
input is cleared late after hitting escape

https://material.angularjs.org/HEAD/demo/autocomplete

On Floating Label:

* Focus Favorite State autocomplete
* Type `a`
* press Escape
* note that `a` isn't cleared
* Focus autocomplete
* Watch it disappear

**Angular Versions**: *
 -  `Angular Version:` master 

**Additional Information**:
 -  `Browser Type: *` chrome 
 -  `Browser Version: *` v52
 -  `OS: *` chromeos
 -  `Stack Traces:`

i'm trying to fix this in https://github.com/angular/material/pull/8904 by decoupling the displayed text with searchText.

clshortfuse:
* Add check for input value when determining to clear text

`searchText` cannot assumed to be equal to input value.
`searchText` isn't set until it passes validation (eg: min-length)

Fixes #8916

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -471,7 +471,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         event.preventDefault();
 
         clearSelectedItem();
-        if ($scope.searchText && hasEscapeOption('clear')) {
+        if (isTextClearable() && hasEscapeOption('clear')) {
           clearSearchText();
         }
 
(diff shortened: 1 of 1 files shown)


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -471,8 +471,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         event.preventDefault();
 
         clearSelectedItem();
-        if ($scope.searchText && hasEscapeOption('clear')) {
-          clearSearchText();
+        if (isTextClearable() && hasEscapeOption('clear')) {
+          clearText();
```
DevVersion: We should still stick with `clearSearchText`.

Using `clearText` against `clearSelectedValue` is unclear and also inconsistent

clshortfuse:
@DevVersion reverted name back

ThomasBurleson:
How does this relate to #8977

clshortfuse:
If #8977 is merged and validation is disabled, I believe this change  may no longer be needed.

The reason is, this change is based around ‘input.value‘ having a different value than ‘searchText‘.

Regardless, this isn't conflicting.

clshortfuse:
fixed in #8977

PR #8920, opened 2016-06-30
fix(autocomplete): improve clear and blur behavior on escape

### LINKED ISSUE #8917: autocomplete: escape always cleans text despite md-escape-options
**Actual Behavior**:
Text is cleared always.
If `md-escape-options` is set to not clear text, it shouldn't

http://codepen.io/shortfuse/pen/dXWNpy

**Angular Versions**: *
 -  `Angular Version:` 1.5.5
 -  `Angular Material Version:`  master

**Additional Information**:
 -  `Browser Type: *` Chrome 
 -  `Browser Version: *` v52
 -  `OS: *` ChromeOS

----

This is because of the input field `type='search'`. There should be a preventDefault when escape is pressed since we'll be handling that interaction, and not the browser.

clshortfuse:
* Prevent browser from using default ESCAPE key interaction for escape.
* Set ctrl.hidden variable on escape

Some browsers clear any input with type search on ESCAPE keypress.
ctrl.hidden variable was not correctly reflecting dropdown hidden state

Fixes #8917

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -482,21 +482,22 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         select(ctrl.index);
         break;
       case $mdConstant.KEY_CODE.ESCAPE:
+        event.preventDefault(); // Prevent browser from always clearing input
         if (!shouldProcessEscape()) return;
         event.stopPropagation();
-        event.preventDefault();
 
         clearSelectedItem();
         if ($scope.searchText && hasEscapeOption('clear')) {
           clearSearchText();
         }
 
+        // Manually hide (needed for mdNotFound support)
+        // doBlur will not update variable, so we will update it manually
+        ctrl.hidden = true;
+
         if (hasEscapeOption('blur')) {
           // Force the component to blur if they hit escape
           doBlur(true);
-        } else {
-          // Manually hide (needed for mdNotFound support)
-          ctrl.hidden = true;
         }
 
         break;
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -400,48 +400,67 @@ describe('<md-autocomplete>', function() {
                   placeholder="placeholder">\
                 <span md-highlight-text="searchText">{{item.display}}</span>\
               </md-autocomplete>';
-      beforeEach( inject(function($timeout) {
+      beforeEach( inject(function($timeout, $material) {
         scope = createScope();
         element = compile(template, scope);
         ctrl = element.controller('mdAutocomplete');
 
+        $material.flushInterimElement();
+
+        // Update the scope
+        element.scope().searchText = 'fo';
+        waitForVirtualRepeat(element);
+
         // Focus the input
         ctrl.focus();
         $timeout.flush();
-        expect(scope.searchText).toBe('');
 
-        scope.$apply('searchText = "test"');
+        expect(ctrl.hidden).toBe(false);
 
-        expect(scope.searchText).toBe('test');
+        expect(scope.searchText).toBe('fo');
 
+        waitForVirtualRepeat(element);
         $timeout.flush();
+        expect(ctrl.hidden).toBe(false);
       }));
 
       afterEach(function() { element.remove() });
-      it('does not clear the value nor blur when hitting escape', inject(function($mdConstant, $document) {
+      it('does not clear the value nor blur when hitting escape', inject(function($mdConstant, $document, $timeout) {
         scope.$apply('escapeOptions = "none"');
         scope.$apply(function() {
           ctrl.keydown(keydownEvent($mdConstant.KEY_CODE.ESCAPE));
+          $timeout.flush();
+          expect(ctrl.hidden).toBe(true);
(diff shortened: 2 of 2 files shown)


DevVersion:
@ThomasBurleson LGTM

--
FYI: This change makes sense, because we're handling the `clearSearchText` manually with specific conditions.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -466,9 +466,9 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         select(ctrl.index);
         break;
       case $mdConstant.KEY_CODE.ESCAPE:
+        event.preventDefault(); // Prevent browser from always clearing input
```
topherfangio: When I first saw this, I assumed that it would always prevent the clearing of the autocomplete, but that doesn't seem to be the case when I actually tested it. Under what conditions does this prevent clearing?

clshortfuse: `preventDefault` stops the default browser event
`stopPropagation` stops the event from bubbling up the DOM tree

Some browsers (ie: Chrome), on `input` elements with `type` set to `search` will clear the `value`. Since we're handling clearing selectively, we're not letting that event pass to the browser.

We are letting ESC to propagate up the DOM tree so other elements can handle it. For example, we want dialogs to close on ESC keydown.

DevVersion: @topherfangio Yeah, you're right - it seems like Firefox doesn't have that behavior. It's showing up in Chrome.

Actually when having a `input[type="search"]`, the input will automatically clear its value on escape key.

It also strips line-breaks (See [here](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input))
```
line-breaks are automatically removed from the input value."
```

---
@clshortfuse Can you find a better solution?

topherfangio: I think I'm confused now 😕 

@clshortfuse So, you are saying that before this fix, there were some cases where the esc key was clearing the input when we didn't want it to. Can you give me an example?

@DevVersion What was I right about? I was stating that my initial thinking was wrong/incorrect 😄

clshortfuse: @topherfangio Yep, the example is in the fixed issue.

There's an older, merged PR that added the ability to instruct ’mdAutocomplete‘ to not clear when you press ESC. The option is ’md-escape-options’. Either Chrome changed their default behavior or somebody changed the ’input’ element type to ’search’ which Chrome clears by default on ESC key down.

topherfangio: @clshortfuse Okay, thanks for the clarification.

Quick note: the esc key still blurs the input even with `md-escape-options="none"`. I would assume that `none` would specifically disable the blurring too, but I honestly don't know how it was implemented originally.

Can you take a quick look at that and see if it's related to this issue?

clshortfuse: I really don't think it should blur if the option is off. I'll have to check it later.

Regardless, I think mdAutocomplete's escape key logic needs work. Also, the input type being search is a forces certain logic instead of being optional.  This isn't tied to just the escape button.

I would argue that input type being search shouldn't be the default and instead it should be text. All of our demos are based around using autocomplete forms, not as a search tool. In fact, we've never written support for the search key which appears on mobile devices. Pressing it on a mdAutocomplete element does nothing and would only confuse users.

Regardless, the change here in this PR is correct, regardless of what we decide to do with the ESCAPE key. We don't need the internal's browser's help to clear text. We handle that ourselves.

topherfangio: That's a fair point. Any chance you can create a test to go along with this?

clshortfuse: I figured out why the tests are passing as is. Typing something would trigger the autocomplete window. Pressing escape closes it and wouldn't clear the text. That's because, with the pop-up open, ‘shouldProcessEscape‘ passes.

I should do a SECOND escape press which would cause ‘shouldProcessEscape‘ to return false (popup is closed) and then the default browser action would kick in.

I'll add the test to this PR, where it should pass and master should fail.

clshortfuse:
@topherfangio I found a bug with `blur` that I fixed where it wouldn't set `ctrl.hidden` to true. I added more spec tests. It seems more robust. PhantomJS doesn't seem to clear search fields on escape, but I believe Firefox does. I've added 2 escape presses in the spec tests. Master would fail here. 

If the travis Firefox tests pass, then we're good.

ThomasBurleson:
How does this relate to #8977 ?

clshortfuse:
@ThomasBurleson, that PR is unrelated. This PR is related to escape keypress. The other is related to md-input-maxlength

topherfangio:
LGTM 👍 

@DevVersion Can you please review as well?

DevVersion:
@topherfangio I forgot to write "LGTM" - I had a discussion with @clshortfuse regarding the move of the `hidden` flag. But we came to the decision, that this is necessary.

@clshortfuse Probably update the comment again, to explain what we archived.

clshortfuse:
@DevVersion I've updated the commit message with

>ctrl.hidden variable was not correctly reflecting dropdown hidden state

To expand, this fixes a regression caused by https://github.com/angular/material/commit/a79fe0e5c368792766034888aba5e84d0841039c

PR #8921, opened 2016-06-30
feat(datePicker): Add a configuration option for the debounce interval.

### LINKED ISSUE #8905: datePicker: model isn't updated until 500ms has passed since the last edit

mkowalchuk:
Fixes #8905

### DIFF
--- a/src/components/datepicker/js/datepickerDirective.js
+++ b/src/components/datepicker/js/datepickerDirective.js
@@ -45,6 +45,7 @@
    * * `required`: whether a required date is not set.
    * * `mindate`: whether the selected date is before the minimum allowed date.
    * * `maxdate`: whether the selected date is after the maximum allowed date.
+   * * `debounceInterval`: ms to delay input processing (since last debounce reset); default value 500ms
    *
    * @usage
    * <hljs lang="html">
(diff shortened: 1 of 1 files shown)


crisbeto:
LGTM @ThomasBurleson.

clshortfuse:
Functionality-wise, it looks good. My only concern is the inconsistency of the variable name with other elements. If I'm not mistaken, mdAutocomplete has something similar, but there it's called ‘wait‘.

crisbeto:
The option in autocomplete is called `delay`, but I think it's different. The one in autocomplete seems to be the delay before results start showing up, the changes in this PR are for the debounce period before the model is updated.

clshortfuse:
👍

PR #8931, opened 2016-07-01
chore(build): module compilation should concatenate multiple theme files

### LINKED ISSUE #8892: Bower build not including all theme files
The builds in the [bower-material repo](https://github.com/angular/bower-material) don't include some of the theme files. For example, the datepicker has both a datepicker and a calendar theme, however the build only generates the datepicker theme.

DevVersion:
* The `build-all-modules` task for `bower-material` did not correctly build the theming stylesheet, if the actual module contains multiple theme stylesheets.

> For example, the `md-datepicker` has two theme stylesheets (`calendar` and `datepicker`).

> - Those files will be renamed to the default theme file name.<br/>
This would mean that all multiple theme stylesheets will retrieve the same filename in the stream.<br/>
We should concatenate all default theme files inside of the stream.

> Here is a little example of the duplicate filenames inside of the stream.
image

---
* The core module in bower-material should also build its theme styles in an extra file. Currently the core module always includes its theming stylesheet in the normal CSS file. This is not correct.

> This is not correct, because the theming service is not able to build the theme from the normal stylesheet.

---
* Moves the checkbox primary mixin, used by the checkbox and select component, to the mixin SCSS, because otherwise we would have this mixin the core modules theme stylesheet which is not correct.

> This is necessary to be able to build each module separately.<br/>
> Otherwise we have to prepend the `themes.scss` to each module compilation manually, to make sure that the checkbox / select module builds properly.

> Also the `theming.scss` will always generate the body selector again, which introduces unnecessary selectors.

---
* Fixed the wrong path pattern for the Gulp Source Glob. The path and pattern syntax were only valid for _old_ versions.

> This causes the detection for themes inside of the `core` module not to work.
Fixes #8892.

cc. @ThomasBurleson @crisbeto


crisbeto:
LGTM @ThomasBurleson

Issue #8933, opened 2016-07-02
Please separate drag and swipe event


Thaina:
I want to use mdgesture to register drag event on some element (canvas)

But it has logic of swipe, only start when the pointer move over some threshold in only one direction. If I select horizontal it cannot start drag event in vertical dragging

Which limit its used case only on carousel. It cannot use to normal movement in arbitrary direction (move a map for example)

Please has an option to let us can use a very basic drag event, no threshold and lock direction

DevVersion:
Related Pull Request: #8934

Thaina:
@DevVersion I have made that pull request such a hack and make things a little complicate. It just fastest way to do so

But I would like if you consider to make a new mdDrag with very basic drag event with no direction or threshold, then have mdSwipe extend it instead

ThomasBurleson:
Will not fix in ngMaterial v1.x: deprecated.

Thaina:
@ThomasBurleson I understand but will it be fixed in ngMaterial 2 ?

DevVersion:
@Thaina In Material 2 there will be no standalone gestures support. Material 2 provides a HammerJS gesture configuration, which allows you to use extremely much gesture events. 

> http://hammerjs.github.io/

Issue #8879, opened 2016-06-28
Is it possible to support drag event


Thaina:
There are only swipe in 4 direction possible. I actually want a drag event that user may touch or click and drag on fixed element, not move it but drag on it, for control canvas and so on

It should fire event start / end / dragging and give information on start position, current position, and offset from the element it start on

Is it possible on angular material to implement this?

DevVersion:
Angular Material has its own Gesture Service, which has support for dragging.

```js
$mdGesture.register(myElement, 'drag');

myElement
  .on('$md.dragstart', onDragStart)
  .on('$md.drag', onDrag)
  .on('$md.dragend', onDragEnd);

function onDrag(event) {
  // Retrieves the dragged delta X.
  let pointer = event.pointer.distanceX;
}
```

See some examples, in the `md-switch` or `md-bottom-sheet`

Thaina:
Thanks. But, I don't know why, it cannot work when try to drag up/down before sideway

https://plnkr.co/edit/aoPoaqIddowhix7733J3?p=preview

If I try to start dragging up first it like event is not be called at all

Thaina:
I was figure it out that it related to `horizontal` property that has only true or false. If I set to true it cannot start from vertical drag and if it false it cannot start from horizontal drag

Thaina:
Should this code here

https://github.com/angular/material/blob/bd1cce42ba16b733af0163011f8bb2429782b5da/src/core/services/gesture/gesture.js#L275

changed to be this

```js
if (this.state.options.horizontal) {
            shouldStartDrag = Math.abs(pointer.distanceX) > this.state.options.minDistance;
            shouldCancel = Math.abs(pointer.distanceY) > this.state.options.minDistance * this.state.options.cancelMultiplier;
}
if (!this.state.options.horizontal || this.state.options.vertical)
{
            shouldStartDrag = Math.abs(pointer.distanceY) > this.state.options.minDistance;
            shouldCancel = Math.abs(pointer.distanceX) > this.state.options.minDistance * this.state.options.cancelMultiplier;
}
```

Thaina:
#8934

PR #8934, opened 2016-07-03
Allow support both horizontal and vertical swipe

Thaina:


Thaina:
I have already sign that. Is it signed?

Thaina:
#8879

EladBezalel:
Horizontal and Vertical swipes are already [working](https://material.angularjs.org/HEAD/demo/swipe)

Thaina:
@EladBezalel It cannot support both in the same element at the same time. We cannot start swipe in vertical by default. And we cannot start swipe in horizontal if we set it as vertical

Simply put, It has no basic drag functionality

DevVersion:
Closing as in comment https://github.com/angular/material/issues/8933#issuecomment-234987351

PR #8953, opened 2016-07-05
fix(theming): potentially generating invalid CSS

crisbeto:
If a theme that ends in a newline gets passed to the theming provider, it will generate a CSS syntax error: a double closing bracket would be added, instead of a single one (e.g. `}}` instead of `}`).


DevVersion:
@ThomasBurleson LGTM.

EladBezalel:
LGTM

PR #8961, opened 2016-07-06
fix(autocomplete): autocompletes input should not clear the view value.

### LINKED ISSUE #8947: md-autocomplete with a md-input-maxlength and md-min-length 0 make all the items appear when maxlength is exceeded
**Actual Behavior**:
 -  `What is the issue?` When I have a md-autocomplete with a md-input-maxlength and a md-min-length 0 and I exceed the maxlength, the field become in error (This is what I want) but probably because the view value is now empty (0), all the values appear.
 -  `What is the expected behavior?` No values should appear when then maxlength is exceeded.

 -  `CodePen Demo which shows your issue:` (http://codepen.io/kcote/pen/yJMdpR)
 -  `Details:` Enter letters to exceed the maxlength (18) and then all the values are showing.

**Angular Versions**: *
 -  `Angular Version:1.5.6`
 -  `Angular Material Version:1.0.9`

**Additional Information**:
 -  `Browser Version: 51.0.2704.103 (64 bit)`
 -  `OS: Windows 10`
 -  `Stack Traces:`

DevVersion:
* Currently when specifying attributes like `md-input-maxlength` and the limit as reached, the ngModelCtrl will clear the `$viewValue`, which causes the
  input to keep the entered text, but the autocomplete already knows that the searchText is empty.

- This causes the dropdown to open, even if the input "should" not show up with all results.
- Also the input value will be cleared, which is not correct.

Fixes #8947.

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -320,6 +320,75 @@ describe('<md-autocomplete>', function() {
       element.remove();
     }));
 
+    describe('md-input-maxlength', function() {
+
+      it('should correctly set the form to invalid', inject(function($timeout) {
+        var scope = createScope(null, {inputId: 'custom-input-id'});
+        var template =
+          '<form name="testForm">' +
+            '<md-autocomplete ' +
+                'md-input-id="{{inputId}}" ' +
+                'md-input-maxlength="2" ' +
+                'md-input-name="testAutocomplete" ' +
+                'md-selected-item="selectedItem" ' +
+                'md-search-text="searchText" ' +
+                'md-items="item in match(searchText)" ' +
+                'md-item-text="item.display" ' +
+                'tabindex="3"' +
+                'md-floating-label="Favorite state">' +
+              '<span md-highlight-text="searchText">{{item.display}}</span>' +
+            '</md-autocomplete>' +
+          '</form>';
+
+        var element = compile(template, scope);
+        var input = element.find('input');
+
+        expect(scope.searchText).toBe('');
+        expect(scope.testForm.$valid).toBe(true);
+
+        scope.$apply('searchText = "Exceeded"');
+
+        expect(scope.testForm.$valid).toBe(false);
+
+        element.remove();
+      }));
+
+      it('should not clear the view value if the input is invalid', inject(function($timeout) {
+        var scope = createScope(null, {inputId: 'custom-input-id'});
+        var template =
+          '<form name="testForm">' +
+            '<md-autocomplete ' +
+                'md-input-id="{{inputId}}" ' +
+                'md-input-maxlength="2" ' +
+                'md-input-name="testAutocomplete" ' +
+                'md-selected-item="selectedItem" ' +
+                'md-search-text="searchText" ' +
+                'md-items="item in match(searchText)" ' +
+                'md-item-text="item.display" ' +
+                'tabindex="3"' +
+                'md-floating-label="Favorite state">' +
+              '<span md-highlight-text="searchText">{{item.display}}</span>' +
+            '</md-autocomplete>' +
+          '</form>';
+
+        var element = compile(template, scope);
+        var input = element.find('input');
+
+        expect(scope.searchText).toBe('');
+        expect(scope.testForm.$valid).toBe(true);
+
+        input.val('Exceeded');
+        input.triggerHandler('change');
+        scope.$digest();
+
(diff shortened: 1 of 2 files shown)


clshortfuse:
Wouldn't this a breaking change? Invalid values would be allowed to be part of of `searchText` which would have adverse consequences on existing code.

DevVersion:
@clshortfuse It shouldn't be. Since this is kind of a major bug for the autocomplete, we have to treat this as a bug fix, and not as a breaking change.

It just guarantees that the autocompletes input won't be cleared unexpectedly.

This change should not affect any existing code - also the autocompletes input is not able to accept any custom validation (yet), so it shouldn't break any existing (supported) codebase yet.

clshortfuse:
Wouldn't searchTextChange fire with invalid options? I've also noticed there are no other min-length checks other than it being 0

DevVersion:
`searchTextChange` will fire, when the search text changes, regardless of the input is invalid or not. Not sure what you mean with other checks. This may be another topic, which isn't affiliated with that change?

clshortfuse:
To expand, let's pretend I have an mdAutocomplete with cache enabled. I use searchTextChange to track what texts are BeIn searched and log each search. I have a min of 3 and max of 20.

Since I use cache, I can't reliable track with the algorithm I use to perform searches. I would depend on searchText always being a validated option.

I would depend on `searchText` blanking always between 3 and 20. Wouldn't this break that?

clshortfuse:
The comment about min check is I have a min length of 5 and input 'banana' into the input field, the searchTextChange won't fire until I reach ‘banan‘. This seems to cause it to fire on ‘b‘, ‘ba‘, ‘ban‘, ‘bana‘ as well. If we had a check for min-length as something other than 0, this PR would fail that.

DevVersion:
The point of supporting those two attributes (`minlength` and `maxlength`) here is to mark the form as invalid if those `$validators` are failing.

This allows developers to show the given `ng-messages`, with the proper `mdInputContainer` message animations.

Normally the autocomplete will not allow developers to forward custom validators, so it's kind of a special case, that we support those two. More will probably follow with #4492.

image

The user / developer cannot expect the `searchText` to only run, when the searchText is valid.
The `change` event will be always firing, once the `searchText` changes.

clshortfuse:
Well, this literally breaks some of my personal code where I expect the functions to only trigger when between min and max length. Considering I have tied an mdAutocomplete to a service where i pay per query / have a query limit, upgrading without knowing it's a major breaking change literally costs me actual money, so I'm going to just leave it at, I disapprove of this PR.

DevVersion:
@clshortfuse Well, pretty interesting.. We actually never had that kind of a "disapproving". Our Team Chat on Slack is just for something like that. If you have something to discuss, just ping me, and ask me your questions.

> have a query limit, upgrading without knowing it's a major breaking change literally costs

Regarding that comment, if you're really "considering" that as a Breaking Change, then we can simply add that to our changelog (See below)

> I expect the functions to only trigger when between min and max length

I'm just wondering, where this is stated? The documentation is pretty clear in that case, and mentions that the `searchText `always changes.

There will be pretty much different edge-cases, which expect other things (that's why the doc is pretty clear)
- **User 1**: Expects the `searchText` announcement to trigger always, because he want to update another binding especially for the `searchText`
- **User 2**: Expects the `searchText` announcement to trigger only if the input is valid.

That's kind of a conflict, between the two users, because both are expecting something different. 
To solve that case, we could easily enhance the `attribute` documentation a bit more, so there won't appear any further questions.

Also the `md-autocomplete` should not clear the input text, if the input turns invalid, this is currently causing a major issue, where the autocompletes input text still shows the text, but the actual `$modelValue` stays the same.

clshortfuse:
I think you understand what I mean about validation. I agree there is a valid issue, but I see breaking changes.

I will correct my statement since I assumed that `searchTextChange` only passed validated values with md-min-length set to something other than 0 but that's not true either. It will pass invalid text if below md-min-length. 

My comment about disapproval is more along the lines of "I see something wrong with it, I'll let the team lead ultimately decide".  Please don't take any offense to it. I just mean that as is, I see breaking changing, that can be reworked to not break. You should probably add on code for queries not to perform searches when md-max-length is reached, because as you addressed, this didn't happen before.

DevVersion:
FYI: This PR will be resubmitted, because we want to start the conversation from scratch again. 

Due to conversation with @robertmesserle and @clshortfuse , we decided to _not_ address Carlos' concerns.

PR #8977, opened 2016-07-08
fix(autocomplete): autocompletes input should not clear the view value.

### LINKED ISSUE #8947: md-autocomplete with a md-input-maxlength and md-min-length 0 make all the items appear when maxlength is exceeded

DevVersion:
* Currently when specifying attributes like `md-input-maxlength` and the limit as reached, the ngModelCtrl will clear the `$viewValue`, which causes the
  input to keep the entered text, but the autocomplete already knows that the searchText is empty.

- This causes the dropdown to open, even if the input "should" not show up with all results.
- Also the input value will be cleared, which is not correct.

--- 
@ThomasBurleson Please read this before reviewing:

- Carlos had a concern, that the autocomplete's behavior will break, since the `autocomplete` did only emit the `searchTextChange` event if it's valid.

- I had a discussion with @robertmesserle about the concerns of @clshortfuse.
  We both agreed that the documentation mentions clearly that the `searchTextChange` will be emitted always (regardless of validity of the searchText)

  The developers should be able to always track the `searchText`.

- Another concern was, that the `autocomplete` currently only queries for results, if the `searchText` is valid (this is not mentioned / stated in the documentation).
 
  The autocomplete should always query for the results (regardless of the validity of the `searchText`).
  Developers can easily reduce the amount of queries, by just modifying their `queryFn`.

- Something to consider is, whether this change should have any special announcement in the `changelog` or not.

Fixes #8947.

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -320,6 +320,75 @@ describe('<md-autocomplete>', function() {
       element.remove();
     }));
 
+    describe('md-input-maxlength', function() {
+
+      it('should correctly set the form to invalid', inject(function($timeout) {
+        var scope = createScope(null, {inputId: 'custom-input-id'});
+        var template =
+          '<form name="testForm">' +
+            '<md-autocomplete ' +
+                'md-input-id="{{inputId}}" ' +
+                'md-input-maxlength="2" ' +
+                'md-input-name="testAutocomplete" ' +
+                'md-selected-item="selectedItem" ' +
+                'md-search-text="searchText" ' +
+                'md-items="item in match(searchText)" ' +
+                'md-item-text="item.display" ' +
+                'tabindex="3"' +
+                'md-floating-label="Favorite state">' +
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@DevVersion - sounds like a breaking change. Thx.

PR #8979, opened 2016-07-08
docs: create retina-ready favicon

soul-wish:
- update existing favicon with a new one
- now it looks much better on retina display

screenshot (before: at the left, after: at the right):

material-favicon


soul-wish:
I signed it ;)

PR #8983, opened 2016-07-08
fix(mdPanel): Propagation, CSS targeting, and dynamic position updating

### LINKED ISSUE #8968: mdPanel: Implement "propagateContainerClicks" config option.
This option should allow users to interact with other elements on the page when the panel is opened.

If `clickOutsideToClose` is also present, the panel should close itself and then click on the underlying element.

If `hasBackdrop` is true, this option should be ignored as the backdrop will visually indicate that clicks will stop at that barrier.

This is currently not possible because the panel container covers the entire screen and captures any click events.

This feature is needed for tooltips and menus.

### LINKED ISSUE #8980: fix(mdPanel): Dynamic update of panel position and CSS class manipulation
**Actual Behavior**:
 -  `What is the issue? *` **This is actually a two-part issue**
  - **Part 1:** There is no current way to update the position of the mdPanel that is originally established in the configuration, without closing the mdPanel, changing the position and re-opening the mdPanel.
  - **Part 2:** The `addClass`, `removeClass`, and `toggleClass` functions of the `MdPanelRef` only target the mdPanel container, the 'outer-wrapper', even though a developer might want to target the mdPanel element instead.

  - **Part 1:** There should be a public function of the mdPanel API that allows you to update the position object in the current panel's configuration and then it should update the position of the mdPanel.
  - **Part 2:** There should be an option on the `addClass`, `removeClass`, and `toggleClass` functions that allow you target the mdPanel element instead of the container.
  - **Part 2 (EDIT):** **An example of why a developer would want to use** the `addClass`, `removeClass`, and `toggleClass` modifications to specifically target the panel element:
    - Targeting the panel element is useful if you want to add CSS classes to the panel element for custom transitions or behavior. For instance, the transition between showing and hiding the panel with an animation that is not supported by the mdPanelAnimation API.

 -  `Details:` N/A

**Angular Versions**: *
 -  `Angular Version:` Latest

**Additional Information**:
 -  `Browser Type: *` All
 -  `Browser Version: *` All
 -  `OS: *` All
 -  `Stack Traces:` N/A

bradrich:
Fixes #8968 by:
 - Adding the `propagateContainerEvents` option to the panel configuration that is defaulted to false. If true, the mdPanel API will allow all touch and pointer events to propagate through the mdPanel container, the 'outer-wrapper', to the elements behind it.

Fixes #8980 by:
- Adding a boolean parameter to the MdPanelRef `addClass`, `removeClass`, and `toggleClass` functions that defaults to false. When true, it will allow the function's primary class actions to be executed on the mdPanel element, instead of the container.
- Adding a public `updatePosition` function that will take in a MdPanelPosition object parameter that it will overwrite the current configuration's position and then call the private `_updatePosition` function.

Ping @ErinCoughlan @EladBezalel

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -108,6 +108,9 @@ angular
  *   - `attachTo` - `{(string|!angular.JQLite|!Element)=}`: The element to
  *     attach the panel to. Defaults to appending to the root element of the
  *     application.
+ *   - `propagateContainerEvents` - `{boolean=}`: Whether pointer or touch
+ *     events should be allowed to propagate 'go through' the container, aka the
+ *     wrapper, of the panel. Defaults to false.
  *   - `panelClass` - `{string=}`: A css class to apply to the panel element.
  *     This class should define any borders, box-shadow, etc. for the panel.
  *   - `zIndex` - `{number=}`: The z-index to place the panel at.
@@ -308,6 +311,8 @@ angular
  * Adds a class to the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} newClass Class to be added.
+ * @param {boolean} toElement Whether or not to add the class to the panel
+ *    element instead of the container.
  */
 
 /**
@@ -317,6 +322,8 @@ angular
  * Removes a class from the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} oldClass Class to be removed.
+ * @param {boolean} fromElement Whether or not to remove the class from the
+ * panel element instead of the container.
  */
 
 /**
@@ -326,6 +333,17 @@ angular
  * Toggles a class on the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} toggleClass Class to be toggled.
+ * @param {boolean} onElement Whether or not to remove the class from the panel
+ *    element instead of the container.
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#updatePosition
+ * @description
+ * Updates the position configuration of a panel. Use this to update the
+ * position of a panel that is open, without having to close and re-open the
+ * panel.
  */
 
 /**
@@ -628,6 +646,7 @@ function MdPanelService($rootElement, $rootScope, $injector, $window) {
     focusOnOpen: true,
     fullscreen: false,
     hasBackdrop: false,
+    propagateContainerEvents: false,
     transformTemplate: angular.bind(this, this._wrapTemplate),
     trapFocus: false,
     zIndex: defaultZIndex
@@ -1057,14 +1076,18 @@ MdPanelRef.prototype.hide = function() {
  * Add a class to the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} newClass Class to be added.
+ * @param {boolean} toElement Whether or not to add the class to the panel
+ *    element instead of the container.
  */
-MdPanelRef.prototype.addClass = function(newClass) {
+MdPanelRef.prototype.addClass = function(newClass, toElement) {
   if (!this._panelContainer) {
     throw new Error('Panel does not exist yet. Call open() or attach().');
   }
 
-  if (!this._panelContainer.hasClass(newClass)) {
+  if (!toElement && !this._panelContainer.hasClass(newClass)) {
     this._panelContainer.addClass(newClass);
+  } else if (toElement && !this._panelEl.hasClass(newClass)) {
+    this._panelEl.addClass(newClass);
   }
 };
 
@@ -1073,14 +1096,18 @@ MdPanelRef.prototype.addClass = function(newClass) {
  * Remove a class from the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} oldClass Class to be removed.
+ * @param {boolean} fromElement Whether or not to remove the class from the
+ *    panel element instead of the container.
  */
-MdPanelRef.prototype.removeClass = function(oldClass) {
+MdPanelRef.prototype.removeClass = function(oldClass, fromElement) {
   if (!this._panelContainer) {
     throw new Error('Panel does not exist yet. Call open() or attach().');
   }
 
-  if (this._panelContainer.hasClass(oldClass)) {
+  if (!fromElement && this._panelContainer.hasClass(oldClass)) {
     this._panelContainer.removeClass(oldClass);
+  } else if (fromElement && this._panelEl.hasClass(oldClass)) {
+    this._panelEl.removeClass(oldClass);
   }
 };
 
@@ -1089,13 +1116,19 @@ MdPanelRef.prototype.removeClass = function(oldClass) {
  * Toggle a class on the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} toggleClass The class to toggle.
+ * @param {boolean} onElement Whether or not to toggle the class on the panel
+ *    element instead of the container.
  */
-MdPanelRef.prototype.toggleClass = function(toggleClass) {
+MdPanelRef.prototype.toggleClass = function(toggleClass, onElement) {
   if (!this._panelContainer) {
     throw new Error('Panel does not exist yet. Call open() or attach().');
   }
 
-  this._panelContainer.toggleClass(toggleClass);
+  if (!onElement) {
+    this._panelContainer.toggleClass(toggleClass);
+  } else {
+    this._panelEl.toggleClass(toggleClass);
+  }
 };
 
 
@@ -1131,9 +1164,14 @@ MdPanelRef.prototype._createPanel = function() {
           self._panelEl = angular.element(
               self._panelContainer[0].querySelector('.md-panel'));
 
-          // Add a custom CSS class.
+          // Add a custom CSS class to the panel element.
           if (self._config['panelClass']) {
             self._panelEl.addClass(self._config['panelClass']);
+          }
+
+          // Handle click and touch events for the panel container.
+          if (self._config['propagateContainerEvents']) {
+            self._panelContainer.css('pointer-events', 'none');
           }
 
           // Panel may be outside the $rootElement, tell ngAnimate to animate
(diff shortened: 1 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -1188,6 +1211,24 @@ MdPanelRef.prototype._addStyles = function() {
 
 
 /**
+ * Update the position configuration of a panel
```
ErinCoughlan: nit: Method descriptions should start with a sentence written in the third person declarative voice. Can you change this to:

"Updates the position configuration of the panel."

src/components/panel/panel.js
```diff
@@ -1188,6 +1211,24 @@ MdPanelRef.prototype._addStyles = function() {
...
+ * @param {MdPanelPosition} position
+ */
+MdPanelRef.prototype.updatePosition = function(position) {
```
ErinCoughlan: You need to add this documentation also to the public documentation above, probably with a description of why you someone would use this. (~line 330)

src/components/panel/panel.js
```diff
@@ -1086,13 +1098,19 @@ MdPanelRef.prototype.removeClass = function(oldClass) {
  * Toggle a class on the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} toggleClass The class to toggle.
+ * @param {boolean} onElement Whether or not to toggle the class on the panel
```
ErinCoughlan: New params need to be added to the public documentation at the top of this file. (For add, remove, and toggle)

bradrich:
@ErinCoughlan Thank you for these comments. I will make your suggested changes.

EladBezalel:
IMO we should remove addClass, removeClass and toggleClass by providing the container and the panel elements as a public api. 
Doesn't makes sense copying this functionality from jqlite. 
@ErinCoughlan

ErinCoughlan:
@EladBezalel That's fine by me.

bradrich:
@topherfangio Squashed my commits on this, even though it needs work.

topherfangio:
@bradrich Awesome; thanks! In general, if you push anything to a PR (even in-progress), you should squash your commits 😄

ErinCoughlan:
@bradrich Tests? You added new functionality, but I have no idea if they work and will continue to work.

bradrich:
@ErinCoughlan All tests are included and passing. Please review and give LGTM.

src/components/panel/panel.spec.js
```diff
@@ -1115,6 +1212,63 @@ describe('$mdPanel', function() {
       mdPanelPosition = $mdPanel.newPanelPosition();
     });
 
+    it('should update the position of an open panel', function() {
```
bradrich: @ErinCoughlan Please give special attention to this test.

src/components/panel/panel.spec.js
```diff
@@ -868,6 +895,76 @@ describe('$mdPanel', function() {
       });
     });
 
+    describe('CSS class logic:', function() {
+      it('should add a class to the container/wrapper when toElement=false',
+        function() {
+        openPanel(DEFAULT_CONFIG);
```
ErinCoughlan: nit: indentation here is off. openPanel and the rest of the code still needs to be indented under function, like this:
```
it('should add a class to the container/wrapper when toElement=false',
  function() {
    openPanel(DEFAULT_CONFIG);
    ...
    expect(PANEL_WRAPPER_CLASS).toHaveCLASS('my-class');
  });

File: src/components/panel/panel.spec.js [hunk 6]
```diff
@@ -868,6 +895,76 @@ describe('$mdPanel', function() {
...
+
+        panelRef.addClass('my-class');
+
+        expect(PANEL_WRAPPER_CLASS).toHaveClass('my-class');
```
ErinCoughlan: Let's add another expect here to be safe. Technically, the class can be add to the wrapper AND element and still pass all of these tests. Same with the below tests.

```
expect(PANEL_WRAPPER_CLASS).toHaveClass('my-class');'
expect(PANEL_EL).not.toHaveClass('my-class');
```

src/components/panel/panel.spec.js
```diff
@@ -1115,6 +1212,62 @@ describe('$mdPanel', function() {
       mdPanelPosition = $mdPanel.newPanelPosition();
     });
 
+    it('should update the position of an open panel', function() {
+      var position = null;
+
+      // Starting off with a absolute positioned panel at the top of the
+      // window.
+      testAbsTop();
```
ErinCoughlan: Each one of these should be a different test. It's hard to follow what the before and after states of each test is with this format. In tests, you want to avoid someone reviewing it having to think. I should be able to look at the test and say "Okay, there's the setup. Some action was taken. And the values changed correctly. Done."

Rarely should you need comments to explain what is happening. That's generally means the test can be better organized (functions that run asserts are generally bad).

Possibly format like this:

```
describe('should update the position of an open panel', function() {
  it('between two relative positions', function() { 
    // Create a relatively positioned panel.
    // Verify the position is initially right.
    // Call updatePosition with new relative position object.
    // Verify the position was updated.
  });

  it('between two absolute positions', function() {
    // Create an absolutely positioned panel.
    // Verify the position is initially right.
    // Call updatePosition with new absolute position object.
    // Verify the position was updated.
  });

  it('from absolute positioning to relative positioning', function() { ... });
});
```

bradrich:
@ErinCoughlan That's it! Whew!

src/components/panel/panel.js
```diff
@@ -1194,6 +1232,24 @@ MdPanelRef.prototype._addStyles = function() {
...
+ */
+MdPanelRef.prototype.updatePosition = function(position) {
+  if (!this._panelContainer) {
+    throw new Error('Panel does not exist yet. Call open() or attach().');
+  }
+
+  var self = this;
+
+  self._config['position'] = position;
+  self._$mdUtil.nextTick(function() {
```
ErinCoughlan: Does this work?

```self._$mdUtil.nextTick(self._updatePosition);```

Comment (DevVersion) [inline src/components/panel/panel.js]: I guess it would work if you bind it to `this`, since `updatePosition` accessed the instance context?

Comment (bradrich) [inline src/components/panel/panel.js]: It does not. I have new tests coming in.

**Comment by bradrich** [comment]
@EladBezalel I am going to add the panel wrapper/container and element to the panelRef in another feature. So, I am going to remove the needs: work label.

**Comment by bradrich** [comment]
@ThomasBurleson There is an issue with the Travis CI. Is it known?

**Comment by DevVersion** [comment]
@bradrich This rarely happens to builds on Saucelabs. I will try to increase the Sauce timeout, so we reduce that flakiness a bit.

File: src/components/panel/panel.js [hunk 9]
```diff
@@ -1194,6 +1232,22 @@ MdPanelRef.prototype._addStyles = function() {
...
 /**
+ * Updates the position configuration of a panel
+ * @param {MdPanelPosition} position
+ */
+MdPanelRef.prototype.updatePosition = function(position) {
+  if (!this._panelContainer) {
+    throw new Error('Panel does not exist yet. Call open() or attach().');
+  }
+
+  var self = this;
```
ErinCoughlan: I think the self variable here is no longer necessary and you can use `this` for both lines below.

src/components/panel/panel.spec.js
```diff
@@ -409,6 +409,34 @@ describe('$mdPanel', function() {
       });
     });
 
+    describe('should cause the propagation of events', function() {
+      var config, wrapper;
+
+      it('to be stopped when propagateContainerEvents=false', function() {
+        config = {
+          propagateContainerEvents: false,
+          template: '<div style="height: 300px">Hello World</div>'
```
ErinCoughlan: Is the template still necessary here? Or does `DEFAULT_TEMPLATE` work?

src/components/panel/panel.spec.js
```diff
@@ -868,6 +895,87 @@ describe('$mdPanel', function() {
...
+        expect(PANEL_WRAPPER_CLASS).not.toHaveClass('my-class');
+      });
+
+      it('should remove a class from the container/wrapper when fromElement=false',
+        function() {
+          openPanel(DEFAULT_CONFIG);
+
+          panelRef.addClass('my-class');
+
+          expect(PANEL_WRAPPER_CLASS).toHaveClass('my-class');
```
ErinCoughlan: super nit: Can you keep the order of these the same for all the tests in this describe? Either put `PANEL_WRAPPER_EL` expect first or `PANEL_EL` first, but don't swap them.

src/components/panel/panel.spec.js
```diff
@@ -868,6 +895,87 @@ describe('$mdPanel', function() {
       });
     });
 
+    describe('CSS class logic:', function() {
+      it('should add a class to the container/wrapper when toElement=false',
+        function() {
```
ErinCoughlan: nit: continuation indents are 4, not 2

src/components/panel/panel.spec.js
```diff
@@ -1115,6 +1223,95 @@ describe('$mdPanel', function() {
...
+        myButton = '<button>myButton</button>';
+        attachToBody(myButton);
+        myButton = angular.element(document.querySelector('button'));
+        myButtonRect = myButton[0].getBoundingClientRect();
+      });
+
+      it('between two absolute positions', function() {
+        var top = '50px';
+        var left = '30px';
+        var position = $mdPanel.newPanelPosition().absolute().top(top).left(left);
```
ErinCoughlan: nit: these will be more readable if you chain them on the next lines, like the relative positioning setup below.

src/components/panel/panel.spec.js
```diff
@@ -1115,6 +1223,95 @@ describe('$mdPanel', function() {
...
+        config['position'] = position;
```
ErinCoughlan: nit: new line before this one

ErinCoughlan:
@ThomasBurleson @bradrich LGTM once nits are resolved

bradrich:
@ErinCoughlan & @ThomasBurleson All requests have been met and everything should be GTG. Please let me know if I have missed anything.

Thanks for everyone's attention.

PR #8986, opened 2016-07-09
wip(tooltip): Tooltip transition to mdPanel

bradrich:
This is a WIP function that will transition the tooltip component to the mdPanel API.

Many of the core pieces of functionality of the original tooltip are still used due to the number of issues that they overcame.

The primary changes were to allow the mdPanel to position, animate, create, open and close the tooltip, instead of relying on proprietary logic within the tooltip code to do those things.

**Needs:**
 - Spec tests
 - Removal of the old tooltip component and renaming of this tooltip to `md-tooltip` instead of `md-tooltip-md-panel`

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -108,6 +108,9 @@ angular
  *   - `attachTo` - `{(string|!angular.JQLite|!Element)=}`: The element to
  *     attach the panel to. Defaults to appending to the root element of the
  *     application.
+ *   - `propagateContainerEvents` - `{boolean=}`: Whether pointer or touch
+ *     events should be allowed to propagate 'go through' the container, aka the
+ *     wrapper, of the panel.
  *   - `panelClass` - `{string=}`: A css class to apply to the panel element.
  *     This class should define any borders, box-shadow, etc. for the panel.
  *   - `zIndex` - `{number=}`: The z-index to place the panel at.
@@ -626,6 +629,7 @@ function MdPanelService($rootElement, $rootScope, $injector, $window) {
     focusOnOpen: true,
     fullscreen: false,
     hasBackdrop: false,
+    propagateContainerEvents: false,
     transformTemplate: angular.bind(this, this._wrapTemplate),
     trapFocus: false,
     zIndex: defaultZIndex
@@ -1054,14 +1058,18 @@ MdPanelRef.prototype.hide = function() {
  * Add a class to the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} newClass Class to be added.
+ * @param {boolean} toElement Whether or not to add the class to the panel
+ *    element instead of the container.
  */
-MdPanelRef.prototype.addClass = function(newClass) {
+MdPanelRef.prototype.addClass = function(newClass, toElement) {
   if (!this._panelContainer) {
     throw new Error('Panel does not exist yet. Call open() or attach().');
   }
 
-  if (!this._panelContainer.hasClass(newClass)) {
+  if (!toElement && !this._panelContainer.hasClass(newClass)) {
     this._panelContainer.addClass(newClass);
+  } else if (toElement && !this._panelEl.hasClass(newClass)) {
+    this._panelEl.addClass(newClass);
   }
 };
 
@@ -1070,14 +1078,18 @@ MdPanelRef.prototype.addClass = function(newClass) {
  * Remove a class from the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} oldClass Class to be removed.
+ * @param {boolean} fromElement Whether or not to remove the class from the panel
+ *    element instead of the container.
  */
-MdPanelRef.prototype.removeClass = function(oldClass) {
+MdPanelRef.prototype.removeClass = function(oldClass, fromElement) {
   if (!this._panelContainer) {
     throw new Error('Panel does not exist yet. Call open() or attach().');
   }
 
-  if (this._panelContainer.hasClass(oldClass)) {
+  if (!fromElement && this._panelContainer.hasClass(oldClass)) {
     this._panelContainer.removeClass(oldClass);
+  } else if (fromElement && this._panelEl.hasClass(oldClass)) {
+    this._panelEl.removeClass(oldClass);
   }
 };
 
@@ -1086,13 +1098,19 @@ MdPanelRef.prototype.removeClass = function(oldClass) {
  * Toggle a class on the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} toggleClass The class to toggle.
+ * @param {boolean} onElement Whether or not to toggle the class on the panel
+ *    element instead of the container.
  */
-MdPanelRef.prototype.toggleClass = function(toggleClass) {
+MdPanelRef.prototype.toggleClass = function(toggleClass, onElement) {
   if (!this._panelContainer) {
     throw new Error('Panel does not exist yet. Call open() or attach().');
   }
 
-  this._panelContainer.toggleClass(toggleClass);
+  if (!onElement) {
+    this._panelContainer.toggleClass(toggleClass);
+  } else {
+    this._panelEl.toggleClass(toggleClass);
+  }
 };
 
 
@@ -1125,9 +1143,14 @@ MdPanelRef.prototype._createPanel = function() {
           self._panelEl = angular.element(
               self._panelContainer[0].querySelector('.md-panel'));
 
-          // Add a custom CSS class.
+          // Add a custom CSS class to the panel element.
           if (self._config['panelClass']) {
             self._panelEl.addClass(self._config['panelClass']);
+          }
+
+          // Handle click and touch events for the panel container.
+          if (self._config['propagateContainerEvents']) {
+            self._panelContainer.css('pointer-events', 'none');
           }
 
           // Panel may be outside the $rootElement, tell ngAnimate to animate
@@ -1188,6 +1211,24 @@ MdPanelRef.prototype._addStyles = function() {
 
 
 /**
+ * Update the position configuration of a panel
+ * @param {MdPanelPosition} position
+ */
+MdPanelRef.prototype.updatePosition = function(position) {
+  if (!this._panelContainer) {
+    throw new Error('Panel does not exist yet. Call open() or attach().');
+  }
+
+  var self = this;
+
+  self._config['position'] = position;
+  self._$mdUtil.nextTick(function() {
+    self._updatePosition();
+  });
+};
+
+
+/**
  * Calculates and updates the position of the panel.
  * @param {boolean=} opt_init
  * @private
@@ -1197,12 +1238,12 @@ MdPanelRef.prototype._updatePosition = function(opt_init) {
 
   if (positionConfig) {
     positionConfig._setPanelPosition(this._panelEl);
-    
+
     // Hide the panel now that position is known.
     if (opt_init) {
       this._panelContainer.addClass(MD_PANEL_HIDDEN);
     }
-    
+
     this._panelEl.css('top', positionConfig.getTop());
     this._panelEl.css('bottom', positionConfig.getBottom());
     this._panelEl.css('left', positionConfig.getLeft());
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -309,6 +309,8 @@ angular
  * Adds a class to the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} newClass Class to be added.
+ * @param {boolean} toElement Whether or not to add the class to the panel
+ *    element instead of the container.
  */
 
 /**
@@ -318,6 +320,8 @@ angular
  * Removes a class from the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} oldClass Class to be removed.
+ * @param {boolean} fromElement Whether or not to remove the class from the
+ * panel element instead of the container.
  */
 
 /**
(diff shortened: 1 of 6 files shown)


EladBezalel:
@bradrich as i told you before, this should include mdPanel changes as you relay on those features. right now the code isn't working and i can't test it.

src/components/tooltipMdPanel/tooltip.js
```diff
@@ -0,0 +1,346 @@
...
+
+  var ENTER_EVENTS = 'focus touchstart mouseenter';
+  var LEAVE_EVENTS = 'blur touchcancel mouseleave';
+  var TOOLTIP_SHOW_DELAY = 0;
+
+  return {
+    restrict: 'E',
+    transclude: true,
+    priority: 210, // Before ngAria
+    template: '<div class="_md-content _md" ng-transclude></div>',
```
EladBezalel: we talked about removing `_md-content`

also don't use private classes cause we're reverting it.

src/components/tooltipMdPanel/tooltip.js
```diff
@@ -0,0 +1,346 @@
...
+
+  function linkFunc(scope, element, attr) {
+
+    $mdTheming(element);
+
+    var parent = $mdUtil.getParentWithPointerEvents(element),
+        attachTo = angular.element(document.body),
+        showTimeout = null,
+        debouncedOnResize = $$rAF.throttle(function() {
+          updatePosition();
```
EladBezalel: `$$rAF.throttle(updatePosition)` isn't enough

src/components/tooltipMdPanel/tooltip.js
```diff
@@ -0,0 +1,346 @@
...
+        }),
+        origin = null,
```
EladBezalel: i would avoid all of those by binding these properties to an object or the directive controller

src/components/tooltipMdPanel/tooltip.js
```diff
@@ -0,0 +1,346 @@
...
+      scope.delay = scope.delay || TOOLTIP_SHOW_DELAY;
+    }
+
+    function manipulateElement() {
+      element.detach();
+      element.attr('role', 'tooltip');
+    }
+
+    function updatePosition() {
+      var direction = scope.direction || 'bottom';
```
EladBezalel: i would do this at `setDefaults` function and use scope.direction instead

src/components/tooltipMdPanel/tooltip.js
```diff
@@ -0,0 +1,346 @@
...
+        });
+
+        attributeObserver.observe(element[0], { attributes: true });
+
+        // Build watcher only if mdVisible is being used
+        if (attr.hasOwnProperty('mdVisible')) {
+          scope.visibleWatcher = scope.$watch('visible', onVisibleChanged);
+        }
+      } else { // MutationObserver not supported
+        scope.visibleWatcher = scope.$watch('visible', onVisibleChanged);
```
EladBezalel: what is the difference?

src/components/tooltipMdPanel/tooltip.js
```diff
@@ -0,0 +1,346 @@
...
+      // adding tooltips while they're invisible. If 'md-visible' become true, at some
+      // point, you'd usually get a lot of inputs.
+      // - We use '.one', not '.on', because this only needs to fire once. If we were
+      // using '.on', it would get thrown into an infinite loop.
+      // - This kicks off the scope's '$destroy' event which finishes the cleanup.
+      element.one('$destroy', onElementDestroy);
+      parent.one('$destroy', onElementDestroy);
+      scope.$on('$destroy', function() {
+        setVisible(false);
+        element.remove();
```
EladBezalel: This is probably an old code

src/components/tooltipMdPanel/tooltip.js
bradrich: I am going to remove the `_md-content` element in my next coding session.

src/components/tooltipMdPanel/tooltip.js
bradrich: Can you please clarify?

src/components/tooltipMdPanel/tooltip.js
bradrich: I think that this could benefit from a full re-write as a component.

src/components/tooltipMdPanel/tooltip.js
bradrich: A lot of the new tooltip code relies on the previous tooltip logic. I studied the code to determine what could be removed and each line seemed like it was written to overcome an issue previously determined.

bradrich:
@EladBezalel You'll have to excuse this oversight. I thought that I had rebased the mdPanel branch into this branch and I must have forgotten to. Please test again and let me know if there are any issues.

src/components/tooltipMdPanel/tooltip.js
EladBezalel: We can't use the component api as we have to support angular 1.3

src/components/tooltipMdPanel/tooltip.js
EladBezalel: You can send updatePosition to the throttle function instead of wrap it

src/components/tooltipMdPanel/tooltip.js
topherfangio: Ideally, we'll have a polyfill that lets us use the `.component()` API, but even rewriting it to use `bindToController`, and `controllerAs: "$ctrl"` would help with the eventual transition.

src/components/panel/panel.js
```diff
@@ -108,6 +108,9 @@ angular
  *   - `attachTo` - `{(string|!angular.JQLite|!Element)=}`: The element to
  *     attach the panel to. Defaults to appending to the root element of the
  *     application.
+ *   - `propagateContainerEvents` - `{boolean=}`: Whether pointer or touch
+ *     events should be allowed to propagate 'go through' the container, aka the
+ *     wrapper, of the panel.
```
topherfangio: In the case of boolean, we usually also specify what the default value is so that users know.

src/components/panel/panel.js
```diff
@@ -306,6 +309,8 @@ angular
  * Adds a class to the panel. DO NOT use this to hide/show the panel.
  *
  * @param {string} newClass Class to be added.
+ * @param {boolean} toElement Whether or not to add the class to the panel
```
topherfangio: Why is this documented at the top of the file and down below on the actual method? Seems to be duplicated?

@ErinCoughlan Is this an internal Google thing to do it twice?

src/components/panel/panel.js
```diff
@@ -1125,11 +1158,16 @@ MdPanelRef.prototype._createPanel = function() {
...
 
-          // Add a custom CSS class.
+          // Add a custom CSS class to the panel element.
           if (self._config['panelClass']) {
             self._panelEl.addClass(self._config['panelClass']);
           }
 
+          // Handle click and touch events for the panel container.
+          if (self._config['propagateContainerEvents']) {
+            self._panelContainer.css('pointer-events', 'none');
```
topherfangio: Not sure if this is applicable here, but just be aware that some of our code calls `$mdUtil.getParentWithPointerEvents()` and it will now skip this. I only mention because I ran into a similar issue earlier with the Speed Dial where it caused it to miss the intended parent.

src/components/tooltipMdPanel/demoBasicUsage/index.html
```diff
@@ -0,0 +1,76 @@
+<div ng-controller="AppCtrl as ctrl" ng-cloak>
+
+  <md-toolbar class="md-accent">
+    <div class="md-toolbar-tools">
+      <h2>Awesome Md App</h2>
```
topherfangio: Minute point, but the `d` here in `Md` should be capitalized 😄

src/components/tooltipMdPanel/tooltip.js
topherfangio: I think @EladBezalel was asking a question in his first post as in:

> ``$rAF.throttle(updatePosition)`` isn't enough?

If you're not passing parameters, you don't need to wrap it with `function() { updatePosition() }`, you can just directly pass the method to be throttled.

src/components/panel/panel.js
ErinCoughlan: The top ones is what is generated into the docs site. The bottom ones are the regular JSDoc for functions. In general, the ng-doc ones would have more details and examples to help devs understand how to use it. The ones below are solely for people working on the code.

Internally, every function needs to have JSDoc, including a descriptions of the method, params, and return statements. We don't have a concept of ng-doc, so we don't run into this public/private issue. The formatting is also slightly different since ngdoc needs types and has the `@description` keyword. There's also `@return` vs `@returns`.

The alternative would be to mix JSDoc and ng-doc within the actual code, but that makes it more difficult to see what is part of the public API vs internal to the panel. We decided early on that it was better to make the distinction so all public documentation is up front and grouped together.

ivanzvakoms:
Not sure if it's ok posting here about issues that this approach has at least one drawback in comparing to current tooltip:
```.md-panel-outer-wrapper {
    height: 100%;
    left: 0;
    position: absolute;
    top: 0;
    width: 100%;
}```
this block is inside of body(attachTo) and it is obvious that this overlapping whole page, so user can't interact with other elements on page when this "tooltip" opened (e.g. md-visible="true")
(tested with on 1.1.0)

File: src/components/panel/panel.js (hunk 8 above)
Comment (ThomasBurleson) [inline src/components/panel/panel.js]: I am concerned that the documentation standards are widely diverging for our ngM1 components.

File: src/components/panel/panel.js (hunk 9 above)
Comment (ThomasBurleson) [inline src/components/panel/panel.js]: 👍  good point.

File: src/components/tooltipMdPanel/demoBasicUsage/index.html [hunk 11]
```diff
@@ -0,0 +1,76 @@
...
+      <span flex></span>
+      <md-button class="md-icon-button test-tooltip" aria-label="Refresh">
+        <md-tooltip-md-panel md-direction="left">Refresh</md-tooltip-md-panel>
+        <md-icon md-svg-src="img/icons/ic_refresh_24px.svg" style="width: 24px; height: 24px;"></md-icon>
+      </md-button>
+    </div>
+  </md-toolbar>
+  <md-content layout-padding style="margin-left: 20px; margin-right: 20px;">
```
ThomasBurleson: please move inline styles to `/demoBasicUsage/styles.css`

src/components/tooltipMdPanel/demoBasicUsage/index.html
```diff
@@ -0,0 +1,76 @@
...
+        <p>
+          Additionally, the Tooltip's <code>md-visible</code> attribute can use data-binding to programmatically
+          show/hide itself. Toggle the checkbox below...
+        </p>
+        <div style="padding-left: 30px;">
+          <md-checkbox ng-model="demo.showTooltip" style="padding-left: 30px;">Insert Drive</md-checkbox>
+        </div>
+      </div>
+
+      <div style="margin-top: 20px; margin-bottom: -5px;">
```
ThomasBurleson: inline style issue again.

src/components/tooltipMdPanel/tooltip.js
ThomasBurleson: what does that mean "next coding session"?

EladBezalel: i think he meant next time he will work on that

src/components/tooltipMdPanel/tooltip.js
```diff
@@ -0,0 +1,346 @@
+/**
+ * @ngdoc module
+ * @name material.components.tooltipMdPanel
+ */
+angular
+  .module('material.components.tooltipMdPanel', [ 'material.core' ])
+  .directive('mdTooltipMdPanel', MdTooltipDirective);
```
ThomasBurleson: no. please do not do this; we should not have an `tooltipMdPanel` component.
Please just use a `wip/panel_tooltip` branch and then modify the tooltip.js.

src/components/tooltipMdPanel/tooltip.js
bradrich: That is correct.

ThomasBurleson:
@bradrich - always label as in-progress while reviews and work is on-going.

### Important
Do not rename component names or create odd new names like *tooltipMdPanel*!

bradrich:
@ThomasBurleson Duly noted. Please excuse my lapse in judgement. I was merely attempting to use the original tooltip logic as a roadmap and felt it simpler to do so with the codebase handy. I assure you that this methodology would never have made it further along in this feature attempt and that I still have plans to completely replace the original tooltip and rename this properly. I hope that you understand.

ThomasBurleson:
@bradrich - after you address this round of comments, please restore the file name to `tooltip.js`, etc. and then push to the branch to update this PR. Thx.

src/components/panel/panel.js
ErinCoughlan: @ThomasBurleson I'm fine with whatever documentation style you want in ngMaterial. The only thing I really feel strongly about is that every method have some form of docs so that anyone reading it can understand what that method does. I've found this really helpful for someone adding to the component after it's created.

The public vs private was so that we didn't use ngdoc on every method (therefore writing every method on the docs site, even private ones). I thought it made the public API more clear to have it at the beginning, but I understand that makes it harder to maintain. I've seen the list everything up front pattern in a few other components, like mdDialog, but I'm open to new suggestions about documentation.

topherfangio:
Closing per Brad/Thomas' request.

PR #8990, opened 2016-07-09
feat(panel): add RTL support

### LINKED ISSUE #8974: mdPanel: Support RTL and LTR positioning.
**Actual Behavior**:
 -  `What is the issue?` *

Developers specify the location in a ltr manner, so when the screen is reversed, panels appear offscreen or in unexpected locations.

Based on the whether the user's orientation, set right or left accordingly when positioning. Currently the panel always sets `left`. [Code](https://github.com/angular/material/blob/master/src/components/panel/panel.js#L1953)

Absolute positioning should be updated as well so that apps work the way developer intended across locales without any additional configuration or thought.

EladBezalel:
- If the application is in RTL mode we swap the `OFFSET` and `ALIGN` - `START` and `END` positions, it tricks the function to switch it origin point and calculate the panel position right.

fixes #8974

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -386,8 +386,8 @@ angular
  * @ngdoc method
  * @name MdPanelPosition#top
  * @description
- * Sets the value of `top` for the panel. Clears any previously set
- * vertical position.
+ * Sets the value of `top` for the panel. Clears any previously set vertical
+ * position.
  * @param {string=} top Value of `top`. Defaults to '0'.
  * @returns {MdPanelPosition}
  */
@@ -396,9 +396,29 @@ angular
  * @ngdoc method
  * @name MdPanelPosition#bottom
  * @description
- * Sets the value of `bottom` for the panel. Clears any previously set
- * vertical position.
+ * Sets the value of `bottom` for the panel. Clears any previously set vertical
+ * position.
  * @param {string=} bottom Value of `bottom`. Defaults to '0'.
+ * @returns {MdPanelPosition}
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelPosition#start
+ * @description
+ * Sets the panel to the start of the page - `left` if `ltr` or `right` for `rtl`. Clears any previously set
+ * horizontal position.
+ * @param {string=} start Value of position. Defaults to '0'.
+ * @returns {MdPanelPosition}
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelPosition#end
+ * @description
+ * Sets the panel to the end of the page - `right` if `ltr` or `left` for `rtl`. Clears any previously set
+ * horizontal position.
+ * @param {string=} end Value of position. Defaults to '0'.
  * @returns {MdPanelPosition}
  */
 
@@ -713,7 +733,7 @@ MdPanelService.prototype.open = function(config) {
  * @returns {MdPanelPosition}
  */
 MdPanelService.prototype.newPanelPosition = function() {
-  return new MdPanelPosition(this._$window);
+  return new MdPanelPosition(this._$injector);
 };
 
 
@@ -1243,10 +1263,10 @@ MdPanelRef.prototype._updatePosition = function(init) {
       this._panelContainer.addClass(MD_PANEL_HIDDEN);
     }
 
-    this._panelEl.css('top', positionConfig.getTop());
-    this._panelEl.css('bottom', positionConfig.getBottom());
-    this._panelEl.css('left', positionConfig.getLeft());
-    this._panelEl.css('right', positionConfig.getRight());
+    this._panelEl.css(MdPanelPosition.absPosition.TOP, positionConfig.getTop());
+    this._panelEl.css(MdPanelPosition.absPosition.BOTTOM, positionConfig.getBottom());
+    this._panelEl.css(MdPanelPosition.absPosition.LEFT, positionConfig.getLeft());
+    this._panelEl.css(MdPanelPosition.absPosition.RIGHT, positionConfig.getRight());
 
     // Use the vendor prefixed version of transform.
     var prefixedTransform = this._$mdConstant.CSS.TRANSFORM;
@@ -1570,12 +1590,15 @@ MdPanelRef.prototype._done = function(callback, self) {
  *   position: panelPosition
  * });
  *
- * @param {!angular.$window} $window
+ * @param {!angular.$injector} $injector
  * @final @constructor
  */
-function MdPanelPosition($window) {
-  /** @private @const */
-  this._$window = $window;
+function MdPanelPosition($injector) {
+  /** @private @const {!angular.$window} */
+  this._$window = $injector.get('$window');
+
+  /** @private {boolean} */
+  this._isRTL = $injector.get('$mdUtil').bidi() === 'rtl';
 
   /** @private {boolean} */
   this._absolute = false;
@@ -1636,11 +1659,48 @@ MdPanelPosition.yPosition = {
 
 
 /**
+ * Possible values of absolute position.
+ * @enum {string}
+ */
+MdPanelPosition.absPosition = {
+  TOP: 'top',
+  RIGHT: 'right',
+  BOTTOM: 'bottom',
+  LEFT: 'left'
+};
+
+
+/**
  * Sets absolute positioning for the panel.
  * @return {!MdPanelPosition}
  */
 MdPanelPosition.prototype.absolute = function() {
   this._absolute = true;
+  return this;
+};
+
+/**
+ * Sets the value of a position for the panel. Clears any previously set position.
+ * @param {string} position Position to set
+ * @param {string=} value Value of the position. Defaults to '0'.
+ * @returns {MdPanelPosition}
+ * @private
+ */
+MdPanelPosition.prototype._setPosition = function(position, value) {
+  if (position === MdPanelPosition.absPosition.RIGHT || position === MdPanelPosition.absPosition.LEFT) {
+    this._left = this._right = '';
+  }
+  else if (position === MdPanelPosition.absPosition.BOTTOM || position === MdPanelPosition.absPosition.TOP) {
+    this._top = this._bottom = '';
+  }
+  else {
+    var positions = Object.keys(MdPanelPosition.absPosition).join().toLowerCase();
+
+    throw new Error('Position must be one of ' + positions + '.');
+  }
+
+  this['_' +  position] = angular.isString(value) ? value : '0';
+
   return this;
 };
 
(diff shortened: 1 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -1950,25 +1954,36 @@ MdPanelPosition.prototype._calculatePanelPosition = function(panelEl, position)
   var targetRight = targetBounds.right;
   var targetWidth = targetBounds.width;
 
+  function reverseXPosition (position) {
+    var start = 'start';
+    var end = 'end';
+
+    if (position.includes(start)) {
+      return position.replace(start, end);
```
ThomasBurleson: is `replace` a **switch** or **overwrite** ?

src/components/panel/panel.js
```diff
@@ -1950,25 +1954,36 @@ MdPanelPosition.prototype._calculatePanelPosition = function(panelEl, position)
...
+    }
+
+    return position.replace(end, start);
+  }
+
+  if (this._$mdUtil.bidi() === 'rtl') {
+    position.x = reverseXPosition(position.x);
+  }
+
   switch (position.x) {
```
ThomasBurleson: I recommend this:

```
function bidi(position) {
  isRTL = this._$mdUtil.bidi() === 'rtl';

  return  isRTL ? this.reversePosition(position) : position;

}

function reversePosition (position) {
  var start = 'start', end = 'end',
      hasStart = position.includes(start);

  return hasStart ? position.replace( start, end ) : position.replace( end, start );
}

switch ( bidi(position.x) )  {
    ...
}
```

src/components/panel/panel.js
```diff
@@ -1950,25 +1954,36 @@ MdPanelPosition.prototype._calculatePanelPosition = function(panelEl, position)
   var targetRight = targetBounds.right;
   var targetWidth = targetBounds.width;
 
+  function reverseXPosition (position) {
+    var start = 'start';
+    var end = 'end';
+
+    if (position.includes(start)) {
```
crisbeto: String.prototype.includes definitely looks nice, but according to [MDN](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/String/includes) it's not supported in IE at all. I think it's better to use `position.indexOf(start) > -1` here.

EladBezalel: Nice catch :+1:

src/components/panel/panel.js
```diff
@@ -710,7 +710,7 @@ MdPanelService.prototype.open = function(opt_config) {
  * @returns {MdPanelPosition}
  */
 MdPanelService.prototype.newPanelPosition = function() {
-  return new MdPanelPosition(this._$window);
+  return new MdPanelPosition(this._$window, this._$injector);
```
ErinCoughlan: If you're going to pass $injector, you can remove $window and just get that from $injector as well.

In general, if there were 3+ items, we used $injector, if there were less, we injected them directly.

src/components/panel/panel.js
```diff
@@ -1950,25 +1954,36 @@ MdPanelPosition.prototype._calculatePanelPosition = function(panelEl, position)
   var targetRight = targetBounds.right;
   var targetWidth = targetBounds.width;
 
+  function reverseXPosition (position) {
+    var start = 'start';
```
ErinCoughlan: Perhaps bail early if position === xPosition.CENTER, since that shouldn't matter with bidi.

src/components/panel/panel.js
```diff
@@ -1950,25 +1954,36 @@ MdPanelPosition.prototype._calculatePanelPosition = function(panelEl, position)
   var targetRight = targetBounds.right;
   var targetWidth = targetBounds.width;
 
+  function reverseXPosition (position) {
```
ErinCoughlan: Please follow the same styles in the rest of this file. Everything is using prototype. No functions are defined like this.

ErinCoughlan: nit: extra space between function name and params

ErinCoughlan:
So this definitely fixes bidi for relative positioning, but not for absolute. If I want a popup on the left side of the screen, I'd still want it to be first in reading order in RTL languages.

What's the plan for fixing absolution positioning?

ThomasBurleson:
@EladBezalel - ping

EladBezalel:
@ErinCoughlan please re-review,
This isn't solving absolute positioning yet,
can you elaborate?

src/components/panel/panel.js
```diff
@@ -1588,12 +1608,15 @@ MdPanelRef.prototype._done = function(callback, self) {
...
  *
- * @param {!angular.$window} $window
+ * @param {!angular.$injector} $injector
  * @final @constructor
  */
-function MdPanelPosition($window) {
-  /** @private @const */
-  this._$window = $window;
+function MdPanelPosition($injector) {
+  /** @private @const {!angular.$window}*/
```
ErinCoughlan: Nit: space after type before */

src/components/panel/panel.js
```diff
@@ -1662,56 +1685,93 @@ MdPanelPosition.prototype.absolute = function() {
...
+    this._left = this._right = '';
+  }
+  else if (position === 'bottom' || position === 'top') {
+    this._top = this._bottom = '';
+  }
+  else {
+    throw new Error('Position must be one of \'top\', \'bottom\', \'left\' or \'right\'.');
+  }
+
+  this['_' +  position] = value || '0';
```
ErinCoughlan: Can you assert that value is actually a string? I don't want anyone setting a function here. This line screams hacker to me.

if (!angular.isString(value)) {
  throw new Error('position value must be a string.');
}

// Written on my phone without back ticks.

src/components/panel/panel.js
```diff
@@ -1662,56 +1685,93 @@ MdPanelPosition.prototype.absolute = function() {
...
 
+/**
+ * Sets the value of a position for the panel. Clears any previously set position.
+ * @param {string} position Position to set
+ * @param {string=} value Value of the position. Defaults to '0'.
+ * @returns {MdPanelPosition}
+ * @private
+ */
+MdPanelPosition.prototype._setPosition = function(position, value) {
+  if (position === 'right' || position === 'left') {
```
ErinCoughlan: I really dislike "random" strings in code, even if it won't be user visible. (Too easy to mistype when editing the code.) What do you think about making a absolutePosition enum with left, right, top, and bottom?

MdPanelPosition.absPosition.

src/components/panel/panel.js
```diff
@@ -1989,6 +2049,20 @@ MdPanelPosition.prototype._setPanelPosition = function(panelEl) {
   }
 };
 
+MdPanelPosition.prototype._reverseXPosition = function (position) {
```
ErinCoughlan: Nit: no space between function and params

ErinCoughlan: Missing jsdoc for this method

src/components/panel/panel.js
```diff
@@ -1989,6 +2049,20 @@ MdPanelPosition.prototype._setPanelPosition = function(panelEl) {
...
+  if (position === MdPanelPosition.xPosition.CENTER) {
+    return;
+  }
+
+  var start = 'start', end = 'end';
```
ErinCoughlan: Nit: one variable declaration per line.

src/components/panel/panel.js
```diff
@@ -1989,6 +2049,20 @@ MdPanelPosition.prototype._setPanelPosition = function(panelEl) {
...
+
+  return position.indexOf(start) > -1 ? position.replace(start, end) : position.replace(end, start);
+};
+
+MdPanelPosition.prototype._bidi = function (position) {
```
ErinCoughlan: nit: Missing jsdoc, two lines between methods, and no space between function and params.

src/components/panel/panel.js
```diff
@@ -2008,25 +2082,21 @@ MdPanelPosition.prototype._calculatePanelPosition = function(panelEl, position)
   var targetRight = targetBounds.right;
   var targetWidth = targetBounds.width;
 
-  switch (position.x) {
+  switch (this._bidi(position.x)) {
```
ErinCoughlan: Please add a test that left and right are reversed if rtl.

src/components/panel/panel.spec.js
```diff
@@ -1538,6 +1538,52 @@ describe('$mdPanel', function() {
...
+        var start = '50px';
+        mdPanelPosition._isRTL = true;
+        config['position'] = mdPanelPosition.absolute().start(start);
+
+        openPanel(config);
+
+        var panelCss = document.querySelector(PANEL_EL).style;
+        expect(panelCss.right).toEqual(start);
+
+        mdPanelPosition._isRTL = false;
```
ErinCoughlan: Does it make sense to add this to an afterEach to be extra safe that it's reset?

src/components/panel/panel.js
```diff
@@ -1662,56 +1697,93 @@ MdPanelPosition.prototype.absolute = function() {
...
+ */
+MdPanelPosition.prototype._setPosition = function(position, value) {
+  if (position === MdPanelPosition.absPosition.RIGHT || position === MdPanelPosition.absPosition.LEFT) {
+    this._left = this._right = '';
+  }
+  else if (position === MdPanelPosition.absPosition.BOTTOM || position === MdPanelPosition.absPosition.TOP) {
+    this._top = this._bottom = '';
+  }
+  else {
+    throw new Error('Position must be one of ' + Object.keys(MdPanelPosition.absPosition).join().toLowerCase() + '.');
```
ErinCoughlan: nit: line too long

src/components/panel/panel.js
```diff
@@ -1662,56 +1697,93 @@ MdPanelPosition.prototype.absolute = function() {
...
+
+
+/**
+ * Sets the panel to the start of the page - `left` if `ltr` or `right` for `rtl`. Clears any previously set
+ * horizontal position.
+ * @param {string=} start Value of position. Defaults to '0'.
+ * @returns {MdPanelPosition}
+ */
+MdPanelPosition.prototype.start = function(start) {
+  return this._setPosition(this._isRTL ? MdPanelPosition.absPosition.RIGHT : MdPanelPosition.absPosition.LEFT, start);
```
ErinCoughlan: nit: line too long

src/components/panel/panel.js
```diff
@@ -1662,56 +1697,93 @@ MdPanelPosition.prototype.absolute = function() {
...
+
+
+/**
+ * Sets the panel to the end of the page - `right` if `ltr` or `left` for `rtl`. Clears any previously set
+ * horizontal position.
+ * @param {string=} end Value of position. Defaults to '0'.
+ * @returns {MdPanelPosition}
+ */
+MdPanelPosition.prototype.end = function(end) {
+  return this._setPosition(this._isRTL ? MdPanelPosition.absPosition.LEFT : MdPanelPosition.absPosition.RIGHT, end);
```
ErinCoughlan: nit: line too long

ErinCoughlan:
LGTM, Approval :smile:

PR #8993, opened 2016-07-10
fix(theming): update palette contrast types to current spec

### LINKED ISSUE #8992: theme: palettes using outdated opacity for light contrast
**Actual Behavior**:
-  `What is the issue?` Opacities for light contrast vary between 87% and 100%
-  `What is the expected behavior?`Opacity for light contrast should be fixed to 100%

**Angular Versions**: *
-  `Angular Version:` N/A

Previously, the spec stated opacities were set to 87% for text on dark background
https://web.archive.org/web/20140628054353/http://www.google.com/design/spec/style/color.html#color-ui-color-application

Now, it's 100% always:
https://material.google.com/style/color.html#color-text-background-colors

clshortfuse:
Previous values used out-of-spec colors for contrast.
Low opacities values caused the design to not to meet AA standards
of mininum 4.5 contrast ratio for default text.

* Set hues with contrast type `light` to `strongLight`.
* Set contrast type for `green` `500` to `dark`.

Fixes #8992, #10164

### DIFF
--- a/src/core/services/theming/theme.palette.js
+++ b/src/core/services/theming/theme.palette.js
@@ -17,7 +17,7 @@ angular.module('material.core.theming.palette', [])
     'A700': '#d50000',
     'contrastDefaultColor': 'light',
     'contrastDarkColors': '50 100 200 300 A100',
-    'contrastStrongLightColors': '400 500 600 700 A200 A400 A700'
+    'contrastStrongLightColors': '400 500 600 700 800 900 A200 A400 A700'
   },
   'pink': {
     '50': '#fce4ec',
@@ -36,7 +36,7 @@ angular.module('material.core.theming.palette', [])
     'A700': '#c51162',
     'contrastDefaultColor': 'light',
     'contrastDarkColors': '50 100 200 A100',
-    'contrastStrongLightColors': '500 600 A200 A400 A700'
+    'contrastStrongLightColors': '300 400 500 600 700 800 900 A200 A400 A700'
   },
   'purple': {
     '50': '#f3e5f5',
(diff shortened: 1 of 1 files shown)


clshortfuse:
Fixes #8992

src/core/services/theming/theme.palette.js
```diff
@@ -17,7 +17,7 @@ angular.module('material.core.theming.palette', [])
     'A700': '#d50000',
     'contrastDefaultColor': 'light',
     'contrastDarkColors': '50 100 200 300 A100',
-    'contrastStrongLightColors': '400 500 600 700 A200 A400 A700'
+    'contrastStrongLightColors': '400 500 600 700 800 900 A200 A400 A700'
```
EladBezalel: how you decided that 800 and 900 should be strong light?

EladBezalel:
I'm not really sure of what are you trying to do,
According to the [Material Design Color Spec](https://material.google.com/style/color.html#color-color-palette) the colors you changed aren't matching

clshortfuse:
@EladBezalel #8992 Spec says opacites 100% now

topherfangio:
Needs internal team discussion with Naomi, Jeremy and Erin.

clshortfuse:
We previously discussed theme changes to be pushed back to 1.2.x

clshortfuse:
Rebased. I caught an a11y issue where some colors would not meet WCAG AA guidelines so perhaps we should considering escalate this.

clshortfuse:
@Splaktar If we're going by the [2014 values](https://material.io/archive/guidelines/style/color.html#color-usability), then secondary is 54% opacity on light for black/dark text. If we're going by the [2019 values](https://material.io/design/color/text-legibility.html#text-backgrounds), that's been bumped up to 60%. It would slightly change your AA values values. (Just an FYI.)

Splaktar:
@clshortfuse in general, we're trying to stick to the 2014-2017 spec in this repo. That said, if it comes down to something being inaccessible vs not, then we'll pick the accessible option. In this case, `0.54` vs `0.6` opacity doesn't seem like it will have a significant impact on a11y.

Thank you for taking a look.

PR #8995, opened 2016-07-10
perf(theming): don't use $watch if register for change is available

clshortfuse:
Register with parent theme controller for theme change if possible.

### DIFF
--- a/src/core/services/theming/theming.js
+++ b/src/core/services/theming/theming.js
@@ -462,7 +462,9 @@ function ThemingProvider($mdColorPalette) {
 
       updateThemeClass(lookupThemeName());
 
-      el.on('$destroy', watchTheme ? $rootScope.$watch(lookupThemeName, updateThemeClass) : angular.noop );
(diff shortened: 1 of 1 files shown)


clshortfuse:
When using `alwaysWatchTheme`, performance suffers on pages with many elements since every themed element would attach a watcher.

http://codepen.io/shortfuse/pen/ZOeObm

src/core/services/theming/theming.js
```diff
@@ -492,6 +494,18 @@ function ThemingProvider($mdColorPalette) {
           el.data('$mdThemeController', ctrl);
         }
       }
+
+      /**
+       * Register change callback with parent mdTheme controller
+       */
+      function registerChangeCallback() {
+        var parentController = parent.controller('mdTheme');
```
EladBezalel: what if `mdTheme` is applied on parent's parent?

clshortfuse: Then it should fall back to the watcher. I was trying to keep the logic the same, not improve upon it. That's why I didn't use `el.data('$mdThemeController')`. If you think I should use it, let me know and I can add it.

PR #9003, opened 2016-07-11
docs(autocomplete): clarify intention of md-autoselect attribute

### LINKED ISSUE #9001: Md-autoselect doesn't select first item in md-autocomplete
Hello!
I'm using md-autocomplete in my project and it's working well apart from that I can't seem to get md-autoselect to work. I wan't it to select first item in the list but it doesn't behave as stated in documentation (If true, the first item will be selected by default).
I tried to make md-autoselect work by experimenting with code in md-autocomplete demos but it didn't work, [http://codepen.io/anon/pen/xOXoRP]. Here I added md-autoselect and deleted placeholder.

DevVersion:
* It was not pretty clear that the `md-autoselect` attribute only selects the first item in the "dropdown", and not selects it inside of the scope / model.

Closes #9001.


clshortfuse:
Alternatively, without the passive voice, we can use:

> When set to `true`, automatically selects the first item upon dropdown open

or 

> When set to `true`, automatically selects the first dropdown item upon open

PR #9010, opened 2016-07-12
feat(theme): add `md-themes-disabled` directive to disable themes

ThomasBurleson:
Similar to `md-layout-css`, this attribute directive would disable
all runtime, startup theme sytle generation and DOM injections

### DIFF
--- a/src/core/services/layout/layout.js
+++ b/src/core/services/layout/layout.js
@@ -174,8 +174,8 @@
     * @ngInject
     */
    function detectDisabledLayouts() {
-    var items = document.querySelectorAll('[md-layouts-disabled]');
-    if ( items.length ) config.enabled = false;
+     var isDisabled = !!document.querySelector('[md-layouts-disabled]');
+     config.enabled = !isDisabled;
    }
 
   /**
--- a/src/core/services/layout/layout.js
+++ b/src/core/services/layout/layout.js
@@ -101,8 +101,25 @@
 
     // Register other, special directive functions for the Layout features:
     module
-      .directive('mdLayoutCss'  , disableLayoutDirective )
-      .directive('ngCloak'      ,  buildCloakInterceptor('ng-cloak'))
+
+      .provider('$$mdLayout'     , function() {
+        // Publish internal service for Layouts
+        return {
+          $get : angular.noop,
+          validateAttributeValue : validateAttributeValue,
+          validateAttributeUsage : validateAttributeUsage,
+          /**
+           * Easy way to disable/enable the Layout API.
+           * When disabled, this stops all attribute-to-classname generations
+           */
+          disableLayouts  : function(isDisabled) {
+            config.enabled =  (isDisabled !== true);
+          }
+        };
+      })
+
+      .directive('mdLayoutCss'        , disableLayoutDirective )
+      .directive('ngCloak'            , buildCloakInterceptor('ng-cloak'))
 
       .directive('layoutWrap'   , attributeWithoutValue('layout-wrap'))
       .directive('layoutNowrap' , attributeWithoutValue('layout-nowrap'))
@@ -126,7 +143,10 @@
       .directive('hideLtMd'       , warnAttrNotSupported('hide-lt-md'))
       .directive('hideLtLg'       , warnAttrNotSupported('hide-lt-lg'))
       .directive('showLtMd'       , warnAttrNotSupported('show-lt-md'))
-      .directive('showLtLg'       , warnAttrNotSupported('show-lt-lg'));
+      .directive('showLtLg'       , warnAttrNotSupported('show-lt-lg'))
+
+      // Determine if
+      .config( detectDisabledLayouts );
 
     /**
      * Converts snake_case to camelCase.
@@ -142,6 +162,21 @@
     }
 
   }
+
+
+  /**
+    * Detect if any of the HTML tags has a [md-layouts-disabled] attribute;
+    * If yes, then immediately disable all layout API features
+    *
+    * Note: this attribute should be specified on either the HTML or BODY tags
+    */
+   /**
+    * @ngInject
+    */
+   function detectDisabledLayouts() {
+    var items = document.querySelectorAll('[md-layouts-disabled]');
+    if ( items.length ) config.enabled = false;
+   }
 
   /**
    * Special directive that will disable ALL Layout conversions of layout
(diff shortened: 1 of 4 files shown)


src/core/services/theming/theming.js
```diff
@@ -538,6 +539,36 @@ function ThemingDirective($mdTheming, $interpolate, $log) {
...
+ * theme stylesheets; e.g. custom themes from Material-Tools:
+ *
+ *       `angular-material.themes.css`
+ *
+ * Another option is to use the ThemingProvider to configure and disable the attribute
+ * conversions; this would obviate the use of the `md-themes-css` directive
+ *
+ */
+function disableThemesDirective() {
+  return {
```
crisbeto: I think that you can avoid the `compile` by doing something like this.:

```js
function disableThemesDirective() {
  disableTheming = true;

  return {
    restrict : 'A',
    priority : '900'
  };
}
```

The directive function gets called only once, the first time it matches an element.

ThomasBurleson:
👍

EladBezalel:
It actually doesn't makes sense for me to have it as a directive.
This is a perfect usecase of service disable function e.g `$animate.enabled`

ThomasBurleson:
We already have `$mdThemingProvider.disableTheming()`
This simply balances that with an HTML-only markup version.

> And it is consistent with the `md-layouts-css` HTML attribute that will disable layout injections.

ThomasBurleson:
@EladBezalel, @topherfangio - plz review.

src/core/services/layout/layout.js
```diff
@@ -101,8 +101,25 @@
...
+        return {
+          $get : angular.noop,
+          validateAttributeValue : validateAttributeValue,
+          validateAttributeUsage : validateAttributeUsage,
+          /**
+           * Easy way to disable/enable the Layout API.
+           * When disabled, this stops all attribute-to-classname generations
+           */
+          disableLayouts  : function(isDisabled) {
+            config.enabled =  (isDisabled !== true);
```
EladBezalel: i'd use `!isDisabled`

src/core/services/layout/layout.js
```diff
@@ -143,6 +163,21 @@
...
+    * Detect if any of the HTML tags has a [md-layouts-disabled] attribute;
+    * If yes, then immediately disable all layout API features
+    *
+    * Note: this attribute should be specified on either the HTML or BODY tags
+    */
+   /**
+    * @ngInject
+    */
+   function detectDisabledLayouts() {
+    var items = document.querySelectorAll('[md-layouts-disabled]');
```
EladBezalel: I think we should only check `<body>` or `<html>` instead of attribute selectors all over

src/core/services/theming/theming.js
```diff
@@ -7,10 +9,24 @@
 angular.module('material.core.theming', ['material.core.theming.palette'])
   .directive('mdTheme', ThemingDirective)
   .directive('mdThemable', ThemableDirective)
+  .directive('mdThemesDisabled', disableThemesDirective )
```
EladBezalel: `mdThemingDisabled` as for `$mdTheming` provider..

src/core/services/theming/theming.js
```diff
@@ -7,10 +9,24 @@
...
 
 /**
+ * Detect if the HTML or the BODY tags has a [md-themes-disabled] attribute
+ * If yes, then immediately disable all theme stylesheet generation and DOM injection
+ */
+/**
+ * @ngInject
+ */
+function detectDisabledThemes($mdThemingProvider) {
+ var isDisabled = document.querySelectorAll('[md-themes-disabled]').length > 0;
```
EladBezalel: Again i think we should check body and html,

We maybe can extract our `bidi` service functionality that checks `dir` attribute on `body` and `html` and make it a generic function with a callback

src/core/services/theming/theming.js
```diff
@@ -202,23 +218,24 @@ var VALID_HUE_VALUES = [
...
-var nonce = null;
-var disableTheming = false;
-
-// Custom styles registered to be used in the theming of custom components.
-var registeredStyles = [];
+var themeConfig = {
+  disableTheming : false,   // Generate our themes at run time; also disable stylesheet DOM injection
+  generateOnDemand : false, // Whether or not themes are to be generated on-demand (vs. eagerly).
+  registeredStyles : [],    // Custom styles registered to be used in the theming of custom components.
+  nonce : null              // Nonce to be added as an attribute to the generated themes style tags.
```
EladBezalel: does it makes sense to leave it null?
also `disableTheming` and `generateOnDemand` are both initialized with false which is the same as not initializing them..

src/core/services/layout/layout.js
crisbeto: The `querySelectorAll` here seems to be redundant. You should be able to do the same with `querySelector` and checking if it returns a result.

src/core/services/theming/theming.js
crisbeto: The `querySelectorAll` isn't necessary here either, You can do the same with `var isDisabled = !!document.querySelector('[md-themes-disabled]')`.

src/core/services/theming/theming.js
```diff
@@ -538,6 +561,35 @@ function ThemingDirective($mdTheming, $interpolate, $log) {
...
+ *
+ * <link rel="stylesheet" href="angular-material.min.css">
+ * <link rel="stylesheet" href="angular-material.themes.css">
+ *
+ * <body md-theme-css>
+ *  ...
+ * </body>
+ *
+ * Note: Using md-themes-css directive requires the developer to load external
+ * theme stylesheets; e.g. custom themes from Material-Tools:
```
topherfangio: If we're going to mention the Material-Tools here, this should probably be a link to the repo.

src/core/services/layout/layout.js
ThomasBurleson: `disableLayouts() == (config.enabled = false)`
`disableLayouts(false) == (config.enabled = true)`
`disableLayouts(true) == (config.enabled = false)`

src/core/services/theming/theming.js
ThomasBurleson: except for `disableTheming` the other keys are unchanged from before.
I think explicit value === `false` is very declarative.

src/core/services/layout/layout.js
EladBezalel: `disableLayouts()` sounds like it should disable layouts ...
also, `disableLayouts(false)` doesn't makes sense :confounded:

DevVersion: Yeah, I guess it's the same problem as in https://github.com/angular/material/pull/8709#discussion_r66444372

ThomasBurleson:
Already merged with SHA ea43da3e2cc2cecc641da9b8a6788479c86cf0b3

PR #9012, opened 2016-07-12
fix(select): Failing tests on Angular 1.6 (snapshot).

topherfangio:
For unknown reasons, the Material tests were failing on Angular 1.6, but not 1.3/4/5. Update test to ensure select is fully closed before attempting to open again to fix the tests.


topherfangio:
@ThomasBurleson Took me a bit to figure out the exact issue, but this seems to fix the failing 1.6 tests.

PR #9013, opened 2016-07-12
docs(input): ngMessage elements should be block

### LINKED ISSUE #8880: Documentation: input + ngMessages
When you're using the `ng-messages` directive an input, and you're *not* using the attribute style on a `div`, the outcome is unexpected and not in line with the documentation.

```html
<div class="ng-message">
```
Creates a block level element.
```html
<ng-message>
```
Creates an inline element.

The result is that the animations don't work correctly. IMHO, it would be nice of this was mentioned in the documentation.

The "fix" for this is to make `ng-message` and `ng-messages` directives block level elements, or to use `div`s as seen in the examples.

```scss
ng-messages {
	display : block;
	ng-message {
		display : block;
	}
}
```

oliversalzburg:
ngMessage and ngMessages elements need to be block-level, otherwise animations applied to them through the input directive will not look right.
Closes #8880

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -39,6 +39,10 @@ angular.module('material.components.input', [
  * <b>Exception:</b> Hidden inputs (`<input type="hidden" />`) are ignored and will not throw an error, so
  * you may combine these with other inputs.
  *
+ * <b>Note:</b> When using `ngMessages` with your input element, make sure the message and container elements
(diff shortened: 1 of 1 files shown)


topherfangio:
LGTM 👍

Thanks so much for submitting this!

@ThomasBurleson I don't know if it needs to be in 1.1.0 or not, but this is a helpful doc addition for users trying to use ngMessages with inputs. Should have 0 side-effect.

PR #9014, opened 2016-07-12
test(mdSelect): fix tests on Angular master (v1.5.8-build.4934)

gkalpak:


ThomasBurleson:
@gkalpak - thx
@topherfangio  - is this a better fix than #9012 ?

gkalpak:
@topherfangio, we were obviously working on this simultaneously :smile:

So, this is essentially very similar to #9012. The main difference is that I split the test for different empty `mdOption`s into separate tests. This is a good idea anyway imo, because you are using the same scope (`$rootScope`) in all cases, which could result into one test's state affecting the subsequent tests.
<sub>(This one also avoids some unnecessary calls, but this is insignificant in the context of tests).</sub>

topherfangio:
That's not a bad idea. I think a merging of the two would be good? I like explicitly calling `closeSelect()` so that it's apparent to the user what we expect to happen.

ThomasBurleson:
@topherfangio - can you merge this with latest from master (which already has 9012 merged).

gkalpak:
Superseeded by #9041.

PR #9022, opened 2016-07-13
update(components): replace Layout attributes with flexbox css

ThomasBurleson:
With new support to disable the Layouts API globally, the md components now use flexbox css instead of the Layout API. This CSS can be easily overriden with future releases of the Layout Flexbox engine.

### DIFF
--- a/src/components/navBar/navBar.scss
+++ b/src/components/navBar/navBar.scss
@@ -13,6 +13,11 @@ $md-nav-bar-height: 48px;
   list-style: none;
   margin: 0;
   padding: 0;
+
+  // Layout [layout='row']
+  box-sizing: border-box;
+  display: flex;
+  flex-direction: row;
 }
 
 .md-nav-item:first-of-type {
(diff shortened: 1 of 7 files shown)


src/components/toast/toast.scss
```diff
@@ -49,6 +50,13 @@ md-toast {
     transition: $swift-ease-out;
 
     @include rtl(justify-content, flex-start, flex-end);
+
+    span {
+      // Layout  [flex]
```
ThomasBurleson: may I should use a new mixin   `@include layout_flex()` to centralize this logic?

src/components/navBar/navBar.scss
```diff
@@ -13,6 +13,11 @@ $md-nav-bar-height: 48px;
   list-style: none;
   margin: 0;
   padding: 0;
+
+  // Layout [layout='row']
+  box-sizing: border-box;
```
ThomasBurleson: may I should use a new mixin `@include layout()` to centralize this logic?

Issue #9021, opened 2016-07-13
mdNavBar: Add responsivity.


menosprezzi:
<h2>Feat Detail</h2>
Implement responsive experience to mdNavBar as there is in mdTabs.

menosprezzi:
The pull request #9023 closes this issue

ThomasBurleson:
@see https://github.com/angular/material/pull/9023#issuecomment-241453954

Issue #8329, opened 2016-05-04
Angular Material very slow in IE 11


mtorres0612:
I've overhauled our application using world's finest Angular Material. It works fast on Chrome and Firefox. 
-  Problem is, it is running slow on IE 11
-  Expected behavior should be the same as the 2 browsers

Below is the gif which illustrates the behavior of our application if run in Chrome:
![animation_chrome](https://cloud.githubusercontent.com/assets/18588219/15004062/0f61cf86-11e5-11e6-9fb2-7f26bef20491.gif)

Below is the gif which illustrates the behavior of our application if run in IE 11:
![animation](https://cloud.githubusercontent.com/assets/18588219/15003932/efa04a52-11e3-11e6-823f-c84d0d12102f.gif)

 Issue Details:
- Slow performance
- Left menu items have inconsistent padding
- mdDialog displays normal in Chrome, but not in IE

Angular Versions / Relevant Script used:
- Angular: https://ajax.googleapis.com/ajax/libs/angularjs/1.4.8/angular.min.js
- Animate: https://ajax.googleapis.com/ajax/libs/angularjs/1.4.8/angular-animate.min.js
- Material: https://ajax.googleapis.com/ajax/libs/angular_material/1.1.0-rc4/angular-material.min.js

Additional Information:
-  Browser Type: IE
-  Browser Version: 11
-  OS: Windows 7

berndvdveen:
+1. Very low performance on any animated transitional behaviour

graphefruit:
@mtorres0612 Have a look here #8218 and here https://github.com/angular/material/issues/864#issuecomment-200325064

This works wonders on my side for performance problems.

Try to disable animations aswell as ripple effects.
They can be disabled on the root element
`<body md-no-ink>`

mtorres0612:
@berndvdveen - thanks for the info. found any workaround on this?
@graphefruit - thank you for the information! In line with this, what am I expecting? a 100% lag-free AM website in IE11?

If in that case yes, then here are the details of my test:

- applied $scope.constant... fix related to #8218
- applied <body md-no-ink>
- results:
   - navigating through different views improved
   - mdDialog popup behavior still lags
   - Tabs also lag when switching between tabs

What else should I apply?

Thank you in advance guys.

graphefruit:
@mtorres0612 100% lag-free? I don't know, try it and you'll see. I reduced my load times from 11 seconds to 4-5 seconds on page navigation in my cordova-app on WP8.

[...]

PR #9023, opened 2016-07-14
feat(mdNavBar): Responsivity added. Close issue #9021

menosprezzi:
<h2> Request Details </h2>
This commits adds pagination when the entire width of this directive is not suported. Like mdTabs, have arrows to slide and work well with gestures (swipe left/right).

### DIFF
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -222,13 +222,24 @@ function MdNavBarController($element, $scope, $timeout, $mdConstant, $mdUtil) {
   var self = this;
 
 
-  // Responsivity actions
+  // Responsivity States
 
   /** @type {!boolean} */
   this.shouldPaginate = this.canPaginate();
 
   /** @type {!number} */
   this.left = 0;
+
+  /**
+    * To avoid multiple parent.clientWidth get calls, cache the value in this scope variable
+    * @type {!number} */
+  this.parentWidth = 0;
+
+  /**
+    * To avoid multiple canvas.scrollWidth get calls, cache the value in this scope variable
+    * @type {!number} */
+  this.navWidth = 0;
+
 
   // need to wait for transcluded content to be available
   var deregisterTabWatch = this._$scope.$watch(function () {
@@ -264,14 +275,12 @@ MdNavBarController.prototype._getElements = function () {
  * @private
  * @returns {!number}
  */
-MdNavBarController.prototype._getNavWidth = function () {
-  var tabs      = this._getElements().tabs;
-  var contWidth = 0;
-  angular.forEach(tabs, function (tab) {
-    contWidth  += tab.offsetWidth;
-  });
-
-  return contWidth;
+MdNavBarController.prototype._updateStates = function () {
+  var elements     = this._getElements();
+  var isNavWidthNotNull = !!elements.canvas && !!elements.canvas.scrollWidth;
+  if(isNavWidthNotNull) this.navWidth = elements.canvas.scrollWidth;
+  var isParentWidthNotNull = !!elements.parent && !!elements.parent.clientWidth;
+  if(isParentWidthNotNull) this.parentWidth = elements.parent.clientWidth;
 }
 
 /**
@@ -279,14 +288,7 @@ MdNavBarController.prototype._getNavWidth = function () {
  * @returns {!boolean}
  */
 MdNavBarController.prototype.canPaginate = function () {
-  var elements             = this._getElements();
-  var navWidth             = this._getNavWidth();
-
-  var parentWidth          = 0;
-  var isParentWidthNotNull = !!elements.parent && !!elements.parent.clientWidth;
-  if(isParentWidthNotNull) parentWidth = elements.parent.clientWidth;
-
-  return navWidth > parentWidth;
+  return this.navWidth > this.parentWidth;
 }
 
 /**
@@ -294,6 +296,7 @@ MdNavBarController.prototype.canPaginate = function () {
  * and the pagination styles defined by the element.
  */
 MdNavBarController.prototype.updatePagination = function () {
+  this._updateStates();
   this.shouldPaginate = this.canPaginate();
 
   // Setting the style.
@@ -321,8 +324,7 @@ MdNavBarController.prototype.updatePagination = function () {
  * @returns {!boolean}
  */
 MdNavBarController.prototype.canPageBack = function () {
-  var elements = this._getElements();
-  return elements.canvas.scrollWidth + this.left <= this._getNavWidth();
+  return this.navWidth + this.left < this.navWidth;
 }
 
 /**
@@ -330,8 +332,7 @@ MdNavBarController.prototype.canPageBack = function () {
  * @returns {!boolean}
  */
 MdNavBarController.prototype.canPageForward = function () {
-  var elements = this._getElements();
-  return elements.canvas.scrollWidth + this.left >= elements.parent.clientWidth;
+  return this.navWidth + this.left >= this.parentWidth;
 }
 
 /**
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -96,77 +96,76 @@ angular.module('material.components.navBar', ['material.core', 'material.compone
  *****************************************************************************/
 
 function MdNavBar($mdAria, $mdUtil, $$mdSvgRegistry) {
-  return {
-    restrict: 'E',
-    transclude: true,
-    controller: MdNavBarController,
-    controllerAs: 'ctrl',
-    bindToController: true,
-    scope: {
-      'mdSelectedNavItem': '=?',
-      'navBarAriaLabel': '@?',
-    },
-    template:
-      '<div class=\'md-nav-wrapper\'>'+
-      '<div class=\'md-nav-pagination\'>'+
-                      '<md-prev-button ' +
-              'aria-label="Previous Page" ' +
-              'aria-disabled="{{!ctrl.canPageBack()}}" ' +
-              'ng-disabled="{{!ctrl.canPageBack()}}" ' +
-              'ng-if="ctrl.shouldPaginate" ' +
-              'ng-click="ctrl.previousPage()"> ' +
-            '<md-icon md-svg-src="'+ $$mdSvgRegistry.mdTabsArrow +'"></md-icon> ' +
-          '</md-prev-button> ' +
+    return {
+        restrict: 'E',
+        transclude: true,
+        controller: MdNavBarController,
+        controllerAs: 'ctrl',
+        bindToController: true,
+        scope: {
+            'mdSelectedNavItem': '=?',
+            'navBarAriaLabel': '@?',
+        },
+        template: '<div class=\'md-nav-wrapper\'>' +
+            '<div class=\'md-nav-pagination\'>' +
+            '<md-prev-button ' +
+            'aria-label="Previous Page" ' +
+            'aria-disabled="{{!ctrl.canPageBack()}}" ' +
+            'ng-disabled="{{!ctrl.canPageBack()}}" ' +
+            'ng-if="ctrl.shouldPaginate" ' +
+            'ng-click="ctrl.previousPage()"> ' +
+            '<md-icon md-svg-src="' + $$mdSvgRegistry.mdTabsArrow + '"></md-icon> ' +
+            '</md-prev-button> ' +
             '<md-next-button ' +
-              'aria-label="Next Page" ' +
-              'aria-disabled="{{!ctrl.canPageForward()}}" ' +
-              'ng-disabled="{{!ctrl.canPageForward()}}" ' +
-              'ng-if="ctrl.shouldPaginate" ' +
-              'ng-click="ctrl.nextPage()"> ' +
-            '<md-icon md-svg-src="'+ $$mdSvgRegistry.mdTabsArrow +'"></md-icon> ' +
-          '</md-next-button> ' +
-      '</div>'+
-      '<div class="md-nav-bar" md-swipe-left="ctrl.nextPage()" md-swipe-right="ctrl.previousPage()">' +
-        '<nav role="navigation">' +
-          '<ul class="_md-nav-bar-list" layout="row" ng-transclude role="listbox"' +
+            'aria-label="Next Page" ' +
+            'aria-disabled="{{!ctrl.canPageForward()}}" ' +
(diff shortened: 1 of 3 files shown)


menosprezzi:
I signed it!

menosprezzi:
Hello @ThomasBurleson @jelbourn @ErinCoughlan @KarenParker @DerekLouie devs. Take a look :smile:

ErinCoughlan:
@gmoothart (original component author): Please take a look at this.

src/components/navBar/navBar.js
```diff
@@ -175,6 +221,132 @@ function MdNavBarController($element, $scope, $timeout, $mdConstant) {
...
+    * and the pagination styles defined by the element.
+    */
+    function updatePagination () {
+        self.shouldPaginate = shouldPaginate();
+
+        // Setting the style.
+        var rootElement = $element[0];
+        var prevButton = angular.element(rootElement.querySelector('md-prev-button'));
+        var nextButton = angular.element(rootElement.querySelector('md-next-button'));
+        console.log($element.css('background-color'));
```
ErinCoughlan: Remove console.log

src/components/navBar/navBar.js
```diff
@@ -115,11 +137,34 @@ function MdNavBar($mdAria) {
           '</ul>' +
         '</nav>' +
         '<md-nav-ink-bar></md-nav-ink-bar>' +
+      '</div>'+
       '</div>',
     link: function(scope, element, attrs, ctrl) {
-      if (!ctrl.navBarAriaLabel) {
-        $mdAria.expectAsync(element, 'aria-label', angular.noop);
-      }
+        if (!ctrl.navBarAriaLabel) {
```
ErinCoughlan: Indentation is off here and below. Try to keep the styles consistent throughout. (Spacing around if statements, indentation, lining up vars, etc)

src/components/navBar/navBar.js
menosprezzi: Thanks! Fixed.

src/components/navBar/navBar.js
```diff
@@ -34,6 +34,9 @@ angular.module('material.components.navBar', ['material.core'])
  * listbox, according to
  * https://www.w3.org/TR/wai-aria-practices/#Site_Navigator_Tabbed_Style
  *
+ * In small screens that not support the entire width of this element, this directive works like md-tabs,
```
ErinCoughlan: nit: reword and remove reference to me-tabs. It doesn't add anything to the behavior for small screens. Something like:

If the width of the mdNavBar is greater than the available screen width, pagination will be used. Pagination supports left and right arrows keys as well as swipe gestures.

src/components/navBar/navBar.js
```diff
@@ -175,6 +221,131 @@ function MdNavBarController($element, $scope, $timeout, $mdConstant) {
...
+    /** @type {Function} */
+    this.previousPage       = previousPage;
+
+    /** @type {Function} */
+    this.canPageBack        = canPageBack;
+
+    /** @type {Function} */
+    this.canPageForward     = canPageForward;
+
+    /** @type {Number} */
```
ErinCoughlan: nit: number type is lowercase

src/components/navBar/navBar.js
```diff
@@ -175,6 +221,131 @@ function MdNavBarController($element, $scope, $timeout, $mdConstant) {
       deregisterTabWatch();
     }
   });
+
+    // Responsivity actions
+
+    /** @type {Object} */
```
ErinCoughlan: This file has explicit nullability for Objects and Functions. Please add them. If it can be null, the type should be ?Object. If it can't be null, the type should be !Object. (Ditto for functions)

src/components/navBar/navBar.js
```diff
@@ -175,6 +221,131 @@ function MdNavBarController($element, $scope, $timeout, $mdConstant) {
...
+
+    /** @type {Function} */
+    this.canPageForward     = canPageForward;
+
+    /** @type {Number} */
+    this.left               = 0;
+
+
+    /**
+    * Gathers references to all of the DOM elements used by this controller.
```
ErinCoughlan: nit: indentation

src/components/navBar/navBar.js
menosprezzi: Sorry! Fixed

src/components/navBar/navBar.js
```diff
@@ -134,47 +178,183 @@ function MdNavBar($mdAria) {
...
 
-  /** @private @const {!angular.Scope} */
-  this._$scope = $scope;
+    /** @type {Function} */
+    this.canPageBack = canPageBack;
 
-  /** @private @const {!Object} */
-  this._$mdConstant = $mdConstant;
+    /** @type {Function} */
+    this.canPageForward = canPageForward;
```
ErinCoughlan: These sound like booleans, not at least a function that returns a Boolean.

If it's the latter, I'd change the type to function(): boolean to be explicit.

src/components/navBar/navBar.js
```diff
@@ -134,47 +178,183 @@ function MdNavBar($mdAria) {
...
+     * @returns {boolean}
+     */
+    function canPageBack() {
+        var elements = getElements();
+        return elements.canvas.scrollWidth + this.left <= getNavWidth();
+    }
+
+    /**
+     * Determines whether or not the right pagination arrow should be enabled.
+     * @returns {*|boolean}
```
ErinCoughlan: Do you really want it to return anything? Should this just be boolean like the above function?

src/components/navBar/navBar.js
```diff
@@ -134,47 +178,183 @@ function MdNavBar($mdAria) {
...
+                'left': self.left + 'px'
+            });
+        }
+    }
+
+    /**
+     * Calculate the real width size of the mdNavBar using each tab of it.
+     * @returns {number}
+     */
+    function getNavWidth() {
```
ErinCoughlan: Looks like most of this file is using prototypes for methods. Watch consistency.

src/components/navBar/navBar.js
```diff
@@ -134,47 +178,183 @@ function MdNavBar($mdAria) {
...
+        var tabs = getElements().tabs
```
ErinCoughlan: Missing semicolon

src/components/navBar/navBar.js
```diff
@@ -134,47 +178,183 @@ function MdNavBar($mdAria) {
...
+        var contWidth = 0;
+        for (var i = 0; i < tabs.length; i++) {
```
ErinCoughlan: How about a forEach loop instead? You only use the index to get the element.

tabs.forEach(function(tab) {
  contWidth += tab.offset;
});

src/components/navBar/navBar.js
```diff
@@ -134,47 +178,183 @@ function MdNavBar($mdAria) {
...
+    }
+
+    /**
+     * Determines if pagination is necessary to display the tabs within the available space.
+     * @returns {boolean}
+     */
+    function shouldPaginate() {
+        var elements = getElements();
+        var navWidth = getNavWidth();
+        var parentWidth = elements.parent != null && elements.parent.clientWidth != null ? elements.parent.clientWidth : 0;
```
ErinCoughlan: This is a confusing line. Consider simplifying by using a real if statement and using !!elements.parent for the nullability checks.

src/components/navBar/navBar.js
menosprezzi: Sorry, i've used the same notation that has in mdTabs

src/components/navBar/navBar.js
```diff
@@ -206,58 +386,63 @@ MdNavBarController.prototype._initTabs = function() {
...
  */
-MdNavBarController.prototype._getTabs = function() {
-  var linkArray = Array.prototype.slice.call(
-      this._navBarEl.querySelectorAll('.md-nav-item'));
-  return linkArray.map(function(el) {
-    return angular.element(el).controller('mdNavItem')
-  });
+MdNavBarController.prototype._getTabs = function () {
+    var linkArray = Array.prototype.slice.call(
+        this._navBarEl.querySelectorAll('.md-nav-item'));
```
ErinCoughlan: Most of these lines aren't really changing. If you fix the indentation, I think the diff will look less overwhelming (and you won't be listed on every line to blame).

src/components/navBar/navBar.js
```diff
@@ -334,107 +519,106 @@ MdNavBarController.prototype.onBlur = function() {
...
-            .text().trim();
+    return {
+        restrict: 'E',
+        require: ['mdNavItem', '^mdNavBar'],
+        controller: MdNavItemController,
+        bindToController: true,
+        controllerAs: 'ctrl',
+        replace: true,
+        transclude: true,
+        template: '<li class="md-nav-item" role="option" aria-selected="{{ctrl.isSelected()}}">' +
```
ErinCoughlan: This is really unreadable now. Consider adding spaces inside the single quotes so formatting won't be lost. Alternatively, don't format unchanged lines.

src/components/navBar/navBar.scss
```diff
@@ -59,3 +58,39 @@ md-nav-extra-content {
   min-height: 48px;
   padding-right: 12px;
 }
+
+/** Responsive pagination **/
+.md-nav-wrapper {
+    border-color: rgba(0,0,0,0.12);
```
ErinCoughlan: Indentation in css also should be 2, not 4.

ErinCoughlan:
@menosprezzi I don't see any tests here. Please add them for all added functionality.

src/components/navBar/navBar.js
menosprezzi: Checked!

src/components/navBar/navBar.js
menosprezzi: Checked!

src/components/navBar/navBar.js
menosprezzi: Checked!

src/components/navBar/navBar.scss
menosprezzi: Checked!

src/components/navBar/navBar.js
```diff
@@ -103,23 +106,68 @@ function MdNavBar($mdAria) {
...
+                  '<md-nav-ink-bar></md-nav-ink-bar> ' +
+                '</div> ' +
+              '</div> ',
+    link: function (scope, element, attrs, ctrl) {
       if (!ctrl.navBarAriaLabel) {
         $mdAria.expectAsync(element, 'aria-label', angular.noop);
       }
+
+      var observer = new MutationObserver(function (mutations) {
+        ctrl.updatePagination();
```
crisbeto: I feel like this in particular might destroy IE performance, because it causes a reflow, every time the DOM is manipulated, which is A LOT. It's because the `updatePagination` method calls `canPaginate` which has a couple of calls to `clientWidth`, each of which causes a reflow. Not to mentioned that `canPaginate` is also being called a few times per digest loop inside of the template.

We have pretty much the same issue with the tabs component. See https://github.com/angular/material/issues/8329

menosprezzi:
@ErinCoughlan Sorry the trouble! Maybe now looks fine.
But about the tests, I still don't know how work with it. I'm going to read about that and implement soon.
Thanks!

src/components/navBar/navBar.js
menosprezzi: Yeees :disappointed: I implemented it based in the mdTabs directive.
Is there any way to solve it?

crisbeto: I'm not sure there's a **good** way to solve it, since we still need some calls to `clientWidth`. Trying to avoid the `canPaginate` calls as much as possible, as well as caching the results` should help.

menosprezzi: I made some improvements caching values from client, like you suggested.
I'll keep the focus to try changes in MutationObserver to another solution.
Thanks!

crisbeto: I'd say that the mutation observer is probably better than a $watch (we did a similar migration from $watch to MutationObserver for the tooltips), but it would be good if it didn't fire as often. For example, try seeing if you can get rid of any of the options in the `config`, without breaking functionality. In addition, you can debounce the callback so it doesn't fire too much if a few mutations happen together, but you might want to play around and see what works, since it might make the UI look unresponsive.

src/components/navBar/navBar.js
```diff
@@ -165,36 +220,163 @@ function MdNavBarController($element, $scope, $timeout, $mdConstant) {
...
+/**
+ * Gathers references to all of the DOM elements used by this controller.
+ * @private
+ * @returns {!Object}
+ */
+MdNavBarController.prototype._getElements = function () {
+  var elements = {};
+  var node = this._$element[0];
+  elements.parent = node.parentElement;
+  elements.canvas = node.querySelector('.md-nav-bar');
```
gmoothart: I think the naming is confusing here, since this is a <div> not a <canvas>

gmoothart:
Can you describe a use-case where you need scrolling - i.e., where the native browser scrolling is not sufficient? Most cases I can think of should really be using md-tabs instead. In general I want to keep the nav-bar as light as we can.

Also, can you add scrolling to the example in demoBasicUsage so I can kick the tires?

menosprezzi:
@gmoothart, I was working in a project that needs a nav-bar and when i switched to mobile I realized that native scrolling was breaking the UI. Why not tabs: in the project, I don't need to change the content itself. Just some buttons with ngClick arranged as a nav-bar (with responsivity) can take the work.

ThomasBurleson:
We will not be adding these features to `md-nav-bar`. If you need tab pagination and navigation buttons, use the `md-tabs`. 

> Note: mdTabs can be used as a simple bar without content body items; see online demo.

menosprezzi:
@ThomasBurleson, but md-nav-bat will still not be responsive. The question is not personal. But Angular Material would have a ui-broker component.

EladBezalel:
@menosprezzi my first expression was: `Yes, This sounds reasonable`
But after some research i found out that by the spec it shouldn't be supported.

According the [Bottom navigation component spec](https://material.google.com/components/bottom-navigation.html):
first
>**Don't.**
>Avoid scrollable content in the bottom navigation bar.

Scrollable -> pagination

sec
>**Don't.**
>Avoid using more than five destinations in bottom navigation as tap targets will be situated too close to one another.

menosprezzi:
@EladBezalel, this make sense. Perhaps, I've not understood the concept of this component.

Issue #5924, opened 2015-11-26
rc5 md-dialog fullscreen works only for small screen


stunaz:
why add the media condition `@media (max-width: $layout-breakpoint-sm)`

```css
@media (max-width: $layout-breakpoint-sm) {
  md-dialog.md-dialog-fullscreen {
    min-height: 100%;
    min-width: 100%;
    border-radius: 0;
  }
}
```

This means we can't force a full screen on a big screen

DevVersion:
I think this is by intention. See at the Material Specs (http://www.google.com/design/spec/components/dialogs.html#dialogs-full-screen-dialogs)

References #5793

PR #9032, opened 2016-07-14
Remove $mdMedia references from mdDialog example

SpikesCafe-google:
cleanup $mdMedia references in example. fullscreen only applies when the screen is xs and sm even when set to true.

See also https://github.com/angular/material/issues/5924

### DIFF
--- a/src/components/dialog/demoBasicUsage/script.js
+++ b/src/components/dialog/demoBasicUsage/script.js
@@ -62,7 +62,7 @@ angular.module('dialogDemo1', ['ngMaterial'])
       parent: angular.element(document.body),
       targetEvent: ev,
       clickOutsideToClose: true,
-      fullscreen: true  // only applies for screen size xs and sm
+      fullscreen: true // Fullscreen only applies for -xs, -sm breakpoints.
     })
     .then(function(answer) {
       $scope.status = 'You said the information was "' + answer + '".';
--- a/src/components/dialog/demoBasicUsage/script.js
+++ b/src/components/dialog/demoBasicUsage/script.js
@@ -2,7 +2,6 @@ angular.module('dialogDemo1', ['ngMaterial'])
 
 .controller('AppCtrl', function($scope, $mdDialog, $mdMedia) {
   $scope.status = '  ';
-  $scope.customFullscreen = $mdMedia('xs') || $mdMedia('sm');
 
   $scope.showAlert = function(ev) {
     // Appending dialog to document.body to cover sidenav in docs app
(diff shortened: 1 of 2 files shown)


src/components/dialog/demoBasicUsage/script.js
```diff
@@ -57,30 +56,19 @@ angular.module('dialogDemo1', ['ngMaterial'])
...
-
     $mdDialog.show({
       controller: DialogController,
       templateUrl: 'dialog1.tmpl.html',
       parent: angular.element(document.body),
       targetEvent: ev,
-      clickOutsideToClose:true,
-      fullscreen: useFullScreen
+      clickOutsideToClose: true,
+      fullscreen: true  // only applies for screen size xs and sm
```
DevVersion: This is a reasonable change. Please update then comment to (because this will actually show up in the demo)

```js
  fullscreen: true // Fullscreen only applies for -xs, -sm breakpoints.
```

src/components/dialog/demoBasicUsage/script.js
```diff
@@ -57,30 +56,19 @@ angular.module('dialogDemo1', ['ngMaterial'])
...
-
     $mdDialog.show({
       controller: DialogController,
       templateUrl: 'dialog1.tmpl.html',
       parent: angular.element(document.body),
       targetEvent: ev,
-      clickOutsideToClose:true,
-      fullscreen: useFullScreen
+      clickOutsideToClose: true,
+      fullscreen: true // Fullscreen only applies for -xs, -sm breakpoints.
```
DevVersion: Edit: Just rethought again. We don't want to force the users on -xs, -sm breakpoints to always use the Fullscreen Dialog.

Those should be able to toggle that functionality within the `checkbox` in the demo HTML.

SpikesCafe-google: Sure! There are several other dialogs in this example that don't use fullscreen. My personal opinion is that they were sufficient to show what mdDialog is about without fullscreen. But I'll make that change.

SpikesCafe-google:
I'm trying to change the index.html

DevVersion:
@SpikesCafe-google Cool thanks. While you're at it, can you probably set the default fullscreen value to `false`?

DevVersion:
This PR looks good to me. Thanks

SpikesCafe-google:
Thanks! I wanted to update the HTML to clarify a little and not hide the checkbox (a little too tricky for an example), but I'm not familiar enough with github to do it. Something like this after the prerendered checkbox:

    <md-checkbox ng-model="customFullscreen" aria-label="Fullscreen Custom Dialog">Use full screen mode for custom dialog</md-checkbox>
    <p class="md-caption">Allows the dialog to consume all available space on small devices</p>

As a replacement for the old fullscreen checkbox.

DevVersion:
I like the idea. Can't you try to edit the file here?
> https://github.com/SpikesCafe-google/material/blob/8d5354ed24d0994a84d42f873269bd328d55b078/src/components/dialog/demoBasicUsage/index.html

SpikesCafe-google:
It says I must be on a branch to make or propose changes to this file.

DevVersion:
@SpikesCafe-google Ah gotcha. Here is a reference to your branch. 
> https://github.com/SpikesCafe-google/material/blob/patch-1/src/components/dialog/demoBasicUsage/index.html

PR #9034, opened 2016-07-14
chore: change closure module prefix

jelbourn:
I'm changing the Closure namespaces from `ng.material.xxxx` to `ngmaterial.xxxx`. As it is now, the `ng.` prefix collides with some other code.


ThomasBurleson:
@jelbourn - lgtm

PR #9041, opened 2016-07-15
refactor(select): Separate option value test into multiple.

topherfangio:
Refactor the various `value=` and `ng-value=` option test into a
separate test for each value type so we can more easily see if a
particular one fails.

Related to #9014.


crisbeto:
I think that you can do something like `expect(whatever, <value goes here>` which will allow you to see which value failed, instead of making a separate test for each.

PR #9049, opened 2016-07-16
fix(datepicker): add missing import

crisbeto:
The datepicker is using $mdAria, however it wasn't importing it, causing an error in some cases.


crisbeto:
Closing this one and merging the fix into #9050.

PR #9050, opened 2016-07-16
fix(datepicker): mark the input as invalid on submit, add missing import

### LINKED ISSUE #8411: md-datepicker ignoring required attribute on pristine input
Codepen is here: http://codepen.io/astynax777/pen/ONqQOY

You'll notice that if you submit the form without first selecting and clearing a date, the validation does not fire on the md-datepicker.

I would like to know a workaround for this in the meantime.

AngularJS stack: 1.5.5
Material: 1.1.0-rc4

crisbeto:
* Fixes invalid datepickers not being highlighted when their parent form is submitted.
* Fixes $mdAria not being imported and causing a JS error occasionally.

Fixes #8411.

### DIFF
--- a/src/components/datepicker/js/datepickerDirective.js
+++ b/src/components/datepicker/js/datepickerDirective.js
@@ -53,7 +53,7 @@
    * </hljs>
    *
    */
-  function datePickerDirective($$mdSvgRegistry) {
+  function datePickerDirective($$mdSvgRegistry, $mdUtil, $mdAria) {
     return {
       template: function(tElement, tAttrs) {
         // Buttons are not in the tab order because users can open the calendar via keyboard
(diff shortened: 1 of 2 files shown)


src/components/datepicker/js/datepickerDirective.js
```diff
@@ -53,7 +53,7 @@
    * </hljs>
    *
    */
-  function datePickerDirective($$mdSvgRegistry) {
+  function datePickerDirective($$mdSvgRegistry, $mdUtil) {
```
DevVersion: Isn't that the same as in #9049. You could probably drop #9049 and just use this one.

crisbeto: #9049 is for ngAria, but yeah it makes sense to merge these two.

Issue #8630, opened 2016-06-02
Enhancement to open multiple dialogs using $mdDialog


denver-HJS:
Hello,

I am inquiring to find out whether or not there are plans to add functionality to the $mdDialog service for showing multiple dialogs. I am aware of [this thread](https://github.com/angular/material/issues/698) but I'd like to rekindle the discussion.

My specific need is that my application requires a virtual keyboard. I implemented this virtual keyboard using a wrapper service around $mdDialog (and a separate directive). This worked flawlessly...up until the point I tried to apply my virtual keyboard to an input that's inside of a dialog. Once the user clicks the input within the dialog it goes away and up comes the keyboard.

**1)** Is there any chance that having multiple dialogs will be implemented in the future?
**2)** Does anyone have suggestions for working around this limitation?

I found the options $mdDialog provides to be extremely useful, so it's disappointing to rearrange my architecture and try to recreate the niceties myself.

```
showKeyboard(keyboardType, inputElement, modelUpdateEvent, customValidation) {
      var s = this;
      let positionKeyboard = () => {
        $('#keyboard-dialog').position({
          of: $(inputElement),
          my: "right top",
          at: "right bottom",
          collision: "flipfit flipfit"
        });
        s._isDragging = false;
        let dragDelay = 1000;
        let dragTimeout = undefined;
        $('#keyboard-dialog').on("mousedown", () => {
          if(s._isDragging == false) {
            dragTimeout = this._$timeout(() => {
              s._isDragging = true;
              s._dragStartTime = moment();
              let draggingMask = $.parseHTML(`<div class="drag-mask"></div>`);
              $(draggingMask).css({
                position: 'absolute',
                width: '100%',
                height: '100%',
                top: '0px',
                left: '0px',
                zIndex: 100
              });
              $('#keyboard-dialog').append(draggingMask);
              $('#keyboard-dialog').css({
                border: '3px solid rgb(86, 180, 239)',
                'box-shadow': '0px 1px 3px rgba(0, 0, 0, 0.05) inset, 0px 0px 8px rgba(82, 168, 236, 0.6)'
              });
              $('#keyboard-dialog').draggable();
            }, dragDelay);
          }
        });

        $('#keyboard-dialog').on("mouseup", () => {
          if(angular.isDefined(dragTimeout)) {
            s._$timeout.cancel(dragTimeout);
            dragTimeout = undefined;
          }
          let now = moment();
          let timespan = moment.duration(now.diff(s._dragStartTime));
[...]

PR #9053, opened 2016-07-16
feat(interimElement): properly handle multiple interims.

### LINKED ISSUE #8624: interimElement: finish animation of previous element
The `$$interimElement` factory currently only supports single `interimElements` at the same time. 

---
This is intentionally, but it doesn't work properly for some components.

- For example, if you quickly open some toasts, the first toasts will overlay each other<br/> (See https://github.com/angular/material/issues/4822, https://github.com/angular/material/issues/5193)

This is caused by the `$$interimElement`'s item stack. 

Once a second interim element is opening, we cancel the previous interim element and show the new one after the animation has **started**

This is not correct, because we should wait for the animation to be **finished**

DevVersion:
* Adds support for multiple interim elements (like dialog)
* When single interim (default) is enabled, then interims should hide properly (as in toasts).

---
When opening interim elements, by rapidly clicking on the open button as an example, the toasts / interims will overlap. But those shouldn't overlap. This is against the specs.

> See https://material.angularjs.org/HEAD/demo/toast

@ThomasBurleson Nothing really changed for the components. It just fixed the issue with overlapping and opens the door of having multiple interims (this was just included into that fix)

Fixes #8624. References #9166 References #8630.

### DIFF
--- a/src/components/menu/js/menuServiceProvider.js
+++ b/src/components/menu/js/menuServiceProvider.js
@@ -34,7 +34,7 @@ function MenuProvider($$interimElementProvider) {
       disableParentScroll: true,
       skipCompile: true,
       preserveScope: true,
-      skipHide: true,
+      multiple: true,
       themable: true
     };
 
--- a/src/core/services/interimElement/interimElement.js
+++ b/src/core/services/interimElement/interimElement.js
@@ -257,7 +257,12 @@ function InterimElementProvider() {
        * A service used to control inserting and removing an element into the DOM.
        *
        */
-      var service, stack = [];
+
+      var service;
+
+      var showPromises = []; // Promises for the interim's which are currently opening.
+      var hidePromises = []; // Promises for the interim's which are currently hiding.
+      var showingInterims = []; // Interim elements which are currently showing up.
 
       // Publish instance $$interimElement service;
       // ... used as $mdDialog, $mdToast, $mdMenu, and $mdSelect
@@ -286,25 +291,34 @@ function InterimElementProvider() {
       function show(options) {
         options = options || {};
         var interimElement = new InterimElement(options || {});
+
         // When an interim element is currently showing, we have to cancel it.
         // Just hiding it, will resolve the InterimElement's promise, the promise should be
         // rejected instead.
-        var hideExisting = !options.skipHide && stack.length ? service.cancel() : $q.when(true);
+        var hideAction = options.multiple ? $q.when(true) : $q.all(showPromises);
 
-        // This hide()s only the current interim element before showing the next, new one
-        // NOTE: this is not reversible (e.g. interim elements are not stackable)
+        if (!options.multiple) {
+          // Wait for all opening interim's to finish their transition.
+          hideAction = hideAction.then(function() {
+            // Wait for all closing and showing interim's to be completely closed.
+            var promiseArray = hidePromises.concat(showingInterims.map(service.cancel));
+            return $q.all(promiseArray);
+          });
+        }
 
-        hideExisting.finally(function() {
+        var showAction = hideAction.then(function() {
 
-          stack.push(interimElement);
-          interimElement
+          return interimElement
             .show()
-            .catch(function( reason ) {
-              //$log.error("InterimElement.show() error: " + reason );
-              return reason;
+            .catch(function(reason) { return reason; })
+            .finally(function() {
+              showPromises.splice(showPromises.indexOf(showAction), 1);
+              showingInterims.push(interimElement);
             });
 
         });
+
+        showPromises.push(showAction);
 
         // Return a promise that will be resolved when the interim
         // element is hidden or cancelled...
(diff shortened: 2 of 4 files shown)


src/components/menu/js/menuServiceProvider.js
```diff
@@ -34,7 +34,7 @@ function MenuProvider($$interimElementProvider) {
       disableParentScroll: true,
       skipCompile: true,
       preserveScope: true,
-      skipHide: true,
```
EladBezalel: Is it documented?

DevVersion: It never was, and was only for internal use.

src/core/services/interimElement/interimElement.js
```diff
@@ -383,26 +408,27 @@ function InterimElementProvider() {
...
+
+        if (parentEl) {
+          // Try to find the interim in the stack which corresponds to the supplied DOM element.
+          var filtered = showingInterims.filter(function(entry) {
+            return entry.options.element[0] === parentEl;
+          });
+
+          // Note: This function might be called when the element already has been removed,
+          // in which case we won't find any matches.
+          if (filtered.length > 0) {
```
EladBezalel: isn't `filtered.length` is enough?

EladBezalel:
Overall lgtm

src/core/services/interimElement/interimElement.js
DevVersion: It should be. I will change it tomorrow.

ThomasBurleson:
Dangerous changes. Must be manually validated (aka user testing) and needs more unit tests.

src/core/services/interimElement/interimElement.js
```diff
@@ -286,26 +291,35 @@ function InterimElementProvider() {
...
         var interimElement = new InterimElement(options || {});
+
         // When an interim element is currently showing, we have to cancel it.
         // Just hiding it, will resolve the InterimElement's promise, the promise should be
         // rejected instead.
-        var hideExisting = !options.skipHide && stack.length ? service.cancel() : $q.when(true);
-
-        // This hide()s only the current interim element before showing the next, new one
-        // NOTE: this is not reversible (e.g. interim elements are not stackable)
+        var hideAction = options.multiple ? $q.when(true) : $q.all(showPromises);
```
EladBezalel: You can use `$q.resolve` instead of `$q.when(true)`

src/core/services/interimElement/interimElement.js
```diff
@@ -286,26 +291,35 @@ function InterimElementProvider() {
...
+        var showAction = hideAction.then(function() {
 
-          stack.push(interimElement);
-          interimElement
+          return interimElement
             .show()
-            .catch(function( reason ) {
-              //$log.error("InterimElement.show() error: " + reason );
-              return reason;
+            .catch(function(reason) { return reason; })
```
EladBezalel: Why are we muting the errors and resolving the promise?

src/core/services/interimElement/interimElement.js
```diff
@@ -325,27 +339,30 @@ function InterimElementProvider() {
...
-        if ( !stack.length ) return $q.when(reason);
         options = options || {};
 
         if (options.closeAll) {
-          var promise = $q.all(stack.reverse().map(closeElement));
-          stack = [];
-          return promise;
+          // We have to make a shallow copy of the array, because otherwise the map will break.
+          return $q.all(showingInterims.slice().reverse().map(closeElement));
         } else if (options.closeTo !== undefined) {
```
EladBezalel: else is redundant

src/core/services/interimElement/interimElement.js
```diff
@@ -325,27 +339,30 @@ function InterimElementProvider() {
...
       function hide(reason, options) {
-        if ( !stack.length ) return $q.when(reason);
         options = options || {};
 
         if (options.closeAll) {
-          var promise = $q.all(stack.reverse().map(closeElement));
-          stack = [];
-          return promise;
+          // We have to make a shallow copy of the array, because otherwise the map will break.
+          return $q.all(showingInterims.slice().reverse().map(closeElement));
```
EladBezalel: Why do we reverse it?

src/core/services/interimElement/interimElement.js
```diff
@@ -325,27 +339,30 @@ function InterimElementProvider() {
...
         } else if (options.closeTo !== undefined) {
-          return $q.all(stack.splice(options.closeTo).map(closeElement));
+          return $q.all(showingInterims.slice(options.closeTo).map(closeElement));
         } else {
```
EladBezalel: else is redundant

src/core/services/interimElement/interimElement.js
DevVersion: We want to show them in reverse, last added interim should hide at first.

src/core/services/interimElement/interimElement.js
```diff
@@ -363,46 +380,76 @@ function InterimElementProvider() {
        *
        */
       function cancel(reason, options) {
```
EladBezalel: cancel, hide and show are pretty similar can't we make a more generic function?

DevVersion: I would rather not make big changes in the PR, since it's already big enough. Could be an extra PR.

src/core/services/interimElement/interimElement.js
```diff
@@ -363,46 +380,76 @@ function InterimElementProvider() {
...
+       */
+      function waitForInterim(callbackFn) {
+        return function() {
+          var fnArguments = arguments;
+
+          if (!showingInterims.length) {
+            // When there are still interim's opening, then wait for the first interim element to
+            // finish its open animation.
+            if (showPromises.length) {
+              return showPromises[0].finally(function () {
```
EladBezalel: `showPromises[0].finally` isn't the same as `$q.all` ?

src/core/services/interimElement/interimElement.js
DevVersion: Not really sure about this. I did not write this.

I assume it's because we are resolving / rejecting the promises twice, and this one is just for internal use.

The real promise is returned below.

src/core/services/interimElement/interimElement.js
```diff
@@ -363,46 +380,76 @@ function InterimElementProvider() {
...
         // promise from being logged to the $exceptionHandler.
         return interim.deferred.promise.catch(angular.noop);
       }
 
+      /**
+       * Creates a function to wait for at least one interim element to be available.
+       * @param callbackFn Function to be used as callback
+       * @returns {Function}
+       */
+      function waitForInterim(callbackFn) {
```
EladBezalel: why are we getting a callback function and not resolving a promise?

src/core/services/interimElement/interimElement.js
DevVersion: Intentionally. We only want to wait for the first promise.

src/core/services/interimElement/interimElement.js
DevVersion: This is function which waits for **one** interim element to be available. Doing it with promises means we need to repeat that logic multiple times (what I had before this part commit)

topherfangio:
Ping @ErinCoughlan for review.

DevVersion:
@ErinCoughlan @topherfangio - Ping again for review, want to push that PR a bit.

src/core/services/interimElement/interimElement.js
ErinCoughlan: Can you explain why you only care about the first promise? This is strange.

ekevoo:
Just adding a link to Keep support to "skipHide" with a deprecation warning #10438.

Issue #877, opened 2014-12-08
Changed start/endsymbols (interpolateProvider) not used in templates


josvos:
When the application has changed the start/endsymbols (in my case I used `$interpolateProvider.startSymbol('{$').endSymbol('$}');` for that), some elements, like dialogs, are not working properly, as the default symbols `{{` and `}}` are hardcoded in the templates.

PR #9054, opened 2016-07-16
chore(): sauce reduce connectivity errors.

DevVersion:
* Reduces the connectivity errors on SauceLabs.


DevVersion:
FYI: Paring with @jelbourn to fix #877

PR #9058, opened 2016-07-17
fix(input): icons not inheriting theme colors

crisbeto:
Fixes the `md-icon` instances inside of a `md-input-container` using the primary color, even though the container is e.g. `md-accent` or `md-warn`.

### DIFF
--- a/src/components/input/input-theme.scss
+++ b/src/components/input/input-theme.scss
@@ -45,9 +45,7 @@ md-input-container.md-THEME_NAME-theme {
     }
 
     &.md-input-focused {
-      label {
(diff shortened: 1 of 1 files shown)


crisbeto:
@EladBezalel Can you take a look at this? It seemed counter-intuitive that the input would have an "accent" color, but the icon would always stay with the "primary". I couldn't find anything about it in the spec.

EladBezalel:
I think it's been noted [here](https://material.google.com/components/text-fields.html#text-fields-single-line-text-field)
image

LGTM

PR #9071, opened 2016-07-19
fix(nav-bar): null check tabs when updating nav-bar

samelliottdlt:
It is possible that _getTabs returns an array of undefined when mapping over linkArray as .controller('mdNavItem') can return undefined if it cannot find the corresponding controller which will cause an error to be shown in the console during the digest cycle.

This change assures that we do not attempt to update the tabs in the case that tabs is an undefined array.

### DIFF
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -222,6 +222,11 @@ MdNavBarController.prototype._initTabs = function() {
 MdNavBarController.prototype._updateTabs = function(newValue, oldValue) {
   var self = this;
   var tabs = this._getTabs();
+
+  // this._getTabs can return null if nav-bar has not yet been initialized
+  if(!tabs)
+    return;
+
   var oldIndex = -1;
   var newIndex = -1;
   var newTab = this._getTabByName(newValue);
@@ -266,11 +271,12 @@ MdNavBarController.prototype._updateInkBarStyles = function(tab, newIndex, oldIn
  * @private
  */
 MdNavBarController.prototype._getTabs = function() {
-  var linkArray = Array.prototype.slice.call(
-      this._navBarEl.querySelectorAll('.md-nav-item'));
-  return linkArray.map(function(el) {
-    return angular.element(el).controller('mdNavItem');
-  });
+  var controllers = Array.prototype.slice.call(
+    this._navBarEl.querySelectorAll('.md-nav-item'))
+    .map(function(el) {
+      return angular.element(el).controller('mdNavItem')
+    });
+  return controllers.indexOf(undefined) ? controllers : null;
 };
 
 /**
--- a/src/components/navBar/navBar.spec.js
+++ b/src/components/navBar/navBar.spec.js
@@ -175,6 +175,17 @@ describe('mdNavBar', function() {
           expect(getTab('tab2').attr('ui-sref-opts'))
               .toBe('{"reload":true,"notify":true}');
         });
+
+    it('does not update tabs if tab controller is undefined', function() {
+      $scope.selectedTabRoute = 'tab1';
+
+      spyOn(Object.getPrototypeOf(ctrl), '_updateInkBarStyles');
+      spyOn(Object.getPrototypeOf(ctrl), '_getTabs').and.returnValue(null);
+      createTabs();
+
+      expect(ctrl._updateInkBarStyles)
+        .not.toHaveBeenCalled();
+    });
   });
 
   describe('inkbar', function() {


SamDLT:
Signed CLA

ThomasBurleson:
Please provide 1-2 unit tests for these. Thx @SamDLT .

SamDLT:
@ThomasBurleson added unit tests

src/components/navBar/navBar.js
```diff
@@ -211,22 +211,25 @@ MdNavBarController.prototype._updateTabs = function(newValue, oldValue) {
...
   var oldIndex = -1;
   var newIndex = -1;
-  var newTab = this._getTabByName(newValue);
-  var oldTab = this._getTabByName(oldValue);
 
-  if (oldTab) {
-    oldTab.setSelected(false);
-    oldIndex = tabs.indexOf(oldTab);
-  }
+  if (tabs) {
```
clshortfuse: To prevent nesting, we can change this to `if (!tabs) return;`. We can also move it up to before `oldIndex` and `newIndex` are initialized.

clshortfuse:
Fixes #9386 

Can you add that to the commit message?

src/components/navBar/navBar.js
```diff
@@ -211,22 +211,25 @@ MdNavBarController.prototype._updateTabs = function(newValue, oldValue) {
...
+    var newTab = this._getTabByName(newValue);
+    var oldTab = this._getTabByName(oldValue);
 
-  if (newTab) {
-    newTab.setSelected(true);
-    newIndex = tabs.indexOf(newTab);
-  }
+    if (oldTab) {
```
gmoothart: can you change the indenting back to +2? Here and elsewhere

src/components/navBar/navBar.js
gmoothart: yes, can you return immediately and as soon as possible?

src/components/navBar/navBar.js
```diff
@@ -253,11 +256,12 @@ MdNavBarController.prototype._updateInkBarStyles = function(tab, newIndex, oldIn
...
-      this._navBarEl.querySelectorAll('.md-nav-item'));
-  return linkArray.map(function(el) {
-    return angular.element(el).controller('mdNavItem')
-  });
+    var linkArray = Array.prototype.slice.call(
+      this._navBarEl.querySelectorAll('.md-nav-item'))
+      .map(function(el) {
+          return angular.element(el).controller('mdNavItem')
+      });
+    return linkArray.indexOf() === -1 ? linkArray : null;
```
gmoothart: `linkArray.indexOf(undefined)` reads a little better to me.

src/components/navBar/navBar.spec.js
```diff
@@ -156,6 +156,17 @@ describe('mdNavBar', function() {
 
       expect($scope.selectedTabRoute).toBe('tab2');
     });
+
+    it('does not update tabs if tab controller is undefined', function() {
+      $scope.selectedTabRoute = 'tab1';
+
+      spyOn(Object.getPrototypeOf(ctrl), '_updateInkBarStyles');
```
gmoothart: is `Object.getPrototypeOf` necessary? I'm surprised you can't just do `spyOn(ctrl, '_upateInkBarStyles');`

src/components/navBar/navBar.js
```diff
@@ -253,11 +256,12 @@ MdNavBarController.prototype._updateInkBarStyles = function(tab, newIndex, oldIn
  * @private
  */
 MdNavBarController.prototype._getTabs = function() {
-  var linkArray = Array.prototype.slice.call(
-      this._navBarEl.querySelectorAll('.md-nav-item'));
-  return linkArray.map(function(el) {
-    return angular.element(el).controller('mdNavItem')
-  });
+    var linkArray = Array.prototype.slice.call(
```
gmoothart: can you unindent this back to +2?

gmoothart: (nit) this isn't a linkArray anymore, it's an array of controllers

SamDLT:
@gmoothart @clshortfuse Thanks for taking a look at this guys. I'll address the comments either tomorrow or Monday.

src/components/navBar/navBar.js
ThomasBurleson: `controllers` instead of `linkArray` plz.

src/components/navBar/navBar.js
```diff
@@ -209,6 +209,10 @@ MdNavBarController.prototype._initTabs = function() {
 MdNavBarController.prototype._updateTabs = function(newValue, oldValue) {
   var self = this;
   var tabs = this._getTabs();
+
+  if(!tabs)
+    return;
```
gmoothart: can you add a comment here explaining why tabs might be null?

src/components/navBar/navBar.spec.js
gmoothart: Pretty sure the Object.getPrototypeOf indirection is unnecessary. Can you remove the 3 instances in this test?

ThomasBurleson:
@SamDLT - FYI, we will not merge until the review feedback is addressed and the PR refreshed with updates.

src/components/navBar/navBar.spec.js
SamDLT: It seems to actually be necessary for some reason. Just ran a few tests:
 `spyOn(ctrl, '_upateInkBarStyles');` results in the test failing with this error:  <toHaveBeenCalled> : Expected a spy, but got Function.
 `spyOn(ctrl.prototype, '_upateInkBarStyles');` also results in the test failing with this error:  <spyOn> : could not find an object to spy upon for _updateInkBarStyles()

The expect can be modified to utilize the ctrl object instead and the test passes as expected.

Any ideas?

gmoothart: Thanks for trying; this is ok.

gmoothart:
@SamDLT can you resolve the merge conflicts and then we can merge this?

SamDLT:
@gmoothart merged and resolved conflict

jelbourn:
@SamDLT can you rebase?

SamDLT:
@jelbourn rebased

PR #9080, opened 2016-07-19
fix(datepicker): solve refocusing problems

### LINKED ISSUE #8960: angular material (datepicker) focus is on body after any keyboard based operation
**Actual Behavior**:

Its seems that ever keyboard based operation with the date-picker don't return the focus back to the date picker input or button.
As a result the tab indexing is back to the first element. 

you can take example with the official demo:
looking at the sources and debugging - its seems that focus operation is triggered on the relevant element but even so - the focus changed to to the document body.

**CodePen**
https://material.angularjs.org/latest/demo/datepicker

**Angular Versions**:  - latest 

**Additional Information**:
 -  `Browser Type: *` chrome

crisbeto:
* Fixes the datepicker not being able to refocus on the calendar icon, because the SVG element inside didn't propagate the focus event.
* Fixes the datepicker sometimes not refocusing on the element that triggered it.

Fixes #8960.

### DIFF
--- a/src/components/datepicker/js/datepickerDirective.js
+++ b/src/components/datepicker/js/datepickerDirective.js
@@ -677,28 +677,29 @@
     if (this.isCalendarOpen) {
       var self = this;
 
-      self.calendarPaneOpenedFrom.focus();
-      self.calendarPaneOpenedFrom = null;
-
-      if (self.openOnFocus) {
-        // Ensures that all focus events have fired before detaching
-        // the calendar. Prevents the calendar from reopening immediately
-        // in IE when md-open-on-focus is set. Also it needs to trigger
-        // a digest, in order to prevent issues where the calendar wasn't
-        // showing up on the next open.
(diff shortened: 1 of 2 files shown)


src/components/datepicker/js/datepickerDirective.js
```diff
@@ -677,28 +677,29 @@
     if (this.isCalendarOpen) {
       var self = this;
 
+      self.detachCalendarPane();
+      self.ngModelCtrl.$setTouched();
+
+      self.documentElement.off('click touchstart', self.bodyClickHandler);
```
ThomasBurleson: Why was this moved outside the detach/reset ?

crisbeto: Mainly because they don't _need_ to be in it. The only one that should fire on the next tick is the `calendarOpen` since it prevents the calendar reopening immediately if the user turned on `openOnFocus`.

PR #9089, opened 2016-07-20
perf(virtual-repeat): fix performance issue in IE

### LINKED ISSUE #9081: mdAutocomplete: Large data set cripples IE
**Actual Behavior**:
 -  `What is the issue? *` A simple autocomplete backed by a query which returns a large array (500+ items) displays the results instantaneously in Chrome and Firefox but takes 30 seconds in IE11.  During this time, the browser is unresponsive.  Our actual scenario has more complex objects in the query result and uses an item-template.  It takes a few seconds in Chrome and about 90 seconds in IE.
 -  `What is the expected behavior?` IE should not take 50x longer to display results.

 -  `CodePen Demo which shows your issue:` http://codepen.io/bpb/pen/LkQaQy
 -  `Details:` This is based on the autocomplete inside a dialog example.  Our use case is actually inside a dialog as well, but I'm not sure what effect the dialog has on this if any.

**Angular Versions**: *
 -  `Angular Version:` 1.5.5
 -  `Angular Material Version:` 1.0.8, 1.1.0-rc.5

**Additional Information**:
 -  `Browser Type: *` Desktop
 -  `Browser Version: *` IE11
 -  `OS: *` Windows 10
 -  `Stack Traces:`

crisbeto:
Removes a requestAnimationFrame call, that for some reason, was taking a really long time. It seems counter-intuitive, but it appears to be consistently slower across all browsers.

For reference, here's some average timings. The way I tested it was to measure the time spent inside the `readItemSize_` method for 10 times in a row, in each browser. Both with and without the rAF.

| Browser       | rAF             | No rAF          |
| ------------- | ----------------| ----------------|
| Chrome        | 119.22ms        | 8.22ms          |
| Firefox       | 92.78ms         | 34.44ms         |
| IE            | 13894.44ms      | 28.89ms         |
| Edge          | 1877.56ms       | 19.33ms         |
| **Average**   | **3996.00ms**   | **22.72ms**     |

Fixes #9081.

### DIFF
--- a/src/components/virtualRepeat/virtual-repeater.js
+++ b/src/components/virtualRepeat/virtual-repeater.js
@@ -620,7 +620,7 @@ VirtualRepeatController.prototype.containerUpdated = function() {
         this.repeatListExpression,
         angular.bind(this, function(items) {
           if (items && items.length) {
-            this.$$rAF(angular.bind(this, this.readItemSize_));
+            this.readItemSize_();
           }
         }));
     if (!this.$rootScope.$$phase) this.$scope.$digest();


crisbeto:
@topherfangio @EladBezalel @DevVersion What do you think about this? IMO it's kind of counter-intuitive, but it seems to help.

EladBezalel:
When removing rAF it seems to provide better records than obviously, IMO, we should remove the `rAF` call :+1:

src/components/virtualRepeat/virtual-repeater.js
```diff
@@ -620,7 +620,7 @@ VirtualRepeatController.prototype.containerUpdated = function() {
         this.repeatListExpression,
         angular.bind(this, function(items) {
           if (items && items.length) {
-            this.$$rAF(angular.bind(this, this.readItemSize_));
+            this.readItemSize_();
```
topherfangio: I would love to know why the removal of `$$rAF` fixes this. My first guess is that `this.itemSize` is undefined/null when using `$$rAF`, but defined when not using it.

@crisbeto Do you think you could take a quick dive to figure out what code is/isn't running with this change that might account for the drastic performance increase?

That said, even if we can't figure it out, I still think we should merge this 😄

crisbeto: Actually with rAF, it reads the `itemSize` correctly the first time, whereas without it, the first time is 0 and the next time comes back proper, presumably because the browser isn't done laying the element out.

That being said, I really have no idea why this helps. Here's a quote from MDN:

> The window.requestAnimationFrame() method tells the browser that you wish to perform an animation and requests that the browser call a specified function to update an animation **before the next repaint**.

My best guess is that sometimes that next repaint might take a long time, which is why the browser delays the execution until it does, causing it to hang.

clshortfuse: I think there's a misunderstanding on what `requestAnimationFrame` does. You use it to register for a callback before the browser repaints the window. This is more relating to syncing graphical output with the renderer. You would use this to ensure you don't redraw the same frame before the browser has displayed it. Or, you can use it to make sure to make sure you don't animate a frame too fast.

The interval at which this is fired is completely browser dependent. One browser might do this at 60fps. Another might do 30fps. There are cases when the browser won't even render a frame, for example, when a tab/window is minimized.

I would look at the initial reasoning as to why rAF was implemented in the first place. It sounds like it was to "wait one frame", but that can be done with mdUtil.nextTick(). That's actually what $$rAF does if the browser doesn't support `requestAnimationFrame`. It'll create an angular timeout running at 16.66ms (1/60th) of a frame.

clshortfuse: Slapping on another comment since we discussed this. 

Basically, $$rAF will return when the browser draws. IE does a smart drawing process where it won't redraw the page unless something changes. Since the DOM element in this case is already drawn, we can pull its dimensions immediately. If we ask to wait for another draw, we don't know when that will fire. Though this is an IE related issue, I can foresee Chrome and Firefox adopting a "redrawOnlyOnChange" features as well, especially since Microsoft is pushing Edge's battery life benefits over Chrome.

The other $$rAF on the container is correct because we are dropping the element into the DOM and we want to wait until it's rendered before it's we can get its size.

Suvarnni:
Hi,
I'm using angular-material.min.js version 1.1.12 [https://ajax.googleapis.com/ajax/libs/angular_material/1.1.12/angular-material.min.js]. 

But still i'm facing the same issue in IE11. I have used more than 8 md-autocomplete controls in a same page.

**The HTML for md-autocomplete is**:

<md-autocomplete md-no-cache="true" md-input-name="autocomplete" md-selected-item="studentName" md-search-text="searchStudent" md-items="item in query(searchStudent , studentMasterList, 'studentName')" md-item-text="item.studentName" md-min-length="0" ng-disabled="details.disableItem.disableFields" md-clear-button="!details.disableItem.disableFields" required> <md-item-template> <span md-highlight-text="searchStudent">{{item.studentName}}</span></md-item-template> <md-not-found> No matches found. </md-not-found> </md-autocomplete>

**The JS for calling Query is:**

$scope.query = function(searchText, masterData, searchKey) {
		var filterObj = {};
      	filterObj[searchKey] = searchText;      	
      	var item = searchText ? $filter('filter')(masterData,filterObj) : masterData;
		return item;		
	};

### **Please help me to resolve this issue. Its taking nearly 3 to 5 sec to load the list.**

Splaktar:
@Suvarnni please open a new issue with a CodePen reproduction and we can take a look.

PR #9095, opened 2016-07-21
feat(chips): md-add-on-blur functionality

### LINKED ISSUE #3364: md-chips: Add a chip onInputBlur
I think it's a good experience for md-chips to convert outstanding text to a chip when input loses focus. 

Here is a use-case when it's relevant. User does not know that he needs to click enter to add a chip. When users clicks "Send email", no email has been captured.

DevVersion:
* Allows developers to convert the remaining input text into a new chip on input blur.
  This is useful for example, when having chips for email addresses.

Closes #3364.

### DIFF
--- a/src/components/chips/js/chipsDirective.js
+++ b/src/components/chips/js/chipsDirective.js
@@ -98,6 +98,8 @@
    * @param {number=} md-max-chips The maximum number of chips allowed to add through user input.
    *    <br/><br/>The validation property `md-max-chips` can be used when the max chips
    *    amount is reached.
+   * @param {boolean=} md-add-on-blur When set to true, remaining text inside of the input will
+   *    be converted into a new chip on blur.
    * @param {expression} md-transform-chip An expression of form `myFunction($chip)` that when called
    *    expects one of the following return values:
    *    - an object representing the `$chip` input string
@@ -224,7 +226,6 @@
         readonly: '=readonly',
         removable: '=mdRemovable',
         placeholder: '@',
-        mdEnableChipEdit: '@',
         secondaryPlaceholder: '@',
         maxChips: '@mdMaxChips',
         transformChip: '&mdTransformChip',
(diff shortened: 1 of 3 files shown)


clshortfuse:
LGTM

hschriek:
md-add-on-blur is only fired, when tabbing out of the chip input field, but not when leaving the field by clicking to another field

DevVersion:
@hschriek Not true. We are just listening for a native `blur` event of the associated input element.

> 
image

FYI: My mouse normally doesn't look like that :smile:

clshortfuse:
@DevVersion I think he means when you type dd, and click on the chip to the left, it doesn't blur the entire component, so the event doesn't get passed. I haven't tested the PR, but I think that's what he means.

DevVersion:
@clshortfuse This isn't a problem. The chip will be added as expected.

ThomasBurleson:
@EladBezalel - please review.

topherfangio:
@DevVersion I think we need to enable this for the contact chips as well? Seems like that will be the most useful case.

Also (I can't easily test with the contact chips), does/should it automatically add if the input is invalid?

For instance, if you use `md-add-on-blur="true"` with email chips, and the user enters `something@somewhere` but leaves off the `.com`, should we still add it?

DevVersion:
@topherfangio As we discussed on Slack, we are now checking the validity of the `chips` ngModel and also of the `userNgModel` (if provided).

That was a great suggestion. Thanks again.

src/components/chips/js/chipsDirective.js
```diff
@@ -224,7 +226,6 @@
         readonly: '=readonly',
         removable: '=mdRemovable',
         placeholder: '@',
-        mdEnableChipEdit: '@',
```
topherfangio: Curious; why was this line removed? Looks like we're still using it above?

DevVersion: Good point - it's because the scope binding is never used and we are only checking the attribute once by using `$attrs`.
> https://github.com/DevVersion/material/blob/3f20904c387e4c7154d67a73d5a8f94213a3607e/src/components/chips/js/chipsController.js#L57

ThomasBurleson:
@jelbourn - gatekeeper ping.

PR #9099, opened 2016-07-21
chore(package.json): add main field to package.json

sarahpads:
Include entry point of module in package.json.
Necessary if angular material was installed from git branch into a project using webpack.
Without this field, webpack is unable to resolve the module.

### DIFF
--- a/package.json
+++ b/package.json
@@ -17,6 +17,7 @@
     "url": "git://github.com/angular/material.git"
   },
   "private": true,
+  "main": "./dist/angular-material.js",
   "style": "./dist/angular-material.css",
   "scss": "./dist/angular-material.scss",
   "devDependencies": {


sarahpads:
And signed.

package.json
```diff
@@ -17,6 +17,7 @@
     "url": "git://github.com/angular/material.git"
   },
   "private": true,
+  "main": "./dist/angular-material.js",
```
DevVersion: Why should we have a main file here? ngMaterial doesn't run inside of NodeJS.

Also a "main" with an associated index file, is already provided in `bower-material`.
> https://github.com/angular/bower-material/blob/master/package.json

sarahpads: I recently had to install angular-material from a specific commit for a patch (https://github.com/angular/material/commit/bd1cce42ba16b733af0163011f8bb2429782b5da) that is not included in the latest npm package.  My project uses webpack to bundle all dependencies together and, without the main field, it was unable to resolve the module.

I was unaware of bower-material.  It looks like switching to that module will also solve the webpack issue.

PR #9102, opened 2016-07-21
perf(virtual-repeat): avoid unnecessary DOM lookups

crisbeto:
The virtual repeater was looking up a collection of elements, but only using the first one. This change switches it to only look for the first element via `querySelector`.


clshortfuse:
Smarter short-circuit evaluation

:lgtm:

PR #9113, opened 2016-07-22
feat(datepicker): allow users to specify the default calendar view

### LINKED ISSUE #9111: md-datepicker option to open at the year/month view
**Actual Behavior**:
 -  `What is the issue? Using the md-datepicker as a date of birth (dob) field we're finding users don't realize that the year can be clicked to go to the year/month view.  This issue is paralleled by the process of choosing a date of birth being longer then it needs to be for a user.  Right now a dob field requires clicking the year, scrolling quickly backwards to the year of birth, choosing the month, then the day of the month.`
 -  `What is the requested behavior?  The request feature is to be able to set the calendar to open to the year/month view, then the user would very intuitively know to scroll backwards a bit to the year and choose the month, then the day of month.  This would be better than emphasizing the year on the month/day view for users since it speeds up the process of filling out a form.`
 - `Example of clicking the md-date-picker and it opened at year/month view`

feature-request-calendar-starting-point-in-year-month-view

 -  `CodePen Demo which shows your issue: Not applicable since it's a feature request`
 -  `Details: Not applicable since it's a feature request` 

**Angular Versions**: 
 -  `Angular Version: 1.5.7` 
 -  `Angular Material Version: 1.1.0-RC5` 

**Additional Information**:
 -  `Browser Type: Chrome, but is irrelevant to feature request`
 -  `Browser Version: Latest, but is irrelevant to feature request`
 -  `OS: Windows, but is irrelevant to feature request`
 -  `Stack Traces: none, this is a feature request`

----

  Add more lines if needed.

crisbeto:
* Allows users to set the default open view of the calendar via the `md-current-view` attribute.
* Removes a few unused dependencies from the datepicker.

Fixes #9111.

### DIFF
--- a/src/components/datepicker/js/calendar.js
+++ b/src/components/datepicker/js/calendar.js
@@ -57,7 +57,8 @@
       scope: {
         minDate: '=mdMinDate',
         maxDate: '=mdMaxDate',
-        dateFilter: '=mdDateFilter'
+        dateFilter: '=mdDateFilter',
+        _currentView: '@mdCurrentView'
       },
       require: ['ngModel', 'mdCalendar'],
       controller: CalendarCtrl,
@@ -116,8 +117,13 @@
     /** @type {!angular.NgModelController} */
     this.ngModelCtrl = null;
 
-    /** @type {String} The currently visible calendar view. */
-    this.currentView = 'month';
+    /**
+     * The currently visible calendar view. Note the prefix on the scope value,
+     * which is necessary, because the datepicker seems to reset the real one value if the
+     * calendar is open, but the value on the datepicker's scope is empty.
+     * @type {String}
+     */
+    this.currentView = this._currentView || 'month';
 
     /** @type {String} Class applied to the selected date cell. */
     this.SELECTED_DATE_CLASS = 'md-calendar-selected-date';
--- a/src/components/datepicker/js/calendar.spec.js
+++ b/src/components/datepicker/js/calendar.spec.js
@@ -93,9 +93,9 @@ describe('md-calendar', function() {
   }
 
   /** Creates and compiles an md-calendar element. */
-  function createElement(parentScope) {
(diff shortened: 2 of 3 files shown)


src/components/datepicker/js/calendar.js
```diff
@@ -57,7 +57,8 @@
       scope: {
         minDate: '=mdMinDate',
         maxDate: '=mdMaxDate',
-        dateFilter: '=mdDateFilter'
+        dateFilter: '=mdDateFilter',
+        _currentView: '@mdCurrentView'
```
DevVersion: Can you not make this optional and do in the controller

```js
this.currentView = this.currentView || 'month';
```

This would drop the unnecessary variable

crisbeto: I tried it out, but for some reason the datepicker kept resetting it after the calendar was opened. E.g. here's what I tried:

```js
var self = this;

console.log(self.currentView); // logs properly

setTimeout(function() {
  console.log(self.currentView); // logs an empty string 
});
```

It's weird, because it doesn't happen if I switch it to a two-way binding or I also add `this.currentView = this.currentView || 'month';` to the datepicker's controller. I went with the underscored property in order to keep all the `currentView` logic within the calendar.

DevVersion: Hmm, I can't really understand why this is resetted. Isn't the `currentView` method only called after some interactions?

crisbeto: I can't understand it either, the only time something interacts with it is when passing it through the template.  My initial guess was because the calendar pane gets transferred from being a child of the datepicker to being directly in the body, but it still happened even after disabling it.

DevVersion: Pretty weird. Not sure if you want to invest some more time, but you could possibly try to hook into that variable and try to figure out where it was changed from.

PR #9115, opened 2016-07-22
feat(layouts): add @mixin for responsive support for rows

### LINKED ISSUE #9112: md-divider: Responsive layout support is not properly supported
**Actual Behavior**:
-  `What is the issue? *` Divider will display as a column if the any responsive configuration of layout is set to `row`.
-  `What is the expected behavior?`Divider should only display as a column if the current layout is in a row configuration.

-  `CodePen Demo which shows your issue:` http://codepen.io/shortfuse/pen/rLdEyR
-  `Details:` The problem is `divider.scss` is based around the `layout-row-*` only existing if it's displayed. The problem is the classes don't disappear when layout changes, locking divider into act as if it's always in a row configuration

**Angular Versions**: *
-  `Angular Version:` 1.1.0-rc5 (master)

**Additional Information**:
-  `Browser Type: *` Chrome
-  `Browser Version: *` v52
-  `OS: *`ChromeOS

---

clshortfuse:
Create @mixin that allows different layout breakpoints

Fixes #9112

### DIFF
--- a/src/core/style/mixins.scss
+++ b/src/core/style/mixins.scss
@@ -326,3 +326,47 @@
     }
   }
 }
+
+// Only use when in row layout
+@mixin when-layout-row($element, $parent: '') {
+  @media (max-width: 599px) {
+    #{$parent}.layout-row:not(.layout-xs-column),
+    #{$parent}.layout-xs-row {
+      & > #{$element} { @content; }
+    }
+  }
+  @media (min-width: 600px) and (max-width: 959px) {
+    #{$parent}.layout-row:not(.layout-gt-xs-column),
+    #{$parent}.layout-gt-xs-row,
+    #{$parent}.layout-sm-row {
+      &:not(.layout-sm-column) > #{$element} { @content; }
+    }
+  }
+  @media (min-width: 960px) and (max-width: 1279px) {
+    #{$parent}.layout-row:not(.layout-gt-xs-column):not(.layout-gt-sm-column),
+    #{$parent}.layout-gt-xs-row:not(.layout-gt-sm-column),
+    #{$parent}.layout-gt-sm-row,
+    #{$parent}.layout-md-row {
+      &:not(.layout-md-column) > #{$element} { @content; }
+    }
+  }
+  @media (min-width: 1280px) and (max-width: 1919px) {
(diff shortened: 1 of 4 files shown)


clshortfuse:
http://codepen.io/shortfuse/pen/AXyZLX (localhost test)

src/core/style/mixins.scss
```diff
@@ -326,3 +326,47 @@
     }
   }
 }
+
+// Only use when in row layout
+@mixin when-layout-row($element) {
```
crisbeto: This seems like a whole lot of selectors (close 2kb), for something as simple as drawing a divider.

clshortfuse: Right, I do have some PRs that would use this to fix issues with md-select, md-checkbox, md-input-container and md-datepicker. I had a tough time trying to squish it down though.

clshortfuse:
I realized I should expand this to `divider-theme.scss` and `radio-button.scss` to make better use of the CSS

clshortfuse:
Yeah, this is looking too bloated. Bumped radio-button from 1KB to 10KB.

src/core/style/mixins.scss
```diff
@@ -326,3 +326,47 @@
     }
   }
 }
+
+// Only use when in row layout
+@mixin when-layout-row($element, $parent: '') {
+  @media (max-width: 599px) {
```
EladBezalel: we have variables for those breakpoints

src/core/style/mixins.scss
```diff
@@ -326,3 +326,47 @@
     }
   }
 }
+
+// Only use when in row layout
+@mixin when-layout-row($element, $parent: '') {
```
EladBezalel: why not use the scss hierarchy instead of specify a parent?

ThomasBurleson:
Mistakenly merged... just now reverted from `master`.

ThomasBurleson:
Rejected as invalid PR

src/core/style/mixins.scss
clshortfuse: I've tried that, but the problem is that @media needs to be root level.

If I do ```
md-radio-group {
  @include when-layout-row(md-radio-button) {
    ...
  }
}```

Then the @media queries would appear under `md-radio-group`, I wasn't sure of a better solution.

PR #9119, opened 2016-07-23
update(autocomplete): md-require-match only turn invalid if search text is provided

### LINKED ISSUE #9072: Md-required-match doesn't work as planned
@DevVersion 

**Actual Behavior**:
 -  `What is the issue? *`
When md-required-match is set to true, and required is not defined or false, in the beginning everything works like planned. When input is empty, there will be no error, when i try to type something, error will be displayed but will be removed when i pick something. The problem starts when i decide to delete everything in the input, then the error still stays displayed, even when input is "".

When md-required-match is true, and md-required is undefined or false, error should be displayed only when input is not empty and item is not selected.

 -  Unable to link to it because of docs issue #9055

DevVersion:
* Currently the `md-require-match` attribute checks whether an item is selected or not. The validity shouldn't be set to false, if the searchText is empty.

Closes #9072.

**Breaking Change**: The autocomplete validator `md-require-match` no longer matches if the search text is empty

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -1326,7 +1326,10 @@ describe('<md-autocomplete>', function() {
       var element = compile(template, scope);
       var ctrl = element.find('md-autocomplete').controller('mdAutocomplete');
 
-      element.scope().searchText = 'fo';
+      // Flush the element gathering.
+      $timeout.flush();
+
+      scope.$apply('searchText = "fo"');
       $timeout.flush();
 
       ctrl.select(0);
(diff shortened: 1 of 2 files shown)


gpopovic:
Any ETA when this gets accepted?

DevVersion:
Not sure @gpopovic - It's a pretty simple update.

EladBezalel:
LGTM :+1:

topherfangio:
LGTM :+1:

topherfangio:
@ThomasBurleson Ready for your review.

ekarudianto:
Please review and release and update soon :)

PR #9130, opened 2016-07-25
fix(checkbox): avoid potential memory leaks, general cleanup

### LINKED ISSUE #7993: mdCheckbox: Memory leak
The event handler [in checkbox.js](https://github.com/angular/material/blob/v1.1.0-rc2/src/components/checkbox/checkbox.js#L85) is leaking.
```
// Attach a click handler in compile in order to immediately stop propagation
// (especially for ng-click) when the checkbox is disabled.
tElement.on('click', function(event) {
  if (this.hasAttribute('disabled')) {
    event.stopImmediatePropagation();
  }
});
```

crisbeto:
* Avoids potential memory leaks in the checkbox component by removing any event handlers from the `compile` function and avoiding references between `compile` and `link`.
* Cleans out a few redundancies in the checkbox link function.

Fixes #7993.

### DIFF
--- a/src/components/checkbox/checkbox.js
+++ b/src/components/checkbox/checkbox.js
@@ -31,10 +31,10 @@ angular
  * @param {expression=} md-indeterminate This determines when the checkbox should be rendered as 'indeterminate'.
  *     If a truthy expression or no value is passed in the checkbox renders in the md-indeterminate state.
  *     If falsy expression is passed in it just looks like a normal unchecked checkbox.
- *     The indeterminate, checked, and unchecked states are mutually exclusive. A box cannot be in any two states at the same time. 
- *     Adding the 'md-indeterminate' attribute overrides any checked/unchecked rendering logic. 
+ *     The indeterminate, checked, and unchecked states are mutually exclusive. A box cannot be in any two states at the same time.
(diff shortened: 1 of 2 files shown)


src/components/checkbox/checkbox.js
```diff
@@ -76,32 +75,35 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
-    // Redirect focus events to the root element, because IE11 is always focusing the container element instead
-    // of the md-checkbox element. This causes issues when using ngModelOptions: `updateOnBlur`
-    container.on('focus', function() {
-      tElement.focus();
-    });
+    return  {
+      pre: function(scope, element) {
+        // Attach a click handler during preLink, in order to immediately stop propagation
+        // (especially for ng-click) when the checkbox is disabled.
+        element.on('click', function(e) {
```
ThomasBurleson: You could also attach click listener during the postlink using the `capture` phase.

crisbeto: I think the point of this one is to prevent any events from firing on the same element (e.g. we have a unit test for an `ng-click`), but from what I understand, `capture` is used to stop the "propagation" down towards child elements, which shouldn't really be an issue. I might be wrong here since I haven't had to use capturing in practise.

PR #9144, opened 2016-07-26
fix(input): ignore hidden inputs when determinating icon orientation

### LINKED ISSUE #8899: md-select with name-attribute and right icon renders md-icon-left because of hidden select (2nd)
I already posted this as issue 8234 but i was closed because it had no demo. So here again, this thime with demo:
```
 <md-input-container>
    <label>Label</label>
    <md-select ng-model="size" name="Test">
      <md-option ng-repeat="size in sizes" value="{{size}}">{{size}}</md-option>
    </md-select>
    <md-icon>help</md-icon>
  </md-input-container>
```
'Name' is required for ng-form support. md-icon is displayed right but container also has 'md-icon-left' class because mdInputContainerDirective looks for 'md-icon ~ isel' and isel is one of 'INPUT', 'TEXTAREA', 'SELECT', 'MD-SELECT', so it detects the additional hidden select which is created because of the name-attribute.
Current Version 1.0.5
Demo: [https://embed.plnkr.co/s3NscQ4s2k1xzrT5pLfq/](https://embed.plnkr.co/s3NscQ4s2k1xzrT5pLfq/)

clshortfuse:
The `md-icon-left` check gives a false position when using `md-select`
with an `md-icon` element at the right. This occurs because `md-select`
creates a hidden `select` element, resulting in the logic incorrectly
creating both `md-icon-left` and `md-icon-right` classes.
- Add check for hidden inputs via CSS selectors

Fixes #8899

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -80,11 +80,20 @@ function mdInputContainerDirective($mdTheming, $parse) {
 
   var INPUT_TAGS = ['INPUT', 'TEXTAREA', 'SELECT', 'MD-SELECT'];
 
+  var HIDDEN_INPUT_SELECTORS = ':not([hidden=""])'
+        + ':not([hidden="true"])'
+        + ':not([type="hidden"])'
+        + ':not([aria-hidden=""])'
+        + ':not([aria-hidden="true"])'
+        + ':not(.md-visually-hidden)';
+
   var LEFT_SELECTORS = INPUT_TAGS.reduce(function(selectors, isel) {
+    isel = isel + HIDDEN_INPUT_SELECTORS;
     return selectors.concat(['md-icon ~ ' + isel, '.md-icon ~ ' + isel]);
   }, []).join(",");
 
   var RIGHT_SELECTORS = INPUT_TAGS.reduce(function(selectors, isel) {
+    isel = isel + HIDDEN_INPUT_SELECTORS;
     return selectors.concat([isel + ' ~ md-icon', isel + ' ~ .md-icon']);
   }, []).join(",");
 
--- a/src/components/input/input.spec.js
+++ b/src/components/input/input.spec.js
@@ -906,5 +906,44 @@ describe('md-input-container directive', function() {
         expect(el.hasClass('md-icon-right')).toBeTruthy();
       });
     });
+
+    [
+      'type="hidden"',
+      'aria-hidden="true"',
+      'aria-hidden',
+      'hidden',
+      'hidden="true"',
+      'class="md-visually-hidden"',
+    ].forEach(function(attribute) {
+      it('ignores all hidden inputs when adding icon classes', function() {
+        var el = compile(
+          '<md-input-container>' +
+          '  <input type="text" name="text">' +
+          '  <md-icon>help</md-icon>' +
+          '  <select ' + attribute + ' >' +
+          '</md-input-container>'
+        );
+
+        expect(el.hasClass('md-icon-left')).toBe(false);
+        expect(el.hasClass('md-icon-right')).toBe(true);
+      });
+    });
+
+    [
+      'aria-hidden="false"',
+      'hidden="false"',
+    ].forEach(function(attribute) {
+      it('does not ignore inputs with falsy hidden attributes when adding icon classes', function() {
+        var el = compile(
+          '<md-input-container>' +
+          '  <input ' + attribute + ' >' +
+          '  <md-icon>help</md-icon>' +
+          '</md-input-container>'
+        );
+
+        expect(el.hasClass('md-icon-left')).toBe(false);
+        expect(el.hasClass('md-icon-right')).toBe(true);
+      });
+    });
   });
 });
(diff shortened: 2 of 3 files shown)


topherfangio:
I can't get the demo in the associated issue to work, but it seems to me that this needs to check for `type="hidden"` as well?

DevVersion:
@clshortfuse - I also think it would be better to check for the visibility by using a check like the followed.
```js
isHidden: function(element, computedStyle, offsetParent) {
   return computedStyle && window.getComputedStyle(element[0]).display === 'none' ||
          offsetParent && element[0].offsetParent == null;
}
```

Just checking for the `hidden` attribute isn't really flexible in my opinion. Thoughts @topherfangio?

clshortfuse:
here's the demo: http://codepen.io/shortfuse/pen/KrZzVr

This is run during code generation, and AFAIK, only we are manipulating this by creating our own hidden element as seen here: https://github.com/angular/material/blob/master/src/components/select/select.js#L220.

I don't think users are creating hidden inputs. We could change the code to `:not(.md-visually-hidden)`, but I figured this is a bit more universal.

topherfangio:
I think it would be ideal if it worked for the following case as well: http://codepen.io/topherfangio/pen/OXENdN?editors=1000

That said, if it's not easy, then I don't think it's worth supporting. But it sounds like you just would have to add `type="hidden"` to your selector as well.

I'm going back and forth on the `md-visually-hidden` class because I can potentially foresee some people using `aria-hidden="TRUE"` or some variation in which case this would fail. Thoughts?

src/components/input/input.js
```diff
@@ -76,11 +76,17 @@ function mdInputContainerDirective($mdTheming, $parse) {
 
   var INPUT_TAGS = ['INPUT', 'TEXTAREA', 'SELECT', 'MD-SELECT'];
 
+  var HIDDEN_INPUT_SELECTORS = ['[hidden=""]', '[hidden="true"]',
+                                '[aria-hidden=""]', '[aria-hidden="true"]',
+                                '[type="hidden"]',
+                                '.md-visually-hidden'];
```
DevVersion: That indention style looks pretty weird here. Can you probably update and follow the patterns like in the `list` component for example?

src/components/input/input.js
```diff
@@ -76,11 +76,18 @@ function mdInputContainerDirective($mdTheming, $parse) {
 
   var INPUT_TAGS = ['INPUT', 'TEXTAREA', 'SELECT', 'MD-SELECT'];
 
+  var HIDDEN_INPUT_SELECTORS = [
+    '[hidden=""]', '[hidden="true"]', '[type="hidden"]',
+    '[aria-hidden=""]', '[aria-hidden="true"]', '.md-visually-hidden'
+  ];
+
   var LEFT_SELECTORS = INPUT_TAGS.reduce(function(selectors, isel) {
+    isel = isel + ':not(' + HIDDEN_INPUT_SELECTORS.join('):not(') + ')';
```
topherfangio: Since you don't need these as an array, I would go ahead and do the `.join('):not(')` when you define the variable so you don't have to do it twice. I think it will also be a tad bit clearer...too many parentheses here :smile:

topherfangio:
Aside from my one comment, this looks good to me :+1:

topherfangio:
@ThomasBurleson Ping for your final review.

ThomasBurleson:
The tests look inadequate. Where is a working CodePen with this demonstrating the fix?

clshortfuse:
Here's a codepen showing the issue. When using `md-icon` after `md-select` causes the `md-icon-left` class to be incorrectly added, creating an empty whitespace:

http://codepen.io/shortfuse/pen/KgPYvo

topherfangio:
@EladBezalel Please review as well.

src/components/input/input.js
```diff
@@ -80,11 +80,20 @@ function mdInputContainerDirective($mdTheming, $parse) {
 
   var INPUT_TAGS = ['INPUT', 'TEXTAREA', 'SELECT', 'MD-SELECT'];
 
+  var HIDDEN_INPUT_SELECTORS = ':not([hidden=""])'
```
EladBezalel: I'm not a fan of all those selectors.. can't we use siblings or something else instead of this list?

clshortfuse: I'm not sure I understand. We are checking for the siblings on lines 92 and 97. The problem is, those checks need to avoid certain attributes. Patching into querySelector is faster than returning all of them and checking the attributes for each one.

That said, the entire concept of `md-icon-left` and `md-icon-right` is only to reduce the CSS and has no value on the Javascript side. Even if we move it to CSS, it will still need this list of selectors. 😑

topherfangio:
@EladBezalel Not really; I tried a bunch of different ways of doing it before I settled on this approach which seemed to work and be relatively performant.

Unless someone wants to spend a bit more time trying to find a better solution, I think this is mergable.

clshortfuse:
@ThomasBurleson going to need your executive decision here.

ThomasBurleson:
Closing as stale and will not merge.

PR #9156, opened 2016-07-27
Zoom directive

cesar-garay:
Hi All, this is my first PR and I need som help.
I created a zoom directive for two images but I need a hand with the unit test.
Where can i found information to run gulp test task?
Thanks in advanced.


cesar-garay:
I signed it!

PR #9160, opened 2016-07-28
feat(mixin) Improved `input-placeholder-color` mixin

IPRIT:
Allows to set style for input state selectors (e.g. :hover, :active, :focus)

Examples:
  
```scss
@include input-placeholder-color(#faf);
```

```scss
@include input-placeholder-color(#faf, (hover: #ff0));
```

```scss
@include input-placeholder-color(#faf, (hover: #ff0, active: #0ff)) {
  transition: all .2s ease-in-out;
}
```

### DIFF
--- a/src/components/input/input-theme.scss
+++ b/src/components/input/input-theme.scss
@@ -7,7 +7,6 @@ md-input-container.md-THEME_NAME-theme {
     }
     color: '{{foreground-1}}';
     border-color: '{{foreground-4}}';
-    text-shadow: '{{foreground-shadow}}';
   }
 
   > md-icon {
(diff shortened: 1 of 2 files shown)


DevVersion:
Where will be that functionality used? Seems like it's just improving the mixin without any required usage?

IPRIT:
@DevVersion, I can offer this effect for focused input. But I need help to finish this.

clshortfuse:
I'm not sure I see the reason for this. Also, `md-input-focused` is where the label gets its color on focus. Users can already override that with custom CSS.

But it seems to want to color the placeholder/hint text on focus. I'm not sure that is part of Material Design spec. Generally, we ask users to file an issue first before proposing changes so we can investigate their validity and best course of action.

DevVersion:
Correct. I kept it open to get the reason for this change, but since you said this requires more work / help and it's kind of an unused change, you should file an issue first.

Thanks for the contribution through.

PR #9168, opened 2016-07-28
fix(select): cast number model values to string hash keys.

### LINKED ISSUE #9124: md-select directive default value can't make the required be valid
**Actual Behavior**:
 -  `What is the issue? *`When set default value to the select directive which is required,the form still invalid.
 -  `What is the expected behavior?`The form should be valid.

 -  `CodePen Demo which shows your issue:`http://codepen.io/wcmboy/pen/BzxJQv
 -  `Details:`In the six line of the AppCtrl controller,I set $scope.myModel default value was 2.But the myForm.$invalid still was ture.

**Angular Versions**: *
 -  `Angular Version:`1.5.5
 -  `Angular Material Version:`v1.1.0-rc.5

**Additional Information**:
 -  `Browser Type: *`Google Chrome
 -  `Browser Version: *`49.0.2623.110 m
 -  `OS: *`Windows 10
 -  `Stack Traces:`

----

  Add more lines if needed.

DevVersion:
* Currently when using the `value` attribute, which is designed to only accept string values. In some cases a number is also a valid option.
  When the developer sets the model to the specific number (number type), to select the given option (number in string), the select will turn invalid.
  Once the types mismatch, we can easily cast it to a string, to be able to compare it properly.

R: @topherfangio 

Fixes #9124.

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -823,8 +823,19 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
         // the current option, which will be added, then we can be sure, that the validation
         // of the option has occurred before the option was added properly.
         // This means, that we have to manually trigger a new validation of the current option.
-        if (angular.isDefined(self.ngModel.$modelValue) && self.hashGetter(self.ngModel.$modelValue) === hashKey) {
-          self.ngModel.$validate();
+        if (angular.isDefined(self.ngModel.$modelValue)) {
+          var modelValue = self.ngModel.$modelValue;
+
+          // In some cases the $modelValue mismatches the type of the current hashKey.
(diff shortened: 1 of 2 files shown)


topherfangio:
@DevVersion This looks good, but can you add a note to the select docs about this behavior since it is a slight change from the existing functionality?

Also, I think this should probably go in 1.2.0. Thoughts?

DevVersion:
@topherfangio Sure - I will add some info about the new behavior. 

I also agree with `1.2.0` - it's not a high priority and it also seems like this could definitely break some existing code-base. 

Let's have a discussion about this change anytime.

DevVersion:
I'm going to close this PR, because I do not like what I did here. I want @topherfangio to take care of that issue, since he is the owner of the `select`.

PR #9169, opened 2016-07-28
fix(input): Add missing Polyfill for ng1.3 and fix tests.

topherfangio:
Angular 1.3 apparently did not include the `AnimateRunner.all()` method. Grab it from latest 1.5.8 code and add to our 1.3 polyfills.

Additional, our $animateCss polyfill for 1.3 modifies the `to` and `from` options that are passed in and sets them to `null`. This was causing the tests to fail. Fix by keeping a cloned reference for test purposes.

### DIFF
--- a/src/components/input/input-animations.spec.js
+++ b/src/components/input/input-animations.spec.js
@@ -1,7 +1,7 @@
 describe('md-input-container animations', function() {
   var $rootScope, $compile, $animate, $animateCss,
     el, pageScope, invalidAnimation, messagesAnimation, messageAnimation,
-    cssTransitionsDisabled = false;
+    cssTransitionsDisabled = false, lastAnimateCall;
 
   // Load our modules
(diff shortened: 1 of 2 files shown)


epelc:
@topherfangio is this optional if you are using 1.5.x?

topherfangio:
@epelc This should only be included if your Angular version is < 1.4 (so 1.3 and earlier).

You having issues in 1.5?

epelc:
@topherfangio I meant could we make this optional(along with any other polyfills) if we do not require them? Perhaps moving to some sort of `angular-material.polyfills.js` file would be good.

No issues. Just wanted to make the build size a little bit smaller for me and anyone else using a newer version of angular.

topherfangio:
@epelc Ah, gotcha. At the moment, no, we don't provide a way to easily remove them, but I'll chat with the team about it this next week and see what they say.

epelc:
@topherfangio I made an issue for you to reference #9378

PR #9173, opened 2016-07-29
feat(md-dialog): Use textarea for prompt dialog

### LINKED ISSUE #9171: dialog(prompt): use textarea to allow for multiple lines of input
The new 'prompt' option for the md-dialog is great, however when I showed it to my client they asked if they could have multiple lines of input.

This could be achieved using a textarea instead of an input, by default it would look and behave the same, but would give the option of allow for multiple lines.

gseabrook:
Use textarea instead of input for prompt dialog to allow for multiple lines of input

Closes #9171

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -428,6 +428,7 @@ function MdDialogDirective($$rAF, $mdTheming, $mdDialog) {
  * - $mdDialogPreset#targetEvent(DOMClickEvent=) - A click's event object. When passed in as an option,
  *     the location of the click will be used as the starting point for the opening animation
  *     of the the dialog.
+ * - $mdDialogPreset#useTextarea(boolean) - Uses a textarea instead of an input for multiple lines of entry
  *
  */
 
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@gseabrook - lgtm. This will be merged into master **after** next week.

DevVersion:
@gseabrook Thanks for the PR again. I just had a quick discussion regarding your PR and I was leaning towards making this `textarea` functionality opt-in.

This means, that the dialog prompt uses by default a single-line input and users can decide whether they want to use a textarea instead.

Could you please update your Pull Request?

gseabrook:
I've updated the pull request to use input by default, with the option of using textarea instead

DevVersion:
Thanks for making the changes. The reason why we want to have a normal input by default is that a normal prompt in the browser (`window.prompt()`) and also a native Android Prompt uses a EditText field (single line) by default

topherfangio:
We had a team discussion and do not believe this fits with our core philosophy.

PR #9179, opened 2016-07-29
feat(bottom-sheet): add isLockedOpen option; general cleanup

### LINKED ISSUE #9084: Add support for md-is-locked-open="<expression>"  in  BottomSheet
BottomSheet is a pretty nice feature, which we use as an edit dialogue.

There is however no way I could see to programmatically remove the escapeToClose and clickOutsideToClose behaviors. I would like to be able to lock the bottomsheet in position (either removing the 2 properties or supporting something like md-is-locked-open) to prevent the user from closing it accidentally.

This was briefly discussed in https://github.com/angular/material/issues/2895 but it does not look like it was addressed?

crisbeto:
- Adds the `isLockedOpen` option to the bottom sheet service, allowing users to disable all non-programmatic ways of closing a bottom sheet.
- Cleans up some redundant logic in the mdBottomSheet service.
- Cleans up the bottom sheet unit test setup.

Fixes #9084.

### DIFF
--- a/src/components/bottomSheet/bottom-sheet.js
+++ b/src/components/bottomSheet/bottom-sheet.js
@@ -90,6 +90,9 @@ function MdBottomSheetDirective($mdBottomSheet) {
  *   - `disableBackdrop` - `{boolean=}`: When set to true, the bottomsheet will not show a backdrop.
  *   - `escapeToClose` - `{boolean=}`: Whether the user can press escape to close the bottom sheet.
  *     Default true.
+ *   - `isLockedOpen` - `{boolean=}`: Disables all default ways of closing the bottom sheet. **Note:** this will override
+ *     the `clickOutsideToClose` and `escapeToClose` options, leaving only the `hide` and `cancel`
+ *     methods as ways of closing the bottom sheet. Defaults to false.
  *   - `resolve` - `{object=}`: Similar to locals, except it takes promises as values
  *   and the bottom sheet will not open until the promises resolve.
  *   - `controllerAs` - `{string=}`: An alias to assign the controller to on the scope.
@@ -150,7 +153,8 @@ function MdBottomSheetProvider($$interimElementProvider) {
       disableBackdrop: false,
       escapeToClose: true,
       clickOutsideToClose: true,
-      disableParentScroll: true
+      disableParentScroll: true,
+      isLockedOpen: false
     };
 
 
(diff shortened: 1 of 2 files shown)


src/components/bottomSheet/bottom-sheet.js
```diff
@@ -189,16 +197,15 @@ function MdBottomSheetProvider($$interimElementProvider) {
         $animate.enter(backdrop, options.parent, null);
       }
 
-      var bottomSheet = new BottomSheet(element, options.parent);
-      options.bottomSheet = bottomSheet;
+      options.cleanupGestures = options.isLockedOpen ? angular.noop : new BottomSheetGestures(element, options.parent);
```
DevVersion: This is pretty unclear.

```js
if (!options.isLockedOpen) {
  options.cleanupGestures = registerGestures(element, options.parent);
}

function registerGestures(element, parent) {
 ...
}
```

crisbeto: I agree about the naming, but the main reason I decided to default to a noop was to make the API consistent, no matter the options.

DevVersion: It doesn't make any sense to have a noop function here. Check for example the dialog interim element / component.

Regarding API, the options are not any API for the developers. At time of `show` those are just internal and used to store things between the different states.

crisbeto: Ah right, I confused it with the hide/cancel/show. I'll switch it up.

src/components/bottomSheet/bottom-sheet.js
```diff
@@ -189,16 +197,17 @@ function MdBottomSheetProvider($$interimElementProvider) {
         $animate.enter(backdrop, options.parent, null);
       }
 
-      var bottomSheet = new BottomSheet(element, options.parent);
-      options.bottomSheet = bottomSheet;
+      if (!options.isLockedOpen) {
+        options.cleanupGestures = registerGestures(element, options.parent);
```
DevVersion: Actually after looking again at the PR, can't you move that declaration up to the `isLockedOpen` check and use a `else`? 

```js
if (options.isLockedOpen) {
  options.clickOutsideToClose = false;
  options.escapeToClose = false;
} else {
  options.cleanupGestures = registerGestures(element, options.parent);
}
```

DevVersion:
LGTM aside from the one comment.

src/components/bottomSheet/bottom-sheet.js
crisbeto: I will, another good point :+1:

PR #9186, opened 2016-07-30
fix(datepicker): calendar opening on window focus with mdOpenOnFocus

crisbeto:
Fixes cases where a closed calendar gets reopened when the user tabs away, on a datepicker with `md-open-on-focus`.


qubiack:
Now user can't open a datepicker again when back to the tab where focus flag `inputFocusedOnWindowBlur` is set as true - it never is set to false.

PR #9191, opened 2016-07-30
fix(list): do not wrap proxied element when using ng-click

### LINKED ISSUE #6152: list: checkbox secondary action gives aria-label warning
I get the warning:
ARIA: Attribute " aria-label ", required for accessibility, is missing on node

My element is:
`<md-checkbox aria-label="Follow" class="md-secondary" ng-init="..." ng-model=".." ng-click="..."></md-checkbox>`

inside md-list-item
When I check the html code of the page I see that the parent of the checkbox is a `<bottom>` with the aria-label

DevVersion:
* Currently when using a proxied element (like `md-checkbox`) as a secondary item and having the `ng-click` attribute on it, the list is wrapping the proxied element inside of an button.
  This isn't valid, because the proxied elements are already clickable.

* Proxied Elements are still not used as proxies for the list, when they use `ng-click` on it.

@ThomasBurleson 
> Regarding the second point here, this is how it currently works (accidentally), but there is no "correct" / "wrong" solution, so I decided to keep that behavior to not break anything. 
> Also I think this should not land in 1.1.0 

Fixes #6152.

### DIFF
--- a/src/components/list/list.js
+++ b/src/components/list/list.js
@@ -349,9 +349,11 @@ function mdListItemDirective($mdAria, $mdConstant, $mdUtil, $timeout) {
       }
 
       function wrapSecondaryItem(secondaryItem, container) {
-        // If the current secondary item is not a button, but contains a ng-click attribute,
-        // the secondary item will be automatically wrapped inside of a button.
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@DevVersion - good to know.

DevVersion:
Will be addressed with #9194

PR #9192, opened 2016-07-30
feat(browser-color): enable browser header coloring

### LINKED ISSUE #8062: meta "theme-color" relative to toolbar theme
Hi,
How to change  `<meta name="theme-color" content="#546E7A">`  content color relative to toolbar theme ?

Any idea is welcome.
Thank you.

EladBezalel:
- provides an easy way to set and change the browser color according the material design spec colors and the defined themes

fixes #8062

### DIFF
--- a/src/core/services/meta/meta.js
+++ b/src/core/services/meta/meta.js
@@ -1,0 +1,120 @@
+/**
+ * @ngdoc service
+ * @name $$mdMeta
+ * @module material.core.meta
+ *
+ * @description
+ *
+ * A provider and a service that simplifies meta tags access
+ *
+ * Note: This is intended only for use with dynamic meta tags such as browser color and title.
+ * Tags that are only processed when the page is rendered (such as `charset`, and `http-equiv`)
+ * will not work since `$$mdMeta` adds the tags after the page has already been loaded.
+ *
+ * ```js
+ * app.config(function($$mdMetaProvider) {
+ *   var removeMeta = $$mdMetaProvider.setMeta('meta-name', 'content');
+ *   var metaValue  = $$mdMetaProvider.getMeta('meta-name'); // -> 'content'
+ *
+ *   removeMeta();
+ * });
+ *
+ * app.controller('myController', function($$mdMeta) {
+ *   var removeMeta = $$mdMeta.setMeta('meta-name', 'content');
+ *   var metaValue  = $$mdMeta.getMeta('meta-name'); // -> 'content'
+ *
+ *   removeMeta();
+ * });
+ * ```
+ *
+ * @returns {$$mdMeta.$service}
+ *
+ */
+angular.module('material.core.meta', [])
+  .provider('$$mdMeta', function () {
+    var head = angular.element(document.head);
+    var metaElements = {};
+
+    /**
+     * Checks if the requested element was written manually and maps it
+     *
+     * @param {string} name meta tag 'name' attribute value
+     * @returns {boolean} returns true if there is an element with the requested name
+     */
+    function mapExistingElement(name) {
+      if (metaElements[name]) {
+        return true;
+      }
+
+      var element = document.getElementsByName(name)[0];
+
+      if (!element) {
+        return false;
+      }
+
+      metaElements[name] = angular.element(element);
+
+      return true;
+    }
+
+    /**
+     * @ngdoc method
+     * @name $$mdMeta#setMeta
+     *
+     * @description
+     * Creates meta element with the 'name' and 'content' attributes,
+     * if the meta tag is already created than we replace the 'content' value
+     *
+     * @param {string} name meta tag 'name' attribute value
+     * @param {string} content meta tag 'content' attribute value
+     * @returns {function} remove function
+     *
+     */
+    function setMeta(name, content) {
+      mapExistingElement(name);
+
+      if (!metaElements[name]) {
+        var newMeta = angular.element('<meta name="' + name + '" content="' + content + '"/>');
+        head.append(newMeta);
+        metaElements[name] = newMeta;
+      }
+      else {
+        metaElements[name].attr('content', content);
+      }
+
+      return function () {
+        metaElements[name].attr('content', '');
+        metaElements[name].remove();
+        delete metaElements[name];
+      };
+    }
+
+    /**
+     * @ngdoc method
+     * @name $$mdMeta#getMeta
+     *
+     * @description
+     * Gets the 'content' attribute value of the wanted meta element
+     *
+     * @param {string} name meta tag 'name' attribute value
+     * @returns {string} content attribute value
+     */
+    function getMeta(name) {
+      if (!mapExistingElement(name)) {
+        throw Error('$$mdMeta: could not find a meta tag with the name \'' + name + '\'');
+      }
+
+      return metaElements[name].attr('content');
+    }
+
+    var module = {
+      setMeta: setMeta,
+      getMeta: getMeta
+    };
+
+    return angular.extend({}, module, {
+      $get: function () {
+        return module;
+      }
+    });
+  });
--- a/src/core/services/meta/meta.spec.js
+++ b/src/core/services/meta/meta.spec.js
@@ -1,0 +1,102 @@
+describe('$$mdMeta', function() {
+  var $$mdMeta;
+
+  beforeEach(module('material.core'));
+
+  beforeEach(function() {
+    inject(function(_$$mdMeta_) {
+      $$mdMeta = _$$mdMeta_;
+    });
+  });
+
+  describe('set meta', function () {
+    beforeEach(function () {
+      angular.element(document.getElementsByTagName('meta')).remove();
+    });
+
+    it('should create the element and append it to the body', function() {
+      var name = 'test';
+      var content = 'value';
+
+      expect(document.getElementsByName(name).length).toEqual(0);
+
+      $$mdMeta.setMeta(name, content);
+
+      expect(angular.element(document.getElementsByName(name)[0]).attr('content')).toBe(content);
+    });
+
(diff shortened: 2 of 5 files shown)


src/core/services/theming/theming.js
```diff
@@ -242,6 +270,53 @@ function ThemingProvider($mdColorPalette) {
...
+   * options are:
+   * `theme`   - A defined theme via `$mdThemeProvider` to use the palettes from. Default is `default` theme
+   * `palette` - Can be any one of the basic material design palettes, extended defined palettes and 'primary',
+   *             'accent', 'background' and 'warn'. Default is `primary`
+   * `hue`     - The hue from the selected palette. Default is `800`
+   *
+   * @param {Object} options Options object for the browser color
+   * @returns {Function} remove function of the browser color
+   */
+  var enableBrowserColor = function (options){
```
DevVersion: Nit: Brace after function

src/core/services/theming/theming.js
```diff
@@ -242,6 +270,53 @@ function ThemingProvider($mdColorPalette) {
 
   // Default theme defined in core.js
 
+  /**
+   *
+   * @param color
+   * @returns {Function}
+   */
+  var setBrowserColor = function (color) {
+    <!-- Chrome, Firefox OS and Opera -->
```
DevVersion: Not sure why you use ta HTML comment-opening sequence here, would be more consistent to use a basic `//` or `/* */`

src/core/services/meta/meta.js
```diff
@@ -0,0 +1,86 @@
...
+    function getMeta(name) {
+      return metaElements[name].attr('content');
+    }
+
+    var module = {
+      setMeta: setMeta,
+      getMeta: getMeta
+    };
+
+    return angular.extend({}, module, {
```
DevVersion: Just for ping: As we discussed on Slack, would be better to use it differently.

```js
module.$get = function() { return module }
return module;
```

DevVersion:
@ThomasBurleson That looks very good. LGTM.

src/core/services/theming/theming.js
EladBezalel: Lol i copied that from the spec, totally would change that

DevVersion: Hahah I thought `wtf, where is the HTML fragment here??` :smile:

ErinCoughlan:
@EladBezalel tests?

EladBezalel:
@ErinCoughlan done

src/core/services/meta/meta.js
```diff
@@ -0,0 +1,116 @@
+/**
+ * @ngdoc service
+ * @name $$mdMeta
+ * @module material.core.meta
+ *
+ * @description
+ *
+ * A provider and a service that simplifies meta tags access
```
topherfangio: Might make a note that certain `<meta>` tags (such as `charset`, `http-equiv="Content-Type"`, and I think `name="viewport"`) cannot be added dynamically because the browser will have already decided how to render (and probably rendered) the page by the time they get added.

I could not find a complete list of the ones that were editable, but this caught me the other day so it's probably not very obvious to most devs :-)

src/core/services/meta/meta.js
```diff
@@ -0,0 +1,116 @@
...
+     *
+     * @description
+     * Gets the 'content' attribute value of the wanted meta element
+     *
+     * @param {string} name meta tag 'name' attribute value
+     * @returns {string} content attribute value
+     */
+    function getMeta(name) {
+      if (!mapExistingElement(name)) {
+        throw Error('$$mdMeta: could not find a meta tag with the name \'' + name + '\'');
```
topherfangio: I see a lot of people take this approach with strings of using single quotes and then escaping them...

Is there a reason you don't just use double-quotes to wrap the outer string? Totally just my curiosity at play here 😄

src/core/services/theming/theming.js
```diff
@@ -93,6 +93,23 @@ function detectDisabledThemes($mdThemingProvider) {
  *     $mdThemingProvider.registerStyles(require('../styles/my-component.theme.css'));
  *   });
  * </hljs>
+ *
+ * ### Browser color
+ *
+ * Enables browser header coloring
+ * for more info please visit:
+ * https://developers.google.com/web/fundamentals/design-and-ui/browser-customization/theme-color
+ *
```
topherfangio: I think it might be helpful to add some notes about how the theme, palette and hue affect the final color.

src/core/services/theming/theming.js
```diff
@@ -282,6 +357,8 @@ function ThemingProvider($mdColorPalette) {
       alwaysWatchTheme = alwaysWatch;
     },
 
+    enableBrowserColor: enableBrowserColor,
+
```
topherfangio: I think it might be helpful to go ahead and provide/document the `setBrowserColor()` method here too. I can see some people just wanting to set it to white/black and it being cumbersome to figure out the proper theme/palette/hue.

topherfangio:
Left a few comments for consideration, but looks pretty great to me!

src/core/services/meta/meta.js
EladBezalel: I just hate double quotes :sweat_smile:

src/core/services/theming/theming.js
EladBezalel: `setBrowserColor` is the same as `enableBrowserColor`

src/core/services/theming/theming.js
```diff
@@ -242,6 +270,53 @@ function ThemingProvider($mdColorPalette) {
 
   // Default theme defined in core.js
 
+  /**
+   *
```
EladBezalel: Forgot documentation

src/core/services/theming/theming.js
topherfangio: Doesn't `enableBrowserColor()` call `setBrowserColor()`?

EladBezalel: Are you talking about [this](https://github.com/angular/material/pull/9192/files#diff-2a98812fedf7173f142f890cb0c29b85R544) or [that](https://github.com/angular/material/pull/9192/files#diff-2a98812fedf7173f142f890cb0c29b85R278)?

src/core/services/meta/meta.js
ErinCoughlan: JS Style Guide also says to always use single quotes.

src/core/services/meta/meta.spec.js
```diff
@@ -0,0 +1,95 @@
+describe('$$mdMeta', function() {
+
+  beforeEach(module('material.core'));
+
+  describe('set meta', function () {
+    beforeEach(function () {
+      angular.element(document.getElementsByTagName('meta')).remove();
+    });
+
+    it('should create the element and append it to the body', inject(function($$mdMeta) {
```
ErinCoughlan: Does it make sense to just inject in a beforeEach instead of within every test (since it's used everywhere)? I've always hated the inject per test design.

Something like this (with or without the extra method).
```
var injectLocals = function($$mdMeta) {
  mdMeta = $$mdMeta;
};

beforeEach(function() {
  inject(injectLocals);
});
 ```

File: src/core/services/theming/theming.js (hunk 7 above)
Comment (EladBezalel) [inline src/core/services/theming/theming.js]: Done by adding:
```js
 setBrowserColor: enableBrowserColor,
```

src/core/services/theming/theming.js
EladBezalel: Done

src/core/services/meta/meta.js
EladBezalel: Done

src/core/services/theming/theming.js
topherfangio: Correct me if I'm wrong, but the `setBrowserColor()` function takes an options object, not a single color, right? So you would still have to pass an options param of theme, palette and hue? That's kinda opposite of what I was suggesting.

EladBezalel: Oh so you we're suggesting that we would enable them the option of setting what ever color they want?
I'm against it, in that case i'd say to them, use it manually.

topherfangio: Normally, I wouldn't fight you on this and I'd say "ok, cool" and move on...but...

I feel like we're being needlessly restrictive by not providing it. Part of the convenience of this service/provider is hiding all of the logic about which meta tags to set for which browsers.

Additionally, I don't really see a downside of exposing it. If they want to use the theme colors, then they should definitely use the `enableBrowserColor()` approach, but if they want to set it to something that isn't part of the theme (or maybe any theme), the giving them a way to set it directly would be nice.

What downsides do you see?

src/core/services/meta/meta.spec.js
```diff
@@ -0,0 +1,102 @@
...
+      var name = 'test';
+      var content = 'value';
+
+      var element = angular.element('<meta name="' + name + '" content="' + content + '"/>');
+      angular.element(document.head).append(element);
+
+      expect($$mdMeta.getMeta(name)).toBe(content);
+    });
+  });
+});
```
ErinCoughlan: no new line

clshortfuse:
How do you feel about a directive based around a class or attribute, that when added to an element, takes it's background color or `md-primary`,`md-warn`,`md-accent` reference and applies that color to the browser color.

For example `<md-toolbar class="browser-color" md-colors="{background: ctrl.currentNavColor}"> </mdtoolbar>`

This is what I use to achieve a similar toolbar affect in Angular Material as done in https://inbox.google.com/

topherfangio:
@clshortfuse I think providing the service is enough. Developers should be able to figure out what color/hue the toolbar will use and match that relatively easily.

topherfangio:
@ThomasBurleson LGTM 👍

PR #9193, opened 2016-07-31
feat(menu): expose close method on element scope; deprecate $mdOpenMenu

### LINKED ISSUE #8446: Enhancement: Support $mdCloseMenu
We can open a menu using:

**ng-mouseover="$mdOpenMenu($event)"**

but it would be great if something like mdCloseMenu was also supported:

**ng-mouseleave="$mdCloseMenu($event)"**

crisbeto:
* Deprecates the `$mdOpenMenu` method in favor of `$mdMenu.open` in order to simplify the API.
* Exposes the `$mdMenu.close` method on the menu's scope, allowing for custom closing behavior.

Fixes #8446.

### DIFF
--- a/src/components/menu/js/menuDirective.js
+++ b/src/components/menu/js/menuDirective.js
@@ -10,9 +10,10 @@
  *
  * Every `md-menu` must specify exactly two child elements. The first element is what is
  * left in the DOM and is used to open the menu. This element is called the trigger element.
- * The trigger element's scope has access to `$mdOpenMenu($event)`
+ * The trigger element's scope has access to `$mdMenu.open($event)`
  * which it may call to open the menu. By passing $event as argument, the
- * corresponding event is stopped from propagating up the DOM-tree.
+ * corresponding event is stopped from propagating up the DOM-tree. Similarly, `$mdMenu.close()`
+ * can be used to close the menu.
  *
  * The second element is the `md-menu-content` element which represents the
  * contents of the menu when it is open. Typically this will contain `md-menu-item`s,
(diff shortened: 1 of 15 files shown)


src/components/menu/js/menuDirective.js
```diff
@@ -123,16 +124,18 @@
  * Sometimes you would like to be able to click on a menu item without having the menu
  * close. To do this, ngMaterial exposes the `md-prevent-menu-close` attribute which
  * can be added to a button inside a menu to stop the menu from automatically closing.
- * You can then close the menu programatically by injecting `$mdMenu` and calling 
- * `$mdMenu.hide()`.
+ * You can then close the menu either by using `$mdCloseMenu()` in the template,
+ * or programatically by injecting `$mdMenu` and calling `$mdMenu.hide()`.
```
DevVersion: Typo in `programmatically`

crisbeto:
I had updated the usage example @EladBezalel, or do you want a "proper" demo?

crisbeto:
@EladBezalel it's updated to have a demo showing how to use the custom triggers.

crisbeto:
Marking as `needs: work` in order to simplify the API as agreed with @ThomasBurleson. The `$mdOpenMenu` and `$mdCloseMenu` scope methods shouldn't be exposed separately, but rather as a part of the `scope.$mdMenu` open. The old methods should still be supported, but show a deperacation warning.

EladBezalel:
I think it deserves a proper demo

crisbeto:
@ThomasBurleson I've updated this to include:
* Deprecation of the `$mdOpenMenu` method via a warning.
* Include the demo requested by @EladBezalel.

DevVersion:
LGTM.

cmacdonnacha:
Nice work on this. Will $mdOpenMenu break if I update from 1.1.1 to 1.1.3?

crisbeto:
It won't break, but it will log a deprecation warning.

cmacdonnacha:
Great thanks. I see that a [breaking change](https://github.com/angular/material/blob/d6b2c9ef8932869a201a5c7305083831853c39cc/CHANGELOG.md#breaking-changes) has been introduced to 1.1.2. Should the minor version be bumped instead of the patch version?

crisbeto:
I'm not sure, I haven't been involved with that particular change.

Issue #8482, opened 2016-05-17
Restructure the list component


crisbeto:
As discussed with @ThomasBurleson, we should try to simplify the list component at some point, because it shuffles the DOM a lot and is overly complicated. This can lead to issues like #6152 and #8172.

DevVersion:
@crisbeto - Elad and I already worked on that.

crisbeto:
Ah cool, is there an issue for it so I can reference it in the ones?

DevVersion:
There is no issue yet, so I think it's good to use your issue now.

- https://github.com/angular/material/issues/7742#issuecomment-201261002

Splaktar:
The CSS was simplified in PR https://github.com/angular/material/pull/9194. It's not clear whether or not we still need to restructure the list component. 

This issue only mentions that the list component
> shuffles the DOM a lot and is overly complicated. This can lead to issues like #6152 and #8172.

Both of the referenced issues are closed, but it's not clear that the second is actually fixed.
I also do not currently have access to the design doc. I may be able to get access, but I'm not sure if it is still relevant.

I'll try to re-visit this when more of the P1, P2, and P3 issues are resolved.

Splaktar:
It turns out that the CSS was not simplified as the previous PR was reverted in PR https://github.com/angular/material/pull/10221 (before ever being released in `1.1.2`).

Here's the status on the issues mentioned here:
 - https://github.com/angular/material/issues/8172: fixed in `1.1.13`
- https://github.com/angular/material/issues/6152: The fix was reverted and this is still reproducible.
- https://github.com/angular/material/issues/8890: The fix was reverted and this is still reproducible.

PR #9194, opened 2016-07-31
refactor(list): simplify list component scss

### LINKED ISSUE #6152: list: checkbox secondary action gives aria-label warning

### LINKED ISSUE #8890: list: dense list cuts off some letters like 'g'
Notice how 'g's get cutoff (at least latest Chrome and Firefox, Windows 10)

http://codepen.io/anon/pen/OXpQmV

DevVersion:
- Simplifies the List components SCSS
- Fixes Dense Mode for List and introduces super elegant approach
- Make SCSS more modular by having a base and dynamic variables
- Remove unnecessary DOM node level for button wrap.
- No longer wrap proxied elements inside of buttons
- Multiple Secondary Items should align properly and have consistent spacing.
- Dense Mode should not cut characaters on different environments.

@EladBezalel and I had a short meeting and I showed him some of the new approaches.

R: @EladBezalel 

Fixes #6152. Fixes #8890 References #8482.

### DIFF
--- a/src/components/list/list-base.scss
+++ b/src/components/list/list-base.scss
@@ -1,0 +1,317 @@
+/** External component calculations */
+$md-checkbox-width: 3 * $baseline-grid !default;
+$md-button-margin: rem(0.800) !default;
+$md-button-spacing: $button-left-right-padding + $md-button-margin !default;
+
+/** Static List Item Variables */
+$md-list-item-indention: $baseline-grid * 7 !default;
+$md-list-item-padding: $baseline-grid * 2 !default;
+$md-list-item-h3-margin: 0 0 0px 0 !default;
+$md-list-item-h4-margin: 3px 0 1px 0 !default;
+$md-list-item-h4-font-weight: 400 !default;
+$md-list-item-p-margin: 0 0 0 0 !default;
+$md-list-item-subheader-height: 6 * $baseline-grid !default;
+
+/** Resets the checkbox component */
+@mixin md-reset-checkbox() {
+  md-checkbox {
+    align-self: center;
+    margin: 0;
+
+    .md-label {
+      display: none;
+    }
+  }
+}
+
+/** Resets the switch component to properly align */
+@mixin md-reset-switch() {
+  md-switch {
+    margin-top: 0;
+    margin-bottom: 0;
+
+    @include rtl-prop(margin-right, margin-left, -6px, 0);
+  }
+}
+
+/** Reset browsers default paragraph styles */
+@mixin md-reset-paragraph() {
+  p {
+    margin: 0;
+    flex: 1;
+  }
+}
+
+/** Reset default md-button behavior like uppercase and text alignment */
+@mixin md-reset-button() {
+  &.md-button {
+    font-size: inherit;
+    text-transform: none;
+
+    @include rtl(text-align, left, right);
+
+    &, > .md-ripple-container {
+      border-radius: 0;
+      margin: 0;
+      padding: 0;
+    }
+  }
+}
+
+/** Mixin to overwrite the size of an md-icon element */
+@mixin md-icon-size($icon-size) {
+  min-width: $icon-size;
+  min-height: $icon-size;
+  height: $icon-size;
+  width: $icon-size;
+
+  svg {
+    width: $icon-size;
+    height: $icon-size;
+  }
+}
+
+/** Creates an overlaying button, which will be used to execute the primary action */
+@mixin md-button-wrap() {
+  &.md-button-wrap {
+    position: relative;
+
+    .md-button-wrap-executor {
+      position: absolute;
+      top: 0;
+      left: 0;
+
+      height: 100%;
+      width: 100%;
+
+      @include md-reset-button();
+    }
+  }
+}
+
+/**
+ * Mixin to style user specified content inside of the list item.
+ */
+@mixin md-list-item-content($list-icon-width, $list-avatar-width, $list-avatar-padding-dense) {
+
+  .md-avatar, .md-avatar-icon, > md-icon:first-child {
+    box-sizing: content-box;
+    margin-top: $baseline-grid;
+    margin-bottom: $baseline-grid;
+    border-radius: 50%;
+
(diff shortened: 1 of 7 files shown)


src/components/list/list-base.scss
```diff
@@ -0,0 +1,300 @@
+/** External component calculations */
+$md-checkbox-width: 3 * $baseline-grid !default;
+//TODO(devversion): use shared variable from buttons.scss
```
EladBezalel: If it's for future work please file an issue and write it's number on the comment

src/components/list/list-base.scss
```diff
@@ -0,0 +1,300 @@
...
+
+      @include md-reset-button();
+    }
+  }
+}
+
+/**
+ * Mixin to style user specified content inside of the list item.
+ */
+@mixin md-list-item-content($_list-icon-width, $_list-avatar-width, $_list-avatar-padding-dense) {
```
EladBezalel: As we talked, I don't really like the `$_` prefix for parameters.. maybe local variables should have it instead?

src/components/list/list.js
```diff
@@ -244,6 +244,7 @@ function mdListItemDirective($mdAria, $mdConstant, $mdUtil, $timeout) {
           wrapIn('div');
         } else if (!tEl[0].querySelector('md-button:not(.md-secondary):not(.md-exclude)')) {
           tEl.addClass('md-no-proxy');
+          tEl.addClass('md-list-item-content');
```
EladBezalel: Are you sure this class should be added during the compile phase?

if so i'd minify the calls to:
```js
tEl.addClass('md-no-proxy md-list-item-content');
```

EladBezalel:
I really like the approach we should consider future work on dense components to behave the same.

> Sort of related question:
> Are we applying `dense` mode automatically on small screens?

@clshortfuse

src/components/list/list.js
DevVersion: Yeah that's intentionally. It allows us to easily apply the custom styles for user content.

src/components/list/list-base.scss
DevVersion: Yeah we both agreed on that. My concern is just that it should be clear whether a variable is dynamic or static.

src/components/list/list-base.scss
DevVersion: Since it's an issue, which is not depending on that PR, it would make sense to create it.
> See #9196

DevVersion:
@EladBezalel - Yeah as discussed in our talk. I'll create a follow-up PR, which makes the dense helpers more generic and usable for other components. I'll create an issue once this PR is available.

> Are we applying dense mode automatically on small screens?

No we don't.

src/components/list/list-base.scss
```diff
@@ -99,6 +99,8 @@ $md-list-item-subheader-height: 6 * $baseline-grid !default;
     margin-top: $baseline-grid;
     margin-bottom: $baseline-grid;
     border-radius: 50%;
+
+    flex-shrink: 0; // Per https://bugs.webkit.org/show_bug.cgi?id=146020
```
DevVersion: It's resolved in the latest Webkit version, but it's not resolved in the latest Safari version (probably in the Technology Preview).

I tested that on my Mac Book Pro and the bug is still valid.

DevVersion:
@topherfangio Friendly ping for review if you find some time ;)

src/components/list/list.spec.js
```diff
@@ -355,25 +341,69 @@ describe('mdListItem directive', function() {
...
+
+    var itemContent = listItem.children().eq(1);
+    var secondaryContainer = itemContent.children().eq(1);
+
+    expect(secondaryContainer).toHaveClass('md-secondary-container');
+
+    // The secondary container should contain the md-checkbox, without any button as parent.
+    var checkboxItem = secondaryContainer.children()[0];
+
+    expect(checkboxItem.nodeName).toBe('MD-CHECKBOX');
```
DevVersion: This does pass, but I will change it.

DevVersion:
@topherfangio Thanks for the review.

1. Yeah I'm aware of that, and the change is intentional (follows the specs - [Image](refactor/list-component-scss))
2. Thanks 😄 
3. Yeah this is also intentional, because the items with a proxied element no longer exceed the height (provides consistency and follows specs [Image](https://i.gyazo.com/da2995b4880f4b2ffa39d7484965d2ff.png))

src/components/list/list.spec.js
```diff
@@ -355,25 +341,69 @@ describe('mdListItem directive', function() {
...
+        '<p>Hello World</p>' +
+        '<md-checkbox class="md-secondary" ng-click="null" ng-model="isChecked"></md-checkbox>' +
+      '</md-list-item>'
+    );
+
+    var checkboxEl = listItem.find('md-checkbox');
+
+    expect($rootScope.isChecked).toBeFalsy();
+
+    var clickListener = listItem[0].querySelector('div');
```
DevVersion: Yeah I agree with that. Just changed and also added a comment :)

src/components/list/list.spec.js
```diff
@@ -355,25 +341,69 @@ describe('mdListItem directive', function() {
...
+      '<md-list-item ng-click="sayHello()">' +
+        '<p>Hello World</p>' +
+        '<md-checkbox class="md-secondary" ng-click="goWild()"></md-checkbox>' +
+      '</md-list-item>'
+    );
+
+    expect(listItem).toHaveClass('md-button-wrap');
+
+    // The item should contain the button executor and the item content.
+    expect(listItem.children().length).toBe(2);
```
topherfangio: 👍

jelbourn:
@DevVersion can you rebase?

DevVersion:
@jelbourn Done.

PR #9203, opened 2016-08-02
fix(icon): implicity trust icon urls given during angular config phase

EladBezalel:
Original author is @jelbourn

### DIFF
--- a/src/components/icon/js/iconService.js
+++ b/src/components/icon/js/iconService.js
@@ -276,13 +276,19 @@
  *
  */
 
-var config = {
-  defaultViewBoxSize: 24,
-  defaultFontSet: 'material-icons',
-  fontSets: []
-};
+
+/**
+ * The configuration for $mdIconProvider. This contains both options for the icon service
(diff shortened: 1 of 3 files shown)


src/components/icon/js/iconService.js
```diff
@@ -554,11 +570,16 @@ function MdIconService(config, $templateRequest, $q, $log, $mdUtil, $sce) {
...
+        if (configUrls.has(url)) {
+          url = $sce.trustAsResourceUrl(url);
+        }
+
         $templateRequest(url, true).then(extractSvg, announceAndReject);
       });
     }
 
-    return dataUrlRegex.test(url)
+    var testUrl = angular.isString(url) ? url : $sce.getTrustedResourceUrl(url);
```
jelbourn: This line isn't necessary any more now that the `trustAs` occurs later.

ThomasBurleson: @jelbourn - what should the line be?

jelbourn: Just remove it entirely and use the `url` directly in the check

PR #9207, opened 2016-08-02
fix(tooltip) correct background and foreground color

### LINKED ISSUE #9159: md-tooltip contrast issue with dark themes
**Actual Behavior**:
 -  `What is the issue? *` The contrast between background and foreground colors isn't high enough for `md-tooltip`s on dark themes. The text is then nearly unreadable.
 -  `What is the expected behavior?` The tooltip text should use the opposite hue in the background palette for dark themes, ie. `background-A700`.

 -  `CodePen Demo which shows your issue:` http://codepen.io/anon/pen/EyROqx
 -  `Details:` As you can see, the text is really hard to read as it is.

**Angular Versions**: *
 -  `Angular Version:` 1.5.8
 -  `Angular Material Version:` 1.1.0-rc5

**Additional Information**:
 -  `Browser Type: *` chrome
 -  `Browser Version: *` 51.0.2704.103 m
 -  `OS: *` Windows 7
 -  `Stack Traces:` -

clshortfuse:
Use Material Design colors for tooltip
 - Set tooltip opacity to 0.90
 - Use background-700 for background color
 - Use background-700-contrast for foreground color

Previously, the background color was set to a foreground-2 and the
foreground color was set to A100 of the background which causes
visibility issues when either changing the background palette from
grey or using a dark theme.

Fixes #9159

### DIFF
--- a/src/components/tooltip/tooltip-theme.scss
+++ b/src/components/tooltip/tooltip-theme.scss
@@ -1,6 +1,6 @@
 md-tooltip.md-THEME_NAME-theme {
-  color: '{{background-A100}}';
+  color: '{{background-700-contrast}}';
   .md-content {
-    background-color: '{{foreground-2}}';
+    background-color: '{{background-700}}';
   }
 }
--- a/src/components/tooltip/tooltip.scss
+++ b/src/components/tooltip/tooltip.scss
@@ -47,7 +47,7 @@ md-tooltip {
     }
     &.md-show, &.md-show-add-active {
       transform: scale(1);
-      opacity: 1;
+      opacity: 0.9;
       transform-origin: center top;
     }
     &.md-show-remove {


clshortfuse:
See https://material.google.com/components/tooltips.html#tooltips-tooltips-desktop-

topherfangio:
@clshortfuse LGTM 👍 

Can you add a quick screenshot to the PR description showing before/after just so we can easily see how it's better?

clshortfuse:
It might take a while but, in the meantime, this is the codepen that shows the issue.

http://codepen.io/anon/pen/EyROqx

clshortfuse:
Before
Before
After
After

ThomasBurleson:
@topherfangio, @clshortfuse - *before* -n- *after* snapshots are GREAT!

PR #9208, opened 2016-08-02
fix(navBar): Validate tab controller is defined before calling find fn

jadengore:
When `_getTabs` is called on a `md-nav-bar` controller, the item's controller may not have been initialized yet. The implementation looks for the controller by using `angular.element` on line 259:

```js
return angular.element(el).controller('mdNavItem')
```

If the controller is not yet defined, we get an `undefined` controller in our list of tab controllers. This means when `_findTab` is called, it will try and call the finder function with `undefined` as the parameter. Since these functions expect prototype methods from the `MdNavItemController`, I end up getting this error:

```
TypeError: Cannot read property 'getName' of undefined
```

This change guards against the tab controller being `undefined` by checking for its existence before calling the finder function.

### DIFF
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -300,7 +300,7 @@ MdNavBarController.prototype.getFocusedTab = function() {
 MdNavBarController.prototype._findTab = function(fn) {
   var tabs = this._getTabs();
   for (var i = 0; i < tabs.length; i++) {
-    if (fn(tabs[i])) {
+    if (tabs[i] && fn(tabs[i])) {
       return tabs[i];
     }
   }
--- a/src/components/navBar/navBar.spec.js
+++ b/src/components/navBar/navBar.spec.js
@@ -156,6 +156,23 @@ describe('mdNavBar', function() {
 
       expect($scope.selectedTabRoute).toBe('tab2');
     });
+
+    it('does not find tab if no controllers exist on tabs', function() {
+        createTabs();
+
+        // Simulate _getTabs returning undefined tab controllers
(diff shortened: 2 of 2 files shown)


jadengore:
@ThomasBurleson Sorry to ping directly, but I'm curious if this will be reviewed/merged. If not I can just close it.

ThomasBurleson:
@jadengore - this PR looks reasonable; but it needs a unit test before merging. Can you do that ?
We will target a merge this for v1.1.2

jadengore:
@ThomasBurleson Added a test that demonstrates the change works. To demonstrate it, the controllers have to return `undefined`, so I mocked `_getTabs` to return `undefined` to demonstrate the error.

clshortfuse:
Possibly related to #9071

We might not have to merge both, since seem to resolve around the same issue of `mdNavItem` having an `undefined` value. Unfortunately, this is why issues should be filed before PRs are built.

jadengore:
@clshortfuse If you do merge #9071 then I should probably check and make sure my tests pass with the change. Otherwise, this will probably still be a necessary change.

gmoothart:
This is also related to #9386

gmoothart:
I think the root of the problem is in _updateTabs running when the nav-bar is not initialized yet (so #9071 looks like a better direction). @jadengore do you have an example of the error being triggered other than in an  _updateTabs call when the nav-bar isn't initialized?

ThomasBurleson:
Closing as invalid change.

jadengore:
@ThomasBurleson Why is this change invalid? Can we at least run these tests against the other proposed fix to make sure this is resolved?

PR #9209, opened 2016-08-02
feat(mdTable):  Data Table component

DevVersion:
Initial Implementation of the Material Design Data Table.

> FYI: @daniel-nagy and I are still working on it.
> Also Credits to @daniel-nagy


src/components/dataTable/cell.scss
```diff
@@ -0,0 +1,65 @@
+$table-head-font-size: 12px;
+$table-body-font-size: 13px;
```
DevVersion: Prefix with `md` and probably use variables from `typeography.scss`

src/components/dataTable/cell.scss
```diff
@@ -0,0 +1,65 @@
...
+table.md-table td {
+  text-align: left;
+  white-space: nowrap;
+  vertical-align: middle;
+
+  &:first-child {
+    padding: 0 24px;
+  }
+
+  &:nth-child(n+2):nth-last-child(n+2) {
```
DevVersion: Add comment explaining why that selector is supposed to match every nth child.

src/components/dataTable/cell.scss
```diff
@@ -0,0 +1,65 @@
...
+    padding: 0 56px 0 0;
+  }
+
+  &:last-child {
+    padding: 0 24px 0 0;
```
DevVersion: Transform padding variables (above as well) into variables

src/components/dataTable/cell.scss
```diff
@@ -0,0 +1,65 @@
...
+    text-align: right;
+  }
+
+  &.md-align-center {
+    text-align: center;
+  }
+
+  &.md-align-justify {
+    text-align: justify;
+  }
```
DevVersion: This could be probably a mixin.

src/components/dataTable/checkbox.scss
```diff
@@ -0,0 +1,10 @@
+table.md-table {
+
+  th, td {
+    md-checkbox {
+      width: 20px;
```
DevVersion: Use variables here as well. 

Also hide checkbox label and align it self in the center (as done in the list as well)

src/components/dataTable/demoBasicUsage/index.html
```diff
@@ -0,0 +1,44 @@
...
+      <thead md-head md-order="ctrl.order" md-on-reorder="$log.info(ctrl.order)">
+        <tr md-row>
+          <th md-cell md-order-by="name">Dessert (100g serving)</th>
+          <th md-cell md-align="right" md-order-by="calories">Calories</th>
+          <th md-cell md-align="right" md-order-by="fat">Fat (g)</th>
+          <th md-cell md-align="right" md-order-by="carbs">Carbs (g)</th>
+          <th md-cell md-align="right" md-order-by="protein">Protein (g)</th>
+          <th md-cell md-align="right" md-order-by="sodium">Sodium (mg)</th>
+          <th md-cell md-align="right" md-order-by="calcium">Calcium (%)</th>
+          <th md-cell md-align="right" md-order-by="iron">Iron (%)</th>
```
DevVersion: We should reduce some rows here, because the demo looks a bit weird with having scrollbars.

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+  };
+
+  function postLink(scope, element, attrs) {
+    attrs.$observe('mdAlign', function(align) {
+      element.data('align', align);
+    });
+  }
+}
+
+function MdTableOrderByDirective() {
```
DevVersion: We should create some comments for each prototype function to explain what it's supposed to do.

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+
+function MdTableOrderByController($element, $scope, $attrs, $compile, $mdUtil, $$mdSvgRegistry) {
+  this.$element = $element;
+  this.$scope = $scope;
+  this.$attrs = $attrs;
+  this.$compile = $compile;
+  this.$mdUtil = $mdUtil;
+  this.$$mdArrowUpIcon = $$mdSvgRegistry.mdArrowUp;
+
+  this.isActiveRegex = new RegExp('^-?' + this.$attrs.mdOrderBy + '$');
```
DevVersion: We should move the comment (I added below) up here, because the regex is declared here.

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+
+  this.$attrs.$observe('mdOrderBy', function (orderBy) {
+    if (orderBy) {
+      this.enableSorting();
+    } else {
+      this.disableSorting();
+    }
```
DevVersion: Can be simplified by using a ternary operator.

src/components/dataTable/js/tableDirective.js
```diff
@@ -0,0 +1,131 @@
...
+
+  $attrs.$observe('multiple', function (multiple) {
+    self.multiple = $mdUtil.parseAttributeBoolean(multiple);
+
+    if (self.multiple) {
+      // This transforms our previous selected value in a array for the multiple selections.
+      self.selected = [].concat(self.selected || []);
+    } else if (angular.isArray(self.selected)) {
+      // Transforms the array of multiple selections into a single selection, by using the first item.
+      self.selected = self.selected.shift();
```
DevVersion: Just doing `self.selected[0]` would be more clear and removes the unnecessary array modification. (Nitpicky)

src/components/dataTable/js/tableDirective.js
```diff
@@ -0,0 +1,131 @@
...
+    }
+  });
+
+  $attrs.$observe('mdRowSelect', function () {
+    self.toggleSelectable(self.isSelectable());
+  });
+}
+
+MdTableController.prototype.getRows = function(element) {
+  var tableElement = element || this.$element[0];
```
DevVersion: I accidentally named that variable `tableElement` - That variable should actually contain `tableBody`, since it's the `<tbody>` element.

src/components/dataTable/js/tableRowDirective.js
```diff
@@ -0,0 +1,326 @@
...
+  cellElement.addClass('md-checkbox-cell');
+
+  this.$element.prepend(cellElement);
+
+  if( this.head) {
+    // We have to call the heads onEnableSelection function, to update the toggle-all checkbox
+    // in the head row.
+    this.head.onEnableSelection();
+  } else if (this.selectCtrl) {
+    //TODO(devversion): revisit the select controller.
```
DevVersion: Pretty weird comment from myself. Can't recall why I should revisit that :smile:

src/components/dataTable/js/tableRowDirective.js
```diff
@@ -0,0 +1,326 @@
...
+  } else if(this.selectCtrl) {
+    this.selectCtrl.setSelectable(false);
+  }
+
+  // Always toggle the checkbox in the first cell of the specified row.
+  var cell = angular.element(this.getCells(0));
+
+  // We only remove the cell from the row, when the cell is marked as a temporary checkbox cell
+  if (cell && cell.hasClass('md-checkbox-cell')) {
+    cell[0].parentNode.removeChild(cell[0]);
```
DevVersion: I guess we could `jqLite` here.

```js
cell.remove();
```

src/components/dataTable/js/tableRowDirective.js
```diff
@@ -0,0 +1,326 @@
...
+  this.$element = $element;
+  this.$scope = $scope;
+  this.$attrs = $attrs;
+  this.$parse = $parse;
+  this.$compile = $compile;
+  this.$mdUtil = $mdUtil;
+
+  var expressionSplit = this.$attrs.mdSelectable.match(/^\s*(\S+)(?:\s+as\s+(\S+))?\s*$/);
+  if (!expressionSplit) {
+    throw 'mdSelectable: Invalid expression. Expected an expression in the form of "item (as alias)?".';
```
DevVersion: Note: We need to revisit the API of the `md-selectable` directive.

src/components/dataTable/cell.scss
```diff
@@ -0,0 +1,65 @@
+$table-head-font-size: 12px;
+$table-body-font-size: 13px;
+
+table.md-table th,
+table.md-table td {
+  text-align: left;
```
EladBezalel: should use our rtl mixin

src/components/dataTable/cell.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  &:nth-child(n+2):nth-last-child(n+2) {
+    padding: 0 56px 0 0;
+  }
+
+  &:last-child {
+    padding: 0 24px 0 0;
+  }
+
+  &.md-align-left {
```
EladBezalel: I assume this is used by users..
Please add `md-align-start` and `md-align-end` to support rtl as well :)

src/components/dataTable/cell.scss
```diff
@@ -0,0 +1,65 @@
...
+
+  &.md-align-justify {
+    text-align: justify;
+  }
+}
+
+table.md-table.md-selectable th,
+table.md-table.md-selectable td {
+  &:first-child {
+    width: 20px;
```
EladBezalel: Is it `20px` because of the checkbox?

src/components/dataTable/cell.scss
```diff
@@ -0,0 +1,65 @@
...
+    padding: 0 0 0 24px;
+  }
+
+  &:nth-child(2) {
```
EladBezalel: please add comment

src/components/dataTable/data-table-theme.scss
```diff
@@ -0,0 +1,45 @@
...
+    }
+  }
+
+  tr {
+    &.md-selected {
+      background: '{{ background-hue-2 }}';
+    }
+
+    &.md-auto-select:not([disabled]):hover {
+      //TODO(devversion): revisit the theming service to provide hue-200 of the background palette.
```
EladBezalel: @DevVersion just so you don't forget :)

src/components/dataTable/icon.scss
```diff
@@ -0,0 +1,37 @@
+table.md-table thead th {
```
EladBezalel: This approach, of having `icon.scss` inside table, feels kind of weird for me.

can't it be with `dataTable.scss`?

src/components/dataTable/icon.scss
```diff
@@ -0,0 +1,37 @@
...
+  md-icon {
+    width: 16px;
+    height: 16px;
+    min-width: 16px;
+    min-height: 16px;
+    font-size: 16px !important;
+    line-height: 16px !important;
```
EladBezalel: variable

src/components/dataTable/icon.scss
```diff
@@ -0,0 +1,37 @@
...
+    &:not(:first-child) {
+      margin-left: 8px;
+    }
+
+    &:not(:last-child) {
+      margin-right: 8px;
+    }
+
+    &.md-sort-icon {
+      margin-bottom: 1px;
```
EladBezalel: `1px`? something isn't right if we're using `1px`

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
+angular
```
EladBezalel: wrap it with iife

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+
+  function postLink(scope, element, attrs, ctrls) {
+    var ctrl = ctrls[0];
+    var tableCtrl = ctrls[1];
+
+    ctrl.initCtrl(tableCtrl);
+  }
+}
+
+function MdTableCellController($element, $scope) {
```
EladBezalel: We can prevent it if we'd just do it in the post link function

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+      this.$element.removeClass('md-align-' + oldAlignment);
+    }
+
+  }.bind(this));
+};
+
+MdTableCellController.prototype.getAlignment = function() {
+
+  // Transform NodeList into an Array.
+  var headRows = [].slice.call(this.table.getHead().rows);
```
EladBezalel: why use `[].slice` ?

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+
+  return  {
+    restrict: 'A',
+    require: ['mdCell', '^^mdHead'],
+    link: postLink
+  };
+
+  function postLink(scope, element, attrs) {
+    attrs.$observe('mdAlign', function(align) {
+      element.data('align', align);
```
EladBezalel: A better approach would be using `mdCell` controller and avoid transferring data via html attributes

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+    });
+  }
+}
+
+function MdTableOrderByDirective() {
+
+  return {
+    restrict: 'A',
+    require: ['mdOrderBy', 'mdCell', '^^mdOrder', '^^mdTable'],
+    scope: {},
```
EladBezalel: why isolating the scope?

src/components/dataTable/js/tableCellDirective.js
EladBezalel: Should be done on the link phase

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+
+  this.isActiveRegex = new RegExp('^-?' + this.$attrs.mdOrderBy + '$');
+
+  this.$attrs.$observe('mdOrderBy', function (orderBy) {
+    if (orderBy) {
+      this.enableSorting();
+    } else {
+      this.disableSorting();
+    }
+  }.bind(this));
```
EladBezalel: we can avoid `.bind` if we'd do it in the link phase

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+  this.clickListenerFn = this.updateSortOrder.bind(this);
+
+  // Watch for position changes to realign the sort icon.
+  this.$scope.$watch(function() {
+    return this.cellCtrl.$element[0].cellIndex;
+  }.bind(this), function() {
+    var icon = this.$element[0].querySelector('.md-sort-icon');
+
+    if (icon) {
+      if (this.$element.data('align') === 'right') {
```
EladBezalel: ternary operator

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+        this.$element.prepend(icon);
+      } else {
+        this.$element.append(icon);
+      }
+    }
+
+  }.bind(this));
```
EladBezalel: can't stand all those `.bind` can't we somehow avoid it? how it's being done on `mdPanel`?

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+};
+
+MdTableOrderByController.prototype.isActive = function() {
+  // The isActiveRegex checks the headCtrl's order property.
```
EladBezalel: This comment should be on top of this function and use `/** */` instead of `//`

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+
+MdTableOrderByController.prototype.getDirection = function() {
+  if (this.isActive()) {
+    // When the current order value is prefixed with a minus, then its order
+    // is descendant.
+    return this.orderCtrl.value.charAt(0) === '-' ? 'md-desc' : 'md-asc';
+  } else {
+    return this.$mdUtil.parseAttributeBoolean(this.$attrs.mdDesc) ? 'md-desc' : 'md-asc';
+  }
+};
```
EladBezalel: feels like the return can be combined into 1 line using ternary operator

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+
+MdTableOrderByController.prototype.compileIcon = function() {
+  return this.$compile(
```
EladBezalel: Maybe we should use a template html with `ng-if`s instead of manual compiling?

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+    '<md-icon class="md-sort-icon" ng-class="$mdOrderBy.getDirection()" ' +
+    '         md-svg-src="' + this.$$mdArrowUpIcon + '">'
+  )(this.$scope);
+};
+
+MdTableOrderByController.prototype.updateSortOrder = function() {
+  this.$scope.$applyAsync(function() {
+    var orderFilter = this.$attrs.mdOrderBy;
+    if (this.isActive()) {
```
EladBezalel: This if should be simplified

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+      var isStaticDescendant = this.$mdUtil.parseAttributeBoolean(this.$attrs.mdDesc);
+      this.orderCtrl.setSortOrder(isStaticDescendant ? '-' + orderFilter : orderFilter);
+    }
+  }.bind(this));
+};
+
+MdTableOrderByController.prototype.enableSorting = function() {
+  this.$element
+    .addClass('md-sort')
+    .on('click', this.clickListenerFn);
```
EladBezalel: what about touch?

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+
+  if (!this.$element.children().length) {
+    this.$element.contents().wrap('<span>');
```
EladBezalel: never head of `wrap`, cool! 
is it supported on `n-1` browsers versions?

src/components/dataTable/js/tableCellDirective.js
```diff
@@ -0,0 +1,213 @@
...
+  }
+
+  if (this.numeric) {
```
EladBezalel: Ternary operator

src/components/dataTable/js/tableDirective.js
```diff
@@ -0,0 +1,131 @@
...
+    bindToController: true,
+    scope: {
+      selected: '=?mdSelected'
+    },
+    compile: compileFn,
+    controller: MdTableController,
+    controllerAs: '$mdTable'
+  };
+
+  function compileFn(tElement) {
```
EladBezalel: I prefer using link function for some reason

src/components/dataTable/js/tableDirective.js
```diff
@@ -0,0 +1,131 @@
...
+  }
+}
+
+function MdTableController($attrs, $element, $scope, $mdUtil) {
+  this.$element = $element;
+  this.$attrs = $attrs;
+  this.$scope = $scope;
+  this.$mdUtil = $mdUtil;
+
+  this.unregisterWatcher = null;
```
EladBezalel: redundant?

src/components/dataTable/js/tableCellDirective.js
DevVersion: There would be no benefit from it. It would just break the Prototype Pattern. Having it here is the right place, since this is called only once as same as a constructor.

src/components/dataTable/js/tableDirective.js
```diff
@@ -0,0 +1,131 @@
...
+
+  $attrs.$observe('mdRowSelect', function () {
+    self.toggleSelectable(self.isSelectable());
+  });
+}
+
+MdTableController.prototype.getRows = function(element) {
+  var tableElement = element || this.$element[0];
+
+  return Array.prototype.filter.call(tableElement.rows, function(row) {
```
EladBezalel: just use `tableElement.rows,filter`

src/components/dataTable/cell.scss
DevVersion: Yeah it is - Should be a variable instead.

src/components/dataTable/js/tableDirective.js
```diff
@@ -0,0 +1,131 @@
...
+    // This makes sure that the row is currently visible in the DOM.
+    // We may need to filter rows leaving with ng-repeat when we incorporate pagination.
+    return !!row.offsetParent;
+  });
+};
+
+MdTableController.prototype.getBodies = function() {
```
EladBezalel: do we expect/support few bodies?

src/components/dataTable/icon.scss
DevVersion: I'm not sure why this is separated, but I can agree with that.

src/components/dataTable/js/tableCellDirective.js
DevVersion: There is no need for that. We automatically wrap each module inside of an IIFE

src/components/dataTable/js/tableDirective.js
```diff
@@ -0,0 +1,131 @@
...
+    });
+
+  } else if (this.unregisterWatcher) {
+    this.unregisterWatcher();
+    this.unregisterWatcher = null;
+  }
+};
+
+MdTableController.prototype.registerChangeListener = function(listener) {
+  return this.changeListeners.push(listener) && listener;
```
EladBezalel: what does `&& listener` means?

src/components/dataTable/js/tableCellDirective.js
DevVersion: We can avoid those things, but it is way clean and clear to follow the prototype pattern.

Also allows us to easily generate definitions from it. And provides consistency and possibilities for API access.

src/components/dataTable/js/tableDirective.js
```diff
@@ -0,0 +1,131 @@
...
+};
+
+MdTableController.prototype.unregisterChangeListener = function(listener) {
```
EladBezalel: I'd expect `registerChnageListener` to return the unregister function

src/components/dataTable/js/tableCellDirective.js
DevVersion: This is used to transform a NodeList into an Array as per comment.

>  `Array.from` is not available in ES5

src/components/dataTable/js/tableCellDirective.js
DevVersion: I definitely agree with that.

src/components/dataTable/js/tableCellDirective.js
DevVersion: No, we want to have it prototype / class based. This makes big implementations more clean and also is a best practice.

See for example `md-panel`.

src/components/dataTable/js/tableHeadDirective.js
```diff
@@ -0,0 +1,117 @@
...
+  // Always toggle the checkbox in the first cell of the specified row.
+  var cell = angular.element(this.getRows(index).cells[0]);
+
+  // When the first cell is not marked as a checkbox cell, then we're skipping
+  // the toggle action.
+  if (!cell.hasClass('md-checkbox-cell')) {
+    return;
+  }
+
+  if (state) {
```
EladBezalel: Ternary operator

src/components/dataTable/js/tableOrderDirective.js
```diff
@@ -0,0 +1,30 @@
...
+}
+
+function MdTableOrderController($scope, $mdUtil) {
+  var self = this;
+
+  self.setSortOrder = function(order) {
+    self.value = order;
+
+    $mdUtil.nextTick(function() {
+      $scope.$eval(self.onReorder);
```
EladBezalel: $parse could fit here @DevVersion ?

src/components/dataTable/js/tableCellDirective.js
DevVersion: They are using `angular.bind` (which is similar to Function.bind in the supported browsers).

Also they use `self` variables, but this is not an elegant solution.

Pretty common syntax for the prototype pattern. 

> See for example the Chrome Developers Supercharged videos etc.

src/components/dataTable/js/tableCellDirective.js
DevVersion: For sure it can. But sometimes we explicitly use a normal `if`, because it allows us to add comments and not exceed the 100 lines limit.

It's just clear and reducing to a minimum with a ternary operator is not always the best solution.

src/components/dataTable/js/tableCellDirective.js
DevVersion: I think the reason is, that we need to append / prepend those icons dynamically and there is no template for the `table`.

Everything is completely up to the user.

src/components/dataTable/js/tableCellDirective.js
DevVersion: Click should be properly triggered on touch devices as well. Especially with the Fastclick solution we use ($mdGesture and soon HammerJS)

src/components/dataTable/js/tableCellDirective.js
DevVersion: It is a basic `jqLite` functionality, so it works for all Angular supported browsers.

src/components/dataTable/js/tableRowDirective.js
```diff
@@ -0,0 +1,326 @@
...
+    }
+  }
+};
+
+MdTableSelectController.prototype.toggleAutoSelect = function(state) {
+  state = this.$mdUtil.parseAttributeBoolean(state);
+
+  this.$element.toggleClass('md-auto-select', !!state);
+
+  if (state) {
```
EladBezalel: Ternary operator

src/components/dataTable/js/tableRowDirective.js
```diff
@@ -0,0 +1,326 @@
...
+    this.$element.on('click', this.$$toggleAutoSelectFn);
```
EladBezalel: Touch supprt?

src/components/dataTable/js/tableDirective.js
DevVersion: No that doesn't work, because `rows` is a native property and is not returning an Array, but a NodeList.

src/components/dataTable/js/tableDirective.js
DevVersion: As far as I know, the data table of Dan supports multiple body elements.

> http://stackoverflow.com/questions/3076708/can-we-have-multiple-tbody-in-same-table

src/components/dataTable/js/tableDirective.js
DevVersion: Doesn't make any sense to me either

src/components/dataTable/js/tableOrderDirective.js
DevVersion: I'm a fan of `$parse` - Should be good to use here :)

src/components/dataTable/js/tableRowDirective.js
DevVersion: _Same as said above_ 

Click should be properly triggered on touch devices as well. 

> Especially with the Fastclick solutions we use ($mdGesture and soon HammerJS)

src/components/dataTable/js/tableCellDirective.js
EladBezalel: i love `self` so much simpler that that

DevVersion: Hm I can understand your opinion. I guess it depends on the habits :smile: 

> https://gist.github.com/jashmenn/b306add36d3e6f0f6483

src/components/dataTable/icon.scss
daniel-nagy: I did this to separate the overrides for the `md-icon` directive, same with `md-checkbox`. I thought it was cleaner.

src/components/dataTable/icon.scss
daniel-nagy: This makes the icon inline with text. Fonts have extra room at the bottom for characters with tails (e.g 'qgyjp')

src/components/dataTable/js/tableCellDirective.js
daniel-nagy: I don't remember why but this directive didn't work unless the scope was isolated.

src/components/dataTable/js/tableCellDirective.js
daniel-nagy: I agree, `bind` is ugly 😛. It also creates a copy of the function,

> Returns a copy of the given function with the specified this value and initial arguments.

which makes me think it is not very inefficient to use it everywhere.

src/components/dataTable/js/tableCellDirective.js
daniel-nagy: You would have a ternary inside a ternary

```javascript
return this.isActive() ? this.orderCtrl.value.charAt(0) === '-' ? 'md-desc' : 'md-asc' : this.$mdUtil.parseAttributeBoolean(this.$attrs.mdDesc) ? 'md-desc' : 'md-asc';
```

src/components/dataTable/js/tableDirective.js
daniel-nagy: I was being clever to push the listener on the array and return it in one line. The last value of the `&&` operator will be returned;

```javascript
let foo = true && 'bar';

console.log(foo) // prints bar
```

src/components/dataTable/js/tableDirective.js
daniel-nagy: Could do that, this used to be used in the `mdSelect` directive but I don't see it being used anymore.

daniel-nagy:
Not sure if I made this [PR](https://github.com/daniel-nagy/material/pull/3) to the correct branch. It contains some of the edits suggested.

src/components/dataTable/js/tableCellDirective.js
ErinCoughlan: Or, since you can now use ES6, you can kill all bind/self. I want to change all of panel to do the same.

src/components/dataTable/icon.scss
EladBezalel: manner of taste, i'm not against and not in favor :)

src/components/dataTable/js/tableCellDirective.js
EladBezalel: Can we use es6 from now on?

src/components/dataTable/demoBasicUsage/style.css
```diff
@@ -0,0 +1,18 @@
+.container {
+  max-width: 100%;
+
+  > md-content {
```
EladBezalel: This is a css file, nesting won't work here :sweat_smile:

src/components/dataTable/js/tableCellDirective.js
daniel-nagy: es6 classes would clean things up nicely

DevVersion: ES6 would be definitely nice. But as long as we support IE11, we are not able to use it without a transpiler.

Also ES6 classes would be great ;) but it would not solve the context issue (still needing a bind, self or arrow fn)

src/components/dataTable/demoBasicUsage/style.css
daniel-nagy: lol, who writes css anymore?

src/components/dataTable/js/tableCellDirective.js
daniel-nagy: Typescript?

EladBezalel: Can we drop support for IE11

daniel-nagy: The application I am developing supports IE 10+, If you drop support for IE 11 I won't be using Angular Material anymore. Probably the same for a lot of other developers.

DevVersion: @daniel-nagy I'd love to have a transpiler instead ;) ES6 to ES5.

Even though TypeScript is awesome for Frameworks (defintion generation etc.) but Material 1.x takes advantage from the no-types ;)

daniel-nagy: @DevVersion You can still have freedom in typescript

```typescript
// TIL Github syntax highlighting supports typescript
let foo: any;
```

however, most of the time you are going to want to specify types for your variables, parameters, return values, etc. It improves the integrity of your code and the IntelliSense is awesome.

DevVersion: @daniel-nagy I always use TypeScript and I love it. Nevertheless most of the team don't want to use Typescript here, and in case of the freedom, it doesn't make any sense to TypeScript if we use non explicit types here.

The IntelliSense would be still better, but at the end exporting Defintions with types like `any` is a No-Go.

src/components/dataTable/js/tableDirective.js
```diff
@@ -0,0 +1,131 @@
...
+  this.changeListeners = [];
+
+  var self = this;
+
+  $attrs.$observe('multiple', function (multiple) {
+    self.multiple = $mdUtil.parseAttributeBoolean(multiple);
+
+    if (self.multiple) {
+      // This transforms our previous selected value in a array for the multiple selections.
+      self.selected = [].concat(self.selected || []);
```
DevVersion: I think we can get rid of that logic, because the `selected` should be always a single value here

```js
self.selected = [self.selected];
```

daniel-nagy: it could be `undefined`, in which case `self.selected := [undefined]` and `self.selected.length := 1` even though none of the items are selected.

talatbaig:
Hi,

Is there any chance this will be included in the next release?

PR #9210, opened 2016-08-02
fix(navbar): add theming support

### LINKED ISSUE #9137: mdNavBar : wrong theme used
**Actual Behavior**:
 -  What is the issue? When a theme is defined via md-theme, the md-nav-ink-bar for the md-nav-bar element does not use the given theme, but the default one
 -  What is the expected behavior? The md-nav-ink-bar element should use the accent color of the theme

 -  http://codepen.io/pyy/pen/QExbrK
 -  Here, the md-nav-ink-bar should be orange instead of the default pink

**Angular Versions**: *
 -  `Angular Version:` Any
 -  `Angular Material Version:` 1.1.0-rc.5

clshortfuse:
* Register theming service on navbar directive linking

Fixes #9137

### DIFF
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -92,7 +92,7 @@ angular.module('material.components.navBar', ['material.core'])
  *                                IMPLEMENTATION                             *
  *****************************************************************************/
 
-function MdNavBar($mdAria) {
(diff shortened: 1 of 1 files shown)


topherfangio:
@ThomasBurleson LGTM 👍

PR #9212, opened 2016-08-02
fix(menu): support dark theme

### LINKED ISSUE #9100: Assigning a color to backgroundPalette renders md-menu unreadable
**Actual Behavior**:
 -  `What is the issue? *` When we assign a color (say orange) to backgroundPalette, the text on md-menu is not barely readable.
 -  `What is the expected behavior?`The color should either be white or black based on the contrastDefaultColor property.

 -  `CodePen Demo which shows your issue:`http://codepen.io/vrathore42/pen/oLqdxp
 -  `Details:` I used the same colors used to define the blue palette and assigned it to customeBlue. I am assigning this pallet to the backgroundPalette property. Since contrastDefaultColor is "light"i would expect the md-menu  text to be white. But that's not the case. 

**Angular Versions**: *
 -  `Angular Version:`
 -  `Angular Material Version:`1.1.0-rc.5

**Additional Information**:
 -  `Browser Type: *` chrome
 -  `Browser Version: *` 51.0.2704.103 (64-bit)
 -  `OS: *` OS X 10.11.6 (El Capitan)
 -  `Stack Traces:` na

----

  Add more lines if needed.

clshortfuse:
Set background color to hue-1
Use contrast values for buttons, icons, and dividers

Fixes #9100


clshortfuse:
Blocked by PR #8872

EladBezalel:
@clshortfuse i'm not sure it's correct by the material design spec,
please take a look at #5744

clshortfuse:
@EladBezalel I see. Strange because that's not how Android's Implementation of Android works.

Still, there is an issue because it's not using the right contrast values causing issues if you use anything other than gray. I'll update it to be lighter but still support changing backgroundPalette.

clshortfuse:
Changed to hue-100 per spec. Now backgrounds properly render regardless of background palette.

Before: Before

After: After

EladBezalel:
@clshortfuse dropdown menus should always be white:
image

clshortfuse:
@EladBezalel 

I took a colorpicker against the dropdown menu and it's not white. It's `#FAFAFA` which is actually `grey-50`. I'm using `background-100` in the current screenshots above, so `background-50` should keep it all to spec. It'll partially tint based on background color if users change `backgroundPalette`. 

What it looks like now: 

Light:
screenshot 2016-09-08 at 2 14 55 pm

Dark:
screenshot 2016-09-08 at 2 12 24 pm

If we want to hardcode the colors (#FAFAFA), it'll create some complications if the user decides to change the foreground color and/or use a dark theme. Perhaps we can add support in mdTheme for referencing by palette name, for example `{{grey-50-background-contrast}}`. As far as I can tell, that would be the safest solution.

clshortfuse:
@EladBezalel can you take a look at it now that I switched to `background-50`. Or wait for the other PR, not sure how we should approach reviewing this.

EladBezalel:
So i'm really not sure what we should assume about background colors..
I'm also not sure if even you have different background color you'd like `background-50` and not `#fafafa`..

I'm afraid it may break some people and maybe we want to postpone it to 1.2 ? i'm not really sure, it's a hard one!

PR #9218, opened 2016-08-03
fix(): refactor rtl-prop mixin to add less CSS

### LINKED ISSUE #9217: Refactor rtl-prop mixin to avoid extra CSS
Currently the `rtl-prop` mixin adds 3 lines worth of CSS resets. We should refactor it, in order to reduce CSS bloat.

crisbeto:
* Refactors the rtl-prop mixin to include less CSS.
* Changes all the instances where it was used to incorporate the new syntax.

**Note:** There are a few instances where the mixin seems used redundantly. These aren't the objective of this PR.

Fixes #9217.

### DIFF
--- a/src/core/style/mixins.scss
+++ b/src/core/style/mixins.scss
@@ -125,15 +125,10 @@
   }
 }
 
-@mixin rtl-prop($ltr-prop, $rtl-prop, $value) {
+@mixin rtl-prop($ltr-prop, $rtl-prop, $value, $reset-value) {
   #{$ltr-prop}: $value;
   [dir=rtl] & {
-    // fallback for auto
-    #{$ltr-prop}: 0;
-    // fallback for IE
-    #{$ltr-prop}: auto;
-    #{$ltr-prop}: initial;
-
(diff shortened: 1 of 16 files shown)


src/core/style/mixins.scss
```diff
@@ -125,15 +125,10 @@
   }
 }
 
-@mixin rtl-prop($ltr-prop, $rtl-prop, $value) {
+@mixin rtl-prop($ltr-prop, $rtl-prop, $value, $reset-value) {
```
clshortfuse: I would say you should add a `null` default: 
`@mixin rtl-prop($ltr-prop, $rtl-prop, $value, $reset-value: null) {`

That was, we can make some segments with even less CSS. 

http://www.sassmeister.com/gist/b87f3d64603d5bceaa29534ffc585457

crisbeto: The point of the PR isn't to reduce the CSS as much as possible, but rather to remove the redundancies while staying as close to the old functionality as possible.

PR #9220, opened 2016-08-03
feat(layouts): add @mixin for responsive support for rows

### LINKED ISSUE #9112: md-divider: Responsive layout support is not properly supported

clshortfuse:
Creates @mixin that only triggers when in row configuration.
- Dividers now properly display in dynamic layout directions.
- Radio Buttons bottom margins are now properly removed when
  in row configuration.
- Input, Select, Radio Buttons and Checkboxes automatically
  add 16px horizontal margin when in a row and not last item.

Previous implementations could not properly detect row configuration
where a layout would change direction based on media breakpoints.

Fixes #9112

### DIFF
--- a/src/core/style/mixins.scss
+++ b/src/core/style/mixins.scss
@@ -337,3 +337,56 @@
     @include rtl($prop, $ltr-dense, $rtl-dense);
   }
 }
+
+// Only use when in row layout
+@mixin when-layout-row($element) {
+  @media (max-width: $layout-breakpoint-xs - 1) {
+    .layout-row:not(.layout-xs-column),
+    .layout-xs-row {
+      & > #{$element} { @content; }
+    }
+  }
+  @media (min-width: $layout-breakpoint-xs) and (max-width: $layout-breakpoint-sm - 1) {
+    .layout-row:not(.layout-gt-xs-column),
+    .layout-gt-xs-row,
+    .layout-sm-row {
+      &:not(.layout-sm-column) > #{$element} { @content; }
+    }
+  }
+  @media (min-width: $layout-breakpoint-sm) and (max-width: $layout-breakpoint-md - 1) {
+    .layout-row:not(.layout-gt-xs-column):not(.layout-gt-sm-column),
+    .layout-gt-xs-row:not(.layout-gt-sm-column),
+    .layout-gt-sm-row,
+    .layout-md-row {
+      &:not(.layout-md-column) > #{$element} { @content; }
+    }
+  }
+  @media (min-width: $layout-breakpoint-md) and (max-width: $layout-breakpoint-lg - 1) {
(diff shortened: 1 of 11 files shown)


clshortfuse:
I recreated this from an old PR #9112 

Now the resultant CSS is actually smaller than master:

 52264 (this PR) vs 52805 (master)

clshortfuse:
Add `High Density` label since this is a multi-component change that reduces horizontal margins. (in other words, denser layouts)

clshortfuse:
@Splaktar Rebased.

src/core/style/mixins.scss
```diff
@@ -337,3 +337,56 @@
...
+    .layout-xl-row {
+      &:not(.layout-gt-lg-column):not(.layout-xl-column) > #{$element} { @content; }
+    }
+  }
+}
+
+// Auto insert object margin
+@mixin auto-horizontal-margin($selector) {
+  @include when-layout-row($selector) {
+    &:not(last-child) {
```
Splaktar, line 388: I believe that this should be `:last-child` to actually work properly.

That said, in my testing, that results in this:
Screen Shot 2020-07-27 at 15 00 05

Using `:first-child` seems to give us what we want:
Screen Shot 2020-07-27 at 15 00 34

Splaktar:
Continued this work in PR https://github.com/angular/material/pull/11981.

src/core/style/mixins.scss
clshortfuse: I think it should be `:last-child` as you noted, or else it would target against `<last-child>` tag name, which would be wrong. I would check against a row with 3 items to ensure the auto margins work. It might look right with `:first-child` because they're only 2 items.

clshortfuse: The rule, in theory, should pad `$default-horizontal-margin` at the start (left on LTR). Left padding should appear on every item, except the first one, so `:not(:first-child)` makes sense. 

Maybe I did a last minute change and had it apply margin at the end originally, and failed to adjust. So yeah, `:not(:first-child)` is good.

PR #9224, opened 2016-08-03
fix(select, input): asterisk should be always red.

DevVersion:
Commit 024e9798b9029ebf1f30df5bed4b2198f96172a6 added support for asterisks for the `md-select`, but also removed the red styling for the asterisk.
As per specs, the asterisk should be always red, except when the input is blurred and not invalid.

This commit makes the asterisk always red and makes the `md-select` asterisk consistent to the inputs.

image

image


clshortfuse:
I think you mean always uses `warn` palette (not necessarily red)

topherfangio:
LGTM 👍

PR #9229, opened 2016-08-04
fix(toolbar): Title text should allow ellipsis.

### LINKED ISSUE #9026: md-toolbar text overflow should be hidden and replaced by ellipsis
**Actual Behavior**:
-  `What is the issue? *` md-toolbar text goes 'multi-line' when it's too long for the view
-  `What is the expected behavior?` Overflow should be hidden and replaced by ellipsis

-  `Details:` Visit: https://material.angularjs.org/latest/demo/toolbar and shrink window to ~400px wide. See here for brief discussion: http://stackoverflow.com/questions/27600115/in-angularjs-material-design-how-to-hide-overflowed-text-in-the-md-toolbar 

**Angular Versions**: *
-  `Angular Version:` All

**Additional Information**:
-  `Browser Type: *` All
-  `Browser Version: *` All
-  `OS: *` All
-  `Stack Traces:` N/A

---

>   Add more lines if needed.

topherfangio:
Currently, the title text of a toolbar wraps around on small screens causing the toolbar to become larger than expected.
- Provide a new `md-truncate` component that applies the appropriate styles
  for using CSS-based text-overflow with an ellipsis.
  
  This component is designed to work nicely with the Toolbar component,
  but is also reusable and can be used in any other component.
- Add docs/demos for the new component.
- Update Toolbar demos to use new `md-truncate` component and show proper
  usage.

Additionally, add the appropriate CSS docs in a new CSS & Styles section which includes a new doc-only directive (`docs-css-api-table`) that creates a clean way to view documentation regarding available CSS classes.

Fixes #9026.

---

_**Edit:** Add screenshot of new docs section._

screen shot 2016-08-03 at 6 56 33 pm

### DIFF
--- a/docs/app/css/style.css
+++ b/docs/app/css/style.css
@@ -872,10 +872,25 @@ table.attributes  tr  td:last-child {
   padding-left: 10px;
 }
 
-table.md-api-table  tr  td:first-child {
+table.md-api-table:not(.md-css-table)  tr  td:first-child {
   font-weight: bold;
 }
 
+table.md-css-table .md-css-selector {
+  display: block;
+  padding: 8px 16px;
+
+  /* Offset the padding of the <td> element */
+  margin: -12px -16px 12px -16px;
+
+  background-color: #0C2238;
+  color: #9ccc65;
+}
+
+/* Fix some odd bottom margin */
+table.md-css-table td p {
+  margin: 0.8em 0;
+}
 
 .layout_note {
   font-size: 0.9em;
--- a/docs/app/js/css-api-table.js
+++ b/docs/app/js/css-api-table.js
@@ -1,0 +1,51 @@
+(function() {
+  angular.module('docsApp')
+    .directive('docsCssApiTable', DocsCssApiTableDirective)
+    .directive('docsCssSelector', DocsCssSelectorDirective);
+
+  function DocsCssApiTableDirective() {
+    return {
+      restrict: 'E',
+      transclude: true,
+
+      bindToController: true,
+      controller: function() {},
+      controllerAs: '$ctrl',
+
+      scope: {},
+
+      template:
+      '<table class="md-api-table md-css-table">' +
+      '  <thead>' +
+      '    <tr><th>Available Selectors</th></tr>' +
+      '  </thead>' +
+      '  <tbody ng-transclude>' +
+      '  </tbody>' +
+      '</table>'
+    }
+  }
+
+  function DocsCssSelectorDirective() {
+    return {
+      restrict: 'E',
+      transclude: true,
+      replace: true,
+
+      bindToController: true,
+      controller: function() {},
+      controllerAs: '$ctrl',
+
+      scope: {
+        code: '@'
+      },
+
+      template:
+      '<tr>' +
+      '  <td>' +
+      '    <code class="md-css-selector">{{$ctrl.code}}</code>' +
+      '    <span ng-transclude></span>' +
+      '  </td>' +
+      '</tr>'
+    }
+  }
+})();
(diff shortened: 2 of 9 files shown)


topherfangio:
@ThomasBurleson @EladBezalel @crisbeto @DevVersion Can you guys review and see if you like the color/styling of the new "CSS & Styles" section of the Toolbar docs?

Also, standard review of my proposed solution to the actual problem 😄

ThomasBurleson:
@topherfangio - this looks very cool. Like your use of the Component API in the `docs-css-api-table` implementation.

ThomasBurleson:
@topherfangio  - lgtm

docs/app/js/css-api-table.js
```diff
@@ -0,0 +1,59 @@
...
+      '  <thead>' +
+      '    <tr><th>Available Selectors</th></tr>' +
+      '  </thead>' +
+      '  <tbody ng-transclude>' +
+      '  </tbody>' +
+      '</table>'
+    }
+  }
+
+  function DocsCssApiTableController() {
```
DevVersion: Nit: Why are there controllers if they aren't used? Same for the selector directive.

DevVersion:
LGTM aside from that minor comment :+1:

docs/app/js/css-api-table.js
ThomasBurleson: can you simply use : 

```js
controller : function() {  },
```

topherfangio:
@ThomasBurleson @DevVersion Waiting for Travis to succeed, but I pushed the requested updates, and fixed a typo 😃

src/components/toolbar/toolbar.scss
```diff
@@ -107,6 +107,19 @@ md-toolbar {
     margin: inherit;
   }
 
+  .md-toolbar-title {
+    display: flex;
+    flex: 1;
+    width: 0;
+
+    span {
```
EladBezalel: Why not use `&:first-child` and make it more generic?

docs/app/css/style.css
```diff
@@ -909,10 +909,25 @@ table.attributes  tr  td:last-child {
...
+  /* Offset the padding of the <td> element */
+  margin: -12px -16px 12px -16px;
+
+  background-color: #0C2238;
+  color: #9ccc65;
+}
+
+/* Fix some odd bottom margin */
+table.md-css-table td p {
+  margin: 0.8em 0 0.8em;
```
crisbeto: Small nitpick: this is the same as `margin: 0.8em 0`.

ThomasBurleson:
@topherfangio - can you fix the failing tests ?

topherfangio:
@ThomasBurleson Um...I "fixed" the tests...actually, I didn't do anything. Someone must have rerun the failed build and it passed?

I ran the tests on my Windows box twice. The first time a `$mdBottomsheet` test failed; the second time, they all passed. So I think these are erroneous failures.

src/components/toolbar/toolbar.scss
topherfangio: Since the only formatting this applies is the overflow, I wanted something that shouldn't apply any additional formatting.

I guess I'm afraid that if we switch it to `:first-child`, people might try to use a `<div>` or something else strange and it will mess up.

I'm not opposed to switching it, and I honestly can't put a finger on why I feel this way, but using the `<span>` just **_feels_** like a better solution to me :confused:

Thoughts?

topherfangio:
Talked with Elad, we think making this an actual directive of `<md-toolbar-title>` makes more sense instead of forcing people to include a class and then add a `<span>` tag inside.

@ThomasBurleson Unless you disagree, I'll get this done tomorrow.

topherfangio:
@ThomasBurleson @EladBezalel Make a more generic `<md-title>` component which handles this case for any component (not just Toolbar).

Do you like the name `md-title`, or should we perhaps name it `md-ellipsis` to be more clear?

EladBezalel:
I said that maybe having an ellipsis directive could be nice but in case we don't want to create more complexibility than I'd go with `<md-toolbar-title>`

ThomasBurleson:
`md-truncate`

topherfangio:
Updated to `md-truncate`. Ping @ThomasBurleson for one more review.

ThomasBurleson:
@topherfangio - lgtm

PR #9231, opened 2016-08-04
feat(mdPanel): Wrapper and Panel elements referenced in the MdPanelRef

### LINKED ISSUE #9109: feat(mdPanel): Wrapper and Panel elements referenced in the MdPanelRef
**Actual Behavior**:
 -  `What is the issue? *`
  - There is not a reference to the Wrapper and Panel elements within the returned MdPanelRef.
  - A jqLite reference of the Wrapper and Panel elements should be returned within the MdPanelRef object. Then several functions, i.e. addClass, removeClass, etc., could be run on the elements without having to have these type functions repeated within the MdPanel API.

 -  `Details:` N/A

**Angular Versions**: *
 -  `Angular Version:` Latest

**Additional Information**:
 -  `Browser Type: *` All
 -  `OS: *` All
 -  `Stack Traces:` N/A

bradrich:
A JQLite reference of the Container/Wrapper and Panel elements have been added to the MdPanelRef object as public properties. It is now possible to use the standard `addClass`, `removeClass`, `toggleClass`, etc methods on those referenced elements.

Fixes #9109

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -145,12 +145,12 @@ angular
  *     when the open() action is finished.
  *   - `onRemoving` - `{function=}`: Callback function used to announce the
  *     close/hide() action is starting.
- *   - `onDomRemoved` - `{function=}`: Callback function used to announce when the
- *     panel is removed from the DOM.
+ *   - `onDomRemoved` - `{function=}`: Callback function used to announce when
+ *     the panel is removed from the DOM.
  *   - `origin` - `{(string|!angular.JQLite|!Element)=}`: The element to
- *     focus on when the panel closes. This is commonly the element which triggered
- *     the opening of the panel. If you do not use `origin`, you need to control
- *     the focus manually.
+ *     focus on when the panel closes. This is commonly the element which
+ *     triggered the opening of the panel. If you do not use `origin`, you need
+ *     to control the focus manually.
  *
  * @returns {MdPanelRef} panelRef
  */
@@ -214,6 +214,12 @@ angular
  *     create.
  *   - `isAttached` - `{boolean}`: Whether the panel is attached to the DOM.
  *     Visibility to the user does not factor into isAttached.
+ *   - `panelContainer` - `{angular.JQLite}`: The wrapper element containing the
+ *     panel. This property is added in order to have access to the `addClass`,
+ *     `removeClass`, `toggleClass`, etc methods.
+ *   - `panelEl` - `{angular.JQLite}`: The panel element. This property is added
+ *     in order to have access to the `addClass`, `removeClass`, `toggleClass`,
+ *     etc methods.
  */
 
 /**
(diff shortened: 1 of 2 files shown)


EladBezalel:
LGTM

src/components/panel/panel.js
```diff
@@ -214,6 +214,12 @@ angular
  *     create.
  *   - `isAttached` - `{boolean}`: Whether the panel is attached to the DOM.
  *     Visibility to the user does not factor into isAttached.
+ *   - `panelContainer` - `{angular.JQLite}`: The wrapper element containing the
```
bradrich: @ErinCoughlan Not sure if you saw these earlier, but just wanted to make sure.

bradrich:
@ErinCoughlan Are you thinking that I should add back in the `addClass`, `removeClass` and `toggleClass` methods, and make them have log errors that direct developers to the new methods? That is a good idea. I will wait on @ThomasBurleson before making that call, unless you give me the go ahead.

EladBezalel:
@ErinCoughlan that's a good idea.. that way we can make sure it's not breaking anyone

ErinCoughlan:
@bradrich @EladBezalel After some thought, I think we should follow the real deprecation strategy here. We don't want to break people and we want to give them some time to migrate to the new methods.

1. Keep the methods functional so the apps won't be broken if they are using them, but mark them `@deprecated`.
2. Throw a warning, not error, that they are deprecated and give the new line to switch to. Probably also have a timeline for when we will remove them. @ThomasBurleson I don't know the normal deprecation timeline for material, but this should match it.
3. After the deprecation period, upgrade the warnings to errors, and make the functions not work.
4. After Y time, remove the methods entirely. Ideally, you would do this Y time after the errors are added. The reason not to remove the methods at first is because the error message says how to migrate to the new versions, so people won't have to think of how to fix their apps.

bradrich:
@ErinCoughlan I agree with this and will move forward with the changes.

bradrich:
@ErinCoughlan Please take a look at the newest commit that contains the deprecation notices and warnings.

src/components/panel/panel.js
```diff
@@ -1071,12 +1089,22 @@ MdPanelRef.prototype.hide = function() {
...
  * @param {string} newClass Class to be added.
  * @param {boolean} toElement Whether or not to add the class to the panel
  *    element instead of the container.
  */
 MdPanelRef.prototype.addClass = function(newClass, toElement) {
+  this._$log.warn(
+      'The addClass method is in the process of being deprecated. ' +
+      'Full deprecation is scheduled for the Angular Material 1.2 release. ' +
+      'To achieve the same results, use the panelContainer or panelEl ' +
+      'JQLite elements that are referenced in MdPanelRef.');
```
ErinCoughlan: I would almost even write the line to replace this one to make it even easier. Not strictly required though.

ErinCoughlan:
LGTM

EladBezalel:
@bradrich can you open an issue at 1.2 milestone so we can keep track of this deprecation ?

bradrich:
@EladBezalel #9310.

PR #9236, opened 2016-08-04
fix(build): switch to cssnano for CSS optimizations

### LINKED ISSUE #9225: css: Dedup CSS using postCSS()
DatePicker and MenuBar has some duplicate CSS properties.
Let's update the build processes and use postCSS rules to cleanup... (similar to ngM2 builds).

crisbeto:
Replaces the deprecated `gulp-minify-css` with `gulp-cssnano`.

Fixes #9225.

CC @topherfangio, @DevVersion, @EladBezalel

Some notes after comparing the output to the one from our previous build:

* The builds take a few seconds longer, but the minified file size is reduced by about 6kb.
* It does shuffle around some styles, but it seems to be playing it pretty safe so I doubt that it'll break anything.
* Apparently the old minifier that we were using was converting instances of `outline: none` to `outline: 0`.  Now the value is preserved properly, but that'll be a minor difference in the output, compared to before. We could go through and convert the values if needed.


topherfangio:
@crisbeto Can you include some of the notes you mentioned during our meeting so we have a bit more info? Particularly interested in your comparisons.

Otherwise, LGTM 👍

crisbeto:
Updated.

DevVersion:
LGTM as well.

jelbourn:
LGTM!

PR #9237, opened 2016-08-04
fix(select): Allow 0 as default value.

### LINKED ISSUE #9232: md-select can't be valid when default value is 0
**Actual Behavior**:
 -  `What is the issue? *`md-select can't be valid when default value is 0
 -  `What is the expected behavior?`md-select should be valid when set default value.

 -  `CodePen Demo which shows your issue:`http://codepen.io/wcmboy/pen/wWYBGZ
 -  `Details:`You can see  the favoriteColor1 is 0 and the favoriteColor2 is 1.The favoriteColor1 is invalid and favoriteColor2 is valid.

**Angular Versions**: *
 -  `Angular Version:`1.5.5
 -  `Angular Material Version:`v1.1.0-rc.5-master-8dcdf82

**Additional Information**:
 -  `Browser Type: *`chome
 -  `Browser Version: *` 52.0.2743.82 m
 -  `OS: *`windows 10

topherfangio:
A recent change introduced a bug which no longer allowed `0` as a valid `ng-value`.

Fix by surrounding associated `if` code with `angular.isDefined(...)` to ensure `0` is treated as a valid value.

Fixes #9232.

_**Note:** It looks like lots of changes to the `.spec.js` file because I moved some of the tests to a more relevant location. In reality, I only added 1 test and moved the others. See my line comment for the actual change._

### DIFF
--- a/src/components/select/select.spec.js
+++ b/src/components/select/select.spec.js
@@ -30,207 +30,256 @@ describe('<md-select>', function() {
     backdrops.remove();
   }));
 
+  describe('basic functionality', function() {
+    it('should have `._md` class indicator', function() {
+      var element = setupSelect('ng-model="val"').find('md-select-menu');
+      expect(element.hasClass('_md')).toBe(true);
+    });
 
-  it('should have `._md` class indicator', function() {
-    var element = setupSelect('ng-model="val"').find('md-select-menu');
-    expect(element.hasClass('_md')).toBe(true);
+    it('should preserve tabindex', function() {
+      var select = setupSelect('tabindex="2" ng-model="val"').find('md-select');
+      expect(select.attr('tabindex')).toBe('2');
+    });
+
+    it('should set a tabindex if the element does not have one', function() {
+      var select = setupSelect('ng-model="val"').find('md-select');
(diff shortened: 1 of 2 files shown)


src/components/select/select.spec.js
```diff
@@ -30,207 +30,256 @@ describe('<md-select>', function() {
...
+      $rootScope.val = 1;
+      var select = setupSelect('ng-model="val"', [1, 2, 3]).find('md-select');
+      var label = select.find('md-select-value')[0];
+      expect(label.textContent).toBe('1');
+      $rootScope.val = 4;
+      $rootScope.$$values = [4, 5, 6];
+      $rootScope.$digest();
+      expect(label.textContent).toBe('4');
+    }));
+
```
topherfangio: Below (like 268) is the added test; everything else already existed; just moved it around a bit for clarity.

topherfangio:
@ThomasBurleson I have marked this as 1.1.0 because I'm pretty sure we broke it in the last release and it seems like a significant enough break that we want to get it fixed before the final push.

DevVersion:
@ThomasBurleson - That looks perfect. I also think it would be good to get that into `1.1.0`.

PR #9240, opened 2016-08-04
chore(select): Add note regarding recent breaking change.

topherfangio:
BREAKING CHANGE

The `<md-select>`'s, `<md-option>` component now acts more like
the default `<option>` attribute of a standard HTML `<select>` by
treating empty `value` and `ng-value` attributes as a special
"reset" case that allows the input to return to the default state
(i.e. the floating label returns to the placeholder position).

If you use the `value` or `ng-value` attributes with no value
as follows **and** expect it to be a valid option,

```html
<md-option value="">All options</md-option>
```

you will need to update your option to have a value like one
of the following:

```html
<md-option value="null">All options</md-option>
<md-option value="undefined">All options</md-option>
<md-option ng-value="''">All options</md-option>
```

Fixes #9718.


topherfangio:
Note that this is a commit-message only modification. There is no associated code change.

ThomasBurleson:
@jelbourn  FYI

ThomasBurleson:
@topherfangio - manually added to the CHANGELOG for v1.1.0. Thx.

PR #9244, opened 2016-08-05
fix(showHide): don't call getComputedStyle on comment node

### LINKED ISSUE #9243: ngShow and ngHide on directives that use comment placeholders will error
To reproduce:
Click the "Toggle" button
http://codepen.io/jelbourn/pen/mEzrOP?editors=1010#0

The problem:
Angular sometimes uses comment elements as placeholders so that a real element doesn't have to be rendered. This means the `element` given to a `link` function will be of type `Comment`. 

[Material's decorating of `ngShow` and `ngHide` uses `getComputedStyle` on the `element`](https://github.com/angular/material/blob/master/src/components/showHide/showHide.js#L25), which fails because it is a comment.

cc @topherfangio as the last one to touch these lines

crisbeto:
Fixes the ngShow and ngHide directive calling getComputedStyle on comment nodes.

Fixes #9243.

### DIFF
--- a/src/components/showHide/showHide.js
+++ b/src/components/showHide/showHide.js
@@ -14,7 +14,7 @@ angular.module('material.components.showHide', [
 
 
 function createDirective(name, targetValue) {
-  return ['$mdUtil', function($mdUtil) {
+  return ['$mdUtil', '$window', function($mdUtil, $window) {
     return {
       restrict: 'A',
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
Very nice!

ThomasBurleson:
@jelbourn - fixed.

PR #9250, opened 2016-08-05
fix(icon): remove trustAs calls in favor of implicit trust conditions

jelbourn:


jelbourn:
@topherfangio or @EladBezalel take a look?

EladBezalel:
Referring your message:
>The changes for mdIcon with `$sce` aren't working out due to security change that was made with angular 1.5
>this change would mean that for people who want to load svg icons, they have to do one of the following:
>* Request icons from the same-origin 
>* Pre-load icons into the `$templateCache`
>* Add the origins for the icon requests to the `$sce` whitelist

So this is actually a breaking change, can't we polyfill the 1.5 fix?

jelbourn:
@EladBezalel this is a breaking change from previous versions. The allowed sources for SVG icons are now more strict, which enforces a stronger security model.

EladBezalel:
LGTM

topherfangio:
LGTM 👍

PR #9251, opened 2016-08-05
fix(autocomplete): fix a couple of js errors and log a warning if the display value isn't a string

### LINKED ISSUE #9242: Autocomplete: useless error thrown when `md-item-text` is missing
To reproduce:
Select a value and then hit the clear button underneath
http://codepen.io/jelbourn/pen/mEzrbJ?editors=1000#anon-login

The issue is this line:
https://github.com/angular/material/blob/master/src/components/autocomplete/js/autocompleteController.js#L338

When `md-item-text` is missing, this ends up being an object instead of a string, so `toLowerCase` is not a function.

If `md-item-text` is _required_, it should throw a more informative error.

If `md-item-text` is _optional_, autocomplete should handle this case gracefully.

crisbeto:
* Fixes a couple of JS errors in the autocomplete that were caused by assumptions that the input would always be a string.
* Adds a warning for cases where the display value doesn't evaluate to a string.
* Cleans up a few redundant expressions.

**Note:** There's still a JS error when clearing the value, but it should be fixed with #9221.

Fixes #9242.

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -35,7 +35,7 @@ describe('<md-autocomplete>', function() {
 
       scope.asyncMatch = function(term) {
         return $timeout(function() {
-          return scope.match(term)
+          return scope.match(term);
         }, 1000);
       };
 
@@ -702,7 +702,7 @@ describe('<md-autocomplete>', function() {
         return {display: item};
       });
       var scope = createScope(myItems);
-      
+
       var template = '\
           <md-autocomplete\
               md-selected-item="selectedItem"\
@@ -740,7 +740,7 @@ describe('<md-autocomplete>', function() {
 
       expect(scope.searchText).toBe('foo ');
       expect(scope.selectedItem).toBe(scope.match(scope.searchText)[0]);
-      
+
       ctrl.clear();
       $timeout.flush();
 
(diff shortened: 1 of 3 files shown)


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -531,7 +531,13 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    * @returns {*}
    */
   function getDisplayValue (item) {
-    return $q.when(getItemText(item) || item);
+    var itemText = getItemText(item) || item;
+
+    if (itemText && !angular.isString(itemText)) {
+      $log.warn('Could not resolve display value to a string. Please check the `md-item-text` attribute.');
```
DevVersion: Can we please prefix that (probably like this) ?

```
md-autocomplete: The resolved text for an item is not a string. Please check the `md-item-text` attribute.`
```

src/components/autocomplete/autocomplete.spec.js
```diff
@@ -1125,6 +1125,33 @@ describe('<md-autocomplete>', function() {
       element.remove();
     }));
 
+    it('should log a warning if the display text does not evaluate to a string',
+      inject(function($log) {
+        $log.reset();
```
DevVersion: Why do we need a log reset here? Shouldn't be necessary since your creating spy after it.

crisbeto: Mostly for insurance, in case some logs got left over from other tests.

src/components/autocomplete/js/autocompleteController.js
crisbeto: Good idea, I'll put it in.

crisbeto: Good idea, I'll put it in.

src/components/autocomplete/autocomplete.spec.js
DevVersion: But that shouldn't trigger a new `$log.warn` call, since those were already executed.

crisbeto: Ah I see what you mean, I'll remove it.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -531,7 +531,14 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    * @returns {*}
    */
   function getDisplayValue (item) {
-    return $q.when(getItemText(item) || item);
+    var itemText = getItemText(item) || item;
+
+    if (itemText && !angular.isString(itemText)) {
```
DevVersion: I'm not really sure about that check, because `$scope.itemText` can also return a promise as a value.
> http://codepen.io/DevVersion/pen/KrGxaQ (Simple example)

I think it would be better to chain into the returned promise and add the check there.

crisbeto: Thanks, I didn't know that was a use case.

DevVersion:
LGTM.

PR #5972, opened 2015-11-28
feat(sidenav): observe component id change

EladBezalel:
Sidenav was registering the first interpolated component id.
Now observing the component id attribute changes, removing the old registration and registering with the new interpolated value.

fixes #4987 and #3171

PR #9255, opened 2016-08-06
fix(sidenav): allow for data bindings in md-component-id

### LINKED ISSUE #9052: mdSidenav renders before scope variables are populated
**Actual Behavior**:
 -  `What is the issue? *`

The html below is loaded by a component through the property templateUrl, and the mdSidenav directive is rendered before {{ ::$ctrl.id }} gets evaluated. So the md-component-id becomes literally the string "{{ ::$ctrl.id }}"

&lt;md-sidenav
    class="md-sidenav-left md-whiteframe-2dp"
    md-component-id="{{ ::$ctrl.id }}"
    md-is-locked-open="$mdMedia('gt-sm')"&gt;
        &lt;md-toolbar&gt;
            &lt;h1 class="text-center"&gt;{{ $ctrl.header }}&lt;/h1&gt;
        &lt;/md-toolbar&gt;
        &lt;md-content&gt;
            &lt;menu-dropdown navigation="$ctrl.navigation"&gt;&lt;/menu-dropdown&gt;
        &lt;/md-content&gt;
&lt;/md-sidenav&gt;

**Angular Versions**: *
 -  `Angular Version:` 1.5.5
 -  `Angular Material Version:` 1.0.9

crisbeto:
Fixes the sidenav not evaluating expressions inside of the `md-component-id` attribute.

Fixes #9052.

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -249,8 +249,7 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
     },
     controller: '$mdSidenavController',
     compile: function(element) {
-      element.addClass('md-closed');
-      element.attr('tabIndex', '-1');
+      element.addClass('md-closed').attr('tabIndex', '-1');
       return postLink;
     }
   };
@@ -481,9 +480,8 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
  * @ngdoc controller
  * @name SidenavController
  * @module material.components.sidenav
- *
  */
-function SidenavController($scope, $element, $attrs, $mdComponentRegistry, $q) {
+function SidenavController($scope, $attrs, $mdComponentRegistry, $q, $interpolate) {
 
   var self = this;
 
@@ -505,5 +503,21 @@ function SidenavController($scope, $element, $attrs, $mdComponentRegistry, $q) {
   self.toggle = function() { return self.$toggleOpen( !$scope.isOpen );  };
   self.$toggleOpen = function(value) { return $q.when($scope.isOpen = value); };
 
-  self.destroy = $mdComponentRegistry.register(self, $attrs.mdComponentId);
+  // Evaluate the component id.
+  var rawId = $attrs.mdComponentId;
+  var hasDataBinding = rawId && rawId.indexOf($interpolate.startSymbol()) > -1;
+  var componentId = hasDataBinding ? $interpolate(rawId)($scope.$parent) : rawId;
+
+  // Register the component.
+  self.destroy = $mdComponentRegistry.register(self, componentId);
+
+  // Watch and update the component, if the id has changed.
+  if (hasDataBinding) {
+    $attrs.$observe('mdComponentId', function(id) {
+      if (id && id !== self.$$mdHandle) {
+        self.destroy(); // `destroy` only deregisters the old component id so we can add the new one.
+        self.destroy = $mdComponentRegistry.register(self, id);
+      }
+    });
+  }
 }
--- a/src/components/sidenav/sidenav.spec.js
+++ b/src/components/sidenav/sidenav.spec.js
@@ -1,14 +1,14 @@
 describe('mdSidenav', function() {
   beforeEach(module('material.components.sidenav'));
 
-  function setup(attrs) {
+  function setup(attrs, skipInitialDigest) {
     var el;
     inject(function($compile, $rootScope) {
       var parent = angular.element('<div>');
       el = angular.element('<md-sidenav ' + (attrs || '') + '>');
       parent.append(el);
       $compile(parent)($rootScope);
-      $rootScope.$apply();
+      !skipInitialDigest && $rootScope.$apply();
     });
     return el;
   }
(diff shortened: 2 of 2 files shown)


DevVersion:
I think @EladBezalel had something similar, including observation of the attribute. https://github.com/angular/material/pull/5972

crisbeto:
Seems like it didn't end up getting merged. @EladBezalel is it ok if I incorporate the changes here?

crisbeto:
Updated, although the approach from #5972 doesn't work for cases where the instance is being looked up in a controller on load.

src/components/sidenav/sidenav.js
```diff
@@ -505,5 +503,19 @@ function SidenavController($scope, $element, $attrs, $mdComponentRegistry, $q) {
...
+  // Evaluate the component id.
+  var rawId = $attrs.mdComponentId;
+  var componentId = rawId && rawId.indexOf($interpolate.startSymbol()) > -1 ?
+    $interpolate(rawId)($scope.$parent) : rawId;
+
+  // Register the component.
+  self.destroy = $mdComponentRegistry.register(self, componentId);
+
+  // Watch and update the component, if the id has changed.
+  $attrs.$observe('mdComponentId', function(id) {
```
ThomasBurleson: Should we only do this if we originally had interpolation symbols ?

src/components/sidenav/sidenav.js
```diff
@@ -505,5 +503,19 @@ function SidenavController($scope, $element, $attrs, $mdComponentRegistry, $q) {
...
+    if (id && id !== self.$$mdHandle) {
+      self.destroy();
+      self.destroy = $mdComponentRegistry.register(self, id);
```
ThomasBurleson: What is ^ this doing ?

src/components/sidenav/sidenav.js
crisbeto: Hmm yeah, good point. I'll change it.

src/components/sidenav/sidenav.js
crisbeto: The `register` method returns a method that can be used to de-register the component. Here we de-register the instance, because it was using the old id, and we register it again with the new one.

ThomasBurleson: Ok... just add a sidecomment: `// $ctrol.destroy() simply de-registers the component`

src/components/sidenav/sidenav.js
```diff
@@ -505,5 +503,21 @@ function SidenavController($scope, $element, $attrs, $mdComponentRegistry, $q) {
   self.toggle = function() { return self.$toggleOpen( !$scope.isOpen );  };
   self.$toggleOpen = function(value) { return $q.when($scope.isOpen = value); };
 
-  self.destroy = $mdComponentRegistry.register(self, $attrs.mdComponentId);
+  // Evaluate the component id.
+  var rawId = $attrs.mdComponentId;
+  var hasDataBinding = rawId && rawId.indexOf($interpolate.startSymbol()) > -1;
```
EladBezalel: Why isn't that check happens on the directive?

src/components/sidenav/sidenav.js
```diff
@@ -505,5 +503,21 @@ function SidenavController($scope, $element, $attrs, $mdComponentRegistry, $q) {
...
+  var componentId = hasDataBinding ? $interpolate(rawId)($scope.$parent) : rawId;
+
+  // Register the component.
+  self.destroy = $mdComponentRegistry.register(self, componentId);
+
+  // Watch and update the component, if the id has changed.
+  if (hasDataBinding) {
+    $attrs.$observe('mdComponentId', function(id) {
+      if (id && id !== self.$$mdHandle) {
```
EladBezalel: $$mdHandle?

src/components/sidenav/sidenav.js
crisbeto: It's not a directive, it's just the attribute that you use to pass the id. I'm not sure it make sense for it to be a directive since it's just a string.

src/components/sidenav/sidenav.js
crisbeto: That's the id that is assigned when an element is registered with the `$mdComponentRegistry`. (https://github.com/angular/material/blob/master/src/core/services/registry/componentRegistry.js#L64)

src/components/sidenav/sidenav.js
EladBezalel: It can be an attribute directive that checks if it should watch for changes or not (like mdColors does)

crisbeto: It still seems redundant and won't be immefiately obvious where the id is coming from. In this case it's right next to the initialization.

src/components/sidenav/sidenav.js
ThomasBurleson: why not call it `self.componentId` ?

crisbeto: `self.componentId` isn't actually being declared, it's only in the `$attrs`, `$$mdHandle` is the one that keeps track of the current id.

src/components/sidenav/sidenav.js
crisbeto: As per the discussion on Slack, let's keep this in here for now and do a more generic solution if the need for one comes up.

Issue #8475, opened 2016-05-16
Allow custom z-index values for elements


ErinCoughlan:
Not all apps use only Angular Material components. They can use other component libraries as well as custom components, which can have different layering strategies.

By adding a z-index override attribute to components, products are able to use them without overriding the z-index on private classes.

There are also defined z-indices in Material, which all components should default to using.
[Material Spec](https://www.google.com/design/spec/what-is-material/elevation-shadows.html#elevation-shadows-elevation-android-)

ThomasBurleson:
![Component z-Index Elevations](https://material-design.storage.googleapis.com/publish/material_v_8/material_ext_publish/0Bzhp5Z4wHba3VG9SaVpNbkpHb2s/whatismaterial_3d_elevation2.png)

bradrich:
This would be easily solved by our transition to mdPanel for all of the Material components. It might complicate the following of the Material spec that @ThomasBurleson added above because it would leave the positioning of items up to the developer, but it would solve several issues related to z-indexes.

PR #9256, opened 2016-08-06
feat(dialog): allow developers to specify zIndex.

### LINKED ISSUE #7101: dialog: z-index doesn't work well over a google map
Just noticed that when I display a dialog over a google map, the small links "Map data...", "Terms of use" and the Google logo at the bottom of the map are displayed OVER the dialog overlay. Their z-index are 1000000 and 1000001. The dialog overlay seems to have a z-index of 79. Why so small?
The same happens with the overlay of the sidebar.
Note: there is no issue under Chrome (and I don't know why because the z-indexes are the same...)

DevVersion:
* Allows developers to specify a custom zIndex on the dialog.
* Cleans up the `disableScrollAround` utility function (possibly fix for $mdPanel)

> Developers are currently not able to specify the z-index of the dialog dynamically and need to change it per overwriting the SCSS variables or overwriting it per CSS selectors.

> For example in #7101 a developer had a custom element, which accidentally overlays the dialog, so he had to change the z-index of the **specific** dialog (not all) 

References #8475. Fixes #7101

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -484,6 +484,7 @@ function MdDialogDirective($$rAF, $mdTheming, $mdDialog) {
  *     custom dialog directive.
  *   - `targetEvent` - `{DOMClickEvent=}`: A click's event object. When passed in as an option,
  *     the location of the click will be used as the starting point for the opening animation
+ *   - `zIndex` - `{number=}`: The z-index to place the dialog at. Defaults to 80.
  *     of the the dialog.
  *   - `openFrom` - `{string|Element|object}`: The query selector, DOM element or the Rect object
  *     that is used to determine the bounds (top, left, height, width) from which the Dialog will
@@ -560,8 +561,11 @@ function MdDialogProvider($$interimElementProvider) {
 
   return $$interimElementProvider('$mdDialog')
     .setDefaults({
-      methods: ['disableParentScroll', 'hasBackdrop', 'clickOutsideToClose', 'escapeToClose',
-          'targetEvent', 'closeTo', 'openFrom', 'parent', 'fullscreen', 'contentElement'],
+      methods: [
+        'disableParentScroll', 'hasBackdrop', 'clickOutsideToClose', 'escapeToClose',
+        'targetEvent', 'closeTo', 'openFrom', 'parent', 'fullscreen', 'contentElement',
+        'zIndex'
+      ],
       options: dialogDefaultOptions
     })
     .addPreset('alert', {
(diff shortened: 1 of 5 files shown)


src/components/dialog/dialog.js
```diff
@@ -653,17 +657,24 @@ function MdDialogProvider($$interimElementProvider) {
...
+            template +
+          '</div>';
 
         /**
-         * The specified template should contain a <md-dialog> wrapper element....
+         * The specified template should contain a <md-dialog> wrapper element.
          */
-        function validatedTemplate(template) {
+        function prepareTemplate(template) {
           if (options.autoWrap && !/<\/md-dialog>/g.test(template)) {
```
crisbeto: It's not something that was added in the PR, but it could be a good chance to change the regex with a faster `indexOf`.

DevVersion: Sounds reasonable - also confirmed that the tests would fail if I make the check wrong.

roymiloh:
@DevVersion, Is it valid for dialog and backdrop only? What if I have a `md-select` inside my `md-dialog` (which I now override his zIndex using this feature to be, say, 100), so when I click on it, a `md-select-menu-container` is created and attached to root element with z-index: 90?
`md-dialog-container` and `md-select-menu-container` are siblings, therefore `md-dialog-container` is going to hide `md-select-menu-container`.

DevVersion:
@roymiloh It is - The intention of that PR (as in the commit description) is to allow a custom zIndex for the _dialog_ 

Whatever you do with the `md-select` inside of the dialog is not related to this PR. It just gives the possibilities.

The main reason for this change is the possibilities to change the zIndex to properly work with the content of the dialog (e.g Leaflet / Maps etc.)

> Other components will / can follow.

roymiloh:
I see. So it doesn't deal with other components at all (which actually makes sense). 

However, wouldn't it be nice to have a feature (or smth related to this feature) that adjusts the relation between `md-dialog` and other components' zIndex? It's not going to be a rare case as you are now ("officially") enabling to change dialog's zIndex through his API. Probably related to another issue though.

DevVersion:
I can see the point of that suggestion, but it will require much work and it is pretty hard to let those components communicate together.

We had something similiar with the autocomplete, when a user opens a dialog from the dropdown.

As the API clearly says that this changes the zIndex of the **dialog** , developers can not expect the zIndex to respect other components.

Let's move that discussion out of that PR since this is probably something for an extra issue (feature request) once the PR is ready.

ThomasBurleson:
Will resubmit as 2 separate PRs.

Splaktar:
Unfortunately, it looks like only 1 of the two PRs mentioned in https://github.com/angular/material/pull/9256#issuecomment-245362371 were ever resubmitted. Thus the ability to specify a z-index from this PR was never merged and there is no open PR for it :(

PR #9257, opened 2016-08-06
chore(): remove final initial values

### LINKED ISSUE #9150: CSS initial value is unsupported in IE
There are a number of scss files that specify "initial" as a value for various properties. IE doesn't support the value "initial" - see http://caniuse.com/#search=initial. This can lead to unexpected behaviour because IE falls back to the next most appropriate css selector.

crisbeto:
Removes the last few instances of the `initial` CSS value which isn't support on all browsers.

Fixes #9150.

### DIFF
--- a/docs/app/css/style.css
+++ b/docs/app/css/style.css
@@ -185,10 +185,6 @@ pre {
 pre > code.highlight {
   padding: 16px;
   font-weight: 400;
-  -webkit-user-select: initial;
-  -moz-user-select: initial;
-  -ms-user-select: initial;
-  user-select: initial;
 }
 
 pre, code {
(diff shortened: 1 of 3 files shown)


docs/app/css/style.css
```diff
@@ -185,10 +185,6 @@ pre {
 pre > code.highlight {
   padding: 16px;
   font-weight: 400;
-  -webkit-user-select: initial;
-  -moz-user-select: initial;
-  -ms-user-select: initial;
```
clshortfuse: `-webkit` and `-moz` prefixes can stay since IE ignores them.

This can be changed to `-ms-user-select: text` without needing to remove anything else.

Unless there's a reason to removing the entire code block?

crisbeto: According to MDN, 'initial' is equal to 'none' but it wasn't working. Also I didn't see a point to disabling marking the text on vode samples.

clshortfuse: Yeah, I don't understand why says MDN `none`. Perhaps Firefox set it to `none` for certain elements. css-tricks.com says the default is `auto`. `auto` option which would probably make more sense than `initial` if we decided to keep the code.

There must have been some CSS that disabled user-select but I can't find anything in master. I don't it's relevant anymore and it looks like the block can be removed safely.

PR #9288, opened 2016-08-11
md-autocomplete with  md-min-length="0" not working

forceuser:
initial handleSearchText() exits on line

```js
if (searchText === previousSearchText) return;
```
because it always ```'' === ''```

and handleQuery() (-> fetchResults()) dont execute until searchText change

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -66,7 +66,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
    * Initialize the controller, setup watchers, gather elements
    */
   function init () {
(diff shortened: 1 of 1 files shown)


DevVersion:
We intentionally changed that line to set the property to an empty string (https://github.com/angular/material/commit/3f0d686d68c09da3ec384bd0eaa78b435627d8ed)

Also I'm not able to see any issues in `master` regarding using `md-min-length=0` on the autocomplete.

forceuser:
Here example http://codepen.io/anon/pen/XKoPWZ
If you focus it will not show any results
But if you input space and then delete it, result will appear

DevVersion:
@forceuser Thanks for the Pull Request, but this is definitely not the right way to fix that issue.

We are going to fix this internally, because we need to have a discussion about the chips integration as well.

PR #9289, opened 2016-08-11
autoComplete: fix for searchText bug

### LINKED ISSUE #9283: md-autocomplete: dropdown not showing on focus if min length is met
`md-autocomplete` doesn't show listed suggestions when the chips are closed, for varying list of suggestions.

**Scenario:** I want to show suggestions only for the chips that are not added. Initially, all the chips are added, so there are no more chips left to add - no suggestions, which is fine. When a chip is closed and the input field is focused, there are still no suggestions but ideally there should be one chip in the suggestions list. The suggestions appear only if one or more characters are typed in. 

`md-autocomplete` should re-render the items even when the query string length is 0, but seems like it is set to 1. This can be inferred by observing the filter function defined in `md-items` which doesn't get triggered, when the query string is empty.

[CodePen Link](http://codepen.io/sgsvenkatesh/pen/LkrdPW)
Delete the added chips and focus on the input field. Observe that no suggestions for the closed chips are being displayed by `md-autocomplete`.

sgsvenkatesh:
Handling query when min-length is zero & no change in searchText is observed. This case happens when there is no change in searchText but matches are available for selection.

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -452,6 +452,9 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     hasFocus = true;
     //-- if searchText is null, let's force it to be a string
     if (!angular.isString($scope.searchText)) $scope.searchText = '';
+    if (getMinLength() === 0 && $scope.searchText === '') {
+      handleQuery();
+    }
     ctrl.hidden = shouldHide();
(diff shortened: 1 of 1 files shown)


DevVersion:
Thanks for the Pull Request @sgsvenkatesh.

Your approach is pretty close to "fix" the real issue. 
> But unfortunately we have to close that PR in favor of #9291.

This is necessary, because your fix isn't fixing the originating issue and we need to have an internal discussion about the autocomplete integration with chips.

PR #9291, opened 2016-08-11
fix(autocomplete): properly show dropdown on focus when minlength is met

### LINKED ISSUE #9283: md-autocomplete: dropdown not showing on focus if min length is met

DevVersion:
Currently the `md-autocomplete` shows its dropdown only by changing the `searchText` from `null` to an empty string.

> This triggered a searchText change, which resulted in triggering a query.

This approach doesn't seem to be very elegant.

> The autocomplete should manually trigger a query and possibly open the dropdown if matches are available.

--
@topherfangio Can you take a look (especially on the chips with autocomplete behavior)
@ThomasBurleson I consider this as too risky for 1.1.0 (specifically the behavior with the chips)

Fixes #9283. Closes #9288. Closes #9289.

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -453,16 +453,11 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   function focus($event) {
     hasFocus = true;
 
-    // When the searchText is not a string, force it to be an empty string.
-    if (!angular.isString($scope.searchText)) {
-      $scope.searchText = '';
+    if (isSearchable() && isMinLengthMet()) {
+      handleQuery();
     }
 
     ctrl.hidden = shouldHide();
-
-    if (!ctrl.hidden) {
-      handleQuery();
-    }
 
     evalAttr('ngFocus', { $event: $event });
   }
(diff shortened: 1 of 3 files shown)


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -589,10 +591,19 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+  }
+
+  /**
+   * Determines whether the autocomplete is able to query within the current state.
+   * @returns {boolean}
+   */
+  function isSearchable() {
+    if (ctrl.loading && !hasMatches()) return false; // No query when query is in progress.
+    else if (hasSelection()) return false;           // No query if there is already a selection
+    else if (!hasFocus) return false;                // No query if the input does not have focus
```
topherfangio: Quick note: I don't think the `else if` is necessary here since you're doing a return. I'm really not sure why I had that in my `shouldHide()` code...

topherfangio:
Code-wise it looks pretty good. I'll need to do some manual testing to make sure it works as expected.

topherfangio:
@ThomasBurleson LGTM :+1:

src/components/chips/js/chipsController.js
```diff
@@ -435,7 +435,17 @@ MdChipsCtrl.prototype.removeChip = function(index) {
...
 MdChipsCtrl.prototype.removeChipAndFocusInput = function (index) {
   this.removeChip(index);
-  this.onFocus();
+
+  if (this.autocompleteCtrl) {
+    // Always hide the autocomplete dropdown before focusing the autocomplete input.
+    // Wait for the input to move horizontally, because the chip was removed.
+    // This can lead to an incorrect dropdown position.
+    this.autocompleteCtrl.hidden = true;
+    this.$mdUtil.nextTick(this.onFocus.bind(this));
```
tuhoojabotti: @DevVersion I updated to 1.1.1 and this breaks, because this.$mdUtil is undefined. It is never assigned in the constructor.

DevVersion: Known issue - see https://github.com/angular/material/pull/9528

PR #9300, opened 2016-08-12
fix(FAB/Toast): Properly reposition FABs when toasts open.

### LINKED ISSUE #9298: fab: Does not properly account for mobile toast position.
**Actual Behavior**:
 -  `What is the issue? *` On desktop devices, a toast in the same position as a FAB will push the FAB down to make room for the toast. On mobile devices, the toast always appears at the bottom, but the FAB still moves based on the original position.
 -  `What is the expected behavior?` On mobile, the FAB should only move if the FAB is at the bottom since that is the only place the toast can appear.

 -  `CodePen Demo which shows your issue:` http://codepen.io/topherfangio/pen/bZzVKy?editors=1000
 -  `Details:` It appears to only affect the `-sm` breakpoint; not the `-xs` breakpoint which is correct. Make sure to reduce the preview/browser window to force the toast to appear on the bottom, **but bigger than the `-xs` breakpoint**.

**Angular Versions**: *
 -  `Angular Version:` 1.5.8
 -  `Angular Material Version:` 1.1.0-RC*

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *` 53
 -  `OS: *` OS X
 -  `Stack Traces:`

----

  Add more lines if needed.

topherfangio:
Previously, there were many cases where a FAB would erroneously reposition itself when a Toast was opened somewhere else in the parent container (i.e. the top-right FAB would open when a top-left Toast opened).
- Add `start`/`end` classes to FAB (i.e. `md-fab-bottom.end`) which
- Fix `top` position to default to `top left` since our CSS does
  not support a full-width `top` Toast.
- Fix Toast positioning CSS to include left/right information.
- Fix CSS privatization issue with Toast `start`/`end` positioning.
  automatically update based on current LTR/RTL settings.
- Refactor Toast to use JS for mobile positioning to reduce CSS
  complexity.
- Refactor FAB SCSS to use variables for better readability.
- Update FAB/Toast docs accordingly.

Additionally, update the Toast Demo with the following related changes:
- Include `start`/`end` options.
- More clearly separate `top`/`bottom` from other positions, and
  use radio button groups instead of checkboxes.
- Refactor to use `$ctrl` instead of `$scope`.
- Add new `Delay` option to control how long the Toast remains open.
- Update FABs to use new `end` option instead of just `right` to
  work better with RTL.

BREAKING CHANGE (minor)

We used to apply a CSS class of `md-toast-open-top` or `md-toast-open-bottom` to the parent container of the toast. If you used one of these two classes, please update your code to use one of the following, more specific, classes instead:
- `md-toast-open-top-left`, `md-toast-open-top-start`
- `md-toast-open-top-right`, `md-toast-open-top-end`
- `md-toast-open-bottom-left`, `md-toast-open-bottom-start`
- `md-toast-open-bottom-right`, `md-toast-open-bottom-end`

Fixes #9298.

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -187,33 +187,81 @@ button.md-button.md-fab {
   }
 }
 
-.md-toast-open-top {
-  .md-button.md-fab-top-left,
-  .md-button.md-fab-top-right {
-    transition: $swift-ease-out;
-    transform: translate3d(0, $button-fab-toast-offset, 0);
-    &:not([disabled]) {
-      &.md-focused,
-      &:hover {
-        transform: translate3d(0, $button-fab-toast-offset - 1, 0);
-      }
+/*
+ * When a toast is open that is on the same side as our FAB, move the FAB out of the way.
+ *
+ * Note: On mobile; the toast will always open with the `md-toast-open-bottom` class, regardless of
+ * requested positioning.
+ */
+.md-toast-open-top-left .md-button.md-fab-top-left,
+.md-toast-open-top-start .md-button.md-fab-top-start,
+.md-toast-open-top-left .md-button.md-fab-top-start,
+.md-toast-open-top-start .md-button.md-fab-top-left,
+[dir=rtl] .md-toast-open-top-end .md-button.md-fab-top-left,
+[dir=rtl] .md-toast-open-top-left .md-button.md-fab-top-end,
+
+.md-toast-open-top-right .md-button.md-fab-top-right,
+.md-toast-open-top-end .md-button.md-fab-top-end,
+.md-toast-open-top-right .md-button.md-fab-top-end,
+.md-toast-open-top-end .md-button.md-fab-top-right,
+[dir=rtl] .md-toast-open-top-start .md-button.md-fab-top-right,
+[dir=rtl] .md-toast-open-top-right .md-button.md-fab-top-start
+{
+  transition: $swift-ease-out;
+  transform: translate3d(0, $button-fab-toast-offset, 0);
+  &:not([disabled]) {
+    &.md-focused,
+    &:hover {
+      transform: translate3d(0, $button-fab-toast-offset - 1, 0);
     }
   }
 }
 
-.md-toast-open-bottom {
-  .md-button.md-fab-bottom-left,
-  .md-button.md-fab-bottom-right {
-    transition: $swift-ease-out;
-    transform: translate3d(0, -$button-fab-toast-offset, 0);
-    &:not([disabled]) {
-      &.md-focused,
-      &:hover {
-        transform: translate3d(0, -$button-fab-toast-offset - 1, 0);
-      }
+// Note: We do not currently allow `md-toast-open-top`, but we do allow `md-toast-open-bottom` for
+// mobile devices.
+.md-toast-open-bottom .md-button.md-fab-bottom-start,
+.md-toast-open-bottom .md-button.md-fab-bottom-end,
+.md-toast-open-bottom .md-button.md-fab-bottom-left,
+.md-toast-open-bottom .md-button.md-fab-bottom-right,
+
+.md-toast-open-bottom-left .md-button.md-fab-bottom-left,
+.md-toast-open-bottom-start .md-button.md-fab-bottom-start,
+.md-toast-open-bottom-left .md-button.md-fab-bottom-start,
+.md-toast-open-bottom-start .md-button.md-fab-bottom-left,
+[dir=rtl] .md-toast-open-bottom-end .md-button.md-fab-bottom-left,
+[dir=rtl] .md-toast-open-bottom-left .md-button.md-fab-bottom-end,
+
+.md-toast-open-bottom-right .md-button.md-fab-bottom-right,
+.md-toast-open-bottom-end .md-button.md-fab-bottom-end,
+.md-toast-open-bottom-right .md-button.md-fab-bottom-end,
+.md-toast-open-bottom-end .md-button.md-fab-bottom-right,
+[dir=rtl] .md-toast-open-bottom-start .md-button.md-fab-bottom-right,
+[dir=rtl] .md-toast-open-bottom-right .md-button.md-fab-bottom-start
+{
+  transition: $swift-ease-out;
+  transform: translate3d(0, -$button-fab-toast-offset, 0);
+  &:not([disabled]) {
+    &.md-focused,
+    &:hover {
+      transform: translate3d(0, -$button-fab-toast-offset - 1, 0);
     }
   }
 }
+
+// Add some RTL overrides for ones we know are ok and don't need to be transformed
+[dir=rtl] .md-toast-open-top-left .md-button.md-fab-top-start,
+[dir=rtl] .md-toast-open-top-start .md-button.md-fab-top-left,
+[dir=rtl] .md-toast-open-top-right .md-button.md-fab-top-end,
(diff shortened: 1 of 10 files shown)


ThomasBurleson:
@clshortfuse, @EladBezalel - can you test/review?

DevVersion:
@ThomasBurleson - I had some discussion with @topherfangio on Slack.

> We remove that media queries, because it's weird that it overwrites the position etc.
> It could be way simpler, by just overwriting the position once (as per attribute)

>  ```js
> var isMobileScreen = !$mdMedia('gt-sm');
> 
> if (isMobileScreen) {
>   options.position = 'bottom';
>   element.addClass('md-toast-mobile');
> }
> ```
> This forces the position to `bottom` and applies a class, to make it full-width and follow the mobile specs

> This means we could also remove the checks for the size in the `onSwipe`

>  ```js
if ((direction === 'down' && options.position.indexOf('top') != -1) ||
>     (direction === 'up' && options.position.indexOf('bottom') != -1)) {
>   return;
> }

> if ((direction === 'left' || direction === 'right') && isMobileScreen) {
  return;
}
```

> Also, it looks like the actual CSS from the toast, is pretty unclear and duplicates a few things.

**Comment by clshortfuse** [comment]
@DevVersion The problem with doing it over JS instead of CSS is that you break browser resize/rotation support. The alternative is to watch the browser size, but to control animations on JS synchronously using media queries, you might be probably better off using CSS to let the browser handle it for better performance.

@topherfangio I can't help but feel that perhaps that you can share some of the CSS rule between the two and simply change the `transform-origin` based on @media breakpoint.

The PR has issues because resizing down from `gt-sm` the toast just fine. Resizing up from `sm` currently has issues.

I would say to remove the [JS media check](https://github.com/angular/material/blob/d9bd2660e1341c16f197e7caa1ad911e30ee5821/src/components/toast/toast.js#L441) and use pure CSS for this. You only need to override with: `@media (max-width: $layout-breakpoint-sm - 1) {`.  See [toast's implementation](https://github.com/angular/material/blob/9a2c47dea0cbb4347d2b6aeec8dd8ff3ae684cd1/src/components/toast/toast.scss#L146) for reference.

I would move that media check inside `onSwipe` anyway since it should be performing a media check when handling the swipe gesture, not assume screen size will remain the same as when the element was created.

File: src/components/button/button.scss [hunk 1]
```diff
@@ -187,15 +187,37 @@ button.md-button.md-fab {
...
+          transform: translate3d(0, $button-fab-toast-offset - 1, 0);
+        }
+      }
+    }
+  }
+}
+
+// Similarly, on small devices, bottom FAB buttons should be pushed out the way even if the toast
+// originally opened on the top.
+@media (max-width: $layout-breakpoint-sm) {
```
clshortfuse: Should be `$layout-breakpoint-sm - 1`

src/components/button/button.scss
```diff
@@ -187,15 +187,37 @@ button.md-button.md-fab {
...
-  .md-button.md-fab-top-right {
-    transition: $swift-ease-out;
-    transform: translate3d(0, $button-fab-toast-offset, 0);
-    &:not([disabled]) {
-      &.md-focused,
-      &:hover {
-        transform: translate3d(0, $button-fab-toast-offset - 1, 0);
+// On small devices, the toast always appears on the bottom even if it should be on the top,
+// so don't move top FABs on small devices.
+@media (min-width: $layout-breakpoint-sm) {
```
clshortfuse: This `@media` check is redundant since what you're trying to achieve is binary (one or the other). For example, it's like changing `if/elseif/else` to `if/else`

DevVersion:
@clshortfuse This is definitely a valid point, but not a really important one in my opinion.

I talked briefly with @topherfangio about that
>  I went the media query route because it was already there, and it handles cases of a window resize while the toast is open (although that’s VERY unlikely to happen to an actual user).

I'm still thinking that this case isn't really important, since it's very unlikely to happen that the browser resizes while a toast is open. And if so, then it's not a big deal if the toast is still at the top.

The user triggered it at the given position and it will stay there (regardless of the screen resize).

This change reduces the SCSS significantly and also makes it **way** clearer.
> Same for the JavaScript of the directive

clshortfuse:
I can agree with that. We can revisit this later if in terms of adding on resize/reflow support, but I doubt it would be an issue. ~~Toasts~~ Snackbars aren't meant to be persistent, and Material Design guideline states:

>For usability reasons, snackbars should not contain the only way to access a core use case. They should not be persistent or be stacked, as they are above other elements on screen.

For reference, the most I could trim the CSS down was to disable the translation.

```
@media (max-width: 959px) {
  .md-toast-open-top .md-button.md-fab-top-left:not([disabled]).md-focused,
  .md-toast-open-top .md-button.md-fab-top-left:not([disabled]):hover,
  .md-toast-open-top .md-button.md-fab-top-right:not([disabled]).md-focused,
  .md-toast-open-top .md-button.md-fab-top-right:not([disabled]):hover,
  .md-toast-open-top .md-button.md-fab-top-left,
  .md-toast-open-top .md-button.md-fab-top-right {
    transform: translate3d(0, 0, 0);
  }
}
```

You'd be able to reduce 1/2 of the CSS if we used multiple classes for direction (`md-fab-top` and `md-fab-left`) and just target `md-fab-top`.

topherfangio:
Made some significant updates. I will likely be updating the docs once #9229 is merged and we have the `docs-css-api-table` directive.

See updated title/description at top for full list of changes. Would love another review from @ThomasBurleson @DevVersion and @clshortfuse.

src/components/button/button.scss
```diff
@@ -187,34 +187,82 @@ button.md-button.md-fab {
...
-      &:hover {
-        transform: translate3d(0, $button-fab-toast-offset - 1, 0);
-      }
+/*
+ * When a toast is open that is on the same side as our FAB, move the FAB out of the way.
+ *
+ * Note: On mobile; the toast will always open with the `md-toast-open-bottom` class, regardless of
+ * requested positioning.
+ */
+.md-toast-open-top-left .md-button.md-fab-top-left,
```
EladBezalel: it feels like all those selectors can be built by a mixin - not saying we should, just suggesting :)

clshortfuse:
I feel this can get a bit CSS heavy. Is it possible to use 4 classes (top, bottom, left, right) and deprecate the paired system ?  We can use directives on the JS side to convert the paired classes to individual orientations. This way, the CSS can target only what it needs to (ie: top only rules). 

Then, your CSS can use translateY and translateX instead of repeating translate(x,y,z).

If not, you should really look into using transform-origin. Note that transform-origin: 0 0 is RTL safe, removing the need for the mixin.

EladBezalel:
@clshortfuse deprecation IMO is not an option

topherfangio:
@clshortfuse Can you elaborate on your `transform-origin` suggestion and how exactly it would be used?

clshortfuse:
@topherfangio Apparently, I was wrong about `transform-origin`. `transform-origin` works on all `transform` properties except for `translate` (ie: `scale`,`skew`,`rotate`), so we can't use it. 

The alternative is to use perspective origin, which allows to shift the perspective of a 3D object. I made an example here: http://codepen.io/shortfuse/pen/vXBrvy/

But I'm going to be honest, while the CSS is very light, the concept of 3D perspective math might make it more trouble than it's worth to maintain.

ThomasBurleson:
Closing as stale PR.

PR #9299, opened 2016-08-12
fix(dialog): remove transition classes after hide

DevVersion:
* Properly remove the transition classes on the dialog element from the transition3d animator.
* Toggle the `fullscreen` class, instead of just adding it. Necessary for prerendered dialogs, because those could change from `fullscreen: true` to `fullscreen: false`

cc. @topherfangio @EladBezalel 

Fixes #9276.

PR #9303, opened 2016-08-13
fix(dialog): do not compile an empty element when using a content element

DevVersion:
I recently noticed while playing with the pre-rendered dialog around, that despite of the pre-rendered dialog, the interim element triggers a compilation in the `$mdCompiler`.
It doesn't introduce any bugs, but it's unnecessary and could damage the performance.

Also moves the content element stuff into its own function, which makes the source more clean and allows us to maintain it easily.

Plays nicely together with #9299.

cc. @EladBezalel @topherfangio


EladBezalel:
overall lgtm

ThomasBurleson:
Dangerous changes... needs thorough manual testing.

crisbeto:
LGTM code-wise, haven't tried running it.

EladBezalel:
Tested manually, every thing is working

PR #9307, opened 2016-08-14
feat(autocomplete): allow developers to specify amount of dropdown items

### LINKED ISSUE #8751: Should be able to set max no. items to display in md-autocomplete dropdown
**Actual behavior**:
-  What is the issue? md-autocomplete only shows 2 or so items max when pulled down.
-  What is the expected behavior? Should be an option to display n items if the real-estate is available. 

You can write some css to make the list taller, but you have to hard-code it so that if the number of entries is smaller than your hard-coded value you end up with blank space. The md-autocomplete should be updated to allow a max, but still shrink when there aren't enough items to reach that max.

-  CodePen Demo which shows your issue :
-  Details:

**Angular Versions**: *
-  Angular Version: *
-  Angular Material Version: *

**Additional Information**:
-  Browser Type: *
-  Browser Version: *
-  OS: *
-  Stack Traces:

----

  Add more lines if needed.

### LINKED ISSUE #9306: autocomplete: item height inconsistency
It seems like the SCSS and JS of the `md-autocomplete` have some different values for the item height.

| CSS | JavaScript |
| ---- | ----------- |
| `48px` | `41px`    |

This deviation causes the autocomplete to not show the correct amount of items, which is specified in the CSS and JS (`5.5` by default) 

Also this introduces some _small_ scrollbars, when there are less items.

> The `5.5` items is not an even number and should be actually `5x`. 
> I'm assuming the 5.5 was introduced to correct the deviation.

DevVersion:
* Adds an attribute, which allows developer to overwrite the amount of items shown in the autocomplete dropdown.
  - When the dropdown doesn't fit into the possible space it will shrink accordingly (as same as before)
  - When the specified amount of items is higher than the current matches then it will shrink (as same as before)

* Fixes the item height deviations between CSS and JavaScript.
* Removed the unnecessary CSS for the dropdown height, because everything will be handled from the JS (same as before)
* Added tests for the dropdown height calculation, which make sure that everything is now calculated properly.

> I really like how the items are now showing up. No cutted items anymore :tada:  

Fixes #9306. Closes #8751.

### DIFF
--- a/src/components/autocomplete/autocomplete.scss
+++ b/src/components/autocomplete/autocomplete.scss
@@ -1,4 +1,5 @@
-$autocomplete-option-height: 48px !default;
+// The default item height is also specified in the JavaScript.
+$md-autocomplete-item-height: 48px !default;
 
 @keyframes md-autocomplete-list-out {
   0% {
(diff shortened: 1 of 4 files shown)


DevVersion:
@ThomasBurleson - Ping for review

kara:
@ThomasBurleson This has passed presubmit, is it ready to merge?

topherfangio:
@kara Apologies, I was in the process of doing a visual review and got side-tracked. LGTM 👍

kara:
@topherfangio No worries! Just didn't want to jump the gun.

PR #9309, opened 2016-08-15
feat(datepicker): configurable start/end dates, consistency improvements

### LINKED ISSUE #9269: [Datepicker] Select date before April 1933
**Actual Behavior**:
 -  `What is the issue? *` Currently it is not possible to pick a date before April 1933.
 -  `What is the expected behavior?` Possibility to select date between 200 years before and 200 year after now ().

 -  `CodePen Demo which shows your issue:` Not needed
 -  `Details:`

1. Open datepicker
2. Scroll up to 1933
3. Try to pick a date before 1933, ex. 18.03.1932 - you cannot scroll.

alternativly:

1. Open datepicker
2. Change to year-month view
3. Scroll up to year below 1933 (ex. 1900)
4. Pick a month
5. Datepicker backs to April 1933.

**Workaround**
It is possible to select a date before April 1933 when min-date is set, but only with Angular-Material 1.0.9.

**Angular Versions**: *
 -  `Angular Version:` 1.5.0
 -  `Angular Material Version:` 1.1.0-rc5

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *` 54
 -  `OS: *` Windows 7

crisbeto:
* Allows for the datepicker's first and last renderable dates to be configured via the $mdDateLocaleProvider. This lets developers specify when the calendar begins and ends.
* Makes the month and year views consistent by having the same start and end dates. Previously the month view started from 1933, whereas the year view started from 1816.
* Simplifies the logic for determining the start and end dates.
* Fixes not being able to use the keyboard to navigate to the first item in the year view, if a minimum date is specified.

Fixes #9269.

### DIFF
--- a/src/components/datepicker/js/calendar.js
+++ b/src/components/datepicker/js/calendar.js
@@ -89,7 +89,7 @@
    * @ngInject @constructor
    */
   function CalendarCtrl($element, $scope, $$mdDateUtil, $mdUtil,
-    $mdConstant, $mdTheming, $$rAF, $attrs) {
+    $mdConstant, $mdTheming, $$rAF, $attrs, $mdDateLocale) {
 
     $mdTheming($element);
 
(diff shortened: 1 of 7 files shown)


jelbourn:
FYI, it's good to have this configurable, but my ultimate plan (that never was high enough of a priority) was to make it truly infinite. This would require some more advanced cycling with the virtual repeat.

jelbourn:
@crisbeto Needs rebase after merging a couple other datepicker PRs

crisbeto:
It's rebased now @jelbourn.

PR #9313, opened 2016-08-15
feat(button) add md-dense support

clshortfuse:
Buttons should include proper touch target area of 48x48px. Developers
should be able to use denser buttons to target mouse users.
- Add mixin for md-dense objects
- Add minimum touch target for buttons
- Add md-dense option for buttons

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -1,8 +1,15 @@
-$button-border-radius: 3px !default;
+// Material Design Button
+// https://material.google.com/components/buttons.html
+
+$button-border-radius: 2px !default;
 $button-fab-border-radius: 50% !default;
 $button-icon-border-radius: $button-fab-border-radius !default;
 
+$button-font-size: $body-font-size-base !default;
+$button-font-size-dense: $body-font-size-base * 13/14 !default;
+
 $button-line-height: rem(3.60) !default;
+$button-line-height-dense: rem(3.20) !default;
 $button-margin: rem(0.600) rem(0.800) !default;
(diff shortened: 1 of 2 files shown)


src/components/button/button.js
```diff
@@ -179,6 +179,7 @@ function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
       });
     }
 
+    element.append('<div class="md-touch-target"></div>');
```
crisbeto: We might get a slight performance boost on this if we do it in `compile`.

clshortfuse: I'm having a bit of difficulty doing this because of the `transclude` option. Anything I append there gets removed. I've tried this, with no success:

```
compile: function (element, attr) {
      element.append('<div class="md-touch-target"></div>');
      return postLink;
    }
```

It seems like in `$compile` `$transclude` has been deprecated as well. It's advised to do interactions it in postLink instead.

>Note: The transclude function that is passed to the compile function is deprecated, as it e.g. does not know about the right outer scope. Please use the transclude function that is passed to the link function instead.

https://docs.angularjs.org/api/ng/service/$compile

crisbeto: Ah, makes sense. Then it shouldn't be a big deal.

clshortfuse:
I removed the touch target options for mdButtons I had created from this PR. I think we should off on that until later, possibly mixing with mdInteraction and HammerJS.

ThomasBurleson:
@clshortfuse - lgtm

PR #9314, opened 2016-08-15
fix(codepens): using unicode characters to avoid codepen's unescaping

### LINKED ISSUE #8761: demo: select codepen preview is not working
When clicking on the codepen button for the first select demo, the demo won't load properly inside of CodePen.

- https://material.angularjs.org/HEAD/demo/select

This is caused by the example directive usage.

```html
The <code><md-select></code> component can be used within a
```

This will be interpreted as a normal `md-select` directive, without any `ng-model`.

If the developer isn't specifying any `ng-model` on the `md-select` we throw an error.

```
Controller 'ngModel', required by directive 'mdSelect', can't be found!
```

cc. @topherfangio

EladBezalel:
- Codepen was unescaping `&lt;` (<) and `&gt;` (>) which caused, on some demos, an unclosed elements (like `<md-select>`).
Used different unicode lookalike characters so it won't be considered as an element

fixes #8761

### DIFF
--- a/docs/app/js/codepen.js
+++ b/docs/app/js/codepen.js
@@ -41,7 +41,14 @@
     function escapeJsonQuotes(json) {
       return JSON.stringify(json)
         .replace(/'/g, "&amp;apos;")
-        .replace(/"/g, "&amp;quot;");
+        .replace(/"/g, "&amp;quot;")
+        /**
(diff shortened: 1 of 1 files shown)


topherfangio:
Lol 👍  LGTM

Issue #8912, opened 2016-06-30
Build: Support Node 6


DevVersion:
NodeJS will put Node.js v6 in LTS.

![image](https://cloud.githubusercontent.com/assets/4987015/16495027/f12eafc8-3eeb-11e6-96a8-c26444686d7c.png)

Node v6 is scheduled for LTS at `2016-10-01`.

---
Currently developers are not able to build the Angular Material sources with Node v6.
We should fix that and make sure that everything will work for Node v6.

Probably Node v5 will be no longer supported for building the Angular Material sources.

clshortfuse:
Upgrading `gulp-if` to `2.0.1` fixes it for me on ChromeOS/Linux.

DevVersion:
Note: Referencing #8850 to update the dependencies.

PR #9320, opened 2016-08-16
build(): add script to remove duplicate CSS properties

crisbeto:
Adds a custom script to remove duplicated CSS properties during builds, because cssnano only does it if the declarations are exactly the same. Note that this will skip prefixed properties and will print a warning with the offending selectors.

CC @jelbourn

### DIFF
--- a/gulp/util.js
+++ b/gulp/util.js
@@ -19,6 +19,8 @@ var ngAnnotate = require('gulp-ng-annotate');
 var insert = require('gulp-insert');
 var gulpif = require('gulp-if');
 var nano = require('gulp-cssnano');
+var postcss = require('postcss');
+var _ = require('lodash');
 var constants = require('./const');
 var VERSION = constants.VERSION;
 var BUILD_MODE = constants.BUILD_MODE;
@@ -33,6 +35,7 @@ exports.filterNonCodeFiles = filterNonCodeFiles;
 exports.readModuleArg = readModuleArg;
 exports.themeBuildStream = themeBuildStream;
 exports.minifyCss = minifyCss;
+exports.dedupeCss = dedupeCss;
 exports.args = args;
 
 /**
@@ -181,6 +184,7 @@ function buildModule(module, opts) {
         // In some cases there are multiple theme SCSS files, which should be concatenated together.
         .pipe(gulpif, /default-theme.scss/, concat(name + '-default-theme.scss'))
         .pipe(sass)
+        .pipe(dedupeCss)
         .pipe(autoprefix)
     (); // Invoke the returning lazypipe function to create our new pipe.
   }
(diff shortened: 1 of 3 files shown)


gulp/util.js
```diff
@@ -223,3 +225,75 @@ function themeBuildStream() {
...
+        // be allowed to be repeated.
+        var hasDuplicates = find(rule.nodes, function(otherProp) {
+          return otherProp !== prop &&
+            prop.prop &&
+            otherProp.prop &&
+            otherProp.prop === prop.prop &&
+            !containsPrefix(prop.prop) &&
+            !containsPrefix(prop.value) &&
+            !containsPrefix(otherProp.prop) &&
+            !containsPrefix(otherProp.value);
```
DevVersion: Would it make sense to remove the unnecessary iteration here?

Current approach
```scss
selector {
  -webkit-transform: someValue; // Iteration 1x (iterate through all properties) -> Do nothing (prefixed)
  -ms-transform: someValue;     // Iteration 2x (iterate through all properties) -> Do nothing (prefixed)
  transform: someValue;         // Iteration 3x (iterate through all properties) -> delete
  transform: anotherValue;      // Iteration 4x (iterate through all properties) -> None found. Do nothing
}
```

New approach
```scss
selector {
  -webkit-transform: someValue; // Iteration 1x (prefixed - skip)
  -ms-transform: someValue;     // Iteration 2x (prefixed - skip)
  transform: someValue;         // Iteration 3x (iterate through all properties) -> delete
  transform: anotherValue;      // Iteration 4x (iterate through all properties) -> None found. Do nothing
}
```

gulp/util.js
```diff
@@ -223,3 +225,75 @@ function themeBuildStream() {
...
+    }
+
+    // Turn the AST back into CSS.
+    return parsed.toResult().css;
+  });
+
+  function containsPrefix(str) {
+    return str && str.indexOf('-') > -1 && !!find(prefixes, function(prefix) {
+      return str.indexOf(prefix) > -1;
+    });
```
DevVersion: Can't you just do?

```js
return !!prefixes.filter(function(prefix) {
  return str.startsWith(prefix);
}).length;
```

gulp/util.js
```diff
@@ -223,3 +225,75 @@ function themeBuildStream() {
...
+  }
+
+  function find(array, predicate) {
```
DevVersion: You can use `lodash` here instead. It's part of our build tools, so no need to install it.

> https://lodash.com/docs#find

gulp/util.js
```diff
@@ -223,3 +225,75 @@ function themeBuildStream() {
...
+    // Walk through all the rules, skipping comments, media queries etc.
+    parsed.walk(function(rule) {
+      if (rule.type !== 'rule' || !rule.nodes || rule.nodes.length < 2) return;
+
+      // Walk all of the properties within a rule.
+      rule.walk(function(prop) {
+        // Check if there's a similar property that comes after the current one.
+        // Note that we skip the prefixed properties and values, because they should
+        // be allowed to be repeated.
+        var hasDuplicates = find(rule.nodes, function(otherProp) {
```
jelbourn: To make this a bit more readable, I would do something like
```js
// If either definition is blank, or if this is the same exact property definition, continue to the next.
if (!prop || !otherProp || prop === otherProp) {
  return false;
}

var isSamePropertyName = prop.prop === otherProp.prop;
var definitionContainsPrefix = 
    [prop.prop, prop.value, otherProp.value].some(value => containsPrefix(value));

return isSamePropertyName && !defintionContainsPrefix;
```

gulp/util.js
jelbourn: Could do this with a regex?
```js
/^-(webkit|moz|ms|o)-.+/
```

DevVersion: I think that would be definitely clearer  than iterating through all prefixes.

gulp/util.js
```diff
@@ -223,3 +225,75 @@ function themeBuildStream() {
...
+          }
+        }
+      });
+    });
+
+    if (removedCount) {
+      gutil.log(
+        gutil.colors.yellow('Removed ' + removedCount + ' duplicate CSS properties from selectors: ' +
+          removedSelectors.map(function(selector) {
+            return '"' + selector + '"';
```
jelbourn: Can you print out the property name and value as well?

jelbourn: You could probably also clean this up w/ an ES6 template string.
(not sure if you have any reason to not use es6 features in these build scripts)

DevVersion: ES6 shouldn't be used right now. The build process will use ES6 once we support Node 6 and probably when v6 becomes LTS.

See #8912

gulp/util.js
```diff
@@ -33,6 +34,7 @@ exports.filterNonCodeFiles = filterNonCodeFiles;
 exports.readModuleArg = readModuleArg;
 exports.themeBuildStream = themeBuildStream;
 exports.minifyCss = minifyCss;
+exports.dedupeCss = dedupeCss;
```
jelbourn: This seems to only be running for the full build of `angular-material.css` now, but I would also need it to run for the individual module css that it outputted from running `build-all-module` (specifically `gulp build-all-module --closure`).

gulp/util.js
crisbeto: It's actually faster too, after I replaced it, the whole method went from ~120ms to 30-50ms. Although I had to tweak it since it didn't pick up the prefixed transitions properly.

gulp/util.js
crisbeto: Yeah, I went with ES5 to be on the safe side, although some of the other build scripts use ES6.

gulp/util.js
crisbeto: It's added everywhere now.

crisbeto:
I've updated it to include the feedback and applied it to the individual module builds as well.

ThomasBurleson:
screen shot 2016-08-17 at 8 25 40 am

PR #9335, opened 2016-08-18
fix(datepicker, menu, slider): remove duplicate properties

crisbeto:
* Reworks the `md-calendar-cell` mixin to allow for the height to be specified, avoiding the need for a duplicate property when overriding.
* Removes the `text-align: start` from the menu bar since, at least according to MDN, it's not supported by IE. Also the rtl mixin right before it does exactly the same.
* Reworks the rtl setup for the inverted slider so it doesn't result in duplicate properties.


ThomasBurleson:
lgtm

PR #9338, opened 2016-08-18
feat(sidenav): configurable scroll prevent target

### LINKED ISSUE #8634: $mdSideNav prevent scrolling on element other than parent on open.
**Actual behavior**:
-  When my sideNav is open, i can scroll thru its content, when i reach the end of the content of the sideNav, the main-content starts to scroll *
-  I dont want the main-content to scroll when i have a sideNav open.

-  CodePen Demo which shows your issue : http://codepen.io/Donniewiko/pen/OXVXjm/
-  Details: When you click on the Button, a sideNav will open. Scrolling isn't prevented on the main content so you can scroll all you want, which isnt expected or wanted. On row 3 I've placed a div which re-creates the problem. In our scenario this div is actually an ngInclude.

**Angular Versions**: *
-  Angular Version: 1.4.8
-  Angular Material Version: 1.0.0

**Additional Information**:
-  Browser Type: all major browsers.*
-  Browser Version: latest*
-  OS: all (both mobile and desktop)*
-  Stack Traces:

I have narrowed the issue down to the method DisableParentScroll: https://github.com/angular/material/blob/master/src/components/sidenav/sidenav.js#L350

In the pen, the parent element of the sideNav is div#troubleBringer. So when i open the sideNav the div will get overflow:hidden. But i want the parent of the parent (md-content) to get overflow:hidden;

I have tried to create my own solution to this issue. I tried to use the $mdSidenav(id).isOpen(); method. But this resulted in unwanted side effects.

---
# Possible solution:

I have little knowledge of pull requests. But i want to propose a solution:
Create an attribute on the sideNav which can be used to provide an alternate element to disable the scroll.
Here is how i fixed it:
SideNav.js:

in the [postLink](https://github.com/angular/material/blob/master/src/components/sidenav/sidenav.js#L255):

    ...
    var preventScrollElement = element.parent();

    if (attr.preventScrollTarget) {
       var domElement = document.querySelectorAll(attr.preventScrollTarget);

       if (domElement.length !== 1) {
          $log.error(attr.preventScrollTarget + " target should target exactly 1 element. " + domElement.length + " elements found!");
       }

       preventScrollElement = angular.element(domElement);
    }
    ...

And the method [DisableParentScroll](https://github.com/angular/material/blob/master/src/components/sidenav/sidenav.js#L350):

      function disableParentScroll(disabled) {
      if ( disabled && !lastParentOverFlow ) {

         lastParentOverFlow = preventScrollElement.css('overflow');
         preventScrollElement.css('overflow', 'hidden');

      } else if (angular.isDefined(lastParentOverFlow)) {

        preventScrollElement.css('overflow', lastParentOverFlow);
        lastParentOverFlow = undefined;

      }
    }

then you can set the attribute on the sideNav like this:

    <md-sidenav prevent-scroll-target="#mainContent" class="md-sidenav-right md-whiteframe-4dp" md-component-id="right">

crisbeto:
* Adds the ability to select which element's scrolling will be disabled when a sidenav is open. This can be useful in the cases where the scrollable container isn't the direct parent of the sidenav.
* Adds unit tests for the parent scroll prevention.

Fixes #8634

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -232,6 +232,8 @@ function SidenavFocusDirective() {
  * @param {expression=} md-is-locked-open When this expression evaluates to true,
  * the sidenav 'locks open': it falls into the content's flow instead
  * of appearing over it. This overrides the `md-is-open` attribute.
+ * @param {string=} md-disable-scroll-target Selector, pointing to an element, whose scrolling will
+ * be disabled when the sidenav is opened. By default this is the sidenav's direct parent.
  *
 * The $mdMedia() service is exposed to the is-locked-open attribute, which
  * can be given a media query or one of the `sm`, `gt-sm`, `md`, `gt-md`, `lg` or `gt-lg` presets.
@@ -261,6 +263,7 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, 
   function postLink(scope, element, attr, sidenavCtrl) {
     var lastParentOverFlow;
     var backdrop;
+    var disableScrollTarget = null;
     var triggeringElement = null;
     var previousContainerStyles;
     var promise = $q.when(true);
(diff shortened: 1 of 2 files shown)


src/components/sidenav/sidenav.js
```diff
@@ -275,8 +278,23 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
       });
     };
 
+    if (attr.mdDisableScrollTarget) {
+      disableScrollTarget = $document[0].querySelector(attr.mdDisableScrollTarget);
+
+      if (disableScrollTarget) {
+        disableScrollTarget = angular.element(disableScrollTarget);
+      } else {
+        $log.warn($mdUtil.supplant('mdSidenav: couldn\'t find element matching ' +
```
ThomasBurleson: Perhaps we should convert all `$log.warn()` to `$log.debug()` since `.debug()` can be disabled globally with `$logProvider.debugEnabled(false)`

crisbeto: I'm not sure we'd want that, because it could lead to cases where this might break in production, but the user wouldn't be able to see what went wrong.

ThomasBurleson: This is a breaking feature then... because the framework will complain using the default configuration (without a disabel scroll target).

crisbeto: It won't complain. If it's not specified, it will default to the direct parent like it does at the moment.

ThomasBurleson: You are correct. I was confused with the code above... but the full code view shows that my concern is invalid. Thx.

PR #9344, opened 2016-08-18
feat(md-input) add mdDense support and realign to spec

clshortfuse:
** WIP DO NOT MERGE **

md-input always had extra margins for placement of labels and messages
without every reducing or removing them if said elements were not present.
Label components were not properly changing based on font-size, but instead,
used a scaling system that had poor dynamic font-size support and introduced
pixel aliasing on certain browsers.

 - Add and remove `md-input-has-label` class dynamically
 - Add and remove `md-input-has-messages` class dynamically
 - Position label with the proper distance from the top per spec
 - Transition label position and font-size depending on focus state per spec
 - Use `will-change` CSS to alert browser of animation change for smoothness
 - Set input height relative to font-size with a minimum of 16px per spec
 - Revert `box-sizing` to default to properly space underline per spec
 - Use dynamic upper margins dependent on label existance
 - Use dynamic lower margins dependent on message existance
 - Correct animation timings for smoother font-size transition
 - Remove unneeded CSS (ie: `order` is only used in flex layouts)
 - Add md-dense support

This commit allows more flexibility with labels and input boxes. Input elements
can now support any font-size. Labels will now resize dynamically, correcting
issues related to text with width longer than space available. Various costly
CSS rules have also been removed. Performance should also be better due to
use of `will-change` and initialization with `transform3d`.

TODO:
- [ ] Increasing click target for touch interaction
- [ ] Create workaround for helper text or just build support for it
- [ ] Prune unneeded CSS
- [ ] Correct horizontal offsets per spec
- [ ] Correct md-select positioning
- [ ] Correct md-autocomplete positioning
- [ ] Correct md-datepicker positioning

Rebase of #8209

Fixes: (TODO)

### DIFF
--- a/docs/app/js/anchor.js
+++ b/docs/app/js/anchor.js
@@ -6,7 +6,7 @@
     .directive('h2', MdAnchorDirective)
     .directive('h1', MdAnchorDirective);
 
-  function MdAnchorDirective($mdUtil) {
+  function MdAnchorDirective($mdUtil, $compile) {
 
     /** @const @type {RegExp} */
     var unsafeCharRegex = /[&\s+$,:;=?@"#{}|^~[`%!'\].\/()*\\]/g;
@@ -14,12 +14,24 @@
     return {
       restrict: 'E',
       scope: {},
-      template: '<a class="docs-anchor" ng-href="#{{ name }}" name="{{ name }}" ng-transclude></a>',
-      transclude: true,
+      require: '^?mdContent',
       link: postLink
     };
 
-    function postLink(scope, element) {
+    function postLink(scope, element, attr, ctrl) {
+
+      // Only create anchors when being inside of a md-content.
+      if (!ctrl) {
+        return;
+      }
+
+      var anchorEl = $compile('<a class="docs-anchor" ng-href="#{{ name }}" name="{{ name }}"></a>')(scope);
+
+      // Wrap contents inside of the anchor element.
+      anchorEl.append(element.contents());
+
+      // Append the anchor element to the directive element.
+      element.append(anchorEl);
 
       // Delay the URL creation, because the inner text might be not interpolated yet.
       $mdUtil.nextTick(createContentURL);
@@ -40,6 +52,6 @@
   }
 
   // Manually specify $inject because Strict DI is enabled.
-  MdAnchorDirective.$inject = ['$mdUtil'];
+  MdAnchorDirective.$inject = ['$mdUtil', '$compile'];
 
 })();


src/components/input/input.scss
```diff
@@ -125,20 +152,14 @@ md-input-container {
...
-
-    // The max-width is necessary, because in some browsers, using this together with
-    // a calc might cause it to overflow the parent. See #7403
-    max-width: 100%;
-
+    font-size: inherit;
+    will-change: transform, font-size; // prep animation for smoothness
+    transform: translate3d(0, $input-label-default-offset, 0);
+    transition: transform $swift-ease-out-duration $swift-ease-out-timing-function,
+                font-size $swift-ease-out-duration $swift-ease-out-timing-function;
```
topherfangio: One thing we need to test: does transitioning the `font-size` cause layout recalculations? I know this is why we used only the translate/scale originally, but I'm not sure if it affects font-size or not.

clshortfuse: I originally set it to `position: relative`, but I ran into the issue of it modifying messing around with width of the component. I couldn't figure out a way to get label to only affect layouts vertically and not horizontally.

`label` has `position: absolute` so it won't affect layout and I just set a really tall margin (now padding in my code) on the input container.

I was more worried about if browsers supported animation of `font-size`. I haven't tested this rebase, but the original one worked fined in Safari, IE10-Edge, Firefox, and Chrome. Color me surprised about that IE 10 compatibility.

clshortfuse:
[reserved for screenshots]

src/components/input/input-animations.spec.js
```diff
@@ -59,8 +58,9 @@ describe('md-input-container animations', function() {
     expectError(getError(), 'pattern');
     expect(container).toHaveClass('md-input-invalid');
     expect(lastAnimateCall.element).toEqual(getError());
-    expect(lastAnimateCall.options.event).toEqual('enter');
-    expect(lastAnimateCall.options.to).toEqual({"opacity": 1, "margin-top": "0"});
+    var computedStyle = getComputedStyle(lastAnimateCall.element[0]);
+    expect(computedStyle.opacity).toEqual('1');
+    expect(computedStyle['margin-top']).toEqual('0px');
```
DevVersion: You can drop that bracket notation here and use the camel case. 

```js
getComputedStyle($0).marginTop
> "100px"
```

src/components/input/input.scss
```diff
@@ -108,10 +172,18 @@ md-input-container {
   }
 
   label:not(.md-container-ignore) {
-    position: absolute;
-    bottom: 100%;
-    @include rtl(left, 0, auto);
-    @include rtl(right, auto, 0);
+    display: block; // fixes Firefox bug with label and auto positioning
+    position:absolute;
```
DevVersion: Nit: Space here is missing.

PR #9345, opened 2016-08-18
fix(prefixer): do not throw an exception if element is undefined

DevVersion:
* The prefixer currently throws an error, when the jqLite element / native element is empty / undefined.
  This was noticeable when having a `md-fab-actions` element without any child elements.

**FYI**: Checking the `nodeType` is more elegant, than comparing the prototype. It's the same way jQuery uses as well.

cc. @ErinCoughlan

### DIFF
--- a/src/core/util/prefixer.js
+++ b/src/core/util/prefixer.js
@@ -49,7 +49,11 @@ function MdPrefixer(initialAttributes, buildSelector) {
   }
 
   function _hasAttribute(element, attribute) {
-    element = element[0] || element;
+    element = _getNativeElement(element);
+
+    if (!element) {
+      return false;
+    }
 
     var prefixedAttrs = _buildList(attribute);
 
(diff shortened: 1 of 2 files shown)


topherfangio:
LGTM :+1:

src/core/util/prefixer.js
```diff
@@ -63,10 +67,30 @@ function MdPrefixer(initialAttributes, buildSelector) {
...
   }
+
+  /**
+   * Parses a jqLite or DOM element into a HTML element.
+   * This is useful when supporting jqLite elements and DOM elements at
+   * same time.
+   * @param element {JQLite|Element} Element to be parsed
+   * @returns {HTMLElement} Parsed HTMLElement
+   */
+  function _parseElement(element) {
```
jelbourn: `parse` isn't really the right word. It implies that you're taking a string recognizing it into something else. I'd use `_getNativeElement`

jelbourn:
LGTM

PR #9347, opened 2016-08-18
fix(select): Ensure `md-no-asterisk` attribute works.

### LINKED ISSUE #9339: md-select: md-no-asterisk not working
**Actual Behavior**:
 -  Having an md-select with the required attribute and a placeholder attribute. 
 -  I would expect to disable the appended asterisk on the placeholder with the md-no-asterisk attribute. But this does not work, the asterisk is appended.

**CodePen** (or steps to reproduce the issue): See behavior description.
 -  `Details:`

**Angular Versions**: *
 -  `Angular Version: 1.5.8`
 -  `Angular Material Version: 1.1.0`

**Additional Information**:
 -  `Browser Version: 52`
 -  `OS: Windows 10`
 -  `Stack Traces:`

topherfangio:
After a recent change to the asterisk colors, the `md-no-asterisk` option was no longer working.

 - Update code to add a new `.md-no-asterisk` CSS class and fix styles.
 - Add appropriate tests.
 - Add `md-no-asterisk="false"` to demo so users can easily turn it on
   and test the fixed functionality.

Fixes #9339.

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -74,7 +74,8 @@ angular.module('material.components.select', [
  * @param {expression=} md-selected-text Expression to be evaluated that will return a string
  * to be displayed as a placeholder in the select input box when it is closed.
  * @param {string=} placeholder Placeholder hint text.
- * @param md-no-asterisk {boolean=} When set to true, an asterisk will not be appended to the floating label.
+ * @param md-no-asterisk {boolean=} When set to true, an asterisk will not be appended to the
+ * floating label. **Note:** This attribute is only evaluated once; it is not watched.
  * @param {string=} aria-label Optional label for accessibility. Only necessary if no placeholder or
  * explicit label is present.
  * @param {string=} md-container-class Class list to get applied to the `.md-select-menu-container`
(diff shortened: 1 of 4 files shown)


src/components/select/select.js
```diff
@@ -74,7 +74,8 @@ angular.module('material.components.select', [
  * @param {expression=} md-selected-text Expression to be evaluated that will return a string
  * to be displayed as a placeholder in the select input box when it is closed.
  * @param {string=} placeholder Placeholder hint text.
- * @param md-no-asterisk {boolean=} When set to true, an asterisk will not be appended to the floating label.
+ * @param md-no-asterisk {boolean=} When set to true, an asterisk will not be appended to the
```
ThomasBurleson: Why is this an opt-out? Would an opt-in not be a better approach?

topherfangio: I have always pushed for the opt-in approach; however, this has been around since before 1.0 (before I joined the team I think) so I don't think we should change the API at this point.

All I did here was add a note about it only being evaluated once; not being watched.

ThomasBurleson:
This seems so inverted backwards....

Issue #9126, opened 2016-07-25
md-autocomplete has an implicit "track by" unless the result is highlighted


mgol:
**Actual Behavior**:
 -  `What is the issue? *`
If you have an `md-autocomplete` element with `md-item-template`, you can't use one-time bindings inside of the template *unless* you also add `md-highlight-text`. It seems as if there's an implicit track-by in the repeater.
 -  `What is the expected behavior?`
The repeater should not reuse elements unless a user explicitly asked for it (via something like `track by`).

**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:`
https://codepen.io/mgol/pen/JKvkjv
https://codepen.io/mgol/pen/EyLRxo
 -  `Details:`
Both of those codepens are almost identical to the first one embedded at https://material.angularjs.org/latest/demo/autocomplete. The only differences are that the first one has `md-highlight-text` removed and both have the `item.display` binding changed to be a one-time one.

If you click in the input and type `s`, the first codepen will display Alabama & Alaska instead of South Carolina & South Dakota

**Angular Versions**: *
 -  `Angular Version:`: 1.5.5 & 1.5.8
 -  `Angular Material Version:` 1.1.0-rc.5 & the version compiled from 0356bed24789e2e9a1c7d7842303031f03e68760

**Additional Information**:
 -  `Browser Type: *` All
 -  `Browser Version: *` All
 -  `OS: *` All
 -  `Stack Traces:` N/A

----
Shortcut to create a [new CodePen Demo](http://codepen.io/team/AngularMaterial/pen/bEGJdd).
Note: `*` indicates required information. Without this information, your issue may be auto-closed.

> Do not modify the titles or questions. Simply add your responses to the ends of the questions.
  Add more lines if needed.

DevVersion:
As discussed with @mgol, we won't change that behavior on the `virtual-repeat`, since the virtual repeat is supposed to recycle the old *blocks*

We will document that on the `virtual-repeat` API and link to it from the autocomplete component.

PR #9360, opened 2016-08-19
docs: Fix many links and other issues.

### LINKED ISSUE #9352: docs: some links leads to 404
There is some links leading to 404 on 1.1.0 docs.

I only found these:
- Link from [\<md-subheader\> directive](https://material.angularjs.org/1.1.0/api/directive/mdSubheader) to `$mdSticky`.
- Links from [\<md-autocomplete\> directive (https://material.angularjs.org/1.1.0/api/directive/mdAutocomplete) to `VirtualRepeat` and `VirtualRepeatContainer`.

I can see it's only happening with internal links.

topherfangio:
There are a few places in the docs that point to `/something/somewhere` which works on localhost, but fails on the live site because the live site adds `/latest` or `/HEAD` to the URL.

 - **Update**: Absolute links to be relative.
 - **Update**: Material Design links to use new url (material.goole.com) to
   avoid redirects (and potentially broken links in the future).

Additionally, fix a few bugs with some of our doc-specific code.

 - **Fix**: Null pointer exception when sidenav is closed (i.e. on mobile).
 - **Fix**: `<hljs>` code snippets removed useful whitespace.

Additionally, update the Getting Started page with formatting/styling improvements and new links.

 - **Update**: Formatting/styling/extra tags.
 - **New**: Add new links for Material Start.
 - **Fix**: Link titles all say `Link opens in a new window`.

Finally, fix various typos/formatting issues including (but not limited to):

 - **Update**: Notes should look the same across pages.
 - **Fix**: Incorrect markdown support in panel docs.
 - **Update**: Add `<hljs>` snippet formatting to panel docs.
 - **New**: Add description to `$mdAriaProvider.disableWarnings()` method.

   This was mainly because it visually looked odd not to have one.

Fixes #9352.

Ping @ThomasBurleson @EladBezalel for general review and @ErinCoughlan for panel review.


src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -35,10 +35,10 @@ angular
  * There is an example below of how this should look.
  *
  * ### Notes
- * The `md-autocomplete` uses the the [VirtualRepeat](/api/directive/mdVirtualRepeatContainer)
+ * The `md-autocomplete` uses the the [VirtualRepeat](api/directive/mdVirtualRepeat)
```
DevVersion: This is intentionally linked to the `mdVirtualRepeatContainer`. 
> The label was just Virtual Repeat because the container is actually the over-all component.

topherfangio: Ok, well I added links on both the VirtualRepeat and the VirtualRepeatContainer to easily jump between them.

I think seeing the docs for the VirtualRepeat is more useful than the container?

DevVersion: The `virtual-repeat` directive itself does actually not contain any useful information about the way it works.

I intentionally linked to the `VirtualRepeatContainer`, because it is responsible for handling all the repeated blocks and includes all insight information.

The most important reason, is just to link to the Virtual Repeat container because of the common issues like https://github.com/angular/material/issues/9126.

I know it's linked below as well - but this is just to make the developer explicitly aware of the Virtual Repeat Container when submitting new issues regarding the dropdown.

> We can keep it as it is, just wanted to clarify why I linked it to the container instead.

docs/app/partials/getting-started.tmpl.html
```diff
@@ -1,87 +1,123 @@
...
+        <a href="https://github.com/angular/material-start/tree/master" target="_blank"
+           title="Material-Start - ES5">Learn with Material-Start: Completed (es5)</a>
+      </li>
+
+      <li>
+        <a href="http://codepen.io/team/AngularMaterial/pen/RrbXyW" target="_blank">Start with a
+        blank CodePen Material Application</a>
+      </li>
+
+      <li style="margin-bottom: 30px;">
```
EladBezalel: maybe a class? (i don't like inline styles)

EladBezalel:
This looks good ! :+1:

ThomasBurleson:
Gatekeeper - LGTM

PR #9363, opened 2016-08-20
fix(virtual-repeat): not re-rendering when switching to a smaller list

### LINKED ISSUE #9315: Scrolled md-virtual-repeat doesn't rerender when the items change
**Actual Behavior**:
 -  `What is the issue?`
Scrolled md-virtual-repeat doesn't rerender when the items change for some configurations of the scroll and the item lists. Occurs when it enters this statement,
https://github.com/angular/material/blame/master/src/components/virtualRepeat/virtual-repeater.js#L700
but the indexes don't change in updateIndexes so this condition fails,
https://github.com/angular/material/blame/master/src/components/virtualRepeat/virtual-repeater.js#L659
And it doesn't reenter the update method and rerender the list with the new items.
Rerenders to show the new items.

http://codepen.io/benkraft/pen/VjgJNQ
 -  `Details:`
Scroll the list down just a couple of pixels, perhaps so the top border is hidden but all the text on the label "black" is still visible. Click "Switch to new color list". Verify that the list did not switch items. You can reload and click the button without scrolling and it indeed should switch to the correct list.

**Angular Versions**: *
 -  `Angular Version:`  
1.5
 v1.0.0-rc5 (and earlier)

**Additional Information**:
 -  `Browser Type: *` 
Chrome
 -  `Browser Version: *`
51.0.2704.103
 -  `OS: *`
OS X 10.11.6
 -  `Stack Traces:`

----

  Add more lines if needed.

crisbeto:
Fixes the virtual repeater not re-rendering, if the list is scrolled slightly and the user switches to a shorter array of items.

Fixes #9315.

### DIFF
--- a/src/components/virtualRepeat/virtual-repeater.js
+++ b/src/components/virtualRepeat/virtual-repeater.js
@@ -696,13 +696,12 @@ VirtualRepeatController.prototype.virtualRepeatUpdate_ = function(items, oldItem
   var itemsLength = items && items.length || 0;
   var lengthChanged = false;
 
-  // If the number of items shrank
+  // If the number of items shrank, keep the scroll position.
   if (this.items && itemsLength < this.items.length && this.container.getScrollOffset() !== 0) {
     this.items = items;
     var previousScrollOffset = this.container.getScrollOffset();
     this.container.resetScroll();
     this.container.scrollTo(previousScrollOffset);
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@jelbourn - LGTM.

PR #9385, opened 2016-08-23
fix(docs): Fix dangling links.

topherfangio:
Dgeni shows a warning when it cannot find the correct link files. Since we are using an Angular app, the only way I have found around it is to use the `ng-href` attribute so Dgeni ignores them.

Ping @ThomasBurleson for review.

### DIFF
--- a/src/components/autocomplete/js/autocompleteDirective.js
+++ b/src/components/autocomplete/js/autocompleteDirective.js
@@ -35,10 +35,10 @@ angular
  * There is an example below of how this should look.
  *
  * ### Notes
- * The `md-autocomplete` uses the the [VirtualRepeat](api/directive/mdVirtualRepeatContainer)
+ * The `md-autocomplete` uses the the <a ng-href="api/directive/mdVirtualRepeatContainer">VirtualRepeat</a>
  * directive for displaying the results inside of the dropdown.<br/>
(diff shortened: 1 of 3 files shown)


src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -35,10 +35,10 @@ angular
  * There is an example below of how this should look.
  *
  * ### Notes
- * The `md-autocomplete` uses the the [VirtualRepeat](api/directive/mdVirtualRepeatContainer)
+ * The `md-autocomplete` uses the the <a ng-href="api/directive/mdVirtualRepeatContainer">VirtualRepeat</a>
```
crisbeto: Couldn't you use the `{@link}` directive here? AFAIK it reverses the URL properly, e.g.

```
{@link material.components.virtualRepeat}
```

topherfangio: @crisbeto I haven't actually seen that before. Is that in Dgeni or Angular?

crisbeto: I'm not sure, but I've been using it for a while. Not sure if it would work in our case, because of the different versions in the URL. Might be worth a shot though.

PR #8850, opened 2016-06-24
chore(): cleanup dependencies

DevVersion:
```
Package                        Current   Wanted  Latest Location
---------------------------------------------------------------------------------
- [X] angularytics              0.3.0    0.3.0   0.4.0  angular-material-source
- [ ] conventional-changelog   0.0.17      git     git  angular-material-source
- [-] esprima                   1.2.5    1.2.5   2.7.2  angular-material-source
- [X] gulp-autoprefixer         1.0.1    1.0.1   3.1.0  angular-material-source
- [X] gulp-connect              2.3.1    2.3.1   4.1.0  angular-material-source
- [X] gulp-filter               1.0.2    1.0.2   4.0.0  angular-material-source
- [X] gulp-if                   1.2.5    1.2.5   2.0.1  angular-material-source
- [X] gulp-insert               0.4.0    0.4.0   0.5.0  angular-material-source
- [X] gulp-jshint              1.12.0   1.12.0   2.0.1  angular-material-source
- [X] gulp-ng-annotate          0.5.3    0.5.3   2.0.0  angular-material-source
- [X] gulp-ng-html2js           0.1.8    0.1.8   0.2.2  angular-material-source
- [X] jquery                    2.2.4    2.2.4   3.0.0  angular-material-source
- [X] jshint-summary            0.3.0    0.3.0   0.4.0  angular-material-source
- [X] karma                   0.13.22  0.13.22   1.0.0  angular-material-source
- [X] karma-firefox-launcher    0.1.7    0.1.7   1.0.0  angular-material-source
- [X] karma-jasmine             0.3.8    0.3.8   1.0.2  angular-material-source
- [X] karma-sauce-launcher     0.2.14   0.2.14   1.0.0  angular-material-source
- [X] lazypipe                  0.2.4    0.2.4   1.0.1  angular-material-source
- [X] lodash                   3.10.1   3.10.1  4.13.1  angular-material-source
- [-] merge-stream              0.1.8    0.1.8   1.0.0  angular-material-source
- [?] prompt-sync               1.0.0    1.0.0   4.1.4  angular-material-source
- [X] through2                  0.6.5    0.6.5   2.0.1  angular-material-source
```
X: Updated Package;
?: Not sure about update
-: Removed Package

Closes #7956

PR #9393, opened 2016-08-23
fix: undo change to unknown symbol for prod build

jelbourn:
PR #8850 made this change, but it seems unintentional because the symbol `cleanCSS` doesn't exist anywhere.

### DIFF
--- a/gulp/util.js
+++ b/gulp/util.js
@@ -130,7 +130,7 @@ function buildModule(module, opts) {
 
   function buildMin() {
     return lazypipe()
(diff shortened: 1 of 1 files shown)


jelbourn:
It's also odd that the CI didn't catch this.

ThomasBurleson:
@DevVersion - any thoughts on this?

DevVersion:
@ThomasBurleson This is valid - sorry. May be introduced while rebasing with the new Nano CSS 
commit. I likely rebased 8 times :smile:

Issue #9395, opened 2016-08-24
ie10 MutationObserver is undefined


j3ski:
**Actual Behavior**:
 -  `What is the issue? *`
When I enter a page which uses md-tabs i get an error: "ReferenceError: 'MutationObserver' is undefined"
 -  `What is the expected behavior?`
I shouldn't be getting the error

**CodePen** (or steps to reproduce the issue): *
 - Go to [the official demo](https://material.angularjs.org/latest/demo/tabs) on ie10
 - [possible culprit](https://github.com/angular/material/blob/875848817bf9820f510647d2b01a347bb4fd8e46/src/components/tabs/js/tabsDummyWrapperDirective.js#L20)

**Angular Versions**: *
 -  `Angular Version: 1.5.8`
 -  `Angular Material Version: 1.1.0`

**Additional Information**:
 -  `Browser Type: ie`
 -  `Browser Version: 10`
 -  `OS: windows 7`
 -  `Stack Traces:`

naveedahmed1:
Can someone please check if the issue still exists? I am receiving below error in my app when running it on IE 10:

```
EXCEPTION: Uncaught (in promise): ReferenceError: 'MutationObserver' is undefined
ReferenceError: 'MutationObserver' is undefined
```

prinzt:
I can confirm the error on IE10.
- Angular 4.0.0
- Material 2.0.0-beta.2

dylanlawrence:
We also had this with IE10,  I was able to add this polyfill - https://github.com/megawac/MutationObserver.js to resolve it but it seems like this 'observing' could just be handled by Rxjs or something that has accounted for these issues.

PR #9397, opened 2016-08-24
fix(tabs): ie10 MutationObserver issue

j3ski:
Check if MutationObserver is in window. If not fall back to DOMSubtreeModified event.

Closes: [9395](https://github.com/angular/material/issues/9395)

### DIFF
--- a/src/components/tabs/js/tabsDummyWrapperDirective.js
+++ b/src/components/tabs/js/tabsDummyWrapperDirective.js
@@ -40,7 +40,8 @@ function MdTabsDummyWrapper ($mdUtil, $window) {
         observer.observe(element[0], config);
         disconnect = observer.disconnect.bind(observer);
       } else {
-        disconnect = element.on('DOMSubtreeModified', mutationCallback);
+        element.on('DOMSubtreeModified', mutationCallback);
+        disconnect = element.off.bind(element, 'DOMSubtreeModified', mutationCallback);
       }
 
       // Disconnect the observer
--- a/src/components/tabs/js/tabsDummyWrapperDirective.js
+++ b/src/components/tabs/js/tabsDummyWrapperDirective.js
@@ -40,8 +40,10 @@ function MdTabsDummyWrapper ($mdUtil, $window) {
         observer.observe(element[0], config);
         disconnect = observer.disconnect.bind(observer);
       } else {
-        element.on('DOMSubtreeModified', mutationCallback);
-        disconnect = element.off.bind(element, 'DOMSubtreeModified', mutationCallback);
+        var debounced = $mdUtil.debounce(mutationCallback, 10);
+        
+        element.on('DOMSubtreeModified', debounced);
+        disconnect = element.off.bind(element, 'DOMSubtreeModified', debounced);
       }
 
       // Disconnect the observer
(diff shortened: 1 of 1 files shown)


j3ski:
I signed it!

src/components/tabs/js/tabsDummyWrapperDirective.js
```diff
@@ -6,38 +6,47 @@ angular
...
+          // the childList mutation, once a <span> element's innerText changes.
+          // The characterData of the <span> element will change.
+          characterData: true
+        };
+
+        observer = new MutationObserver(mutationCallback);
+        observer.observe(element[0], config);
+        disconnect = observer.disconnect.bind(observer);
+      } else {
+        element.on('DOMSubtreeModified', mutationCallback);
```
crisbeto: @j3ski I haven't tried this out yet, but I believe that `DOMSubtreeModified` is deprecated and fires **very** often. It might be a good idea to see if this can either be switched to an Angular watcher that "only" fires once per digest, or look into debouncing the callback (You can use the `$mdUtil.debounce` method).

j3ski: @crisbeto Yes, `DOMSubtreeModified` is deprecated, but this is a fix for browsers which don't yet support `MutationObserver`. I re-implemented the functionality removed [here](https://github.com/angular/material/commit/77958a173d38aa321d9929dd8071e6ac1ef0c84c). I don't think this can be switched to an Angular watcher, as the MutationObserver watches for subtree modifications, not attributes, but debouncing sounds like a good idea. Initially the `DOMSubtreeModified` would fire up to 9x more often than `MutationObserver` in my case. Debounced it fires roughly the same amount of times as `MutationObserver` (provided MO did fire), however it fires on more occasions.

src/components/tabs/js/tabsDummyWrapperDirective.js
```diff
@@ -6,38 +6,49 @@ angular
...
+          // the childList mutation, once a <span> element's innerText changes.
+          // The characterData of the <span> element will change.
+          characterData: true
+        };
+
+        observer = new MutationObserver(mutationCallback);
+        observer.observe(element[0], config);
+        disconnect = observer.disconnect.bind(observer);
+      } else {
+        var debounced = $mdUtil.debounce(mutationCallback, 10);
```
crisbeto: This looks good, although could you change the `debounce` signature to `$mdUtil.debounce(mutationCallback, 15, null, false)`. This is because a couple of reasons:

1. Passing in `false` as the last param means that it won't trigger a digest every time it gets invoked, which should help with the performance.
2. Bumping the time is more of a nitpick, but I feel that it's something we can get away with since it's still less than 1 frame (which is ~16ms).

j3ski: Done.
Checked it, reduced the number of calls to mutationCallback by 1 in one scenario. Also the app seems to work smoother when loading the tabs, but it can be just my impression.

crisbeto: Great, I'll test it out manually a bit later.

crisbeto:
Code and functionality-wise LGTM @ThomasBurleson.

PR #9410, opened 2016-08-25
feat(datepicker): add timezone support

### LINKED ISSUE #8448: feat(datePicker): UTC mode
Right now it assumes all dates are local time. This would be a mode/flag to make it assume all dates are UTC time

### LINKED ISSUE #8936: md-datepicker sending the day before the one that I select
Hello 

I've seen closed issue https://github.com/angular/material/issues/4636 named « md-datepicker sending the day before the one that I select ». 

 @jadjoubran, @jelbourn and @altsel, maybe I have an other solution closer of https://material.angularjs.org/1.0.9/api/service/$mdDateLocaleProvider. Maybe it's better than @altsel solution (11 Nov 2015) , because it uses `$mdDateLocaleProvider.formatDate`
```
        $mdDateLocaleProvider.formatDate = function(date) {
            var m = moment(date);
            if (!m.isValid()) {
                return '';
            }
            var dateView = moment(date).format('X'), // return unix timestamp of `date' argument
            dateCurrentScript = moment().format('X'); // return current unix timestamp
            if (dateView + 1 > dateCurrentScript )
                return '';
            else
                return m.format('L');
        };
```

What do you think about that (maybe my code is totally stupid, (sorry)) ? 

If you think this solution is good, maybe we can add it in https://github.com/angular/material/issues/4636 ? But for that, we must ask to @ThomasBurleson or an other AngularMaterial collaborator to  unlock  https://github.com/angular/material/issues/4636 conversation.

Julio

crisbeto:
Adds the ability to specify the datepicker's timezone via the `ng-model-options` attribute, e.g. for UTC, the element would look like this:

```
<md-datepicker ng-model="date" ng-model-options="{ timezone: 'utc' }"></md-datepicker>
```

This works by re-using the logic from Angular's `input[type="date"]` directive.

Fixes #8448.
Fixes #8936.

### DIFF
--- a/src/components/datepicker/js/datepickerDirective.js
+++ b/src/components/datepicker/js/datepickerDirective.js
@@ -8,7 +8,6 @@
   // TODO(jelbourn): forward more attributes to the internal input (required, autofocus, etc.)
   // TODO(jelbourn): something better for mobile (calendar panel takes up entire screen?)
   // TODO(jelbourn): input behavior (masking? auto-complete?)
-  // TODO(jelbourn): UTC mode
 
 
   angular.module('material.components.datepicker')
@@ -20,6 +19,8 @@
    * @module material.components.datepicker
    *
    * @param {Date} ng-model The component's model. Expects a JavaScript Date object.
+   * @param {Object=} ng-model-options Allows tuning of the way in which `ng-model` is being updated. Also allows
+   * for a timezone to be specified. <a href="https://docs.angularjs.org/api/ng/directive/ngModelOptions#usage">Read more at the ngModelOptions docs.</a>
    * @param {expression=} ng-change Expression evaluated when the model value changes.
    * @param {expression=} ng-focus Expression evaluated when the input is focused or the calendar is opened.
    * @param {expression=} ng-blur Expression evaluated when focus is removed from the input or the calendar is closed.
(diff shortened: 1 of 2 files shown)


src/components/datepicker/js/datepickerDirective.js
```diff
@@ -20,6 +19,8 @@
    * @module material.components.datepicker
    *
    * @param {Date} ng-model The component's model. Expects a JavaScript Date object.
+   * @param {Object=} ng-model-options Allows tuning the of way in which `ng-model` is being updated. Also allows
```
DevVersion: Should this be `the way of`?

crisbeto: It should.

ThomasBurleson:
@jelbourn - as the original author, I think it best if you review.

src/components/datepicker/js/datepickerDirective.js
```diff
@@ -365,17 +369,30 @@
...
-  DatePickerCtrl.prototype.configureNgModel = function(ngModelCtrl, mdInputContainer) {
+  DatePickerCtrl.prototype.configureNgModel = function(ngModelCtrl, mdInputContainer, inputDirective) {
     this.ngModelCtrl = ngModelCtrl;
     this.mdInputContainer = mdInputContainer;
 
+    // The input needs to be [type="date"] in order to be picked up by Angular.
+    this.$attrs.$set('type', 'date');
+
+    // Invoke the `input` directive link function, adding a stub for the element.
+    inputDirective[0].link.pre(this.$scope, {
```
ThomasBurleson: Bit of black magic here. Can you comment/explain why this is needed ?

crisbeto: Will do.

crisbeto: Done.

PR #9411, opened 2016-08-25
add mdReverse to mdFabActions

dcbartlett:
feat(fabActions): add md-reverse to mdFabActions

Add your code like this:

    ```html
    <md-fab-actions md-reverse>
    ```

### DIFF
--- a/src/components/fabActions/fabActions.js
+++ b/src/components/fabActions/fabActions.js
@@ -29,11 +29,23 @@
       restrict: 'E',
 
       require: ['^?mdFabSpeedDial', '^?mdFabToolbar'],
-
+  
       compile: function(element, attributes) {
         var children = element.children();
 
         var hasNgRepeat = $mdUtil.prefixer().hasAttribute(children, 'ng-repeat');
+        var hasReverse = attributes.mdReverse || attributes.mdReverse == "";
+
+        // We don't want to be able to use reverse with NgRepeat
+        // You should be reversing it in your controller if you are using NgRepeat
+        if (hasReverse && !hasNgRepeat) {
+          var childrenArr = [].slice.call(children);
+          element.empty();
+          childrenArr.reverse().forEach(function(child){
+            element.append(child);
+          });
+          children = element.children();
+        }
 
         // Support both ng-repeat and static content
         if (hasNgRepeat) {
--- a/src/components/fabActions/fabActions.js
+++ b/src/components/fabActions/fabActions.js
@@ -29,23 +29,15 @@
       restrict: 'E',
 
       require: ['^?mdFabSpeedDial', '^?mdFabToolbar'],
+
+      scope: {
+        reverse: "=mdReverse"
+      },
   
-      compile: function(element, attributes) {
+      compile: function (element, attributes) {
         var children = element.children();
 
         var hasNgRepeat = $mdUtil.prefixer().hasAttribute(children, 'ng-repeat');
-        var hasReverse = attributes.mdReverse || attributes.mdReverse == "";
-
-        // We don't want to be able to use reverse with NgRepeat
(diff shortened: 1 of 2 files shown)


dcbartlett:
Ugh, forgot to add the tests.

dcbartlett:
I also noticed that this wants to merge into master.  I'll gladly resubmit to a different branch, just didn't know where to put this.

topherfangio:
Master is definitely the correct branch 😃 

However, I am a bit unclear as to why this is useful. Especially if it is not a boolean expression that gets watched.

If you want it reversed all the time, and it doesn't work with `ng-repeat` (as evidenced by your comment in the code), why not just reverse the items in the HTML?

> We don't want to be able to use reverse with NgRepeat

Can you clarify the use-case a bit for me?

dcbartlett:
Well,  Semantically, the HTML should be structured in the way that it should be read if the menu is left-to-right.  However, if your menu comes from the right (or bottom) you would want the menu items reversed for presentation.

You're probably right in that it should monitor the the attribute for a Boolean change.  Changing order when the Boolean changes would allow for us to use this to change both direction and order based on screen size.  Desktop could use a right hand menu (opens to the left) with a reversed order, and mobile/tablet could use a top-down (opens to the bottom) with a standard order.

I'll update this to support the Boolean watching.

dcbartlett:
@topherfangio should this support passing in the attribute without a variable?  Or should it only work if a true/false value is passed in?  I'm not 100% sure if supporting an attribute by assuming true is what is preferred in angular material.

for example:
```html
<md-fab-actions md-reverse>
```

If we do support this, anyone using the feature without passing a value directly, would not be able to update the value at a later time (because they didn't bind anything).  So i'm not going to support that by default unless we should.

dcbartlett:
UGH....stupid isolated scope...I'll push a fix for that.  I'm going to squash all this once i get it nailed out i spose.

topherfangio:
Can you elaborate on the following statement a bit? I'm not 100% sure I understand.

> However, if your menu comes from the right (or bottom) you would want the menu items reversed for presentation.

Can you maybe make a Codepen with a before/after scenario so I can see it better? Particularly, why would you want them "reversed for presentation"?

dcbartlett:
Sure thing,

This codepen should display what i'm going for a bit easier. (this is running my code but doesn't have a fix for the scope issue yet).
http://codepen.io/anon/pen/yJmdgb

topherfangio:
Thanks for the Codepen. After seeing it, though, I think I disagree at a fundamental level that we should be doing that. I think that the HTML should match the FAB visually, but rather than thinking left-to-right, you should be thinking "from the fab outward".

Particularly, I think this is important from an accessibility standpoint. When you tab through the options, they should start at the FAB and work outward from there.

Based on the Codepen, I definitely see the example of how it would work. I'm just trying to figure out in what scenario you would want to do that. If you disagree with my statements above, can you show me some screenshots (or just describe) your use-case for why this makes sense?

I'm not 100% against this if it makes sense, I just feel like I'm missing something here 😄

dcbartlett:
My thought on it was, i'm using conditionals to wrap my md-buttons so that they can be used both in md-toolbar (they should be left-to-right there) and in md-fab-speed-dial (they need to be reveres from a left opening fab).  I wanted to be able to re-use the html without needing to have to copies to maintain.  I suppose i could keep two copies as its not that much maintenance, but thought there may be more uses for a reverse on the md-fab-actions.

I do see your point of it being semantic from the fab - out.  I wasn't looking at it from that stand point.  I'm on the fence on this one and could go either way myself, so at this point, i'd have to wonder if the rest of the community sees a benefit.

Thanks for your time and feedback.  I'll probably have more commits soon (found a few bugs i want to squash).

ThomasBurleson:
Will not fix/merge.

PR #9412, opened 2016-08-25
fix(layout): Revert overzealous IE11 flexbox fix.

### LINKED ISSUE #9354: Demo in layout tips no longer fixes Chrome issue.
**Actual Behavior**:
 -  `What is the issue? *` The second demo which shows the chrome flex/height bug fixed is not working.
 -  `What is the expected behavior?` The demo should properly show each container (not the toolbar) at 50% of the remaining available space.

 -  `Details:` https://material.angularjs.org/latest/layout/tips

**Angular Versions**: *
 -  `Angular Version:` N/A

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *` 52.0.2743.116 m
 -  `OS: *` Windows
 -  `Stack Traces:`

----

  Add more lines if needed.

topherfangio:
A recent change which was supposed to fix an IE11 bug accidentally altered the `flex-basis` for the flex children of all `.layout-column` flex containers in all browsers.

Unfortunately, there is no IE11-specific fix.

 - Revert the original change.
 - Update the Layout Troubleshooting page with a new demo.
 - Update the Layout Children docs to mention the new demo.
 - Remove old Chrome-specific demo which was incidentally fixed
   by reverting this change.

Reverts ce0ebbf.

Fixes #9354.

### DIFF
--- a/docs/app/partials/layout-tips.tmpl.html
+++ b/docs/app/partials/layout-tips.tmpl.html
@@ -108,97 +108,91 @@
   <br/>
   <hr/>
 
-  <h3>Layout Column and Container Heights</h3>
+  <h3 id="layout-column-0px-ie11">IE11 - Layout Column, 0px Height</h3>
 
   <p>
-    In Flexbox, some browsers will determine size of the flex containers based on the size of their
-    content. This is particularly noticable if you have a non-flex item (say a toolbar), followed by
-    two flex items in a column layout.
+    In Internet Explorer 11, when you have a column layout with unspecified height and flex items
+    inside, the browser can get confused and incorrectly calculate the height of each item (and thus
+    the container) as <code>0px</code>, making the items overlap and not take up the proper amount
+    of space.
   </p>
 
-  <docs-demo demo-title="Flex Height - Odd (Chrome )" class="small-demo colorNested">
+  <p class="layout_note">
+    <b>Note:</b> The flex items below actually do have some height. This is because our doc-specific
+    CSS provides a small bit of padding (<code>8px</code>). We felt that the extra padding made for
+    a better demo of the actual issue.
+  </p>
+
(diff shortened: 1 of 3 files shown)


docs/app/partials/layout-tips.tmpl.html
```diff
@@ -108,98 +108,92 @@
...
 
-  <h3>Layout Column and Container Heights</h3>
+  <h3 id="layout-column-0px-ie11">IE11 - Layout Column, 0px Height</h3>
 
   <p>
-    In Flexbox, some browsers will determine size of the flex containers based on the size of their
-    content. This is particularly noticable if you have a non-flex item (say a toolbar), followed by
-    two flex items in a column layout.
+    In Internet Explorer 11, when you have a column layout with unspecified height and flex items
+    inside, the browser can get confused and incorrectly calculate the height of each item (and thus
```
DevVersion: Should that be `and incorrectly calculates`?

DevVersion:
That looks very good to me. Especially how everything is explained with an associated demo.

ThomasBurleson:
@topherfangio - lgtm, nice work here sir!

docs/app/partials/layout-tips.tmpl.html
topherfangio: #### Random English Silliness

I honestly think either would work here, but how I judge it is this: remove the "get confused and" phrase to see how it sounds:

> the browser can incorrectly calculate the height...

verses

> the browser can incorrectly calculates the height...

Honestly, seeing as I've never seen it **not** get confused, it might be more correct to just say

> the browser gets confused and incorrectly calculates the height...

Stupid english 😝

DevVersion: lol, thanks for the explanation :smile: Keep whatever you think. You are the native speaker!

PR #9413, opened 2016-08-26
feat(themes): register theme on the fly

### LINKED ISSUE #2965: $mdThemingProvider - registering a theme on the fly
Hi there!
First of all, I love Angular Material. I think you guys are doing an excellent job. Keep it up.

Secondly, and the reason I'm starting this issue is because I want my users to be able to register a new theme by adding 3 colour palletes together. I would like to search that in the database... so I will have active just one theme but I would need it to work with an ajax function...
The question is... is there a way to do that? Will it? Could you add a resolve method just like the routes?

If you want to see what I would be using it for... just check this out... https://github.com/rgolea/angular-deployd-admin/commit/3dd2a6f922eda5b1ec3befa30d3852fd4bc91756

I'm making an angular-material dashboard (free and open source for anyone who may need it). But in order to take full advantage of the angular-material capabilities I would need to do that option.

Thank you very much.

EladBezalel:
- Added `defineTheme` functionality to `$mdTheming` service
- Added support for promise and a function that returns a promise on `md-theme` directive to support async build of themes

fixes #2965

### DIFF
--- a/src/core/services/theming/theming.js
+++ b/src/core/services/theming/theming.js
@@ -538,7 +538,7 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
    * @param {el=} element to apply theming to
    */
   /* @ngInject */
-  function ThemingService($rootScope, $log) {
+  function ThemingService($rootScope, $mdUtil, $q, $log) {
         // Allow us to be invoked via a linking function signature.
     var applyTheme = function (scope, el) {
           if (el === undefined) { el = scope; scope = undefined; }
@@ -546,12 +546,45 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
           applyTheme.inherit(el, el);
         };
 
-    applyTheme.THEMES = angular.extend({}, THEMES);
-    applyTheme.PALETTES = angular.extend({}, PALETTES);
+    Object.defineProperty(applyTheme, 'THEMES', {
+      get: function () {
+        return angular.extend({}, THEMES);
+      }
+    });
+    Object.defineProperty(applyTheme, 'PALETTES', {
+      get: function () {
+        return angular.extend({}, PALETTES);
+      }
+    });
     applyTheme.inherit = inheritTheme;
     applyTheme.registered = registered;
     applyTheme.defaultTheme = function() { return defaultTheme; };
     applyTheme.generateTheme = function(name) { generateTheme(THEMES[name], name, themeConfig.nonce); };
+    applyTheme.defineTheme = function(name, options) {
+      options = options || {};
+
+      var theme = registerTheme(name);
+
+      if (options.primary) {
+        theme.primaryPalette(options.primary);
+      }
+      if (options.accent) {
+        theme.accentPalette(options.accent);
+      }
+      if (options.warn) {
+        theme.warnPalette(options.warn);
+      }
+      if (options.background) {
+        theme.backgroundPalette(options.background);
+      }
+      if (options.dark){
+        theme.dark();
+      }
+
+      this.generateTheme(name);
+
+      return $q.resolve(name);
+    };
     applyTheme.setBrowserColor = enableBrowserColor;
 
     return applyTheme;
@@ -568,14 +601,24 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
      * Get theme name for the element, then update with Theme CSS class
      */
     function inheritTheme (el, parent) {
-      var ctrl = parent.controller('mdTheme');
-      var attrThemeValue = el.attr('md-theme-watch');
-      var watchTheme = (alwaysWatchTheme || angular.isDefined(attrThemeValue)) && attrThemeValue != 'false';
+      var ctrl = parent.controller('mdTheme') || el.data('$mdThemeController');
 
       updateThemeClass(lookupThemeName());
 
-      if ((alwaysWatchTheme && !registerChangeCallback()) || (!alwaysWatchTheme && watchTheme)) {
-        el.on('$destroy', $rootScope.$watch(lookupThemeName, updateThemeClass) );
+      if (ctrl) {
+        var watchTheme =
+          alwaysWatchTheme || ctrl.$shouldWatch || $mdUtil.parseAttributeBoolean(el.attr('md-theme-watch'));
+
+        var unwatch = ctrl.registerChanges(function (name) {
+          updateThemeClass(name);
+
+          if (!watchTheme) {
+            unwatch();
+          }
+          else {
+            el.on('$destroy', unwatch);
+          }
+        });
       }
 
       /**
@@ -583,7 +626,6 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
        */
       function lookupThemeName() {
         // As a few components (dialog) add their controllers later, we should also watch for a controller init.
-        ctrl = parent.controller('mdTheme') || el.data('$mdThemeController');
         return ctrl && ctrl.$mdTheme || (defaultTheme == 'default' ? '' : defaultTheme);
       }
 
@@ -606,24 +648,12 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
           el.data('$mdThemeController', ctrl);
         }
       }
-
-      /**
-       * Register change callback with parent mdTheme controller
-       */
-      function registerChangeCallback() {
-        var parentController = parent.controller('mdTheme');
-        if (!parentController) return false;
-        el.on('$destroy', parentController.registerChanges( function() {
-          updateThemeClass(lookupThemeName());
-        }));
-        return true;
-      }
     }
 
   }
 }
 
-function ThemingDirective($mdTheming, $interpolate, $log) {
+function ThemingDirective($mdTheming, $interpolate, $parse, $mdUtil, $q, $log) {
   return {
     priority: 100,
     link: {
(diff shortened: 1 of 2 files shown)


EladBezalel:
@topherfangio we should find a better approach

src/core/services/theming/theming.spec.js
```diff
@@ -787,6 +798,38 @@ describe('md-theme directive', function() {
...
+    var ctrl = el.data('$mdThemeController');
+    expect(ctrl.$mdTheme).toBe('unregistered');
+    $rootScope.$apply('themey = "blue"');
+    expect(ctrl.$mdTheme).toBe('blue');
+  }));
+
+  it('does not warns when an unregistered theme is used', function() {
+    inject(function($log, $compile, $rootScope) {
+      spyOn($log, 'warn');
+      $compile('<div md-deferred-theme="unregistered"></div>')($rootScope);
```
ErinCoughlan: You actually would want this to warn because the theme isn't registered. This approach loses that information.

ErinCoughlan:
@EladBezalel I have another idea. Instead of using a string in md-theme, can you use an object, at least in the deferred class? That way we know if it will be created since the object needs to come from somewhere.

src/core/services/theming/theming.js
```diff
@@ -649,16 +678,44 @@ function ThemingDirective($mdTheming, $interpolate, $log) {
...
+
             ctrl.$mdTheme = theme;
 
-            registeredCallbacks.forEach(function (cb) {
-              cb();
-            });
-          }
+            // Iterating backwards to support unregistering during iteration
+            // http://stackoverflow.com/a/9882349/890293
+            for (var i = registeredCallbacks.length - 1; i >= 0; i -= 1) {
```
DevVersion: Doesn't the following work?
```js
registeredCallbacks.reverse().forEach(function(callback) {
```

Note: `forEach` keeps the right order.

EladBezalel: Maybe :sweat_smile: i'll check

src/core/services/theming/theming.js
```diff
@@ -538,20 +538,53 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
...
         // Allow us to be invoked via a linking function signature.
     var applyTheme = function (scope, el) {
           if (el === undefined) { el = scope; scope = undefined; }
           if (scope === undefined) { scope = $rootScope; }
           applyTheme.inherit(el, el);
         };
 
-    applyTheme.THEMES = angular.extend({}, THEMES);
-    applyTheme.PALETTES = angular.extend({}, PALETTES);
+    Object.defineProperty(applyTheme, 'THEMES', {
```
ThomasBurleson: Why this complexity... unless you want generate the THEMES object on demand (each time the getter is accessed).

EladBezalel: Yes, generating it once doesn't support themes that are registered on the fly

src/core/services/theming/theming.js
```diff
@@ -649,16 +678,44 @@ function ThemingDirective($mdTheming, $interpolate, $log) {
...
+        var getThemeInterpolation = function () { return $interpolate(attrs.mdTheme)(scope); };
+
+        var setParsedTheme = function (interpolation) {
+          var theme = $parse(interpolation)(scope) || interpolation;
+
+          if (typeof theme === 'string') {
+            ctrl.$setTheme(theme);
+          }
+
+          var promise = $q.resolve(theme);
```
ThomasBurleson: use ternary here plz:

```js
var promise = $q.when( (typeof theme === 'function') ?  theme() : theme );
```

EladBezalel: Totally

topherfangio:
Looks pretty interesting. Any usage docs so that we can play with it and test it out? :smiley:

EladBezalel:
Actually I was testing on button demo,  had a timeout to resolve a new defined theme after 2 seconds

topherfangio:
@EladBezalel Ok. Can you share that code at least so that we can do some testing on our local machines?

Also, this will need docs before we merge it, right?

EladBezalel:
There are no docs for the theming service at all.. I wanted to add it but it requires even more work on that.. 
I can fix the comments and work on another PR for documentation to mdTheming service

EladBezalel:
Run `gulp watch site --dev` and go to 
http://codepen.io/EladBezalel/pen/YGKBYE
You can also remove md-theme-watch to see how it's not being changed

src/core/services/theming/theming.js
```diff
@@ -568,22 +601,30 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
...
-      var attrThemeValue = el.attr('md-theme-watch');
-      var watchTheme = (alwaysWatchTheme || angular.isDefined(attrThemeValue)) && attrThemeValue != 'false';
+      var ctrl = parent.controller('mdTheme') || el.data('$mdThemeController');
 
       updateThemeClass(lookupThemeName());
 
-      if ((alwaysWatchTheme && !registerChangeCallback()) || (!alwaysWatchTheme && watchTheme)) {
-        el.on('$destroy', $rootScope.$watch(lookupThemeName, updateThemeClass) );
+      if (ctrl) {
+        var watchTheme = alwaysWatchTheme || ctrl.$shouldWatch || $mdUtil.parseAttributeBoolean(el.attr('md-theme-watch'));
```
ErinCoughlan: nit: line too long here :)

src/core/services/theming/theming.js
```diff
@@ -538,20 +538,53 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
...
+    Object.defineProperty(applyTheme, 'PALETTES', {
+      get: function () {
+        return angular.extend({}, PALETTES);
+      }
+    });
     applyTheme.inherit = inheritTheme;
     applyTheme.registered = registered;
     applyTheme.defaultTheme = function() { return defaultTheme; };
     applyTheme.generateTheme = function(name) { generateTheme(THEMES[name], name, themeConfig.nonce); };
+    applyTheme.defineTheme = function (name, options) {
```
ErinCoughlan: nit: Why the extra space after "function"? This seems strange. It's inconsistent throughout... :(

I dislike inconsistency.

src/core/services/theming/theming.spec.js
```diff
@@ -785,6 +797,40 @@ describe('md-theme directive', function() {
       expect($log.warn.calls.count()).toBe(0);
     });
   });
+
+  it('should accept string as a theme', function() {
+    inject(function($compile, $rootScope, $mdTheming) {
```
ErinCoughlan: More inconsistency here. I'm not going to argue about the inject part, but please be consistent with this vs line 773. I don't really care which you pick as long as the whole file is the same.

src/core/services/theming/theming.spec.js
```diff
@@ -785,6 +797,40 @@ describe('md-theme directive', function() {
...
+      $rootScope.$apply();
+      $mdTheming(el);
+      expect(el.hasClass('md-default-theme')).toBeFalsy();
+      expect(el.hasClass('md-red-theme')).toBeTruthy();
+    });
+  });
+
+  it('should accept $q promise as a theme', function() {
+    inject(function($compile, $rootScope, $q, $mdTheming) {
+      $rootScope.promise = $q.resolve('red');
```
ErinCoughlan: So I thought there was a method on `$mdTheming` that returns a promise (which eventually is the resolve). To make this test more robust and closer to real use, can you use that method instead of just resolve with a string.

src/core/services/theming/theming.spec.js
```diff
@@ -453,6 +453,7 @@ describe('$mdThemeProvider with custom styles', function() {
       // Verify that $MD_THEME_CSS is still set to '/**/' in the test environment.
       // Check angular-material-mocks.js for $MD_THEME_CSS latest value if this test starts to fail.
       expect($MD_THEME_CSS).toBe('/**/');
+      expect($MD_THEME_CSS).toBe('/**/');
```
ErinCoughlan: duplicate line?

src/core/services/theming/theming.spec.js
```diff
@@ -768,23 +780,47 @@ describe('md-theme directive', function() {
...
+  it('should accept $q promise as a theme', inject(function($compile, $rootScope, $q, $mdTheming) {
+    $rootScope.promise = $mdTheming.defineTheme('red', { primary: 'red' });
+    var el = $compile('<div md-theme="promise"></div>')($rootScope);
+    $mdTheming(el);
+    $rootScope.$apply();
+    expect(el.hasClass('md-default-theme')).toBeFalsy();
+    expect(el.hasClass('md-red-theme')).toBeTruthy();
+  }));
+
+  it('should accept a function that returns a promise as a theme', inject(function($compile, $rootScope, $q, $mdTheming) {
```
ErinCoughlan: nit: line too long

ErinCoughlan:
LGTM

src/core/services/theming/theming.js
```diff
@@ -649,16 +679,39 @@ function ThemingDirective($mdTheming, $interpolate, $log) {
...
               $log.warn('attempted to use unregistered theme \'' + theme + '\'');
             }
+
+
             ctrl.$mdTheme = theme;
 
-            registeredCallbacks.forEach(function (cb) {
-              cb();
+            // Iterating backwards to support unregistering during iteration
+            // http://stackoverflow.com/a/9882349/890293
```
DevVersion: Isn't that comment now invalid? since we are now using a simple `reverse` ?

EladBezalel: It's still explaining why are we iterating backwards

DevVersion: The URL does in my case only show how to look backwards. But does not explain why we do this.

EladBezalel: >The array is being re-indexed when you do a .splice(), which means you'll skip over an index when one is removed, and your cached .length is obsolete.

DevVersion: It's not a big deal. But where do we explicitly use a `splice` here? No splice and no `.length` anymore.

src/core/services/theming/theming.js
hansl: Reading this code, the following assertion happen, which seems like the wrong logic:

```javascript
applyTheme.THEMES !== applyTheme.THEMES
```

PR #9417, opened 2016-08-26
fix(menu-bar): use checked icon from $$mdSvgRegistry

### LINKED ISSUE #9407: In md-menu-item type="checkbox" the checkbox is not visible
My md-menu-item type="checkbox" never shows the actual checkbox. It seems to work in the Demo (https://material.angularjs.org/latest/demo/menuBar), but doesn't when I go to the associated CodePen.
I can see md-icon md-svg-icon="check". In demo it's somehow populated from core-icons.svg, but  empty in my case.
I'm using v.1.1.0. Tried it with with the latest AngularJS and 1.5.5 you are using.
Please help.

DevVersion:
* Using the icon from the `$$mdSvgRegistry` ensures that the icon is properly shipped to the developer, without having them to include the Material Icons library.

Fixes #9407.


ThomasBurleson:
@DevVersion - lgtm

PR #9418, opened 2016-08-26
fix(demos): update core-icons svg in assets cache to latest changes.

DevVersion:
* It seems like the `core-icons.svg` set has been updated, but not in the assets cache, which means that the required icons cannot show up in the codepens.

@ThomasBurleson This also requires you to update the `svg-assets-cache.js` in the AWS.

References #9407.


ThomasBurleson:
@DevVersion - lgtm

PR #9428, opened 2016-08-27
fix(menu-bar): unable to close menu when clicking on toolbar

### LINKED ISSUE #8965: Toolbar menu items doesn't close on click
When a `<md-menu-content>` inside a `<md-menu>` on a `<md-toolbar>` is opened. It can't be closed by clicking on the md-toolbar.

This seems because some component has added to the toolbar styles the rule z-index:100 which produces this element to be over the md-scroll-mask which has a z-index of 50, resulting on this mask can never be clicked on the part of the layout where the md-toolbar stands.

**Angular Versions**: *
 -  `Angular Version: 1.5` 
 -  `Angular Material Version: 1.1.0-rc4`

**Additional Information**:
 -  `Browser Type: Chrome, Safary`
 -  `Browser Version: 51.0.2704.106`
 -  `OS: Linux and MacOS`

crisbeto:
fix(menu-bar): unable to close menu when clicking on toolbar

* Fixes not being able to close a toolbar menu by clicking on the wrapping toolbar. Works by adding an extra click handler on the toolbar that closes the menu. Note that the same effect could be achieved by removing the `z-index` altogether, but that could break the functionality that allows users to jump from one menu to the next without clicking.
* Moves the menu bar's z-index to CSS so it's easier to manage.
* Adds call to `disableOpenOnHover` when the scope is destroyed, in order to clean up leftover event handlers.

Fixes #8965.

Cheers to @topherfangio for catching a potentially breaking error.

### DIFF
--- a/src/components/menuBar/js/menuBarController.js
+++ b/src/components/menuBar/js/menuBarController.js
@@ -66,6 +66,7 @@ MenuBarController.prototype.init = function() {
   }));
 
   $scope.$on('$destroy', function() {
+    self.disableOpenOnHover();
     while (deregisterFns.length) {
       deregisterFns.shift()();
     }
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
@crisbeto - my impression is that Ryan wanted the parent with a menubar to float above everything else in the UI layers; consider the classic menubar usage where it has should priority access above all except modals and tooltips.

src/components/menuBar/js/menuBarController.js
```diff
@@ -84,11 +84,11 @@ MenuBarController.prototype.enableOpenOnHover = function() {
...
   this.openOnHoverEnabled = true;
 
-  var parentToolbar;
-  if (parentToolbar = this.parentToolbar) {
+  var parentToolbar = this.parentToolbar;
+  if (parentToolbar) {
     parentToolbar.dataset.mdRestoreStyle = parentToolbar.getAttribute('style');
     parentToolbar.style.position = 'relative';
-    parentToolbar.style.zIndex = 100;
+    parentToolbar.style.zIndex = 98; // puts it directly below the backdrop
```
ThomasBurleson: Since we set the `z-index` in the SCSS for our other components, let's do the same here and simply add a classname that updates the position/z-index.

crisbeto: Sounds good.

crisbeto:
I'm not sure that was the point @ThomasBurleson. This `z-index` only gets set when the menu is open and then gets cleared when it closes.

topherfangio:
@crisbeto That code exists to ensure that you can open a menu inside a menu bar and then hover over the menus next to it.

This change will currently break the MenuBar demo.

crisbeto:
I could reproduce it as well. I'll try a suggestion from @topherfangio to listen for click events on the toolbar.

crisbeto:
Updated to keep the old behaviour while fixing the issue when clicking on the toolbar.

topherfangio:
LGTM 👍

Issue #9342, opened 2016-08-18
mdDatepicker: mdInputContainer ngMessages unaligned with other input types


mckenzielong:
**Actual Behavior**:
 -  `What is the issue? *` mdDatepicker + mdInputContainer has inconsistent ngMessages placement when compared to other inputs.
 -  `What is the expected behavior?` The error messages div is positioned in line with the other containers

**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:` http://codepen.io/mckenzielong/pen/bZJdrq
 -  `Details:` Click each input to trigger the required error. Note the datepicker's error div is positioned slightly lower than the others.

**Angular Versions**: *
 -  `Angular Version:` 1.5.8
 -  `Angular Material Version:` 1.1.0

**Additional Information**:
 -  `Browser Type: *` Opera / Chrome
 -  `Browser Version: *` 39 /52
 -  `OS: *` Windows 7 64bit
 -  `Stack Traces:` N/A

PR #9432, opened 2016-08-27
fix(datepicker): improved overlay positioning

crisbeto:
The current approach for positioning the datepicker's calendar works by expanding the input container in order to fill a gap in the floating calendar pane and using negative margins to prevent the UI from jumping around. This comes with some drawbacks:
* It causes the input element to shift down slightly when the calendar gets opened.
* It can shift all of the elements surrounding the datepicker, despite the negative margins.
* It makes it hard to refactor the positioning.

This new approach solves all of the above-mentioned issues by using an element with a box-shadow to create the overlay, avoiding the need for expanding the container.

Referencing #9342.

#### Current approach:
current

#### New approach:
new

### DIFF
--- a/src/components/datepicker/datePicker-theme.scss
+++ b/src/components/datepicker/datePicker-theme.scss
@@ -60,12 +60,18 @@
     }
   }
 
-  .md-datepicker-open .md-datepicker-input-container,
-  .md-datepicker-input-mask-opaque {
-    background: '{{background-hue-1}}';
-  }
-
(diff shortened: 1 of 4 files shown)


ThomasBurleson:
@jelbourn - can you review also plz. LGTM

ThomasBurleson:
@crisbeto  - will need to rebase after PR #9410

crisbeto:
Alright, I'll rebase when one of them gets merged.

PR #9436, opened 2016-08-28
fix(panel): Panel issue after minification.

### LINKED ISSUE #8948: mdPanel issue after minification.
**Actual Behavior**:
 -  `What is the issue?` mdPanel stops working after minfying code. I have a grunt task that minifies my files and groups bower libraries into single js files. Before this process mdPanel works fine but afterwards i get an error in the vendor.js file. I use mdDialog (and many other material services) in my code and they work fine after this process. 

I get the following error: [$injector:unpr] Unknown provider: aProvider <- a (in my vendor.js file)

The error occurs here: (in the function... f.apply(b, a)
`    this.$get = ["$window", function(c) {
            function e(a) {
                return a instanceof Error && (a.stack ? a = a.message && -1 === a.stack.indexOf(a.message) ? "Error: " + a.message + "\n" + a.stack : a.stack : a.sourceURL && (a = a.message + "\n" + a.sourceURL + ":" + a.line)),
                a
            }
            function f(a) {
                var b = c.console || {}
                  , f = b[a] || b.log || n
                  , g = !1;
                try {
                    g = !!f.apply
                } catch (h) {}
                return g ? function() {
                    var a = [];
                    return d(arguments, function(b) {
                        a.push(e(b))
                    }),
                    f.apply(b, a)
                }
                : function(a, b) {
                    f(a, null == b ? "" : b)
                }
            }`

 -  `What is the expected behavior?` mdPanel should work correctly after minifying / grouping.

 -  `Details:`

**Angular Versions**: *
 -  `Angular Version:`1.5
 -  `Angular Material Version:`1.1.0 RC5

**Additional Information**:
 -  `Browser Type: *`Chrome

----

  Add more lines if needed.

bradrich:
Panel API now works properly after minification within a standard build process.

Fixes #8948 

Ping @DevVersion, @ErinCoughlan

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -965,7 +965,7 @@ MdPanelRef.prototype.attach = function() {
   var self = this;
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
-    var onDomAdded = self.config['onDomAdded'] || angular.noop;
+    var onDomAdded = self.config.onDomAdded || angular.noop;
     var addListeners = function(response) {
         self.isAttached = true;
         self._addEventListeners();
@@ -995,7 +995,7 @@ MdPanelRef.prototype.detach = function() {
   }
 
   var self = this;
-  var onDomRemoved = self.config['onDomRemoved'] || angular.noop;
+  var onDomRemoved = self.config.onDomRemoved || angular.noop;
 
   var detachFn = function() {
     self._removeEventListeners();
@@ -1066,7 +1066,7 @@ MdPanelRef.prototype.show = function() {
 
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
-    var onOpenComplete = self.config['onOpenComplete'] || angular.noop;
+    var onOpenComplete = self.config.onOpenComplete || angular.noop;
 
     self._$q.all([
       self._backdropRef ? self._backdropRef.show() : self,
@@ -1098,10 +1098,10 @@ MdPanelRef.prototype.hide = function() {
 
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
-    var onRemoving = self.config['onRemoving'] || angular.noop;
+    var onRemoving = self.config.onRemoving || angular.noop;
 
     var focusOnOrigin = function() {
-      var origin = self.config['origin'];
+      var origin = self.config.origin;
       if (origin) {
         getElement(origin).focus();
       }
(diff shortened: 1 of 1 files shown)


bradrich:
@DevVersion, @ErinCoughlan:  I need some help testing this. How do I `bower install` this codebase in order to test minification? Thomas seemed to think that the issue was with the usage of references such as `self.config['scope']` instead of `self.config.scope`. There were several littered throughout the entire document, that I removed. I reran the Karma tests and everything looked good, but I am afraid that I don't know how to pull in only this codebase to test.

Thanks for your help on this.

src/components/panel/panel.js
```diff
@@ -965,7 +965,7 @@ MdPanelRef.prototype.attach = function() {
   var self = this;
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
-    var onDomAdded = self.config['onDomAdded'] || angular.noop;
+    var onDomAdded = self.config.onDomAdded || angular.noop;
```
ThomasBurleson: This is not the best approach... Instead use the constructor

```js
function MdPanelRef(config, $injector) {
   this.config = this.validateConfig(config);
}

MdPanelRef.prototype.validateConfig = function(config) {

config.onDomAdded = config.onDomAdded || angular.noop;
config.onDomRemoved = config.onDomRemoved  || angular.noop;
// etc.

return config;
}
```

ThomasBurleson:
@bradrich - use the [code snippet here](https://github.com/angular/material/issues/8948#issuecomment-230343137) as a starting point.

crisbeto:
@bradrich I'm pretty sure that minifcation won't touch the property names. You can run the docs site locally with the minified files via `gulp watch site`.  

I wasn't able to get the error locally, but I have a few ideas what might be causing it. There are a few places where, instead of injecting the dependencies directly and adding `@ngInject`, only `$injector.get` is being used. These are the cases:

* https://github.com/angular/material/blob/master/src/components/panel/panel.js#L841
* https://github.com/angular/material/blob/master/src/components/panel/panel.js#L1691
* https://github.com/angular/material/blob/master/src/components/panel/panel.js#L2264

DevVersion:
@crisbeto I can agree with that.

Anyways the `MdPanelRef` using the `$injector` shouldn't be a problem, since the variable is properly transferred from the `MdPanelService` to the `MdPanelRef` per constructor.

Also no minification should change the string literals inside of the `$injector` call.

crisbeto:
It won't change the string literal, but it will change the `$injector` being injected as a function argument.

DevVersion:
Yeah those variables can be renamed due minification, but all references to that previous `$injector` variable we be updated as well, so that shouldn't be the problem.

The `$injector` is injected in the `MdPanelService`, which is properly annotated and then passed to the `MdPanelRef` per constructor, so there shouldn't be any issues with that.

bradrich:
@DevVersion You are correct. The `$injector` service is passed in properly and I do not believe that the issue would be with that. 

I don't have any issues with testing the functionality of the above changes, I just don't know how to `bower install` the above changes into a project to test minification.

bradrich:
This PR is no longer valid as #8948 was not an issue with our codebase.

PR #9437, opened 2016-08-28
fix(autocomplete): properly run animation for dialog in demo.

DevVersion:
The autocomplete demo with the dialog currently logs a warning for the `ng-cloak` being present on the dialog template.

  Removing `ng-cloak` removes that warning and also fixes the warning in the console.

### DIFF
--- a/src/components/autocomplete/demoInsideDialog/dialog.tmpl.html
+++ b/src/components/autocomplete/demoInsideDialog/dialog.tmpl.html
@@ -1,4 +1,5 @@
-<md-dialog aria-label="Autocomplete Dialog Example" ng-cloak>
+<md-dialog aria-label="Autocomplete Dialog Example">
+
   <md-toolbar>
     <div class="md-toolbar-tools">
       <h2>Autocomplete Dialog Example</h2>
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@DevVersion - the `ng-cloak` was used to hide Layout Flexbox CSS changes that initial occur when the dialog is shown. Layout API has a ngCloak interceptor...

DevVersion:
@ThomasBurleson - There is actually not explicit use of the Layout engine here, but anyways I added it as same as we did with the following commits.

- b9db09d5f9e59d90245ecc8986ca00a0faa6887e
- 37905c0848c4cba6c05726cbaf43b91390dad979

src/components/autocomplete/demoInsideDialog/dialog.tmpl.html
```diff
@@ -9,7 +10,7 @@
     </div>
   </md-toolbar>
 
-  <md-dialog-content>
+  <md-dialog-content ng-cloak>
```
ThomasBurleson: I think the `ng-cloak` is only useful on the `<md-dialog-content>` wrapper. All others are superfluous.

DevVersion: I can't understand why we even need ngCloak here -Will remove those though..

ThomasBurleson: The cloak hides the UI [at initialization] while more complex FlexBox layouts adjust and resize. The toolbar and actions are very simple and IMO do not require the hide trick. But if you want to be consistent, keep all three (3).

PR #9438, opened 2016-08-28
fix(autocomplete): remove autofocus ambiguity.

DevVersion:
* The autocomplete opens the dropdown when the `md-autofocus` attribute is specified.

* The documentation is not mentioning anything about it (it probably was), but anyways this is interfering with the normal functionality of the `md-autofocus`  attribute, because it is used, to mark an element to be focused upon opening. (e.g `$mdSidenav`, `$mdBottomSheet`, `$mdDialog`)

* This also fixes an issue where the autocomplete dropdown shows up twice, when having the 
autocomplete inside of a dialog.

This change now provides consistency for the `md-autofocus` attribute on the autocomplete as well.

@ThomasBurleson This **can** be marked as a breaking change, but the functionality was not documented for long time.. so not sure.

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -278,6 +278,36 @@ describe('<md-autocomplete>', function() {
       element.remove();
     }));
 
+    it('should forward focus to the input element with md-autofocus', inject(function($timeout) {
+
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
@DevVersion - lgtm

PR #9449, opened 2016-08-29
fix(button): fix rendering issue with borders and webkit browsers

### LINKED ISSUE #9154: Ripple does not fill button with md-cornered
**Actual Behavior**:
-  `What is the issue?` Ripple does not fill button with md-cornered
-  `What is the expected behavior?` Ripple should fill entire button

-  `CodePen Demo which shows your issue:` http://codepen.io/svencowart/pen/JKkARL

**Angular Versions**:
-  `Angular Version:` 1.4.8
-  `Angular Material Version:` 1.1.0-RC5

clshortfuse:
Webkit has a rendering bug with child elements whos parent uses a
rounded corner. In this case, this is a ripple on a rounded button.
To circumvent this issue, we applied a 1x1 PNG image mask on the
ripple element.

The image mask is causing issues with Chrome rendering opaque under
certain conditions. An image mask is only needed for Safari 8, which
is no longer supported. Instead we will just Safari and Chrome to
use a compositing layer by forcing an empty Z-axis translation.

* Inherit border radius from buttons
* Remove image mask CSS
* Use translateZ to force webkit to create a compositing layer

Fixes #9154, #10086

References:

https://bugs.webkit.org/show_bug.cgi?id=30475

### DIFF
--- a/src/components/button/button.scss
+++ b/src/components/button/button.scss
@@ -115,13 +115,6 @@ button.md-button::-moz-focus-inner {
     padding: $baseline-grid;
     width: $icon-button-width;
     border-radius: $button-icon-border-radius;
-    .md-ripple-container {
-      border-radius: $button-icon-border-radius;
-      background-clip: padding-box;
-      overflow: hidden;
-      // The following hack causes Safari/Chrome to respect overflow hidden for ripples
-      -webkit-mask-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAAA5JREFUeNpiYGBgAAgwAAAEAAGbA+oJAAAAAElFTkSuQmCC');
-    }
   }
 
   &.md-fab {
@@ -145,13 +138,6 @@ button.md-button::-moz-focus-inner {
 
     transition: $swift-ease-in;
     transition-property: background-color, box-shadow, transform;
-    .md-ripple-container {
-      border-radius: $button-fab-border-radius;
-      background-clip: padding-box;
-      overflow: hidden;
-      // The following hack causes Safari/Chrome to respect overflow hidden for ripples
-      -webkit-mask-image: url('data:image/png;base64,iVBORw0KGgoAAAANSUhEUgAAAAEAAAABCAIAAACQd1PeAAAAGXRFWHRTb2Z0d2FyZQBBZG9iZSBJbWFnZVJlYWR5ccllPAAAAA5JREFUeNpiYGBgAAgwAAAEAAGbA+oJAAAAAElFTkSuQmCC');
-    }
 
     &.md-mini {
       line-height: $button-fab-mini-line-height;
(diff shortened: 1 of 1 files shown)


topherfangio:
Can you add a bit more detail to the commit message about why the old hack was there and why this is an improvement?

Otherwise LGTM :+1:

topherfangio:
@ThomasBurleson Ready for your review. Carlos and I have both done manual testing on various browsers.

clshortfuse:
I've done a bit more research on the issue. There is a webkit issue with one element bleeding through a parent element (ripple bleeding through corner). The previous fix created used a 1x1 pixel `mask-image` which worked, but in reality wasn't really doing anything to the actual rendering. What was going on, is that in webkit, certain operations will for elements to create their own backing surface (aka a compositing layer).

When you add a mask to an image, the browser will break up the elements into layers and overlay a mask layer over the element. Technically, yes, we are going to force 3D, but for the express objective of making sure the element has a compositing layer, eliminating the overlay bleed issue. 

There are many ways to force a composition layer, and transform3D is just another one, that's much safer than applying a mask.

References: https://www.chromium.org/developers/design-documents/gpu-accelerated-compositing-in-chrome (from RenderLayers to GraphicsLayers)
https://chromium.googlesource.com/chromium/blink/+/master/Source/platform/graphics/CompositingReasons.h#84
https://trac.webkit.org/wiki/Accelerated%20rendering%20and%20compositing
https://github.com/WebKit/webkit/blob/master/Source/WebCore/platform/graphics/GraphicsLayer.h#L283

@DevVersion ping, you brought up concerns. Basically, before it was 3D + mask layer. Now it's just 3D (no mask needed)

DevVersion:
@clshortfuse This sounds valid. LGTM then!

clshortfuse:
Well, well, well. Seems like it was right to retest all this stuff with Browserstack. Safari 8 needs both compositing and image mask, not just compositing. (Safari 9 and Chrome are okay with just `transformZ`).

Sure, `-webkit-transform:translateZ(0)` is 14 characters less in CSS, but this way, we can support Safari 8 as well.

I've rewritten the code and commit, using this codepen for reviewing: http://codepen.io/shortfuse/full/dpYQaz

EladBezalel:
Is it me or then ripple isn't working correctly ?

src/components/button/button.scss
```diff
@@ -170,11 +156,13 @@ button.md-button::-moz-focus-inner {
   }
 
   .md-ripple-container {
-    border-radius: $button-border-radius;
+    border-radius: inherit;
```
EladBezalel: Why do we inherit? we want it to always be 50%

clshortfuse: `border-radius` isn't the shape of the ripple, it's the border limit to which the ripple will reach on its container. See #9154

topherfangio: Also, if they use `md-cornered`, then it shouldn't be `50%`, it should be `0`.

clshortfuse:
I reverted back to using `translateZ` because the image mask is causing issues with Chrome. We don't need it the image mask anymore since that was only for Safari 8.

marshall007:
@clshortfuse what's the status on this? Your patch is working great for us in Chrome. Do we just need some manual testing in other browsers?

PR #9451, opened 2016-08-29
fix(tabs): support flexible content layout

### LINKED ISSUE #9206: Md-Tab Content width issue within mdDialog
**Actual Behavior**:
-  `What is the issue? *` Md-Tab content within an mdDialog does not initially gets the full width. After interaction with an element, the content "jumps" to full width.
-  `What is the expected behavior?` The content should have its expected width from the beginning.

-  `CodePen Demo which shows your issue:` http://jsfiddle.net/aftq6sgo/ and http://codepen.io/DevVersion/pen/WxmKJL (provided by DevVersion)
-  `Details:` Open the Dialog. You see how the green Div (with an input field) has the full width. However, the blue one (which is within die tab container), has some issue. As soon as you click in the input field, the width changes.
  I used a simple <input> for the example. The issue exists with anything within the tab container.

**Angular Versions**: *
-  `Angular Version:` 1.5.7
-  `Angular Material Version:` 1.0.9

**Additional Information**:
-  `Browser Type: *` Chrome
-  `Browser Version: *` 51
-  `OS: *` Windows 8
-  `Stack Traces:` -

See the issue in action:

github_issue

clshortfuse:
Certain browsers had issues with resizing tab content md-tabs' parent
would change size or have an unspecified size during creation (for
example in a dialog). Converting to flexbox solves this issue.
- Set md-tab-content to flexbox
- Set md-tab-content's div to fill parent
- Set md-pagination-wrapper to flexbox

Fixes #9206, #9704, #9779

### DIFF
--- a/src/components/tabs/tabs.scss
+++ b/src/components/tabs/tabs.scss
@@ -120,7 +120,6 @@ md-tabs-wrapper {
   &.md-stretch-tabs {
     md-pagination-wrapper {
       width: 100%;
-      display: flex;
       flex-direction: row;
       md-tab-item {
         flex-grow: 1;
@@ -157,7 +156,7 @@ md-tabs-canvas {
 md-pagination-wrapper {
   @include pie-clearfix;
   height: $tabs-header-height;
-  display: block;
+  display: flex;
   transition: transform $swift-ease-in-out-duration $swift-ease-in-out-timing-function;
   position: absolute;
   width: 999999px;
@@ -181,7 +180,7 @@ md-tabs-content-wrapper {
 }
 
 md-tab-content {
-  display: block;
+  display: flex;
   position: absolute;
   top: 0;
   left: 0;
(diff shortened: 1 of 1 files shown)


topherfangio:
I feel like this might break some existing users who expect it to be block?

Also, have you tested this across all the browsers we support to make sure it works as expected?

clshortfuse:
@topherfangio OK on Chrome, Firefox, IE 10, IE11 and Edge.

The reason I didn't just use `display:flex` was so we can use our layout engine which has more predictable results. I have yet to test on on Safari which honestly might be the only browser to present issues do to its `flex-shrink` bug. https://bugs.webkit.org/show_bug.cgi?id=146020

clshortfuse:
I've reworked this to not use mdLayout and instead the `display:flex` in CSS. 

This should not break layouts since user content is wrapped by  `md-tab-content`, which is what we are modifying here. Despite we are adding CSS to `> .div`, the `flex` rule only applies to its relation to its parent, not its children.

We are also using `flex: 1 0` to avoid any possible issues due to Safari's flexbug rendering issue: https://github.com/philipwalton/flexbugs#1-minimum-content-sizing-of-flex-items-not-honored

PR #9452, opened 2016-08-29
fix(tabs) dummy tabs should not have acccessibilty roles

### LINKED ISSUE #9450: [a11y] md-tabs has twice as many tabs as it should
**Actual Behavior**:
 -  `What is the issue? *`  There are twice as many role="tab" inside md-tabs as there should be.  This causes ChromeVox to read "Tab 2 of 4" with only 2 tabs.
 -  `What is the expected behavior?`  It should read tab 2 of 2

 -  `CodePen Demo which shows your issue:` http://codepen.io/anon/pen/zBgVmJ
 -  `Details:` Turn on ChromeVox and focus a tab.

**Angular Versions**: *
 -  `Angular Version:` 1.5

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *` 52
 -  `OS: *` Ubuntu
 -  `Stack Traces:`

clshortfuse:
* Remove role="tab" from dummy tabs

Fixes #9450

### DIFF
--- a/src/components/tabs/js/tabsDirective.js
+++ b/src/components/tabs/js/tabsDirective.js
@@ -159,7 +159,6 @@ function MdTabs ($$mdSvgRegistry) {
                   'class="md-tab" ' +
                   'tabindex="-1" ' +
                   'id="tab-item-{{::tab.id}}" ' +
-                  'role="tab" ' +
                   'aria-controls="tab-content-{{::tab.id}}" ' +
                   'aria-selected="{{tab.isActive()}}" ' +
                   'aria-disabled="{{tab.scope.disabled || \'false\'}}" ' +
--- a/src/components/tabs/tabs.spec.js
+++ b/src/components/tabs/tabs.spec.js
@@ -354,7 +354,6 @@ describe('<md-tabs>', function () {
       expect(tabs.find('md-tabs-canvas').attr('role')).toBe('tablist');
 
       expect(tabItem.attr('id')).toBeTruthy();
-      expect(tabItem.attr('role')).toBe('tab');
       expect(tabItem.attr('aria-controls')).toBe(tabContent.attr('id'));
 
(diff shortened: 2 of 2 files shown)


clshortfuse:
I'm not sure understanding why the spec tests are actively checking for md-dummy-tab to have a role='tab'.

crisbeto:
I'm not sure either. It could have something to do with them forwarding focus to the real tabs.

topherfangio:
@clshortfuse Can you do some testing by removing the `role` from the non-dummy tabs and see if keyboard accessibility still works?

clshortfuse:
@topherfangio I wouldn't say it works properly because I found a keyboard bug with dynamic heights in master where you can't navigating while the animation is running, but yes, **keyboard works like master**. 

Travis Fails because of a spec test is expecting a dummy tab to have a role. I'm not sure why this would be expected behavior. I'll need some clarification about this.

https://github.com/angular/material/blob/875848817bf9820f510647d2b01a347bb4fd8e46/src/components/tabs/tabs.spec.js#L357

topherfangio:
@clshortfuse Is the dynamic height bug you mentioned fixed by the dummy tabs having the role?

clshortfuse:
@topherfangio I discovered that keyboard events aren't processed while a tab is in a transition phase and dynamic height is on. In other words, go to demo and spam right, enter and right, enter. It's unrelated to this PR, just something I found.

I'm going to just remove the `role` check in dummy tabs and add a spec test that ensures dummy tabs do not have `role` attribute.

I will say, there is another way to solve the issue, which is to set a dummy tab with a hidden attribute. Technically, dummy tabs are hidden and it may be a better solution, but it creates a lot more problems than just removing the `role` attribute.

clshortfuse:
Spec tests added.

PR #9453, opened 2016-08-29
feat(tooltip): add the ability to specify custom trigger events

crisbeto:
Adds the `md-enter-trigger` and `md-leave-trigger` properties, which allow users to specify which browser events will open/close a tooltip instance. This can be useful for cases where the user might want to trigger a tooltip on an event, different from the default ones, or to disable one or more of the default events.

@ThomasBurleson note that this isn't something that was requested in an issue, but could it be useful in the future.

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -32,6 +32,8 @@ angular
  * parent. Defaults to 0ms on non-touch devices and 75ms on touch.
  * @param {boolean=} md-autohide If present or provided with a boolean value, the tooltip will hide on mouse leave, regardless of focus
  * @param {string=} md-direction Which direction would you like the tooltip to go?  Supports left, right, top, and bottom.  Defaults to bottom.
+ * @param {string=} md-enter-trigger Space-separated list of events that will cause a tooltip to appear. Defaults to `"focus touchstart mouseenter"`.
+ * @param {string=} md-leave-trigger Space-separated list of events that will cause a tooltip to become hidden. Defaults to `"blur touchcancel mouseleave"`.
  */
 function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdTheming, $rootElement,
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
This functionality/API needs to be inside mdPanel.

ErinCoughlan:
@crisbeto @EladBezalel I think I misunderstood this PR earlier. You're changing what is used to open the panel, so there's not going to be a way to add it to MdPanel. MdPanel has all functionality after you call open, but nothing to do with when open is called.

Since this is needed for other components, I think the correct location may be a utils or internally shared component that handles events opening panels.

Does this make sense? Or am I still confused?

crisbeto:
@ErinCoughlan before I opened this PR, I checked with @bradrich about what moving the tooltip to panel actually meant. As far as I understood, it would only change the positioning logic, but the enter/leave event handlers would remain the same. That's why I though that this would be a good idea to add it as a configurable option.

ErinCoughlan:
@crisbeto I think I agree. Here's the panel/tooltip split I'm thinking about. My biggest question currently is how to add this functionality for fabs and other components as well as tooltips without duplicate work. That's why we were thinking add it to panel because eventually everything will be panel.

Tooltip:
- Handles what triggers the panel to open.
- Handles what triggers the panel to close.
- Sets/changes the config for positioning, animation, and panelRef.
- Controls panel content.

Panel:
- Actually performs the open and close actions.
- All animation logic based on config. 
- All positioning logic based on config.

Any component switching to a panel will follow the same split, which is why I'm looking for utils that make adding functionality easier. Whenever something is implemented twice, we're going to end up with bugs and inconsistency.

crisbeto:
That's what I was thinking about as well. It doesn't really make sense to have the event handlers in mdPanel since it's supposed to be as generic as possible.

bradrich:
Agreed on all said. I am going to update my tooltip WIP ASAP.

PR #9455, opened 2016-08-30
chore(build): minify theme template for ng constant

### LINKED ISSUE #9066: fix(scss): should allow multiline comments
Multiline comments:
```js
/**
 * yay
 */
```
Can't be compiled (in some node versions) with [`node-sass`](https://github.com/sass/node-sass) into 1 line comment,

DevVersion:
Now minifies the theme css for the Angular constant.

As discussed:
>- Now minifying the SCSS for the $MD_THEME_CSS constant. (new)
> - Not minifying the SCSS, when exporting a release build for bower-material. (currently)

R: @EladBezalel 

Fixes #9066.

### DIFF
--- a/gulp/util.js
+++ b/gulp/util.js
@@ -229,6 +229,7 @@ function themeBuildStream() {
       .pipe(utils.hoistScssVariables())
       .pipe(sass())
       .pipe(dedupeCss())
+      .pipe(minifyCss())
       .pipe(utils.cssToNgConstant('material.core', '$MD_THEME_CSS'));
 }
 


crisbeto:
This could be a good opportunity to get rid of the homebrew minification from here https://github.com/angular/material/blob/master/scripts/gulp-utils.js#L292

topherfangio:
@crisbeto That doesn't do minification, does it? It looks like it's actually building JS and inserting the string into the middle of it.

ThomasBurleson:
@crisbeto - what do you propose as a workaround ?

DevVersion:
@topherfangio Exactly, I can't see any minification as well.

crisbeto:
Yes, I meant the line breaks. They should have been removed already.

DevVersion:
@crisbeto I would rather keep it right now, because theoretically we don't know that the constant is already minified before.

crisbeto:
True, although this is the only place that it's used. Either way, it's not a big deal.

PR #9457, opened 2016-08-30
fix(tabs): scroll blocks in pagination (related to #5439)

enne30:
Fixes md-tabs scroll blocking happening when single tab is larger than containing canvas (e.g. tabs containing long text elements).

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -333,8 +333,12 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
       tab = elements.tabs[ i ];
       if (tab.offsetLeft + tab.offsetWidth > totalWidth) break;
     }
-    if (viewportWidth > tab.offsetWidth) ctrl.offsetLeft = fixOffset(tab.offsetLeft);
-    else ctrl.offsetLeft = fixOffset(tab.offsetLeft + (tab.offsetWidth - viewportWidth + 1));
+    
+    if (viewportWidth > tab.offsetWidth) {
+        ctrl.offsetLeft = fixOffset(tab.offsetLeft);
+    } else {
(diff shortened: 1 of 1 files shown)


enne30:
I signed it!

topherfangio:
This looks good.

Can you please update it to follow our style guide though? Particularly, the `if/else` statements should look like the following:

```js
if (test) {
  // code here
} else {
  // more code here
}
```

Thanks!

enne30:
Amended style, thank you for quick response :)

topherfangio:
Manually confirmed that this is working. Without it, tabs with names longer than the canvas break the pagination and it does not scroll at all.

That said, the [spec](https://material.google.com/components/tabs.html#tabs-specs) currently says you should have a max-width of 264dp, and it looks like we're not following that.

Additionally, if the text is too long to fit, we're supposed to wrap the text and then truncate at the max-width.

This is a good temporary fix, but I think the long-term solution would be to update tabs to fully support the spec.

@enne30 If you get a chance before this gets merged, can you please:

 - add a comment to each of your `if/else` statements just explaining what they are checking for,
 - use 2 spaces instead of 4 for indentation,
 - and then squash your commits once you're finished?

Thanks so much!

enne30:
@topherfangio Squashed new commit! :D

Let me know if something's still wrong, this is my very first contribution and my english is *not* excellent, too :p

topherfangio:
@enne30 Thanks for the updates!

It's not a huge deal, but when we squash commits, we usually squash all of the commits. This can be done pretty easily by running `git rebase -i master` and then using only the first commit, and merging all of the commit messages into a single message.

Once you're done, you can force-push to your `patch-2` branch on your forked repo and it should update this PR.

But again, GitHub has a shiny new button that does this for us, so if you run into trouble or don't have the time, no worries 😃

enne30:
@topherfangio : sorry, I really didn't know this git option

Thanks for useful information :)

topherfangio:
@enne30 No problem! Thanks so much for the contribution!

PR #9458, opened 2016-08-30
fix(util): properly determine viewport top offset

### LINKED ISSUE #9370: Scroll Bug.
When I trying to open the menu or something like it I have a problem with page scrolling. This bug gives me a pain for a long time.
Maybe this problem is already discovered by you but I want to write my observations all the same.
Ok. I checked the code with Chrome debugger and discovered following:
When the script trying to modify body's style it uses $mdUtils.scrollTop function. 

This function has a line with small bug: 
`
      return scrollTop || Math.abs(element[0].getBoundingClientRect().top);
`
_angular-material.js:1032 (version v1.1.0-rc.5)_

scrollTop variable can be equal zero, that in this case it will be converted to boolean and then second expression will be evaluated. In my case the returned value was 60.
You can use strict equation with null and undefined instead.

I wrote this small directive for temporary fix this bug till Angular Material will be updated:

`

    app.directive('fixScroll', function ($window) {
        return {
            restrict: 'A',
            link: function (scope, el, attrs) {
                angular.element($window).on('scroll', function () {
                    if (document.body.scrollTop === 0) {
                        document.body.scrollTop = 1;
                    }
                });
            }
        }
    });
`

It simply prevents the scrolling to 0.

DevVersion:
* Properly determines the viewport top offset, without any consistency issues for the `scrollTop`.
* The `scrollTop` method had a wrong name when passing an element, because it resolved the client rects in relative to the viewport.
* Now it determines the viewport top offset from the window (instead of using the body element - this is deprecated)
* Also makes sure that the calculations in the source are not too long (extra variables)

**FYI**: Mostly some cleanup in the `disableBodyScroll` function.

Fixes #9370.

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -572,7 +572,8 @@ function inputTextareaDirective($mdUtil, $window, $mdAria, $timeout, $mdGesture)
 
         function onDrag(ev) {
           if (!isDragging) return;
-          element.css('height', startHeight + (ev.pointer.y - dragStart) - $mdUtil.scrollTop() + 'px');
+
+          element.css('height', (startHeight + ev.pointer.distanceY) + 'px');
         }
 
         function onDragEnd(ev) {
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -133,18 +133,12 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
     },
 
     /**
-     * Calculate the positive scroll offset
-     * TODO: Check with pinch-zoom in IE/Chrome;
-     *       https://code.google.com/p/chromium/issues/detail?id=496285
+     * Determines the absolute position of the viewport.
+     * Useful when making client rectangles absolute.
+     * @returns {number}
      */
-    scrollTop: function(element) {
-      element = angular.element(element || $document[0].body);
-
-      var body = (element[0] == $document[0].body) ? $document[0].body : undefined;
-      var scrollTop = body ? body.scrollTop + body.parentElement.scrollTop : 0;
-
-      // Calculate the positive scroll offset
-      return scrollTop || Math.abs(element[0].getBoundingClientRect().top);
+    getViewportTop: function() {
+      return window.scrollY || window.pageYOffset || 0;
     },
 
     /**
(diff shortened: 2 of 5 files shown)


src/components/input/input.js
```diff
@@ -569,7 +569,8 @@ function inputTextareaDirective($mdUtil, $window, $mdAria, $timeout, $mdGesture)
 
         function onDrag(ev) {
           if (!isDragging) return;
-          element.css('height', startHeight + (ev.pointer.y - dragStart) - $mdUtil.scrollTop() + 'px');
+
+          element.css('height', (startHeight + ev.pointer.distanceY) + 'px');
```
topherfangio: The math seems different here. Can you add a note to your commit message (and maybe here) about why it's been changed?

DevVersion: It's under the hood the same. Kristyian just calculated the distance manually, but `$mdGestures` already does this under the hood (common usage)

src/core/util/util.spec.js
```diff
@@ -223,10 +223,33 @@ describe('util', function() {
         // Restore the scrolling.
         enableScrolling();
         window.scrollTo(0, 0);
-        document.body.removeChild(element[0]);
+
+        element.remove();
       }));
     });
 
+    describe('getViewportTop', function() {
```
topherfangio: Can we add some tests to make sure everything is correct even if they use `<md-content>` to scroll (rather than the body)?

DevVersion: The viewport is not the `md-content`. `Window` is always the viewport, and `getBoundingClientRect()` is always in relative to the viewport.

topherfangio:
LGTM but is there a specific issue we can manually test to make sure it's fixed?

DevVersion:
@topherfangio There is no provided codepen or something else, we just saw in #9370 that his point is valid and at that the method is doing wrong stuff with the ClientRect

src/core/util/util.js
```diff
@@ -251,37 +245,39 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
+          body.style.overflow = 'hidden';
+        }
 
         return function restoreScroll() {
-          body.style.cssText = restoreBodyStyle;
-          htmlNode.style.cssText = restoreHtmlStyle;
-          body.scrollTop = scrollOffset;
-          htmlNode.scrollTop = scrollOffset;
+          // Reset the inline style CSS to the previous.
+          body.style.cssText = prevBodyStyle;
```
EladBezalel: Actually I've encountered the need to copy the styles from one element to another TODAY!
And `cssText` didn't seemed to be working.. is it tested?

DevVersion: I did not introduce the `cssText` in this PR - also when I took a look at it, it seemed to work a like a charm - It's a basic thing for the `CSSStyleDeclaration` prototype.
> https://developer.mozilla.org/en-US/docs/Web/API/CSSStyleDeclaration/cssText

EladBezalel: Looks like it's also supported on IE8 so we're fine 
http://stackoverflow.com/a/4494571/890293
cool

topherfangio:
This looks good to me and is ready for presubmit, but without any way to test, I'm just a bit hesitant to give it a go ahead.

@DevVersion Is there something specific that we could see broken if there was something wrong with the code?

DevVersion:
@topherfangio Nothing was broken, because we did not use the function with an specific element. The function just combined two different things into one (which is confusing and could cause issues)

kara:
@DevVersion  Can you rebase this?

DevVersion:
@kara Yeah. Done

kara:
Thanks!

PR #9462, opened 2016-08-30
fix(compiler): remove manual controllerAs logic

crisbeto:
Currently the `controllerAs` logic in the compiler is done manually, however the same result can be achieved by passing the `controllerAs` value to the `$controller` factory.


DevVersion:
Very simple change. LGTM.

topherfangio:
LGTM 👍 

Although perhaps we should note that the last two parameters (`true` and `controllerAs`) are undocumented and considered private?

https://github.com/angular/angular.js/blob/master/src/ng/controller.js#L99

crisbeto:
True, but I checked back to 1.3 and they haven't been changed so we should be fine. Also we were using the third parameter already.

PR #9484, opened 2016-09-01
fix(theming): match preceding selectors as well

### LINKED ISSUE #9480: Tab font colors are broken in latest HEAD
**Actual Behavior**:
 -  `What is the issue? *` Tab font colors are broken in current HEAD both for active tabs and non-active.
 -  `What is the expected behavior?` Correct tab colors as it was very recently.

 -  `CodePen Demo which shows your issue:` https://material.angularjs.org/HEAD/demo/tabs
 -  `Details:` See demo.

**Angular Versions**: *
 -  `Angular Version:` 1.5.8
 -  `Angular Material Version:` HEAD

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `OS: *` macOS
 -  `Stack Traces:` None.

DevVersion:
* The theming service currently only matches from the beginning of the `md-default-theme`, but all preceding selectors without whitespaces are being ignored.

* The regex is doing more than it actually needs. Having an extra group for the target is unnecessary   (Since we only want to concatenate the start and the end without the `.md-default-theme` class).

* Kristiyan also helped with this effort and we figured out, that NanoCSS accidentally removes the theme color expressions from the `border` CSS property.

Shout-out again to @crisbeto 
  

Fixes #9480

### DIFF
--- a/gulp/util.js
+++ b/gulp/util.js
@@ -84,13 +84,15 @@ function autoprefix () {
   ]});
 }
 
-function minifyCss() {
-  return nano({
+function minifyCss(extraOptions) {
+  var options = {
     autoprefixer: false,
     reduceTransforms: false,
     svgo: false,
     safe: true
-  });
+  };
+
+  return nano(_.assign(options, extraOptions));
 }
 
 function buildModule(module, opts) {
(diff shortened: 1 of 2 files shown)


crisbeto:
I've been testing it out together with Paul and it LGTM.

DevVersion:
Ping @EladBezalel for ping. This was our discussion in the meeting and was planned to go into `1.1.2` instead.

EladBezalel:
As long it was tested by @crisbeto and @DevVersion locally - LGTM

PR #9490, opened 2016-09-01
docs(): viewport should be not scalable

DevVersion:
* Normally modern webapp do not allow the developer to zoom into the page on mobile.
  This applies the most common meta tag to the docs page and makes the demos look way better on mobile.

* Components, which have a text input (like datepicker, autocomplete or normal input) no longer zoom in, when clicking on them.

I had a quick chat with @topherfangio about it, and we were not 100% sure, why we didn't apply that meta tag before.

### DIFF
--- a/docs/config/template/index.template.html
+++ b/docs/config/template/index.template.html
@@ -7,7 +7,7 @@
   Angular Material
 </title>
 <link rel="icon" type="image/x-icon" href="favicon.ico" />
-<meta name="viewport" content="initial-scale=1" />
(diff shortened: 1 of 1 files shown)


clshortfuse:
As an added bonus, this will remove the 300ms click delay on mobile devices.

@DevVersion Why not also add `maximum-scale=1` to keep scaling consistent on rotation.

See https://developer.mozilla.org/en-US/docs/Mozilla/Mobile/Viewport_meta_tag

Issue #9414, opened 2016-08-26
ContentEditable is still not focusable on mobile


kpgarrod:
**Actual Behavior**:
 -  `What is the issue? *`
Using ngWig editor on a page with angular-material, the ngWig editor window (contentEditable) cannot be brought into focus by clicking on it. 
 -  `What is the expected behavior?`
It should be possible to bring the ngWig editor window into focus (and therefore edit the content) by clicking on it 
**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:` http://codepen.io/kpgarrod/pen/BKVeZV
 -  `Details:` Try to edit the 'Biography' field on an Android (4.4 e.g.) device.

**Angular Versions**: *
 -  `Angular Version:` 1.5
 -  `Angular Material Version:` 1.1.0

**Additional Information**:
 -  `Browser Type: *` Android browser
 -  `Browser Version: *` 
 -  `OS: *` Android 4.4
 -  `Stack Traces:`

----
Shortcut to create a [new CodePen Demo](http://codepen.io/team/AngularMaterial/pen/bEGJdd).
Note: `*` indicates required information. Without this information, your issue may be auto-closed.

> Do not modify the titles or questions. Simply add your responses to the ends of the questions.
  Add more lines if needed.

DevVersion:
That's a valid issue - The problem is related to our Fastclick Solution.

---

As a workaround you can disable the click hijacking and everything works as expected (tried it out)
> http://codepen.io/DevVersion/pen/XKLLrY

```js
$mdGestureProvider.skipClickHijack();
```

Anyways that issue will be hopefully fixed when dropping `$mdGestures` and using another Fastclick solution like HammerJS. See #9362

kpgarrod:
@DevVersion that certainly seems to work for me. I really appreciate the quick response :). Are there any implications of skipping clickHijack?

DevVersion:
I'm going to merge this issue into #9362 now.

Issue #9362, opened 2016-08-19
Drop $mdGestures by using HammerJS


DevVersion:
It is planned to drop our internal gesture service and depend on HammerJS.

This could improve performance and also provides us a better approach of solving the Fastclick issue.
> Also drops the `skipClickHijack`

Splaktar:
This would be nice to have, but it's unclear at this time if this is viable given resources and the significant impact and effort involved.

Splaktar:
This is also related to https://github.com/angular/material/issues/2328.

PR #9491, opened 2016-09-01
Bypass click/mouse handlers for any contenteditable elements

hanthomas:
Fixes issues #9414 and #5937.  I can't see any reason why ngMaterial would need to hijack clicks for elements that have contenteditable attribute set.

### DIFF
--- a/src/core/services/gesture/gesture.spec.js
+++ b/src/core/services/gesture/gesture.spec.js
@@ -715,6 +715,10 @@ describe('$mdGesture', function() {
         target: el[0]
       });
 
+      expect(spyClick).not.toHaveBeenCalled();
+
+      el.trigger('click');
+
       expect(spyClick).toHaveBeenCalled();
     }));
 
@@ -726,12 +730,20 @@ describe('$mdGesture', function() {
         target: el[0]
       });
 
+      expect(spyMouseDown).not.toHaveBeenCalled();
+
+      el.trigger('mousedown');
+
       expect(spyMouseDown).toHaveBeenCalled();
 
       $document.triggerHandler({
         type: 'mouseup',
         target: el[0]
       });
+
+      expect(spyMouseUp).not.toHaveBeenCalled();
+
+      el.trigger('mouseup');
 
       expect(spyMouseUp).toHaveBeenCalled();
     }));
--- a/src/core/services/gesture/gesture.spec.js
+++ b/src/core/services/gesture/gesture.spec.js
@@ -717,7 +717,7 @@ describe('$mdGesture', function() {
 
       expect(spyClick).not.toHaveBeenCalled();
 
-      el.trigger('click');
+      el.triggerHandler('click');
 
       expect(spyClick).toHaveBeenCalled();
     }));
@@ -732,7 +732,7 @@ describe('$mdGesture', function() {
 
       expect(spyMouseDown).not.toHaveBeenCalled();
 
-      el.trigger('mousedown');
+      el.triggerHandler('mousedown');
 
       expect(spyMouseDown).toHaveBeenCalled();
 
@@ -743,7 +743,7 @@ describe('$mdGesture', function() {
 
       expect(spyMouseUp).not.toHaveBeenCalled();
 
-      el.trigger('mouseup');
+      el.triggerHandler('mouseup');
 
       expect(spyMouseUp).toHaveBeenCalled();
     }));
--- a/src/core/services/gesture/gesture.spec.js
+++ b/src/core/services/gesture/gesture.spec.js
@@ -690,13 +690,16 @@ describe('$mdGesture', function() {
 
   describe('contenteditable', function() {
 
-    // Click tests should only be enabled when `$$hijackClicks == true` (for mobile)
-
     it('should not hijack click on contenteditable element', inject(function($document, $mdGesture) {
       var spy = jasmine.createSpy('click');
       var el = angular.element('<div contenteditable>');
 
       el.on('click', spy);
+
+      $document.triggerHandler({
+        type: 'click',
+        target: el[0]
+      });
 
       expect(spy).toHaveBeenCalled();
     }));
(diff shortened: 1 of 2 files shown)


hanthomas:
I signed it!

hanthomas:
I signed it!

ThomasBurleson:
Pending/related to HammerJS efforts.

ThomasBurleson:
@hanthomas - please add 1 or more unit tests.

hanthomas:
@ThomasBurleson - Unit tests have been added to gesture.spec.js.

src/core/services/gesture/gesture.spec.js
```diff
@@ -688,4 +688,36 @@ describe('$mdGesture', function() {
 
   });
 
+  describe('contenteditable', function() {
+
+    // Click tests should only be enabled when `$$hijackClicks == true` (for mobile)
+
+    it('should not hijack click on contenteditable element', inject(function($document, $mdGesture) {
+      if ( $mdGesture.$$hijackClicks ) {
```
DevVersion: Where does this property exist? Never saw this one in our gesture service.

Also the tests should enable hijacking by default.

src/core/services/gesture/gesture.spec.js
```diff
@@ -688,4 +688,36 @@ describe('$mdGesture', function() {
...
+
+        el.on('click', spy);
+
+        expect(spy).toHaveBeenCalled();
+      }
+
+    }));
+
+    it('should not hijack mousedown/mouseup on contenteditable element', inject(function($document, $mdGesture) {
+      if ( $mdGesture.$$hijackClicks ) {
```
DevVersion: Same for here

src/core/services/gesture/gesture.spec.js
```diff
@@ -688,4 +688,36 @@ describe('$mdGesture', function() {
...
+        var spy1 = jasmine.createSpy('mousedown');
+        var spy2 = jasmine.createSpy('mouseup');
+        var el = angular.element('<div contenteditable>');
+
+        el.on('mousedown', spy1);
```
DevVersion: Where do you actually trigger the event? The tests seem to pass because you created that `$$hijackClicks` check, which always doesn't pass

src/core/services/gesture/gesture.spec.js
hanthomas: Hmm... you're right.  I used one of the existing tests as a template, but apparently that property no longer exists.  I'll make the correction.

src/core/services/gesture/gesture.spec.js
hanthomas: See reply to previous comment

src/core/services/gesture/gesture.spec.js
hanthomas: True.  So, once I take that condition out, it should pass if the mousedown and mouseup handlers are triggered.

DevVersion: I think it gets more difficult for this tests.

You need the element you want to trigger the mouse event on to the `document.body` to be able to test it with `$mdGesture`.

Then you need to dispatch a fake mouse event and determine if the triggered event from the event handlers is the same as the fake mouse event.

hanthomas: You're right... plus, now the tests fail.  So, it looks like I have some work to do on this.  Thanks for the explanation.

hanthomas:
@DevVersion - It took a few tries, but I think the unit tests are correct now.  My basic approach was to make sure that ngMaterial allows the mouse events to be processed by the handler attached to the element instead of the click hijacker.

src/core/services/gesture/gesture.spec.js
```diff
@@ -688,4 +688,64 @@ describe('$mdGesture', function() {
 
   });
 
+  describe('contenteditable', function() {
+
+    var spyClick, spyMouseDown, spyMouseUp;
+    var el;
+
+    beforeEach(function() {
+      inject(function($mdGesture) {
```
DevVersion: You can simplify this

```js
beforeEach(inject(function($mdGesture) {
}));
```

src/core/services/gesture/gesture.spec.js
```diff
@@ -688,4 +688,64 @@ describe('$mdGesture', function() {
...
+        el = angular.element('<div contenteditable="true">');
+
+        el.on('click', spyClick);
+        el.on('mousedown', spyMouseDown);
+        el.on('mouseup', spyMouseUp);
+      });
+    });
+
+    it('should not hijack click on contenteditable element', inject(function($document, $mdGesture) {
+      $document.triggerHandler('$$mdGestureReset');
```
DevVersion: Move this to the `beforeEach`, since you are repeating it for each test

src/core/services/gesture/gesture.spec.js
```diff
@@ -688,4 +688,64 @@ describe('$mdGesture', function() {
...
+      $document.triggerHandler({
+        type: 'click',
+        target: el[0]
+      });
+
+      expect(spyClick).not.toHaveBeenCalled();
+
+      el.triggerHandler('click');
+
+      expect(spyClick).toHaveBeenCalled();
```
DevVersion: Shouldn't those checks also check for the event to not being prevented. 

```js
// Something like this
var event = spyClick.calls.mostRecent().args[0];
bool = event.defaultPrevented

```

DevVersion:
I will close that PR for now, because this is not a high priority on our queue right now, because it is planned to get rid of the `$mdGesture` service in the future (See #9362).

Anyways, Thanks for the PR!

kmturley:
This issue also affects regular <select> elements when using angular-material for other elements on the same page.

Using this code fixes it:

    function mouseInputHijacker(ev) {
      var isKeyClick = !ev.clientX && !ev.clientY;
      if (!isKeyClick && !ev.$material && !ev.isIonicTap
        && !isInputEventFromLabelClick(ev) && !ev.target.nodeName === 'SELECT') {
        ev.preventDefault();
        ev.stopPropagation();
      }
    }

Would be good to have a directive to disable click hijacking on specific elements

PR #9496, opened 2016-09-02
feat(autocomplete): support ng-trim on the underlaying input

### LINKED ISSUE #4492: mdAutocomplete: pass ng-* applicable attributes to input

DevVersion:
* A lot of developers requested a way to pass the `ng-trim` attribute to the underlaying input element.
  This is reasonable, because ngTrim always trims the searchText and can cause deviations with the search results.

* Also added a more generic way of forwarding the the (currently) three attributes to the input element, without doing this in the template.

Fixes #4492

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -374,6 +374,31 @@ describe('<md-autocomplete>', function() {
       element.remove();
     }));
 
+    it('should support ng-trim for the search input', inject(function() {
+      var scope = createScope(null, {inputId: 'custom-input-id'});
+      var template =
+        '<md-autocomplete ' +
+            'md-selected-item="selectedItem" ' +
+            'md-search-text="searchText" ' +
+            'md-items="item in match(searchText)" ' +
+            'md-item-text="item.display" ' +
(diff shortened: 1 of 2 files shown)


src/components/autocomplete/autocomplete.spec.js
```diff
@@ -374,6 +374,31 @@ describe('<md-autocomplete>', function() {
...
+            'placeholder="placeholder">' +
+          '<span md-highlight-text="searchText">{{item.display}}</span>' +
+        '</md-autocomplete>';
+
+      var element = compile(template, scope);
+      var input = element.find('input');
+
+      expect(input.attr('ng-trim')).toBe("false");
+
+      scope.$apply('searchText = "Start           End"');
```
ThomasBurleson: ngTrim does not trim whitespace inside does it? Only start/end?

DevVersion: Yeah that's a valid one. I was a bit confused because browsers also trim whitespace between text nodes.

> http://codepen.io/DevVersion/pen/pEoVjA?editors=1111

PR #9497, opened 2016-09-02
fix(variables): rem function should use global $font-size variable

### LINKED ISSUE #9486: Could $font-size be removed from @function rem($multiplier)
As Angular Material doesn't account for density I've used the following code to set the density for desktop apps (0.75x):

$font-size: 8px;
$baseline-grid: 6px;
@import "../../../bower_components/angular-material/modules/scss/angular-material.scss";

This only works if the $font-size variable is removed from the rem() function. The sass still compiles since "$font-size: 10px !default;" is defined before the rem() function is used for the first time.

The thing is when the variable is defined in the rem() function I can not override is from outside

DevVersion:
The `rem` SCSS function currently uses a hard-coded font size variable, which makes it impossible for developers to overwrite the global font-size.

> The rem function should be moved behind the initialization of the global `$font-size` variable, and should use this one.

Fixes #9486.

### DIFF
--- a/src/core/style/variables.scss
+++ b/src/core/style/variables.scss
@@ -1,14 +1,12 @@
-// Function
-//-- Must be defined before variables
-@function rem($multiplier) {
-  $font-size: 10px;
-  @return $multiplier * $font-size;
-}
-
 // Typography
 // ------------------------------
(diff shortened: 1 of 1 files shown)


dehart:
<md-datepicker> and <md-select> still use hard-coded size in px

PR #9512, opened 2016-09-03
fix(sidenav): notify child components when the element is opened

### LINKED ISSUE #7309: virtualRepeat not rendering in md-sidenav
The  virtual-repeat shows when the sidenav is initially rendered open, but if the sidenav is initially rendered closed, and opened programatically then the virtual-repeat does not not render.

crisbeto:
Triggers a resize event whenever a sidenav is opened. This allows components like virtualRepeat and textarea to resize properly.

Fixes #7309.

@topherfangio you and @DevVersion had a discussion on the issue about having a more generic solution to this. What do you think about having the changes from here until that happens?

### DIFF
--- a/src/components/sidenav/sidenav.js
+++ b/src/components/sidenav/sidenav.js
@@ -243,7 +243,8 @@ function SidenavFocusDirective() {
  *   - `<md-sidenav md-is-locked-open="$mdMedia('min-width: 1000px')"></md-sidenav>`
  *   - `<md-sidenav md-is-locked-open="$mdMedia('sm')"></md-sidenav>` (locks open on small screens)
  */
-function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate, $compile, $parse, $log, $q, $document) {
+function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming,
+  $animate, $compile, $parse, $log, $q, $document, $window) {
   return {
     restrict: 'E',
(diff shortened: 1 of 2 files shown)


DevVersion:
I think this is good for now. But I would not close that issue as for this PR.

topherfangio:
LGTM 👍

src/components/sidenav/sidenav.js
```diff
@@ -365,6 +367,8 @@ function SidenavDirective($mdMedia, $mdUtil, $mdConstant, $mdTheming, $animate,
       ]).then(function() {
         // Perform focus when animations are ALL done...
         if (scope.isOpen) {
+          // Notifies child components that the sidenav was opened.
+          ngWindow.triggerHandler('resize');
```
ThomasBurleson: So after the SideNav open animation finishes, you fire a `resize` event that causes reflows for the entire window?

crisbeto: It's not really a resize event, it just fires all of the resize handlers. An alternative to this can be to fire off a `$md-resize` event which is only for the virtual repeaters. I decided to go with this so other components could take advantage of it.

ThomasBurleson: The handlers have to be, however, attached to the `window` right ?

crisbeto: True, however we can assume that all the handlers are registered on the window, because the "resize" event doesn't work on anything else.

Issue #9506, opened 2016-09-02
md-tooltip wont work until scrolled


djfd:
**Actual Behavior**: 
 -  `What is the issue? *` a set of md-tooltips sharing their visibility state through md-visible attribute (initially set to true) do not disappear on blur. After scrolling (and thus disappearing) they become to operate as expected

UPD. adding ng-cloak to outer DIV gives unexpected result: both tooltips are shown in wrong location
 -  `What is the expected behavior?` it is expected that they have to disappear on blur.

**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:` http://codepen.io/djFD/pen/rrNQjZ http://s.codepen.io/djFD/debug/rrNQjZ
 -  `Details:` right after page load hover on and off over gray stripes, nothing happens, tooltips still visible. now scroll a few, then play with hover/blur. all works fine at this moment

**Angular Versions**: *
 -  `Angular Version:`1.5.7
 -  `Angular Material Version:`1.1.0

**Additional Information**:
 -  `Browser Type: *`firefox
 -  `Browser Version: *`48.0
 -  `OS: *`linux
 -  `Stack Traces:`nope

crisbeto:
After looking into it, it seems like this behaviour is intentional. The point of `md-visible` is to allow programmatic toggling of the tooltip, which is why it doesn't bind the "leave" events (the exception being "scroll"), in order to prevent the user from closing it manually. E.g. you can try toggling the "Insert drive" checkbox in [the demo](https://material.angularjs.org/HEAD/demo/tooltip) and hovering over the tooltip container.

djfd:
well, maybe intentional, thank you for clarification. But what about misplaced tootlips in a presence of ng-cloak on containing div?

crisbeto:
That's because the parent still hasn't been laid out when the tooltip is being positioned. This is a little tricky to solve without layout thrashing, because Material can't know in advance when the parent will become visible. It's possible to solve by watching for the element's `offsetParent`, however [each read on that property can cause a reflow](https://gist.github.com/paulirish/5d52fb081b3570c81e3a).

djfd:
hmm, I could not agree, but too lazy now to open new issue. please check http://codepen.io/djFD/pen/EgxqjL all the same conditions (ng-cloak) but works correct

crisbeto:
[...]

PR #9514, opened 2016-09-04
perf(tooltip): reduce amount of event listeners on the window

crisbeto:
Until now, each tooltip used to register 3 events on the window (scroll, resize and blur), even if it isn't being displayed. These can really add up on a page with a lot of tooltips. This change introduces a service that keeps track of the event handlers and only registers 1 global handler on the window for each event.

Referencing #9506.

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -3,8 +3,9 @@
  * @name material.components.tooltip
  */
 angular
-    .module('material.components.tooltip', [ 'material.core' ])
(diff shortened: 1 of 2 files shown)


topherfangio:
This looks good, but I think it would be nice to have some unit tests. Mostly for the service, but a simple one that ensures the tooltip properly deregisters the handlers on scope destroy would also be great (i.e. mock the service and make sure the dereg call is made).

crisbeto:
Updated to include tests @topherfangio

PR #9516, opened 2016-09-04
feat(compiler): pass $element to controller instantiation

### LINKED ISSUE #9507: mdCompiler: Pass $element to controller
As same as in a DDO the mdCompiler should pass the $element to the instantiated controller.

Thanks to @daniel-nagy for that issue.

> I asked Daniel if he want to work on that, so just having this issue as for tracking.

DevVersion:
* Pass the `$element` local to the instantiated controller, as same as a normal Angular 1.x DDO.

Closes #9507.

### DIFF
--- a/src/core/services/compiler/compiler.js
+++ b/src/core/services/compiler/compiler.js
@@ -61,7 +61,7 @@ function mdCompilerService($q, $templateRequest, $injector, $compile, $controlle
     *   - `link` - `{function(scope)}`: A link function, which, when called, will compile
     *     the element and instantiate the provided controller (if given).
     *   - `locals` - `{object}`: The locals which will be passed into the controller once `link` is
-    *     called. If `bindToController` is true, they will be coppied to the ctrl instead
+    *     called. If `bindToController` is true, they will be copied to the ctrl instead
     *   - `bindToController` - `bool`: bind the locals to the controller, instead of passing them in.
     */
   this.compile = function(options) {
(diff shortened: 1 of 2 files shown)


src/core/services/compiler/compiler.js
```diff
@@ -112,12 +112,19 @@ function mdCompilerService($q, $templateRequest, $injector, $compile, $controlle
         link: function link(scope) {
           locals.$scope = scope;
 
-          //Instantiate controller if it exists, because we have scope
+          // Instantiate controller if it exists, because we have scope
           if (controller) {
-            var invokeCtrl = $controller(controller, locals, true, controllerAs);
+
+            var ctrlLocals = angular.extend(locals, {
+              $element: element
```
ThomasBurleson: What about the `$attr`?

DevVersion: That's unfortunately not possible, because the `Attributes` constructor is internal-only.

> https://github.com/angular/angular.js/blob/master/src/ng/compile.js#L1508

Also doesn't make any sense to have the `$attrs` here, because the template is always static, and the user can't specify attributes dynamically.

crisbeto: `$attrs` might make sense once we start supporting content elements.

DevVersion: It's still not possible. Though `content` elements are still not dynamic, those are just to avoid compilation each time and are `pre-rendered` elements.

DevVersion:
@ThomasBurleson @crisbeto Any other feedback to address ?

crisbeto:
Nope, it LGTM.

PR #9517, opened 2016-09-04
docs(datepicker): split and clean up demos

crisbeto:
* Splits up the datepicker demos into two: one for standard usage and one for validations.
* Switches the datepicker demos to use controllerAs.
* Adds examples for a couple more options.


EladBezalel:
LGTM

PR #9531, opened 2016-09-06
update(css): components should use rem font-sizes

### LINKED ISSUE #9498: Component font-sizes should use the internal rem function
Most of the components use the internal `rem` function, which uses variables, which can be overwritten globally (very useful)

Some components do not use the `rem` function, and just use a hard-coded value of the size.

If a developer now changes the `$font-size` variable, some components like the `menu` are not changing anything.

**Affected Components (per RegExp)**
- Autocomplete
- Calendar
- Grid List
- Tabs
- Menu
- Slider
- Select
- Toast
- Card
- Input
- Flat Input Mixin

cc. @ThomasBurleson

DevVersion:
- The components should use our internal `rem` function to allow developers to globally overwrite the font of all components with a native-like `rem` behavior.
- Also includes extra variables for the font-size instead of hard-coded values, so developer can overwrite a _specific_ font of a component / part.

**FYI**: Ignore the prefixes of the variables. Right now those are just made consistent to the previous ones, but it is planned to prefix everything with `md-`

Closes #9498.

### DIFF
--- a/src/components/autocomplete/autocomplete.scss
+++ b/src/components/autocomplete/autocomplete.scss
@@ -1,5 +1,7 @@
 // The default item height is also specified in the JavaScript.
 $md-autocomplete-item-height: 48px !default;
+$md-autocomplete-button-font-size: rem(1.2) !default;
+$md-autocomplete-dropdown-font-size: rem(1.4) !default;
 
 @keyframes md-autocomplete-list-out {
   0% {
@@ -140,7 +142,7 @@ md-autocomplete {
     border: none;
     border-radius: 50%;
     padding: 0;
-    font-size: 12px;
+    font-size: $md-autocomplete-button-font-size;
     background: transparent;
     margin: auto 5px;
     &:after {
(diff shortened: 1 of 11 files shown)


jelbourn:
This seems to be affecting something in one of the consuming apps that is definitely not a material components, so I'm holding off on this one to investigate.

DevVersion:
Cool thx in advance :) Also why is this considered as a Breaking Change? Shouldn't be in theory one.

jelbourn:
Re-ran this today, now causing way more screenshot diff failures. @DevVersion @ThomasBurleson I think we should deem this PR too risky to merge.

DevVersion:
Alright. If it causes issues then I agree with closing that. 

There might be also some better approach for the REM units (and how they determine based on the body)

ThomasBurleson:
@jelbourn, @DevVersion - moved to Milestone 1.2.0.

dinglittle:
now we use the material develop our web on weixin , but can't use the rem , what can i do about img like log icon ? help me , wish reply

Splaktar:
This change will just have too much of a breaking impact on the library at this time.

PR #9538, opened 2016-09-07
feat(panel): Panel grouping

### LINKED ISSUE #8971: mdPanel: Implement group logic.
### Tasks

Add this config option to the panel:
```
 * TODO(ErinCoughlan): Add the following config options.
 *   - `groupName` - `{string=}`: Name of panel groups. This group name is
 *     used for configuring the number of open panels and identifying specific
 *     behaviors for groups. For instance, all tooltips will be identified
 *     using the same groupName.
```

Keep track of what panels are opened for a particular group (in order).

Implement/Extend the following methods:
```
/**
 * @ngdoc method
 * @name $mdPanel#setGroupMaxOpen
 * @description
 * Sets the maximum number of panels in a group that can be opened at a given
 * time.
 *
 * @param groupName {string} The name of the group to configure.
 * @param maxOpen {number} The max number of panels that can be opened.
 */
```
`setGroupMaxOpen` is a global configuration for a specific group. When this max is reached, the panel should close the panel that was first opened, keeping only a specific number of panels open at a time. 

An alternative to closing the first panel is to close the panel the user interacted with first. For example, assume there is a max of 2 panels open. The user interacts with panel 1, opens panel 2, goes back to panel 1, then opens panel 3. Panel 2 should close because it was interacted with the farthest back. I can see benefits to either implementation and I think either is understandable from a user's point of view. Perhaps make it configurable.

`clickOutsideToClose` and `pressEscapeToClose` should optionally close either all of the panels in a group or only the last panel. Add a config option for specifying which behavior is desired.

### Rationale for this Feature

Grouping allows the user to specify logic for a collection of panels. For instance, all popups in the header can be considered one group with a max of one popup open at the same time. Dialogs within dialogs are another example of groups. A "dialog group" may be able to have a max of 3 opened at a time, and then after this point, old dialogs close. This prevents the panel from storing a ridiculous amount of data in memory or polluting the DOM with several panels. This helps performance and gives the user more flexibility about how to use panels.

bradrich:
Panels are now capable of being grouped together by any designation the user sees fit. Grouping allows the user to specify logic for a collection of panels. For instance, all popups in the header can be considered one group with a maximum of one popup open at a time. Dialogs within dialogs are another example of groups. This implementation prevents the panel from storing a ridiculous amount of data in memory or polluting the DOM with several panels. This helps performance and gives the user more flexibility about how to use panels.

Fixes #8971

Ping @ErinCoughlan

### DIFF
--- a/src/components/panel/demoGroups/index.html
+++ b/src/components/panel/demoGroups/index.html
@@ -1,0 +1,46 @@
+<div ng-controller="PanelGroupsCtrl as ctrl" ng-cloak>
+
+  <md-toolbar class="md-accent">
+    <div class="md-toolbar-tools">
+      <md-button
+        class="md-icon-button"
+        aria-label="Settings"
+        ng-click="ctrl.showToolbarMenu($event, ctrl.settings)">
+        <md-icon md-svg-icon="img/icons/menu.svg"></md-icon>
+      </md-button>
+      <h2>Toolbar with grouped panels (Maximum open: 2)</h2>
+      <span flex></span>
+      <md-button
+        class="md-icon-button"
+        aria-label="Favorite"
+        ng-click="ctrl.showToolbarMenu($event, ctrl.favorite)">
+        <md-icon md-svg-icon="img/icons/favorite.svg"></md-icon>
+      </md-button>
+      <md-button
+        class="md-icon-button"
+        aria-label="More"
+        ng-click="ctrl.showToolbarMenu($event, ctrl.more)">
+        <md-icon md-svg-icon="img/icons/more_vert.svg"></md-icon>
+      </md-button>
+    </div>
+  </md-toolbar>
+
+  <md-content layout-padding>
+    <p>
+      Panels can be added to a group. Groups are used to configure specific
+      behaviors on multiple panels. To add a panel to a group, use the
+      <code>$mdPanel.newPanelGroup</code> method, or simply add a group name
+      to the configuration object passed into the <code>$mdPanel.create</code>
+      method.
+    </p>
+    <p>
+      Grouping allows for methods to be applied to several panels at once, i.e.
+      closing all panels within the toolbar group, or destroying all panels
+      within a dialog group. With the <code>maxOpen</code> property, you can
+      also limit the number of panels allowed open within a specific group. This
+      can be useful in limiting the number of menu panels allowed open at a
+      time, etc.
+    </p>
+  </md-content>
+
+</div>
--- a/src/components/panel/demoGroups/script.js
+++ b/src/components/panel/demoGroups/script.js
@@ -1,0 +1,88 @@
+(function() {
+  'use strict';
+
+  angular
+    .module('panelGroupsDemo', ['ngMaterial'])
+    .controller('PanelGroupsCtrl', PanelGroupsCtrl)
+    .controller('PanelMenuCtrl', PanelMenuCtrl);
+
+  function PanelGroupsCtrl($mdPanel) {
+    this.settings = {
+      name: 'settings',
+      items: [
+        'Home',
+        'About',
+        'Contact'
+      ]
+    };
+    this.favorite = {
+      name: 'favorite',
+      items: [
+        'Add to Favorites'
+      ]
+    };
+    this.more = {
+      name: 'more',
+      items: [
+        'Account',
+        'Sign Out'
+      ]
+    };
+
+    this.showToolbarMenu = function($event, menu) {
+      var template = '' +
+          '<div class="menu-panel" md-whiteframe="4">' +
+          '  <div class="menu-content">' +
+          '    <div class="menu-item" ng-repeat="item in ctrl.items">' +
+          '      <button class="md-button">' +
+          '        <span>{{item}}</span>' +
+          '      </button>' +
+          '    </div>' +
+          '    <md-divider></md-divider>' +
+          '    <div class="menu-item">' +
+          '      <button class="md-button" ng-click="ctrl.closeMenu()">' +
+          '        <span>Close Menu</span>' +
+          '      </button>' +
+          '    </div>' +
+          '  </div>' +
+          '</div>';
+
+      var position = $mdPanel.newPanelPosition()
+          .relativeTo($event.srcElement)
+          .addPanelPosition(
+            $mdPanel.xPosition.ALIGN_START,
+            $mdPanel.yPosition.BELOW
+          );
+
+      $mdPanel.newPanelGroup('toolbar', {
+        maxOpen: 2
+      });
+
+      var config = {
+        id: 'toolbar_' + menu.name,
+        attachTo: angular.element(document.body),
+        controller: PanelMenuCtrl,
+        controllerAs: 'ctrl',
+        template: template,
+        position: position,
+        panelClass: 'menu-panel-container',
+        locals: {
+          items: menu.items
+        },
+        openFrom: $event,
+        focusOnOpen: false,
+        zIndex: 100,
+        propagateContainerEvents: true,
+        groupName: 'toolbar'
+      };
+
+      $mdPanel.open(config);
+    }
+  }
+
+  function PanelMenuCtrl(mdPanelRef) {
+    this.closeMenu = function() {
+      mdPanelRef && mdPanelRef.close();
+    }
+  }
+})();
--- a/src/components/panel/demoGroups/style.global.css
+++ b/src/components/panel/demoGroups/style.global.css
@@ -1,0 +1,77 @@
+.menu-panel-container {
+  pointer-events: auto;
+}
+
+.menu-panel {
+  width: 256px;
+  background-color: #fff;
+  border-radius: 4px;
+}
+
+.menu-panel .menu-divider {
+  width: 100%;
+  height: 1px;
+  min-height: 1px;
+  max-height: 1px;
+  margin-top: 4px;
+  margin-bottom: 4px;
+  background-color: rgba(0, 0, 0, 0.11);
+}
+
+.menu-panel .menu-content {
+  display: flex;
+  flex-direction: column;
+  padding: 8px 0;
+  max-height: 305px;
+  overflow-y: auto;
+  min-width: 256px;
+}
+
+.menu-panel .menu-item {
+  display: flex;
+  flex-direction: row;
+  min-height: 48px;
+  height: 48px;
+  align-content: center;
+  justify-content: flex-start;
+}
+.menu-panel .menu-item > * {
+  width: 100%;
+  margin: auto 0;
+  padding-left: 16px;
+  padding-right: 16px;
+}
+.menu-panel .menu-item > a.md-button {
+  padding-top: 5px;
+}
+.menu-panel .menu-item > .md-button {
+  display: inline-block;
+  border-radius: 0;
+  margin: auto 0;
+  font-size: 15px;
+  text-transform: none;
+  font-weight: 400;
+  height: 100%;
+  padding-left: 16px;
+  padding-right: 16px;
+  width: 100%;
+  text-align: left;
+}
+.menu-panel .menu-item > .md-button::-moz-focus-inner {
+  padding: 0;
+  border: 0;
+}
+.menu-panel .menu-item > .md-button md-icon {
+  margin: auto 16px auto 0;
+}
+.menu-panel .menu-item > .md-button p {
+  display: inline-block;
+  margin: auto;
+}
+.menu-panel .menu-item > .md-button span {
+  margin-top: auto;
+  margin-bottom: auto;
+}
+.menu-panel .menu-item > .md-button .md-ripple-container {
+  border-radius: inherit;
+}
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -155,10 +155,13 @@ angular
  *     on when the panel closes. This is commonly the element which triggered
  *     the opening of the panel. If you do not use `origin`, you need to control
  *     the focus manually.
+ *   - `groupName` - `{string=}`: Name of a panel group. This group name is used
+ *     for configuring the number of open panels and identifying specific
+ *     behaviors for groups. For instance, all tooltips could be identified
+ *     using the same groupName.
  *
  * @returns {!MdPanelRef} panelRef
  */
-
 
 /**
  * @ngdoc method
@@ -176,7 +179,6 @@ angular
  *     to an instance of the panel.
  */
 
-
 /**
  * @ngdoc method
  * @name $mdPanel#newPanelPosition
@@ -187,7 +189,6 @@ angular
  * @returns {!MdPanelPosition} panelPosition
  */
 
-
 /**
  * @ngdoc method
  * @name $mdPanel#newPanelAnimation
@@ -196,6 +197,37 @@ angular
  * the animation config object.
  *
  * @returns {!MdPanelAnimation} panelAnimation
+ */
+
+/**
+ * @ngdoc method
+ * @name $mdPanel#newPanelGroup
+ * @description
+ * Creates a panel group and adds it to a tracked list of panel groups.
+ *
+ * @param {string} groupName Name of the group to create.
+ * @param {!Object=} config Specific configuration object that may contain the
+ *     following properties:
+ *
+ *   - `maxOpen` - `{number=}`: The maximum number of panels that are allowed to
+ *     be open within a defined panel group.
+ *
+ * @returns {!Object<string,
+ *     {panels: !Array<!MdPanelRef>,
+ *     openPanels: !Array<!MdPanelRef>,
+ *     maxOpen: number}>} panelGroup
+ */
+
+/**
+ * @ngdoc method
+ * @name $mdPanel#setGroupMaxOpen
+ * @description
+ * Sets the maximum number of panels in a group that can be opened at a given
+ * time.
+ *
+ * @param {string} groupName The name of the group to configure.
+ * @param {number} maxOpen The maximum number of panels that can be
+ *     opened. Infinity can be passed in to remove the maxOpen limit.
  */
 
 
@@ -355,9 +387,28 @@ angular
 
 /**
  * @ngdoc method
+ * @name MdPanelRef#addToGroup
+ * @description
+ * Adds a panel to a group if the panel does not exist within the group already.
+ * A panel can only exist within a single group.
+ *
+ * @param {string} groupName The name of the group to add the panel to.
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#removeFromGroup
+ * @description
+ * Removes a panel from a group if the panel exists within that group. The group
+ * must be created ahead of time.
+ *
+ * @param {string} groupName The name of the group.
+ */
+
+/**
+ * @ngdoc method
  * @name MdPanelRef#registerInterceptor
  * @description
- *
  * Registers an interceptor with the panel. The callback should return a promise,
  * which will allow the action to continue when it gets resolved, or will
  * prevent an action if it is rejected. The interceptors are called sequentially
@@ -374,7 +425,6 @@ angular
  * @ngdoc method
  * @name MdPanelRef#removeInterceptor
  * @description
- *
  * Removes a registered interceptor.
  *
  * @param {string} type Type of interceptor to be removed.
@@ -386,7 +436,6 @@ angular
  * @ngdoc method
  * @name MdPanelRef#removeAllInterceptors
  * @description
- *
  * Removes all interceptors. If a type is supplied, only the
  * interceptors of that type will be cleared.
  *
@@ -762,8 +811,19 @@ function MdPanelService($rootElement, $rootScope, $injector, $window) {
   /** @private @const */
   this._$window = $window;
 
+  /** @private @const */
+  this._$mdUtil = this._$injector.get('$mdUtil');
+
   /** @private {!Object<string, !MdPanelRef>} */
   this._trackedPanels = {};
+
+  /**
+   * @private {!Object<string,
+   *     {panels: !Array<!MdPanelRef>,
+   *     openPanels: !Array<!MdPanelRef>,
+   *     maxOpen: number}>}
+   */
+  this._groups = Object.create(null);
 
   /**
    * Default animations that can be used within the panel.
@@ -807,9 +867,9 @@ MdPanelService.prototype.create = function(config) {
     return this._trackedPanels[config.id];
   }
 
-  // If no ID is set within the passed-in config, then create an arbitrary ID.
   this._config = {
-    id: config.id || 'panel_' + this._$injector.get('$mdUtil').nextUid(),
+    // If no ID is set within the passed-in config, then create an arbitrary ID.
+    id: config.id || 'panel_' + this._$mdUtil.nextUid(),
     scope: this._$rootScope.$new(true),
     attachTo: this._$rootElement
   };
@@ -818,6 +878,10 @@ MdPanelService.prototype.create = function(config) {
   var panelRef = new MdPanelRef(this._config, this._$injector);
   this._trackedPanels[config.id] = panelRef;
   this._config.scope.$on('$destroy', angular.bind(panelRef, panelRef.detach));
+
+  if (this._config.groupName) {
+    panelRef.addToGroup(this._config.groupName);
+  }
 
   return panelRef;
 };
@@ -853,6 +917,78 @@ MdPanelService.prototype.newPanelPosition = function() {
  */
 MdPanelService.prototype.newPanelAnimation = function() {
   return new MdPanelAnimation(this._$injector);
+};
+
+
+/**
+ * Creates a panel group and adds it to a tracked list of panel groups.
+ * @param groupName {string} Name of the group to create.
+ * @param config {!Object=} Specific configuration object that may contain the
+ *     following properties:
+ *
+ *   - `maxOpen` - `{number=}`: The maximum number of panels that are allowed
+ *     open within a defined panel group.
+ *
+ * @returns {!Object<string,
+ *     {panels: !Array<!MdPanelRef>,
+ *     openPanels: !Array<!MdPanelRef>,
+ *     maxOpen: number}>} panelGroup
+ */
+MdPanelService.prototype.newPanelGroup = function(groupName, config) {
+  if (!this._groups[groupName]) {
+    config = config || {};
+    var group = {
+      panels: [],
+      openPanels: [],
+      maxOpen: config.maxOpen > 0 ? config.maxOpen : Infinity
+    };
+    this._groups[groupName] = group;
+  }
+  return this._groups[groupName];
+};
+
+
+/**
+ * Sets the maximum number of panels in a group that can be opened at a given
+ * time.
+ * @param {string} groupName The name of the group to configure.
+ * @param {number} maxOpen The maximum number of panels that can be
+ *     opened. Infinity can be passed in to remove the maxOpen limit.
+ */
+MdPanelService.prototype.setGroupMaxOpen = function(groupName, maxOpen) {
+  if (this._groups[groupName]) {
+    this._groups[groupName].maxOpen = maxOpen;
+  } else {
+    throw new Error('mdPanel: Group does not exist yet. Call newPanelGroup().');
+  }
+};
+
+
+/**
+ * Determines if the current number of open panels within a group exceeds the
+ * limit of allowed open panels.
+ * @param {string} groupName The name of the group to check.
+ * @returns {boolean} true if open count does exceed maxOpen and false if not.
+ * @private
+ */
+MdPanelService.prototype._openCountExceedsMaxOpen = function(groupName) {
+  if (groupName && this._groups[groupName]) {
+    var group = this._groups[groupName];
+    return group.maxOpen > 0 && group.openPanels.length > group.maxOpen;
+  }
+  return false;
+};
+
+
+/**
+ * Closes the first open panel within a specific group.
+ * @param {string} groupName The name of the group.
+ * @private
+ */
+MdPanelService.prototype._closeFirstOpenedPanel = function(groupName) {
+  if (groupName) {
+    this._groups[groupName].openPanels[0].close();
+  }
 };
 
 
@@ -985,9 +1121,15 @@ MdPanelRef.prototype.open = function() {
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
     var show = self._simpleBind(self.show, self);
+    var checkGroupMaxOpen = function() {
+      if (self._$mdPanel._openCountExceedsMaxOpen(self.config.groupName)) {
+        self._$mdPanel._closeFirstOpenedPanel(self.config.groupName);
+      }
+    };
 
     self.attach()
         .then(show)
+        .then(checkGroupMaxOpen)
         .then(done)
         .catch(reject);
   });
@@ -1031,9 +1173,9 @@ MdPanelRef.prototype.attach = function() {
     var done = self._done(resolve, self);
     var onDomAdded = self.config['onDomAdded'] || angular.noop;
     var addListeners = function(response) {
-        self.isAttached = true;
-        self._addEventListeners();
-        return response;
+      self.isAttached = true;
+      self._addEventListeners();
+      return response;
     };
 
     self._$q.all([
@@ -1101,6 +1243,9 @@ MdPanelRef.prototype.detach = function() {
  * Destroys the panel. The Panel cannot be opened again after this.
  */
 MdPanelRef.prototype.destroy = function() {
+  if (this.config.groupName) {
+    this.removeFromGroup(this.config.groupName);
+  }
   this.config.scope.$destroy();
   this.config.locals = null;
   this._interceptors = null;
@@ -1115,7 +1260,7 @@ MdPanelRef.prototype.destroy = function() {
 MdPanelRef.prototype.show = function() {
   if (!this.panelContainer) {
     return this._$q(function(resolve, reject) {
-      reject('Panel does not exist yet. Call open() or attach().');
+      reject('mdPanel: Panel does not exist yet. Call open() or attach().');
     });
   }
 
@@ -1132,11 +1277,18 @@ MdPanelRef.prototype.show = function() {
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
     var onOpenComplete = self.config['onOpenComplete'] || angular.noop;
+    var addToGroupOpen = function() {
+      if (self.config.groupName) {
+        var group = self._$mdPanel._groups[self.config.groupName];
+        group.openPanels.push(self);
+      }
+    };
 
     self._$q.all([
       self._backdropRef ? self._backdropRef.show() : self,
       animatePromise().then(function() { self._focusOnOpen(); }, reject)
     ]).then(onOpenComplete)
+      .then(addToGroupOpen)
       .then(done)
       .catch(reject);
   });
@@ -1151,7 +1303,7 @@ MdPanelRef.prototype.show = function() {
 MdPanelRef.prototype.hide = function() {
   if (!this.panelContainer) {
     return this._$q(function(resolve, reject) {
-      reject('Panel does not exist yet. Call open() or attach().');
+      reject('mdPanel: Panel does not exist yet. Call open() or attach().');
     });
   }
 
@@ -1164,7 +1316,18 @@ MdPanelRef.prototype.hide = function() {
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
     var onRemoving = self.config['onRemoving'] || angular.noop;
-
+    var hidePanel = function() {
+      self.panelContainer.addClass(MD_PANEL_HIDDEN);
+    };
+    var removeFromGroupOpen = function() {
+      if (self.config.groupName) {
+        var group = self._$mdPanel._groups[self.config.groupName];
+        var index = group.openPanels.indexOf(self);
+        if (index > -1) {
+          group.openPanels.splice(index, 1);
+        }
+      }
+    };
     var focusOnOrigin = function() {
       var origin = self.config['origin'];
       if (origin) {
@@ -1172,15 +1335,12 @@ MdPanelRef.prototype.hide = function() {
       }
     };
 
-    var hidePanel = function() {
-      self.panelContainer.addClass(MD_PANEL_HIDDEN);
-    };
-
     self._$q.all([
       self._backdropRef ? self._backdropRef.hide() : self,
       self._animateClose()
           .then(onRemoving)
           .then(hidePanel)
+          .then(removeFromGroupOpen)
           .then(focusOnOrigin)
           .catch(reject)
     ]).then(done, reject);
@@ -1201,13 +1361,14 @@ MdPanelRef.prototype.hide = function() {
  */
 MdPanelRef.prototype.addClass = function(newClass, toElement) {
   this._$log.warn(
-      'The addClass method is in the process of being deprecated. ' +
+      'mdPanel: The addClass method is in the process of being deprecated. ' +
       'Full deprecation is scheduled for the Angular Material 1.2 release. ' +
       'To achieve the same results, use the panelContainer or panelEl ' +
       'JQLite elements that are referenced in MdPanelRef.');
 
   if (!this.panelContainer) {
-    throw new Error('Panel does not exist yet. Call open() or attach().');
+    throw new Error(
+        'mdPanel: Panel does not exist yet. Call open() or attach().');
   }
 
   if (!toElement && !this.panelContainer.hasClass(newClass)) {
@@ -1231,13 +1392,14 @@ MdPanelRef.prototype.addClass = function(newClass, toElement) {
  */
 MdPanelRef.prototype.removeClass = function(oldClass, fromElement) {
   this._$log.warn(
-      'The removeClass method is in the process of being deprecated. ' +
+      'mdPanel: The removeClass method is in the process of being deprecated. ' +
       'Full deprecation is scheduled for the Angular Material 1.2 release. ' +
       'To achieve the same results, use the panelContainer or panelEl ' +
       'JQLite elements that are referenced in MdPanelRef.');
 
   if (!this.panelContainer) {
-    throw new Error('Panel does not exist yet. Call open() or attach().');
+    throw new Error(
+        'mdPanel: Panel does not exist yet. Call open() or attach().');
   }
 
   if (!fromElement && this.panelContainer.hasClass(oldClass)) {
@@ -1261,13 +1423,14 @@ MdPanelRef.prototype.removeClass = function(oldClass, fromElement) {
  */
 MdPanelRef.prototype.toggleClass = function(toggleClass, onElement) {
   this._$log.warn(
-      'The toggleClass method is in the process of being deprecated. ' +
+      'mdPanel: The toggleClass method is in the process of being deprecated. ' +
       'Full deprecation is scheduled for the Angular Material 1.2 release. ' +
       'To achieve the same results, use the panelContainer or panelEl ' +
       'JQLite elements that are referenced in MdPanelRef.');
 
   if (!this.panelContainer) {
-    throw new Error('Panel does not exist yet. Call open() or attach().');
+    throw new Error(
+        'mdPanel: Panel does not exist yet. Call open() or attach().');
   }
 
   if (!onElement) {
@@ -1382,7 +1545,8 @@ MdPanelRef.prototype._addStyles = function() {
  */
 MdPanelRef.prototype.updatePosition = function(position) {
   if (!this.panelContainer) {
-    throw new Error('Panel does not exist yet. Call open() or attach().');
+    throw new Error(
+        'mdPanel: Panel does not exist yet. Call open() or attach().');
   }
 
   this.config['position'] = position;
@@ -1662,7 +1826,8 @@ MdPanelRef.prototype._animateOpen = function() {
     var done = self._done(resolve, self);
     var warnAndOpen = function() {
       self._$log.warn(
-          'MdPanel Animations failed. Showing panel without animating.');
+          'mdPanel: MdPanel Animations failed. ' +
+          'Showing panel without animating.');
       done();
     };
 
@@ -1694,7 +1859,8 @@ MdPanelRef.prototype._animateClose = function() {
     };
     var warnAndClose = function() {
       self._$log.warn(
-          'MdPanel Animations failed. Hiding panel without animating.');
+          'mdPanel: MdPanel Animations failed. ' +
+          'Hiding panel without animating.');
       done();
     };
 
@@ -1830,6 +1996,44 @@ MdPanelRef.prototype._done = function(callback, self) {
 };
 
 
+/**
+ * Adds a panel to a group if the panel does not exist within the group already.
+ * A panel can only exist within a single group.
+ * @param {string} groupName The name of the group.
+ */
+MdPanelRef.prototype.addToGroup = function(groupName) {
+  if (!this._$mdPanel._groups[groupName]) {
+    this._$mdPanel.newPanelGroup(groupName);
+  }
+
+  var group = this._$mdPanel._groups[groupName];
+  var index = group.panels.indexOf(this);
+
+  if (index < 0) {
+    group.panels.push(this);
+  }
+};
+
+
+/**
+ * Removes a panel from a group if the panel exists within that group. The group
+ * must be created ahead of time.
+ * @param {string} groupName The name of the group.
+ */
+MdPanelRef.prototype.removeFromGroup = function(groupName) {
+  if (!this._$mdPanel._groups[groupName]) {
+    throw new Error('mdPanel: The group ' + groupName + ' does not exist.');
+  }
+
+  var group = this._$mdPanel._groups[groupName];
+  var index = group.panels.indexOf(this);
+
+  if (index > -1) {
+    group.panels.splice(index, 1);
+  }
+};
+
+
 /*****************************************************************************
  *                               MdPanelPosition                             *
  *****************************************************************************/
@@ -1961,7 +2165,7 @@ MdPanelPosition.prototype._setPosition = function(position, value) {
     var positions = Object.keys(MdPanelPosition.absPosition).join()
         .toLowerCase();
 
-    throw new Error('Position must be one of ' + positions + '.');
+    throw new Error('mdPanel: Position must be one of ' + positions + '.');
   }
 
   this['_' +  position] = angular.isString(value) ? value : '0';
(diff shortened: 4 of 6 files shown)


src/components/panel/panel.js
```diff
@@ -806,6 +850,75 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+ * Creates a panel group and adds it to a tracked list of panel groups.
+ * @param groupName {string} Name of the group to create.
+ * @param config (!Object=) Specific configuration object that may contain the
+ *     following properties:
+ *
+ *   - `maxOpen` - `{number=}`: The maximum number of panels that are allowed
+ *     open within a defined panel group.
+ */
+MdPanelService.prototype.newPanelGroup = function(groupName, config) {
+  if (!this._groups[groupName]) {
```
crisbeto: I think that it might be better to use `hasOwnProperty` here. This might cause undefined behaviour in the edge case if somebody names their group `toString` or something similar.

src/components/panel/panel.js
```diff
@@ -923,9 +1043,23 @@ MdPanelRef.prototype.open = function() {
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
     var show = self._simpleBind(self.show, self);
+    var setOpened = function(response) {
+      self.isOpened = true;
+      return response;
+    };
+    var checkGroupMaxOpen = function(response) {
```
crisbeto: You can avoid having `return response` twice by doing it like this:

```js
 if (self.config.groupName) {
   self._$mdPanel._checkGroupMaxOpen(self.config.groupName);
 }
 
return response;
```

src/components/panel/panel.js
```diff
@@ -2550,3 +2702,28 @@ function getElement(el) {
...
+
+/**
+ * Moves an array element from one array position to another position.
+ * http://stackoverflow.com/questions/5306680
+ *     /move-an-array-element-from-one-array-position-to-another
+ * @param {number} oldIndex The original position of the array element.
+ * @param {number} newIndex The new position that the array element will move
+ *     to.
+ */
+Array.prototype.move = function(oldIndex, newIndex) {
```
crisbeto: I think that it's generally a bad idea to add stuff to the native prototype. We should turn this into a regular function.

src/components/panel/panel.js
```diff
@@ -2550,3 +2702,28 @@ function getElement(el) {
...
+  while (oldIndex < 0) {
```
crisbeto: Using a loop here and for `newIndex` seem like overkill. Couldn't you subtract the index and the array length and use that to figure out how many elements you need to add?

src/components/panel/panel.js
```diff
@@ -806,6 +850,75 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+ */
+MdPanelService.prototype.newPanelGroup = function(groupName, config) {
+  if (!this._groups[groupName]) {
+    config = config || {};
+    var group = {
+      panels: [],
+      maxOpen: config.maxOpen > 0 ? config.maxOpen : false
+    };
+    this._groups[groupName] = group;
+  }
```
crisbeto: A nice addition to this method could be to return the group that was created.

src/components/panel/panel.js
```diff
@@ -806,6 +850,75 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+MdPanelService.prototype._checkGroupMaxOpen = function(groupName) {
+  if (this._groups[groupName] && this._groups[groupName].maxOpen > 0) {
+    var openCount = 0,
+        firstOpenedPanel = null,
+        firstOpenedPanelIndex = null;
+
+    angular.forEach(this._groups[groupName].panels, function(panel, index) {
+      if (panel.isOpened) {
+        openCount++;
+        if (openCount === 1 && !firstOpenedPanel) {
```
crisbeto: Instead of checking against the variable that was just incremented (`openCount`), couldn't we just check that the `index === 0` to determine that it's first?

src/components/panel/panel.spec.js
```diff
@@ -1053,6 +1053,125 @@ describe('$mdPanel', function() {
     });
   });
 
+  describe('grouping logic:', function() {
```
DevVersion: Why is there a `:` in the describe? - This does not follow the normal jasmine pattern.

DevVersion: Ping before I delete my comment: It seems like this is with other describes here as well. This is odd.

src/components/panel/demoGroups/index.html
```diff
@@ -0,0 +1,44 @@
...
+  <md-content layout-padding>
+    <p>
+      Panels can be added to a group. Groups are used to configure specific
+      behaviors on multiple panels. To add a panel to a group, use the
+      <code>$mdPanel.newPanelGroup</code> method, or simply add a group name
+      to the configuration object passed into the <code>$mdPanel.create</code>
+      method.
+    </p>
+    <p>
+      Adding panels to groups prevents the panel API from storing a ridiculous
```
ErinCoughlan: reword - "rediculous amount of data" sounds too vague as does the added flexibility. This is your sales pitch for groups.

src/components/panel/demoGroups/script.js
```diff
@@ -0,0 +1,95 @@
...
+
+      $mdPanel.open(config);
+    }
+  }
+
+  function PanelMenuCtrl(mdPanelRef) {
+    this.closeDialog = closeDialog;
+
+    function closeDialog() {
+      mdPanelRef && mdPanelRef.close();
```
ErinCoughlan: destroy as well?

src/components/panel/demoGroups/script.js
```diff
@@ -0,0 +1,95 @@
...
+        propagateContainerEvents: true,
+        groupName: group
+      };
+
+      $mdPanel.open(config);
+    }
+  }
+
+  function PanelMenuCtrl(mdPanelRef) {
+    this.closeDialog = closeDialog;
```
ErinCoughlan: this.closeDialog = function() { ... };

src/components/panel/demoGroups/style.global.css
```diff
@@ -0,0 +1,80 @@
+.menu-panel-container {
+  pointer-events: all;
```
ErinCoughlan: Why does the user need to specify this? Isn't it part of panel?

src/components/panel/demoGroups/index.html
```diff
@@ -0,0 +1,44 @@
+<div ng-controller="PanelGroupsCtrl as ctrl" ng-cloak>
+
+  <md-toolbar class="md-accent">
+    <div class="md-toolbar-tools">
+      <md-button
+        class="md-icon-button"
+        aria-label="Settings"
+        ng-click="ctrl.showMenu($event, 'settings', 'toolbar')">
```
ErinCoughlan: Use constants from the controller instead of strings for 'settings', 'favorite', and 'more'. It's a better code example. (More robust, less likely for typos or errors.)

Also since all of these are "toolbar" groups, I'd expect a more common use case is to declare the group name directly in the open function rather then passing it in. I would think when a user calls open, it's mapped to one method that knows what group it should be.

src/components/panel/demoGroups/script.js
```diff
@@ -0,0 +1,95 @@
...
+        controller: PanelMenuCtrl,
+        controllerAs: 'ctrl',
+        template: template,
+        position: position,
+        panelClass: 'menu-panel-container',
+        locals: {
+          items: this.items
+        },
+        openFrom: $event,
+        clickOutsideToClose: false,
```
ErinCoughlan: Any of these that default, please remove from the config. It makes it unnecessarily complicated.

src/components/panel/panel.js
```diff
@@ -355,6 +382,16 @@ angular
  * @param {!MdPanelPosition} position
  */
 
+/**
+ * @ngdoc method
+ * @name MdPanelRef#addToGroup
+ * @description
+ * Add a panel to a group, either automatically during its original creation, or
```
ErinCoughlan: Add -> Adds

Remove the "automatically during its original creation". We only care about the public API in this description. The users shouldn't have to care what happens behind the scenes during creation.

src/components/panel/panel.js
```diff
@@ -721,6 +758,9 @@ function MdPanelService($rootElement, $rootScope, $injector, $window) {
   /** @private {!Object<string, !MdPanelRef>} */
   this._trackedPanels = {};
 
+  /** @private {!Object<string, !Object<{panels:Array<!MdPanelRef>, maxOpen:number}>>} */
```
ErinCoughlan: {!Object<string, {panels: !Array<!MdPanelRef>, maxOpen: number}>}

This is the correct type. You are explict about the object keys, so you don't write Object. Also fixed nullability on Array.

src/components/panel/panel.js
ErinCoughlan: Agreed, this will also make it easier to extend the group config later to include group-wide settings for `clickOutsideToClose` and other panel config options.

src/components/panel/panel.js
```diff
@@ -806,6 +850,75 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+      panels: [],
+      maxOpen: config.maxOpen > 0 ? config.maxOpen : false
+    };
+    this._groups[groupName] = group;
+  }
+};
+
+
+/**
+ * Set the maximum number of panels in a group that can be opened at a given
```
ErinCoughlan: Set -> Sets

src/components/panel/panel.js
```diff
@@ -806,6 +850,75 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+ * time.
+ * @param {string} groupName The name of the group to configure.
+ * @param {number|boolean} maxOpen The maximum number of panels that can be
+ *     opened. If false is passed in, the maximum limit will be removed.
+ */
+MdPanelService.prototype.setGroupMaxOpen = function(groupName, maxOpen) {
+  if (this._groups[groupName]) {
```
ErinCoughlan: If group doesn't exist, throw an error to create the group first?

src/components/panel/panel.js
```diff
@@ -806,6 +850,75 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+
+/**
+ * Determines if opening a panel within a group will exceed the limit of allowed
+ * open panels. If so, the first panel opened within the group will be closed.
+ * @param {string} groupName The name of the group to check.
+ * @private
+ */
+MdPanelService.prototype._checkGroupMaxOpen = function(groupName) {
+  if (this._groups[groupName] && this._groups[groupName].maxOpen > 0) {
+    var openCount = 0,
```
ErinCoughlan: This feels silly to do this tracking/iteration every time a panel is opened.

Since the only ways to open/close are through other methods, what about adding a field to the groups object for "currentNumOpen", which increments when a panel is opened with that groupId, and decrements when a panel is closed. Then this method simply becomes:

```
var group = this._groups[groupName];
return group.maxOpen < group.currentNumOpen;
```

src/components/panel/panel.js
ErinCoughlan: Another option is to make `checkGroupMaxOpen` pass when there isn't a group, which simplifies this method.

src/components/panel/panel.js
```diff
@@ -923,9 +1043,23 @@ MdPanelRef.prototype.open = function() {
...
+        self._$mdPanel._checkGroupMaxOpen(self.config.groupName);
+        return response;
+      } else {
+        return response;
+      }
+    };
 
     self.attach()
         .then(show)
+        .then(setOpened)
```
ErinCoughlan: Shouldn't you only `setOpened` once you successfully open a panel (after checking max is right).

src/components/panel/panel.js
```diff
@@ -923,9 +1043,23 @@ MdPanelRef.prototype.open = function() {
   return this._$q(function(resolve, reject) {
     var done = self._done(resolve, self);
     var show = self._simpleBind(self.show, self);
+    var setOpened = function(response) {
```
ErinCoughlan: Why are you passing a response? I don't think it's necessary. This will always resolve, so we shouldn't care.

src/components/panel/panel.js
```diff
@@ -1012,6 +1146,7 @@ MdPanelRef.prototype.detach = function() {
 
     self.panelContainer.remove();
     self.isAttached = false;
+    self.isOpened = false;
```
ErinCoughlan: This should be in close, not detach (unless you move setting it to true to attach instead of open, which I don't recommend). Open and close go together. Attach and detach go together.

src/components/panel/panel.js
```diff
@@ -1663,6 +1798,23 @@ MdPanelRef.prototype._done = function(callback, self) {
 };
 
 
+/**
+ * Adds a panel to a group if the panel does not already exist within the group.
+ * The only way for a panel to be apart of a group and then to be removed from
```
ErinCoughlan: Well this seems suspicious... Should we add a removeFromGroup method? I think it follows the same use case/user journey as dynamic adding.

src/components/panel/panel.spec.js
```diff
@@ -1053,6 +1053,125 @@ describe('$mdPanel', function() {
...
+        'group hasn\'t been created yet', function() {
+          var config = {
+            groupName: 'test'
+          };
+
+          openPanel(config);
+
+          expect($mdPanel._groups['test']).toExist();
+
+          closePanel();
```
ErinCoughlan: This should be in the `afterEach`. The test should end with the expects.

src/components/panel/panel.spec.js
```diff
@@ -1053,6 +1053,125 @@ describe('$mdPanel', function() {
...
+      var panelFromGroup = $mdPanel._groups['test'].panels[0];
+
+      expect(panel1).toEqual(panelFromGroup);
+    });
+
+    it('should add a panel to a group using the config option groupName',
+        function() {
+          $mdPanel.newPanelGroup('test');
+
+          expect($mdPanel._groups['test']).toExist();
```
ErinCoughlan: This expect doesn't matter. It's not relevant to the given tests (you've already tested that creating a group works). Ditto for all other tests besides the create newPanelGroup one.

src/components/panel/panel.spec.js
```diff
@@ -1053,6 +1053,125 @@ describe('$mdPanel', function() {
...
+    it('should add a panel to a group using the addToGroup method', function() {
+      $mdPanel.newPanelGroup('test');
+
+      expect($mdPanel._groups['test']).toExist();
+
+      var panel1 = $mdPanel.create(DEFAULT_CONFIG);
+
+      panel1.addToGroup('test');
+
+      var panelFromGroup = $mdPanel._groups['test'].panels[0];
```
ErinCoughlan: These statements need to be grouped in an intelligent way.

// Setup

// Execute

// Assert

So this test should look like this:
```
$mdPanel.newPanelGrou('test');
var panel = $mdPanel.create(DEFAULT_CONFIG);

panel.addToGroup('test');

var panelFromGroup = $mdPanel._groups['test'].panels[0];
expect(panel).toEqual(panelFromGroup);
```

src/components/panel/panel.spec.js
```diff
@@ -1053,6 +1053,125 @@ describe('$mdPanel', function() {
...
+
+          expect($mdPanel._groups['test']).toExist();
+
+          var config = {
+            groupName: 'test'
+          };
+
+          var panel1 = $mdPanel.create(config);
+
+          var panelFromGroup = $mdPanel._groups['test'].panels[0];
```
ErinCoughlan: I think there needs to be a util for this. It's complicated and confusing.

Perhaps, `getGroup(groupName)`?

Then you should use `toContain`:
`expect(getGroup('test')).toContain(panel1);`

bradrich:
@ErinCoughlan @crisbeto Thank you for the comments. I will get to them ASAP and re-push.

src/components/panel/panel.js
bradrich: @crisbeto Do you think that we should add this to `$mdUtil`? Or, do you think that it should be confined to this API?

crisbeto: I think it should be confined to the panel until we find a case where it'd be useful in another component.

src/components/panel/panel.js
bradrich: @crisbeto Not necessarily. Imagine this scenario:

 - The toolbar group has a `maxOpen` of 3 panels.
 - The user opens `panel1`, then opens `panel2`, and when they open `panel3`, `panel1` closes. The new order inside of the group went from `panel1`, `panel2`, `panel3` to `panel2`, `panel3`, `panel1`.
 - If no ID was provided for any of the panels, when the user clicks again on the icon that opened `panel1`, it is `panel4` that gets opened, not `panel1`. Then, `panel2` is closed. The new order is `panel3`, `panel1`, `panel4`, `panel2`.
 - If the user clicks on the icon that opened `panel2` earlier, it will open `panel5` instead. Then, `panel3` is closed. The new order is `panel1`, `panel4`, `panel2`, `panel5`, `panel3`.
 - If the user then clicks to open another panel, `panel6` is created and added to the group, and the first opened is not `panel1`, which is the first panel in the group, it is `panel4`, the second panel in the group. 

That is why I had to have a special iterator. The whole problem came from having to support a group of panels that have an ID, meaning that it is the same panel opening again and again, and a group of panels that do not have an ID and each click is creating a new panel.

Please let me know if that doesn't make sense.

src/components/panel/panel.spec.js
```diff
@@ -1053,6 +1053,125 @@ describe('$mdPanel', function() {
...
+          expect(panel1).toEqual(panelFromGroup);
+        });
+
+    it('should set the maximum number of panels allowed open within a group ' +
+        'using the newPanelGroup option', function() {
+          $mdPanel.newPanelGroup('test', {
+            maxOpen: 1
+          });
+
+          expect($mdPanel._groups['test']).toExist();
```
ErinCoughlan: unnecessary expect

src/components/panel/panel.spec.js
```diff
@@ -1053,6 +1053,125 @@ describe('$mdPanel', function() {
...
+          expect($mdPanel._groups['test'].maxOpen).toEqual(1);
+        });
+
+    it('should set the maximum number of panels allowed open within a group ' +
+        'using the setGroupMaxOpen method', function() {
+          $mdPanel.newPanelGroup('test');
+
+          expect($mdPanel._groups['test']).toExist();
```
ErinCoughlan: unnecessary expect

src/components/panel/panel.spec.js
```diff
@@ -1053,6 +1053,125 @@ describe('$mdPanel', function() {
...
+          expect(panel2.isOpened).toEqual(true);
+          expect(panel3.isOpened).toEqual(true);
+
+          panel1.open();
+          flushPanel();
+          expect(panel1.isOpened).toEqual(true);
+          expect(panel2.isOpened).toEqual(false);
+          expect(panel3.isOpened).toEqual(true);
+
+          panel1.close();
```
ErinCoughlan: Should be part of test cleanup in the afterEach

src/components/panel/demoGroups/script.js
bradrich: @ErinCoughlan I do not think so. Destroying a panel makes it unable to be re-opened, and in this demo, I am applying an ID to each opened panel, which means it can be closed and re-opened from the click event.

src/components/panel/panel.js
crisbeto: Looking at it again, it does. I had missed the part that the increment of `openCount` is inside the if statement.

src/components/panel/demoGroups/style.global.css
bradrich: @ErinCoughlan When specifying `propagateContainerEvents` in the panel config, the panel CSS applies `pointer-events: none` to the entire panel DOM element. This was added to the `menu-panel-container` class and then the class was added with the `panelClass` property of the panel config, in order to allow for the user to interact with the menu.

crisbeto: @bradrich You should keep in mind that `pointer-events: all` seems [to be limited to SVG](https://developer.mozilla.org/en-US/docs/Web/CSS/pointer-events). You should use `auto` if you want to reset it.

ErinCoughlan: @bradrich This is also a strange edge case that I don't think is intuitive to people using `propagateContainerEvents`. They'd have to look at the source code to figure it out. The user shouldn't have to care about how we implemented the propagation, it should just work.

src/components/panel/demoGroups/index.html
bradrich: @ErinCoughlan This was originally going to be a larger demo where another group was going to be in the content area below the toolbar. The `showMenu` function was opened up in order to allow for multiple groups to be added and their panels opened. I do not understand your second paragraph.

src/components/panel/panel.js
bradrich: @ErinCoughlan I also changed this on line 224 as well.

src/components/panel/demoGroups/index.html
ErinCoughlan: Demos should be something we expect users to _want_ to do - the common case, not the everything case. All I'm saying in the second paragraph is that the common case is to declare the group in the open method, not pass it in. Most people will have exactly one open method per panel, not one method that's used to open multiple panels.

src/components/panel/panel.js
bradrich: @crisbeto This is to allow for negative numbers to be passed in. See:  http://stackoverflow.com/questions/5306680/move-an-array-element-from-one-array-position-to-another

src/components/panel/demoGroups/style.global.css
bradrich: @ErinCoughlan In order to have pointer events reach anything beyond the panel, isn't it necessary to include `propagateContainerEvents: true`? And, in doing so, you allow `pointer-events` to reach through the panel as well. This will result in a menu, such as in the example, that is unable to be interacted with. To catch `pointer-events` only on the panel itself, I added the `pointer-events` CSS class.

src/components/panel/panel.js
crisbeto: I don't see a case where we'd use it though. The only place where we use `move` is [here ](https://github.com/angular/material/pull/9538/files#diff-b71bb3e10759daf665e48e9bc558dc99R912) and I don't

src/components/panel/panel.js
bradrich: @ErinCoughlan I thought that the response was necessary in order to continue the promise stream?

src/components/panel/panel.js
bradrich: @ErinCoughlan There are some merits. I will add it.

src/components/panel/demoGroups/style.global.css
ErinCoughlan: @bradrich My point here is that we should be doing this automatically, not making the user set it. No one want a panel that they can't interact with. That defeats the purpose of having a panel in most cases and to me, means that `propagateContainerEvents` was implemented incorrectly. It should only propagate the container, not through panel. Does that make sense? If I had realized this was a side effect earlier, I would have forced a better implementation.

src/components/panel/panel.js
ErinCoughlan: @bradrich Promises continue if they resolve, which it does because this is synchronous and wrapped in a then. You never call reject here.

src/components/panel/panel.js
bradrich: @ErinCoughlan I didn't understand this comment. Can you elaborate?

bradrich:
@ErinCoughlan @crisbeto I have updated everything within the PR. Please review.

src/components/panel/panel.js
```diff
@@ -1663,6 +1809,44 @@ MdPanelRef.prototype._done = function(callback, self) {
...
+  if (!this._$mdPanel._groups[groupName]) {
+    throw new Error('The group does not exist.');
+  }
+
+  var group = this._$mdPanel._groups[groupName];
+  var index = group.panels.indexOf(this);
+
+  if (index > -1) {
+    group.panels.splice(index, 1);
+    console.debug(group.panels);
```
DevVersion: Why is there a debug logging?

bradrich: @DevVersion Oops.

src/components/panel/panel.js
```diff
@@ -806,6 +860,79 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+ * @param {string} groupName The name of the group to check.
+ * @private
+ */
+MdPanelService.prototype._checkGroupMaxOpen = function(groupName) {
+  if (groupName &&
+      this._groups[groupName] &&
+      this._groups[groupName].maxOpen > 0) {
+        var openCount = 0,
+            firstOpenedPanel = null,
+            firstOpenedPanelIndex = null;
```
DevVersion: When I joined the team, I was explicitly forced to use separate variable statements (more clear and AFAIK follows the code conventions)

```js
var openCount = 0;
var firstOpendedPanel = null;
```

> See https://github.com/angular/material/pull/5589#discussion_r44260777 and https://github.com/angular/material/pull/5589#discussion_r44260777

src/components/panel/panel.js
```diff
@@ -806,6 +860,79 @@ MdPanelService.prototype.newPanelAnimation = function() {
 
 
 /**
+ * Creates a panel group and adds it to a tracked list of panel groups.
+ * @param groupName {string} Name of the group to create.
+ * @param config (!Object=) Specific configuration object that may contain the
```
DevVersion: Invalid jsdoc, should be swift braces instead

```
* @param {Object}
```

src/components/panel/panel.js
```diff
@@ -200,6 +201,34 @@ angular
...
  */
 
+/**
+ * @ngdoc method
+ * @name $mdPanel#newPanelGroup
+ * @description
+ * Creates a panel group and adds it to a tracked list of panel groups.
+ *
+ * @param groupName {string} Name of the group to create.
+ * @param config (!Object=) Specific configuration object that may contain the
```
DevVersion: Invalid jsdoc, should be curly bracket instead

```
* @param {Object}
```

bradrich:
@DevVersion Updated. Thank you for catching those.

src/components/panel/demoGroups/style.global.css
```diff
@@ -0,0 +1,80 @@
...
+  pointer-events: auto;
+}
+
+.menu-panel {
+  width: 256px;
+  background-color: #ffffff;
+  border-radius: 4px;
+  box-shadow: 0px 1px 5px 0px rgba(0, 0, 0, 0.2),
+      0px 2px 2px 0px rgba(0, 0, 0, 0.14),
+      0px 3px 1px -2px rgba(0, 0, 0, 0.12);
```
DevVersion: This seems to be a sample shadow just for the demo, why not using `md-whiteframe` instead to show best practices?

bradrich: @DevVersion This is CSS. I am unsure how to use Sass variables without it being Sass.

src/components/panel/panel.js
bradrich: @ErinCoughlan This is fixed. Comment just didn't disappear because this particular line wasn't altered.

src/components/panel/demoGroups/style.global.css
DevVersion: I meant just applying to the template

```html
<div class="menu-panel" md-whiteframe="4">
```

bradrich: @DevVersion Ah, I didn't realize that was a possibility.

src/components/panel/panel.js
ErinCoughlan: @bradrich It seemed like the ordering was wrong, but I guess since you just close the original panel, it's not as important.

src/components/panel/panel.js
```diff
@@ -1663,6 +1809,42 @@ MdPanelRef.prototype._done = function(callback, self) {
...
+    throw new Error('The group does not exist.');
+  }
+
+  var group = this._$mdPanel._groups[groupName];
+  var index = group.panels.indexOf(this);
+
+  if (index > -1) {
+    group.panels.splice(index, 1);
+  } else {
+    throw new Error('The panel does not exist within the group.');
```
ErinCoughlan: I'd argue that if the panel wasn't in the group, then the remove action was successful so we shouldn't throw an error here.

src/components/panel/panel.js
```diff
@@ -806,6 +860,79 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+  if (groupName &&
+      this._groups[groupName] &&
+      this._groups[groupName].maxOpen > 0) {
+        var openCount = 0;
+        var firstOpenedPanel = null;
+        var firstOpenedPanelIndex = null;
+
+        angular.forEach(this._groups[groupName].panels, function(panel, index) {
+          if (panel.isOpened) {
+            openCount++;
```
ErinCoughlan: Same comment as before, why are we iterating here and not keeping track of open panels as part of the group config?

src/components/panel/panel.js
```diff
@@ -806,6 +860,79 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+};
+
+
+/**
+ * Determines if opening a panel within a group will exceed the limit of allowed
+ * open panels. If so, the first panel opened within the group will be closed.
+ * @param {string} groupName The name of the group to check.
+ * @private
+ */
+MdPanelService.prototype._checkGroupMaxOpen = function(groupName) {
```
ErinCoughlan: I think this should be two methods... I don't expect a "check" method to have side effects such as closing panels.

I would say one for `checkGroupMaxOpen`, then other for `closeOldestPanel`.

src/components/panel/panel.js
ErinCoughlan: @bradrich You're ordering the panels whenever you exceed `maxOpen` so the last `currentNumOpen` in the list are open.

src/components/panel/panel.js
bradrich: @ErinCoughlan That is a very good point.

src/components/panel/panel.js
bradrich: @ErinCoughlan It has to do less with the number of open panels and the determining of which panel in the group is the `firstOpenedPanel`. I could keep the opened panel count somewhere else, but I would still need to iterate in order to find the first opened panel. Is there a better way to get this setting?

src/components/panel/panel.js
bradrich: @ErinCoughlan Could it be that the method is just misnamed? It is checking to see if the opening of a panel exceeds the `maxOpen` limit, but the method is really for the reordering of the panels within the group and the closing of the first opened panel. Do you have a naming suggestion?

bradrich:
@ErinCoughlan @crisbeto @DevVersion Working on a different way to handle the tracking of the open panels within a group in order to eliminate the iteration and reordering. I will re-push asap.

bradrich:
@ErinCoughlan @crisbeto @DevVersion New method of tracking open panels and closing the appropriate panel after maxOpen is reached has been completed. Please review when you have a chance.

src/components/panel/panel.js
```diff
@@ -200,6 +201,34 @@ angular
  * @returns {!MdPanelAnimation} panelAnimation
  */
 
+/**
+ * @ngdoc method
+ * @name $mdPanel#newPanelGroup
+ * @description
+ * Creates a panel group and adds it to a tracked list of panel groups.
+ *
+ * @param groupName {string} Name of the group to create.
```
crisbeto: The pattern that we use when declaring `@param` in all the other strings is `@param <type> <name>`, a few of the docstrings here use `@param <name> <type>`.

src/components/panel/panel.js
```diff
@@ -200,6 +201,34 @@ angular
...
+ * @param config {!Object=} Specific configuration object that may contain the
+ *     following properties:
+ *
+ *   - `maxOpen` - `{number=}`: The maximum number of panels that are allowed
```
crisbeto: "...that are allowed open" -> "...that are allowed to be open"

src/components/panel/panel.js
```diff
@@ -721,6 +768,9 @@ function MdPanelService($rootElement, $rootScope, $injector, $window) {
   /** @private {!Object<string, !MdPanelRef>} */
   this._trackedPanels = {};
 
+  /** @private {!Object<string, {panels: !Array<!MdPanelRef>, maxOpen: number}>} */
+  this._groups = {};
```
crisbeto: I know that I mentioned yesterday that we should use `hasOwnProperty` when checking whether a group exists, but I think I have a better solution. If we switch this line to `this._groups = Object.create(null)`, it would create an object without a prototype, which means that we can safely use the bracket notation and can revert the `hasOwnProperty` change from yesterday.

src/components/panel/panel.js
```diff
@@ -806,6 +860,62 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+ * time.
+ * @param {string} groupName The name of the group to configure.
+ * @param {number|boolean} maxOpen The maximum number of panels that can be
+ *     opened. If false is passed in, the maximum limit will be removed.
+ */
+MdPanelService.prototype.setGroupMaxOpen = function(groupName, maxOpen) {
+  if (this._groups[groupName]) {
+    this._groups[groupName].maxOpen = maxOpen;
+  } else {
+    throw new Error('Group does not exist yet. Call newPanelGroup().');
```
crisbeto: This is more of a question: Perhaps we could prefix the error messages with `mdPanel:` so it's a little more obvious where errors are coming from?

src/components/panel/panel.js
```diff
@@ -806,6 +860,62 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+MdPanelService.prototype.newPanelGroup = function(groupName, config) {
+  if (!this._groups.hasOwnProperty(groupName)) {
+    config = config || {};
+    var group = {
+      panels: [],
+      openPanels: [],
+      maxOpen: config.maxOpen > 0 ? config.maxOpen : false
+    };
+    this._groups[groupName] = group;
+    return this._groups[groupName];
```
crisbeto: Perhaps move the `return this._groups[groupName]` outside the `if` statement? This would return the group even if exists already.

src/components/panel/panel.js
```diff
@@ -806,6 +860,62 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+ *   - `maxOpen` - `{number=}`: The maximum number of panels that are allowed
+ *     open within a defined panel group.
+ */
+MdPanelService.prototype.newPanelGroup = function(groupName, config) {
+  if (!this._groups.hasOwnProperty(groupName)) {
+    config = config || {};
+    var group = {
+      panels: [],
+      openPanels: [],
+      maxOpen: config.maxOpen > 0 ? config.maxOpen : false
```
crisbeto: Another question: What do you think about having the default value for `maxOpen` be `Infinity`. This could spare us some issues when comparing numbers to `false`.

src/components/panel/panel.js
```diff
@@ -1663,6 +1795,40 @@ MdPanelRef.prototype._done = function(callback, self) {
 };
 
 
+/**
+ * Adds a panel to a group if the panel does not already exist within the group.
```
crisbeto: The doc string mention "if the panel does not already exist within the group", but I don't see anything that checks that the panel isn't part of the group.

src/components/panel/panel.js
```diff
@@ -2550,3 +2716,22 @@ function getElement(el) {
       document.querySelector(el) : el;
   return angular.element(queryResult);
 }
+
+
+/**
+ * Moves an array element from one array position to another position.
```
crisbeto: Might want to remove this if it isn't being used anymore.

src/components/panel/panel.js
```diff
@@ -757,8 +807,8 @@ MdPanelService.prototype.create = function(config) {
     return this._trackedPanels[config.id];
   }
 
-  // If no ID is set within the passed-in config, then create an arbitrary ID.
   this._config = {
+    // If no ID is set within the passed-in config, then create an arbitrary ID.
     id: config.id || 'panel_' + this._$injector.get('$mdUtil').nextUid(),
```
crisbeto: This looks like it would call `$injector.get('$mdUtil')` every time somebody doesn't pass in an id. It might be more appropriate to cache it once on init.

src/components/panel/panel.js
ErinCoughlan: Agreed. Keeping the types the same is going to help a lot in the future, also helps with comparing undefined vs false and so forth.

src/components/panel/panel.js
ErinCoughlan: I think `$mdUtil` should be injected directly like `$window` and `$rootScope` in the constructor. `$injector` was only meant to be used when creating a new panelPosition or panelAnimation, which also require injected variables.

src/components/panel/demoGroups/index.html
```diff
@@ -0,0 +1,46 @@
+<div ng-controller="PanelGroupsCtrl as ctrl" ng-cloak>
+
+  <md-toolbar class="md-accent">
+    <div class="md-toolbar-tools">
+      <md-button
+        class="md-icon-button"
+        aria-label="Settings"
+        ng-click="ctrl.showMenu($event, ctrl.settings, 'toolbar')">
```
ErinCoughlan: 'toolbar' is still here. It should be written in the `showMenu` method, not passed in.

src/components/panel/demoGroups/index.html
```diff
@@ -0,0 +1,46 @@
...
+        <md-icon md-svg-icon="img/icons/menu.svg"></md-icon>
+      </md-button>
+      <h2>Toolbar with grouped panels (Maximum open: 2)</h2>
+      <span flex></span>
+      <md-button
+        class="md-icon-button"
+        aria-label="Favorite"
+        ng-click="ctrl.showMenu($event, ctrl.favorite, 'toolbar')">
```
ErinCoughlan: `ctrl.favorite.name` - Ditto for settings and more.

src/components/panel/demoGroups/script.js
```diff
@@ -0,0 +1,88 @@
...
+    this.showMenu = function($event, menu, group) {
+      var template = '' +
+          '<div class="menu-panel" md-whiteframe="4">' +
+          '  <div class="menu-content">' +
+          '    <div class="menu-item" ng-repeat="item in ctrl.items">' +
+          '      <button class="md-button">' +
+          '        <span>{{item}}</span>' +
+          '      </button>' +
+          '    </div>' +
+          '    <div class="menu-divider"></div>' +
```
ErinCoughlan: md-divider?

src/components/panel/demoGroups/style.global.css
```diff
@@ -0,0 +1,77 @@
+.menu-panel-container {
+  pointer-events: auto;
+}
+
+.menu-panel {
+  width: 256px;
+  background-color: #ffffff;
```
ErinCoughlan: nit: Just `#fff`.

src/components/panel/panel.js
```diff
@@ -200,6 +201,34 @@ angular
...
+ * Creates a panel group and adds it to a tracked list of panel groups.
+ *
+ * @param groupName {string} Name of the group to create.
+ * @param config {!Object=} Specific configuration object that may contain the
+ *     following properties:
+ *
+ *   - `maxOpen` - `{number=}`: The maximum number of panels that are allowed
+ *     open within a defined panel group.
+ *
+ * @returns {!Object} group
```
ErinCoughlan: I think you need a type for `Object` here. I have no idea what "group" is going to contain or how to use it when I get it back. Do I have access to my config? Are there special functions I can call?

src/components/panel/panel.js
```diff
@@ -721,6 +768,9 @@ function MdPanelService($rootElement, $rootScope, $injector, $window) {
   /** @private {!Object<string, !MdPanelRef>} */
   this._trackedPanels = {};
 
+  /** @private {!Object<string, {panels: !Array<!MdPanelRef>, maxOpen: number}>} */
```
ErinCoughlan: I think this type is wrong now since there are more properties. (Missing `openPanels`.)

src/components/panel/panel.js
```diff
@@ -806,6 +860,62 @@ MdPanelService.prototype.newPanelAnimation = function() {
 
 
 /**
+ * Creates a panel group and adds it to a tracked list of panel groups.
+ * @param groupName {string} Name of the group to create.
+ * @param config {!Object=} Specific configuration object that may contain the
+ *     following properties:
+ *
+ *   - `maxOpen` - `{number=}`: The maximum number of panels that are allowed
+ *     open within a defined panel group.
```
ErinCoughlan: Missing `@returns`

src/components/panel/panel.js
ErinCoughlan: @bradrich I'm fine with a rename, although I think the clarity is going to be a lot better with two methods. I can't think of a good name for a combined method, at least nothing that coveys the purpose of the method.

```
if (this.countExceedsMaxOpen(group)) { this.closeOldestPanel(group) };
```

src/components/panel/panel.js
```diff
@@ -806,6 +860,62 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+
+
+/**
+ * Determines if opening a panel within a group will exceed the limit of allowed
+ * open panels. If so, the first panel opened within the group will be closed.
+ * @param {string} groupName The name of the group to check.
+ * @private
+ */
+MdPanelService.prototype._checkGroupMaxOpen = function(groupName) {
+  if (groupName && this._groups[groupName]) {
```
ErinCoughlan: groupName must be defined based on the types.

src/components/panel/panel.js
```diff
@@ -355,6 +384,24 @@ angular
  * @param {!MdPanelPosition} position
  */
 
+/**
+ * @ngdoc method
+ * @name MdPanelRef#addToGroup
+ * @description
+ * Adds a panel to a group. The group must be created ahead of time.
```
ErinCoughlan: You probably should mention that a panel can only be part of one group, and this should be enforced when adding a panel to a group. (Although I'm not a fan of this decision, it's fine for an initial version.)

All of your logic is based around config.groupName, which is a string, not an array of strings.

src/components/panel/panel.spec.js
```diff
@@ -1053,6 +1053,114 @@ describe('$mdPanel', function() {
...
+          expect(getGroupPanels('test')).not.toContain(panel);
+        });
+
+    it('should set the maximum number of panels allowed open within a group ' +
+        'using the newPanelGroup option', function() {
+          $mdPanel.newPanelGroup('test', {
+            maxOpen: 1
+          });
+
+          expect($mdPanel._groups['test'].maxOpen).toEqual(1);
```
ErinCoughlan: This as a util too.

src/components/panel/panel.spec.js
```diff
@@ -1053,6 +1053,114 @@ describe('$mdPanel', function() {
...
+
+          var config = {
+            groupName: 'test'
+          };
+
+          var panel = $mdPanel.create(config);
+          expect(getGroupPanels('test')).toContain(panel);
+        });
+
+    it('should remove a panel from a group using the removeFromGroup method',
```
ErinCoughlan: Missing tests:

```
it('should remove a panel from a group on panel destroy', ...
it('should update open panels when a panel is closed', ... // Similar to the closing the first panel when you hit maxOpen,
// but you manually close one and make sure the final state is still correct (that another one isn't also closed).
it('should throw if trying to set maxOpen on a group that doesn/'t exist', ...
it('should only create a group once', ... // Create a group using config, then create a group with newPanelGroup.

// For a future PR because I don't think you have this functionality yet...
it('should reopen a forced close panel when another is closed', ...
```

src/components/panel/panel.spec.js
```diff
@@ -1053,6 +1053,114 @@ describe('$mdPanel', function() {
...
+            groupName: 'test'
+          };
+
+          var panel1 = $mdPanel.create(config);
+          var panel2 = $mdPanel.create(config);
+          var panel3 = $mdPanel.create(config);
+
+          panel1.open();
+          flushPanel();
+          expect(getGroupOpenPanels('test')).toContain(panel1);
```
ErinCoughlan: You should not be checking the internal state of `$mdPanel` to test this. Open the panels, make sure they are in the DOM and visible. Open too many panels, and make sure the the first one is no longer visible.

src/components/panel/panel.spec.js
```diff
@@ -1053,6 +1053,114 @@ describe('$mdPanel', function() {
...
   });
 
+  describe('grouping logic:', function() {
+    it('should create a group using the newPanelGroup method', function() {
+      $mdPanel.newPanelGroup('test');
+
+      expect($mdPanel._groups['test']).toExist();
+    });
+
+    it('should create a group using the config option groupName when the ' +
```
ErinCoughlan: Another missing test:

```it('should not create a group using the config option when the group is already defined', ...```

File: src/components/panel/panel.js (hunk 40 above)
Comment (bradrich) [inline src/components/panel/panel.js]: @ErinCoughlan How do you feel about this approach?

File: src/components/panel/demoGroups/index.html (hunk 48 above)
Comment (bradrich) [inline src/components/panel/demoGroups/index.html]: @ErinCoughlan I can't do that as I am accessing `menu.items` from the passed-in variable on line 70.

File: src/components/panel/panel.js (hunk 51 above)
Comment (bradrich) [inline src/components/panel/panel.js]: @ErinCoughlan Would I leave the definition of `MdPanelGroup` here? `MdPanelRef`, etc. are defined by their public variables. This is the only place where the `MdPanelGroup` will be returned.

Comment (ErinCoughlan) [inline src/components/panel/panel.js]: @bradrich You just want a more specific type than Object. Objects can have "puppies", but I'm pretty sure if I write "$mdPanel.newPanelGroup('my-group').puppies" it will break.

File: src/components/panel/panel.js (hunk 54 above)
Comment (bradrich) [inline src/components/panel/panel.js]: @ErinCoughlan Can you explain this comment?

File: src/components/panel/panel.js (hunk 55 above)
Comment (bradrich) [inline src/components/panel/panel.js]: @ErinCoughlan It is not a problem moving to supporting a panel being in multiple groups. Should I build this in?

Comment (ErinCoughlan) [inline src/components/panel/panel.js]: Not in this PR.

File: src/components/panel/panel.spec.js (hunk 58 above)
Comment (bradrich) [inline src/components/panel/panel.spec.js]: @ErinCoughlan Can you explain this comment?

**Comment by bradrich** [comment]
@ErinCoughlan @crisbeto I have updated everything that we reviewed and it is ready to go. If everything looks good, please respond with an LGTM.

**Comment by crisbeto** [comment]
Sounds good, I'll do a final review a little later.

**Comment by crisbeto** [comment]
Did a final look and LGTM. I think you only missed the one comment about a typo in the docs.

File: src/components/panel/panel.js [hunk 60]
```diff
@@ -806,6 +873,78 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+ * @param config {!Object=} Specific configuration object that may contain the
+ *     following properties:
+ *
+ *   - `maxOpen` - `{number=}`: The maximum number of panels that are allowed
+ *     open within a defined panel group.
+ *
+ * @returns {!Object<string,
+ *     {panels: !Array<!MdPanelRef>,
+ *     openPanels: !Array<!MdPanelRef>,
+ *     maxOpen: number}>}
```
ErinCoughlan: Probably mention that this is returning the created panel group.

src/components/panel/panel.js
```diff
@@ -806,6 +873,78 @@ MdPanelService.prototype.newPanelAnimation = function() {
...
+
+
+/**
+ * Determines if the current number of open panels within a group exceeds the
+ * limit of allowed open panels.
+ * @param {string} groupName The name of the group to check.
+ * @returns {boolean} true if open count does exceed maxOpen and false if not.
+ * @private
+ */
+MdPanelService.prototype._openCountExceedsGroupMaxOpen = function(groupName) {
```
ErinCoughlan: How about just: `openCountExceedsMaxOpen` because of the param groupName, we know it's for groups.

Also, `closeFirstOpenedPanel` is shorter as well, but still has the same point.

src/components/panel/panel.js
```diff
@@ -1663,6 +1832,44 @@ MdPanelRef.prototype._done = function(callback, self) {
...
+
+
+/**
+ * Removes a panel from a group if the panel exists within that group. The group
+ * must be created ahead of time.
+ * @param {string} groupName The name of the group.
+ */
+MdPanelRef.prototype.removeFromGroup = function(groupName) {
+  if (!this._$mdPanel._groups[groupName]) {
+    throw new Error('mdPanel: The group does not exist.');
```
ErinCoughlan: include the groupName in the error message to make it more useful?

src/components/panel/panel.spec.js
```diff
@@ -1049,6 +1049,169 @@ describe('$mdPanel', function() {
...
+    it('should update open panels when a panel is closed', function() {
+      $mdPanel.newPanelGroup('test');
+
+      var config = {
+        groupName: 'test'
+      };
+
+      var panel = $mdPanel.create(config);
+
+      panel.open();
```
ErinCoughlan: Please use the test helper methods for these:

`openPanel(config);` and `closePanel()`.

src/components/panel/panel.spec.js
```diff
@@ -1049,6 +1049,169 @@ describe('$mdPanel', function() {
...
+          var panel3 = $mdPanel.create(config);
+
+          panel1.open();
+          flushPanel();
+          expect(panel1.isAttached).toEqual(true);
+
+          panel2.open();
+          panel3.open();
+          flushPanel();
+          expect(panel1.isAttached).toEqual(false);
```
ErinCoughlan: Also expect that the other two are attached.

ErinCoughlan:
A few minor comments left. Once resolved, LGTM.

src/components/panel/panel.spec.js
bradrich: @ErinCoughlan Unless I am mistaken, this change will disrupt the expect tests that I am running in this test. I used this method because I needed to create a panel variable to then check the open panels array for afterward. Please let me know if I didn't write this test correctly.

bradrich:
@ErinCoughlan @crisbeto All changes and nits have been accounted for. Thanks for the solid review.

src/components/panel/panel.spec.js
ErinCoughlan: @bradrich Unless I'm really confused about something in that util, you get the panelRef back, so your tests still work. Well, you have to rename panel to panelRef...

```
openPanel(config);
expect(getPanels).toContain(panelRef);
```

jelbourn:
Passed presubmit, but needs rebase.

bradrich:
@jelbourn Rebase complete, please continue.

bradrich:
@DevVersion Are the TravisCI checks failing due to something that I can fix? I have rebased and there is still a failure...

DevVersion:
@bradrich Yeah that's not a CI issue. You have a syntax error in Line 2694

```
Uncaught SyntaxError: Unexpected token )
at src/components/panel/panel.spec.js:2694
```

bradrich:
@DevVersion Ah, found it. It was the callInterceptors function that didn't have its final `}` bracket. Not sure who wrote that function.

bradrich:
@jelbourn There was a missing bracket that I discovered in the test files and removed. Please continue.

DevVersion:
@bradrich Glad you fixed it. The tests in master pass, so it probably was a rebase mistake.

bradrich:
@DevVersion it was definitely on me, there were several merge conflicts and I missed an error.

PR #9547, opened 2016-09-07
fix(util): disableScrollAround should not remove disabled scroll mask

DevVersion:
* When the scroll mask is disabled per the new options which were recently added, the specified element will be used as scroll mask (under the hood)
  If restoring scroll, the scroll mask will be removed from the DOM (even the disabled one) - This is invalid.

This causes an issue when restoring the scroll, while the element is still showing up (e.g dialog with z-index)

@ThomasBurleson Also I need to correct my self, the issue within the `$mdPanel` is not reproducable (**but present**), because the element will be anyways detached by the panel.

Part of #9256

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -198,44 +198,56 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
      *     use the passed parent element.
      */
     disableScrollAround: function(element, parent, options) {
-      $mdUtil.disableScrollAround._count = $mdUtil.disableScrollAround._count || 0;
-      ++$mdUtil.disableScrollAround._count;
-      if ($mdUtil.disableScrollAround._enableScrolling) return $mdUtil.disableScrollAround._enableScrolling;
-      var body = $document[0].body,
-        restoreBody = disableBodyScroll(),
-        restoreElement = disableElementScroll(parent);
+      options = options || {};
 
-      return $mdUtil.disableScrollAround._enableScrolling = function() {
-        if (!--$mdUtil.disableScrollAround._count) {
+      $mdUtil.disableScrollAround._count = Math.max(0, $mdUtil.disableScrollAround._count || 0);
+      $mdUtil.disableScrollAround._count++;
+
+      if ($mdUtil.disableScrollAround._restoreScroll) {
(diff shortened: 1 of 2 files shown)


src/core/util/util.js
```diff
@@ -204,44 +204,56 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
+        return $mdUtil.disableScrollAround._restoreScroll;
+      }
+
+      var body = $document[0].body;
+      var restoreBody = disableBodyScroll();
+      var restoreElement = disableElementScroll(parent);
 
-      return $mdUtil.disableScrollAround._enableScrolling = function() {
+      return $mdUtil.disableScrollAround._restoreScroll = function() {
         if (!--$mdUtil.disableScrollAround._count) {
```
ThomasBurleson: This seems dangerous. How about 

```js
if ( --$mdUtil.disableScrollAround._count < 1 ) {
....
}
```

src/core/util/util.js
```diff
@@ -204,44 +204,56 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
      *     use the passed parent element.
      */
     disableScrollAround: function(element, parent, options) {
+      options = options || {};
+
       $mdUtil.disableScrollAround._count = $mdUtil.disableScrollAround._count || 0;
```
ThomasBurleson: Please fix:
```js
$mdUtil.disableScrollAround._count = $mdUtil.disableScrollAround._count > 0 ? $mdUtil.disableScrollAround._count : 0;

Comment (DevVersion) [inline src/core/util/util.js]: Yeah `Math.max` is better than a ternary or the OR operator.

File: src/core/util/util.js (hunk 1 above)
Comment (DevVersion) [inline src/core/util/util.js]: Yeah will change while I'm at it.

File: src/core/util/util.js (hunk 2 above)
Comment (DevVersion) [inline src/core/util/util.js]: So after checking, this does not work, because `count` is sometimes undefined. 

```js
Math.max(0, $mdUtil.disableScrollAround._count || 0);
```

Doing this ensures, that the count is not smaller than zero.

> Ternary is unnecessary here, duplicating the count three times is too much.

DevVersion:
@ThomasBurleson Addressed the comments you added.

Issue #9366, opened 2016-08-21
mdPanel - allow to specify a content element.


ErinCoughlan:
**Actual Behavior**:
 -  `What is the issue? *`
    In mdPanel, you can't specify an content element. You can in dialog, so we should find a way to make that code work for both components, probably via the mdCompiler.

 -  `What is the expected behavior?`
This allows developers, to specify a pre-rendered/compiled element, which will be not compiled each time a dialog opens.

It is possible to specify:

- A DOM Element
- A String as CSS Selector

When using an element, which is already present in the DOM, it will be anchored to the dialog, which means, that we temporary fetch it from the DOM into the dialog and move it back to its old DOM position upon close.

**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:`
 -  `Details:`

**Angular Versions**: *
 -  `Angular Version:`
 -  `Angular Material Version:`

**Additional Information**:
 -  `Browser Type: *`
 -  `Browser Version: *`
 -  `OS: *`
 -  `Stack Traces:`

----
Shortcut to create a [new CodePen Demo](http://codepen.io/team/AngularMaterial/pen/bEGJdd).
Note: `*` indicates required information. Without this information, your issue may be auto-closed.

> Do not modify the titles or questions. Simply add your responses to the ends of the questions.
  Add more lines if needed.

ThomasBurleson:
@DevVersion - Can you help Brad here. I think you did something similar for Dialog?

DevVersion:
@ThomasBurleson For sure - I actually planned to move that logic to the compiler anyways, because then we could also use that feature in the `bottom-sheet`, ` toast` and other interim elements.

As we discussed in the last Team Meeting, it would be better to start working on porting a simple component ('bottom-sheet`) to the panel, and concurrently work on the `contentElement` functionality for the panel (to be able to continue porting with the dialog)

bradrich:
@DevVersion I will ping you when I start on this. Thanks!

ThomasBurleson:
@DevVersion - sounds reasonable. I do not have clarity regarding the issues and impacts... but proceed if it makes sense and we can review the PR.

crisbeto:
This was resolved via #9551. There is still some work related to the content element, but that will be tracked in a separate issue.

PR #9551, opened 2016-09-08
feat(compiler): support for content elements

DevVersion:
- This moves the `contentElement` logic from the `$mdDialog` (as introduced in https://github.com/angular/material/commit/135cb3a290bee19a1395fc3a7440599c5ff5aef5) to the `$mdCompiler`.

  This allows the `$mdPanel` to take advantage of this functionality and does not block the port of the dialog to the `$mdPanel`

**FYI**: Don't be too nitpicky about the docs, those were mostly not written by me (rather fix those in a new PR)

Base for #9366

### DIFF
--- a/src/core/services/compiler/compiler.js
+++ b/src/core/services/compiler/compiler.js
@@ -1,139 +1,327 @@
+/**
+ * @ngdoc module
+ * @name material.core.compiler
+ * @description
+ * Angular Material template and element compiler.
+ */
 angular
   .module('material.core')
-  .service('$mdCompiler', mdCompilerService);
+  .service('$mdCompiler', MdCompilerService);
 
-function mdCompilerService($q, $templateRequest, $injector, $compile, $controller) {
-  /* jshint validthis: true */
+/**
+ * @ngdoc service
+ * @name $mdCompiler
+ * @module material.core.compiler
+ * @description
+ * The $mdCompiler service is an abstraction of Angular's compiler, that allows developers
+ * to easily compile an element with options like in a Directive Definition Object.
+ *
+ * > The compiler powers a lot of components inside of Angular Material.
+ * > Like the `$mdPanel` or `$mdDialog`.
+ *
+ * @usage
+ *
+ * Basic Usage with a template
+ *
+ * <hljs lang="js">
+ *   $mdCompiler.compile({
+ *     templateUrl: 'modal.html',
+ *     controller: 'ModalCtrl',
+ *     locals: {
+ *       modal: myModalInstance;
+ *     }
+ *   }).then(function (compileData) {
+ *     compileData.element; // Compiled DOM element
+ *     compileData.link(myScope); // Instantiate controller and link element to scope.
+ *   });
+ * </hljs>
+ *
+ * Example with a content element
+ *
+ * <hljs lang="js">
+ *
+ *   // Create a virtual element and link it manually.
+ *   // The compiler doesn't need to recompile the element each time.
+ *   var myElement = $compile('<span>Test</span>')(myScope);
+ *
+ *   $mdCompiler.compile({
+ *     contentElement: myElement
+ *   }).then(function (compileData) {
+ *     compileData.element // Content Element (same as above)
+ *     compileData.link // This does nothing when using a contentElement.
+ *   });
+ * </hljs>
+ *
+ * > Content Element is a significant performance improvement when the developer already knows that the
+ * > compiled element will be always the same and the scope will not change either.
+ *
+ * The `contentElement` option also supports DOM elements which will be temporary removed and restored
+ * at its old position.
+ *
+ * <hljs lang="js">
+ *   var domElement = document.querySelector('#myElement');
+ *
+ *   $mdCompiler.compile({
+ *     contentElement: myElement
+ *   }).then(function (compileData) {
+ *     compileData.element // Content Element (same as above)
+ *     compileData.link // This does nothing when using a contentElement.
+ *   });
+ * </hljs>
+ *
+ * The `$mdCompiler` can also query for the element in the DOM itself.
+ *
+ * <hljs lang="js">
+ *   $mdCompiler.compile({
+ *     contentElement: '#myElement'
+ *   }).then(function (compileData) {
+ *     compileData.element // Content Element (same as above)
+ *     compileData.link // This does nothing when using a contentElement.
+ *   });
+ * </hljs>
+ *
+ */
+function MdCompilerService($q, $templateRequest, $injector, $compile, $controller) {
+  /** @private @const {!angular.$q} */
+  this.$q = $q;
 
-  /*
-   * @ngdoc service
-   * @name $mdCompiler
-   * @module material.core
-   * @description
-   * The $mdCompiler service is an abstraction of angular's compiler, that allows the developer
-   * to easily compile an element with a templateUrl, controller, and locals.
-   *
-   * @usage
-   * <hljs lang="js">
-   * $mdCompiler.compile({
-   *   templateUrl: 'modal.html',
-   *   controller: 'ModalCtrl',
-   *   locals: {
-   *     modal: myModalInstance;
-   *   }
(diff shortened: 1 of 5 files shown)


ThomasBurleson:
@topherfangio, @ErinCoughlan - plz review.

topherfangio:
LGTM. Definitely agree that the docs need some cleanup, but otherwise, I think this is good.

src/core/services/compiler/compiler.js
```diff
@@ -1,139 +1,316 @@
...
+/**
+ * @ngdoc service
+ * @name $mdCompiler
+ * @module material.core.compiler
+ * @description
+ * The $mdCompiler service is an abstraction of Angular's compiler, that allows developers
+ * to easily compile an element with options like in a Directive Definition Object.
+ *
+ * > The compiler powers a lot of components inside of Angular Material.
+ * > Like the `$mdPanel` or `$mdDialog`.
```
ErinCoughlan: Hehe, this line is great. :dancer: 

I'm fine with fixing the docs later, but some love would be nice.

src/core/services/compiler/compiler.js
```diff
@@ -1,139 +1,316 @@
...
+ *     contentElement: '#myElement'
+ *   }).then(function (compileData) {
+ *     compileData.element // Content Element (same as above)
+ *     compileData.link // This does nothing when using a contentElement.
+ *   });
+ * </hljs>
+ *
+ */
+function MdCompilerService($q, $templateRequest, $injector, $compile, $controller) {
+  this.$q = $q;
```
ErinCoughlan: Mark these are private? Annotate types?

```@private @const {!angular.$injector}```

File: src/core/services/compiler/compiler.js [hunk 3]
```diff
@@ -1,139 +1,316 @@
...
+  if (options.contentElement) {
+    return this._prepareContentElement(options);
+  } else {
+    return this._compileTemplate(options);
+  }
+
+};
+
+/**
+ * @private
```
ErinCoughlan: optional nit: Put `@private` last. I tend to think the description and params are more important than the fact the method is private.

src/core/services/compiler/compiler.js
```diff
@@ -1,139 +1,316 @@
...
+/**
+ * @ngdoc method
+ * @name $mdCompiler#compile
+ * @description
+ *
+ * A method to compile a HTML template with the Angular compiler.
+ * The `$mdCompiler` is wrapper around the Angular compiler and provides extra functionality
+ * like controller instantiation or async resolves.
+ *
+ * @param {object} options An options object, with the following properties:
```
ErinCoughlan: You may want to add nullability to some of these based on the usage.

In this method, options has to be defined or it will throw an error "can't access contentElement of undefined". So this would be `@param {!Object}`

src/core/services/compiler/compiler.js
```diff
@@ -1,139 +1,316 @@
...
-              $element: element
-            });
+};
 
-            var invokeCtrl = $controller(controller, ctrlLocals, true, controllerAs);
+/**
+ * @private
+ * Method to compile an element with the given options.
+ * @param locals {Object} Locals to be injected to the controller if present
+ * @param element {JQLite} Element to be compiled and linked
```
ErinCoughlan: Usually we write these as `@param {type} variableName Description.`

src/core/services/compiler/compiler.js
```diff
@@ -1,139 +1,316 @@
...
+        parent.appendChild(element);
+      } else {
+        // When the element had a sibling, which marks the previous position of the element
+        // in the DOM, we insert it correctly before the sibling, to have the same index as
+        // before.
+        parent.insertBefore(element, nextSibling);
+      }
+    }
+  }
+};
```
ErinCoughlan: Missing newline at end of file?

src/core/services/compiler/compiler.js
```diff
@@ -1,139 +1,316 @@
...
+
+  return compileData;
+
+};
+
+/**
+ * Fetches an element removing it from the DOM and using it temporary for the compiler.
+ * Elements which were fetched will be restored after use.
+ * @param options {Object} Options to be used for the compilation.
+ * @returns {{element: JQLite, restore: Function}}
```
ErinCoughlan: nit: `!angular.JQLite`, and `!Function`.

src/core/services/compiler/compiler.js
DevVersion: Hehe, I only have love available for the `$$interimElement` stuff 😄 

> As planned, we are improving this in a follow-up PR

DevVersion:
@topherfangio Yeah, as said I did not write most of the docs, and I wanted to keep this PR as clean as possible, since changes to the compiler are very dangerous.

> There will be a follow-up PR for improving the documentation ;)

src/core/services/compiler/compiler.js
```diff
@@ -1,139 +1,326 @@
...
+ *   }).then(function (compileData) {
+ *     compileData.element // Content Element (same as above)
+ *     compileData.link // This does nothing when using a contentElement.
+ *   });
+ * </hljs>
+ *
+ */
+function MdCompilerService($q, $templateRequest, $injector, $compile, $controller) {
+  /** @private @const {!angular.$q} */
+  this.$q = $q;
```
crisbeto: Probably down to preference, but if these are private, shouldn't they be prefixed with an underscore?

src/core/services/compiler/compiler.js
```diff
@@ -1,139 +1,326 @@
...
+ * @param {!Object} options Compile options
+ * @returns {!Object} Compile data with link function.
+ * @private
+ */
+MdCompilerService.prototype._compileTemplate = function(options) {
+
+  var self = this;
+  var templateUrl = options.templateUrl;
+  var template = options.template || '';
+  var resolve = angular.extend({}, options.resolve);
```
crisbeto: What's the purpose of the `extend` with an empty object here? If it's for copying, there's `angular.copy`.

src/core/services/compiler/compiler.js
DevVersion: Hm, I do not like to have them in the Prototype, we also don't have them in the chips.

I personally like them, when working with TypeScript and Getters (as a accumulator)

src/core/services/compiler/compiler.js
DevVersion: I did not create this line.

From my understanding it's just to provide a default empty object and to ensure that it's always an object.

That won't work with `angular.copy` then, and we also don't want to have a deep copy for resolves.

src/core/services/compiler/compiler.js
DevVersion: Done ;)

src/core/services/compiler/compiler.js
crisbeto: In that case `options.resolve || {}` should work the same.

crisbeto:
A couple of very minor notes, but otherwise LGTM.

src/core/services/compiler/compiler.js
DevVersion: FYI: Had a quick chat with @crisbeto about it, and `extend` seems to be the right way, because we want to create a flat clone (not deep).

src/core/services/compiler/compiler.js
```diff
@@ -1,139 +1,327 @@
...
-              angular.extend(invokeCtrl.instance, locals);
-            }
+/**
+ * Fetches an element removing it from the DOM and using it temporary for the compiler.
+ * Elements which were fetched will be restored after use.
+ * @param {!Object} options Options to be used for the compilation.
+ * @returns {{element: !JQLite, restore: !Function}}
+ * @private
+ */
+MdCompilerService.prototype.fetchContentElement = function(options) {
```
ErinCoughlan: Private methods should have underscores in the name.

DevVersion: Yeah I missed that one. Solved now!

PR #9554, opened 2016-09-08
feat(datepicker): allow date strings as the source for ng-model

### LINKED ISSUE #6253: md-datepicker: parseDate before throwing error
The datepicker currently throws error if model given is not a date instance. 
With parseDate function available, it would be helpful check if valid dateString is given before throwing an error pull #5503.

Possible replacement of line 256-259 of src/components/datepicker/datePicker.js:
      
      if (value && !(value instanceof Date)) {
        var parseValue = self.dateLocale.parseDate(value);
        if (isNaN(parseValue)) {
            throw Error('The ng-model for md-datepicker must be a Date instance. ' +
            'Currently the model is a: ' + (typeof value));
        } else {
            value = parseValue;
        }
      }

Up for discussion

### LINKED ISSUE #9535: date picker: support string representation of date as a model
**Actual Behavior**:
-  `What is the issue? *` The date picker invalidates the valid ISO 8601 formatted string. It should allow to parseDate in mdLocaleProvider.
-  `What is the expected behavior?` The parseDate function should fire prior to date picker throwing the error for string type or inbuilt support for valid ISO 8601 string format.
  This has been already raised earlier: #6253

**Angular Versions**: *
-  `Angular Version:` 1.5.x

**Additional Information**:
-  `Browser Type: *` Chrome
-  `OS: *` MAC , El Captain
-  `Stack Traces:`

---

>   Add more lines if needed.

crisbeto:
This change adds the ability to have a date string (e.g. "2016-01-02") as the source for the datepicker. Until now only Date objects were allowed.

Fixes #6253.
Fixes #9535.

### DIFF
--- a/src/components/datepicker/js/datepickerDirective.js
+++ b/src/components/datepicker/js/datepickerDirective.js
@@ -18,7 +18,7 @@
    * @name mdDatepicker
    * @module material.components.datepicker
    *
-   * @param {Date} ng-model The component's model. Expects a JavaScript Date object.
+   * @param {Date} ng-model The component's model. Expects either a JavaScript Date object or a value that can be parsed into one (e.g. a ISO 8601 string).
    * @param {Object=} ng-model-options Allows tuning of the way in which `ng-model` is being updated. Also allows
    * for a timezone to be specified. <a href="https://docs.angularjs.org/api/ng/directive/ngModelOptions#usage">Read more at the ngModelOptions docs.</a>
    * @param {expression=} ng-change Expression evaluated when the model value changes.
@@ -444,9 +444,18 @@
 
     // Responds to external changes to the model value.
     self.ngModelCtrl.$formatters.push(function(value) {
+      var parsedValue = angular.isDefined(value) ? Date.parse(value) : null;
+
+      // `parsedValue` is the time since epoch if valid or `NaN` if invalid.
+      if (!isNaN(parsedValue) && angular.isNumber(parsedValue)) {
+        value = new Date(parsedValue);
+      }
+
       if (value && !(value instanceof Date)) {
-        throw Error('The ng-model for md-datepicker must be a Date instance. ' +
-            'Currently the model is a: ' + (typeof value));
+        throw Error(
+          'The ng-model for md-datepicker must be a Date instance or a value ' +
+          'that can be parsed into a date. Currently the model is of type: ' + typeof value
+        );
       }
 
       self.onExternalChange(value);
--- a/src/components/datepicker/js/datepickerDirective.spec.js
+++ b/src/components/datepicker/js/datepickerDirective.spec.js
@@ -105,15 +105,20 @@ describe('md-datepicker', function() {
     expect(controller.inputElement.placeholder).toBe('Fancy new placeholder');
   });
 
-  it('should throw an error when the model is not a date', function() {
-    expect(function() {
-      pageScope.myDate = '2015-01-01';
-      pageScope.$apply();
-    }).toThrowError('The ng-model for md-datepicker must be a Date instance. ' +
-        'Currently the model is a: string');
+  it('should forward the aria-label to the generated input', function() {
+    createDatepickerInstance('<md-datepicker ng-model="myDate" aria-label="Enter a date"></md-datepicker>');
+    expect(controller.ngInputElement.attr('aria-label')).toBe('Enter a date');
   });
 
-  it('should support null and undefined values', function() {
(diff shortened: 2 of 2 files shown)


src/components/datepicker/js/datepickerDirective.js
```diff
@@ -402,9 +402,17 @@
 
     // Responds to external changes to the model value.
     self.ngModelCtrl.$formatters.push(function(value) {
+      var parsedValue = angular.isDefined(value) ? Date.parse(value) : null;
+
+      if (parsedValue) {
+        value = self.dateLocale.parseDate(parsedValue);
```
clshortfuse: Isn't this redundant with `Date.parse(value)` above? Also, `mdDateLocaleProvider`.`parseDate` should expect a string as well, and we're passing a `Date` object here.

src/components/datepicker/js/datepickerDirective.js
```diff
@@ -402,9 +402,17 @@
 
     // Responds to external changes to the model value.
     self.ngModelCtrl.$formatters.push(function(value) {
+      var parsedValue = angular.isDefined(value) ? Date.parse(value) : null;
+
+      if (parsedValue) {
```
clshortfuse: You should change this to `parsedValue == null || isNaN(parsedValue)` or else "1970-01-01T00:00:00.000Z" won't be validated

src/components/datepicker/js/datepickerDirective.js
crisbeto: The difference between `Date.parse` and `new Date` is that `parse` is a little more strict about what it accepts. E.g. the constructor will parse `null`, `false`, etc. into January 1st 1970, whereas `parse` with consider them invalid. Also `parse` returns the time since epoch, which is why it needs to be passed through `parseDate`.

As for the reason why it's not doing the logic in `parseDate`, it's because this should only be done when something, different from the directive, modifies the model.

src/components/datepicker/js/datepickerDirective.js
crisbeto: Good point, although it's better to do `angular.isNumber(parsedValue)`, we generally try to avoid/factor out double equals.

src/components/datepicker/js/datepickerDirective.js
clshortfuse: Okay, I see. If it isn't too much trouble, since you're making a change for L#407, perhaps you can renamed the variable as `parsedNumber` to explain that a number is expected to return. I didn't realize `Date.parse` returned a Number the first time I saw it.

crisbeto: Just did as a part of the change to the comparison.

src/components/datepicker/js/datepickerDirective.js
clshortfuse: Whoops, I inverted the equality in the expression. Note that `angular.isNumber(NaN)` returns `true` so it should be checked second. For example `if (!isNaN(parsedValue) && angular.isNumber(parsedValue))`

crisbeto: I went with the simpler `parsedValue >= 0` which should cover both cases.

crisbeto: Actually that was a bad idea, I'm switching to the isNaN + angular.isNumber.

ThomasBurleson:
@topherfangio - plz review.

src/components/datepicker/js/datepickerDirective.js
```diff
@@ -402,9 +402,18 @@
...
+      if (!isNaN(parsedValue) && angular.isNumber(parsedValue)) {
+        value = new Date(parsedValue);
+      }
+
       if (value && !(value instanceof Date)) {
-        throw Error('The ng-model for md-datepicker must be a Date instance. ' +
-            'Currently the model is a: ' + (typeof value));
+        throw Error(
+          'The ng-model for md-datepicker must be a Date instance or a value ' +
+          'that can be parsed into a date. Currently the model is a: ' + typeof value
```
clshortfuse: nitpick: Change `Currently the model is a` to `Current model: ` because if it's an `Array` it should be `an` instead of `a`.

ThomasBurleson:
@crisbeto - please confirm this PR is still valid/needed. And rebase to lastest in master. 
Thank you.

crisbeto:
Rebased @ThomasBurleson.

ThomasBurleson:
LGTM

PR #9566, opened 2016-09-10
chore(panel): fix failing unit test

crisbeto:
When running the panel tests separately from all the others, the one about disabling scrolling was failing. This wasn't happening when running it together with all the other components, because some other component (not sure exactly which one, but it seemed like either dialog or autocomplete) wasn't cleaning up the DOM properly and was leaving a scroll mask lying around.

This change removes the DOM lookup for the scroll mask, because it looks like it got disabled in PR #8901, however the test didn't get updated.

@bradrich can you confirm that this is desired behaviour?

### DIFF
--- a/src/components/panel/panel.spec.js
+++ b/src/components/panel/panel.spec.js
@@ -675,14 +675,10 @@ describe('$mdPanel', function() {
       spyOn($mdUtil, 'disableScrollAround').and.callThrough();
 
       openPanel(config);
-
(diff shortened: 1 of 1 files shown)


crisbeto:
**Note:** The CI is expected to fail now since this is pending a fix to the autocomplete tests by @DevVersion.

DevVersion:
I based #9568 on **this** PR, so this one should definitely go in before.

PR #9568, opened 2016-09-10
fix(autocomplete): two specs leak the scroll mask element

DevVersion:
* The scroll mask was not removed by two specs of the autocomplete and could cause unexpected issues in other suites, when setting focus on those.


DevVersion:
@ThomasBurleson Please do merge this **after** #9566, because this PR is based on @crisbeto's

PR #9570, opened 2016-09-10
feat(panel): configurable animation duration

### LINKED ISSUE #9177: mdPanel: Ability to modify the duration of a mdPanelAnimation
**Actual Behavior**:
 -  `What is the issue? *`
  -  Within the current mdPanelAnimation API, there is no way to modify the duration of the chosen animation.
  -  There are currently three built-in animations that can be used:  SLIDE, SCALE, and FADE. There is, however, no way to modify the duration of the animation. Adding the duration will allow for the animation to be faster or slower than the default 0.3s.

 -  `CodePen Demo which shows your issue:`  N/A
 -  `Details:`  N/A

**Angular Versions**: *
 -  `Angular Version:`  Latest
 -  `Angular Material Version:`  Latest

**Additional Information**:
 -  `Browser Type: *`  All
 -  `Browser Version: *`  All
 -  `OS: *`  All
 -  `Stack Traces:`  All

Ping @ErinCoughlan.

crisbeto:
* Adds the ability to specify the duration of an animation the panel.
* Fixes the `MdPanelAnimation` not showing up in the docs.

Fixes #9177.

### DIFF
--- a/src/components/panel/demoPanelAnimations/index.html
+++ b/src/components/panel/demoPanelAnimations/index.html
@@ -1,8 +1,8 @@
 <div class="demo-md-panel-animation md-padding" ng-controller="AnimationCtrl as ctrl">
   <h2>Animations</h2>
   <div layout="row">
-    <div flex="33">
-      OpenFrom:
+    <div flex="25">
+      <h3>OpenFrom:</h3>
       <md-radio-group ng-model="ctrl.openFrom">
         <md-radio-button value="button">Button</md-radio-button>
         <md-radio-button value="corner">Top/Left Corner</md-radio-button>
@@ -10,8 +10,8 @@
       </md-radio-group>
     </div>
 
-    <div flex="33">
-      CloseTo:
+    <div flex="25">
+      <h3>CloseTo:</h3>
       <md-radio-group ng-model="ctrl.closeTo">
         <md-radio-button value="button">Button</md-radio-button>
         <md-radio-button value="corner">Top/Left Corner</md-radio-button>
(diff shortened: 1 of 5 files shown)


crisbeto:
@ErinCoughlan @bradrich this currently works on the panel itself, but do you think we should also make this configurable for the backdrop?

bradrich:
I would say yes. The backdrop is in itself a panel as well. I haven't had the chance to look through the code.

crisbeto:
My worry about it is that the backdrop animation config would have to be passed through the panel, not the `newPanelAnimation`. This could make the API confusing. Otherwise we can also take the duration from the panel's `animation` object.

bradrich:
Is there a case where someone would want the backdrop to have a different duration?

crisbeto:
The only case I can think of is if you wanted to prevent the user from clicking/scrolling around while a longer animation was running.

bradrich:
That is true. Maybe we should make the backdrop animation always show at a default 300ms. Then all duration variations could be only on the panel opening above the backdrop. How does that sound?

I am also really looking forward to seeing how you pulled off the animation duration change.

crisbeto:
Removing the `needs: discussion` since we decided to match the backdrop duration with the panel's.

ErinCoughlan:
Nice work! LGTM.

crisbeto:
Updated to have the backdrop inherit the panel's animation duration. Not a huge fan of having to save what the original duration value was, but I couldn't think of a better way of passing it to the backdrop (aside of specifying the units).

crisbeto:
@ThomasBurleson this should be good to go.

ErinCoughlan:
@ThomasBurleson I think this PR is good to go. Can we presubmit/merge it soon?

PR #9574, opened 2016-09-11
feat(panel): add interceptors API and onClose hook

### LINKED ISSUE #9557: mdPanel: Ability to register listeners that are called onClose and can cancel event.
Use Case: I am using mdPanel as a input, and want to warn the user on close that there are unsaved edits. If the user says "Do it anyways", I want to continue the close event. If the user says "Oops, stop.", I want to cancel the close event.

Currently: I can register listeners that are called after close has begun, but nothing before it starts which would allow my to resolve or reject the close event. And I can only register those listeners when I configure the panel, not after it's been created. (`onRemoving` and `onDomRemoved`).

The issue: Panels can be closed using the escape key or by clicking on the background. There's no way for me to hook into that logic and prevent the event from happening (without rewriting that logic on my own.)

Proposed API:
Extend mdPanelRef to keep track of "close listeners". Add three methods:
1. `registerListener(listener: () => angular.IPromise<any>)`
2. `removeListener(listener: () => angular.IPromise<any>)`
3. `removeAllListeners()`

Call the listeners at the beginning of the close promise chain.
```
MdPanelRef.prototype._callListeners = function() {
  var promise = this.$q.when();
  if (this.listeners) {
    for (let listener of this.listeners) {
      promise = promise.then(listener);
    }
  }
  return promise;
};

...

this._callListeners().then(close);
```

crisbeto:
* Adds an API that can be used for registering interceptors on the panel.
* Adds support for an `onClose` interceptor that allows the user to prevent a panel from closing.

Fixes #9557.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -355,6 +355,47 @@ angular
  * @param {!MdPanelPosition} position
  */
 
+/**
+ * @ngdoc method
+ * @name MdPanelRef#registerInterceptor
+ * @description
+ *
+ * Registers an interceptor with the panel. The callback should return a promise,
+ * which will allow the action to continue when it gets resolved, or will
+ * prevent an action if it is rejected. The interceptors are called sequentially
+ * and it reverse order. `type` must be one of the following
+ * values available on `$mdPanel.interceptorTypes`:
+ * * `CLOSE` - Gets called before the panel begins closing.
+ *
+ * @param {string} type Type of interceptor.
+ * @param {!angular.$q.Promise<any>} callback Callback to be registered.
+ * @returns {!MdPanelRef}
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#removeInterceptor
+ * @description
+ *
+ * Removes a registered interceptor.
+ *
+ * @param {string} type Type of interceptor to be removed.
+ * @param {function(): !angular.$q.Promise<any>} callback Interceptor to be removed.
+ * @returns {!MdPanelRef}
+ */
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#removeAllInterceptors
+ * @description
+ *
+ * Removes all interceptors. If a type is supplied, only the
+ * interceptors of that type will be cleared.
+ *
+ * @param {string=} type Type of interceptors to be removed.
+ * @returns {!MdPanelRef}
+ */
+
 
 /*****************************************************************************
  *                               MdPanelPosition                            *
@@ -740,6 +781,12 @@ function MdPanelService($rootElement, $rootScope, $injector, $window) {
    * @type {enum}
    */
   this.yPosition = MdPanelPosition.yPosition;
+
+  /**
+   * Possible values for the interceptors that can be registered on a panel.
+   * @type {enum}
+   */
+  this.interceptorTypes = MdPanelRef.interceptorTypes;
 }
 
 
@@ -909,7 +956,18 @@ function MdPanelRef(config, $injector) {
 
   /** @private {Function?} */
   this._restoreScroll = null;
+
+  /**
+   * Keeps track of all the panel interceptors.
+   * @private {!Object}
+   */
+  this._interceptors = Object.create(null);
 }
+
+
+MdPanelRef.interceptorTypes = {
+  CLOSE: 'onClose'
+};
 
 
 /**
@@ -941,13 +999,15 @@ MdPanelRef.prototype.close = function() {
   var self = this;
 
   return this._$q(function(resolve, reject) {
-    var done = self._done(resolve, self);
-    var detach = self._simpleBind(self.detach, self);
+    self._callInterceptors(MdPanelRef.interceptorTypes.CLOSE).then(function() {
+      var done = self._done(resolve, self);
+      var detach = self._simpleBind(self.detach, self);
 
-    self.hide()
-        .then(detach)
-        .then(done)
-        .catch(reject);
+      self.hide()
+          .then(detach)
+          .then(done)
+          .catch(reject);
+    }, reject);
   });
 };
 
@@ -1039,6 +1099,7 @@ MdPanelRef.prototype.detach = function() {
 MdPanelRef.prototype.destroy = function() {
   this.config.scope.$destroy();
   this.config.locals = null;
+  this._interceptors = null;
 };
 
 
@@ -1634,6 +1695,106 @@ MdPanelRef.prototype._animateClose = function() {
     animationConfig.animateClose(self.panelEl)
         .then(done, warnAndClose);
   });
+};
+
+/**
+ * Registers a interceptor with the panel. The callback should return a promise,
+ * which will allow the action to continue when it gets resolved, or will
+ * prevent an action if it is rejected.
+ * @param {string} type Type of interceptor.
+ * @param {!angular.$q.Promise<!any>} callback Callback to be registered.
+ * @returns {!MdPanelRef}
+ */
+MdPanelRef.prototype.registerInterceptor = function(type, callback) {
+  var error = null;
+
+  if (!angular.isString(type)) {
+    error = 'Interceptor type must be a string, instead got ' + typeof type;
+  } else if (!angular.isFunction(callback)) {
+    error = 'Interceptor callback must be a function, instead got ' + typeof callback;
+  }
+
+  if (error) {
+    throw new Error('MdPanel: ' + error);
+  }
+
+  var interceptors = this._interceptors[type] = this._interceptors[type] || [];
+
+  if (interceptors.indexOf(callback) === -1) {
+    interceptors.push(callback);
+  }
+
+  return this;
+};
+
+/**
+ * Removes a registered interceptor.
+ * @param {string} type Type of interceptor to be removed.
+ * @param {Function} callback Interceptor to be removed.
+ * @returns {!MdPanelRef}
+ */
+MdPanelRef.prototype.removeInterceptor = function(type, callback) {
+  var index = this._interceptors[type] ?
+    this._interceptors[type].indexOf(callback) : -1;
+
+  if (index > -1) {
+    this._interceptors[type].splice(index, 1);
+  }
+
+  return this;
+};
+
+
+/**
+ * Removes all interceptors.
+ * @param {string=} type Type of interceptors to be removed.
+ *     If ommited, all interceptors types will be removed.
+ * @returns {!MdPanelRef}
+ */
+MdPanelRef.prototype.removeAllInterceptors = function(type) {
+  if (type) {
+    this._interceptors[type] = [];
+  } else {
+    this._interceptors = Object.create(null);
+  }
+
+  return this;
+};
+
+
+/**
+ * Invokes all the interceptors of a certain type sequantially in
+ *     reverse order. Works in a similar way to `$q.all`, except it
+ *     respects the order of the functions.
+ * @param {string} type Type of interceptors to be invoked.
+ * @returns {!angular.$q.Promise<!MdPanelRef>}
+ * @private
+ */
+MdPanelRef.prototype._callInterceptors = function(type) {
+  var self = this;
+  var $q = self._$q;
+  var interceptors = self._interceptors && self._interceptors[type] || [];
+
+  return interceptors.reduceRight(function(promise, interceptor) {
+    var isPromiseLike = interceptor && angular.isFunction(interceptor.then);
+    var response = isPromiseLike ? interceptor : null;
+
+    /**
+    * For interceptors to reject/cancel subsequent portions of the chain, simply
+    * return a `$q.reject(<value>)`
+    */
+    return promise.then(function() {
+      if (!response) {
+        try {
+          response = interceptor(self);
+        } catch(e) {
+          response = $q.reject(e);
+        }
+      }
+
+     return response;
+    });
+  }, $q.resolve(self));
 };
 
 
--- a/src/components/panel/panel.spec.js
+++ b/src/components/panel/panel.spec.js
@@ -2147,6 +2147,213 @@ describe('$mdPanel', function() {
     });
   });
 
+  describe('interceptor logic: ', function() {
+    var interceptorTypes;
+
+    beforeEach(function() {
+      interceptorTypes = $mdPanel.interceptorTypes;
+      openPanel();
+    });
+
+    it('should throw when registering an interceptor without a type', function() {
+      expect(function() {
+        panelRef.registerInterceptor();
+      }).toThrowError('MdPanel: Interceptor type must be a string, instead got undefined');
+    });
+
+    it('should throw when registering an interceptor without a callback', function() {
+      expect(function() {
+        panelRef.registerInterceptor(interceptorTypes.CLOSE);
+      }).toThrowError('MdPanel: Interceptor callback must be a function, instead got undefined');
+    });
+
+    it('should execute the registered interceptors', function() {
+      var obj = { callback: function() {} };
+
+      spyOn(obj, 'callback');
+
+      panelRef.registerInterceptor(interceptorTypes.CLOSE, obj.callback);
+      callInteceptors('CLOSE');
+
+      expect(obj.callback).toHaveBeenCalledWith(panelRef);
+    });
+
+    it('should execute the interceptors in reverse order', function() {
+      var results = [];
+      var obj = { callback: function() {} };
+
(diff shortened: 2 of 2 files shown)


crisbeto:
@ErinCoughlan @bradrich do you think we should move the other callbacks to this API? This would be a breaking change, though.

src/components/panel/panel.js
```diff
@@ -1636,6 +1685,91 @@ MdPanelRef.prototype._animateClose = function() {
...
+};
+
+
+/**
+ * Invokes all the listeners from a specific type.
+ * @param {string} type Type of listeners to be invoked.
+ * @returns {!angular.$q.Promise<!MdPanelRef>}
+ * @private
+ */
+MdPanelRef.prototype._callListeners = function(type) {
```
ThomasBurleson: please change to `_notifyListeners`

src/components/panel/panel.js
```diff
@@ -1636,6 +1685,91 @@ MdPanelRef.prototype._animateClose = function() {
...
 
+/**
+ * Registers a listener with the panel. The callback should return a promise,
+ * which will allow the action to continue when it gets resolved, or will
+ * prevent an action if it is rejected.
+ * @param {string} type Type of listener.
+ * @param {!angular.$q.Promise<!any>} callback Callback to be registered.
+ * @returns {!MdPanelRef}
+ */
+MdPanelRef.prototype.registerListener = function(type, callback) {
```
ThomasBurleson: I think these methods should be moved to a `/utils/CallbackRegistry.js` for generic usages.

```
this.callbacks = new CallbackRegistry(this);

this.callbacks.add(type, callback);
this.callbacks.remove(type, callback);

this.callbacks.notify(type);
this.callbacks.removeAll(type);

```

crisbeto: Perhaps this ties in with a more generic use for the service introduced with #9514? I went with this approach mainly because of the spec by @ErinCoughlan in #9557. 

As an aside, I was initially considering something similar to the browser API where you could do `event.preventDefault` to stop the panel from closing, but using promises should be more flexible since it allows you to determine the close condition asynchronously.

src/components/panel/panel.js
ErinCoughlan: Actually, we had a huge debate at work over naming, and technically these are interceptors, not listeners because they can approve/reject the close. So I think it should be called `_callInterptors`. Notify sounds like we don't care about the response, but we do.

src/components/panel/panel.js
ErinCoughlan: @crisbeto We had debated following the Event based model, but promises worked differently so we went with this API. (We meaning my team since I made a similar API internally and they had to debate/approve.)

src/components/panel/panel.js
```diff
@@ -355,6 +355,45 @@ angular
...
+ * @ngdoc method
+ * @name MdPanelRef#registerListener
+ * @description
+ *
+ * Registers a listener with the panel. The callback should return a promise,
+ * which will allow the action to continue when it gets resolved, or will
+ * prevent an action if it is rejected. Currently supported listener types:
+ * * `onClose` - Gets called before the panel begins closing.
+ *
+ * @param {string} type Type of listener.
```
ErinCoughlan: You should make an enum for the possible types to prevent user errors and allow better type checking. (Really useful in typescript typings.)

src/components/panel/panel.js
```diff
@@ -355,6 +355,45 @@ angular
...
+ * @param {!angular.$q.Promise<!any>} callback Callback to be registered.
```
ErinCoughlan: nit: ```{!angular.$q.Promise<any>}```

src/components/panel/panel.js
crisbeto: Could you give me an example?

src/components/panel/panel.js
```diff
@@ -909,6 +948,13 @@ function MdPanelRef(config, $injector) {
 
   /** @private {Function?} */
   this._restoreScroll = null;
+
+  /**
+   * Keeps track of all the panel listeners.
+   * @type {!Object}
```
ErinCoughlan: nit: combine the type and `@private` annotations.

```@private {!Object}```

File: src/components/panel/panel.js [hunk 6]
```diff
@@ -941,13 +987,15 @@ MdPanelRef.prototype.close = function() {
...
 
   return this._$q(function(resolve, reject) {
-    var done = self._done(resolve, self);
-    var detach = self._simpleBind(self.detach, self);
-
-    self.hide()
-        .then(detach)
-        .then(done)
-        .catch(reject);
+    self._callListeners('onClose').then(function() {
```
ErinCoughlan: enum will help you here too.

src/components/panel/panel.js
```diff
@@ -1636,6 +1685,91 @@ MdPanelRef.prototype._animateClose = function() {
...
+ * prevent an action if it is rejected.
+ * @param {string} type Type of listener.
+ * @param {!angular.$q.Promise<!any>} callback Callback to be registered.
+ * @returns {!MdPanelRef}
+ */
+MdPanelRef.prototype.registerListener = function(type, callback) {
+  var error = null;
+
+  if (!angular.isString(type)) {
+    error = 'Listener type must be a string, instead got ' + typeof type;
```
ErinCoughlan: I think we wanted to prefix error messages with MdPanel to make it more obvious where they come from.

src/components/panel/panel.js
```diff
@@ -355,6 +355,45 @@ angular
...
+
+/**
+ * @ngdoc method
+ * @name MdPanelRef#removeListener
+ * @description
+ *
+ * Removes a registered listener.
+ *
+ * @param {string} type Type of listener to be removed.
+ * @param {Function} callback Listener to be removed.
```
ErinCoughlan: `{!Function}` or my preferred version because it says what the form of the function is: `function(): !angular.$q.Promise<any>`

src/components/panel/panel.js
```diff
@@ -1636,6 +1685,91 @@ MdPanelRef.prototype._animateClose = function() {
...
+    error = 'Listener type must be a string, instead got ' + typeof type;
+  } else if (!angular.isFunction(callback)) {
+    error = 'Listener callback must be a function, instead got ' + typeof callback;
+  }
+
+  if (error) {
+    throw new Error(error);
+  }
+
+  var listeners = this._listeners[type] = this._listeners[type] || [];
```
ErinCoughlan: Does this work:

```
var listeners = this._listeners[type] || [];
```

crisbeto: It does, but that way it won't initialize it if it doesn't exist.

src/components/panel/panel.js
```diff
@@ -1636,6 +1685,91 @@ MdPanelRef.prototype._animateClose = function() {
...
+  return this;
+};
+
+/**
+ * Removes a registered listener.
+ * @param {string} type Type of listener to be removed.
+ * @param {Function} callback Listener to be removed.
+ * @returns {!MdPanelRef}
+ */
+MdPanelRef.prototype.removeListener = function(type, callback) {
```
ErinCoughlan: A simple version of this method:

```
this._listeners[type] = this._listeners[type].filter(c => c !== callback);
```

src/components/panel/panel.js
```diff
@@ -1636,6 +1685,91 @@ MdPanelRef.prototype._animateClose = function() {
...
+  }
+
+  return this;
+};
+
+
+/**
+ * Removes all listeners.
+ * @param {string=} type Type of listeners to be removed.
+ * If ommited, all listeners types will be removed.
```
ErinCoughlan: nit: indent this 4 spaces

src/components/panel/panel.js
```diff
@@ -1636,6 +1685,91 @@ MdPanelRef.prototype._animateClose = function() {
...
+
+  if (listeners && listeners.length) {
+    var invokedListeners = listeners.map(function(currentListener) {
+      return currentListener(this);
+    }, this);
+
+    return this._$q.all(invokedListeners);
+  }
+
+  return this._$q.when(this);
```
ErinCoughlan: when -> resolve

src/components/panel/panel.js
```diff
@@ -1636,6 +1685,91 @@ MdPanelRef.prototype._animateClose = function() {
...
+ */
+MdPanelRef.prototype._callListeners = function(type) {
+  var listeners = this._listeners && this._listeners[type];
+
+  if (listeners && listeners.length) {
+    var invokedListeners = listeners.map(function(currentListener) {
+      return currentListener(this);
+    }, this);
+
+    return this._$q.all(invokedListeners);
```
ErinCoughlan: How does order work here? I thought the listeners should be called in reverse order of them being registered, so something like:

```
var promise = this._$q.resolve(this);

if (listeners) {
  for (var i = listeners.length - 1; i >= 0; i--) {
    var listener = listener[i];
    promise = promise.then(listeners);
  }
}

return listener;
```

src/components/panel/panel.js
ErinCoughlan: Look at MdPanelPosition xPosition and yPosition.

src/components/panel/panel.js
crisbeto: Sounds good, I suppose the `registerListener` etc. will be renamed as well.

src/components/panel/panel.js
crisbeto: While it's a little simpler, it would overwrite the listeners array every time.

src/components/panel/panel.js
crisbeto: Reversing them before calling shouldn't be a big deal, however I don't know if we should count on the order being correct, because they're using promises and that implies that it's async.

ErinCoughlan: If there's a then between each promise, the order is enforced since the previous one must resolve/reject before the then is called.

crisbeto: Alright, that makes sense. I'll refactor it.

src/components/panel/panel.js
crisbeto: We did, although that's probably more appropriate once we have most of these PRs merged.

src/components/panel/panel.js
ThomasBurleson: This functionality ^ needs to be refactored as a external util class.

crisbeto:
@ErinCoughlan @bradrich I've updated the PR with the following:
* The "listeners" are renamed "interceptors".
* The callbacks are now called one after another, instead of in parallel.
* The interceptor names are in an enum like the positions and animations.
* Some of the other feedback on the documentation.

src/components/panel/panel.js
```diff
@@ -1636,6 +1697,106 @@ MdPanelRef.prototype._animateClose = function() {
...
+
+/**
+ * Invokes all the interceptors of a certain type sequantially in
+ *     reverse order. Works in a similar way to `$q.all`, except it
+ *     respects the order of the functions.
+ * @param {string} type Type of interceptors to be invoked.
+ * @returns {!angular.$q.Promise<!MdPanelRef>}
+ * @private
+ */
+MdPanelRef.prototype._callInterceptors = function(type) {
```
ThomasBurleson: Let me be gentle and honest: this is HORRIBLE code here! 

Please use:

```js
MdPanelRef.prototype._callInterceptors = function(type) {
 var self = this, $q = this._$q,  seed = this.$q.resolve(self),
     isPromiseLike = function (obj) { return obj && isFunction(obj.then); },  // Same as in angular core
     gatherInterceptorsFor = function(interceptors, type) { return interceptors[type] || [ ]; },
     interceptors = gatherInterceptorsFor(self._interceptors || [ ], type);

     /**
      * Create a chain/sequence of interceptor callbacks
      */
     return interceptors.reduce(function(promise, interceptor){
       var response = isPromiseLike(interceptor) ? interceptor : null;

       /**
        * @todo - this chain does not allow interceptor results to reject/cancel
        *         subsequent portions of the chain.
        */
       return promise.then(function(result) {
         if ( !response ) {
           try {
                response = interceptor(self);
           } catch(e) {
             response = $q.rejected(e);
           }
         }
         return response;
       });

     },seed);
```

src/components/panel/panel.js
```diff
@@ -941,13 +999,15 @@ MdPanelRef.prototype.close = function() {
...
   return this._$q(function(resolve, reject) {
-    var done = self._done(resolve, self);
-    var detach = self._simpleBind(self.detach, self);
-
-    self.hide()
-        .then(detach)
-        .then(done)
-        .catch(reject);
+    self._callInterceptors(MdPanelRef.interceptorTypes.CLOSE).then(function() {
+      var done = self._done(resolve, self);
```
ThomasBurleson: `function _done()` is VERY confusing. Similar to the angular core code `valueFn`, use

```js
function _callbackFn() { return function callbackRef() { return callback(self); } }
```

src/components/panel/panel.js
crisbeto: I'll look into rewriting that method, but the main idea was to ensure that: 
1. The interceptors execute in the proper sequence. This proposal will run them in parallel and only the callbacks will be chained. 
2. Allows breaking the chain with a rejection.

src/components/panel/panel.js
```diff
@@ -941,13 +999,15 @@ MdPanelRef.prototype.close = function() {
...
+      var detach = self._simpleBind(self.detach, self);
+
+      self.hide()
+          .then(detach)
+          .then(done)
```
ThomasBurleson: Declaring locals vars here is obfuscating IMO. 

```
self.hide()
    .then( self._callbackFn(resolve, self) )
    .then( self._simpleBind(self.detach, self) )
    .catch( reject )
```

crisbeto: This is how it used to work before, I only wrapped it in the interceptor.

src/components/panel/panel.js
ThomasBurleson: This reduce function runs the interceptor callbacks in a sequence.

crisbeto: Alright, my bad, that works pretty well. At first glance last night it seemed like it would only chain the callbacks (`.then`), but not promises themselves (if that makes sense).

ThomasBurleson: @crisbeto - we need significant unit testing for this `_callInterceptors()` logic:

* queue with promises
* queue with interceptor exceptions
* queue with interceptor response values or promises
* empty queue

etc.

src/components/panel/panel.spec.js
```diff
@@ -2147,6 +2147,192 @@ describe('$mdPanel', function() {
...
+      openPanel();
+      spyOn(obj, 'callback');
+
+      panelRef.registerInterceptor(interceptorTypes.CLOSE, obj.callback);
+      panelRef._callInterceptors(interceptorTypes.CLOSE);
+
+      expect(obj.callback).toHaveBeenCalledWith(panelRef);
+    });
+
+    it('should execute the interceptors in reverse order', function() {
```
ThomasBurleson: Why should it execute in reverse order ?

src/components/panel/panel.js
ThomasBurleson: The unit tests says that the interceptors should be executed in **reverse** order?
Then you will need:

```js
return interceptorsForType(self._interceptors || [ ], type)
      .reverse(    )
      .reduce(... )
```

src/components/panel/panel.spec.js
```diff
@@ -2147,6 +2147,192 @@ describe('$mdPanel', function() {
...
+
+      function makePromise(value) {
+        return function() {
+          results.push(value);
+          return $q.resolve();
+        };
+      }
+    });
+
+    it('should reject if one of the promises fails', function() {
```
ThomasBurleson: Also need a test if an interceptor callback throws an exception.

src/components/panel/panel.spec.js
crisbeto: This was agreed upon with @ErinCoughlan. I think it's also the way Angular runs `$formatters` and `$parsers`.

src/components/panel/panel.js
crisbeto: Will do regarding the unit tests, as for doing it in reverse order, there's the `reduceRight` method.

ThomasBurleson: LOL - yes, I forgot that one: `reduceRight()`

crisbeto:
@ThomasBurleson the `_callInterceptors` method has been refactored according to the discussion and I've added a few extra unit tests.

PR #9585, opened 2016-09-12
feat(tooltip): Tooltip using MdPanel API

bradrich:
Tooltip has been reformated from an Angular directive to an Angular component. It is also using the MdPanel API to position, animate open and closed, and show the tooltip as a result of user interaction.

Ping @ErinCoughlan @crisbeto

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -1,393 +1,366 @@
+var MdTooltipComponent = {
+  controller: MdTooltipController,
+  bindings: {
+    mdDelay: '<?',
+    mdVisible: '<?',
+    mdAutohide: '<?',
+    mdDirection: '<?'
+  }
+};
+
+
 /**
  * @ngdoc module
  * @name material.components.tooltip
  */
 angular
-  .module('material.components.tooltip', [ 'material.core' ])
-  .directive('mdTooltip', MdTooltipDirective)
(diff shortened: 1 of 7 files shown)


src/components/tooltip/tooltip.js
```diff
@@ -1,393 +1,366 @@
...
       }
-      content.css('transform-origin', origin);
-    }
 
-    function onVisibleChanged (isVisible) {
-      if (isVisible) showTooltip();
-      else hideTooltip();
-    }
+      function onDestroy() {
+        angular.element($window)
```
crisbeto: You can remove these. They're handled by the `$$mdTooltipRegistry`.

src/components/tooltip/tooltip.js
```diff
@@ -1,393 +1,366 @@
...
+        $element.remove();
 
-    function addAriaLabel (override) {
-      if ((override || !parent.attr('aria-label')) && !parent.text().trim()) {
-        var rawText = override || element.text().trim();
-        var interpolatedText = $interpolate(rawText)(parent.scope());
-        parent.attr('aria-label', interpolatedText);
+        // Trigger the handler in case the tooltip was still visible.
+        leaveEventHandler();
+        document.removeEventListener('scroll', windowScrollEventHandler, true);
```
crisbeto: This is also handled by the tooltip registry.

bradrich:
There is no current polyfill for components yet, so we can't use the component API.

PR #9586, opened 2016-09-12
fix(switch): invalid container margin in RTL

DevVersion:
* The switch does not properly display in RTL mode - this is due to the recent changes to the RTL mixin, which introduced a reset value.
* Inside of the switch the container should have a 0px reset value

| Before | After |
| ------ | ------ |
|
image | image |

### DIFF
--- a/src/components/switch/switch.scss
+++ b/src/components/switch/switch.scss
@@ -44,7 +44,7 @@ md-switch {
     height: $switch-height;
     position: relative;
     user-select: none;
-    @include rtl-prop(margin-right, margin-left, 8px, auto);
+    @include rtl-prop(margin-right, margin-left, 8px, 0px);
     float: left;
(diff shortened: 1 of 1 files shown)


clshortfuse:
LGTM

The `margin` propery has a default value of `0`, not `auto`, which is what we want here.

src/components/switch/switch.scss
```diff
@@ -44,7 +44,7 @@ md-switch {
     height: $switch-height;
     position: relative;
     user-select: none;
-    @include rtl-prop(margin-right, margin-left, 8px, auto);
+    @include rtl-prop(margin-right, margin-left, 8px, 0px);
```
clshortfuse: nit: use `0` instead of `0px` (minifying CSS will do this anyway)

PR #9606, opened 2016-09-14
fix(tooltip): properly interpolate tooltip text to prevent possible XSS

### LINKED ISSUE #9605: tooltip: XSS due to interpolating values twice
Tooltip element values are interpolated twice if an expression is found in the interpolated value of the tooltip element. As such, XSS might be possible if the value is derived from user input.

The code responsible for interpolating an interpolated value can be found at https://github.com/angular/material/blob/master/src/components/tooltip/tooltip.js#L164.

A PoC to show that expressions are evaluated twice can also be found here: http://jsfiddle.net/bf3u47f3/8/. 
Notice that the `aria-label` attribute has a value of `1337`, which signifies that the value was interpolated twice.

quanyang:
- Change tooltip controller to watch only for changes in data value (if
  any) of the tooltip text.
- Modify tooltip controller so that interpolated values would not be
  interpolated again, preventing possible XSS.

Fixes #9605


src/components/tooltip/tooltip.js
```diff
@@ -104,8 +104,9 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
 
     function addAriaLabel(override) {
```
ThomasBurleson: @DevVersion - please add comment (in another PR) that describes what `override` does?  

> One has to read the entire component code to release that it is used as the changeHandler in a `.$watch( )`; so the `override` is the value interpolated from $watch.

jelbourn:
Closed in favor of #10159 due to rebase issues

PR #9609, opened 2016-09-14
fix(panel): allow numbers in offset methods

### LINKED ISSUE #9604: mdPanel: assume that number is in pixels
Currently passing in a plain number doesn't work with the `withOffsetX` and `withOffsetY` methods, because they only accept CSS units. In this case we should append the `px` unit automatically.

Referencing #9564.

crisbeto:
Allows for numbers to be pass to the `withOffsetX` and `withOffsetY` methods, assuming that the units are pixels.

Fixes #9604.

CC @ErinCoughlan @bradrich

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -676,7 +676,7 @@ angular
  * @description
  * Sets the value of the offset in the x-direction.
  *
- * @param {string} offsetX
+ * @param {string|number} offsetX
  * @returns {!MdPanelPosition}
  */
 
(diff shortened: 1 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -2553,3 +2553,12 @@ function getElement(el) {
       document.querySelector(el) : el;
   return angular.element(queryResult);
 }
+
+/**
+ * Adds units to a number value.
+ * @param {number} value
```
ErinCoughlan: `@param {string | number} value`

crisbeto:
@ThomasBurleson this should be ready for presubmit.

PR #9613, opened 2016-09-15
feat(navBar): add disabled attribute

maxjoehnk:
- Add disabled attribute
- Update demo to showcase the disabled attribute


maxjoehnk:
I signed it

maxjoehnk:
I've signed it!

src/components/navBar/navBar.js
```diff
@@ -414,6 +418,7 @@ function MdNavItem($$rAF) {
       'mdNavHref': '@?',
       'mdNavSref': '@?',
       'name': '@',
+      'disabled': '=?ngDisabled',
```
DevVersion: I think we should rather observe the `disabled` attribute itself, instead of the `ngDisabled` attribute`.

> `ngDisabled` changes the `disabled` attribute at runtime 

We also do this on the autocomplete and other components, this also allows developers to apply the `disabled` attribute, without having to use the `ng-disabled` attribute.


```html
<md-x-item disabled="disabled">
```

src/components/navBar/navBar.js
```diff
@@ -82,6 +82,7 @@ angular.module('material.components.navBar', ['material.core'])
  * clicked. Renders as a `ui-sref`.
  * @param {string=} name The name of this link. Used by the nav bar to know
  * which link is currently selected.
+ * @param {boolean=} ngDisabled If present and expression evaluates to truthy, disabled item selection.
```
DevVersion: If addressing the comment below, then this should be `disabled`.

src/components/navBar/navBar.js
maxjoehnk: I'm using the ngDisabled attribute here because of the tabs directive which is also using ngDisabled.
I feel like the behaviour of those components should be as equal as possible.
So I think when we use the disabled attribute here we should change it in the tabs component too.

DevVersion: Yeah that's why `disabled` doesn't work on the tab item as well.

- http://codepen.io/DevVersion/pen/rrrBzL

Watching for the `disabled` attribute ensures that `ngDisabled` and the normal `disabled` attribute work.

maxjoehnk: I've pushed support for the disabled attribute, but the implementation does feel kinda hacky because I can't just get the disabled attribute from the element.
I've tried using $observe on the disabled attribute (and then using disabled: '@?' instead of nativeDisabled but I guess this version should be more performant.

I will add tests when the implementation is finished and reviewed.

src/components/navBar/navBar.js
```diff
@@ -434,6 +439,21 @@ function MdNavItem($$rAF) {
...
+      var stopDisabledWatch = angular.noop;
+
+      if (angular.isDefined(attrs.disabled)) {
+        mdNavItem.disabled = true;
+      }
+      else if (attrs.ngDisabled) {
+        stopDisabledWatch = scope.$parent.$watch(attrs.ngDisabled, function() {
+          mdNavItem.disabled = $parse(attrs.ngDisabled)(scope.$parent);
+        });
+      }
```
DevVersion: I thought about something similar to the autocomplete

```js
$attrs.$observe('disabled', function (value) {
  mdNavItem.disabled= $mdUtil.parseAttributeBoolean(value, false); 
});
```

This should be everything you need to do.

maxjoehnk: I didn't use $observe because it adds another watcher, but now that I think about it I can remove the ngDisabled watcher so it doesn't matter anyway.

src/components/navBar/demoBasicUsage/index.html
```diff
@@ -13,3 +14,11 @@
       External content for `<span>{{currentNavItem}}</span>`
     </div>
   </md-content>
+
+  <div class="md-padding" layout="row" layout-sm="column" layout-align="left center"
+       style="padding-top: 0;">
+    <md-checkbox ng-model="data.secondLocked" aria-label="Disable item two?" style="margin: 5px;">
+      Disable item two?
```
DevVersion: Can we change this to `Disable Second Item` only?

src/components/navBar/demoBasicUsage/index.html
```diff
@@ -13,3 +14,11 @@
       External content for `<span>{{currentNavItem}}</span>`
     </div>
   </md-content>
+
+  <div class="md-padding" layout="row" layout-sm="column" layout-align="left center"
+       style="padding-top: 0;">
```
DevVersion: Please remove the `layout` and `layout-align` here (we don't need the alignment and a layout for now)

src/components/navBar/navBar.js
DevVersion: @maxjoehnk Ping for this comment as well.

src/components/navBar/navBar.js
```diff
@@ -414,6 +418,7 @@ function MdNavItem($$rAF) {
       'mdNavHref': '@?',
       'mdNavSref': '@?',
       'name': '@',
+      'disabled': '@?'
```
DevVersion: This can be removed now, because we have the $observe.

maxjoehnk:
I've implemented the requested changes.
The disabled item is still not really distinguishable from the enabled items.
Should it just be more transparent or how shall we approach this?
And shouldn't this be implemented by the button component already?

maxjoehnk:
TIL rebasing isn't a good idea on a pull request... -.-'

maxjoehnk:
Since I've fucked up this pull request I've created a new one.
@DevVersion This should be ready to merge, I've fixed the styling and the demo page.

#10202

PR #9615, opened 2016-09-15
feat(panel): allow passing in a function to the offset methods

### LINKED ISSUE #9608: mdPanel: allow callback in offset functions
The `withOffsetX` and `withOffsetY` methods currently require the offset to be known ahead of time, however this doesn't work too well if the offset is based on the panel itself. This can be worked around by setting the offset in the `open` callback, however it can cause the panel to jump on slower browsers (IE and Edge). The methods should support passing in a function that allows for the offset to be determined dynamically. Example:

```js
mdPanelPosition.withOffsetX(function(panelRef) {
  return dynamicallyDeterminedOffset;
});
```

Referencing #9564.

crisbeto:
* Allows for a function to be passed to the `withOffsetX` and `withOffsetY` methods which will help when trying to determine the offset dynamically.
* Switches to assigning the offset before the position, in order to avoid tiny jumps in the UI on slower browsers.

Fixes #9608.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -1337,6 +1337,12 @@ MdPanelRef.prototype._updatePosition = function(init) {
   var positionConfig = this.config['position'];
 
   if (positionConfig) {
+    // Use the vendor prefixed version of transform.
+    // Note that the offset should be assigned before the position, in
+    // order to avoid tiny jumps in the panel's position, on slower browsers.
+    var prefixedTransform = this._$mdConstant.CSS.TRANSFORM;
+    this.panelEl.css(prefixedTransform, positionConfig.getTransform());
+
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@ErinCoughlan - ping for review.

src/components/panel/panel.js
```diff
@@ -2003,7 +2005,7 @@ MdPanelPosition.prototype._validateXPosition = function(xPosition) {
 /**
  * Sets the value of the offset in the x-direction. This will add to any
  * previously set offsets.
- * @param {string} offsetX
+ * @param {string|Function(MdPanelPosition): string} offsetX
```
ErinCoughlan: nit: Function -> function

PR #9631, opened 2016-09-17
fix(panel): consolidate redundant validations

crisbeto:
- The panel has the `_validateXPosition` and `_validateYPosition` methods which do the exact same thing, except on a different object. This moves that logic to a function in order to avoid the duplicated code.
- Removes a use of double equals in the validation function.

CC @ErinCoughlan @bradrich


crisbeto:
@ThomasBurleson this should also be ready for presubmit.

Issue #9273, opened 2016-08-09
md-dialog: delayed focus on opening dialog


clshortfuse:
**Actual Behavior**:
 -  `What is the issue? *` dialogs have a long animation delay when opening and the default button is not focused until animation closed.
 -  `What is the expected behavior?` default button should have focus during it's animation state

**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:` https://material.angularjs.org/latest/demo/dialog
 -  `Details:` Click Alert Dialog and press (Enter) not too long after. If you don't wait until the OK button turn gray (have focus), it'll click again.

**Angular Versions**: *
 -  `Angular Version:` 1.5.5
 -  `Angular Material Version:` 1.1.0-rc5 (master)

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *` V52
 -  `OS: *` ChromeOS
 -  `Stack Traces:`

----

This is an accessibility issue where the dialog is fully present to the user, but the user must wait for a  before committing an action. This may sound trivial for dialog buttons where the user rarely ever sees, but for repeated tasks (especially those that use prompts based on keyevents), it becomes rather confusing.

For comparison, check out Chrome's `window.alert()` system. The OK button gets focus before the animation is complete, and you can press (Enter) before animation is done. Custom dialogs probably need the ability to disable this feature, but our default ones should work like this. 

At the very, whatever had focused should be trapped to ensure that if you hit Enter early, you're not sending events to the previous element. Holding the events and waiting a digest cycle after animation ends is also another option, though I honestly rather see focus without waiting for the animation.

https://www.w3.org/TR/wai-aria-practices/#trap_focus

clshortfuse:
From: https://www.w3.org/TR/wai-aria-practices/#make_modal

>Before the dialog is shown, create and display the dialog underlay. Connect an onkeypress event handler to the DOM document.documentElement. This will catch all keystrokes on the document and allow trapping keyboard focus within the dialog. Size and position the dialog box in the viewport above the underlay, make it visible and set focus to the first tab focusable item in the dialog box.

The key here is to hook keypress events before making it visible.

DevVersion:
So I tried to work on that the last days and it seems to require a lot of work.

> To get it working we need to rework the `md-autofocus` directive, because if we want to auto-focus at the time the animation starts the attribute expression needs to be evaluated before the next digest runs.

[...]

PR #9636, opened 2016-09-18
fix(autofocus) properly parse attribute expression

DevVersion:
The autofocus directive currently supports an empty attribute value because `jqLite` toggles the class if the `force` param is `undefined`.

> This means if you change it several times to the same `undefined` value it does incorrectly update the autofocus class (See associated test)

Also the autofocus directive does now determine the class immediately and does not require a second digest to run (The previous delay could cause issues with the $mdCompiler element linking).

References #9273.

@ThomasBurleson: This is a preparation fix to address the aforementioned issue with the $mdPanel and just to make the autofocus directive _faster_.

### DIFF
--- a/src/core/util/autofocus.js
+++ b/src/core/util/autofocus.js
@@ -90,18 +90,40 @@ angular.module('material.core')
  * </div>
  * </hljs>
  **/
-function MdAutofocusDirective() {
+function MdAutofocusDirective($parse) {
   return {
     restrict: 'A',
-    link: postLink
+    link: {
+      pre: preLink
+    }
   };
-}
 
-function postLink(scope, element, attrs) {
-  var attr = attrs.mdAutoFocus || attrs.mdAutofocus || attrs.mdSidenavFocus;
+  function preLink(scope, element, attr) {
+    var attrExp = attr.mdAutoFocus || attr.mdAutofocus || attr.mdSidenavFocus;
 
-  // Setup a watcher on the proper attribute to update a class we can check for in $mdUtil
-  scope.$watch(attr, function(canAutofocus) {
-    element.toggleClass('md-autofocus', canAutofocus);
-  });
+    // Initially update the expression by manually parsing the expression as per $watch source.
+    updateExpression($parse(attrExp)(scope));
+
+    // Only watch the expression if it is not empty.
+    if (attrExp) {
+      scope.$watch(attrExp, updateExpression);
+    }
+
+    /**
+     * Updates the autofocus class which is used to determine whether the attribute
+     * expression evaluates to true or false.
+     * @param {string|boolean} value Attribute Value
+     */
+    function updateExpression(value) {
+
+      // Rather than passing undefined to the jqLite toggle class function we explicitly set the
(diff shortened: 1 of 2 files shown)


src/core/util/autofocus.js
```diff
@@ -90,18 +90,37 @@ angular.module('material.core')
...
+    link: {
+      pre: preLink
+    }
   };
-}
 
-function postLink(scope, element, attrs) {
-  var attr = attrs.mdAutoFocus || attrs.mdAutofocus || attrs.mdSidenavFocus;
+  function preLink(scope, element, attr) {
+    var attrExp = attr.mdAutoFocus || attr.mdAutofocus || attr.mdSidenavFocus;
```
crisbeto: You've got `attr.mdAutofocus` twice here.

src/core/util/autofocus.js
```diff
@@ -90,18 +90,37 @@ angular.module('material.core')
...
+    /**
+     * Updates the autofocus class which is used to determine whether the attribute
+     * expression evaluates to true or false.
+     * @param {string|boolean} value Attribute Value
+     */
+    function updateExpression(value) {
+
+      // Rather than passing null or undefined to the jqLite toggle class function we explicitly set the
+      // value to true. Otherwise the class will be manually toggled and can cause issues.
+      if (value == null) {
```
crisbeto: Could avoid the double equals? I'm currently trying to factor them out as a part of the JSHint fixes.

src/core/util/autofocus.js
DevVersion: This is from before, and is **required** because we currently support `md-autofocus` and `md-auto-focus`.

src/core/util/autofocus.js
DevVersion: Double equal is used here to also match `undefined`.

```js
undefined === null // => false
```

src/core/util/autofocus.js
crisbeto: Yes, but this is exactly the same. The mapping would be `md-autofocus` -> `mdAutofocus` and `md-auto-focus` -> `mdAutoFocus`.

DevVersion: Yeah, that's what we have right now?

- `attr.mdAutoFocus`
- `attr.mdAutofocus`

src/core/util/autofocus.js
```diff
@@ -90,18 +90,37 @@ angular.module('material.core')
...
 
-function postLink(scope, element, attrs) {
-  var attr = attrs.mdAutoFocus || attrs.mdAutofocus || attrs.mdSidenavFocus;
+  function preLink(scope, element, attr) {
+    var attrExp = attr.mdAutoFocus || attr.mdAutofocus || attr.mdSidenavFocus;
+
+    // Initially update the expression by manually parsing the expression as per $watch source.
+    updateExpression($parse(attrExp)(scope));
+
+    scope.$watch(attrExp, updateExpression);
```
crisbeto: Perhaps we can avoid the watch here if the attribute is empty?

DevVersion: Yeah I forgot about this.

PR #9637, opened 2016-09-18
update(dialog): remove .md-actions deprecation

### LINKED ISSUE #9633: dialog: deprecate .md-actions class
Just a simple reminder to make sure that the `.md-actions` class will be removed in the future.

Regardless of using `mdPanel` or still being at the old one.

> I would remove them right now, to get rid of unnecessary logic step by step.

cc. @ThomasBurleson

DevVersion:
* Removes the `.md-actions` deprecation from pre- v1.0.0
* Simplifies the method to search for a close button

Closes #9633

### DIFF
--- a/src/components/dialog/dialog.scss
+++ b/src/components/dialog/dialog.scss
@@ -80,7 +80,7 @@ md-dialog {
     }
   }
 
-  .md-actions, md-dialog-actions {
+  md-dialog-actions {
     display: flex;
     order: 2;
     box-sizing: border-box;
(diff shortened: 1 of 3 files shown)


src/components/dialog/dialog.scss
```diff
@@ -80,7 +80,7 @@ md-dialog {
     }
   }
 
-  .md-actions, md-dialog-actions {
```
ThomasBurleson: we should leave the style classes for deprecated support (while we removing the warnings).
Otherwise this is a breaking change.

DevVersion: We had them since pre v1.0.0 that's why I intentionally removed them for now. The target is to remove the deprecation completely (two major versions~)

PR #9567, opened 2016-09-10
fix(panel): element not being removed when scope is destroyed

crisbeto:
The panel element wasn't being removed when it's scope is destroyed (e.g. when navigating to a different page).

Fixes #8683.

PR #9641, opened 2016-09-19
feat(datepicker): migrate to mdPanel

### LINKED ISSUE #9564: mdPanel: Rewrite mdDatepicker to use the panel.
Tracking issue for using the mdPanel in datepicker. Any related issues that appear during the migration should refer to this issue as a primary use case.

crisbeto:
Migrates the datepicker's positioning logic to mdPanel. All the functionality should work as before.

Fixes #9564.

## Notes
This PR technically works and it is ready for code review, however in order to achieve feature parity with the current datepicker we need the following:

1. Merging in #9615. I've been developing this PR with the changes from #9615 already in place so I can confirm that it works.
2. Intelligent positioning that doesn't allow the calendar to go outside the viewport. I'll be working on this next (issue #7878). This should also include a class to be used to slightly re-style the calendar if it has been repositioned.
3. Currently the calendar has some functionality that makes it scale down if it's bigger than the screen. I've skipped this for now, but I suppose we need to decide whether we want it.


crisbeto:
Also I think the test failures are because the old panels aren't being destroyed when their scope is destroyed. That should be fixed with #9567. I'll check if it can be fixed in the tests here as well.

src/components/datepicker/js/datepickerDirective.js
```diff
@@ -291,12 +278,23 @@
...
 
     /** @type {boolean} */
-    this.isDisabled;
-    this.setDisabled($element[0].disabled || angular.isString($attrs.disabled));
+    this.isDisabled = false;
+    this.setDisabled($element[0].disabled || $attrs.hasOwnProperty('disabled'));
 
-    /** @type {boolean} Whether the date-picker's calendar pane is open. */
+    /** @type {boolean} Whether the date-picker's calendar pane is open. Used internally. */
     this.isCalendarOpen = false;
```
ErinCoughlan: Used internally, therefore should be marked private?

src/components/datepicker/js/datepickerDirective.js
```diff
@@ -291,12 +278,23 @@
...
+    /**
+     * Whether calendar should be open. Used when triggering the calendar externally.
+     * It needs to be separate from `isCalendarOpen` in order to avoid some mixups
+     * where it might trigger the calendar a couple of times in a row.
+     * @type {boolean}
+     */
+    this.isOpen = false;
+
+    /** @type {boolean} Used to prevent infinite loops when using mdOpenOnFocus. */
+    this.preventInputFocus = false;
```
ErinCoughlan: If you're using panels focusOnOpen, is this needed?

src/components/datepicker/js/datepickerDirective.js
```diff
@@ -338,6 +330,25 @@
...
+      .withOffsetY(angular.bind(this, this.getPanelYOffset));
+
+    /** @type {!MdPanelRef} Reference to the mdPanel instance of the calendar. */
+    this.panelRef = $mdPanel.create({
+      attachTo: document.body,
+      template: CALENDAR,
+      clickOutsideToClose: true,
+      escapeToClose: true,
+      focusOnOpen: false,
+      scope: this.$scope.$new(),
```
ErinCoughlan: I think this is the default, so you can remove it if you want. Being explicit is also fine though. Up to you.

src/components/datepicker/js/datepickerDirective.js
```diff
@@ -606,182 +622,105 @@
...
-    // on each end to make it overlap slightly. The overlay's background is added in
-    // the theme in the form of a box-shadow with a huge spread.
-    this.inputMask.css({
-      position: 'absolute',
-      left: this.leftMargin + 'px',
-      top: this.topMargin + 'px',
-      width: (elementRect.width - 1) + 'px',
-      height: (elementRect.height - 2) + 'px'
-    });
+    var self = this;
```
ErinCoughlan: So much code goes bye-bye. This is amazing!

src/components/datepicker/js/datepickerDirective.js
```diff
@@ -606,182 +622,105 @@
...
+      // on each end to make it overlap slightly. The overlay's background is added in
+      // the theme in the form of a box-shadow with a huge spread.
+      angular.element(inputMask).css({
+        position: 'absolute',
+        left: self.leftMargin + 'px',
 
-    calendarPane.style.left = paneLeft + 'px';
-    calendarPane.style.top = paneTop + 'px';
-    document.body.appendChild(calendarPane);
+        // TODO(crisbeto): this should use getPanelYOffset once we ditch the units
```
ErinCoughlan: Can this be done now?

src/components/datepicker/js/calendar.js
```diff
@@ -197,13 +200,18 @@
 
     var boundKeyHandler = angular.bind(this, this.handleKeyEvent);
 
-    // Bind the keydown handler to the body, in order to handle cases where the focused
-    // element gets removed from the DOM and stops propagating click events.
-    angular.element(document.body).on('keydown', boundKeyHandler);
+    if (this._panelRef) {
+      // Bind the keydown handler to the body, in order to handle cases where the focused
+      // element gets removed from the DOM and stops propagating key events.
+      angular.element(document.body).on('keydown', boundKeyHandler);
```
crisbeto: The calendar currently focuses the active date's cell (I believe for accessibility reasons), which causes any keyboard events to be captured by the calendar. The issue is that the calendar uses `mdVirtualRepeat` which will remove the cell eventually, if you scroll long enough. This causes the browser to shift focus back to the body which stops the keyboard events from working.

src/components/datepicker/js/datepickerDirective.js
crisbeto: Yes, this is also a private property. When the calendar closes, it restores focus to the element that triggered it to open. With `mdOpenOnFocus` that would cause the calendar to open up again since it refocuses the input. We use this flag to prevent such cases. It used to be possible to do this by delaying the close handler, but it got a little trickier with `mdPanel` since it's also async.

src/components/datepicker/js/datepickerDirective.js
crisbeto: The default creates a scope that's a child of `$rootScope`, however we need it to be a child of the datepicker's scope.

src/components/datepicker/js/datepickerDirective.js
crisbeto: It's waiting on #9609 to be merged.

src/components/datepicker/js/datepickerDirective.js
```diff
@@ -338,6 +330,25 @@
     /** @type {Number} Extra margin for the top of the floating calendar. Gets determined on the first open. */
     this.topMargin = null;
 
+    /** @type {!MdPanelPosition} Position object for the mdPanel instance. */
+    this.panelPosition = $mdPanel.newPanelPosition()
+      .relativeTo(this.inputContainer)
+      .addPanelPosition($mdPanel.xPosition.ALIGN_START, $mdPanel.yPosition.ALIGN_TOPS)
+      .withOffsetX(-this.leftMargin + 'px')
+      .withOffsetY(angular.bind(this, this.getPanelYOffset));
+
```
henrymana: The current version of the md-datepicker hides the popup when the window is re-sized. I noticed that the md-panel stays in the original position when the window is resized and using "relativeTo" (it stops being relativeTo the original element).

Can you confirm what the behavior is for the new implementation?

henrymana: Cool, I see it know... (window.addEventListener(this.windowEventName, this.windowEventHandler) ). Thank you.

crisbeto:
This is no longer blocked by #9651, however there are still some issues with the logic for constraining the calendar to the viewport. I'll leave a comment about the issues on #9564.

PR #9647, opened 2016-09-20
feat($mdDialog): configure prompt input type

### LINKED ISSUE #9646: $mdDialog.prompt() input type should be configurable
**Actual Behavior**:
 -  `What is the issue? *`

The `mdDialog.prompt` template has an `input` element which has no `type` attribute by default.

The `$mdDialog` service should offer the option to add a `type` to the `input` element.
The value can be restricted to a limited set of values (e.g. `text`, `email`, `tel`, `password`...).
Without this option, I have to create a custom dialog which adds a template and a controller file to my code base.

Here is the input I talk about:
https://github.com/angular/material/blob/master/src/components/dialog/dialog.js#L597

 -  `CodePen Demo which shows your issue:` http://codepen.io/thomaskerbrat/pen/ALpvYO
 -  `Details:` If you comment the `.inputType('password')` line, the CodePen just works as expected.

**Angular Versions**: *
 -  `Angular Version:` 1.5.8

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *` 53.0.2785.116 m
 -  `OS: *` Windows 7
 -  `Stack Traces:`

----

  Add more lines if needed.

ThomasKerbrat:
Enable the developer to choose the type of input for the prompt dialog.
Add $mdDialog().prompt().inputType() method to set the input type.
Remove the need to create a dedicated controller and template to specify the type of input.

Fix #9646

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -457,6 +457,7 @@ function MdDialogDirective($$rAF, $mdTheming, $mdDialog) {
  * - $mdDialogPreset#targetEvent(DOMClickEvent=) - A click's event object. When passed in as an option,
  *     the location of the click will be used as the starting point for the opening animation
  *     of the the dialog.
+ * - $mdDialogPreset#inputType(string) - Sets the type of the prompt input.
  *
  */
 
@@ -577,7 +578,7 @@ function MdDialogProvider($$interimElementProvider) {
     })
     .addPreset('prompt', {
       methods: ['title', 'htmlContent', 'textContent', 'initialValue', 'content', 'placeholder', 'ariaLabel',
-          'ok', 'cancel', 'theme', 'css'],
+          'ok', 'cancel', 'theme', 'css', 'inputType'],
       options: advancedDialogOptions
     });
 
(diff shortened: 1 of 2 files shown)


ThomasKerbrat:
I signed it!

DevVersion:
Thanks for the contribution! 

We recently had a similar discussion, which was about allowing different input types for the prompt preset.

> The conclusion was, that we don't want to support other input types here, because it won't match with the native `window.prompt` anymore, and we don't want to be responsible for all the other types and options available.

Developers should create their own preset or use a custom template instead.

I'm going to keep that PR open for now, because I'm up for some discussion :)

ThomasBurleson:
Closing. Will not merge; see notes above.

ThomasKerbrat:
@DevVersion  Thank you for the feedback.
I was unaware of the custom presets, but that won't help me here.

@ThomasBurleson I have an objection to make.
I understand that without this change, I have to create a custom dialog via the `$mdDialog.show({ ... })` interface. Thus, I have to implement a controller and a template.
The issue is that I don't have access to the properties on the `dialog` object in the template. Here is the source file I am referring to:
https://github.com/angular/material/blob/master/src/components/dialog/dialog.js#L588

That means the custom dialog won't be sensitive to the theme.
It is set with `md-theme="{{ dialog.theme }}"` in the link above.

So, how I am supposed to implement a dialog myself without reinventing the wheel?

PR #9650, opened 2016-09-20
fix(chips): Add basic accessibility support.

### LINKED ISSUE #8867: contact chips: Contact Chips demo no longer filters based on input.
**Actual Behavior**:
-  `What is the issue? *` When typing in the search field for the Contact Chips demo, the results are not filtered based on the selected text.
-  `What is the expected behavior?` The results should be filtered to only those that match the search.

-  `CodePen Demo which shows your issue:` https://material.angularjs.org/1.1.0-rc.5/demo/chips
-  `Details:`

**Angular Versions**: *
-  `Angular Version:` 1.5.6
-  `Angular Material Version:` 1.1.0-RC5

**Additional Information**:
-  `Browser Type: *` All
-  `Browser Version: *` All
-  `OS: *` All
-  `Stack Traces:`

---

>   Add more lines if needed.

### LINKED ISSUE #8897: Chips not accessible with a screen reader
**Actual Behavior**:
While chips work great with the keyboard alone, they are inaccessible with screen readers for a few reasons:
1. The chips control is hidden with `aria-hidden="true"`, so it essentially doesn't exist for a screen reader user. How will they interact with chips without some other affordance?
2. There is no way to focus on existing chips with the arrow keys when focused in an input field like you can with the keyboard. 
3. Pressing delete sends focus to the last chip, but nothing is announced; probably because the entire control has `aria-hidden="true"`. 
4. The demo input fields are not labeled.

**What is the expected behavior?**
Chips should be usable with a screen reader running. Users should be able to determine what a chip is, how many there are, and how to delete them without seeing the screen.

This component needs an accessible UX review to decide the appropriate roles and interactions for screen reader users. Off the top of my head I'm not sure what the right roles would be–ARIA menus and tabs aren't really appropriate, even if they might help with the arrow keys in a screen reader. You're focused on an input field the whole time, so it's not immediately obvious what the right pattern would be. Keep in mind the responsive nature of this framework means it should also work with a screen reader on a touch device, but I recognize even getting it to work on desktop will be a challenge.

**Details** (or steps to reproduce the issue): *
Visit the Chips demo on https://material.angularjs.org in Safari or Chrome with Voiceover running. On Windows, the issue is reproducible with NVDA and JAWS. 

Screenshot: Arrow keys in a screen reader do not select chips. Focus is in the input field, but virtual cursor is on h2 heading after pressing the left arrow key

Arrow keys in Voiceover do not select chips

**Angular Versions**: *
-  Angular Version:
-  Angular Material Version:
  All

**Additional Information**:
-  `Browser Type: *`
-  `Browser Version: *`
-  `OS: *`
-  `Stack Traces:`
  All

---

>   Add more lines if needed.

### LINKED ISSUE #9391: Chips have poor accessibility
It appears that the chips don't have any accessibility treatment other than a few `aria-hidden` and `aria-label` instances. Follow up with me for more details.

### LINKED ISSUE #9556: md-chips has aria-hidden='true' which makes it hidden from the screenreader
**Actual Behavior**:
-  aria-hidden='true' is applied to the md-chips directive, which causes the the content inside md-chips does not get announced by the screenreader.
-  Expect no aria-hidden='true' on the md-chips, so that the items inside it can be read by screenreader.

**Angular Versions**: *
-  `Angular Version:` 1.5.8
-  `Angular Material Version:` 1.5.8

**Additional Information**:
-  `Browser Type: *` Chrome

### LINKED ISSUE #9649: demos: Chips demo hits slow network resource.
**Actual Behavior**:
-  `What is the issue? *` When viewing the Chips Demos page, one of the demos (the contact chips demo) hits http://lorempixel.com to load fake user images. Sometimes, this request can take up to 10 seconds and is extremely irritating during development.
-  `What is the expected behavior?` We should use a faster service and/or cache the images somehow.

-  `CodePen Demo which shows your issue:` https://material.angularjs.org/1.1.1/demo/chips
-  `Details:` Open the inspector and you will see a multitude of long-pending image requests (they usually download quickly, but the connection waiting seems to take forever).

Additionally, there are are warnings because the images are loaded over HTTP instead of HTTPS.

**Angular Versions**: *
-  `Angular Version:` All

**Additional Information**:
-  `Browser Type: *` Chrome
-  `Browser Version: *` 53.0.2785.116
-  `OS: *` Mac & Windows
-  `Stack Traces:`

---

>   Add more lines if needed.

topherfangio:
Previously, chips had no, or very broken, support for accessibility.

Make many updates to fix broken functionality and add required features.
- Remove existing `aria-hidden` attributes.
- Dynamically apply required `role` and `aria-owns` attributes.
- Ensure `tabindex` is set properly on all elements. Particularly,
  you can now tab to, and navigate through, a set of chips in
  readonly mode.
- Provide new `input-aria-label` option to identify the inputs.
- Provide new `container-hint` option used when the element is in
  readonly mode.
- Provide new `md-chip-append-delay` option which controls the delay
  (in ms) before a user can add a new chip (fixes screen readers).
- Fix issue with `delete-hint` never being read by screen readers
  because it was outside of the chip.
- Fix keyboard navigation to properly wrap when moving both left
  or right.
- Fix keyboard navigation when in readonly mode.
- Fix issue where the wrong chip may be focused because the old
  chip was still in the DOM.
- Fix issue where `onAdd` callback passed incorrect `$index` (it
  was 1-based instead of 0-based).

Additionally update the Chips docs with new Accessibility info and
a few other minor changes to the docs/demos.
- Remove the "WORK IN PROGRESS" notice.
- Fix mispelling of "maximum" on first chips demo.
- Fix Contact Chips demo filtering.
- Fix Contact Chips demo images to use new/faster service.

BREAKING CHANGES

**MAJOR**

In order to make the chips fully accessible and work properly across
a wide variety of screen readers, we have added a 300ms delay after
appending a chip before we re-focus the input.

This is necessary because some screen readers change operational modes
when the enter key is pressed inside of an input and this causes the
screen reader to move into "navigation" mode and no longer apply
keystrokes to the input.

Additionally, this ensures that the newly added chip is properly
announced to the screen readers.

You **may** alter this value with the new `md-chip-append-delay`
attribute, however using a value less than `300` can cause issues
on JAWS and NVDA and will make your application inaccessible to those
users.

Note 1: This issue only seems to affect chips appended when using the
`enter` key. If you override the `md-separator-keys` and disable the
`enter` key (and enable something like `,` or `;`), you may be able
to reduce this delay to `0` and achieve past functionality.

Note 2: This issue does not appear to affect VoiceOver or ChromeVox,
so if you are only targeting those users, you may be able to reduce
the delay to `0`.

**MINOR**

We consider the following to be minor breaking changes since we never
expected these attributes/elements to be utilized by developers.

Nonetheless, we want to ensure that they are documented.
- The `role` of the `.md-chip-content` has been modified from `button`
  to `option` so that it works with the new `listbox` role of it's
  parent.
  
  If you rely on this role being `button`, you should update your code
  accordingly.
- The delete hint on removable chips now resides inside of the
  `div.md-chip-content` rather than the parent `md-chip` element where
  it could not be read by screen readers.
  
  If you interact with this element (in CSS or JS) please update your
  selectors/code to account for the new DOM hierarchy.

Fixes #9391. Fixes #9556. Fixes #8897. Fixes #8867. Fixes #9649.

### DIFF
--- a/src/components/chips/demoBasicUsage/index.html
+++ b/src/components/chips/demoBasicUsage/index.html
@@ -5,7 +5,7 @@
 
     <form name="fruitForm">
       <md-chips ng-model="ctrl.roFruitNames" name="fruitName" readonly="ctrl.readonly"
-                md-removable="ctrl.removable" md-max-chips="5">
+                md-removable="ctrl.removable" md-max-chips="5" placeholder="Enter a fruit...">
         <md-chip-template>
           <strong>{{$chip}}</strong>
           <em>(fruit)</em>
--- a/src/components/chips/js/chipsDirective.js
+++ b/src/components/chips/js/chipsDirective.js
@@ -26,6 +26,9 @@
    * is also placed as a sibling to the chip content (on which there are also click listeners) to
    * avoid a nested ng-click situation.
    *
+   * <!-- Note: We no longer want to include this in the site docs; but it should remain here for
+   * future developers and those looking at the documentation.
+   *
    * <h3> Pending Features </h3>
    * <ul style="padding-left:20px;">
    *
@@ -56,10 +59,7 @@
    *   </ul>
    * </ul>
    *
-   * <span style="font-size:.8em;text-align:center">
-   *   Warning: This component is a WORK IN PROGRESS. If you use it now,
-   *   it will probably break on you in the future.
-   * </span>
+   * //-->
    *
    * Sometimes developers want to limit the amount of possible chips.<br/>
    * You can specify the maximum amount of chips by using the following markup.
@@ -82,7 +82,21 @@
    *   </md-chips>
    * </hljs>
    *
-   * @param {string=|object=} ng-model A model to bind the list of items to
+   * ### Accessibility
+   *
+   * The `md-chips` component supports keyboard and screen reader users since Version 1.1.2. In
+   * order to achieve this, we modified the chips behavior to select newly appended chips for
+   * `300ms` before re-focusing the input and allowing the user to type.
+   *
+   * For most users, this delay is small enough that it will not be noticeable but allows certain
+   * screen readers to function properly (JAWS and NVDA in particular).
+   *
+   * We introduced a new `md-chip-append-delay` option to allow developers to better control this
+   * behavior.
+   *
+   * Please refer to the documentation of this option (below) for more information.
+   *
+   * @param {string=|object=} ng-model A model to which the list of items will be bound.
    * @param {string=} placeholder Placeholder text that will be forwarded to the input.
    * @param {string=} secondary-placeholder Placeholder text that will be forwarded to the input,
    *    displayed when there is at least one item in the list
@@ -112,11 +126,28 @@
    * @param {expression=} md-on-select An expression which will be called when a chip is selected.
    * @param {boolean} md-require-match If true, and the chips template contains an autocomplete,
    *    only allow selection of pre-defined chips (i.e. you cannot add new ones).
+   * @param {string=} input-aria-label A string read by screen readers to identify the input.
+   * @param {string=} container-hint A string read by screen readers informing users of how to
+   *    navigate the chips. Used in readonly mode.
    * @param {string=} delete-hint A string read by screen readers instructing users that pressing
    *    the delete key will remove the chip.
    * @param {string=} delete-button-label A label for the delete button. Also hidden and read by
    *    screen readers.
    * @param {expression=} md-separator-keys An array of key codes used to separate chips.
+   * @param {string=} md-chip-append-delay The number of milliseconds that the component will select
+   *    a newly appended chip before allowing a user to type into the input. This is **necessary**
+   *    for keyboard accessibility for screen readers. It defaults to 300ms and any number less than
+   *    300 can cause issues with screen readers (particularly JAWS and sometimes NVDA).
+   *
+   *    _Available since Version 1.1.2._
+   *
+   *    **Note:** You can safely set this to `0` in one of the following two instances:
+   *
+   *    1. You are targeting an iOS or Safari-only application (where users would use VoiceOver) or
+   *    only ChromeVox users.
+   *
+   *    2. If you have utilized the `md-separator-keys` to disable the `enter` keystroke in
+   *    favor of another one (such as `,` or `;`).
    *
    * @usage
    * <hljs lang="html">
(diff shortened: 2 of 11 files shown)


src/components/chips/js/chipsDirective.js
```diff
@@ -146,25 +149,33 @@
...
         <md-chip ng-repeat="$chip in $mdChipsCtrl.items"\
             index="{{$index}}"\
             ng-class="{\'md-focused\': $mdChipsCtrl.selectedChip == $index, \'md-readonly\': !$mdChipsCtrl.ngModelCtrl || $mdChipsCtrl.readonly}">\
           <div class="md-chip-content"\
-              tabindex="-1"\
-              aria-hidden="true"\
+              tabindex="{{$mdChipsCtrl.ariaTabIndex == $index ? 0 : -1}}"\
+              id="{{$mdChipsCtrl.contentIdFor($index)}}"\
+              role="option"\
+              aria-live="assertive"\
```
jelbourn: I don't know if `aria-live` is a good idea here. Generally you only want one live region for the whole app. 

@marcysutton do you have any thoughts on this?

src/components/chips/demoBasicUsage/index.html
```diff
@@ -5,7 +5,7 @@ <h2 class="md-title">Use a custom chip template.</h2>
 
     <form name="fruitForm">
       <md-chips ng-model="ctrl.roFruitNames" name="fruitName" readonly="ctrl.readonly"
-                md-removable="ctrl.removable" md-max-chips="5">
+                md-removable="ctrl.removable" md-max-chips="5" placeholder="Enter a fruit...">
         <md-chip-template>
           <strong>{{$chip}}</strong>
```
jelbourn: In JAWS w/ IE11, upon adding a new chip, the screen reader announces "Left brace left brace dollar chip right brace right brace". Is the template being added to the dom before being linked?

marcysutton:
IE11 and JAWS is a known issue that may be fixed by using ng-bind: https://github.com/angular/angular.js/issues/15079

It works in Chrome and what AT? You should try Safari and Voiceover. NVDA is typically used with Firefox. Mobile Safari and Voiceover would be good to check as well. For Firefox and JAWS, I'll have to take a closer look when I'm in front of my computer.

It is a good idea to limit use of aria-live to specific announcer elements to avoid conflicting messages. Too many similar messages will also get swallowed in JAWS, so we often use two for each politeness level. You can see this implemented in ngA11y. https://github.com/dequelabs/ngA11y/blob/master/src/nga11yannounce.js

ThomasBurleson:
@topherfangio  ping to finish.

topherfangio:
@jelbourn After discussion with team and @marcysutton, we have added a `300ms` delay after adding the chips to fix some JAWS/NVDA issues.

The latest updates include this change as well as all of the related documentation. Can you review this change and the new docs?

Also, I have tested in again JAWS but only have 40 minute mode so it was pretty basic. Can you please re-test and see if you have any additional issues?

I still want to add a few test and hopefully remove one TODO, but this is pretty close to final.

ThomasBurleson:
@jelbourn - ping for review.

jelbourn:
@topherfangio can you rebase?

PR #9651, opened 2016-09-20
update(panel): constrain panel to viewport boundries

### LINKED ISSUE #7878: mdPanel: Add intelligent positioning to the panel

crisbeto:
Prevents the panel from going outside the viewport by adjusting the position.
If developers want more control over how the panel gets repositioned, they can specify addition fallback positions via `addPanelPosition`.

Related to #9641.

Fixes #7878.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -2156,6 +2156,12 @@ MdPanelPosition.absPosition = {
   LEFT: 'left'
 };
 
+/**
+ * Margin between the edges of a panel and the viewport.
+ * @const {number}
+ */
+MdPanelPosition.viewportMargin = 8;
+
 
 /**
  * Sets absolute positioning for the panel.
@@ -2525,6 +2531,9 @@ MdPanelPosition.prototype._reduceTranslateValues =
  * @private
  */
 MdPanelPosition.prototype._setPanelPosition = function(panelEl) {
+  // Remove the class in case it has been added before.
+  panelEl.removeClass('_md-panel-position-adjusted');
+
   // Only calculate the position if necessary.
   if (this._absolute) {
     return;
@@ -2539,11 +2548,48 @@ MdPanelPosition.prototype._setPanelPosition = function(panelEl) {
     this._actualPosition = this._positions[i];
     this._calculatePanelPosition(panelEl, this._actualPosition);
     if (this._isOnscreen(panelEl)) {
-      break;
+      return;
+    }
+  }
+
+  // Class that can be used to re-style the panel if it was repositioned.
+  panelEl.addClass('_md-panel-position-adjusted');
+  this._constrainToViewport(panelEl);
+};
+
+
+/**
+ * Constrains a panel's position to the viewport.
+ * @param {!angular.JQLite} panelEl
+ * @private
+ */
+MdPanelPosition.prototype._constrainToViewport = function(panelEl) {
+  var margin = MdPanelPosition.viewportMargin;
+
+  if (this.getTop()) {
+    var top = parseInt(this.getTop());
+    var bottom = panelEl[0].offsetHeight + top;
+    var viewportHeight = this._$window.innerHeight;
+
+    if (top < margin) {
+      this._top = margin + 'px';
+    } else if (bottom > viewportHeight) {
+      this._top = top - (bottom - viewportHeight + margin) + 'px';
+    }
+  }
+
+  if (this.getLeft()) {
+    var left = parseInt(this.getLeft());
+    var right = panelEl[0].offsetWidth + left;
+    var viewportWidth = this._$window.innerWidth;
+
+    if (left < margin) {
+      this._left = margin + 'px';
+    } else if (right > viewportWidth) {
+      this._left = left - (right - viewportWidth + margin) + 'px';
     }
   }
 };
-
 
 /**
  * Switches between 'start' and 'end'.
--- a/src/components/panel/panel.spec.js
+++ b/src/components/panel/panel.spec.js
@@ -13,6 +13,8 @@ describe('$mdPanel', function() {
   var DEFAULT_CONFIG = { template: DEFAULT_TEMPLATE };
   var PANEL_ID_PREFIX = 'panel_';
   var SCROLL_MASK_CLASS = '.md-scroll-mask';
+  var ADJUSTED_CLASS = '_md-panel-position-adjusted';
+  var VIEWPORT_MARGIN = 8;
 
   /**
    * @param {!angular.$injector} $injector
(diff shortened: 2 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -743,6 +743,12 @@ function MdPanelService($rootElement, $rootScope, $injector, $window) {
    * @type {enum}
    */
   this.yPosition = MdPanelPosition.yPosition;
+
+  /**
+   * Margin between the edges of the panel and the viewport
+   * @type {number}
+   */
+  this.viewportMargin = MdPanelPosition.viewportMargin;
```
ErinCoughlan: I don't think this one needs to be re-exported via `$mdPanel`, since there are no values for people to access or change.

src/components/panel/panel.js
```diff
@@ -1767,6 +1773,12 @@ MdPanelPosition.absPosition = {
   LEFT: 'left'
 };
 
+/**
+ * Margin between the edges of a panel and the viewport.
+ * @type {number}
```
ErinCoughlan: Should this use `@const`?

src/components/panel/panel.js
```diff
@@ -1767,6 +1773,12 @@ MdPanelPosition.absPosition = {
...
+ */
+MdPanelPosition.viewportMargin = 5;
```
ErinCoughlan: 5 -> 8 (everything in Material design is based on 8s.)

https://material.google.com/layout/responsive-ui.html#responsive-ui-grid

src/components/panel/panel.js
crisbeto: The main reason was to expose it to the unit tests. It can also be a global variable there that can be easily changed if tests start failing because of the wrong margin.

crisbeto:
Included the feedback from Erin.

crisbeto:
@ThomasBurleson this should be ready for presubmit.

src/components/panel/panel.js
```diff
@@ -2539,13 +2548,50 @@ MdPanelPosition.prototype._setPanelPosition = function(panelEl) {
     this._actualPosition = this._positions[i];
     this._calculatePanelPosition(panelEl, this._actualPosition);
     if (this._isOnscreen(panelEl)) {
-      break;
+      return;
     }
   }
+
+  // Class that can be used to re-style the panel if it was repositioned.
+  panelEl.addClass('_md-panel-position-adjusted');
```
henrymana: I see the test case uses it... nevermind.

src/components/panel/panel.js
```diff
@@ -2539,13 +2548,50 @@ MdPanelPosition.prototype._setPanelPosition = function(panelEl) {
...
+    var viewportHeight = this._$window.innerHeight;
+
+    if (top < margin) {
+      this._top = margin + 'px';
+    } else if (bottom > viewportHeight) {
+      this._top = top - (bottom - viewportHeight + margin) + 'px';
+    }
+  }
+
+  if (this.getLeft()) {
```
crisbeto: Good question. IMO the `constrainToViewport` method is a last-ditch effort to keep the panel in the viewport and it may not be successful (e.g. the element is wider than the viewport). Before `mdPanel` resorts to it, it tries all of the available positions to see which one stays in the viewport. If you want a bit more control, you can add extra positions via `addPanelPosition` and it'll try those first. Here's an example:

``js
$mdPanel.newPanelPosition()
  .relativeTo('something')
  .addPanelPosition(this._mdPanel.xPosition.ALIGN_START, this._mdPanel.yPosition.BELOW)
  .addPanelPosition(this._mdPanel.xPosition.ALIGN_END, this._mdPanel.yPosition.BELOW)
  .addPanelPosition(this._mdPanel.xPosition.ALIGN_START, this._mdPanel.yPosition.ABOVE);
```

This will try to position it start/below, end/below, start/above before resorting to constraining it.

Also I think you asked yesterday about the "position adjusted" class. This is useful in the cases where the panel was constrained, because it allows you to do some specific styling (e.g. we use it in the datepicker to hide a part of the overlay).

Comment (henrymana) [inline src/components/panel/panel.js]: I do use multiple positions. The panel does not really go off the screen, but it touches the right edge of the viewport with no margin (so I set a negative offsetX). Thinking about it more, that could probably be handled by css instead. Thank you for this change. I am already trying out your changes.

**Comment by henrymana** [comment]
I noticed earlier that the md-panel position is not updated when disableParentScroll is set to true (and as a result the panel falls outside the viewport boundaries). In my case the md-panel is at the bottom of the screen and this results in part of the panel being offscreen. 

The panel is positioned correctly within the viewport when disableParentScroll is set to false. 

Looking at the code, I can see this was a pre-existing behavior. I haven't finished debugging but the issue may have to do with the positionUpdate performed when the onScroll event handler is added in line 1783 ( this._$window.addEventListener('scroll', onScroll, true) ).

Should re-positioning work when the parent scroll is disabled? Should I open a separate issue to handle this?

**Comment by crisbeto** [comment]
I'm not sure about it, but it would be great if you made an issue so it's easier to track and discuss. It does sound strange though.

**Comment by henrymana** [comment]
Ok, I will create a small test to replicate and create a new issue later in the week. Thank you.

File: src/components/panel/panel.js [hunk 6]
```diff
@@ -2539,13 +2548,50 @@ MdPanelPosition.prototype._setPanelPosition = function(panelEl) {
...
 
 /**
+ * Constrains a panel's position to the viewport.
+ * @param {!angular.JQLite} panelEl
+ * @private
+ */
+MdPanelPosition.prototype._constrainToViewport = function(panelEl) {
+  var margin = MdPanelPosition.viewportMargin;
+
+  if (this.getTop()) {
```
henrymana: The md-datepicker code has a comment regarding "disabled body scrolling" which I think is applicable here as well. I need to understand the code more, but I think this may be a gap in the constrainToViewport implementation.

Comment from: https://github.com/angular/material/pull/9641/files#diff-31bf126782bc06608d149fe72237bf4eL640
```
If ng-material has disabled body scrolling (for example, if a dialog is open),
then it's possible that the already-scrolled body has a negative top/left. In this case,
we want to treat the "real" top as (0 - bodyRect.top). In a normal scrolling situation,
though, the top of the viewport should just be the body's scroll position.
```

crisbeto: I doubt it, because the panel is always positioned relatively to the viewport (`position: fixed`). The current datepicker implementation positions the calendar relatively to the body (via `position: absolute`).

henrymana:
I am no longer able to replicate the issue I mentioned earlier with disableParentScroll (I had a fresh checkout of the master branch and your branch only.). I am not sure why it failed earlier but it seems to be working now. Thank you again.

PR #9657, opened 2016-09-21
Autocomplete with icon

matteo-pando:
When a floating label is defined I added the possibility to include also a left icon, defining a md-icon attribute.
For now only svg icon supported


DevVersion:
Thanks for the pull request! 

Unfortunately this is not the approach we want to go, because we would rather 'd like to have a template for the icon instead.

Also we need to have a discussion with the team first, before we are going to support such enhancement.

matteo-pando:
Ok, understood.
Maybe I can open an issue to request this feature.
Thank you.

PR #9663, opened 2016-09-21
fix(build): fix errors on angular 1.3

crisbeto:
* Fixes a test in the autocomplete that was failing, because it wasn't triggering a digest.
* Adds a shim for `$q.resolve`, which was first introduced in Angular 1.4.1, in order to prevent similar errors in the future.


ThomasBurleson:
lgtm

PR #9695, opened 2016-09-26
fix(select): Fix duplicates in label.

### LINKED ISSUE #9442: md-select: selected item label is shown twice when md-option list changes
**Actual Behavior**:
 -  `What is the issue? *` If an item is selected and the option list (`<md-option ng-repeat="...">`) changes the select box contains the label of the selected item twice (as if it was selected multiple times): "value, value".  
When later on a digest cycle runs for the scope of the md-select (e.g. by opening the select box again) the label is updated correctly.
 -  `What is the expected behavior?` The selected value should only be shown once as label.

 -  `CodePen Demo which shows your issue:` http://codepen.io/anon/pen/WxqkyA
 -  `Details:` Click the "Update options" button to see the issue

While checking if it's related to changes in our application code (which it is not) I already found out that the `syncLabelText()` method is called multiple times. The first time is okay, but the next calls create the wrong label because the `<md-option>` elements are contained twice in the `<md-select-menu>` and both elements with the current value are marked as "selected". Looks like there's a mix of old and new `<md-option>` elements in the DOM.

**Angular Versions**: *
 -  `Angular Version:` 1.4.12, 1.5.5
 -  `Angular Material Version:` 1.1.0 stable (reproducable since 1.1.0-rc2)

**Additional Information**:
 -  `Browser Type: *` any
 -  `Browser Version: *` any
 -  `OS: *` any
 -  `Stack Traces:`

----

  Add more lines if needed.

topherfangio:
Occasionally, `ng-repeat` appears to modify the DOM in such a way that it allows duplicates of selected items. When this happens, the `md-select` has issues displaying the proper label.

Add a quick filter to ensure there are no duplicate selections in our labels.

NOTE: I was unable to reproduce this in a spec; it appears to be related to the scope digests and never happens in the spec. I did update one of the demos to do some manual testing and ensure this fix works as expected, but it didn't have any real use inside of the demo so I removed it before submission.

Fixes #9442.

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -833,6 +833,25 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
 
         return start + change * (-2 * tc + 3 * ts);
       }
+    },
+
+    /**
+     * Provides an easy mechanism for removing duplicates from an array.
+     *
+     *    var myArray = [1, 2, 2, 3, 3, 3, 4, 4, 4, 4];
+     *
+     *    $mdUtil.uniq(myArray) => [1, 2, 3, 4]
+     *
+     * @param {array} array The array whose unique values should be returned.
+     *
+     * @returns {array} A copy of the array containing only unique values.
+     */
+    uniq: function(array) {
+      if (!array) { return; }
+
+      return array.filter(function(value, index, self) {
+        return self.indexOf(value) === index;
+      });
     }
   };
 
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -174,7 +174,7 @@ angular.module('material.components.select', [
  * </div>
  * </hljs>
  */
-function SelectDirective($mdSelect, $mdUtil, $mdConstant, $mdTheming, $mdAria, $compile, $parse) {
+function SelectDirective($mdSelect, $mdUtil, $mdConstant, $mdTheming, $mdAria, $parse) {
   var keyCodes = $mdConstant.KEY_CODE;
   var NAVIGATION_KEYS = [keyCodes.SPACE, keyCodes.ENTER, keyCodes.UP_ARROW, keyCodes.DOWN_ARROW];
 
@@ -408,7 +408,7 @@ function SelectDirective($mdSelect, $mdUtil, $mdConstant, $mdTheming, $mdAria, $
       }
 
       scope.$watch(function() {
-          return selectMenuCtrl.selectedLabels();
+        return selectMenuCtrl.selectedLabels();
       }, syncLabelText);
 
       function syncLabelText() {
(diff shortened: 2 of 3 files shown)


src/core/util/util.js
```diff
@@ -799,6 +799,23 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
+     *    myArray.filter($mdUtil.uniq) => [1, 2, 3, 4]
+     *
+     * @param value
+     * @param index
+     * @param self
+     *
+     * @returns {boolean}
+     */
+    uniq: function(value, index, self) {
+      return self.indexOf(value) === index;
```
crisbeto: This is a pretty simple check and I don't think it makes sense to have it as a util, considering that it's only used it one place as well.

topherfangio: It is basic, but I've wanted it a few times and used some other approach. Although it's currently only in one place, I think there are lots of other places where we could leverage it.

crisbeto: Wouldn't it make sense to have the filtering inside of `uniq` in that case? That's how lodash's implementation works.

topherfangio: @crisbeto You mean something like `$mdUtil.uniq(myArray)` and have it call the `filter()` function internally? I guess so. I'll update the code.

crisbeto: Exactly. I'm not sure that just having the predicate function makes sense in a service.

ErinCoughlan:
@topherfangio What's the status on this PR? We have some Googlers also reporting this issue.

src/core/util/util.js
```diff
@@ -832,6 +832,23 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
+    },
+
+    /**
+     * When used with the Array.filter() method; provides an easy mechanism for removing duplicates
+     * from an array.
+     *
+     *    var myArray = [1, 2, 2, 3, 3, 3, 4, 4, 4, 4];
+     *    myArray.filter($mdUtil.uniq) => [1, 2, 3, 4]
+     *
+     * @param value
```
ErinCoughlan: Can you add types to these params?

src/core/util/util.js
ErinCoughlan: I think `uniq` should return the filtered list without any duplicates. Otherwise, it seems like a lot of extra overhead for using the util.

topherfangio:
@ErinCoughlan Fixed `mdUtil.uniq()` and added param types as requested.

This is a bit of a band-aid which removes duplicate labels caused by an issue with `ng-repeat` allowing duplicate elements in the DOM. I tried a few other fixes, but I think a more proper fix will require major rewrites of the select component.

topherfangio:
Testing can be done by checking out this PR locally and viewing http://codepen.io/topherfangio/pen/yaKWAB?editors=1010 which pulls Material from localhost.

PR #9697, opened 2016-09-26
fix(panel): invoke offset callbacks later

crisbeto:
Invokes the panel's offset functions slightly later, in order to allow for the _actualPosition to be assigned.
This lets the user make better decisions regarding the offset.

### DIFF
--- a/src/components/panel/panel.spec.js
+++ b/src/components/panel/panel.spec.js
@@ -1880,6 +1880,60 @@ describe('$mdPanel', function() {
         });
       });
 
+      it('should have assigned the actual position by the time the offset' +
+        'methods have been called', function() {
+          var positionSnapshot = null;
+          var obj = {
+            getOffsetX: function(mdPanelPosition) {
+              positionSnapshot = angular.copy(mdPanelPosition.getActualPosition());
+            }
+          };
+          var position = mdPanelPosition
+              .relativeTo(myButton[0])
+              .addPanelPosition(xPosition.ALIGN_START, yPosition.ALIGN_TOPS)
+              .withOffsetX(obj.getOffsetX);
+
+          config['position'] = position;
+
+          openPanel(config);
+
+          expect(positionSnapshot).toEqual({
+            x: xPosition.ALIGN_START,
+            y: yPosition.ALIGN_TOPS
+          });
+      });
+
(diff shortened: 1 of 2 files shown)


src/components/panel/panel.spec.js
```diff
@@ -1880,6 +1880,31 @@ describe('$mdPanel', function() {
...
+              return 0;
+            }
+          };
+
+          spyOn(obj, 'getOffsetX').and.callThrough();
+
+          var position = mdPanelPosition
+              .relativeTo(myButton[0])
+              .addPanelPosition(xPosition.ALIGN_START, yPosition.ALIGN_TOPS)
+              .withOffsetY(obj.getOffsetX);
```
ErinCoughlan: Please keep the X and Y the same here.

src/components/panel/panel.spec.js
```diff
@@ -1880,6 +1880,31 @@ describe('$mdPanel', function() {
         });
       });
 
+      it('should have assigned the actual position by the time the offset' +
+        'methods have been called', function() {
+          var hasActualPosition = false;
+
+          var obj = {
+            getOffsetX: function(position) {
```
ErinCoughlan: Consider making this entire method a spy and then doing something like this:

```
var getOffsetX = jasmine.createSpy('position');
var position = mdPanelPosition
    .relativeTo(myButton[0])
    .addPanelPosition(xPosition.ALIGN_START, yPosition.ALIGN_TOPS)
    .withOffsetX(getOffsetX);

config['position'] = position;

openPanel(config);

var positionObj = jasmine.objectContaining({
  _actualPosition: {
    x: xPosition.ALIGN_START,
    y: yPosition.ALIGN_TOPS
  }
};

expect(getOffsetY).toHaveBeenCalledWith(positionObj);
```

An even better way is to write this test using `getActualPosition` since that's actually the method we can access. Typescript makes private stuff really private, so you can't even write tests with them.

crisbeto: I initially tried something similar, but the problem with this approach is that it's testing against the final object, which means that the test will pass even without the current changes.

crisbeto:
I'll close the PR for now since this approach isn't correct. This will break the functionality where we take the offsets into account when checking whether a panel is on screen.

PR #9698, opened 2016-09-27
fix(autocomplete): apply theming to input element

### LINKED ISSUE #9447: autocomplete: input element should use the same theme
**Actual Behavior**:
-  `What is the issue? *` Classes like md-accent (for floating label autocompletes) are not passed along to the input container.  There is a similar feature for passing classes to the menu container: `md-menu-class`.  This would allow the use theming engine with the existing inputs (as well as additional styles for the input.)  For me, the big draw here is the theming support. 
-  `What is the expected behavior?` Add an option like `md-input-class` that allows us to forward classes to the md-input-container.

-  `CodePen Demo which shows your issue:` http://codepen.io/mckenzielong/pen/vKorwJ
-  `Details:` Note: mdAccent isn't applied, and the custom (and ugly) class isn't used on the mdAutocomplete input (as expected)

**Angular Versions**: *
-  `Angular Version:` 1.5.8

**Additional Information**:
-  `Browser Type: *` N/A
-  `Browser Version: *` N/A
-  `OS: *` N/A
-  `Stack Traces:` N/A

Ping @DevVersion (as per #3623)

---

>   Add more lines if needed.

mckenzielong:
Fixes #9447

### DIFF
--- a/src/components/autocomplete/autocomplete-theme.scss
+++ b/src/components/autocomplete/autocomplete-theme.scss
@@ -1,3 +1,24 @@
+@mixin md-autocomplete-input($input-color) {
+  md-input-container.md-input-focused {
+    .md-input {
+      border-color: $input-color;
+    }
+    label,
+    md-icon {
+      color: $input-color;
+    }
+  }
+}
+@mixin md-autocomplete-progress($container-color, $bar-color) {
+  md-progress-linear {
+    .md-container {
+      background-color: $container-color;
+    }
+    .md-bar {
+      background-color: $bar-color;
+    }
+  }
+}
 md-autocomplete.md-THEME_NAME-theme {
   background: '{{background-hue-1}}';
   &[disabled]:not([md-floating-label]) {
@@ -15,6 +36,14 @@ md-autocomplete.md-THEME_NAME-theme {
   }
   input {
     color: '{{foreground-1}}';
+  }
+  &.md-accent {
+    @include md-autocomplete-input('{{accent-color}}');
+    @include md-autocomplete-progress('{{accent-100}}', '{{accent-color}}');
+  }
+  &.md-warn {
+    @include md-autocomplete-input('{{warn-A700}}');
+    @include md-autocomplete-progress('{{warn-100}}', '{{warn-color}}');
   }
 }
 .md-autocomplete-suggestions-container.md-THEME_NAME-theme {


mckenzielong:
@DevVersion similar to what we discussed in #9447 as option 1, but this isn't a full solution... I realized that the loading bar wouldn't be covered by this case (for example docs + accent yields proper coloured input, but incorrect styled loading bar)

We could filter the classes (md-warn and md-accent) as you mentioned and forward those to the input container and loading bar -- a similar solution could be applicable to md-input-container + md-select.

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -218,6 +218,7 @@ function MdAutocomplete ($$mdSvgRegistry) {
         // Retrieve the state of using a md-not-found template by using our attribute, which will
         // be added to the element in the template function.
         ctrl.hasNotFound = !!element.attr('md-has-not-found');
+        $mdTheming(element);
```
DevVersion: Why do we need to do this again? The controller already does register the theme.

> https://github.com/angular/material/blob/master/src/components/autocomplete/js/autocompleteController.js#L70

src/components/autocomplete/autocomplete-theme.scss
```diff
@@ -13,6 +13,44 @@ md-autocomplete.md-THEME_NAME-theme {
...
+    }
+    md-progress-linear {
+      .md-container {
+        background-color: '{{warn-100}}';
+      }
+      .md-bar {
+        background-color: '{{warn-color}}';
+      }
+    }
+  }
```
DevVersion: If we are doing something like that, then we should create a mixin for it, so we don't have to repeat it three times.

mckenzielong: @DevVersion not sure if you mean in autocomplete-theme.scss or if you mean on in input and linear progress that could be included?

Was hoping that the mixins used in md-tabs would have been a good starting point -- upon closer inspection they just repeated the same code in mixins: https://github.com/angular/material/blob/317c1c8d95bca192397336fc0f65dab402c23680/src/components/tabs/tabs-theme.scss#L1  <- were you looking for something more like that?

DevVersion: Yeah I meant something like that :+1: This just removes the unnecessary repetitions and it's easier to maintain.

DevVersion:
I just pulled the PR and tried it out. Seems like the progress bar did not get the correct color.

mckenzielong:
@DevVersion Sorry, my bad.  Fixed.

DevVersion:
I tried it out and it works 👍 Still not very a fan of that solution, but it might be easiest for now.

Splaktar:
@mckenzielong I'm sorry that this got stuck for so long. If you can rebase it, I'd be happy to try to get this merged in 1.1.11. 

If you aren't able to rebase it, I can try to take over this work and get it updated merged in.

mckenzielong:
@Splaktar i'll take a look 👍

PR #9721, opened 2016-09-29
fix(panel): use prefixed transform property

crisbeto:
Fixes a JS error in the panel, if the user's browser uses the prefixed `transform` property.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -2087,6 +2087,9 @@ function MdPanelPosition($injector) {
   /** @private {boolean} */
   this._isRTL = $injector.get('$mdUtil').bidi() === 'rtl';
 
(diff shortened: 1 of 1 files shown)


src/components/panel/panel.js
```diff
@@ -2087,6 +2087,9 @@ function MdPanelPosition($injector) {
   /** @private {boolean} */
   this._isRTL = $injector.get('$mdUtil').bidi() === 'rtl';
 
+  /** @private @const {!angular.$mdConstant} */
```
DevVersion: Why is `$mdConstant` in the angular namespace?

crisbeto: I'm not sure. I found it being used this way in [another place in mdPanel](https://github.com/angular/material/blob/master/src/components/panel/panel.js#L1051).

PR #9731, opened 2016-09-30
﻿fix(autocomplete): reduce unecessary re-positioning calls

DevVersion:
- The autocomplete right now always re-positions the virtual repeat containers while the autocomplete is hidden.
- This is unecessary because the autocomplete will automatically position itself upon open.

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -21,7 +21,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       hasFocus             = false,
       fetchesInProgress    = 0,
       enableWrapScroll     = null,
-      inputModelCtrl       = null;
+      inputModelCtrl       = null,
+      debouncedOnResize    = $mdUtil.debounce(onWindowResize);
 
   // Public Exported Variables with handlers
   defineProperty('hidden', handleHiddenChange, true);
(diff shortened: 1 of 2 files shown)


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -203,7 +207,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       $mdUtil.enableScrolling();
     }
 
-    angular.element($window).off('resize', positionDropdown);
+    angular.element($window).off('resize', onWindowResize);
```
crisbeto: I think it's generally a good idea to debounce the resize handlers since they can get called very often.

DevVersion: It could be, but this can also damage the smoothness (we have right) now.

If so, then we should also do this change in the tabs
> https://github.com/angular/material/blob/f7d6d1007f608005c83fc22a4f59fd4868aca2dd/src/components/tabs/js/tabsController.js#L109

crisbeto: I don't know about the tabs, but we do it in most other components (backdrop, datepicker, dialog, select, slider, tooltip, virtual repeat).

DevVersion: Okay, so I now made the changes, because I did not see any regressions with the smoothness. 
> Please review again

crisbeto: You could also inline the `onWindowResize` function since it isn't being used anywhere else.

DevVersion: We need to have the same fn-Object to be able to de-register the listener properly. 

> Also having as a normal function is more clean...

crisbeto: What I mean is doing this:

```js
debouncedOnResize = $mdUtil.debounce(function() {
  if (!ctrl.hidden) {
   positionDropdown();
  }
});
```

This still lets you de-register it properly. As for being cleaner, it's probably down to preference.

DevVersion: Yeah I know what you mean, but what I said, having as an extra function is more clean (having it as an variable at the top is ugly).

crisbeto:
LGTM

PR #9733, opened 2016-10-01
fix(datepicker): add aria-owns and aria-expanded support

### LINKED ISSUE #9727: md-datepicker is missing the aria-owns attribute for its md-calendar.
**Actual Behavior**:
 -  `What is the issue? *`  md-datepicker is missing the aria-owns attribute for its md-calendar
-  `What is the expected behavior?`  md-datepicker should have an aria-owns attribute with the id of  its md-calendar

 -  `CodePen Demo which shows your issue:`  Inspect https://material.angularjs.org/latest/demo/datepicker you will notice that the md-datepicker elements do not have an aria-owns attribute.
 -  `Details:`

**Angular Versions**: *
 -  `Angular Version:` 1.5.8
 -  `Angular Material Version: ` 1.1.1

**Additional Information**:
 -  `Browser Type: *` Any.
 -  `Browser Version: *` Any
 -  `OS: *` Any
 -  `Stack Traces:` N/A

----

  Add more lines if needed.

crisbeto:
* Adds the `aria-owns` attribute to the datepicker, linking it to the calendar pane. This is necessary since the calendar isn't a child of the datepicker in the DOM.
* Adds the `aria-expanded` attribute, indicating when the calendar is open.
* Moves the accessibility tests to their own separate suite.

Fixes #9727.

### DIFF
--- a/src/components/datepicker/js/datepickerDirective.js
+++ b/src/components/datepicker/js/datepickerDirective.js
@@ -89,13 +89,15 @@
             (ariaLabelValue ? 'aria-label="' + ariaLabelValue + '" ' : '') +
             'class="md-datepicker-input" ' +
             'aria-haspopup="true" ' +
+            'aria-expanded="{{ctrl.isCalendarOpen}}" ' +
+            'aria-owns="{{::ctrl.calendarPaneId}}"' +
             'ng-focus="ctrl.setFocused(true)" ' +
             'ng-blur="ctrl.setFocused(false)"> ' +
             triangleButton +
         '</div>' +
 
         // This pane will be detached from here and re-attached to the document body.
-        '<div class="md-datepicker-calendar-pane md-whiteframe-z1">' +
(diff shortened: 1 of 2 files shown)


src/components/datepicker/js/datepickerDirective.js
```diff
@@ -89,13 +89,15 @@
             (ariaLabelValue ? 'aria-label="' + ariaLabelValue + '" ' : '') +
             'class="md-datepicker-input" ' +
             'aria-haspopup="true" ' +
+            'aria-expanded="{{!!ctrl.isCalendarOpen}}" ' +
```
DevVersion: `isCalendarOpen` should be already a boolean, and also the expression automatically checks for truthiness.

> https://github.com/angular/material/blob/a73ab43c437ec7468d51da54f159b9f2845ec07b/src/components/datepicker/js/datepickerDirective.js#L298

crisbeto: Good point. Updated.

PR #9734, opened 2016-10-01
﻿fix(autocomplete): dropdown shrink restore

### LINKED ISSUE #9726: md-chips + md-autocomplete: dropdown not always good positioned
**Actual Behavior**:
-  `What is the issue?` In the current version of angular material, I see some issues with the positioning of the dropdown for the autocomplete suggestions when this is inside of a md-chips. Sometimes the dropdown is not visible because its height and position are not properly calculated.
-  `What is the expected behavior?` The dropdown positioning should be calculated properly and shown when there are suggestions.

  WRONG (master): http://codepen.io/Borland/pen/kkogpO
  CORRECT (1.1.0): http://codepen.io/Borland/pen/BLdVzN
-  `Details:` The issue is not easy to reproduce, but in our project, after updating angular material from 1.1.0 to the latest version in master it started to happen.
  We're having an md-chips with an autocomplete. The suggested items of the autocomplete depends on the model of the chips (so that we don't show the already included chips). We're using md-min-length="0" to show the suggestions right away and md-no-cache="true" to be able to show different suggestions after each model change. Then, sometimes (usually when the md-chips is close to the bottom, i don't know if that would be a clue) the dropdown is not visible (because it has height: 0, and a wrong bottom property, again I'm not sure that if will help you) although it has still suggestions to show.

Here is the codepen showing the wrong behavior: http://codepen.io/Borland/pen/kkogpO
You just need to focus the md-chips, click on any of the suggested vegetables and then you'll see that no dropdown is visible anymore.
Instead, this is the behavior in version 1.1.0:  http://codepen.io/Borland/pen/BLdVzN

Wrong:
screen shot 2016-09-30 at 06 29 29

Correct:
screen shot 2016-09-30 at 06 29 11

**Angular Versions**: *
-  `Angular Version:` 1.5+

**Additional Information**:
-  `Browser Type: *` Chrome
-  `OS: *` MacOS X
-  `Stack Traces:`

DevVersion:
The `VirtualRepeat` determines the original size (when shrank) of the dropdown from the computed style.

Commit b114302717341638bb627792c197ec3c43e069ff removed the fixed height from the SCSS, because of the dynamic dropdown height.

Setting the height to `100%` allows the `VirtualRepeat` to properly determine the dimension of the dropdown.

Fixes #9726.

### DIFF
--- a/src/components/autocomplete/autocomplete.scss
+++ b/src/components/autocomplete/autocomplete.scss
@@ -184,6 +184,10 @@ md-autocomplete {
   position: absolute;
   box-shadow: 0 2px 5px rgba(black, 0.25);
   z-index: $z-index-tooltip;
+
+  // Expand the virtualRepeatContainer as much as the max-height from the JavaScript allows.
+  // This is necessary for the virtualRepeatContainer to be able to grow back.
+  height: 100%;
 }
 
 .md-virtual-repeat-container.md-not-found {
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -21,8 +21,8 @@ describe('<md-autocomplete>', function() {
   function createScope(items, scopeData, matchLowercase) {
 
     items = items || ['foo', 'bar', 'baz'].map(function(item) {
-      return { display: item };
-    });
+        return { display: item };
+      });
 
     inject(function($rootScope, $timeout) {
       scope = $rootScope.$new();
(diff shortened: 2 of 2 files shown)


DevVersion:
@ThomasBurleson Ping for review, because this is kind of an important issue.

ThomasBurleson:
@DevVersion - do we have a real-world issue that demonstrates this fix as valid ?

DevVersion:
@ThomasBurleson Yeah this shows the issue (http://codepen.io/Borland/pen/kkogpO) and also if you remove the code and run the tests those will fail.

> I investigated a lot into the VirtualRepeat, and since the `scroller / sizer` is absolute, the VirtualRepeat is not able to determine the height (from the children) without an explicit height definition (as previously in https://github.com/angular/material/commit/b114302717341638bb627792c197ec3c43e069ff)

kara:
@DevVersion Can you rebase?

DevVersion:
@kara Done.

PR #9736, opened 2016-10-02
feat(datepicker): add the ability to restrict users to a calendar view

### LINKED ISSUE #9260: Don't want to show md-calendar-month on date picker
Is there an any option that user can only select year and month in datepicker, the date views should not shown to user. once the user select the year and month is close the calendar and it show the select date in MM/YYYY format   that also in some specific field not in complete application.

**Show only year and month view**
year-month-view

**Hide date view**
date-views

**regrading #9246**

I try parseDate and formatDate methods on the $mdDateLocaleProvider but it format the date when date is selected from the calendar, if date enter using keyboard in number format (08052016) it not get formatted as (08/05/2016)  and if make the change in mdDatelocalProvider than the date format change every where in application, just need to change the date format in some specific field only

crisbeto:
- Adds the ability to restrict users to a calendar view. For example, it's now possible to let users only select a month, instead of having to pick out a date within the month. This can be useful for cases like a credit card form, where the day of the month doesn't necessarily make sense.
- Fixes an issue where the year view wouldn't highlight the proper cell, if the model was changed externally and the date was different from the first day of the month.

Fixes #9260.

### DIFF
--- a/src/components/datepicker/calendar.scss
+++ b/src/components/datepicker/calendar.scss
@@ -127,7 +127,7 @@ md-calendar {
   font-weight: 500; // Roboto Medium
   @include rtl(padding, 0 0 0 $md-calendar-side-padding + $md-calendar-month-label-padding, rtl-value( 0 0 0 $md-calendar-side-padding + $md-calendar-month-label-padding));
 
-  md-calendar-month &:not(.md-calendar-month-label-disabled) {
+  &.md-calendar-label-clickable {
     cursor: pointer;
   }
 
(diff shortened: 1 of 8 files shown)


ThomasBurleson:
@jelbourn - can you review plz ?

nightcrawler91:
How come this is still without reviewing since October?

ThomasBurleson:
@EladBezalel - please test and review.

EladBezalel:
@crisbeto mind rebasing it?

PR #9737, opened 2016-10-02
﻿feat(): support multiple as a default preset method

DevVersion:
* Dialogs, Bottom-Sheets and Toasts have presets, which all depend on the `build` preset automatically.

* Due to the new `muliple` property in the `$$interimElement` factory, developers can explicitly pass the multiple property per JSON, but not as a preset method.

@ThomasBurleson: This just creates an alias for the given option.

### DIFF
--- a/src/core/services/interimElement/interimElement.js
+++ b/src/core/services/interimElement/interimElement.js
@@ -53,7 +53,7 @@ function InterimElementProvider() {
      * all interim elements will come with the 'build' preset
      */
     provider.addPreset('build', {
-      methods: ['controller', 'controllerAs', 'resolve',
+      methods: ['controller', 'controllerAs', 'resolve', 'multiple',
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@DevVersion - what is the issue that this resolves ?

DevVersion:
@ThomasBurleson This is a followup PR to the #9053, which just provides the "new available" option in the default preset.

ThomasBurleson:
@DevVersion - If we consider our planned changes to $mdCompile and use of the Component API & mdPanel, is this necessary ?

DevVersion:
@ThomasBurleson This does not interfere with anything of that. It just provides the alias of the `multiple` flag, which was not added in the original PR.

Right now you can do

```js
$mdDialog({
  multiple: true
}
```

But to be consistent with all "available" options for the $$interimElement, the preset should also have a method for this

```js
$mdDialog(
  $mdDialog
    .build()
    .multiple(true)
);
```

PR #9742, opened 2016-10-03
feat(tooltip): Tooltip use MdPanel API

### LINKED ISSUE #9563: mdPanel: Rewrite mdTooltip to use the panel.
Tracking issue for using the mdPanel in tooltip. Any related issues that appear during the migration should refer to this issue as a primary use case.

bradrich:
Tooltip has been refactored to use the MdPanel API to position, animate open and closed, and show the tooltip as a result of user interaction.

Breaking Change: Tooltips with not content will no longer show and result in an error.

Fixes #9563

### DIFF
--- a/src/components/tooltip/tooltip-theme.scss
+++ b/src/components/tooltip/tooltip-theme.scss
@@ -1,6 +1,6 @@
-md-tooltip.md-THEME_NAME-theme {
+.md-tooltip.md-THEME_NAME-theme {
   color: '{{background-700-contrast}}';
-  .md-content {
+  &.md-panel {
     background-color: '{{background-700}}';
   }
 }
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -3,151 +3,300 @@
  * @name material.components.tooltip
  */
 angular
-  .module('material.components.tooltip', [ 'material.core' ])
-  .directive('mdTooltip', MdTooltipDirective)
-  .service('$$mdTooltipRegistry', MdTooltipRegistry);
+    .module('material.components.tooltip', [
+      'material.core',
+      'material.components.panel'
+    ])
+    .directive('mdTooltip', MdTooltipDirective)
+    .service('$$mdTooltipRegistry', MdTooltipRegistry);
+
 
 /**
  * @ngdoc directive
  * @name mdTooltip
  * @module material.components.tooltip
  * @description
- * Tooltips are used to describe elements that are interactive and primarily graphical (not textual).
+ * Tooltips are used to describe elements that are interactive and primarily
+ * graphical (not textual).
  *
  * Place a `<md-tooltip>` as a child of the element it describes.
  *
- * A tooltip will activate when the user focuses, hovers over, or touches the parent.
+ * A tooltip will activate when the user hovers over, focuses, or touches the
+ * parent element.
  *
  * @usage
  * <hljs lang="html">
- * <md-button class="md-fab md-accent" aria-label="Play">
- *   <md-tooltip>
- *     Play Music
- *   </md-tooltip>
- *   <md-icon md-svg-src="img/icons/ic_play_arrow_24px.svg"></md-icon>
- * </md-button>
+ *   <md-button class="md-fab md-accent" aria-label="Play">
+ *     <md-tooltip>Play Music</md-tooltip>
+ *     <md-icon md-svg-src="img/icons/ic_play_arrow_24px.svg"></md-icon>
+ *   </md-button>
  * </hljs>
  *
- * @param {expression=} md-visible Boolean bound to whether the tooltip is currently visible.
- * @param {number=} md-delay How many milliseconds to wait to show the tooltip after the user focuses, hovers, or touches the
- * parent. Defaults to 0ms on non-touch devices and 75ms on touch.
- * @param {boolean=} md-autohide If present or provided with a boolean value, the tooltip will hide on mouse leave, regardless of focus
- * @param {string=} md-direction Which direction would you like the tooltip to go?  Supports left, right, top, and bottom.  Defaults to bottom.
+ * @param {number=} md-z-index The visual level that the tooltip will appear
+ *     in comparison with the rest of the elements of the application.
+ * @param {expression=} md-visible Boolean bound to whether the tooltip is
+ *     currently visible.
+ * @param {number=} md-delay How many milliseconds to wait to show the tooltip
+ *     after the user hovers over, focuses, or touches the parent element.
+ *     Defaults to 0ms on non-touch devices and 75ms on touch.
+ * @param {boolean=} md-autohide If present or provided with a boolean value,
+ *     the tooltip will hide on mouse leave, regardless of focus.
+ * @param {string=} md-direction The direction that the tooltip is shown,
+ *     relative to the parent element. Supports top, right, bottom, and left.
+ *     Defaults to bottom.
  */
-function MdTooltipDirective($timeout, $window, $$rAF, $document, $mdUtil, $mdTheming, $animate,
-  $interpolate, $mdConstant, $$mdTooltipRegistry) {
+function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
+    $mdUtil, $mdTheming, $mdPanel, $$mdTooltipRegistry) {
 
   var ENTER_EVENTS = 'focus touchstart mouseenter';
   var LEAVE_EVENTS = 'blur touchcancel mouseleave';
-  var SHOW_CLASS = 'md-show';
-  var TOOLTIP_SHOW_DELAY = 0;
-  var TOOLTIP_WINDOW_EDGE_SPACE = 8;
+  var TOOLTIP_DEFAULT_Z_INDEX = 100;
+  var TOOLTIP_DEFAULT_SHOW_DELAY = 0;
+  var TOOLTIP_DEFAULT_DIRECTION = 'bottom';
+  var TOOLTIP_DIRECTIONS = {
+    top: { x: $mdPanel.xPosition.CENTER, y: $mdPanel.yPosition.ABOVE },
+    right: { x: $mdPanel.xPosition.OFFSET_END, y: $mdPanel.yPosition.CENTER },
+    bottom: { x: $mdPanel.xPosition.CENTER, y: $mdPanel.yPosition.BELOW },
+    left: { x: $mdPanel.xPosition.OFFSET_START, y: $mdPanel.yPosition.CENTER }
+  };
 
   return {
     restrict: 'E',
-    transclude: true,
-    priority: $mdConstant.BEFORE_NG_ARIA,
-    template: '<div class="md-content _md" ng-transclude></div>',
+    priority: 210, // Before ngAria
     scope: {
-      delay: '=?mdDelay',
-      visible: '=?mdVisible',
-      autohide: '=?mdAutohide',
-      direction: '@?mdDirection'    // only expect raw or interpolated string value; not expression
+      mdZIndex: '=?mdZIndex',
+      mdDelay: '=?mdDelay',
+      mdVisible: '=?mdVisible',
+      mdAutohide: '=?mdAutohide',
+      mdDirection: '@?mdDirection' // Do not expect expressions.
     },
-    compile: function(tElement, tAttr) {
-      if (!tAttr.mdDirection) {
-        tAttr.$set('mdDirection', 'bottom');
-      }
-
-      return postLink;
-    }
+    link: linkFunc
   };
 
-  function postLink(scope, element, attr) {
+  function linkFunc(scope, element, attr) {
+    // Set constants.
+    var parent = $mdUtil.getParentWithPointerEvents(element);
+    var debouncedOnResize = $$rAF.throttle(updatePosition);
+    var mouseActive = false;
+    var origin, position, panelPosition, panelRef, autohide, showTimeout,
+        elementFocusedOnWindowBlur = null;
 
+    // Initialize the theming of the tooltip.
     $mdTheming(element);
 
-    var parent        = $mdUtil.getParentWithPointerEvents(element),
-        content       = angular.element(element[0].querySelector('.md-content')),
-        tooltipParent = angular.element(document.body),
-        showTimeout   = null,
-        debouncedOnResize = $$rAF.throttle(function () { updatePosition(); });
-
-    if ($animate.pin) $animate.pin(element, parent);
-
-    // Initialize element
-
+    // Set defaults
     setDefaults();
-    manipulateElement();
-    bindEvents();
 
-    // Default origin transform point is 'center top'
-    // positionTooltip() is always relative to center top
-    updateContentOrigin();
-
-    configureWatchers();
+    // Set parent aria-label.
     addAriaLabel();
 
+    // Remove the element from its current DOM position.
+    element.detach();
+    element.attr('role', 'tooltip');
 
-    function setDefaults () {
-      scope.delay = scope.delay || TOOLTIP_SHOW_DELAY;
-    }
+    updatePosition();
+    bindEvents();
+    configureWatchers();
 
-    function updateContentOrigin() {
-      var origin = 'center top';
-      switch (scope.direction) {
-        case 'left'  : origin =  'right center';  break;
-        case 'right' : origin =  'left center';   break;
-        case 'top'   : origin =  'center bottom'; break;
-        case 'bottom': origin =  'center top';    break;
+    function setDefaults() {
+      scope.mdZIndex = scope.mdZIndex || TOOLTIP_DEFAULT_Z_INDEX;
+      scope.mdDelay = scope.mdDelay || TOOLTIP_DEFAULT_SHOW_DELAY;
+      if (!TOOLTIP_DIRECTIONS[scope.mdDirection]) {
+        scope.mdDirection = TOOLTIP_DEFAULT_DIRECTION;
       }
-      content.css('transform-origin', origin);
     }
 
-    function onVisibleChanged (isVisible) {
-      if (isVisible) showTooltip();
-      else hideTooltip();
+    function addAriaLabel(override) {
+      if (override || !parent.attr('aria-label')) {
+        var rawText = override || element.text().trim();
+        var interpolatedText = $interpolate(rawText)(parent.scope());
+        parent.attr('aria-label', interpolatedText);
+      }
     }
 
-    function configureWatchers () {
-      if (element[0] && 'MutationObserver' in $window) {
+    function updatePosition() {
+      setDefaults();
+
+      // If the panel has already been created, remove the current origin
+      // class from the panel element.
+      if (panelRef && panelRef.panelEl) {
+        panelRef.panelEl.removeClass(origin);
+      }
+
+      // Set the panel element origin class based off of the current
+      // mdDirection.
+      origin = 'md-origin-' + scope.mdDirection;
+
+      // Create the position of the panel based off of the mdDirection.
+      position = TOOLTIP_DIRECTIONS[scope.mdDirection];
+
+      // Using the newly created position object, use the MdPanel
+      // panelPosition API to build the panel's position.
+      panelPosition = $mdPanel.newPanelPosition()
+          .relativeTo(parent)
+          .addPanelPosition(position.x, position.y);
+
+      // If the panel has already been created, add the new origin class to
+      // the panel element and update it's position with the panelPosition.
+      if (panelRef && panelRef.panelEl) {
+        panelRef.panelEl.addClass(origin);
+        panelRef.updatePosition(panelPosition);
+      }
+    }
+
+    function bindEvents() {
+      // Add a mutationObserver where there is support for it and the need
+      // for it in the form of viable host(parent[0]).
+      if (parent[0] && 'MutationObserver' in $window) {
+        // Use a mutationObserver to tackle #2602.
         var attributeObserver = new MutationObserver(function(mutations) {
-          mutations
-            .forEach(function (mutation) {
-              if (mutation.attributeName === 'md-visible') {
-                if (!scope.visibleWatcher)
-                  scope.visibleWatcher = scope.$watch('visible', onVisibleChanged );
-              }
-              if (mutation.attributeName === 'md-direction') {
-                updatePosition(scope.direction);
-              }
+          if (isDisabledMutation(mutations)) {
+            $mdUtil.nextTick(function() {
+              setVisible(false);
             });
+          }
         });
 
-        attributeObserver.observe(element[0], { attributes: true });
-
-        // build watcher only if mdVisible is being used
-        if (attr.hasOwnProperty('mdVisible')) {
-          scope.visibleWatcher = scope.$watch('visible', onVisibleChanged );
-        }
-      } else { // MutationObserver not supported
-        scope.visibleWatcher = scope.$watch('visible', onVisibleChanged );
-        scope.$watch('direction', updatePosition );
+        attributeObserver.observe(parent[0], {
+          attributes: true
+        });
       }
 
-      var onElementDestroy = function() {
-        scope.$destroy();
-      };
+      elementFocusedOnWindowBlur = false;
+
+      $$mdTooltipRegistry.register('scroll', windowScrollEventHandler, true);
+      $$mdTooltipRegistry.register('blur', windowBlurEventHandler);
+      $$mdTooltipRegistry.register('resize', debouncedOnResize);
+
+      scope.$on('$destroy', onDestroy);
+
+      // To avoid 'synthetic clicks', we listen to mousedown instead of
+      // 'click'.
+      parent.on('mousedown', mousedownEventHandler);
+      parent.on(ENTER_EVENTS, enterEventHandler);
+
+      function isDisabledMutation(mutations) {
+        mutations.some(function(mutation) {
+          return mutation.attributeName === 'disabled' && parent[0].disabled;
+        });
+        return false;
+      }
+
+      function windowScrollEventHandler() {
+        setVisible(false);
+      }
+
+      function windowBlurEventHandler() {
+        elementFocusedOnWindowBlur = document.activeElement === parent[0];
+      }
+
+      function enterEventHandler($event) {
+        // Prevent the tooltip from showing when the window is receiving
+        // focus.
+        if ($event.type === 'focus' && elementFocusedOnWindowBlur) {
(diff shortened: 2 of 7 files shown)


src/components/tooltip/tooltip.spec.js
```diff
@@ -436,26 +421,33 @@ describe('<md-tooltip> directive', function() {
...
+    }
+    $rootScope.$apply(
+      'testModel.isVisible = ' + (isVisible ? 'true' : 'false')
+    );
     $material.flushOutstandingAnimations();
   }
 
   function findTooltip() {
-    return angular.element(document.body).find('md-tooltip');
+    // return angular.element(document.body).find('.md-tooltip');
```
DevVersion: Remove unnecessary comment.

src/components/tooltip/tooltip.spec.js
```diff
@@ -436,26 +421,33 @@ describe('<md-tooltip> directive', function() {
...
 
   function showTooltip(isVisible) {
-    if (angular.isUndefined(isVisible)) isVisible = true;
-
-    $rootScope.$apply('testModel.isVisible = ' + (isVisible ? 'true' : 'false') );
+    if (angular.isUndefined(isVisible)) {
+      isVisible = true;
+    }
+    $rootScope.$apply(
+      'testModel.isVisible = ' + (isVisible ? 'true' : 'false')
```
DevVersion: This seems to be a bit ugly. 

```js
$rootScope.testModel.isVisible = !!isVisible;
$rootScope.$apply();
```

src/components/tooltip/tooltip.spec.js
```diff
@@ -275,83 +272,68 @@ describe('<md-tooltip> directive', function() {
...
+          triggerEvent('focus,mousedown');
+
+          expect($document[0].activeElement).toBe(element[0]);
+          expect($rootScope.testModel.isVisible).toBe(true);
+
+          triggerEvent('mouseleave');
+          expect($rootScope.testModel.isVisible).toBe(false);
+
+          // Clean up document.body.
+          $document[0].body.removeChild(element[0]);
```
DevVersion: Can be simplified to

```js
element.remove();
```

src/components/tooltip/tooltip.spec.js
```diff
@@ -1,20 +1,30 @@
...
+    $timeout = $injector.get('$timeout');
+    $mdPanel = $injector.get('$mdPanel');
+    $$mdTooltipRegistry = $injector.get('$$mdTooltipRegistry');
+  };
+
+  beforeEach(function() {
+    module(
+      'material.components.tooltip',
+      'material.components.button',
+      'material.components.panel'
```
DevVersion: Panel should be a dependency on the tooltip itself (so you can remove it here).

> This is **important** for material-tools for example or general the modules we provide (closure or classic)

src/components/tooltip/tooltip.scss
```diff
@@ -6,66 +6,69 @@ $tooltip-top-margin-lg: rem(1.4) !default;
 $tooltip-top-margin-sm: rem(2.4) !default;
 $tooltip-lr-padding-lg: rem(0.8) !default;
 $tooltip-lr-padding-sm: rem(1.6) !default;
-$tooltip-max-width: rem(3.20) !default;
+$tooltip-max-width: rem(3.2) !default;
```
DevVersion: I had this comment on the previous PR as well - Why do we use `rem` here? rem should be for font sizes.

src/components/tooltip/tooltip.js
```diff
@@ -442,9 +479,9 @@ function MdTooltipRegistry() {
 
   /**
    * Removes an event handler from the service.
-   * @param  {String} type Type of event handler.
-   * @param  {Function} handler The event handler itself.
-   * @param  {Boolean} useCapture Whether the event handler used event capturing.
+   * @param {string} type Type of event handler.
+   * @param {function} handler The event handler itself.
+   * @param {boolean} useCapture Whether the event handler used event capturing.
```
DevVersion: Those parameters in the JSDoc shouldn't be nullable? (Same for other docs)

src/components/tooltip/tooltip.js
```diff
@@ -3,391 +3,428 @@
...
+          }
+        }
 
-    function updatePosition() {
-      if ( !scope.visible ) return;
+        function showTooltip() {
+          // Do not show the tooltip if the text is empty.
+          if (!element[0].textContent.trim()) {
+            throw new Error('Text for the tooltip has not been provided. ' +
+                'Please include text within the mdTooltip element.');
```
DevVersion: This needs to be marked as a breaking change.

src/components/tooltip/tooltip.js
```diff
@@ -3,391 +3,428 @@
...
-        element.removeClass(SHOW_CLASS);
-        if (!scope.visible) element.detach();
-      });
-    }
+        function onVisibleChanged(isVisible) {
+          if (isVisible) {
+            showTooltip();
+          } else {
+            hideTooltip();
+          }
```
DevVersion: Can be simplified with a ternary operator.

```js
isVisible ? showTooltip() : hideTooltip();
```

src/components/tooltip/tooltip.js
```diff
@@ -3,391 +3,428 @@
...
-      };
+          function leaveEventHandler() {
+            autohide = scope.hasOwnProperty('mdAutohide') ?
+                scope.mdAutohide :
+                attr.hasOwnProperty('mdAutohide');
+
+            if (
+              autohide ||
+              mouseActive ||
+              $document[0].activeElement !== parent[0]
```
DevVersion: This does not look very good.  

As per the [Code guidelines](https://google.github.io/styleguide/javascriptguide.xml) you should start the autohide directly after the `if (`.

```js
if (autohide || mouseActive ..... // TO LONG
    $document[0].activeElement !== parent[0]) {
```

src/components/tooltip/tooltip.js
```diff
@@ -3,391 +3,428 @@
...
+      var ENTER_EVENTS = 'focus touchstart mouseenter';
+      var LEAVE_EVENTS = 'blur touchcancel mouseleave';
+      var TOOLTIP_SHOW_DELAY = 0;
+      var TOOLTIP_DIRECTION = 'bottom';
+
+      return {
+        restrict: 'E',
+        // transclude: true,
+        priority: 210, // Before ngAria
+        // template: '<div class="md-content" ng-transclude></div>',
```
DevVersion: Remove comments (above as well) if those are not valid anymore.

src/components/tooltip/tooltip.js
```diff
@@ -3,391 +3,428 @@
  * @name material.components.tooltip
  */
 angular
-  .module('material.components.tooltip', [ 'material.core' ])
-  .directive('mdTooltip', MdTooltipDirective)
-  .service('$$mdTooltipRegistry', MdTooltipRegistry);
+    .module('material.components.tooltip', ['material.core'])
```
DevVersion: As mentioned in another comment, this should depend on the panel module as well.

src/components/tooltip/tooltip.spec.js
crisbeto: Alternatively it can also be done as:

```js
$rootScope.$apply(function() {
  $rootScope.testModel.isVisible = !!isVisible;
});
```

src/components/tooltip/tooltip.js
```diff
@@ -3,391 +3,428 @@
...
+          mdDirection: '@?mdDirection' // Do not expect expressions.
+        },
+        link: linkFunc
+      };
 
-    // Initialize element
+      function linkFunc(scope, element, attr) {
+        // Set constants.
+        var parent = $mdUtil.getParentWithPointerEvents(element);
+        var attachTo = angular.element(document.body);
```
crisbeto: From what I'm seeing, `attachTo` is only being used once in the `mdPanel` config. It might be more appropriate to inline it.

src/components/tooltip/tooltip.js
bradrich: @DevVersion This is something that I have been struggling with in my own code as well. Where in the Code Guidelines is this reference? I was unable to find it.

src/components/tooltip/tooltip.js
bradrich: @DevVersion Not necessarily. In the previous version of the tooltip, the tooltip would not show if there was no text assigned. This only adds the throw.

DevVersion: That's why it should be marked as an breaking change. We intentionally added this behavior (enhancement) on request of users.

**See:**: https://github.com/angular/material/issues/8021

src/components/tooltip/tooltip.js
bradrich: @DevVersion FWIU `string` and `boolean` are non-nullable by definition. I have added the `!` symbol before the Function.

src/components/tooltip/tooltip-theme.scss
```diff
@@ -1,6 +1,6 @@
-md-tooltip.md-THEME_NAME-theme {
-  color: '{{background-700-contrast}}';
-  .md-content {
-    background-color: '{{background-700}}';
+.md-tooltip.md-THEME_NAME-theme {
+  color: '{{background-A100}}';
```
bradrich: @crisbeto That is a good question. I am not sure, but the result was the same. I will have to look into why I changed them.

src/components/tooltip/tooltip.js
bradrich: @crisbeto Good catch. I have removed some others that were also only used in one place.

src/components/tooltip/tooltip.js
DevVersion: @bradrich The specs are not anchor able and not very specific about things, but most of the time the code-sample explains a lot. (We also follow this kind of pattern in all other components - Provide consistency)

src/components/tooltip/tooltip.js
ErinCoughlan: lowercase function is also non-nullable by default. Capital Function is nullable by default.

Since you're using the lowercase one, you should be more specific about the params and return types.

`function(paramType): returnType`

src/components/tooltip/tooltip.js
bradrich: @DevVersion I didn't realize. What is the necessary action to take for a breaking change? Just want to make sure that I handle it properly.

DevVersion: Normally you just amend the commit and add to the description

```
Breaking Change: Tooltips no longer show when the content is empty.
```

src/components/tooltip/tooltip.js
bradrich: @DevVersion I understand. I have found that the four-space continuation makes things slightly more unreadable compared to the opening of the parenthesis, but this is nothing that I will debate.

DevVersion: @bradrich Yeah, I agree four spaces is a pain.

ErinCoughlan: Helpful advice via Clang format (I ran it on the entire file - here's the result for this method):

```
      function leaveEventHandler() {
        autohide = scope.hasOwnProperty('mdAutohide') ?
            scope.mdAutohide :
            attr.hasOwnProperty('mdAutohide');

        if (autohide || mouseActive ||
            $document[0].activeElement !== parent[0]) {
          // When a show timeout is currently in progress, then we have
          // to cancel it, otherwise the tooltip will remain showing
          // without focus or hover.
          if (showTimeout) {
            $timeout.cancel(showTimeout);
            setVisible.queued = false;
            showTimeout = null;
          }

          parent.off(LEAVE_EVENTS, leaveEventHandler);
          parent.triggerHandler('blur');
          setVisible(false);
        }
        mouseActive = false;
      }
```

src/components/tooltip/tooltip.js
bradrich: @ErinCoughlan The functions that are being sent in are event handlers. There doesn't seem to be any params or return values, except for the `debouncedOnResize` function which is `$$rAF.throttle(updatePosition)`. For these cases, what would I need to add to make it more readable?

src/components/tooltip/tooltip.js
bradrich: So they do the four space continuation on the `if` statement header, but after that, the inner functionality drops back to a single indentation. @ErinCoughlan, is this formatter something that I might have access to?

bradrich:
I have updated everything requested, but GitHub must have changed something. Normally, the inline comments would disappear after the commented line was changed in the next pushed commit. Now, all of your comments are sticking around...

src/components/tooltip/tooltip.scss
```diff
@@ -1,71 +1,74 @@
...
-$tooltip-lr-padding-lg: rem(0.8) !default;
-$tooltip-lr-padding-sm: rem(1.6) !default;
-$tooltip-max-width: rem(3.20) !default;
+$tooltip-fontsize-lg: 10px !default;
+$tooltip-fontsize-sm: 14px !default;
+$tooltip-height-lg: 22px !default;
+$tooltip-height-sm: 32px !default;
+$tooltip-top-margin-lg: 14px !default;
+$tooltip-top-margin-sm: 24px !default;
+$tooltip-lr-padding-lg: 08px !default;
```
crisbeto: This should be `8px` (without the leading zero).

crisbeto:
Overall LGTM on the second look. I'll try to do some manual testing and final review tomorrow.

src/components/tooltip/tooltip.js
```diff
@@ -3,392 +3,396 @@
...
+ *     <md-icon md-svg-src="img/icons/ic_play_arrow_24px.svg"></md-icon>
+ *   </md-button>
  * </hljs>
  *
- * @param {expression=} md-visible Boolean bound to whether the tooltip is currently visible.
- * @param {number=} md-delay How many milliseconds to wait to show the tooltip after the user focuses, hovers, or touches the
- * parent. Defaults to 0ms on non-touch devices and 75ms on touch.
- * @param {boolean=} md-autohide If present or provided with a boolean value, the tooltip will hide on mouse leave, regardless of focus
- * @param {string=} md-direction Which direction would you like the tooltip to go?  Supports left, right, top, and bottom.  Defaults to bottom.
+ * @param {expression=} md-visible Boolean bound to whether the tooltip is
```
ErinCoughlan: boolean=

src/components/tooltip/tooltip.js
```diff
@@ -3,392 +3,396 @@
...
-        // use an mutationObserver to tackle #2602
+        // Use a mutationObserver to tackle #2602.
         var attributeObserver = new MutationObserver(function(mutations) {
-          if (mutations.some(function (mutation) {
-              return (mutation.attributeName === 'disabled' && parent[0].disabled);
-            })) {
-              $mdUtil.nextTick(function() {
-                setVisible(false);
-              });
+          if (mutations.some(function(mutation) {
```
ErinCoughlan: This is a tad confusing in an if statement. Consider making a function (isDisabled or something) to check this.

src/components/tooltip/tooltip.js
```diff
@@ -3,392 +3,396 @@
...
-    function hideTooltip() {
-      $animate.removeClass(content, SHOW_CLASS).then(function(){
-        element.removeClass(SHOW_CLASS);
-        if (!scope.visible) element.detach();
-      });
-    }
+      var panelAnimation = $mdPanel.newPanelAnimation()
+          .openFrom(parent)
+          .closeTo(parent)
+          .withAnimation({
```
ErinCoughlan: Would using the built in scale animation with a custom duration work in place of this?

Ideally, I'd want to have to use custom classes rarely, so if there's a way to make panel smarter to cover tooltips, I think it should.

src/components/tooltip/tooltip.js
```diff
@@ -3,392 +3,396 @@
...
-    function updatePosition() {
-      if ( !scope.visible ) return;
+      var panelConfig = {
+        attachTo: attachTo,
+        template: element.html(),
+        propagateContainerEvents: true,
+        panelClass: 'md-tooltip ' + origin,
+        animation: panelAnimation,
+        position: panelPosition,
+        zIndex: 100,
```
ErinCoughlan: At some point, consider making this customizable for uses of the tooltip. It's a problem my team has run into several times.

src/components/tooltip/tooltip.js
```diff
@@ -3,392 +3,396 @@
...
+          .withAnimation({
+            open: 'md-show',
+            close: 'md-hide'
+          });
 
-    function updatePosition() {
-      if ( !scope.visible ) return;
+      var panelConfig = {
+        attachTo: attachTo,
+        template: element.html(),
```
ErinCoughlan: Should we implement the content element in panel to make this easier?

src/components/tooltip/tooltip.js
```diff
@@ -406,9 +410,9 @@ function MdTooltipRegistry() {
   };
 
   /**
-   * Global event handler that dispatches the registered
-   * handlers in the service.
-   * @param  {Event} event Event object passed in by the browser.
+   * Global event handler that dispatches the registered handlers in the
+   * service.
+   * @param {Event} event Event object passed in by the browser
```
ErinCoughlan: !Event

src/components/tooltip/tooltip.js
```diff
@@ -420,19 +424,16 @@ function MdTooltipRegistry() {
...
    * Registers a new handler with the service.
-   * @param  {String} type Type of event to be registered.
-   * @param  {Function} handler Event handler
-   * @param  {Boolean} useCapture Whether to use event capturing.
+   * @param {string} type Type of event to be registered.
+   * @param {!Function} handler Event handler.
+   * @param {boolean} useCapture Whether to use event capturing.
    */
   function register(type, handler, useCapture) {
     var array = listeners[type] = listeners[type] || [];
```
ErinCoughlan: This is a horrible variable name. Variable names should be descriptive of what it contains, not it's type. (I know if was here before, but consider changing to handlers or listenersForType.)

src/components/tooltip/tooltip.scss
```diff
@@ -1,71 +1,74 @@
...
 
-md-tooltip {
-  position: absolute;
-  z-index: $z-index-tooltip;
-  overflow: hidden;
+.md-tooltip {
   pointer-events: none;
   border-radius: 4px;
+  overflow: hidden;
+
```
ErinCoughlan: What's with all the new lines between "sections"?

src/components/tooltip/tooltip.js
bradrich: @ErinCoughlan I haven't had a chance to study the new animation work that you and Kristiyan have put together. Can we add a future issue to Github to go back and add the new animation handlers?

src/components/tooltip/tooltip.js
bradrich: @ErinCoughlan In this case, I am grabbing the internal content of the `<md-tooltip></md-tooltip>` element and transposing it into the panel element. I think that it would complicate things to introduce the content element when all you want is the tooltip text here.

src/components/tooltip/tooltip.js
bradrich: I agree. I am adding this now. This actually came up in a demo with Topher.

src/components/tooltip/tooltip.scss
bradrich: I went ahead and removed 90% of them.

src/components/tooltip/tooltip.js
```diff
@@ -3,392 +3,405 @@
...
+      // 'click'.
+      parent.on('mousedown', mousedownEventHandler);
+      parent.on(ENTER_EVENTS, enterEventHandler);
 
-      scope.$on('$destroy', function() {
-          $$mdTooltipRegistry.deregister('scroll', windowScrollHandler, true);
-          $$mdTooltipRegistry.deregister('blur', windowBlurHandler);
-          $$mdTooltipRegistry.deregister('resize', debouncedOnResize);
+      function isDisabledMutation() {
+        mutations.some(function(mutation) {
```
crisbeto: @bradrich This throws an error. Seems like you forgot to pass in the `mutations`.

src/components/tooltip/tooltip.js
```diff
@@ -3,392 +3,405 @@
...
+        panelClass: 'md-tooltip ' + origin,
+        animation: panelAnimation,
+        position: panelPosition,
+        zIndex: scope.mdZIndex,
+        focusOnOpen: false
+      };
 
-      updateContentOrigin();
-      positionTooltip();
+      panelRef = $mdPanel.create(panelConfig);
```
crisbeto: This seems very inefficient, because it creates a new panel ref on every open. I think we should create one only after the first open and then re-use it.

crisbeto:
@bradrich it seems like the margins on the tooltip has changed. Here's a side-by-side (left is what the current tooltip looks like):

untitled-1

src/components/tooltip/tooltip.js
bradrich: You are correct about the inefficiency. I forgot to add the ID parameter to the panel configuration. That will make it only create the panel once for each of the tooltips.

bradrich:
@crisbeto You are correct about the new spacing. According to the [Material Guidelines](https://material.google.com/components/tooltips.html#tooltips-tooltips-desktop), there is supposed to be 14 pixels between the element and the tooltip on a desktop. 24 pixels if you are on a mobile device. I have positioned the tooltip according to those requirements.

bradrich:
@crisbeto @ErinCoughlan Please take a look at the final revisions and provide an LGTM if everything is GTG. Thank you.

ThomasBurleson:
@crisbeto , @ErinCoughlan  - ping request for lgtm.

src/components/tooltip/tooltip.js
```diff
@@ -3,392 +3,418 @@
...
 
-      element.css({
-        left: newPosition.left + 'px',
-        top: newPosition.top + 'px'
-      });
+      var content = element.html().trim();
+
+      // Convert the content to camelCase to create the tooltip panel id.
+      // Note: This will also remove any non-alpha characters.
+      var id = content
```
crisbeto: @bradrich Why do we have to use the content to generate an id for the panel? This is seems like a lot of troubke for something that could easily be done with `"tooltip-" + $mdUtil.nextUid()`.

bradrich: @crisbeto This is a very good point. I have made this alteration.

src/components/tooltip/tooltip.js
```diff
@@ -3,392 +3,407 @@
...
     restrict: 'E',
-    transclude: true,
     priority: 210, // Before ngAria
-    template: '<div class="md-content _md" ng-transclude></div>',
     scope: {
-      delay: '=?mdDelay',
-      visible: '=?mdVisible',
-      autohide: '=?mdAutohide',
-      direction: '@?mdDirection'    // only expect raw or interpolated string value; not expression
+      mdZIndex: '=?mdZIndex',
```
ErinCoughlan: optional: since you are no longer renaming these, you don't need to write the attribute name again:

```
mdZIndex: '=?',
mdDelay: '=?', 
...
```

src/components/tooltip/tooltip.js
```diff
@@ -3,392 +3,407 @@
...
   var ENTER_EVENTS = 'focus touchstart mouseenter';
   var LEAVE_EVENTS = 'blur touchcancel mouseleave';
-  var SHOW_CLASS = 'md-show';
-  var TOOLTIP_SHOW_DELAY = 0;
-  var TOOLTIP_WINDOW_EDGE_SPACE = 8;
+  var TOOLTIP_DEFAULT_Z_INDEX = 100;
+  var TOOLTIP_DEFAULT_SHOW_DELAY = 0;
+  var TOOLTIP_DEFAULT_DIRECTION = 'bottom';
+  var TOOLTIP_DIRECTIONS = {
+    top: { x: 'CENTER', y: 'ABOVE' },
```
ErinCoughlan: This is be more robust if you use the enum defined in mdPanel instead of writing them here. If, for instance, I decide that the text shouldn't be capitalized and change it, this entire component will break and it shouldn't.

src/components/tooltip/tooltip.spec.js
```diff
@@ -275,83 +274,68 @@ describe('<md-tooltip> directive', function() {
...
+          triggerEvent('focus,mousedown');
+
+          expect($document[0].activeElement).toBe(element[0]);
+          expect($rootScope.testModel.isVisible).toBe(true);
+
+          triggerEvent('mouseleave');
+          expect($rootScope.testModel.isVisible).toBe(false);
+
+          // Clean up document.body.
+          element.remove();
```
ErinCoughlan: Do you want to do something like this in the `afterEach` to be extra safe from someone forgetting to clean up?

src/components/tooltip/tooltip.js
ThomasBurleson: @bradrich - please debounce the showing/hiding of the tooltip.

ThomasBurleson:
@bradrich - ping regard last requests from @ErinCoughlan .

bradrich:
@ThomasBurleson All clear!

bradrich:
@ThomasBurleson This one is all clear and good to go. Conflicts have been resolved.

src/components/tooltip/tooltip-theme.scss
clshortfuse: The colors shouldn't be changed here. It seems you didn't rebase properly here after this PR:

https://github.com/angular/material/pull/9207

bradrich: @clshortfuse The result of the tooltip using the colors that I have specified is the same exact color scheme of the tooltips that are already present in the material demos. Changing them back to the `background-700` color doesn't work.

clshortfuse: @bradrich I don't know what you mean by "have the same color scheme", but the material design spec says `background-700`. https://material.google.com/components/tooltips.html#tooltips-tooltips-desktop-

By having these colors changed back, you'd be reintroducing old issues with dark themes and non-gray background palettes.

src/components/tooltip/tooltip.scss
```diff
@@ -1,71 +1,67 @@
...
+  }
+  
+  &.md-show-add {
+    transform: scale(0);
+  }
   &.md-show {
     transition: $swift-ease-out;
-    pointer-events: auto;
+    transform: scale(1);
+    opacity: 1;
```
clshortfuse: Opacity should be `0.9` (90%) as in master

bradrich:
@ThomasBurleson Checking on the status of this PR.

PR #9746, opened 2016-10-04
fix(md-calendar): boundKeyHandler preventDefault on other input elements

NgaiKaKit:
If use the md-calendar directly in the body without datepicker, boundKeyHandler will disable some keyEvent on other input elements in the page. So only apply the handleKeyEvent on the document.body when the md-calendar is inside datepicker. Otherwise, apply on the calendar element only.

### DIFF
--- a/src/components/datepicker/js/calendar.js
+++ b/src/components/datepicker/js/calendar.js
@@ -198,15 +198,17 @@
     var boundKeyHandler = angular.bind(this, this.handleKeyEvent);
 
 
-    var handleKeyElement = this.$element;
 
     // If use the md-calendar directly in the body without datepicker,
     // handleKeyEvent will disable other inputs on the page.
     // So only apply the handleKeyEvent on the body when the md-calendar inside datepicker,
     // otherwise apply on the calendar element only.
 
+    var handleKeyElement;
     if (this.$element.parent().hasClass('md-datepicker-calendar')) {
       handleKeyElement = angular.element(document.body);
+    } else {
+      handleKeyElement = this.$element;
     }
 
     // Bind the keydown handler to the body, in order to handle cases where the focused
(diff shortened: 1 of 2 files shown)


hnomka:
I signed it!

src/components/datepicker/js/calendar.js
```diff
@@ -197,12 +197,24 @@
...
+
+    var handleKeyElement = this.$element;
+
+    // If use the md-calendar directly in the body without datepicker,
+    // handleKeyEvent will disable other inputs on the page.
+    // So only apply the handleKeyEvent on the body when the md-calendar inside datepicker,
+    // otherwise apply on the calendar element only.
+
+    if (this.$element.parent().hasClass('md-datepicker-calendar')) {
+      handleKeyElement = angular.element(document.body);
```
crisbeto: It might be a little more readable if you defined the `handleKeyEvent` in an `else` block. Also you can refer to `this.$element` as just `$element`:

```js
if ($element.parent().hasClass('md-datepicker-calendar')) {
  handleKeyElement = angular.element(document.body);
} else {
  handleKeyElement = $element;
}
```

src/components/datepicker/js/calendar.spec.js
```diff
@@ -113,7 +113,14 @@ describe('md-calendar', function() {
   function dispatchKeyEvent(keyCode, opt_modifiers) {
     var mod = opt_modifiers || {};
 
-    angular.element(document.body).triggerHandler({
+    var dispatchElement;
+      if (angular.element(element).parent().hasClass('md-datepicker-calendar')) {
```
crisbeto: You don't need this check. The calendar here will always be on it's own (without the datepicker).

NgaiKaKit:
I signed it!

PR #9748, opened 2016-10-04
﻿fix(autocomplete): margin consistency with input

### LINKED ISSUE #9692: md-autocomplete alignment problem using layout-margin
**Actual Behavior**:
 -  `What is the issue? *`
When using md-autocomplete (floating label) in an element with layout='row' and layout-margin='' it doesn't align correctly with the other form elements.
 -  `What is the expected behaviour?`
The md-autocomplete field aligns with other form elements

http://codepen.io/oostmeijer/pen/xEdxop
 -  `Details:`
Keep your screen wide enough to have both fields on one row. The autocomplete field is lower than the first 'normal' form field.

**Angular Versions**: *
 -  `Angular Version:` 1.5.7

**Additional Information**:
 -  `Browser Type: *` Safari, Chrome
 -  `Browser Version: *`Versie 9.1.1 (11601.6.17), Versie 53.0.2785.116 (64-bit)
 -  `OS: *` OSX 10.11.5
 -  `Stack Traces:`

----

DevVersion:
* Curently the autocomplete integrates as expected with the input container, because the underlaying input container handles all the margin.

* But when using `layout-margin`, then there will be some inconsistency, because the attribute changes the following margins
  - `md-input-container` (Normal Input - 18px -> 8px)
  - `md-autocomplete` (Autocomplete  - 0px -> 8px)

The issue is, that the `md-autocomplete` still uses the deault input container margin, because it's *inside*  of the autocomplete.

This means that the autocomplete is now higher than the input container.

| Default Input (no extra margin) | Default autocomplete (no extra margin) |
| ---------------------------------- | -------------------------------------------- |
|  |  |

> Basically you see that the margin is outside on the input-container.

---
@ThomasBurleson Not sure it's worth making this change, because this means we need to take care of the input container margin in the autocomplete SCSS.

> The solution before, was completely in sync with the `md-input-container` 
> I personally would go with the previous approach and not fix the issue (possible approach PR )

Fixes #9692.

### DIFF
--- a/src/components/autocomplete/autocomplete.scss
+++ b/src/components/autocomplete/autocomplete.scss
@@ -1,5 +1,6 @@
 // The default item height is also specified in the JavaScript.
-$md-autocomplete-item-height: 48px !default;
+$md-autocomplete-item-height: 48px !default
+$md-autocomplete-input-container-margin: 18px 0;
 
 @keyframes md-autocomplete-list-out {
   0% {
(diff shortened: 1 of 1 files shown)


ThomasBurleson:
@DevVersion - based on your research and feedback, I agree that we should not change this.
Closing as will not fix.

PR #9773, opened 2016-10-06
fix(docs): prevent tabbing over hidden content; better animation handling

### LINKED ISSUE #8896: Docs version picker makes sidenav hard to use with a keyboard or screen reader
**Actual Behavior**:
 -  `What is the issue?` *
Version picker links in the docs sidenav aren't hidden from keyboard and screen reader users when the list is visually closed. That means it's still a huge pain to try to navigate the sidebar with a keyboard. This issue was open previously as https://github.com/angular/material/issues/4106, and it still exists.

The `menu-toggle-list` should be hidden with `display: none` after animating completes. This could be done with CSS keyframes if it impacts the animation. Alternatively, links could be removed from the tab order with `tabindex="-1"` on each control and `aria-hidden="true"` on the parent, but that would take more work.

 -  `Details:`
Go to any page on https://material.angularjs.org and try to tab through the sidenav. Observe your keyboard focus go through every item in the (closed) version picker, and then the expand/collapse links which work correctly. 

See the version picker announced even when it is closed:
Version picker announced even though it is closed

**Angular Versions**: *
 -  `Angular Version:`
Current / All past versions 

**Additional Information**:
 -  `Browser Type: *`
 -  `Browser Version: *`
 -  `OS: *`
 -  `Stack Traces:`
All - Visible with keyboard alone, but also with Voiceover, NVDA, JAWS, etc.
----

  Add more lines if needed.

crisbeto:
* Fixes users being able to tab through the hidden content on the docs site.
* Switches the docs site's menu accordions to use `$animateCss` for their transitions. This allows us to reliably know when an animation has started and finished.

Fixes #8896.

### DIFF
--- a/docs/app/css/style.css
+++ b/docs/app/css/style.css
@@ -355,8 +355,7 @@ body[dir=rtl] .menu-heading {
   overflow: hidden;
   position: relative;
   z-index: 1;
-  transition: 0.75s cubic-bezier(0.35, 0, 0.25, 1);
-  transition-property: height;
+  height: 0;
 }
 .docs-menu .menu-toggle-list a.md-button {
   display: block;
--- a/docs/app/js/app.js
+++ b/docs/app/js/app.js
@@ -492,7 +492,7 @@ function(SERVICES, COMPONENTS, DEMOS, PAGES, $location, $rootScope, $http, $wind
   };
 })
 
-.directive('menuToggle', [ '$timeout', '$mdUtil', function($timeout, $mdUtil) {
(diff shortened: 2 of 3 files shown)


ThomasBurleson:
@EladBezalel , @DevVersion, @topherfangio - who can test the A11Y fixes for this ?

topherfangio:
@ThomasBurleson I'll do some testing for the a11y today.

topherfangio:
Tested on Firefox/NVDA and IE11/JAWS. LGTM on both 👍

PR #9777, opened 2016-10-07
docs(icon): l10n translations mangles interpolated font ligature

### LINKED ISSUE #9750: docs: use ng-bind for md-icon ligature documentation
**Actual Behavior**:
 -  `What is the issue? *` Current documentation for md-icon uses Angular text markup (eg: `{{...}}`)
 -  `What is the expected behavior?` ng-bind should be used, to keep logic modular inside the md-icon declaration. This is also an important change for users that use multilingual support in their applications and use our demo code as a template.

 -  `CodePen Demo which shows your issue:` https://material.angularjs.org/1.1.0/demo/icon

**Angular Versions**: *

 -  `Angular Material Version:`  1.1.0+

----

ping @ErinCoughlan

clshortfuse:
For md-icon font ligatures, `ng-bind` should be used, to keep logic modular inside the md-icon declaration. This is also an
important change for users that use multilingual support in their applications and use our demo code as a template

* Update icon demo to use ng-bind instead of inline text interpolization
* Update icon documentation to explain ng-bind usage and its importance with localization support

`ng-bind` is more also more efficient than text interpolization.

Fixes #9750

### DIFF
--- a/src/components/icon/demoFontIconsWithLigatures/index.html
+++ b/src/components/icon/demoFontIconsWithLigatures/index.html
@@ -12,8 +12,8 @@
         <md-icon ng-style="{color: font.color}"
             aria-label="{{ font.name }}"
             class="material-icons step"
-            ng-class="it.size">
-          {{ font.name }}
+            ng-class="it.size"
+            ng-bind="font.name">
         </md-icon>
       </div>
     </div>
--- a/src/components/icon/js/iconDirective.js
+++ b/src/components/icon/js/iconDirective.js
@@ -48,13 +48,18 @@ angular
  *    rel="stylesheet">`
  * </li>
  * <li>
- *   Use either (a) font-icon class names or (b) font ligatures to render the font glyph by using
- *   its textual name
+ *   Use either (a) font-icon class names or (b) a fontset and a font ligature to render the font glyph by
+ *   using its textual name _or_ numerical character reference. Note that `material-icons` is the default fontset when
+ *   none is specified.
  * </li>
- * <li>
- *   Use `<md-icon md-font-icon="classname" />` or <br/>
- *   use `<md-icon md-font-set="font library classname or alias"> textual_name </md-icon>` or <br/>
- *   use `<md-icon md-font-set="font library classname or alias"> numerical_character_reference </md-icon>`
+ * <li> Use any of the following templates: <br/>
+ *   <ul>
+ *     <li>`<md-icon md-font-icon="classname"></md-icon>`</li>
+ *     <li>`<md-icon md-font-set="font library classname or alias">textual_name</md-icon>`</li>
+ *     <li>`<md-icon> numerical_character_reference </md-icon>`</li>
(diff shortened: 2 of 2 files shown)


clshortfuse:
Review with: http://codepen.io/shortfuse/pen/GjQxap

topherfangio:
This needs a note in the demo, and probably the docs as well. If we do something an "unstandard" way, we should document why we do it that way.

Also, would love @ErinCoughlan's thoughts on this.

clshortfuse:
@topherfangio We can probably link to https://docs.angularjs.org/api/ng/directive/ngBind 

To be honest, I think we should `ng-bind` whenever possible, if we're not mixing content "mid-sentence". IMO, it looks cleaner to do it all during the declaration. I'm just assuming, but It's probably more efficient than `{{...}}` since it we don't ask Angular to parse the text content and perform text replace operations. 

I'm thinking something along the lines of stating that we bind to the variable instead of replacing the text since we don't expect the actual string to be human-readable text (ie: `&#xE91D;`). And it allows better interoperability with multilingual support.

ErinCoughlan:
We need to be explicit with why we are doing this - performance benefits, multilingual support, etc. Ideally, we would use the material docs as a resource to point to saying "this is the right way to use md-icon and this is why"

ThomasBurleson:
A use case [that justifies this change] has not been presented. Documentation-alone is not sufficient since interpolation works fine. Rejecting this PR as "noise".

ErinCoughlan:
@ThomasBurleson Should we add a new attribute to md-icon instead that binds the text and makes it the obvious solution? We also suggested `md-font-value` before learning that `ng-bind` also works.

The issue here is that using ligatures, when they become translated, no longer match the icons and then the user sees very large, unrelated text with poor alignment.

Using ng-bind lets the font value be part of `md-icon` rather than random text on the page that gets translated. However, this solution is not obvious because it's not mentioned in the docs or used in any of the demos.

What would you like to do here? I feel strongly that we need a solution. I initially wanted the separate attribute and I think I still slightly prefer that, but after talking to the team, I was convinced this was easier.

ThomasBurleson:
@ErinCoughlan - to paraphrase: 

> The issue here is how to prevent translations of font-icon ligatures when used in interpolated strings?

Using `ng-bind` seems like a very reasonable solution. I do not think that either a modification to the mdIcon API or  a global use of `ng-bind` is warranted. I do think we should improve the documentation to discuss this issue and highlight the `ng-bind` workaround.

ThomasBurleson:
@clshortfuse - please add documentation regarding use of `ng-bind` with translation issues. Extra docs can go in https://material.angularjs.org/latest/api/directive/mdIcon#Localization section (before Usage).

ThomasBurleson:
@clshortfuse - any progress ?

clshortfuse:
I've made changes to the documentation by adding more examples and a localization section. Adding something to the documention about how one thing (interpolization) is better than another (ng-bind) is rather awkward, but I did add it to the commit message.

PR #9780, opened 2016-10-07
fix(icon): codepen demo issue

### LINKED ISSUE #9561: demos(): "Edit on Codepen" icon not working in icon demo
The "Edit on Codepen" functionality on the "Load Svg Icons From Url" [demo for icons](https://material.angularjs.org/HEAD/demo/icon) doesn't generate anything.

DevVersion:
* The icon demo with the SVG files from URLs was not working for the codepen checkout, because it seems like the "inlined" SVG has some escaping issues (especially when being wrapped inside of single qoutes inside of HTML).

* Moving the SVG template into the Script (as same as for the decoded SVG) does fix the issue, and is also makes the demo look better.

Fixes #9561.

### DIFF
--- a/src/components/icon/demoLoadSvgIconsFromUrl/index.html
+++ b/src/components/icon/demoLoadSvgIconsFromUrl/index.html
@@ -1,4 +1,4 @@
-<div ng-controller="DemoCtrl" ng-cloak layout="column" layout-margin  ng-cloak>
+<div ng-controller="DemoCtrl" ng-cloak layout="column" layout-margin>
     <p>The simplest way to display a single SVG icon is by referencing it by URL:</p>
     <p>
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
Nice!

src/components/icon/demoLoadSvgIconsFromUrl/index.html
```diff
@@ -1,4 +1,4 @@
-<div ng-controller="DemoCtrl" ng-cloak layout="column" layout-margin  ng-cloak>
+<div ng-controller="DemoCtrl" ng-cloak layout="column" layout-margin>
```
ThomasBurleson: Why did you remove the `ng-cloak` which is used to hide flickers due to Layout API activity.

PR #9787, opened 2016-10-08
fix(docs, dialog, interim, panel): compatibility with latest angular snapshot

crisbeto:
* Fixes the docs site not working against the latest snapshot.
* Fixes the `dialog` throwing an error about `ngSanitize` not being present and then compiling the contents anyway.
* Makes the `interimElement` compatible both with the latest changes to promises from Angular PR #15213 and the current promise handling in Angular <= 1.5.
* Reintroduces the panel changes from #9688 which were overwritten accidentally.
* Fixes an error that was being swallowed during unit tests, but was still showing up in the `$exceptionHandler`.

### DIFF
--- a/src/core/services/interimElement/interimElement.js
+++ b/src/core/services/interimElement/interimElement.js
@@ -244,8 +244,8 @@ function InterimElementProvider() {
   }
 
   /* @ngInject */
-  function InterimElementFactory($document, $q, $$q, $rootScope, $timeout, $rootElement, $animate,
-                                 $mdUtil, $mdCompiler, $mdTheming, $injector ) {
+  function InterimElementFactory($document, $q, $rootScope, $timeout, $rootElement, $animate,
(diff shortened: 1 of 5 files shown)


DevVersion:
LGTM for the  `$$interimElement` and `dialog` stuff.

src/core/services/interimElement/interimElement.js
```diff
@@ -682,7 +686,12 @@ function InterimElementProvider() {
           // Trigger onComplete callback when the `show()` finishes
           var notifyComplete = options.onComplete || angular.noop;
 
-          notifyShowing(options.scope, element, options, controller);
+          // Necessary for consistency between Angular 1.5 and 1.6.
+          try {
+            notifyShowing(options.scope, element, options, controller);
+          } catch (e) {
+            return $q.reject(e);
```
ThomasBurleson: 👍

PR #9790, opened 2016-10-09
fix(radio-group): wrong aria-checked value on load when used with ng-value

### LINKED ISSUE #9400: Toggling radio buttons makes all of them appear checked to JAWS
Have a report that, with an `<md-radio-group>`, all of the options are appearing checked at the same time with JAWS. We should investigate with IE11, Edge, Chrome, and Firefox.

The radio button template looks something like

``` html
<md-radio-group ng-model="ctrl.element" class="element-options"
                ng-change="ctrl.updateChoice()">
  <md-radio-button ng-value="'EARTH'" class="md-primary">
    Running on the earth
  </md-radio-button>
  <md-radio-button ng-value="'WIND'" class="md-primary">
    Dancing in the wind
  </md-radio-button>
  <md-radio-button ng-value="'WATER'" class="md-primary">
    Swimming in the water
  </md-radio-button>
</md-radio-group>
```

crisbeto:
- Fixes all of the radio buttons within a radio group having `aria-checked` set to true on load, when using `ng-value`.
- Cleans up the `radioButton` directive.

Fixes #9400.

### DIFF
--- a/src/components/radioButton/radio-button.js
+++ b/src/components/radioButton/radio-button.js
@@ -62,7 +62,7 @@ function mdRadioGroupDirective($mdUtil, $mdConstant, $mdTheming, $timeout) {
   function linkRadioGroup(scope, element, attr, ctrls) {
     element.addClass('_md');     // private md component indicator for styling
     $mdTheming(element);
-    
+
     var rgCtrl = ctrls[0];
     var ngModelCtrl = ctrls[1] || $mdUtil.fakeNgModel();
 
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@crisbeto - ping regard `elemet` mis-spelling.

crisbeto:
Fixed @ThomasBurleson @jelbourn. We should really get that linter set up...

ThomasBurleson:
@crisbeto - sgtm. Btw - I missed the spelling issue myself while reviewing the code.

PR #9800, opened 2016-10-10
chore(docs): autoprefix docs stylesheet

### LINKED ISSUE #9785: docs: use autoprefixer for docs gulpfile
Currently the docs gulp tasks are very small and do not even autoprefix the CSS, and all the properties are manually prefixed right now.

DevVersion:
* Uses the shared autoprefixer from the gulp utils to autoprefix the docs CSS (instead of always doing it manually as in #9786)

> Kept a few special properties like (`-webkit-font-smoothing` because those can't be autoprefixed)

Closes #9785

### DIFF
--- a/docs/app/css/style.css
+++ b/docs/app/css/style.css
@@ -37,11 +37,6 @@ body.docs-body {
   position: fixed !important;
 
   transform: scale(0);
-
-  /* Manually prefix because the docs won't be prefixed. See #9785 */
-  -webkit-transition: transform 0.2s;
-  -moz-transition: transform 0.2s;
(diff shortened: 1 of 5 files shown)


docs/app/css/style.css
```diff
@@ -669,7 +626,6 @@ ul.buckets li a {
   font-weight: 500;
   padding: 16px 0;
   text-decoration: none;
-  -webkit-transition: all 0.45s cubic-bezier(0.35, 0, 0.25, 1);
```
DevVersion: FYI: This was intentionally removed, because it will be set in the selector below (this was an unnecessary property)

PR #9801, opened 2016-10-10
fix(card): fix alignment with avatar icons in Safari

### LINKED ISSUE #9147: Card avatar alignment incorrect in Safari
When using an avatar in the heading of an md-card, the svg does not align to the middle of the icon, and so the border looks incorrect

This can be seen by viewing the Angular Material docs in safari and looking at the page for 'Card'.

Material version - 1.1.0-RC5
Safari version -  9.1 (11601.5.17.1)

screen shot 2016-07-27 at 11 48 22

clshortfuse:
Safari has an alignment issue with SVG icons when its parent has
any padding, causing its content to shift placement

* Force SVG elements to inherit height and width properties.

Fixes #9147

### DIFF
--- a/src/components/card/card.scss
+++ b/src/components/card/card.scss
@@ -37,6 +37,11 @@ md-card {
 
       md-icon {
         padding: 8px;
+        > svg {
+          // Safari workaround for any SVG with padded parent
+          height: inherit;
+          width: inherit;
+        }
       }
 
       & + md-card-header-text {


clshortfuse:
Review with: http://codepen.io/shortfuse/full/GjxLYj/

Tested OK on IE 11, Safari 9, Chrome 53, Firefox 49, Android 4.4.2

If somebody knows an alternative `-webkit-` solution, please share. `-webkit-calc(100%)` did not work.

topherfangio:
@clshortfuse I don't see a `-webkit-` anything in your changes. What are you talking about? Can you explain a bit more?

clshortfuse:
@topherfangio The idea being, if we are targeting a Safari issue, we would **ideally** limit the CSS to just Safari. 

The only possible solution I found the one in the commit, that was applies to all browsers. As I said before, though, I didn't notice any issue in any of our supported browsers.

PR #9803, opened 2016-10-10
fix(datepicker): remove dependency on $mdGesture

### LINKED ISSUE #9793: datepicker: remove dependency to $mdGesture
In favor of #9362 the Datepicker component should not depend on the `$mdGesture` service.

> The datepicker right now, just uses the service to detect whether the current agent is a touch device or not.

> This check can be inlined, because after #9362 nothing will have a similar check.

Notice that this is just a PR to make the integration to HammerJS easier.

crisbeto:
Removes the datepicker's dependency on `$mdGesture` by inlining the check. This is a preparation for the switch over to HammerJS.

Fixes #9793.

### DIFF
--- a/src/core/services/gesture/gesture.js
+++ b/src/core/services/gesture/gesture.js
@@ -79,8 +79,6 @@ function MdGesture($$MdGestureHandler, $$rAF, $timeout) {
   var self = {
     handler: addHandler,
     register: register,
-    isIos: isIos,
-    isAndroid: isAndroid,
     // On mobile w/out jQuery, we normally intercept clicks. Should we skip that?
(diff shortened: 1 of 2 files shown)


src/core/services/gesture/gesture.js
```diff
@@ -79,8 +79,6 @@ function MdGesture($$MdGestureHandler, $$rAF, $timeout) {
   var self = {
     handler: addHandler,
     register: register,
-    isIos: isIos,
```
Ardeshir81: I think deleting it wasn't a nice thing. people may be using it outside of angular material. as a normal service which is capable of identifying if visitor is on iOS or Android.

Issue #9798, opened 2016-10-10
Form Layout Issues


mattpawson:
**Actual Behavior**:
 -  `What is the issue? 

When attempting to build a form that utilises a number of different components the layout doesn't appear to render correctly.

The datepicker doesn't fill the whole of the row in the same manner as select and text boxes do, while the autocomplete's height is larger than that normal textbox / select box.

 -  `What is the expected behavior?`

I expected all the input elements to be of equal width / height when building the form.

**CodePen** (or steps to reproduce the issue): *
http://codepen.io/mattpawson/pen/ALybpq/

**Angular Versions**: *
 -  `Angular Version:` 1.5.8
 -  `Angular Material Version:` 1.1.1

**Additional Information**:
 -  `Browser Version: *` Firefox 49.0.1, Chrome: 53.0.2785
 -  `OS: *` Windows 10

crisbeto:
The datepicker isn't meant to stretch the full width, because it changes the background on the input, in order to make the calendar masking work. That being said, if you know that your form won't stretch too much, you can get it working by adding the following to your CSS (note that I haven't included any vendor prefixes):

```css
md-datepicker {
    display: flex;
    align-items: center;
}

.md-datepicker-input-container {
    flex: 1;
}
```

[Here's a forked codepen.](http://codepen.io/crisbeto/pen/rrdROO) Also it seems like the arrow isn't quite aligned, I'll try to fix that one.

mattpawson:
Thanks for your response. I like the calendar icon, it gives the user a good indicator of the type of input, so I'll probably aim to leave it there. I did manage to achieve something similar, but it wasn't as elegant as yours:

	.md-datepicker-input-container {
	  width: calc(100% - 64px);
	  margin-left: 0px;
	}

Any news on the autocomplete's, and why their height is so much taller than the other elements?

crisbeto:
I'm not too sure about the autocomplete since I'm not responsible for it, but @DevVersion should know more about it.

crisbeto:
I'll close the issue for now, but we can continue the discussion.

PR #9807, opened 2016-10-10
fix(datepicker,select): arrow button consistency

crisbeto:
- Fixes the datepicker's arrow being off a few pixels inside `md-input-container`, as well as being too close to the bottom when not in a `md-input-container`.
- Fixes the select's arrow icon being too dark, because it inherits from the closest `color` rule.

Relates to #9798.

For reference, here's what it looked like before:
before

And here's what it looks like after:
after

### DIFF
--- a/src/components/datepicker/datePicker.scss
+++ b/src/components/datepicker/datePicker.scss
@@ -194,10 +194,10 @@ md-datepicker {
 .md-datepicker-triangle-button {
   position: absolute;
   @include rtl-prop(right, left, 0, auto);
-  top: $md-date-arrow-size;
+  bottom: -$md-date-arrow-size / 2;
 
   // TODO(jelbourn): This position isn't great on all platforms.
-  @include rtl(transform, translateY(-25%) translateX(45%), translateY(-25%) translateX(-45%));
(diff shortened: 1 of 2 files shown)


src/components/datepicker/datePicker.scss
```diff
@@ -194,10 +194,10 @@ md-datepicker {
 .md-datepicker-triangle-button {
   position: absolute;
   @include rtl-prop(right, left, 0, auto);
-  top: $md-date-arrow-size;
+  bottom: -$md-date-arrow-size / 2;
 
   // TODO(jelbourn): This position isn't great on all platforms.
-  @include rtl(transform, translateY(-25%) translateX(45%), translateY(-25%) translateX(-45%));
+  @include rtl(transform, translateX(45%), translateX(-45%));
```
crisbeto: I'm not really sure what the TODO is referring to. I removed it since it's unnecessary.

topherfangio:
This looks good, but what demo can we look at to see the changes? Or was your image from a Codepen?

crisbeto:
My image is from me running the docs site locally. I shuffled some HTML around in the input demo.

PR #9817, opened 2016-10-11
fix(subheader): add accessibility support

### LINKED ISSUE #9392: Subheader has no accessibility treatment
It appears that the subheader has no accessibility treatment. Follow up with me for more details.

crisbeto:
Previously, due to the way the sticky headers work, lists with an `md-subheader` were being read out by screen readers as follows:
- Header 2
- Header 1
- Header 1
- Content 1
- Header 2
- Content 2

This can be confusing for users since it doesn't reflect the real structure of the page, as well as not reading out the headers as being headers.
These changes add the ARIA attributes that allow for the content to be read out as:
- Header 1
- Content 1
- Header 2
- Content 2

Fixes #9392.

### DIFF
--- a/src/components/subheader/subheader.js
+++ b/src/components/subheader/subheader.js
@@ -52,7 +52,7 @@ angular
  * </hljs>
  */
 
-function MdSubheaderDirective($mdSticky, $compile, $mdTheming, $mdUtil) {
+function MdSubheaderDirective($mdSticky, $compile, $mdTheming, $mdUtil, $mdAria) {
   return {
     restrict: 'E',
     replace: true,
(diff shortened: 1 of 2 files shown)


src/components/subheader/subheader.js
```diff
@@ -78,6 +78,12 @@ function MdSubheaderDirective($mdSticky, $compile, $mdTheming, $mdUtil) {
         return angular.element(el[0].querySelector('.md-subheader-content'));
       }
 
+      // Set the ARIA attributes on the original element since it keeps it's original place in
+      // the DOM, whereas the clones are in reverse order. Should be done after the outerHTML,
+      // in order to avoid having multiple element be marked as headers.
+      attr.$set('role', 'heading');
+      attr.$set('aria-level', '2');
```
jelbourn: This shouldn't be hard-coded to 2; it should be whatever header-level is appropriate for that section of the DOM. We should expose an attribute that lets users specify the level and probably console.log a warning if it's missing.

ThomasBurleson: @jelbourn - why not using existing aria-level (if specified) or fallback to default level of `2`? ... and warn user that the aria-level attribute is missing (if not specified).

crisbeto: Makes sense. We _could_ try to figure it out automatically, but it would probably be too slow/error prone.

crisbeto:
Changed to allow for the `aria-level` to be overwritten and to default to 2. I went with 2 since it's a **sub**header.

ThomasBurleson:
@crisbeto - lgtm.

jelbourn:
I really don't think just saying "2" as the default is a good idea. This can lead to having an invalid header structure.

@marcysutton do you have any thoughts on this?

crisbeto:
It's the clone being hidden @marcysutton. Here's what is being read out in the current setup:

- Header 2
- Header 1
- Header 1
- Content 1
- Header 2
- Content 2

PR #9818, opened 2016-10-11
perf(navbar): reduces amount of watchers and buttons per item

crisbeto:
With the current approach, the navbar creates 3 child buttons and uses 3 `ng-if` statements to determine which one to render. This seems wasteful.
This PR moves that logic to the compilation phase which will only be triggered once (per occurrence in the DOM) and won't add any more watches.

### DIFF
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -388,27 +388,44 @@ function MdNavItem($$rAF) {
     controllerAs: 'ctrl',
     replace: true,
     transclude: true,
-    template:
-      '<li class="md-nav-item" role="option" aria-selected="{{ctrl.isSelected()}}">' +
-        '<md-button ng-if="ctrl.mdNavSref" class="_md-nav-button md-accent"' +
-          'ng-class="ctrl.getNgClassMap()"' +
-          'tabindex="-1"' +
-          'ui-sref="{{ctrl.mdNavSref}}">' +
-          '<span ng-transclude class="_md-nav-button-text"></span>' +
-        '</md-button>' +
-        '<md-button ng-if="ctrl.mdNavHref" class="_md-nav-button md-accent"' +
-          'ng-class="ctrl.getNgClassMap()"' +
-          'tabindex="-1"' +
-          'ng-href="{{ctrl.mdNavHref}}">' +
-          '<span ng-transclude class="_md-nav-button-text"></span>' +
-        '</md-button>' +
-        '<md-button ng-if="ctrl.mdNavClick" class="_md-nav-button md-accent"' +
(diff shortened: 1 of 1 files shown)


src/components/navBar/navBar.js
```diff
@@ -388,27 +388,45 @@ function MdNavItem($$rAF) {
...
+    template: function(tElement, tAttrs) {
+      var hasNavClick = tAttrs.mdNavClick;
+      var hasNavHref = tAttrs.mdNavHref;
+      var hasNavSref = tAttrs.mdNavSref;
+
+      // Cannot specify more than one nav attribute
+      if ((hasNavClick ? 1:0) + (hasNavHref ? 1:0) + (hasNavSref ? 1:0) > 1) {
+        throw Error(
+          'Must specify exactly one of md-nav-click, md-nav-href, ' +
+          'md-nav-sref for nav-item directive'
```
clshortfuse: I know it was like this in master, but it should read `'or md-nav-sref for nav-item directive'`

src/components/navBar/navBar.js
```diff
@@ -388,27 +388,45 @@ function MdNavItem($$rAF) {
...
+        );
+      }
+
+      var navigationAttribute = '';
```
clshortfuse: You don't need to initialize this with a value.

src/components/navBar/navBar.js
```diff
@@ -388,27 +388,45 @@ function MdNavItem($$rAF) {
...
-      '</li>',
+    template: function(tElement, tAttrs) {
+      var hasNavClick = tAttrs.mdNavClick;
+      var hasNavHref = tAttrs.mdNavHref;
+      var hasNavSref = tAttrs.mdNavSref;
+
+      // Cannot specify more than one nav attribute
+      if ((hasNavClick ? 1:0) + (hasNavHref ? 1:0) + (hasNavSref ? 1:0) > 1) {
+        throw Error(
+          'Must specify exactly one of md-nav-click, md-nav-href, ' +
```
clshortfuse: Since the user can pass 0 nav attributes, it doesn't seem this is the right wording. I would suggest `Must not specify more than one of...`

src/components/navBar/navBar.js
crisbeto: Yes I do, otherwise I'd have to do a check before adding it to the template.

clshortfuse: You perform a null check in line number 415

crisbeto:
Updated to include the feedback.

ThomasBurleson:
@crisbeto - darn it... I should have noticed this! Lol - nice catch.

PR #9819, opened 2016-10-12
feat(panel): Adding a hook on close success.

DerekLouie:
Hello,

@ErinCoughlan @bradrich 

Here is a small PR to add an "on close success" hook for users who want to run some code after a panel successfully closes.

It also adds an optional parameter to the "close" function which allows the user to specify a string describing why the close happened (standardized strings have been added to the "escapeToClose" and "clickOutsideToClose" functions. 

Please Review.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -126,6 +126,12 @@ angular
  *     outside the panel to close it. Defaults to false.
  *   - `escapeToClose` - `{boolean=}`: Whether the user can press escape to
  *     close the panel. Defaults to false.
+ *   - `onCloseSuccess` - `{function(!panelRef, string)=}`: Function that is
+ *     called after the close successfully finishes. The first parameter passed
+ *     into this function is the current panelRef and the 2nd is an optional
+ *     string explaining the close reason. The currently supported closeReasons
+ *     can be found in the MdPanelRef.closeReasons enum. These are by default
+ *     passed along by the panel.
  *   - `trapFocus` - `{boolean=}`: Whether focus should be trapped within the
  *     panel. If `trapFocus` is true, the user will not be able to interact
  *     with the rest of the page until the panel is dismissed. Defaults to
@@ -865,6 +871,18 @@ function MdPanelService($rootElement, $rootScope, $injector, $window) {
    * @type {enum}
    */
   this.interceptorTypes = MdPanelRef.interceptorTypes;
+
+  /**
+   * Possible values for closing of a panel.
+   * @type {enum}
+   */
+  this.closeReasons = MdPanelRef.closeReasons;
+
+  /**
+   * Possible values of absolute position.
+   * @type {enum}
+   */
+  this.absPosition = MdPanelPosition.absPosition;
 }
 
 
@@ -1153,20 +1171,24 @@ MdPanelRef.prototype.open = function() {
 
 /**
  * Closes the panel.
+ * @param {string} closeReason The event type that triggered the close.
  * @returns {!angular.$q.Promise<!MdPanelRef>} A promise that is resolved when
  *     the panel is closed and animations finish.
  */
-MdPanelRef.prototype.close = function() {
+MdPanelRef.prototype.close = function(closeReason) {
   var self = this;
 
   return this._$q(function(resolve, reject) {
     self._callInterceptors(MdPanelRef.interceptorTypes.CLOSE).then(function() {
       var done = self._done(resolve, self);
       var detach = self._simpleBind(self.detach, self);
+      var onCloseSuccess = self.config['onCloseSuccess'] || angular.noop;
+      onCloseSuccess = angular.bind(self, onCloseSuccess, self, closeReason);
 
       self.hide()
           .then(detach)
           .then(done)
+          .then(onCloseSuccess)
           .catch(reject);
     }, reject);
   });
@@ -1698,7 +1720,7 @@ MdPanelRef.prototype._configureEscapeToClose = function() {
         ev.stopPropagation();
         ev.preventDefault();
 
-        self.close();
+        self.close(MdPanelRef.closeReasons.ESCAPE);
       }
     };
 
@@ -1741,7 +1763,7 @@ MdPanelRef.prototype._configureClickOutsideToClose = function() {
         ev.stopPropagation();
         ev.preventDefault();
 
-        self.close();
+        self.close(MdPanelRef.closeReasons.CLICK_OUTSIDE);
       }
     };
 
@@ -2048,6 +2070,14 @@ MdPanelRef.prototype.removeFromGroup = function(groupName) {
   }
 };
 
+/**
+ * Possible default closeReasons for the close function.
+ * @enum {string}
+ */
+MdPanelRef.closeReasons = {
+  CLICK_OUTSIDE: 'clickOutsideToClose',
+  ESCAPE: 'escapeToClose',
+};
 
 /*****************************************************************************
  *                               MdPanelPosition                             *
--- a/src/components/panel/panel.spec.js
+++ b/src/components/panel/panel.spec.js
@@ -564,6 +564,75 @@ describe('$mdPanel', function() {
       expect(PANEL_EL).not.toExist();
     });
 
+    it('should call onCloseSuccess if provided after the panel finishes ' +
+        'closing', function() {
+          var closeReason, closePanelRef;
+          var onCloseSuccessCalled = false;
+
+          var onCloseSuccess = function(panelRef, reason) {
+            closePanelRef = panelRef;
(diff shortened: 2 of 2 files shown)


crisbeto:
@ErinCoughlan wouldn't this be more appropriate to pass to the `onClose` interceptor? That way you could use the reason for closing to determine whether to block it.

ErinCoughlan:
@crisbeto Interceptors intercept the close events, so they are run before close is finished and can block the close from happening. The goal of this PR is to allow the user to run events _after_ the panel is closed, similar to the animation hooks for `onDomRemoved`.

This already works for explicit close events because the user can chain off of the close call. But on pressing escape or clicking on the backdrop, we don't give the user any way to know these events happened or allow them to do work after it's closed. The most common use case of this is a dialog where when the panel is closed you need to reject/resolve the promise from open. This currently is not possible.

src/components/panel/panel.js
```diff
@@ -126,6 +126,9 @@ angular
  *     outside the panel to close it. Defaults to false.
  *   - `escapeToClose` - `{boolean=}`: Whether the user can press escape to
  *     close the panel. Defaults to false.
+ *   - `postClose` - `{function=}`: Function that is called after close
```
bradrich: I think naming this `onClose` would be more appropriate.

crisbeto:
Alright, that makes sense. Regarding the promise: didn't we have a task somewhere to have it be a part of the `mdPanel` API (returning a promise on init and then resolving/rejecting it)?

src/components/panel/panel.js
```diff
@@ -1153,20 +1156,24 @@ MdPanelRef.prototype.open = function() {
 
 /**
  * Closes the panel.
+ * @param {string} closeEvent The event type that triggered the close.
```
bradrich: This parameter would most likely not be a string. Events are normally `MouseEvent`, or `AngularEvent`.

src/components/panel/panel.spec.js
```diff
@@ -1043,6 +1043,7 @@ describe('$mdPanel', function() {
         clickPanelContainer();
 
         expect(myButton).toBeFocused();
+
```
bradrich: Why the extra lines here?

src/components/panel/panel.spec.js
```diff
@@ -1055,10 +1056,55 @@ describe('$mdPanel', function() {
         pressEscape();
 
         expect(myButton).toBeFocused();
+
       });
     });
   });
 
+  describe('postClose logic:', function() {
+      it('postClose when provided', function () {
```
bradrich: These tests should have more descriptive names. See the rest of the `it` tests, they all have the form of `should do something when something...`. These tests are failing in Travis and it wasn't immediately apparent without reading the tests what the test was trying to accomplish.

src/components/panel/panel.js
```diff
@@ -126,6 +126,9 @@ angular
  *     outside the panel to close it. Defaults to false.
  *   - `escapeToClose` - `{boolean=}`: Whether the user can press escape to
  *     close the panel. Defaults to false.
+ *   - `postClose` - `{function=}`: Function that is called after close
+ *     is done. The first parameter passed into this function is the closeEvent
+ *     and the 2nd is the current panelRef.
```
ErinCoughlan: I thought `panelRef` was first because that has to be defined and `closeEvent` is optional.

You should also mention the two events for `pressEscapeToClose` and `clickBackdropToClose` or no one will know how to use them.

src/components/panel/panel.js
ErinCoughlan: {function(panelRef, string)=} is the correct type, especially since this has params.

src/components/panel/panel.js
```diff
@@ -126,6 +126,9 @@ angular
  *     outside the panel to close it. Defaults to false.
  *   - `escapeToClose` - `{boolean=}`: Whether the user can press escape to
  *     close the panel. Defaults to false.
+ *   - `postClose` - `{function=}`: Function that is called after close
+ *     is done. The first parameter passed into this function is the closeEvent
```
ErinCoughlan: Possible re-word to "after close has successfully finished". "done" is too vague for this, since it will not be called if the panel fails the close.

src/components/panel/panel.js
ErinCoughlan: You should probably not mention "event" here, since that means something very different to most people, as @bradrich mentioned. Perhaps, closeReason, closeCause, or even just "reason" or "cause" would work.

src/components/panel/panel.js
```diff
@@ -1704,7 +1711,7 @@ MdPanelRef.prototype._configureEscapeToClose = function() {
         ev.stopPropagation();
         ev.preventDefault();
 
-        self.close();
+        self.close("escapeToClose");
```
ErinCoughlan: There should be an enum that the user can access for the different close causes (like the enum for positioning or animations).

String values are dangerous because it allows for typos.

src/components/panel/panel.spec.js
ErinCoughlan: Follow the same test naming as the other events we fire, like `onOpenComplete` or `onRemoving`.

src/components/panel/panel.spec.js
```diff
@@ -1055,10 +1056,55 @@ describe('$mdPanel', function() {
...
+        var obj = {
+          postClose: function() {},
```
ErinCoughlan: Why do you need to make an object here? Does it suffice to just create a spy, then pass it to the config?

```
var onCloseSuccess = jasmine.createSpy('onCloseSuccess');
var config = angular.extend({  'postClose': onCloseSuccess });

openPanel(config);
closePanel();

expect(onCloseSuccess).toHaveBeenCalledWith(panelRef, undefined);
```

DerekLouie:
Hello! All the aforementioned comments have been addressed, please take another look at this PR at your earliest convenience :)

src/components/panel/panel.js
bradrich: @ErinCoughlan Good catch! I was thinking that it was going to be the event that the occurred to get there, not a string representation.

bradrich:
LGTM

src/components/panel/panel.js
```diff
@@ -126,6 +126,12 @@ angular
  *     outside the panel to close it. Defaults to false.
  *   - `escapeToClose` - `{boolean=}`: Whether the user can press escape to
  *     close the panel. Defaults to false.
+ *   - `onCloseSuccess` - `{function(panelRef, string)=}`: Function that is
```
ErinCoughlan: !panelRef

src/components/panel/panel.js
```diff
@@ -126,6 +126,12 @@ angular
...
+ *     called after the close successfully finishes. The first parameter passed
+ *     into this function is the current panelRef and the 2nd is an optional
+ *     string explaining the close reason. Currently 'pressEscapeToClose' and
```
ErinCoughlan: Aren't these enums? that the user can get access to?

src/components/panel/panel.js
```diff
@@ -2054,6 +2064,14 @@ MdPanelRef.prototype.removeFromGroup = function(groupName) {
   }
 };
 
+/**
+ * Possible default closeReasons for the close function.
+ * @enum {string}
+ */
+MdPanelRef.closeReasons = {
```
ErinCoughlan: You'll want this to be exported on the mdPanel service as well so users can access them. (Follow what the other enums did.)

src/components/panel/panel.spec.js
```diff
@@ -1059,6 +1059,77 @@ describe('$mdPanel', function() {
     });
   });
 
+  describe('onCloseSuccess logic:', function() {
```
ErinCoughlan: I don't think these tests should be it's own group. This functionality is part of the close logic, and therefore should be individual tests after the other close callback tests.
 Like this:
https://github.com/angular/material/blob/master/src/components/panel/panel.spec.js#L877

src/components/panel/panel.spec.js
```diff
@@ -1059,6 +1059,77 @@ describe('$mdPanel', function() {
...
+            };
+
+            var config = angular.extend( {'onCloseSuccess': onCloseSuccess,
+                clickOutsideToClose: true, }, DEFAULT_CONFIG);
+
+            openPanel(config);
+            clickPanelContainer();
+
+            expect(onCloseSuccessCalled).toBe(true);
+            expect(closeReason).toBe('clickOutsideToClose');
```
ErinCoughlan: Use the enums in the tests too, not strings.

DerekLouie:
Fixed the enums / documentation comments. Please take another look at your earliest convenience :).

Issue #7963, opened 2016-04-09
a11y: dialog is always restoring focus.


DevVersion:
The dialog is currently always restoring the focus when it's closing.

``` js
      function detachAndClean() {
        angular.element($document[0].body).removeClass('md-dialog-is-showing');
        element.remove();

        if (!options.$destroy) options.origin.focus();
      }
```

This is a weird behavior, when clicking for example on a list-item which opens a dialog as an action.
So if the dialog is closing, then the focus will be restored back to the button inside the list-item, which triggered the `ngClick` action.

This causes the button, to have a focus effect, which shouldn't be. We're only showing the focus effect if there was no mouse click the last `100ms`.

In my opinion, we should have different accessibility actions, when the last interaction was from a keyboard / mouse or touch.

We currently have a PR pending, which adds the required functionality. (#7965) 
It was an old PR, and I reworked it again. 

This PR would allow fixes for:
- The dialog focus restore with a list-item (#7960)
- Focus Effect for Switches (#7837)
- Button Shows Focus when Switching browser tabs

cc. @marcysutton @ThomasBurleson @EladBezalel

ThomasBurleson:
Fixed by #7965

DevVersion:
Status Update: Pull Request is in and solved the Sidenav issues (as intended) but now I can move on and use the new method for the Button and switches as well.

PR #9826, opened 2016-10-12
fix(button): properly show focus effect

### LINKED ISSUE #8749: Md-button with target blank stays focused after click
**Actual behavior**:
-  What is the issue?

When `md-button` is used with `target="_blank"` it stays focused after click. Using `md-tooltip` with it and hovering after click makes it visible until we click outside the button.

`md-button` should lose focus after click and tooltip should not be visible when button is not hovered.

**CodePen** or Steps to reproduce the issue:
-  CodePen Demo which shows your issue : http://codepen.io/anon/pen/WxwpwE
-  Details: Similar behavior reported in kubernetes/dashboard#877

**Angular Versions**: 
-  Angular Version: 1.5.3
-  Angular Material Version: 1.1.0-rc2

**Additional Information**:
-  Browser Type: Chrome/Firefox
-  Browser Version: Latest
-  OS: Ubuntu 14.04

DevVersion:
* Only apply the focus effect for keyboard interaction (as same as it does currently, but not with a random timeout).

Similar to #9827 

 Fixes #8749. References #7963

### DIFF
--- a/src/components/button/button.js
+++ b/src/components/button/button.js
@@ -114,7 +114,7 @@ function MdAnchorDirective($mdTheming) {
  *  </md-button>
  * </hljs>
  */
-function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
+function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $mdInteraction) {
 
   return {
     restrict: 'EA',
@@ -162,20 +162,17 @@ function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
     });
 
     if (!element.hasClass('md-no-focus')) {
-      // restrict focus styles to the keyboard
-      scope.mouseActive = false;
-      element.on('mousedown', function() {
-        scope.mouseActive = true;
-        $timeout(function(){
-          scope.mouseActive = false;
-        }, 100);
-      })
-      .on('focus', function() {
-        if (scope.mouseActive === false) {
+
+      element.on('focus', function() {
+
+        // Only show the focus effect when being focused through keyboard interaction or programmatically
+        if (!$mdInteraction.isUserInvoked() || $mdInteraction.getLastInteractionType() === 'keyboard') {
           element.addClass('md-focused');
         }
-      })
-      .on('blur', function(ev) {
+
+      });
+
+      element.on('blur', function() {
         element.removeClass('md-focused');
       });
     }
--- a/src/components/button/button.spec.js
+++ b/src/components/button/button.spec.js
@@ -59,20 +59,52 @@ describe('md-button', function() {
     expect(button.hasClass('md-button')).toBe(true);
   }));
 
-  it('should not set focus state on mousedown', inject(function ($compile, $rootScope){
+  it('should apply focus effect with keyboard interaction', inject(function ($compile, $rootScope){
     var button = $compile('<md-button>')($rootScope.$new());
+    var body = angular.element(document.body);
+
     $rootScope.$apply();
-    button.triggerHandler('mousedown');
-    expect(button[0]).not.toHaveClass('md-focused');
+
+    // Fake a keyboard interaction for the $mdInteraction service.
+    body.triggerHandler('keydown');
+    button.triggerHandler('focus');
+
+    expect(button).toHaveClass('md-focused');
+
+    button.triggerHandler('blur');
+
+    expect(button).not.toHaveClass('md-focused');
   }));
 
-  it('should set focus state on focus and remove on blur', inject(function ($compile, $rootScope){
+  it('should apply focus effect when programmatically focusing', inject(function ($compile, $rootScope){
     var button = $compile('<md-button>')($rootScope.$new());
+
     $rootScope.$apply();
+
     button.triggerHandler('focus');
-    expect(button[0]).toHaveClass('md-focused');
+
+    expect(button).toHaveClass('md-focused');
+
     button.triggerHandler('blur');
-    expect(button[0]).not.toHaveClass('md-focused');
+
+    expect(button).not.toHaveClass('md-focused');
+  }));
+
+  it('should not apply focus effect with mouse interaction', inject(function ($compile, $rootScope){
+    var button = $compile('<md-button>')($rootScope.$new());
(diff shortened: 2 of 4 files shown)


src/components/button/button.js
```diff
@@ -162,20 +162,15 @@ function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
...
+      element.on('focus', function() {
+        // Only show the focus effect when being focused through keyboard interaction.
+        if ($mdInteraction.getLastInteractionType() === 'keyboard') {
           element.addClass('md-focused');
         }
-      })
-      .on('blur', function(ev) {
+      });
+
+      element.on('blur', function() {
```
DevVersion: Note: This is intentionally now a new statement, because otherwise the indention is ugly and this is just more clear..

src/components/button/button.js
```diff
@@ -162,20 +162,15 @@ function MdButtonDirective($mdButtonInkRipple, $mdTheming, $mdAria, $timeout) {
...
-        $timeout(function(){
-          scope.mouseActive = false;
-        }, 100);
-      })
-      .on('focus', function() {
-        if (scope.mouseActive === false) {
+
+      element.on('focus', function() {
+        // Only show the focus effect when being focused through keyboard interaction.
+        if ($mdInteraction.getLastInteractionType() === 'keyboard') {
```
clshortfuse: Wouldn't this change the existing interaction if a developer focuses an element over Javascript (such as when opening a dialog or a custom component)?

DevVersion: This does not change anything to the existing behavior, just that we now check the global interaction to detect the last interaction.

Focusing manually per `.focus` does not work right now as well, because the `mousedown` event didn't get fired.

This is the approach we wanted to go with the `$mdInteraction` service, as proposed by Marcy. Similar to #9827

clshortfuse: I just some testing and `mousedown` gets fired after `focus` on Chrome. Instead of tracking mouseactive as a `boolean`, I would suggest a timestamp on last mousedown event. Then attach a timeout to the adding of `md-focused`. If mousedown was triggered within a certain time (say 100ms), don't add the class.

clshortfuse: I've had pretty good success with this change I was testing: 
```
if (!element.hasClass('md-no-focus')) {
  // prevent focus on mousedown
  scope.lastMouseDown = null;
  element.on('mousedown', function(event) {
    scope.lastMouseDown = $mdUtil.now();
  })
  .on('focus', function(event) {
    scope.focusTimeout = $timeout(function() {
      if (scope.lastMouseDown && ($mdUtil.now() - scope.lastMouseDown < 500)) {
        return;
      }
    }, 100);
  })
  .on('blur', function(ev) {
    if (scope.focusTimeout) {
      $timeout.cancel(scope.focusTimeout);
      scope.focusTimeout = null;
    }
    element.removeClass('md-focused');
  });
}
```

DevVersion:
Going to mark this as in progress now, as we have some discussions about it.

src/components/checkbox/checkbox.js
```diff
@@ -54,7 +54,7 @@ angular
  * </hljs>
  *
  */
-function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $mdUtil, $timeout) {
+function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $mdUtil, $timeout, $mdInteraction) {
```
crisbeto: Seems like the `$mdInteraction` import is unused.

src/core/services/interaction/interaction.js
```diff
@@ -129,4 +132,18 @@ MdInteractionService.prototype.onBufferInputEvent = function(event) {
...
+
+/**
+ * @ngdoc method
+ * @name $mdInteraction#isProgrammatic
+ * @description Method to detect whether any interaction happened recently.
+ * @param {number=} checkDelay Time to check for any interaction to have been triggered.
+ * @returns {boolean} Whether there was any interaction or not.
+ */
+MdInteractionService.prototype.isProgrammatic = function(checkDelay) {
+  var delay = angular.isDefined(checkDelay) ? checkDelay : 15;
```
crisbeto: `angular.isNumber` might be more appropriate here.

src/core/services/interaction/interaction.js
```diff
@@ -129,4 +132,18 @@ MdInteractionService.prototype.onBufferInputEvent = function(event) {
...
+};
+
+/**
+ * @ngdoc method
+ * @name $mdInteraction#isProgrammatic
+ * @description Method to detect whether any interaction happened recently.
+ * @param {number=} checkDelay Time to check for any interaction to have been triggered.
+ * @returns {boolean} Whether there was any interaction or not.
+ */
+MdInteractionService.prototype.isProgrammatic = function(checkDelay) {
```
crisbeto: I know that we decided on the naming for this together, but I'm still not sure that `isProgrammatic` is very obvious. It seems like it implies that the service is programmatic. That being said, I don't have any better suggestions at the moment, perhaps other team member can pitch in?

DevVersion: Yeah, Carlos had a great idea here. - `isUserInvoked`.

src/components/button/button.spec.js
```diff
@@ -66,12 +66,51 @@ describe('md-button', function() {
...
+  }));
+
+  it('should apply focus effect when programmatically focusing', inject(function ($compile, $rootScope){
+    var button = $compile('<md-button>')($rootScope.$new());
+
+    $rootScope.$apply();
+
+    button.triggerHandler('focus');
+
+    expect(button[0]).toHaveClass('md-focused');
```
crisbeto: AFAIK you don't have to fetch the DOM node here. `expect(button).toHaveClass('md-focused')` should work as well.

DevVersion: Valid and changed it.. But just for your information, a lot of other tests are doing the same and I don't want to diff the whole test now.

crisbeto: That's why I didn't make it a blocking request. It's basically the same, it just looks nicer.

DevVersion: Yeah, changed it for the tests I've added.

PR #9827, opened 2016-10-12
fix(checkbox): properly show focus effect

DevVersion:
* Right now (as same as with the button) the focus will be automatically restored when switching the browser tabs (this is bad UX).

* Using the new $mdInteraction service could remove the random $timeout and just fix this issue easily.

References #7963. References #8749.


ThomasBurleson:
👍  nice work!

Issue #9675, opened 2016-09-22
dialog: conversion to $mdPanel


DevVersion:
Also going to create an issue for the `$mdDialog` conversion, since I didn't create one yet.

PR #9829, opened 2016-10-13
feat(panel): add contentElement option

### LINKED ISSUE #9757: mdPanel - integrate with $mdCompiler changes
* Recently we made some refactoring on the `$mdCompiler` and it now supports `contentElement`s

---
#### Task 1
Rather than manually removing the element from the DOM, the panel should use the `cleanup` method, to integrate with the `contentElement` functionality.

```js
const compileData = $mdCompiler.compile(...);

/* Removes or restores the element */
compileData.cleanup();
```

#### Task 2

To properly support `contentElement`s, the transition classes need to be reverted before calling the `cleanup` function.

You can either
- Manually remove the transition classes
- Reversely run the `$animate` runner.

#### Task 3

Add documentation to the panel, to let users know how they can use a `contentElement`.

---

This currently blocks the port of the dialog to the panel (#9675).

crisbeto:
Hooks up the `contentElement` option from `$mdCompile` to `$mdPanel`, allowing for a pre-compiled element to be used as the panel's content.

Relates to #9675.
Fixes #9757.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -93,6 +93,8 @@ angular
  *     [$sce service](https://docs.angularjs.org/api/ng/service/$sce).
  *   - `templateUrl` - `{string=}`: The URL that will be used as the content of
  *     the panel.
+ *   - `contentElement` - `{(string|!angular.JQLite|!Element)=}`: Pre-compiled
+ *     element to be used as the panel's content.
  *   - `controller` - `{(function|string)=}`: The controller to associate with
  *     the panel. The controller can inject a reference to the returned
  *     panelRef, which allows the panel to be closed, hidden, and shown. Any
@@ -1028,6 +1030,24 @@ MdPanelService.prototype._wrapTemplate = function(origTemplate) {
 };
 
 
+/**
+ * Wraps a content element in a md-panel-outer wrapper and
+ * positions it off-screen. Allows for proper control over positoning
+ * and animations.
+ * @param {!angular.JQLite} contentElement Element to be wrapped.
+ * @return {!angular.JQLite} Wrapper element.
+ * @private
+ */
+MdPanelService.prototype._wrapContentElement = function(contentElement) {
+  var wrapper = angular.element('<div class="md-panel-outer-wrapper">');
+
+  contentElement.addClass('md-panel').css('left', '-9999px');
+  wrapper.append(contentElement);
+
+  return wrapper;
+};
+
+
 /*****************************************************************************
  *                                 MdPanelRef                                *
  *****************************************************************************/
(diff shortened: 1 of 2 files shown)


crisbeto:
I don't think adding the docs to the dialog make sense at the moment since it's not using `$mdPanel` yet.

src/components/panel/panel.js
```diff
@@ -1457,6 +1505,51 @@ MdPanelRef.prototype.toggleClass = function(toggleClass, onElement) {
...
+ * the DOM. Helps normalize differences in the compilation process between
+ * using a string template and a content element.
+ * @returns {!angular.$q.Promise<!MdPanelRef>} Promise that is resolved when
+ *     the element has been compiled and added to the DOM.
+ * @private
+ */
+MdPanelRef.prototype._compile = function() {
+  var self = this;
+
+  // Compile the element via $mdCompiler. Note that whent using a
```
ErinCoughlan: typo: whent -> when

src/components/panel/panel.spec.js
```diff
@@ -2713,6 +2713,70 @@ describe('$mdPanel', function() {
...
+        contentElement: angular.element('<div>'),
+        position: $mdPanel.newPanelPosition().absolute().center()
+      };
+    });
+
+    it('should wrap the content element in the proper HTML and assign ' +
+      'the wrapper to the panel reference', function() {
+        openPanel(config);
+
+        expect(panelRef.panelEl.parent()).toHaveClass(PANEL_WRAPPER_CLASS.substring(1));
```
ErinCoughlan: I think I would prefer to fix the variables over using substrings like this. It's confusing from the variable name that you're just trying to get rid of the "." What do you think about the following?

The _CLASS should just be the class names, not the selectors. Then the not _CLASS versions should be the selectors. I think that flows better with the expects anyways. ```expect(PANEL_WRAPPER_CLASS).toHaveClass(HIDDEN_CLASS)``` doesn't make sense to me.

So,
```
var PANEL_WRAPPER = ".md-panel-outer-wrapper";
var PANEL_WRAPPER_CLASS = "md-panel-outer-wrapper";
```

ErinCoughlan: I'd suggest a new line before the `.toHaveClass` in this case to help with readability.

crisbeto: Yes that sounds good. My worry was ending up with multiple constants for the same class selector. I agree that the substring approach is kind of dirty.

bradrich:
LGTM. Very clean approach, easy to read. Good job!

src/components/panel/panel.js
```diff
@@ -1117,6 +1137,26 @@ function MdPanelRef(config, $injector) {
...
+   * restore the element to it's proper place in the DOM.
+   * @private {!Function}
+   */
+  this._compilerCleanup = null;
+
+  /**
+   * Used for saving and restoring an element's inline styles.
+   * @private {string}
+   */
+  this._restoreStyles = '';
```
ThomasBurleson: Personally I would prefer `this._cache = { styles : {}, classes: [ ] };`

crisbeto:
Updated to include the feedback.

src/components/panel/panel.js
```diff
@@ -1117,6 +1137,23 @@ function MdPanelRef(config, $injector) {
...
+   * Cleanup function, provided by `$mdCompiler` and assigned after the element
+   * has been compiled. When `contentElement` is used, the function is used to
+   * restore the element to it's proper place in the DOM.
+   * @private {!Function}
+   */
+  this._compilerCleanup = null;
+
+  /**
+   * Cache for saving and restoring element inline styles, CSS classes etc.
+   * @type {!Object}
```
ErinCoughlan: @type {{styles: string, classes: string}}

PR #9830, opened 2016-10-13
feat(panel): Allow panels to be part of multiple groups.

### LINKED ISSUE #9565: mdPanel: Allow panels to be part of multiple groups.
Followup of #8971 and #9538, which add basic group functionality.

Example use case: Menus within a Dialog. The menu should close when the dialog closes, so they need to be part of the same group. However, the menu can also be part of a menu group, which has it's own maxOpen and configuration options.

bradrich:
The groupName parameter has been changed from allowing only a string to allowing a string or an array of strings. All of the necessary functionality has been changed to accompany the array.

Fixes #9565

Ping @ErinCoughlan @crisbeto and @DerekLouie for review.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -14,6 +14,7 @@ angular
  *                            PUBLIC DOCUMENTATION                           *
  *****************************************************************************/
 
+
 /**
  * @ngdoc service
  * @name $mdPanel
@@ -155,10 +156,11 @@ angular
  *     on when the panel closes. This is commonly the element which triggered
  *     the opening of the panel. If you do not use `origin`, you need to control
  *     the focus manually.
- *   - `groupName` - `{string=}`: Name of a panel group. This group name is used
- *     for configuring the number of open panels and identifying specific
- *     behaviors for groups. For instance, all tooltips could be identified
- *     using the same groupName.
+ *   - `groupName` - `{(string|!Array<string>)=}`: A group name or an array of
+ *     group names. The group name is used for creating a group of panels. The
+ *     group is used for configuring the number of open panels and identifying
+ *     specific behaviors for groups. For instance, all tooltips could be
+ *     identified using the same groupName.
  *
  * @returns {!MdPanelRef} panelRef
  */
@@ -685,6 +687,7 @@ angular
  *                               MdPanelAnimation                            *
  *****************************************************************************/
 
+
 /**
  * @ngdoc type
  * @name MdPanelAnimation
@@ -895,7 +898,12 @@ MdPanelService.prototype.create = function(config) {
   this._config.scope.$on('$destroy', angular.bind(panelRef, panelRef.detach));
 
   if (this._config.groupName) {
-    panelRef.addToGroup(this._config.groupName);
+    if (angular.isString(this._config.groupName)) {
+      this._config.groupName = [this._config.groupName];
+    }
+    angular.forEach(this._config.groupName, function(group) {
+      panelRef.addToGroup(group);
+    });
   }
 
   return panelRef;
@@ -987,7 +995,7 @@ MdPanelService.prototype.setGroupMaxOpen = function(groupName, maxOpen) {
  * @private
  */
 MdPanelService.prototype._openCountExceedsMaxOpen = function(groupName) {
-  if (groupName && this._groups[groupName]) {
+  if (this._groups[groupName]) {
     var group = this._groups[groupName];
     return group.maxOpen > 0 && group.openPanels.length > group.maxOpen;
   }
@@ -1001,9 +1009,7 @@ MdPanelService.prototype._openCountExceedsMaxOpen = function(groupName) {
  * @private
  */
 MdPanelService.prototype._closeFirstOpenedPanel = function(groupName) {
-  if (groupName) {
-    this._groups[groupName].openPanels[0].close();
-  }
+  this._groups[groupName].openPanels[0].close();
 };
 
 
@@ -1137,8 +1143,12 @@ MdPanelRef.prototype.open = function() {
     var done = self._done(resolve, self);
     var show = self._simpleBind(self.show, self);
     var checkGroupMaxOpen = function() {
-      if (self._$mdPanel._openCountExceedsMaxOpen(self.config.groupName)) {
-        self._$mdPanel._closeFirstOpenedPanel(self.config.groupName);
+      if (self.config.groupName) {
+        angular.forEach(self.config.groupName, function(group) {
+          if (self._$mdPanel._openCountExceedsMaxOpen(group)) {
+            self._$mdPanel._closeFirstOpenedPanel(group);
+          }
+        });
       }
     };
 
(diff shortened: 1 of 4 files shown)


src/components/panel/panel.js
```diff
@@ -895,7 +898,12 @@ MdPanelService.prototype.create = function(config) {
   this._config.scope.$on('$destroy', angular.bind(panelRef, panelRef.detach));
 
   if (this._config.groupName) {
-    panelRef.addToGroup(this._config.groupName);
+    if (angular.isString(this._config.groupName)) {
+      this._config.groupName = [this._config.groupName];
```
crisbeto: It seems like the `_groupName` got converted into an array. I think it's more appropriate to call it something like `_groups` since an array implies multiple groups.

src/components/panel/panel.js
```diff
@@ -987,7 +995,7 @@ MdPanelService.prototype.setGroupMaxOpen = function(groupName, maxOpen) {
  * @private
  */
 MdPanelService.prototype._openCountExceedsMaxOpen = function(groupName) {
-  if (groupName && this._groups[groupName]) {
+  if (this._groups[groupName]) {
```
crisbeto: Using `indexOf` here should be more appropriate. This will break with a group name of `toString`.

src/components/panel/panel.js
```diff
@@ -1001,9 +1009,7 @@ MdPanelService.prototype._openCountExceedsMaxOpen = function(groupName) {
  * @private
  */
 MdPanelService.prototype._closeFirstOpenedPanel = function(groupName) {
-  if (groupName) {
-    this._groups[groupName].openPanels[0].close();
-  }
+  this._groups[groupName].openPanels[0].close();
```
crisbeto: Might be appropriate to have a check here that there are any open panels.

src/components/panel/panel.js
```diff
@@ -155,10 +156,11 @@ angular
  *     on when the panel closes. This is commonly the element which triggered
  *     the opening of the panel. If you do not use `origin`, you need to control
  *     the focus manually.
- *   - `groupName` - `{string=}`: Name of a panel group. This group name is used
- *     for configuring the number of open panels and identifying specific
- *     behaviors for groups. For instance, all tooltips could be identified
- *     using the same groupName.
+ *   - `groupName` - `{(string|string[])=}`: A single name of a panel group. Can
```
crisbeto: nit: You're saying "a single name of a group" and "can also be an array" in two consecutive sentences. Might be better to say something along the lines of "A group name or an array of groups".

ThomasBurleson:
When you ping for review, please mark as "needs: review" and "in-progress".

src/components/panel/panel.js
bradrich: The `groupName` here refers to the parameter that the user passes from the panel configuration. Since it can be either a single groupName or an array of groupNames, I was concerned about changing its name.

@ErinCoughlan what do you think about the naming convention here? Let's get the whole team's buy-in.

src/components/panel/panel.js
bradrich: This function is actually only fired from the `_openCountExceedsMaxOpen` method above it. That method checks if the current number of open panels within the group exceeds the maximum limit and if so, the `_closeFirstOpenedPanel` method is called to close the first opened panel.

bradrich:
@ErinCoughlan @crisbeto @DerekLouie Please review changes and provide feedback or LGTM.

src/components/panel/panel.js
crisbeto: Oh, sorry, I must've misread it as `this._groupName`.

bradrich:
@ErinCoughlan Waiting on your final review for this PR.

src/components/panel/panel.js
```diff
@@ -155,10 +156,11 @@ angular
  *     on when the panel closes. This is commonly the element which triggered
  *     the opening of the panel. If you do not use `origin`, you need to control
  *     the focus manually.
- *   - `groupName` - `{string=}`: Name of a panel group. This group name is used
- *     for configuring the number of open panels and identifying specific
- *     behaviors for groups. For instance, all tooltips could be identified
- *     using the same groupName.
+ *   - `groupName` - `{(string|string[])=}`: A group name or an array of group
```
ErinCoughlan: !Array<string>

src/components/panel/panel.js
ErinCoughlan: I think `groupName` works here.

src/components/panel/panel.spec.js
```diff
@@ -1139,18 +1164,38 @@ describe('$mdPanel', function() {
           expect(getGroupPanels('test')).not.toContain(panel);
         });
 
-    it('should remove a panel from a group on panel destroy', function() {
```
ErinCoughlan: Why are you removing the single group tests? The config is different, so I think you still need a test or two here.

src/components/panel/panel.spec.js
```diff
@@ -1194,6 +1239,26 @@ describe('$mdPanel', function() {
...
+            groupName: ['test1', 'test2']
+          };
+
+          openPanel(config);
+          flushPanel();
+          expect(getGroupOpenPanels('test1')).toContain(panelRef);
+          expect(getGroupOpenPanels('test2')).toContain(panelRef);
+
+          closePanel();
+          flushPanel();
```
ErinCoughlan: I thought there was a `flushPanel` at the end of the `closePanel`. Why do you need another one?

src/components/panel/panel.spec.js
```diff
@@ -1222,6 +1287,45 @@ describe('$mdPanel', function() {
...
           panel3.close();
         });
+
+    it('should close the first open panel of any group that the panel is in ' +
+        'when more than the maxium number of panels is opened', function() {
+          $mdPanel.newPanelGroup('test1', {
+            maxOpen: 2
+          });
+          $mdPanel.newPanelGroup('test2', {
+            maxOpen: 2
```
ErinCoughlan: Please choose different `maxOpen` for each group so we can also verify the maxes are being set of the right panels.

src/components/panel/panel.spec.js
```diff
@@ -1222,6 +1287,45 @@ describe('$mdPanel', function() {
...
+          var config1 = {
+            groupName: ['test1', 'test2']
+          };
+          var config2 = {
+            groupName: 'test1'
+          };
+
+          var panel1 = $mdPanel.create(config1);
+          var panel2 = $mdPanel.create(config1);
+
```
ErinCoughlan: Why the extra new line here?

bradrich:
@ErinCoughlan Ping to approve your latest review.

src/components/panel/panel.spec.js
ErinCoughlan: It's still here? And there's no response to the question?

src/components/panel/panel.spec.js
```diff
@@ -1222,6 +1299,69 @@ describe('$mdPanel', function() {
...
+          var config3 = {
+            groupName: 'test2'
+          };
+
+          var panel1 = $mdPanel.create(config1);
+          var panel2 = $mdPanel.create(config1);
+          var panel3 = $mdPanel.create(config2);
+          var panel4 = $mdPanel.create(config3);
+          var panel5 = $mdPanel.create(config3);
+          var panel6 = $mdPanel.create(config3);
```
ErinCoughlan: This just got way too complicated. How about one group with maxOpen = 1 and one group with maxOpen =2? I don't understand why you need 6 panels to test this, even with the current values.

It would also be useful to name these related to what their config is. Below when you say panel4, I have no idea which config that is using, so it's hard to understand what you are testing and even harder to know whether it is correct.

bradrich:
@ErinCoughlan Latest ping!

bradrich:
@ThomasBurleson Final LGTM has been given.

PR #9855, opened 2016-10-19
change margin of md-input-container to md-autocomplete

NgaiKaKit:
When applying md-floating-label to md-autocomplete,
the vertical margin 18px of md-input-container apply inside the md-autocomplete element.
If there are a md-autocomplete following a normal md-input-container, the space between will be 36px.
But I believe it should be same behaviour of multiple md-input-container in a form, which the space bewteen is 18px;

### DIFF
--- a/src/components/autocomplete/autocomplete.scss
+++ b/src/components/autocomplete/autocomplete.scss
@@ -46,11 +46,13 @@ md-autocomplete {
   }
   &[md-floating-label] {
     border-radius: 0;
+    margin: 18px 0px;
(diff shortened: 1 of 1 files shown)


DevVersion:
Thanks for the PR, but I don't think we can make such a change without any discussion.

The margin not being set on the autocomplete, seems to be intentionally and personally makes sense for me.

* Because it does not support any message spacers
* The label does not float up (so no need for extra space)

Please open an issue first, and then we can discuss about this.

PR #9856, opened 2016-10-19
fix(toolbar): scroll shrink applied to inner md-content

myagoo:
I have a md-select nested within my toolbar which has scroll shrink activated.
When onChangeScrollShrink is called, closestContent contains the md-select's md-content, which is later moved to the body, but it's already too late, scroll listener have already been applied :/

### DIFF
--- a/src/components/toolbar/toolbar.js
+++ b/src/components/toolbar/toolbar.js
@@ -122,7 +122,7 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
          *
          */
         function onChangeScrollShrink(shrinkWithScroll) {
-          var closestContent = element.siblings('md-content');
+          var closestContent = element.parent().children('md-content');
 
           // If we have a content element, fake the call; this might still fail
           // if the content element isn't a sibling of the toolbar
--- a/src/components/toolbar/toolbar.js
+++ b/src/components/toolbar/toolbar.js
@@ -122,7 +122,7 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
          *
          */
         function onChangeScrollShrink(shrinkWithScroll) {
-          var closestContent = element.parent().find('md-content');
+          var closestContent = element.siblings('md-content');
 
           // If we have a content element, fake the call; this might still fail
           // if the content element isn't a sibling of the toolbar
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -525,9 +525,38 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
         parent.$$childHead = parent.$$childTail = child;
       }
     },
+    
+    /**
+     * Get an element direct children matching a given tag name
+     *
+     * @param element Element to start walking the DOM from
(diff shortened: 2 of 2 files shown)


ThomasBurleson:
@DevVersion, @crisbeto - plz investigate.

DevVersion:
Just investigated a bit, and I can see the issue you experience.

_Just to summarize a few things_

* `md-select` creates its own `md-content` element for the dropdown
* The toolbar accidentally targets the `md-content` from the `md-select`.

> Only supporting the immediate children makes sense for me (but needs a breaking change note and unit tests)

src/components/toolbar/toolbar.js
```diff
@@ -122,7 +122,7 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
          *
          */
         function onChangeScrollShrink(shrinkWithScroll) {
-          var closestContent = element.parent().find('md-content');
+          var closestContent = element.parent().children('md-content');
```
DevVersion: Angular's jqLite does not support `.children` with a selector - [See here](https://docs.angularjs.org/api/ng/function/angular.element)

myagoo:
Is it Ok to use classic dom traversal ?

DevVersion:
@myagoo Yeah that should be fine here. I can't think of any easy way to do this right now. @crisbeto any thoughts?

```js
// Get Child Nodes and cast to normal Array (not NodeList)
[].slice
  .call(element.parent().children())
// Chain to filter for specific DOM types (can also use toLowerCase etc.)
  .filter(function(node) { return node.tagName === 'MD-LIST-ITEM'; }); 
```

ThomasBurleson:
@DevVersion - encapsulate into distinct function call plz

crisbeto:
Wouldn't `element.parent()[0].querySelectorAll('> md-list-item')` work?

DevVersion:
@ThomasBurleson Yeah I'd like to have a generic `$mdUtil` function for this (I hit this issues a few times) 
@crisbeto Unfortunately this does not work. The browser will throw an error 
> Failed to execute 'querySelectorAll' on 'Element': '> md-list-item' is not a valid selector.

myagoo:
I'll be more than pleased to s
implement this function if you tell me the API you want

DevVersion:
I personally would have done something like

```js
// Not type of a NodeList -> normal Array
var children = $mdUtil.getChildren(element, 'md-content');
```

Not really sure about the name.

crisbeto:
In that case your approach should be the best @DevVersion. I would rewrite it like this though, it avoids having to cast it to a proper array first:

```js
Array.prototype.filter.call(element[0].parentNode.children, function(node) { 
  return node.tagName === 'MD-LIST-ITEM'; 
});
```

myagoo:
I assume inputed element and ouputed children are jqlite wrappers ?

myagoo:
Ho sorry, did not read the comment... (What a sad thing to say)

DevVersion:
@myagoo No problems and with the snippet from crisbeto you get DOM nodes 😄 @crisbeto Yeah looks good.

myagoo:
I recreated this Pull Request here #9871 Could you delete this one ?

PR #9870, opened 2016-10-20
feat(md-chips): expose event to onRemove function

### LINKED ISSUE #8847: chips: md-chip-remove and $event.stopPropagation() not working
Hello, actually working a lot with md-chip which i like pretty much. I encounter several problems. Most of time because the doc is not complete. This time I couldn't find a working solutions in any other place.

my code :

```
<button class="md-chip-remove"
    md-chip-remove
    ng-click="$event.stopPropagation(); ctrl.function($chip);">
    // ng-click="ctrl.function($chip, $event);">
    <md-icon md-svg-icon="md-close"></md-icon>
 </button>
```

I use sweetAlert when the user click on the close 'x' of a chip when he wants to remove it. He can click on cancel or confirm the suppression.
the problem is, $event.stopPropagation(); doesn't stop the propagation of the event and remove the md-chip in every cases. Even before the user can click on cancel or confirm.

The only thing I could do is to add again the removed chip to my list of chips if the user clicked on cancel. But that not something i ll be satisfied with.

we got :
md-on-add   expression  
An expression which will be called when a chip has been added.

md-on-remove    expression  
An expression which will be called when a chip has been removed.

But we got nothing that is called so we could stop the add or remove action.
something like : md-on-click boolean
which will do the action only when the boolean is true and will wait if it's false.
it's a kind of delay.

But all this wouldn't be necessary if we could stop the propagation on the remove :)

gseabrook:
Expose event object as parameter to onRemove function to allow for stopping propagation

Closes #8847

### DIFF
--- a/src/components/chips/js/chipsDirective.js
+++ b/src/components/chips/js/chipsDirective.js
@@ -108,7 +108,7 @@
    * @param {expression=} md-on-add An expression which will be called when a chip has been
    *    added.
    * @param {expression=} md-on-remove An expression which will be called when a chip has been
-   *    removed.
+   *    removed with $chip, $index and $event available as parameters.
    * @param {expression=} md-on-select An expression which will be called when a chip is selected.
    * @param {boolean} md-require-match If true, and the chips template contains an autocomplete,
    *    only allow selection of pre-defined chips (i.e. you cannot add new ones).
@@ -190,7 +190,7 @@
       <button\
           class="md-chip-remove"\
           ng-if="$mdChipsCtrl.isRemovable()"\
-          ng-click="$mdChipsCtrl.removeChipAndFocusInput($$replacedScope.$index)"\
+          ng-click="$mdChipsCtrl.removeChipAndFocusInput($$replacedScope.$index, $event)"\
           type="button"\
           aria-hidden="true"\
           tabindex="-1">\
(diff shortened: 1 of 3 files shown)


gseabrook:
Apologies for the delay, I was away for a week. I've made the requested changes and merged into the previous commit.

The 'db' variable came from an earlier test that used the same button, so I've also renamed the variable in both tests to chip button

gseabrook:
@DevVersion I've made the requested changes, do I need to do something to change the PR status?

fschwalm:
Any news about the progress of this feature? 

Thanks!

src/components/chips/js/chipsDirective.js
```diff
@@ -108,7 +108,7 @@
    * @param {expression=} md-on-add An expression which will be called when a chip has been
    *    added.
    * @param {expression=} md-on-remove An expression which will be called when a chip has been
-   *    removed.
+   *    removed with $chip, $index and $event available as parameters.
```
Splaktar: Can you please wrap these parameter names with backticks? It will help make the docs more clear.

Splaktar:
This PR is quite old and needs rebasing. I don't want to bother you with it after so long, so I'll rebase it, make the minor docs changes and create a new PR. It'll still preserve your commit though 😄 Thank you very much for the contribution!

Splaktar:
Continuing this in https://github.com/angular/material/pull/11192.

Issue #9509, opened 2016-09-03
$mdUtil getClosest() failed to find nearest <svg /> node


djfd:
**Actual Behavior**:
 -  `What is the issue? *`Enclosing <SVG /> node not found
 -  `What is the expected behavior?`expected to be found

**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:` http://codepen.io/djFD/pen/qaBrbL (please check console outpiut)
 -  `Details:` Tested in two browsers, firefox and chromium, all SVG internal nodes (including svg itself) have their nodeNames in lowercase (HTML nodenames are in uppercase), and there is a code https://github.com/angular/material/blob/master/src/core/util/util.js#L511 that converts expected parent node to uppercase. 

However nodes' localNames are in lowercase, maybe need to check lowercased target against of localName property?

**Angular Versions**: *
 -  `Angular Version:`1.5.7
 -  `Angular Material Version:`1.1.0

**Additional Information**:
 -  `Browser Type: *`firefox/chromium
 -  `Browser Version: *`48.0/51.0.2704.79
 -  `OS: *`linux
 -  `Stack Traces:`--

crisbeto:
Nice catch, I'll patch the method in Material. Until it gets released, you can work around it yourself by passing in a function to `getClosest`:

```js
$mdUtil.getClosest(el, function(parent) {
  return parent.nodeName.toLowerCase() === 'lower-case-name-goes-here';
}, true))
```

PR #9871, opened 2016-10-20
fix(toolbar): scroll shrink is applied to md-content within toolbar

myagoo:
I have a `md-select` nested within my `md-toolbar` which has scroll shrink activated.
When `onChangeScrollShrink` is called, `closestContent` contains the `md-select`'s `md-content`, which is later moved to the body, but it's already too late, scroll listeners have already been applied :/

Duplicate of #9856
This time I cloned the repo instead of inline editing via github...

### DIFF
--- a/src/components/toolbar/toolbar.js
+++ b/src/components/toolbar/toolbar.js
@@ -122,7 +122,7 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
          *
          */
         function onChangeScrollShrink(shrinkWithScroll) {
-          var closestContent = element.parent().find('md-content');
+          var closestContent = $mdUtil.getSiblings(element, 'md-content');
 
           // If we have a content element, fake the call; this might still fail
           // if the content element isn't a sibling of the toolbar
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -526,6 +526,21 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
       }
     },
 
+    /**
+     * Get an element siblings matching a given tag name
+     *
+     * @param el Element to start walking the DOM from
+     * @param tagName Tag name to match
+     */
+    getSiblings: function getSiblings(element, tagName) {
+      var upperCasedTagName = tagName.toUpperCase();
+      if (element instanceof angular.element) element = element[0];
+      var siblings = Array.prototype.filter.call(element.parentNode.children, function(node) {
+        return element !== node && node.tagName.toUpperCase() === upperCasedTagName;
+      });
+      return angular.element(siblings);
+    },
+
     /*
      * getClosest replicates jQuery.closest() to walk up the DOM tree until it finds a matching nodeName
      *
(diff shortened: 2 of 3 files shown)


src/core/util/util.js
```diff
@@ -526,6 +526,34 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
       }
     },
 
+    /**
+     * Get an element direct children matching a given tag name
+     *
+     * @param element Element to start walking the DOM from
+     * @param tagName Tag name to match
+     */
+    getChildren: function getChildren(element, tagName) {
```
crisbeto: Doesn't look like `getChildren` is being used anywhere. I agree that it may be useful, but it's not worth having it if it's not used.

src/components/toolbar/toolbar.js
```diff
@@ -122,7 +122,7 @@ function mdToolbarDirective($$rAF, $mdConstant, $mdUtil, $mdTheming, $animate) {
          *
          */
         function onChangeScrollShrink(shrinkWithScroll) {
-          var closestContent = element.parent().find('md-content');
+          var closestContent = angular.element($mdUtil.getSiblings(element, 'md-content'));
```
crisbeto: It might be more convenient `getSiblings` returned a jqLite wrapper in the first place, e.g.

```js
var siblings Array.prototype.filter.call(element.parentNode.children, ...);
return angular.element(siblings);
```

src/core/util/util.js
```diff
@@ -526,6 +526,34 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
+     * Get an element siblings matching a given tag name
+     *
+     * @param el Element to start walking the DOM from
+     * @param tagName Tag name to match
+     */
+    getSiblings: function getSiblings(element, tagName) {
+      var upperCasedTagName = tagName.toUpperCase();
+      if (element instanceof angular.element) element = element[0];
+      return Array.prototype.filter.call(element.parentNode.children, function(node) {
+        return element !== node && node.tagName === upperCasedTagName;
```
crisbeto: Could you call `toUpperCase` on the `node.tagName` as well? In certain situations the tagName/nodeName can be lowercase (see #9509).

crisbeto:
LGTM @ThomasBurleson. @myagoo could you try signing the CLA? Otherwise we're not allowed to merge your PR.

myagoo:
I made those commits with my corporate email adress. How can I solve this ?

myagoo:
Apparently, adding my corporate email address as a secondary email address on GitHub did the trick :D

ThomasBurleson:
@myagoo - please squash your commits.

myagoo:
Is this ok ?

crisbeto:
It seems like your PR still has 2 commits @myagoo, it's only supposed to have 1. You can try looking up guides on how to squash commits ([like this one](https://ariejan.net/2011/07/05/git-squash-your-latests-commits-into-one/)).

myagoo:
I know how to squash them (i done it already), but I thought it would be good to keep separate the feature add commit from the bugfix commit.

crisbeto:
Nope, it should be one. You can have a list of fixes inside your commit message, e.g.:

```markdown
* Fixes an issue.
* Fixes some other issue.
```

PR #9873, opened 2016-10-20
fix(layout.scss): Prevent duplicate layout-row and layout-column rules

mallowigi:
I've found out that the .layout-row and .layout-column css rules are duplicated, by being inside the @for loop without any interaction with the $i variable. This creates unnecessary duplicate declarations.


ThomasBurleson:
@clshortfuse - let's confirm this fix and check all the layout demos accordingly.

ThomasBurleson:
@clshortfuse  - ping for review.

PR #9875, opened 2016-10-20
fix(constant): remove dependency on $sniffer

crisbeto:
The `$mdConstant` service uses Angular's `$sniffer` service to determine the browser's prefix via the `vendorPrefix`.
`vendorPrefix` was removed with https://github.com/angular/angular.js/commit/35482babd9eb0970edbc99f223e04594a9d09364.
These changes provide the same functionality, without depending on `$sniffer`. They also fix the current CI failures.

### DIFF
--- a/src/core/util/constant.js
+++ b/src/core/util/constant.js
@@ -5,12 +5,12 @@ angular.module('material.core')
  * Factory function that creates the grab-bag $mdConstant service.
  * @ngInject
  */
-function MdConstantFactory($sniffer, $window, $document) {
+function MdConstantFactory() {
 
-  var vendorPrefix = $sniffer.vendorPrefix;
+  var prefixTestEl = document.createElement('div');
+  var vendorPrefix = getVendorPrefix(prefixTestEl);
   var isWebkit = /webkit/i.test(vendorPrefix);
   var SPECIAL_CHARS_REGEXP = /([:\-_]+(.))/g;
-  var prefixTestEl = document.createElement('div');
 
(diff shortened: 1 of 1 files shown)


src/core/util/constant.js
```diff
@@ -33,6 +33,17 @@ function MdConstantFactory($sniffer, $window, $document) {
...
+  function getVendorPrefix() {
+    var prop, match;
+    var vendorRegex = /^(Moz|webkit|ms)(?=[A-Z])/;
+
+    for (prop in prefixTestEl.style) {
+      if (match = vendorRegex.exec(prop)) {
+        return match[0];
+      }
+    }
+  }
```
ThomasBurleson: what is returned if not a match?  `""` ?

src/core/util/constant.js
```diff
@@ -5,12 +5,12 @@ angular.module('material.core')
  * Factory function that creates the grab-bag $mdConstant service.
  * @ngInject
  */
-function MdConstantFactory($sniffer, $window, $document) {
+function MdConstantFactory() {
 
-  var vendorPrefix = $sniffer.vendorPrefix;
+  var prefixTestEl = document.createElement('div');
+  var vendorPrefix = getVendorPrefix();
```
crisbeto: Done. Also added it to `hasStyleProperty`.

src/core/util/constant.js
crisbeto: It's `undefined`. That's what got returned before as well.

PR #9877, opened 2016-10-20
feat(autocomplete): Ability to hide clear icon.

### LINKED ISSUE #4841: md-autocomplete: make clear button optional
Our UI designer is asking us to remove the clear button from the autocomplete component in a specific situation. I am wondering if there should be an attribute for ````md-autocomplete```` such as ````disable-clear="true"```` or if we should just use CSS to hide that button in the particular case.

Thoughts?

topherfangio:
Per multiple users' requests, we have added the `md-hide-clear-button`
class to the autocomplete so that users can hide the clear button.

Fixes #4841.

### DIFF
--- a/src/components/autocomplete/autocomplete.scss
+++ b/src/components/autocomplete/autocomplete.scss
@@ -235,6 +235,13 @@ md-autocomplete {
   }
 }
 
+// Allow hiding of the clear button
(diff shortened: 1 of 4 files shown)


topherfangio:
Closing in favor of #9892 which allows both hiding/showing.

PR #9885, opened 2016-10-21
fix(chips): use empty chip buffer if not a string

### LINKED ISSUE #9867: error inside onInputBlur, when using md-chips with md-autocomplete
**Actual Behavior**:
-  `What is the issue? *` Click inside the chips input. Then click outside to blur. It throws an error.
-  `What is the expected behavior?` shouldn't throw error on blur

-  `CodePen Demo which shows your issue:` http://codepen.io/ex37/pen/jrQkvr
-  `Details:` This demo is a part of Angular Material's documentation. This problem was introduced in version 1.1.1, and was not there before.

**Angular Versions**: *
-  `Angular Version:` 1.5.6

**Additional Information**:
-  `Browser Type: *` Chrome
-  `Browser Version: *` 54.0.2840.59 (64-bit)
-  `OS: *` macOS Sierra
-  `Stack Traces:` 

```
angular-material.min.js:14 Uncaught TypeError: Cannot read property 'trim' of null(…)
e.onInputBlur   @   angular-material.min.js:14
(anonymous function)    @   angular.js:1256
Qf  @   angular.js:3456
d   @   angular.js:3444
```

---

>   Add more lines if needed.

DevVersion:
- Sometimes the ngModel of the custom input element (provided by the user) does not have any initial $viewValue.
- The chips component always expects the chipBuffer to be a string and a non-string reference can cause Runtime Exceptions.

Fixes #9867.

### DIFF
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -383,12 +383,15 @@ MdChipsCtrl.prototype.useOnSelectExpression = function() {
  * model of an {@code md-autocomplete}, or, through some magic, the model
  * bound to any inpput or text area element found within a
  * {@code md-input-container} element.
- * @return {Object|string}
+ * @return {string}
  */
 MdChipsCtrl.prototype.getChipBuffer = function() {
-  return !this.userInputElement ? this.chipBuffer :
-      this.userInputNgModelCtrl ? this.userInputNgModelCtrl.$viewValue :
-          this.userInputElement[0].value;
+  var chipBuffer =  !this.userInputElement ? this.chipBuffer :
+                     this.userInputNgModelCtrl ? this.userInputNgModelCtrl.$viewValue :
+                     this.userInputElement[0].value;
+
+  // Ensure that the chip buffer is always a string. For example, the input element buffer might be falsy.
+  return angular.isString(chipBuffer) ? chipBuffer : '';
 };
 
 /**
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -1224,6 +1224,19 @@ describe('<md-chips>', function() {
               expect(scope.items.length).toBe(4);
               expect(scope.items[3]).toBe('Grape');
             }));
+
+          it('should use an empty string if ngModel value is falsy', inject(function($timeout) {
+            var element = buildChips(NG_MODEL_TEMPLATE);
+            var ctrl = element.controller('mdChips');
+
+            $timeout.flush();
+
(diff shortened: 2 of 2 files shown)


src/components/chips/js/chipsController.js
```diff
@@ -383,12 +383,15 @@ MdChipsCtrl.prototype.useOnSelectExpression = function() {
...
 MdChipsCtrl.prototype.getChipBuffer = function() {
-  return !this.userInputElement ? this.chipBuffer :
-      this.userInputNgModelCtrl ? this.userInputNgModelCtrl.$viewValue :
-          this.userInputElement[0].value;
+  var chipBuffer =  !this.userInputElement ? this.chipBuffer :
+                     this.userInputNgModelCtrl ? this.userInputNgModelCtrl.$viewValue :
+                     this.userInputElement[0].value;
+
+  // Ensure that the chip buffer is always a string. For example, the input element buffer might be falsy.
+  return angular.isString(chipBuffer) ? chipBuffer : '';
```
topherfangio: Is it ever possible for `chipBuffer` to be something like `3` instead of the string `'3'`? If so, this may fail?

DevVersion: Yeah, it could happen if the developer uses a custom input element or for example the `md-autocomplete`

topherfangio: Indeed; I'm just wondering if we could alter this to cast the input to a string instead of returning an empty string.

So, perhaps something like:

`return angular.isDefined(chipBuffer) ? chipBuffer + '' : ''`

This will ensure that it is always a string, or the empty string, but should handle more cases. Thoughts?

DevVersion: I thought a lot about this, and figured that just returning an empty string is the safest & cleanest approach IMO

I don't think casting would open more question than it solves, because then users might get weird chips like `[Object]`, rather than not showing any chip.
> I used the `isString` to avoid objects being used as truthy chip buffers.

ppham27:
Should this JSDoc annotation be changed if this.chipBuffer is not necessarily a string?

See: https://github.com/DevVersion/material/blob/4c0c7f9cc0fa7f24a108b08d849cf905183e6451/src/components/chips/js/chipsController.js#L79

topherfangio:
@ppham27 I think that's okay the way it is. Although you can technically force it to be something other than a string, it *should* always be a string.

DevVersion:
@ppham27 Yeah this is fine. Also the `chipBuffer` variable is always a string, because it is from our internal ngModel and can't be overwritten with custom ngModel formatters by the user.

PR #9886, opened 2016-10-21
fix(panel): clickOutsideToClose with propagateContainerEvents

### LINKED ISSUE #9388: mdPanel -allow clickOutsideToClose works with propagateContainerEvents
**Actual Behavior**:
-  Currently mdPanel by default use body element as parent and have wrapper(`.md-panel-outer-wrapper`) with styles `{ height: 100%; left: 0; position: absolute; top: 0; width: 100%; }` so we must use `propagateContainerEvents: true` (`pointer-events: none;`) to be able to interact with page elements. But in this case `clickOutsideToClose:true` doesn't work at all.
-  When `clickOutsideToClose:true` is specified along with `propagateContainerEvents: true` it should close our panel when user clicks somewhere on page, but also propagate this event to a target.

-  You could observe this in demo example in docs for 1.1.0
-  N/A

bradrich:
If both the `clickOutsideToClose` and `propagateContainerEvents` parameters are set to true within the panel configuration, then the panel will be closed when a click happens outside of the panel.

Fixes #9388

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -1825,15 +1825,17 @@ MdPanelRef.prototype._configureEscapeToClose = function() {
  */
 MdPanelRef.prototype._configureClickOutsideToClose = function() {
   if (this.config['clickOutsideToClose']) {
-    var target = this.panelContainer;
-    var sourceElem;
+    var target = this.config['propagateContainerEvents'] ?
+        angular.element(document.body) :
+        this.panelContainer;
+    var sourceEl;
 
     // Keep track of the element on which the mouse originally went down
     // so that we can only close the backdrop when the 'click' started on it.
-    // A simple 'click' handler does not work,
-    // it sets the target object as the element the mouse went down on.
+    // A simple 'click' handler does not work, it sets the target object as the
+    // element the mouse went down on.
     var mousedownHandler = function(ev) {
-      sourceElem = ev.target;
+      sourceEl = ev.target;
     };
 
     // We check if our original element and the target is the backdrop
@@ -1841,7 +1843,15 @@ MdPanelRef.prototype._configureClickOutsideToClose = function() {
     // panel we don't want to panel to close.
     var self = this;
     var mouseupHandler = function(ev) {
-      if (sourceElem === target[0] && ev.target === target[0]) {
+      if (self.config['propagateContainerEvents']) {
+
+        // We check if the sourceEl of the event is the panel element or one
+        // of it's children. If it is not, then close the panel.
+        if (sourceEl !== self.panelEl[0] && !self.panelEl[0].contains(sourceEl)) {
+          self.close();
+        }
+
+      } else if (sourceEl === target[0] && ev.target === target[0]) {
         ev.stopPropagation();
         ev.preventDefault();
 
@@ -2713,6 +2723,7 @@ MdPanelPosition.prototype._constrainToViewport = function(panelEl) {
   }
 };
 
+
 /**
  * Switches between 'start' and 'end'.
  * @param {string} position Horizontal position of the panel
@@ -2902,6 +2913,7 @@ MdPanelAnimation.prototype.closeTo = function(closeTo) {
   return this;
 };
 
+
 /**
  * Specifies the duration of the animation in milliseconds.
  * @param {number|{open: number, close: number}} duration
@@ -2926,6 +2938,7 @@ MdPanelAnimation.prototype.duration = function(duration) {
     if (angular.isNumber(value)) return value / 1000;
   }
 };
+
 
 /**
  * Returns the element and bounds for the animation target.
--- a/src/components/panel/panel.spec.js
+++ b/src/components/panel/panel.spec.js
@@ -544,6 +544,20 @@ describe('$mdPanel', function() {
       expect(PANEL_EL).not.toExist();
     });
 
+    it('should close when clickOutsideToClose set to true and ' +
+        'propagateContainerEvents is also set to true', function() {
+          var config = {
+            propagateContainerEvents: true,
+            clickOutsideToClose: true
+          };
+
+          openPanel(config);
+
+          clickPanelContainer(getElement('body'));
+
+          expect(PANEL_EL).not.toExist();
+        });
+
     it('should not close when escapeToClose set to false', function() {
       openPanel();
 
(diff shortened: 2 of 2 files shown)


src/components/panel/panel.js
```diff
@@ -1737,14 +1741,36 @@ MdPanelRef.prototype._configureClickOutsideToClose = function() {
...
 
         self.close();
       }
     };
 
+    var findSourceElem = function(elems, sourceElem) {
+      var found = false;
+      angular.forEach(elems, function(elem) {
+        if (elem === sourceElem) {
+          found = true;
```
ErinCoughlan: How about this? It quits early when it finds a match and removes the extra variable.

```
var findSourceEl = function(els, sourceEl) {
  for (var el, i = 0; el = els[i]; i++) {
    if (el === sourceEl) {
      return true;
    }
  });
  return false;
};
```

src/components/panel/panel.js
```diff
@@ -1737,14 +1741,36 @@ MdPanelRef.prototype._configureClickOutsideToClose = function() {
     // panel we don't want to panel to close.
     var self = this;
     var mouseupHandler = function(ev) {
-      if (sourceElem === target[0] && ev.target === target[0]) {
+      if (self.config['propagateContainerEvents']) {
+
+        // We check if the sourceElem of the event is the panel element or one
+        // of it's children. If it is not, then close the panel.
+        var found = findSourceElem(
```
ErinCoughlan: El instead of Elem. We use el throughout this file, so it's strange to switch naming conventions here.

src/components/panel/panel.spec.js
```diff
@@ -542,6 +542,20 @@ describe('$mdPanel', function() {
...
+    it('should close when clickOutsideToClose set to true and ' +
+        'propagateContainerEvents is also set to true', function() {
+          var config = {
+            propagateContainerEvents: true,
+            clickOutsideToClose: true
+          };
+
+          openPanel(config);
+
+          clickOutsideOfPanel();
```
ErinCoughlan: Can this be clickPanelContainer? Or can clickPanelContainer be updated so you're not duplicating logic. (The only thing that's changing is the element you're clicking on. Maybe something like this?

```
function clickPanelContainer(rootEl) {
  rootEl = rootEl || document.body;
  const selector = '.md-panel:not(._md-panel-backdrop)';
  angular.element(rootEl).find(selector).parent().simulate('click');
  angular.element(rootEl).find(selector).parent().simulate('mousedown');
  angular.element(rootEl).find(selector).parent().simulate('mouseup');
  flushPanel();
}
```
Benefits of this version is that you don't need the panelRef, which removes one of the if statements. I know this version can replace the current one and you won't need to change any tests.

or 

```
function clickPanelContainer(container) {
    container = container || angular.element('.md-panel-outer-wrapper');

    container.triggerHandler({
      type: 'mousedown',
      target: container[0]
    });

    container.triggerHandler({
      type: 'mouseup',
      target: container[0]
    });

    flushPanel();
  }
```

src/components/panel/panel.js
```diff
@@ -1721,30 +1723,53 @@ MdPanelRef.prototype._configureEscapeToClose = function() {
...
     // because if the original was the backdrop and the target was inside the
     // panel we don't want to panel to close.
     var self = this;
     var mouseupHandler = function(ev) {
-      if (sourceElem === target[0] && ev.target === target[0]) {
+      if (self.config['propagateContainerEvents']) {
+
+        // We check if the sourceEl of the event is the panel element or one
+        // of it's children. If it is not, then close the panel.
+        var found = findSourceEl(
```
crisbeto: You could simplify it and avoid the loop if you did something like this:

```js
var found = sourceEl === self.panelEl[0] || self.panelEl[0].contains(sourceEl);
```

src/components/panel/panel.spec.js
```diff
@@ -542,6 +542,20 @@ describe('$mdPanel', function() {
...
     });
 
+    it('should close when clickOutsideToClose set to true and ' +
+        'propagateContainerEvents is also set to true', function() {
+          var config = {
+            propagateContainerEvents: true,
+            clickOutsideToClose: true
+          };
+
+          openPanel(config);
```
crisbeto: It might make the test slightly more robust if you had a check that the `PANEL_EL` exists, right after opening it. This should avoid cases where the test will pass, if the `openPanel` function didn't open it up for some reason.

src/components/panel/panel.spec.js
```diff
@@ -2823,12 +2837,12 @@ describe('$mdPanel', function() {
...
   }
 
-  function clickPanelContainer() {
+  function clickPanelContainer(container) {
     if (!panelRef) {
       return;
     }
 
-    var container = panelRef.panelContainer;
+    var container = container || panelRef.panelContainer;
```
crisbeto: It's a bad practise to have variables with the same name as function arguments.

src/components/panel/panel.js
bradrich: I was looking for the `.contains()` method and didn't realize that I would have to access the element to get to it. Good catch and I appreciate you showing it to me.

src/components/panel/panel.spec.js
bradrich: This should actually read as `container = container || panelRef.panelContainer`. I forgot to remove the `var` before it.

crisbeto: Alright, it makes sense without the `var`.

src/components/panel/panel.spec.js
crisbeto: While it's true that the opening logic is well tested, we should try to keep the tests as independent as possible.

src/components/panel/panel.spec.js
bradrich: @ThomasBurleson I think that there was some confusion as we extended this function. Previously, the container variable was declared internally and assigned to the `panelRef.panelContainer`. Now, you are able to pass in a container to click on. But, if you don't pass one in, it needs to still assign it to the default `panelRef.panelContainer`.

src/components/panel/panel.spec.js
ErinCoughlan: @crisbeto I go back and forth on this one. Because the open existence is tested earlier, we get a better test signal when opens fails because the failure is in the test that tests open. If we write it in every test, then all will fail and the messages don't point to the actual problem.

src/components/panel/panel.spec.js
```diff
@@ -544,6 +544,20 @@ describe('$mdPanel', function() {
...
+    it('should close when clickOutsideToClose set to true and ' +
+        'propagateContainerEvents is also set to true', function() {
+          var config = {
+            propagateContainerEvents: true,
+            clickOutsideToClose: true
+          };
+
+          openPanel(config);
+
+          clickPanelContainer(angular.element(document.body));
```
ErinCoughlan: Can you move the `angular.element` part to the clickPanelContainer? I don't think it particularly matters for this test, so it seems to fit better in the helper method.

I'm wondering if we want something like `getElement` to make this more flexible.
https://github.com/angular/material/blob/72d068593b7f353d32f8308870d2b0d1c7c759c9/test/angular-material-spec.js#L287

bradrich: Went ahead and added this one, as I thought that it was a more extended implementation and useful.

bradrich:
@ThomasBurleson This is ready for merge.

src/components/panel/panel.spec.js
crisbeto: I wouldn't say it's necessary for every test, but it makes sense for this one since the test would fail if there isn't an open panel. This may be down to personal preference, though.

bradrich:
@ThomasBurleson Checking on the status of this PR.

PR #9892, opened 2016-10-22
feat(autocomplete): option to toggle the clear button

### LINKED ISSUE #2727: autocomplete: x icon issues and missing md-floating-label attribute in docs
According to the spec the "x" icon shouldn't be shown when your using `md-floating-label`. Also `md-floating-label` isn't in the docs list of attributes.

Spec for floating label style autocomplete(it's at the very bottom of the autocomplete stuff)
http://www.google.com/design/spec/components/text-fields.html#text-fields-auto-complete-text-field

Docs page missing `md-floating-label` under attributes
https://material.angularjs.org/#/api/material.components.autocomplete/directive/mdAutocomplete

Also the x icon should be just an x instead of the current circle with an x. This can be seen on the spec link above. Here is the icon from the official google material icons repo 
https://github.com/google/material-design-icons/blob/master/navigation/svg/production/ic_close_24px.svg

### LINKED ISSUE #4841: md-autocomplete: make clear button optional

DevVersion:
*  Adds a new attribute, which allows developers to explicitly show the clear button for all types of autocomplete's.

Closes #4841. Closes #2727

### DIFF
--- a/src/components/autocomplete/autocomplete.scss
+++ b/src/components/autocomplete/autocomplete.scss
@@ -1,5 +1,7 @@
 // The default item height is also specified in the JavaScript.
 $md-autocomplete-item-height: 48px !default;
+$md-autocomplete-clear-size: 30px !default;
+$md-autocomplete-input-offset: 20px !default;
 
 @keyframes md-autocomplete-list-out {
   0% {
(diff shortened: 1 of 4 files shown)


ThomasBurleson:
@DevVersion  - Questions:

1) Is this consistent with the **hide** clear button api ?
2) Does this work with the hide features; if so which one supercedes ?

DevVersion:
@topherfangio Not sure about this. I just tried it out again and it works for me - Let's pair on chat about it.

topherfangio:
Note that this will also close #4841 

@DevVersion Can you please add that to the commit description so it gets closed when this is merged?

DevVersion:
@topherfangio Done! Thanks for looking into it.

PR #9893, opened 2016-10-23
fix(datepicker): ensure that all month/year elements have the expected height

### LINKED ISSUE #9863: md datepicker scroll behaviour
**Actual Behavior**:
 -  `What is the issue? * Scrolling dates with limits does not have correct behaviour`
 -  `What is the expected behavior? When scrolling up should go to previous month and not next.`

 -  `CodePen Demo which shows your issue:`[CodePen Demo](http://codepen.io/MigFerreira/pen/kkQgBp)
 -  `Details: Select the from date and scroll slowly up, it will be visible it goes to the next month instead                 of previous`

**Angular Versions**: *
 -  `Angular Version: 1.5.8`
 -  `Angular Material Version: 1.1.1`

**Additional Information**:
 -  `Browser Type: Desktop Chrome`
 -  `Browser Version: 54`
 -  `OS: Windows 7`
 -  `Stack Traces:`

crisbeto:
The datepicker's calendar uses `md-virtual-repeat` to render itself. The virtual repeater expects that all elements that it has rendered will have a certain height, however this may not be the case if it renders an empty month/year element in the calendar. If this happens (it's usually if the current value and the maximum date are in the same month), the calendar can start jumping while scrolling.

This change adds an empty `<tr>` that will ensure that even empty elements have a height. When the element is compiled, the empty `<tr>` will be overwritten.

Fixes #9863.

### DIFF
--- a/src/components/datepicker/js/calendarMonth.js
+++ b/src/components/datepicker/js/calendarMonth.js
@@ -32,7 +32,13 @@
                   'md-month-offset="$index" ' +
                   'class="md-calendar-month" ' +
                   'md-start-index="monthCtrl.getSelectedMonthIndex()" ' +
-                  'md-item-size="' + TBODY_HEIGHT + '"></tbody>' +
+                  'md-item-size="' + TBODY_HEIGHT + '">' +
+
+                // The <tr> ensures that the <tbody> will always have the
+                // proper height, even if it's empty. If it's content is
+                // compiled, the <tr> will be overwritten.
+                '<tr aria-hidden="true" style="height:' + TBODY_HEIGHT + 'px;"></tr>' +
+              '</tbody>' +
             '</table>' +
           '</md-virtual-repeat-container>' +
         '</div>',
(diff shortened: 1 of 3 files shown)


ThomasBurleson:
@jelbourn - ping for final review.

src/components/datepicker/js/calendarMonth.js
```diff
@@ -32,7 +32,13 @@
                   'md-month-offset="$index" ' +
                   'class="md-calendar-month" ' +
                   'md-start-index="monthCtrl.getSelectedMonthIndex()" ' +
-                  'md-item-size="' + TBODY_HEIGHT + '"></tbody>' +
+                  'md-item-size="' + TBODY_HEIGHT + '">' +
+
+                // The <tr> ensures that the <tbody> will always have the
+                // proper height, even if it's empty. If it's content is
+                // compiled, the <tr> will be overwritten.
+                '<tr style="height:' + TBODY_HEIGHT + 'px;"></tr>' +
```
jelbourn: Empty `<tr>` should have `aria-hidden`

crisbeto: Updated.

PR #9895, opened 2016-10-23
feat(panel): add the ability to update the animation of an existing panel

crisbeto:
- Adds the ability to update the animation of a panel, without having to re-create it.
- Adds a unit test for verifying that the backdrop animation duration matches the one of the panel. This was missed in #9570.


ThomasBurleson:
@ErinCoughlan, @bradrich - ping for review.

bradrich:
LGTM

PR #9909, opened 2016-10-24
fix(autocomplete): Fix messages not appearing.

### LINKED ISSUE #9468: ng-message of md-autocomplete is invisible
**Actual Behavior**:
If you enter something  and remove it, a required error message was shown. 
If you try it a second time, the error message wasn't shown, because of the following style 

`md-input-container .md-input-message-animation:not(.ng-animate) {
    opacity: 0;
    margin-top: -100px;
}.`

Used the floating label example of the demos

topherfangio:
In certain scenarios, the `ng-messages` of an autocomplete would fail to animate properly leaving them in a state where they were in the DOM, but not visible to the user.

This was ultimately caused by the animations not properly retrieving the messages element to animate, but was also tangentially caused by the animations not properly calling the `done` callback upon failure.

There are three associated fixes:
1. Log a warning if the messages element or it's children cannot
   be found.
2. Ensure the `done` callback always fires even on failures.
3. Ensure the `getMessagesElement()` method correctly returns when
   the element passed IS the messages element itself.

Fixes #9468.

### DIFF
--- a/test/angular-material-mocks.js
+++ b/test/angular-material-mocks.js
@@ -16,6 +16,12 @@
 
 'use strict';
 
+  // Allow our code to know when they are running inside of a test so they can expose extra services
+  // that should NOT be exposed to the public but that should be tested.
+  //
+  // As an example, see input.js which exposes some animation-related methods.
+  window._mdMocksIncluded = true;
+
 /**
  * @ngdoc module
  * @name ngMaterial-mock
--- a/src/components/input/input-animations.spec.js
+++ b/src/components/input/input-animations.spec.js
@@ -1,5 +1,5 @@
 describe('md-input-container animations', function() {
-  var $rootScope, $compile, $animateCss, $material,
+  var $rootScope, $compile, $animateCss, $material, $$mdInput,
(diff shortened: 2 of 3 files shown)


test/angular-material-mocks.js
```diff
@@ -16,6 +16,12 @@
 
 'use strict';
 
+  // Allow our code to know when they are running inside of a test so they can expose extra services
+  // that should NOT be exposed to the public but that should be tested.
+  //
+  // As an example, see input.js which exposes some animation-related methods.
+  window._mdMocksIncluded = true;
```
topherfangio: I am open to alternate ways of doing this; if you have a better suggestion, I'd love to hear it!

ThomasBurleson:
@topherfangio - can you fix this plz so we can merge it.

topherfangio:
@ThomasBurleson Updated. I can use this for the test methods (and have done so), but for the animations, I need the injector to retrieve my dependencies.

Will this work?

ThomasBurleson:
@topherfangio - lgtm

PR #9925, opened 2016-10-26
fix(navbar): tabs not being read out by screen readers

### LINKED ISSUE #9383: Nav Bar & Tabs - Tabindex for Inactive Elements
**Actual Behavior**:
-  `What is the issue? *` Currently the `tabindex` for inactive elements in the Nav Bar & Tab components is set to -1. This renders inactive tabs essentially invisible to screen readers.
  If the `tabindex` was set to zero, the same as the active tab, both the Nav Bar & Tab components would be more accessible to all users (not just screen readers).

-  Inspect element on either the Tabs or Nav Bar demos to view the current values. If you use the tab key on your keyboard to cycle through interactive elements on the rendered page, you will see that the elements with `tabindex="-1"`are completely skipped over.

**Angular Versions**: *
-  `Angular Version:`

**Notes:** If the core team would be willing to review it, my team would be happy to open a PR for this change.

tabindex

crisbeto:
- Fixes the tabs in the navbar not being read out by screen readers. This is due to the fact that they were being highlighted as focused, but focus wasn't actually being passed to them.
- Some minor cleanup in the navbar.

Fixes #9383.

### DIFF
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -57,7 +57,7 @@ angular.module('material.components.navBar', ['material.core'])
  *</hljs>
  * <hljs lang="js">
  * (function() {
- *   ‘use strict’;
+ *   'use strict';
  *
  *    $rootScope.$on('$routeChangeSuccess', function(event, current) {
  *      $scope.currentLink = getCurrentLinkFromRoute(current);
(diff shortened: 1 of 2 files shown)


crisbeto:
CC @ErinCoughlan

frkfir:
Still doesn't work after the changes I've made like suggestion here.
The navbar not being read out by the screen reader.
This due to the fact that they were being highlighted as focused, but focus wasn't actually being passed to them.
@EladBezalel

PR #9934, opened 2016-10-27
feat(core): add support for hybrid ngMd1+ngMd2 applications

### LINKED ISSUE #9824: update(css): update ngM1 CSS to avoid conflicts with ngM2 stylesheets
Teams migrating ngM1 applications to ng2 + ngM2 may experience css style collisions when both ngM1 + ngM2 stylesheets are loaded.
### ngM2 CSS Changes

Styles will be updated to support concepts/approaches described in  [BEM - Block Element Modifiers](http://getbem.com/naming/).
- In material2, change all our css classes to be prefixed either as `md--` or `md-card--`
- stop using **element selectors** in material2, and always use classes
-  add a css class called `md2` to each material1 component (elements only)
### ngM1 CSS Changes
- Change any ngM1 styles that use an **element selector**  to also have `:not(.md2)`
-  These changes are needed in both ngMaterial 1.0.9 and 1.1.x 

> Using `:not` doesn't increase specificity when the given expression doesn't match anything. So that shouldn't break any existing consumption based on specificity. This will prevent material1 css from affecting anything in material2, and then we can both use the `md-` prefix for elements.

clshortfuse:
Angular Material 1 CSS should not conflict with Angular Material 2 CSS

* Use PostCSS to change Element based CSS selectors to target via a class
* Create directive for components that conflict with MD2
* Use directive to add a class during postLink phase
* Update md-icon spec tests to support extra classes
* Update md-toolbar spec test to support multiple directives
* Use mdCheckboxDirective to add "md-checkbox" class

Fixes #9824

Breaking Change:

In CSS, class selectors have higher specificity than element selectors by a factor of 10. This may break some custom user styles.

### DIFF
--- a/gulp/tasks/build-scss.js
+++ b/gulp/tasks/build-scss.js
@@ -39,6 +39,7 @@ exports.task = function() {
       .pipe(concat('angular-material.scss'))
       .pipe(gulp.dest(dest))            // raw uncompiled SCSS
       .pipe(sass())
+      .pipe(util.applyMd2CompatibilityCss())
       .pipe(util.dedupeCss())
       .pipe(util.autoprefix())
       .pipe(insert.prepend(config.banner))
@@ -69,6 +70,7 @@ exports.task = function() {
         .pipe(insert.prepend(config.banner))
         .pipe(gulp.dest(layoutDest))      // raw uncompiled SCSS
         .pipe(sass())
+        .pipe(util.applyMd2CompatibilityCss())
         .pipe(util.dedupeCss())
         .pipe(util.autoprefix())
         .pipe(rename({ extname : '.css'}))
@@ -91,6 +93,7 @@ exports.task = function() {
           .pipe(sassUtils.hoistScssVariables())
           .pipe(gulp.dest(layoutDest))     // raw uncompiled SCSS
           .pipe(sass())
+          .pipe(util.applyMd2CompatibilityCss())
           .pipe(util.dedupeCss())
           .pipe(util.autoprefix())
           .pipe(rename({ extname : '.css'}))
--- a/gulp/util.js
+++ b/gulp/util.js
@@ -35,6 +35,7 @@ exports.readModuleArg = readModuleArg;
 exports.themeBuildStream = themeBuildStream;
 exports.minifyCss = minifyCss;
 exports.dedupeCss = dedupeCss;
+exports.applyMd2CompatibilityCss = applyMd2CompatibilityCss;
 exports.args = args;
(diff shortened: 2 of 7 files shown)


clshortfuse:
Manually verified with the following command:

```
grep -rn -e "^md" -e "[^._-\[]md" -e "(md" src/ --include=*.scss | sort -t: -k1,1 -k2,2n | grep -e "md" -e "not(.md2)"
```

clshortfuse:
It seems as though adding `.not:(.md2)` raises specificity of our CSS rules by a factor of 1. This can lead to breaking changes with custom theming. 

A clear is example with our demo site. There is a CSS rule for `.site-sidenav md-content` that is being overridden by `md-sidenav:not(.md2) md-content:not(.md2)`. Just have one psuedo selector is enough to override the demo site's CSS rule (`md-sidenav:not(.md2) md-content`).

clshortfuse:
I modified the code to use a variable called `$md1-element-suffix` instead with a default value of `''`. 

MD2 compatibility won't work out of the box, but if you compile MD1 on your own, provide an `override.scss` file with the following line included:
```
$md1-element-suffix: ":not(.md2)";
```

Then when you build, use the override argument like so:

```
gulp build --release --override=override.scss
```

DevVersion:
@clshortfuse Do you think this suffixing is also possible with a PostCSS plugin?

<details>
<summary>Here is some skeleton code, which already worked (but needs to be validated / improved)</summary>
**Note** that the Regex here might not work properly, but it might be a possible approach.
```js
function suffixCss() {

  return insert.transform(function(contents) {
    var parsed = postcss.parse(contents);

    parsed.walkRules(function(rule) {

      if (rule.parent && rule.parent.name === 'keyframes') {
        return;
      }

      rule.selectors = rule.selectors.map(function(selector) {
        return selector.replace(/(md-)(\w|-)+/g, '$&:not(.md2)');
      });

    });
    
    // Turn the AST back into CSS.
    return parsed.toResult().css;

  });

}
```
</details>

clshortfuse:
@DevVersion Regex could work, but you'd have to be very careful. Not all element selectors begin and end with a space. They could begin with `}` `,` or `>` and end with `{` `,` or `>`. I'm not sure if I forgot any others, but I was trying to be "safer than sorry".

DevVersion:
Exactly, that's why my Regexp (which was just to show what I mean) works independently of the preceding / trailing characters.

ThomasBurleson:
Closing as stale and not needed.

PR #9940, opened 2016-10-28
feat(select): add the ability to pre-select the only option in the list

### LINKED ISSUE #9626: select: select option if there is only 1 option
Feature request: automatically select option from md-select if there is just 1 option.
Maybe this could be an optional setting on the directive.

**Actual Behavior**:
-  `What is the issue? *`
  I want to improve the usability of my form by pre-selecting values from md-select items if they only have 1 option.

  http://codepen.io/joranbodt/pen/dpXNoE
-  `Details:`

**Angular Versions**: *
-  `Angular Version:`

**Additional Information**:
-  `Browser Type: *`
-  `Browser Version: *`
-  `OS: *`
-  `Stack Traces:`

crisbeto:
- Adds the `md-select-only-option` attribute, that will cause `<md-select>` to pre-select the first option, if it only has one option in it's list.
- Fixes an ugly way of adding extra classes to the select container.
- Removes a few variables and arguments that weren't being used.

Fixes #9626.


ThomasBurleson:
@crisbeto - please verify this PR is still valid/needed and rebase. Thank you.

crisbeto:
It's rebased @ThomasBurleson.

src/components/select/select.js
```diff
@@ -882,6 +893,7 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
       } else {
         $element.find('md-content').removeAttr('aria-activedescendant');
       }
+      self.setMultiple(self.isMultiple);
```
Splaktar, line 896:
```suggestion
```

PR #9941, opened 2016-10-29
feat(panel): allow transform to be animated on an offset panel

### LINKED ISSUE #9905: feat(panel): allow transform to be animated on elements with an offset
If a panel has an offset (`position.withOffsetX` or `position.withOffsetY`), then it's `transform` can't be animated, because it uses inline transforms to do the offsetting. I'm considering applying the animations to an inner container, that has nothing to do with the panel positioning.

This issue is blocking for #9641.

CC @ErinCoughlan @bradrich @DerekLouie

crisbeto:
Previously, if a panel had a position with an offset (e.g. `$mdPanel.newPanelPosition().center()`), the positioning would break any `transform` animations on the panel. This was due to the fact that `mdPanel` uses inline `transform` styles to do the offsetting. These changes introduce a wrapper around the panel (`.md-panel-inner-wrapper`), which will handle all of the positioning, allowing for any animations to be applied to the `.md-panel` itself.

Relates to #9641.
Fixes #9905.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -236,7 +236,7 @@ angular
  *   - `locals` - `{Object=}`: An object containing key/value pairs. The keys
  *     will be used as names of values to inject into the controller. For
  *     example, `locals: {three: 3}` would inject `three` into the controller,
- *     with the value 3. 'mdPanelRef' is a reserved key, and will always 
+ *     with the value 3. 'mdPanelRef' is a reserved key, and will always
  *     be set to the created MdPanelRef instance.
  *   - `resolve` - `{Object=}`: Similar to locals, except it takes promises as
  *     values. The panel will not open until all of the promises resolve.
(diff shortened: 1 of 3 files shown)


src/components/panel/panel.js
```diff
@@ -2605,37 +2619,37 @@ MdPanelPosition.prototype.getTransform = function() {
 
 /**
  * Sets the `transform` value for a panel element.
- * @param {!angular.JQLite} panelEl
+ * @param {!angular.JQLite} element
```
bradrich: For consistency sake, I would just use `el` here, and then all the way through.

crisbeto: Done.

ThomasBurleson:
@crisbeto - please rebase and I will do a final review.

kara:
@crisbeto Rebase?

crisbeto:
Rebased @kara.

Splaktar:
Please rebase once more.

crisbeto:
Rebased.

Splaktar:
I'm going to move this over to https://github.com/angular/material/pull/11390 since it needs rebasing again and will also need some changes to fix tooltip styling.

PR #9944, opened 2016-10-30
fix(panel): panel not being constrained to viewport on repeat openings

### LINKED ISSUE #9942: mdPanel: logic for keeping element on screen doesn't work in certain cases
The logic that keeps a panel within the viewport doesn't seem to work in certain cases. Here's an example from the demo page:

groups

Note that while this happens on the "groups", the "Select menu" from the "Basic usage" example seems to work reliably.

CC @ErinCoughlan @bradrich

crisbeto:
The panel wasn't being constrained to the viewport when opening it multiple times in a row, due to the "constrain to viewport" logic only running if the `_actualPosition` isn't known. These changes make it so the panel always gets constrained to the viewport.

Fixes #9942.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -2679,7 +2679,7 @@ MdPanelPosition.prototype._reduceTranslateValues =
  * @private
  */
 MdPanelPosition.prototype._setPanelPosition = function(panelEl) {
-  // Remove the class in case it has been added before.
+  // Remove the "position adjusted" class in case it has been added before.
   panelEl.removeClass('_md-panel-position-adjusted');
 
   // Only calculate the position if necessary.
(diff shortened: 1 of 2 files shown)


bradrich:
LGTM.

scriptsure:
so i have an issue not sure if it is related to your work on this... but i am finding that the when a panel is opened that is beyond the vertical size of the window, the bottom works perfect, the top does not... rides off view.... here is a codepen.... just click the select menu button

aaron

ErinCoughlan:
@scriptsure mdPanel uses `position: fixed`, which means it cannot be larger than the viewport and be guaranteed to function properly and show all items. I'd suggest for your select-menu that you restrict the height and scroll within the panel instead of relying on the browser scroll. (There's also no codepen link, but I remember you asking a similar question in another issue. Correct me if I'm wrong though.)

scriptsure:
ops sorry... yeah that case was surrounding on small devices. i figured that one out with some css...

http://codepen.io/scriptsure/pen/vydYGK

PR #9964, opened 2016-11-02
feat(autocomplete): pass md-input-max/min-length to the input with no…

theLufenk:
…n-floating label

Adds the md-input-minlength and md-input-maxlength attributes as ng-minlength and ng-maxlength to the input element of the autocomplete directive when a floating label is not used.
A use case : It would help in validating the input when the value of Search Text is being used if no matches are found.

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -596,7 +596,7 @@ describe('<md-autocomplete>', function() {
 
     describe('md-input-maxlength', function() {
 
-      it('should correctly set the form to invalid', inject(function($timeout) {
+      it('should correctly set the form to invalid if floating label is present', inject(function($timeout) {
         var scope = createScope(null, {inputId: 'custom-input-id'});
         var template =
           '<form name="testForm">' +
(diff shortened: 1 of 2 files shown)


theLufenk:
I believed it would be taken up with #4492 . However it wasn't.
As of now, there is no way to add length validation to the text input element.

The value of Search Text is exposed , hence it can be used in a scenario when we want the autocomplete to serve as both a suggestion list as well as text input .

DevVersion:
Thanks for the pull request - The reason why the `ng-min/max-length` attributes are not supported a normal autocomplete is that we don't support `ng-messages` for a normal autocomplete.

My question here is, how will you validate the "underlaying" input?

theLufenk:
The property is exposed as [FORM_NAME].[MD_INPUT_NAME].$error.maxlength nonetheless.

DevVersion:
@theLufenk Ah I see. From this perspective it makes sense to support those `validator` attributes for the normal autocomplete.

Can you please some tests for those two attributes?

theLufenk:
Added tests. The spec was missing tests for md-input-minlength. Added them too.
I suppose checking for presence of ng-maxlength/ng-minlength attribute in the render input element and $valid state of container forms would be sufficient ?

DevVersion:
@theLufenk Perfect! I think the tests which are only checking for the `attribute` on the input are not necessary when having the tests for the `$valid` state.

> If you have time it would be good to have them removed (those are unnecessary and bloating the specs)

theLufenk:
Done and rebased.

PR #9973, opened 2016-11-04
chore(components): use $onInit hook for binding assignments

DevVersion:
<details>
<summary>What changed?</summary>
A few components are using the `bindToController` option for the scope bindings.

Recently a Breaking Change commit landed in the AngularJS master branch, which changed the way the bindings are assigned.

The bindings are no longer pre-assigned in the constructors, and will be only set before the `$onInit` lifecycle hook.

Angular Material supports Angular versions which do not support lifecycle hooks and so we can't ensure that Snapshot and the older versions work together with `$onInit` approach.
</details>

---
* Adds the $onInit hook to all affected components and manually dispatches the $onInit hook for AngularJS 1.4 and older versions.

* Fixes two checks in the `datepicker`, which required the controller to have a binding set after $onInit (unnecessary here)

Fixes the Breaking Change: https://github.com/angular/angular.js/commit/bcd0d4d896d0dfdd988ff4f849c1d40366125858

cc. @gkalpak @mgol @petebacondarwin


src/components/datepicker/js/calendar.js
```diff
@@ -205,18 +200,41 @@
       angular.element(document.body).off('keydown', boundKeyHandler);
     });
 
-    if (this.minDate && this.minDate > $mdDateLocale.firstRenderableDate) {
+    // For older Angular versions, where the bindings are pre-assigned, manually call the $onInit hook.
```
mgol: This sounds like you're calling `$onInit` manually because bindings are pre-assigned and the real reason is that this hook is not supported in older versions and that's why you have to call it. The fact that bindings are pre-assigned only means that it will work even though normally the hook would be invoked later. How about:
> For Angular 1.4 and older, where there are no lifecycle hooks but bindings are pre-assigned, manually call the $onInit hook.

src/components/datepicker/js/calendarMonthBody.js
```diff
@@ -33,8 +33,8 @@
         // of repeated items that are linked, and then those elements have their bindings updated.
         // Since the months are not generated by bindings, we simply regenerate the entire thing
         // when the binding (offset) changes.
-        scope.$watch(function() { return monthBodyCtrl.offset; }, function(offset, oldOffset) {
-          if (offset !== oldOffset) {
+        scope.$watch(function() { return monthBodyCtrl.offset; }, function(offset) {
+          if (angular.isNumber(offset)) {
```
mgol: This change (here & in other places) doesn't seem to be mentioned in the commit message or issue description.

mgol:
In the commit message:
> Angular Material supports Angular versions, which do not support lifecycle hooks

There should be no comma before `which`.

> a Breaking Chang commit

`Chang` -> `Change`

> Recently in the AngularJS master branch was added a Breaking Chang commit

How about "Recently a Breaking Change commit landed in the AngularJS master branch"?

mgol:
LGTM, assuming it fixes the Jenkins build. 👍

ThomasBurleson:
@DevVersion - only Calendar and Tabs need this? What about Panel, etc. ?

DevVersion:
@ThomasBurleson Yeah. This change is only necessary for directives which use `bindToController` and have scope bindings set.

> Panel, Dialog, Toast etc. depend on our internal compiler and aren't affected from this breaking change (In the future they might, if we support the component API)

src/components/datepicker/js/calendar.js
```diff
@@ -205,18 +200,42 @@
...
+    // manually call the $onInit hook.
+    if (angular.version.major === 1 && angular.version.minor <= 4) {
+      this.$onInit();
+    }
+
+  }
+
+  /**
+   * Angular Lifecycle hook for newer Angular versions.
+   * Bindings are no longer assigned in the controller, but in the $onInit hook.
```
gkalpak: Maybe: "Bindings are not guaranteed to have been assigned in the controller, but they are in the `$onInit` hook."

src/components/fabSpeedDial/fabController.js
```diff
@@ -26,7 +26,20 @@
...
+     */
+    vm.$onInit = function() {
+      setupDefaults();
+    };
+
+    // For Angular 1.4 and older, where there are no lifecycle hooks but bindings are pre-assigned,
+    // manually call the $onInit hook.
+    if (angular.version.major === 1 && angular.version.minor <= 4) {
+      this.$onInit();
+    }
```
gkalpak: It feels a little odd to call `$onInit()` before the end of the controller initialization. It will also make the difference between 1.4 vs 1.6 more significant. AFAICT, it doesn't matter with the current implementation, but it makes the code more brittle (e.g. a change inside `setupListeners/Watchers()` or `fireInitialAnimations()` might have unexpected results.

Was there a specific reason for not moving all initialization inside `$onInit()` or not at least calling `$onInit()` at the end of the controller constructor?

mgol:
Still LGTM.

src/components/datepicker/js/datepickerDirective.js
```diff
@@ -387,9 +379,34 @@
         }
       });
     }
+
+    // For Angular 1.4 and older, where there are no lifecycle hooks but bindings are pre-assigned,
+    // manually call the $onInit hook.
+    if (angular.version.major === 1 && angular.version.minor <= 4) {
```
crisbeto: Might be a good idea to move this into a constant so it's easier to re-use.

petebacondarwin:
Can we get this merged @ThomasBurleson ?

src/components/datepicker/js/datepickerDirective.js
mgol: We can improve on that, sure, but let's not delay this PR further as we need it landed for the Core 1.6 sync.

PR #10001, opened 2016-11-10
$mdUtil.now() not working on android

dejakob:
I noticed there were a lot of bugs logged on our website from android devices on the now method. I refactored it a bit...

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -71,7 +71,7 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
     }
 
     return function() {
-      return (new Date()).getTime();
+      return new Date().getTime();
     }
   }
 
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -63,7 +63,7 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
    */
   function getNowMethod() {
     if (window.performance && typeof window.performance.now === 'function') {
-      return angular.bind(window.performance, window.performance.now);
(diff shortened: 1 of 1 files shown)


src/core/util/util.js
```diff
@@ -57,12 +57,27 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
+    if (window.performance && typeof window.performance.now === 'function') {
+      return angular.bind(window.performance, window.performance.now);
+    }
+
+    if (typeof Date.now === 'function') {
+      return Date.now;
+    }
+
+    return function() {
+      return (new Date()).getTime();
```
DevVersion: The parentheses can be removed around the  `new Date()`

src/core/util/util.js
```diff
@@ -57,12 +57,27 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
 
   }
 
+  /**
+   * Bulletproof method to get a function to return the current timestamp
+   * @returns {Function}
+   */
+  function getNowMethod() {
+    if (window.performance && typeof window.performance.now === 'function') {
+      return angular.bind(window.performance, window.performance.now);
```
DevVersion: We should be able to use the `Function.bind` method, since it supports all browsers we support.
> See https://developer.mozilla.org/de/docs/Web/JavaScript/Reference/Global_Objects/Function/bind

```js
return window.performance.now.bind(window.performance)
```

ThomasBurleson:
@dejakob  - please rebase with HEAD and squash your commits. This LGTM.

dejakob:
Closed because no longer applicable

PR #10007, opened 2016-11-11
feat(panel): Panel configuration provider

### LINKED ISSUE #10006: feat(panel): Panel configuration provider
**Actual Behavior**:
 -  `What is the issue? *` There is no way to create default configurations for panel elements.
 -  `What is the expected behavior?` A user should be able to add a provider to the config block of their application and create one or more custom configurations for upcoming panel elements.

 -  `Details:` N/A

**Angular Versions**: *
 -  `Angular Version:` Latest

**Additional Information**:
 -  `Browser Type: *` All
 -  `Browser Version: *` All
 -  `OS: *` All
 -  `Stack Traces:` All

bradrich:
Allows users to be able to add the `$mdPanelConfigProvider` to the config block of their application. They can use this provider to create one or more configurations for upcoming panel elements.

Fixes #10006

### DIFF
--- a/src/components/panel/demoConfigProvider/index.html
+++ b/src/components/panel/demoConfigProvider/index.html
@@ -1,0 +1,51 @@
+<div ng-controller="PanelProviderCtrl as ctrl" ng-cloak>
+
+  <md-content layout-padding>
+
+    <div layout="row" layout-align="space-between center" flex>
+      <md-button
+        class="md-fab md-primary"
+        aria-label="Navigation"
+        ng-click="ctrl.showMenu($event, ctrl.navigation)">
+        <md-icon md-svg-icon="img/icons/menu.svg"></md-icon>
+      </md-button>
+      <md-button
+        class="md-fab md-accent"
+        aria-label="Favorites"
+        ng-click="ctrl.showMenu($event, ctrl.favorites)">
+        <md-icon md-svg-icon="img/icons/favorite.svg"></md-icon>
+      </md-button>
+      <md-button
+        class="md-fab md-background"
+        aria-label="More"
+        ng-click="ctrl.showMenu($event, ctrl.more)">
+        <md-icon md-svg-icon="img/icons/more_vert.svg"></md-icon>
+      </md-button>
+    </div>
+
+    <p>
+      Configurations for upcoming panel elements can be created at the
+      <code>.config</code> level of your Angular application. To create a
+      configuration object, use the <code>$mdPanelConfigProvider</code>
+      dependency within a <code>.config</code> method and call
+      <code>$mdPanelConfigProvider.createConfig</code> with an object containing
+      a configuration name and options as it's argument. This object will be
+      stored for you so that you can retrieve it by using the
+      <code>$mdPanel.getConfig</code> method with the name of the configuration
+      object that you created as it's argument.
+    </p>
+    <p>
+      The configuration object takes all of the options found within the
+      <code>$mdPanel.create</code> method; however, it does not make sense to
+      include any options that depend on user interaction, panel positioning, or
+      panel animation.
+    </p>
+    <p>
+      This will help you reduce the necessary lines of configuration code that
+      are required to create a panel when you are wanting to have multiple
+      panels that are largely the same.
+    </p>
+
+  </md-content>
+
+</div>
--- a/src/components/panel/demoConfigProvider/script.js
+++ b/src/components/panel/demoConfigProvider/script.js
@@ -1,0 +1,95 @@
+(function() {
+  'use strict';
+
+  angular
+      .module('panelProviderDemo', ['ngMaterial'])
+      .config(PanelProviderConfig)
+      .controller('PanelProviderCtrl', PanelProviderCtrl)
+      .controller('PanelMenuCtrl', PanelMenuCtrl);
+
+  function PanelProviderConfig($mdPanelConfigProvider) {
+    var config = {
+      name: 'demoConfig',
+      options: {
+        attachTo: angular.element(document.body),
+        controller: PanelMenuCtrl,
+        controllerAs: 'ctrl',
+        template: '' +
+            '<div class="menu-panel" md-whiteframe="4">' +
+            '  <div class="menu-content">' +
+            '    <div class="menu-item" ng-repeat="item in ctrl.items">' +
+            '      <button class="md-button">' +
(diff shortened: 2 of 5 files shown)


src/components/panel/panel.spec.js
```diff
@@ -94,6 +94,84 @@ describe('$mdPanel', function() {
...
+      name: 'testConfig',
+      options: DEFAULT_CONFIG
+    };
+    var config2 = {
+      name: 'testConfig2',
+      options: DEFAULT_CONFIG
+    };
+
+    it('should have the $mdPanelConfigProvider available',
+        inject(function($mdPanelConfig) {
```
ErinCoughlan: Tests shouldn't have inject within in. These should all be in a beforeEach at the top of the file.

src/components/panel/demoConfigProvider/index.html
```diff
@@ -0,0 +1,51 @@
...
+      dependency within a <code>.config</code> method and call
+      <code>$mdPanelConfigProvider.createConfig</code> with an object containing
+      a configuration name and options as it's argument. This object will be
+      stored for you so that you can retrieve it by using the
+      <code>$mdPanel.getConfig</code> method with the name of the configuration
+      object that you created as it's argument.
+    </p>
+    <p>
+      The configuration object takes all of the options found within the
+      <code>$mdPanel.create</code> method; however, it does not make sense to
```
ErinCoughlan: Rather it's impossible to include options related to user interaction, panel positioning, or panel animations.

src/components/panel/demoConfigProvider/script.js
```diff
@@ -0,0 +1,95 @@
...
+    this.more = {
+      name: 'more',
+      items: [
+        'Account',
+        'Sign Out'
+      ]
+    };
+
+    $mdPanel.newPanelGroup('menus', {
+      maxOpen: 2
```
ErinCoughlan: I'm thinking that this group config can have all the options you are trying to initialize before hand. The only case, to me, that makes sense for a shared config is groups, so having the logic associated seems better than in separate parts of the application.

src/components/panel/demoConfigProvider/script.js
```diff
@@ -0,0 +1,95 @@
...
+    });
+
+    this.showMenu = function($event, menu) {
+      var config = $mdPanel.getConfig('demoConfig');
```
ErinCoughlan: So you have to get the config in order to extend it? I think it's better for this to be automatic. This is extra steps a user needs to take.

If I'm writing an application and want to do this, it makes more sense for more to make a global constant for the object than use this provider.

src/components/panel/demoConfigProvider/script.js
```diff
@@ -0,0 +1,95 @@
...
+      };
+      config.openFrom = $event;
+
+      $mdPanel.open(config);
+    };
+  }
+
+  function PanelMenuCtrl(mdPanelRef) {
+    this.closeMenu = function() {
+      mdPanelRef && mdPanelRef.close();
```
ErinCoughlan: Depending on reuse, you may also want ```.then(ref => ref.destroy());```

src/components/panel/panel.js
```diff
@@ -17,6 +18,118 @@ angular
...
+ *   }
+ * })(angular);
+ * </hsjs>
+ */
+
+/**
+ * @ngdoc method
+ * @name $mdPanelConfig#createConfig
+ * @description
+ * Takes the passed in panel configuration object and adds it to the `_configs`
```
ErinCoughlan: The user shouldn't be aware of private variables or how we are storing the configs. Please reword.

src/components/panel/panel.js
```diff
@@ -17,6 +18,118 @@ angular
...
+ * object of the provider.
+ *
+ * @param {!Object<string, !Object>} config Specific configuration object that
```
ErinCoughlan: I'm not sure if we defined a type for the config, but if we did it would be useful to write here along with which options are not supported.

src/components/panel/panel.js
```diff
@@ -17,6 +18,118 @@ angular
...
+ *     is required to contain a name string and an options object. The options
+ *     object can contain any and all of the parameters available within the
+ *     `$mdPanel.create` configuration. However, it does not make sense to
+ *     include parameters that depend on certian interactions of the user,
```
ErinCoughlan: Same comment as above. Not possible, not doesn't make sense.

src/components/panel/panel.js
```diff
@@ -784,22 +897,71 @@ angular
...
 var FOCUS_TRAP_TEMPLATE = angular.element(
     '<div class="_md-panel-focus-trap" tabindex="0"></div>');
 
 
 /**
+ * A provider that is used for building configurations for the panel.
+ * @final @constructor @ngInject
+ */
+function MdPanelConfigProvider() {
+  /** @private {!Object} */
```
ErinCoughlan: I think this is supposed to be a map? Please fix the type.

src/components/panel/panel.js
```diff
@@ -784,22 +897,71 @@ angular
...
+ * Takes the passed in panel configuration object and adds it to the `_configs`
+ * object of the provider.
+ * @param {!Object<string, !Object>} config Specific configuration object that
+ *     is required to contain a name string and an options object. The options
+ *     object can contain any and all of the parameters available within the
+ *     `$mdPanel.create` configuration. However, it does not make sense to
+ *     include parameters that depend on certain interactions of the user,
+ *     positioning, or animation.
+ */
+MdPanelConfigProvider.prototype.createConfig = function(config) {
```
ErinCoughlan: I think `createConfig` should take two params to help avoid the confusion with fields should be in the config. You're creating one at a time so it really shouldn't matter.

Also, consider a builder if you want to chain the creates.

```createConfig(name, config)```

File: src/components/panel/panel.js [hunk 11]
```diff
@@ -784,22 +897,71 @@ angular
...
+  }
+};
+
+
+/**
+ * Default $get method required by an Angular provider.
+ * @method $get
+ * @returns {!Object} An object that expresses the provider dependency.
+ */
+MdPanelConfigProvider.prototype.$get = function() {
```
ErinCoughlan: If it's a provider, you shouldn't need to override this method. The default one should suffice. There's nothing special here.

PR #10015, opened 2016-11-14
fixes the validation for date picker in IE

### LINKED ISSUE #9994: datepicker: validation behaves differently in IE
**Actual Behavior**:
You can test it here in angular material demos for IE and Chrome:
https://material.angularjs.org/1.1.1/demo/datepicker
 The issue is when you input an invalid value in the date picker in IE with ngMessages, it says "Only Weekends are allowed", then when you add a letter in the input, it doesn't validate it (for example 11/9/2016 and after that input a letter, 11/9/2016t). I was expecting to have this message showed "The entered value is not a date!" but it didn't. It wasn't happening when I was using Angular Material 1.0.8.
In Chrome, it validates correctly so if you add a letter in datepicker input, it says "The entered value is not a date!". This only became an issue from angular material 1.1.0 in IE.

** steps to reproduce the issue**
 -  https://material.angularjs.org/1.1.1/demo/datepicker
 -  `Details:` adding a letter in input "With ngMessages", the validation message stays "Only Weekends are allowed" instead of "The entered value is not a date!".

richardhughes:
Adds in validation for the input as a last check as IE interpreted the date '11/14/2016d' as valid which is incorrect.

closes #9994

### DIFF
--- a/src/components/datepicker/js/datepickerDirective.js
+++ b/src/components/datepicker/js/datepickerDirective.js
@@ -571,7 +571,29 @@
       this.ngModelCtrl.$setValidity('valid', date == null);
     }
 
+    var input = this.inputElement.value;
+    var parsedDate = this.locale.parseDate(input);
+
+    if (!this.isInputValid(input, parsedDate) && this.ngModelCtrl.$valid) {
+      this.ngModelCtrl.$setValidity('valid', date == null);
+    }
+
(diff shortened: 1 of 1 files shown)


topherfangio:
This looks good, but can you add some unit tests that show how this fixes the failure?

ThomasBurleson:
@topherfangio - can you add 1-2 unit tests for this PR?

ThomasBurleson:
Closing as stale PR.

PR #10017, opened 2016-11-14
fix(progressCircular): update animation to spec

### LINKED ISSUE #9879: progressCircular: spinner is too slow relative to spec
feedback from the core design team -- ProgressCircular is about 3x too slow.

have shared a better example with @ThomasBurleson. Can we get this updated?

See https://material.angularjs.org/latest/demo/progressCircular

clshortfuse:
Previous animation did not properly animate the circle stroke

* Use persistent SVG path for all animations
* Change `stroke-dashoffset` to match spec on every requested frame
* Rotate object -90 degrees every iteration
* Correct overall counter clockwise animation timing in SCSS

Instead of creating a new SVG path every frame, we will instead simply
animate the `stroke-dashoffset` parameter. This allows us to match spec
while also increasing performance by simplifying logic per frame.

Fixes #9879

### DIFF
--- a/src/components/progressCircular/js/progressCircularDirective.js
+++ b/src/components/progressCircular/js/progressCircularDirective.js
@@ -62,7 +62,6 @@ function MdProgressCircularDirective($window, $mdProgressCircular, $mdTheming,
             $window.webkitCancelRequestAnimationFrame ||
             angular.noop;
 
-  var DEGREE_IN_RADIANS = $window.Math.PI / 180;
   var MODE_DETERMINATE = 'determinate';
   var MODE_INDETERMINATE = 'indeterminate';
   var DISABLED_CLASS = '_md-progress-circular-disabled';
@@ -103,7 +102,7 @@ function MdProgressCircularDirective($window, $mdProgressCircular, $mdTheming,
     var path = angular.element(node.querySelector('path'));
     var startIndeterminate = $mdProgressCircular.startIndeterminate;
     var endIndeterminate = $mdProgressCircular.endIndeterminate;
-    var rotationIndeterminate = 0;
+    var iterationCount = 0;
     var lastAnimationId = 0;
     var lastDrawFrame;
     var interval;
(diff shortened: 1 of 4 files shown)


clshortfuse:
It's a little difficult to see on a recording, but you should be able to see the animation works differently now.

This PR:

progresscircle pr

Master:
progresscircle master

crisbeto:
To be honest, I doubt that this will have much of an effect on performance. If anything, it may end up being slightly slower, because while it does do less math, it does more DOM manipulation (2 attributes for every frame, as opposed to one with the old approach). What does this approach let us do that the old one didn't?

clshortfuse:
@crisbeto By changing the path, you're telling the browser to render the entire path again, which internally means, interpolation of points to screen and all the mathematics involved with it.

Changing rotation, or adding dash strokes doesn't change the underlying rendered object's framework. 

Regardless, I ran a benchmark and path manipulation is about 4x slower: https://jsperf.com/svg-dash-vs-path

Also, the transform attribute is only changed every iteration, which is 1333ms now, so the majority of time it's still one DOM manipulation. I'm also not counting the reduced complexity from removing the need to run sin/cos functions.

The animation is exactly as spec. It's not something I came up with to try to match it, but it's the actual logic rewritten for RAF.

The only thing I'm not sure is correct is the easing curve (since material ease is [0.4, 0, 0.2, 1] but I decided against changing it. If it's not 100% accurate, it's still pretty close.

crisbeto:
Alright, I see what you mean. Regarding the extra rotation, couldn't you do it with CSS?

clshortfuse:
@crisbeto There was a reason why I opted for SVG transform vs CSS transform. IIRC, it has to do with animation synchronization (not mixing SVG with CSS) and not having to deal with the funky transform origin on IE.

crisbeto:
It makes sense if it needs to be synchronized, otherwise the transform origin is pretty easy to work around: you have to set it with values explicitly and in all 3 dimensions, e.g. `transform-origin: 150px 150px 150px`.

ThomasBurleson:
@clshortfuse - status plz ?

PR #10023, opened 2016-11-15
fix(select): block xss on md-select-label

jelbourn:


src/components/select/select.js
```diff
@@ -337,17 +338,41 @@ function SelectDirective($mdSelect, $mdUtil, $mdConstant, $mdTheming, $mdAria, $
...
+          isSelectLabelFromUser = true;
         }
 
         var target = valueEl.children().eq(0);
-        target.html(text);
+
+        if (attr.mdSelectedHtml) {
+          if (!$injector.has('$sanitize')) {
+              throw Error('The ngSanitize module must be loaded in order to use ' +
+                  'md-treat-selected-text-as-html.');
```
rjamet: If there's no $sanitize, the user will get an [$sce:unsafe](https://docs.angularjs.org/error/$sce/unsafe) error from getTrustedHtml, which points to the sanitizer already. It can make sense to not have it (codesize), and rely on $sce.trustAsHtml only instead, and you'll fail hard for them at this point. So from a purely security POV I wouldn't throw here or look for $sanitize, and just let $sce.getTrustedHtml do its thing.

rjamet:
Besides from the small comment, looks good :) thanks for the super-quick patch !

jelbourn:
@ThomasBurleson @ErinCoughlan can you review?

src/components/select/select.js
```diff
@@ -337,17 +343,39 @@ function SelectDirective($mdSelect, $mdUtil, $mdConstant, $mdTheming, $mdAria, $
...
         }
 
         var target = valueEl.children().eq(0);
-        target.html(text);
+
+        if (attr.mdSelectedHtml) {
+            // Using getTrustedHtml will run the content through $sanitize if it is not already
+            // explicitly trusted. If the ngSanitize module is not loaded, this will
+            // *correctly* throw an sce error.
+            target.html($sce.getTrustedHtml(text));
```
ThomasBurleson: Nice 👍

ThomasBurleson:
@jelbourn - lgtm.

Splaktar:
This was a breaking change for users upgrading from 1.1.1 to 1.1.4/1.1.5 as mentioned in https://github.com/angular/material/issues/10912. I'm going to update the changelog to add this to the breaking changes for 1.1.2.

PR #10024, opened 2016-11-16
fix(icon): fix aria roles and attributes

### LINKED ISSUE #9629: icon: a11y support appears broken
Please see review here:

Aria-label has no effect.

``` js
    var label = attr.alt || attr.mdFontIcon || attr.mdSvgIcon || element.text();
    if ( !attr['aria-label'] ) {
    if (label !== '' && !parentsHaveText() ) {
```

Both aria-label and label are empty.
1. The logic is flawed.
   
   `if (label !== '' && !parentsHaveText() ) {`

Just because the parents have text does not mean that that text is going to be announced. It needs to be checked that the parent is a control or some kind of element such that the icon is not relevant.

clshortfuse:
mdIcon was improperly assuming its parent's text content would be announced didn't apply ARIA roles properly

* Apply `img` role by default
* Don't change ARIA if it's already valid
* Prioritize `alt` attribute for aria-label
* Apply `aria-hidden` attr if parent or parent's parent has ARIA label and is not part of role or tagName whitelist
* Fallback to icon name as an aria label
* Apply `aria-hidden` attr instead of `aria-hidden` when all checks fail
* Add spec tests to ensure new functionality
* Remove improper spec test based on text content

Fixes #9629

### DIFF
--- a/src/components/icon/js/iconDirective.js
+++ b/src/components/icon/js/iconDirective.js
@@ -218,20 +218,29 @@ function mdIconDirective($mdIcon, $mdTheming, $mdAria, $sce) {
     // If using a font-icon, then the textual name of the icon itself
     // provides the aria-label.
 
-    var label = attr.alt || attr.mdFontIcon || attr.mdSvgIcon || element.text();
     var attrName = attr.$normalize(attr.$attr.mdSvgIcon || attr.$attr.mdSvgSrc || '');
 
-    if ( !attr['aria-label'] ) {
+    /* Provide a default accessibility role of img */
+    if (!attr.role) {
+      $mdAria.expect(element, 'role', 'img');
+      /* manually update attr variable */
+      attr.role = 'img';
+    }
 
-      if (label !== '' && !parentsHaveText() ) {
-
-        $mdAria.expect(element, 'aria-label', label);
-        $mdAria.expect(element, 'role', 'img');
-
-      } else if ( !element.text() ) {
-        // If not a font-icon with ligature, then
-        // hide from the accessibility layer.
-
+    /* Don't process ARIA if already valid */
+    if ( attr.role === "img" && !attr.ariaHidden && !$mdAria.hasAriaLabel(element) ) {
+      var iconName;
+      if (attr.alt) {
+        /* Use alt text by default if available */
+        $mdAria.expect(element, 'aria-label', attr.alt);
+      } else if ($mdAria.parentHasAriaLabel(element, 2)) {
+        /* Parent has ARIA so we will assume it will describe the image */
+        $mdAria.expect(element, 'aria-hidden', 'true');
+      } else if (iconName = (attr.mdFontIcon || attr.mdSvgIcon || element.text())) {
+        /* Use icon name as aria-label */
+        $mdAria.expect(element, 'aria-label', iconName);
+      } else {
+        /* No label found */
         $mdAria.expect(element, 'aria-hidden', 'true');
       }
     }
@@ -247,19 +256,7 @@ function mdIconDirective($mdIcon, $mdTheming, $mdAria, $sce) {
             element.append(svg);
           });
         }
-
       });
-    }
-
-    function parentsHaveText() {
-      var parent = element.parent();
-      if (parent.attr('aria-label') || parent.text()) {
-        return true;
-      }
-      else if(parent.parent().attr('aria-label') || parent.parent().text()) {
-        return true;
-      }
-      return false;
     }
 
     function prepareForFontIcon() {
--- a/src/components/icon/icon.spec.js
+++ b/src/components/icon/icon.spec.js
@@ -301,6 +301,16 @@ describe('MdIcon directive', function() {
 
     describe('with ARIA support', function() {
 
+      it('should apply "img" role by default', function() {
+        el = make('<md-icon md-svg-icon="android" ></md-icon>');
+        expect(el.attr('role')).toEqual('img');
+      });
+
+      it('should apply not replace current role', function() {
+        el = make('<md-icon md-svg-icon="android" role="presentation" ></md-icon>');
+        expect(el.attr('role')).toEqual('presentation');
+      });
+
       it('should apply aria-hidden="true" when parent has valid label', function() {
         el = make('<button aria-label="Android"><md-icon md-svg-icon="android"></md-icon></button>');
         expect(el.find('md-icon').attr('aria-hidden')).toEqual('true');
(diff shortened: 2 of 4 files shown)


clshortfuse:
@ThomasBurleson @nbtorcol I haven't built the unit tests yet since I want ensure this is the proper logic before I do.

nbtorcol:
Let me get back to you tomorrow.

nbtorcol:
This is better, but it is still not correct. I will try to explain more
what is wrong.



• Apply  img  role by default -- yes

• Don't change ARIA if it's already valid -- yes

• Prioritize  alt  attribute for aria-label -- yes

• Apply  presentation  role if parent or parent's parent has ARIA label –
no, see note 1 below

• Fallback to icon name as an aria label -- yes

• Apply  presentation  role instead of  aria-hidden  when all checks fail –
no, see note 2 below



1.       First and foremost, the original—aria-hidden—was better. If it is
not of significance to screen readers, then just make it go away
completely; don’t turn it into something else. Second, as I mentioned
before, this logic is flawed. Yes, the ARIA spec says that you can put
aria-label/aria-labelledby/aria-describedby on any element, regardless of
tagName or role, but that is not true for some of the Windows screen
readers, or more correctly the browsers that provide the information to the
screen readers. Please consider reading this article:
http://www.ssbbartgroup.com/blog/how-browsers-interact-with-screen-readers-and-where-aria-fits-in-the-mix/.
There are several ARIA roles that support these ARIA attributes, but, for
this situation, the best criteria would be to only use the ARIA roles that
are controls. I sent example code to this effect, and I do not understand
why it was not even used as a starting point. If this turns out to be to
complex, then I would recommend just not trying to do this sort of analysis
and making sure that aria-hidden works, so that people can hide the icons
on their own if they think it necessary.

2.       As above, aria-hidden would be better.

clshortfuse:
@nbtorcol I'm sorry I didn't use the code your provided (on the issue page), but that was because it was private link on a Google Drive when I started writing the PR (for future reference, something like pastebin.com would serve better). The use of `presentation=role` instead of `aria-hidden` is from research that it is more reliable than `aria-hidden` (especially with IE and JAWS). After more research, it seems I can safely use both (and even possible `tabindex=-1`). See: http://john.foliot.ca/aria-hidden/

If I understand correctly, you suggest we should only perform a parent aria check based on a whitelist of tagNames and/or or ARIA. It would reduce the specificity of our upwards search.

As for the third point of respecting `aria-hidden, I have added a check for the mdIcon itself, but I never checked the parent. I will add this to the code.

nbtorcol:
@nbtorcol <https://github.com/nbtorcol> I'm sorry I didn't use the code
your provided (on the issue page), but that was because it was private link
on a Google Drive when I started writing the PR (for future reference,
something like pastebin.com would serve better).



Apologies, that was partially an oversight on my part. I will look at that
website, but what I use is driven almost entirely by the level of
accessibility. Given that I am doing this as a side project at work, I
don’t have a lot of time to learn new stuff.



The use of presentation=role instead of aria-hidden is from research that
it is more reliable than aria-hidden (especially with IE and JAWS). After
more research, it seems I can safely use both (and even possible tabindex=-1
).



The tabIndex attribute is only relevant for controls and should not be used
with icons.



See: http://john.foliot.ca/aria-hidden/



That article is four years old; both the screen readers and browsers have
changed a lot in that time. Please base your conclusions off of more recent
articles or conduct your own experiments; NVDA is free, and JAWS has a
demo. I can help test if needed and/or point you to relevant articles and
information from people.



If I understand correctly, you suggest we should only perform a parent aria
check based on a whitelist of tagNames and/or or ARIA. It would reduce the
specificity of our upwards search.



Yes.



As for the third point of respecting `aria-hidden, I have added a check for
the mdIcon itself, but I never checked the parent. I will add this to the
code.



Not sure what you mean here; sorry if my prior comments were confusing.

ThomasBurleson:
@clshortfuse - status update plz ?

clshortfuse:
@nbtorcol I've added the changed we've discussed. I removed `role=presentation` since that from an old version of NVDA that didn't support `aria-hidden`. 

I'm a little confused about the whitelisting `type` attribute, since as far as I could tell, we'd pretty much be performing HTML validation for valid inputs, since we're are listing every possible `type` value. 

I feel a blacklist would work better, or else our code will eventually become incompatible when newer versions of the HTML spec and when new ARIA roles are created.

nbtorcol:
Okay. Give me a little time to put together the blacklist.



*From:* Carlos Lopez [mailto:notifications@github.com]
*Sent:* Thursday, December 08, 2016 9:19 PM
*To:* angular/material <material@noreply.github.com>
*Cc:* nbtorcol <tnicole@google.com>; Mention <mention@noreply.github.com>
*Subject:* Re: [angular/material] fix(icon): fix aria roles and attributes
(#10024)



@nbtorcol <https://github.com/nbtorcol> I've added the changed we've
discussed. I removed role=presentation since that from an old version of
NVDA that didn't support aria-hidden.

I'm a little confused about the whitelisting type attribute, since as far
as I could tell, we'd pretty much be performing HTML validation for valid
inputs, since we're are listing every possible type value.

I feel a blacklist would work better, or else our code will eventually
become incompatible when newer versions of the HTML spec and when new ARIA
roles are created.

—
You are receiving this because you were mentioned.
Reply to this email directly, view it on GitHub
<https://github.com/angular/material/pull/10024#issuecomment-265935980>,
or mute
the thread

ThomasBurleson:
@clshortfuse - let's finish this PR asap or close it.

clshortfuse:
@jelbourn Can you review this? The solution is ready to merge with the current whitelist solution. Blacklist could be more future-proof, but not entirely needed.

nbtorcol:
I will get the blacklist to you today. I was gone on vacation for a while
and then had a lot to do when I got back.

nbtorcol:
Do you still have access to the Google doc that I made before? I updated
the code there.

clshortfuse:
@nbtorcol Yes, I do have access. I'm updating the PR now. Thank you!

nbtorcol:
Sounds good. Please let me know if you have any questions.

clshortfuse:
@nbtorcol I just want to confirm a couple of concerns

* ARIA ROLES `menu` and `scrollbar` are now being blacklisted, but were whistlisted before
* INPUT TYPES  `number` and `time` are now being blacklisted, but were whistlisted before
* HTML TAGS `menu` and `time` are now being blacklisted, but were whitelisted before
* I couldn't find where to put `tu`, since it's neither a role, type, or tag.

Perhaps we should 

* return false if aria role blacklist fails
* return true if any input type is used
* return false if html tag blacklist fails

Also, I would need some clarification about hwo to use the shared role and tag names.

nbtorcol:
Oops, number was supposed to not be part of the blacklist. Sorry.

Time, as an HTML tag fails.

Some of the ARIA roles that were included but now blacklisted are because I
ran an actual test with screen readers, and those are the results.



   - return false if aria role blacklist fails
   - return true if any input type is used
   - return false if html tag blacklist fails



Sounds good. In that case, per my code, you can just drop retrieving the
type attribute.

clshortfuse:
@nbtorcol @ThomasBurleson All changes have been made and we are ready for review.

nbtorcol:
Do you mean committed?

src/components/icon/js/iconDirective.js
```diff
@@ -247,19 +256,135 @@ function mdIconDirective($mdIcon, $mdTheming, $mdAria, $sce) {
...
+    function checkAriaAttributes(domAttrs) {
+      return domAttrs['aria-label'] || domAttrs['aria-labelledby'] || domAttrs['aria-describedby'];
+    }
+
+    function performAriaCheck(domElement) {
+      if (!checkAriaAttributes(domElement.attributes)) {
+        return false;
+      }
+      /* Perform role blacklist check */
+      if (domElement.attributes.role) {
```
ThomasBurleson: Please move this to the aria utils. This complexity should not be in the iconDirective.

clshortfuse:
@nbtorcol Generally, we ask another team member to check our work before we commit directly to our master branch

@ThomasBurleson I made the changes, as requested. Both functions are now accessible by any component via the $mdAria service. Also, the number of parents searches are based on a passed variable.

ThomasBurleson:
LGTM

PR #10031, opened 2016-11-17
fix(mdPanel): Panel and tooltip theming

### LINKED ISSUE #10030: fix(mdPanel): Theming is not initialized on certain panel implementations.
**Actual Behavior**:
 -  `What is the issue? *` If theming doesn't get initialized on `mdPanel` components, i.e. `mdTooltip`, then the color styling falls back to the default `mdPanel` styling.
 -  `What is the expected behavior?` Theming should always be initialized.

 -  Before theming problem:
pasted image at 2016_11_16 05_13 pm
 -  After theming problem:
pasted image at 2016_11_16 05_13 pm 1

**Angular Versions**: *
 -  `Angular Version:` Latest
 -  `Angular Material Version:` Master

**Additional Information**:
 -  `Browser Type: *` All
 -  `OS: *` All
 -  `Stack Traces:` N/A

----

  Add more lines if needed.

bradrich:
- Initialized theming through `$mdTheming` on the panel element and container after the panel is created.
 - Fixed the panel and tooltip theme Sass files to remove unthemed styles and complicated syntax.
 - Removed an unnecessary `$mdTheming` call from the tooltip directive.

Fixes #10030

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -453,15 +453,15 @@ angular
  * @returns {!MdPanelRef}
  */
 
- /**
-  * @ngdoc method
-  * @name MdPanelRef#updateAnimation
-  * @description
-  * Updates the animation configuration for a panel. You can use this to change
-  * the panel's animation without having to re-create it.
-  *
-  * @param {!MdPanelAnimation} animation
-  */
+/**
+ * @ngdoc method
+ * @name MdPanelRef#updateAnimation
+ * @description
+ * Updates the animation configuration for a panel. You can use this to change
+ * the panel's animation without having to re-create it.
+ *
+ * @param {!MdPanelAnimation} animation
+ */
 
 
 /*****************************************************************************
@@ -1108,6 +1108,9 @@ function MdPanelRef(config, $injector) {
   /** @private @const {!angular.$mdUtil} */
   this._$mdUtil = $injector.get('$mdUtil');
 
+  /** @private @const {!angular.$mdTheming} */
+  this._$mdTheming = $injector.get('$mdTheming');
+
   /** @private @const {!angular.Scope} */
   this._$rootScope = $injector.get('$rootScope');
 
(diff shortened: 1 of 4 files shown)


ErinCoughlan:
This appears to fix the tooltip, but not the panel backdrop. The theme is still not applied to the backdrop.

bradrich:
It seemed to work in my tests. How are you isolating it to get these results?

bradrich:
Were you able to see where the `$mdTheming` calls were made on the `panelContainer` and the `panelEl`? The `panelContainer` is the backdrop of the panel API?

ErinCoughlan:
I tested this by running the presubmit and patching it into our code where I could see the tooltip and a standalone panel.

So you can also see the standalone panel, if you temporarily add `hasBackdrop` to the tooltip panel config, no backdrop appears.

ErinCoughlan:
For more information, `mdTheming` is called for the tooltip panel and the tooltip outer wrapper, but not for the backdrop panel or the backdrop outer wrapper.

ErinCoughlan:
Ahha! Got it!

For the backdrop, there is no positionConfig, so the `addStyles` method resolves without hitting the `$mdTheming` call.

If you add the two `$mdTheming` lines at the beginning of `hideAndResolve` in addition to the current location, the backdrop appears the right color.

bradrich:
Very cool. I am updating this now.

bradrich:
Would it be better to put it here:  https://github.com/angular/material/blob/master/src/components/panel/panel.js#L1662? The `hideAndResolve` function could be called more than once. However, I don't think that the `$mdTheming` call happening more than once is a bad thing...

bradrich:
@ErinCoughlan Give this a check and see what you think.

src/components/panel/panel.js
```diff
@@ -1679,18 +1687,33 @@ MdPanelRef.prototype._addStyles = function() {
...
+
       resolve(self);
     });
   });
 };
 
 
 /**
+ * Sets the `$mdTheming` classes on the `panelContainer` and `panelEl`.
+ * @method setTheming
```
ErinCoughlan: This line isn't needed, `@method` is only used for the ngdoc, but this isn't public.

ThomasBurleson:
@jelbourn - important fix.

PR #10056, opened 2016-11-23
feat(dialog): optionally disable confirm in prompt

oliversalzburg:
Allows the user to call `.require()` on the dialog options to have the "Okay" button disabled in a prompt until a value is entered.

Please me know if I'm on the right path with this :P

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -475,6 +475,7 @@ function MdDialogDirective($$rAF, $mdTheming, $mdDialog) {
  * - $mdDialogPreset#initialValue(string) - Sets the initial value for the prompt input.
  * - $mdDialogPreset#ok(string) - Sets the prompt "Okay" button text.
  * - $mdDialogPreset#cancel(string) - Sets the prompt "Cancel" button text.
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@oliversalzburg - you added the `required` to the preset for `confirm`, yet the template modifications checks for prompt:

```html
ng-disabled="{{ ::dialog.$type == \'prompt\' && ::dialog.required && !dialog.result }}"
```

oliversalzburg:
@ThomasBurleson The wrong documentation should be fixed. I started on the tests, but I ran into an issue I don't understand.

```
Error: [$parse:syntax] Syntax Error: Token ':' not a primary expression at column 29 of the expression [dialog.$type == 'prompt' && ::dialog.required && !dialog.result] starting at [::dialog.required && !dialog.result].
```

I don't understand why it doesn't like the `::` here, which is used all over the place in the template.

oliversalzburg:
@mdklab Did 2015ae8c9098d17c6ee9dc46859d62026071eae7 make this obsolete?

Splaktar:
Yes, I think that https://github.com/angular/material/pull/10626 implemented this and solved https://github.com/angular/material/issues/10135. Unfortunately that PR didn't include any tests.

PR #10061, opened 2016-11-24
fix(autocomplete): change highlight color to match material standards

### LINKED ISSUE #10060: Highlight hardly highlights anything
**Actual Behavior**:
When using `md-highlight-text`, the resulting highlight isn't actually highlighting anything. At least the contrast is so low, that hardly any effect is visible.

image

You can observe the problem directly at https://material.angularjs.org/HEAD/demo/autocomplete

**Additional Information**:
Chrome stable, Windows

Drumarcco:
Change highlight color to provide a better contrast on highlighted text on items.

Closes #10060

[Standard example](https://material.google.com/components/text-fields.html#text-fields-auto-complete-text-field)
image

With this PR:
image

### DIFF
--- a/src/components/autocomplete/autocomplete-theme.scss
+++ b/src/components/autocomplete/autocomplete-theme.scss
@@ -19,7 +19,7 @@ md-autocomplete.md-THEME_NAME-theme {
   li {
     color: '{{background-900}}';
     .highlight {
-      color: '{{background-600}}';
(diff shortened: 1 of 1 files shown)


clshortfuse:
I believe we should be using foreground colors and opacities rather than background colors (since they are meant to be used with text). Not using them is why we have this issue in the first place.

PR #10079, opened 2016-12-02
feat(input): add animation to color change

nnmrts:
feat(input): add animation to color change

add animation to color change when clicking inside an `md-input` field
previous behaviour was an instant color change, which doesn't harmonize with material design guidelines


nnmrts:
I signed it!

clshortfuse:
@nnmrts Do you have a source for these bezier curve numbers?

nnmrts:
@clshortfuse actually not, just used a random bezier curve, but perhaps we can use some of these:
https://material.io/guidelines/motion/duration-easing.html#duration-easing-natural-easing-curves

src/components/input/input.scss
```diff
@@ -181,6 +181,7 @@ md-input-container {
     -ms-flex-preferred-size: $input-line-height; //IE fix
     border-radius: 0;
     border-style: solid; // Firefox fix
+    transition: border-color 0.4s cubic-bezier(.25,.8,.25,1);
```
Splaktar, line 184:
```suggestion
    transition: border-color $swift-ease-out-duration $swift-ease-out-timing-function;
```
We've got some variables defined already. They can be used to make this a bit easier.

Splaktar:
This build is failing because we switched from TravisCI to CircleCI. Rebasing the PR would resolve that, but it doesn't need to be rebased to merge this one line change.

Splaktar:
@googlebot I consent.

nnmrts:
Wow.

PR #10083, opened 2016-12-02
Remove mdMaxlength directive forcing ng-trim

christrude:
Adding ng-trim="false" to an input will cause the 'required' state to be fulfilled by empty spaces, essentially allowing a user to put in no content and pass required validation.  Allowing angular to trim, and trimming manually in the renderCharCount function works better (IMO).

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -649,8 +649,9 @@ function mdMaxlengthDirective($animate, $mdUtil) {
 
       // Stop model from trimming. This makes it so whitespace
       // over the maxlength still counts as invalid.
-      attr.$set('ngTrim', 'false');
-
+      // Edit: Setting the trim is not needed, as the whitespace over maxlength will be trimmed 
+      // attr.$set('ngTrim', 'false');
+ 
       scope.$watch(attr.mdMaxlength, function(value) {
         maxlength = value;
         if (angular.isNumber(value) && value > 0) {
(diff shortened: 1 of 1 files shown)


christrude:
I signed it!

src/components/input/input.js
```diff
@@ -649,8 +649,9 @@ function mdMaxlengthDirective($animate, $mdUtil) {
 
       // Stop model from trimming. This makes it so whitespace
       // over the maxlength still counts as invalid.
-      attr.$set('ngTrim', 'false');
-
+      // Edit: Setting the trim is not needed, as the whitespace over maxlength will be trimmed 
+      // attr.$set('ngTrim', 'false');
+
```
DevVersion: If this code won't be used anymore. Then this should be removed.

src/components/input/input.js
```diff
@@ -684,7 +685,7 @@ function mdMaxlengthDirective($animate, $mdUtil) {
 
       // Force the value into a string since it may be a number,
       // which does not have a length property.
-      charCountEl.text(String(element.val() || value || '').length + ' / ' + maxlength);
+      charCountEl.text(String(element.val().trim || value || '').length + ' / ' + maxlength);
```
DevVersion: Instead of always trimming the text we should make this more dependent to the model

* Similar to the `ng-maxlength` directive: [See source](https://github.com/angular/angular.js/blob/master/src/ng/directive/validators.js#L266)

This would mean that the counter respects the `ng-trim` attribute.

Splaktar:
Even after rebasing, this branch still causes 8 test failures in addition to the review feedback above not being addressed in a couple of years. I'm going to close out this PR and create a new one based on it so that we can get things fixed up and resolve the issue.

src/components/input/input.js
Splaktar: I looked at the `ng-maxlength` directive linked here, but I don't really see it interacting with the `ng-trim` attribute in any way.

Issue #2455, opened 2015-04-21
md-fab-top-center and md-fab-bottom-center


soutlink:
Shoud be a .md -fab -top -center and .md -fab -bottom -center class ?

I mean , here (http://www.google.com/design/spec/components/buttons-floating-action-button.html#buttons-floating-action-button-behavior) is a fab button centered at bottom ( like in lollipop dialer )

This behavior its not possible at this moment (I think)

mprentic-harris:
Yes, this would be nice.

danieladz:
yeah, I would like this too

barraq:
:+1:

topherfangio:
The team is focused on getting things ready for v1.0, so this is a little bit low on the priority list.

If anyone is interested in submitting a PR to help us get this out sooner, it would certainly be welcomed :-)

ThomasBurleson:
This issue is closed as part of our deprecation effort.
    For details, see our post [Spring Cleaning 2016](https://groups.google.com/forum/#!topic/ngMaterial/3Ip9nd68bfk).

PR #10097, opened 2016-12-07
Add fab positions bottom-center and top-center

JTWalraven:
This is a simple change to add fab positions for bottom-center and top-center as mentioned in issue  #2455. Center fab buttons are a part of Material as described here: https://material.google.com/components/buttons-floating-action-button.html#buttons-floating-action-button-behavior

### DIFF
--- a/src/core/style/mixins.scss
+++ b/src/core/style/mixins.scss
@@ -151,14 +151,18 @@
     right: $right;
     bottom: $bottom;
     left: $left;
+    margin-left: auto;
+    margin-right: auto;
     position: absolute;
   }
 }
 
 @mixin fab-all-positions() {
   @include fab-position(bottom-right, auto, ($button-fab-width - $button-fab-padding)/2, ($button-fab-height - $button-fab-padding)/2, auto);
+  @include fab-position(bottom-center, auto, 0, ($button-fab-height - $button-fab-padding)/2, 0);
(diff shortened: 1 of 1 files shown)


JTWalraven:
I agree!

ThomasBurleson:
@EladBezalel - please review and test.

EladBezalel:
@JTWalraven thanks for the contrib. and sorry for the late response, lgtm.

@ThomasBurleson we i happen to notice 2 things: 
1. we don't document those classes anywhere.
2. we should add support for `start` and `end` for rtl

JTWalraven:
@EladBezalel I am glad I could help out. Thank you for reviewing and what you all have done with this project!

ThomasBurleson:
@EladBezalel - what do you mean start/end ?

EladBezalel:
`md-fab-bottom-start` would be left on ltr and right on rtl, 
`md-fab-bottom-end` would be right on ltr and left on rtl
just like we have on panel

jelbourn:
@JTWalraven Unfortunately modifying the styles in this mixin is a breaking change; it causes visual differences in a large number of Google apps we tested. Because of this we won't be able to merge this change.

JTWalraven:
@jelbourn Is there any additional information you could provide that might help me fix this? It is likely caused by the margin-left and margin-right changes.

PR #10103, opened 2016-12-07
Revert "fix(progressCircular): update animation to spec"

mmalerba:
Reverts angular/material#10017. This breaks Google Analytics


clshortfuse:
See fix in #10108

PR #10108, opened 2016-12-07
fix(progressCircular): fix arc bleeding through container

### LINKED ISSUE #10107: progressCircular: svg arc is bleeding through container
**Actual Behavior**:
 -  `What is the issue? *` svg arg clips through container
 -  `What is the expected behavior?` not to bleed

 -  `CodePen Demo which shows your issue:` https://material.angularjs.org/HEAD/demo/progressCircular
 -  `Details:` use inspect to see bleed

**Angular Versions**: *

----

Filing issue for PR

CC @ErinCoughlan

clshortfuse:
SVG arc was not properly sized which caused it to bleed through it's container.
In some configurations, the circle would be cropped

* Account for strokeWidth when measuring dash length
* Set offset in arc path equal to half of stroke width

Fixes #10107

### DIFF
--- a/src/components/progressCircular/js/progressCircularDirective.js
+++ b/src/components/progressCircular/js/progressCircularDirective.js
@@ -199,13 +199,13 @@ function MdProgressCircularDirective($window, $mdProgressCircular, $mdTheming,
       path.attr('stroke-width', strokeWidth);
       path.attr('stroke-linecap', 'square');
       if (scope.mdMode == MODE_INDETERMINATE) {
-        path.attr('d', getSvgArc(diameter, true));
-        path.attr('stroke-dasharray', diameter * $window.Math.PI * 0.75);
-        path.attr('stroke-dashoffset', getDashLength(diameter, 1, 75));
+        path.attr('d', getSvgArc(diameter, strokeWidth, true));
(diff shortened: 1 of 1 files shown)


clshortfuse:
Before:
screenshot 2016-12-07 at 5 06 48 pm

After
screenshot 2016-12-07 at 5 05 46 pm

crisbeto:
Judging by the second screenshot, it seems like the top part is still bleeding out a little. This usually isn't an issue on Chrome, but IE is `overflow: hidden` by default.

clshortfuse:
@crisbeto The change does revert back to the proper offset notation in the original:

```
v1.1.1:  M25,2.5A22.5,22.5 0 1,1 ... (varies by position)
This PR: M25,2.5A22.5,22.5 0 1 1 2.5,25A22.5,22.5 0 0 1 25,2.5
```

I'm not sure how much we can work against subpixel bleeding. I think we can examine that for another PR.

clshortfuse:
@crisbeto I think that bleed is a Chrome Inspect Tool visual bug. As you suspected `overflow:hidden` highlights the problem in the `master` branch. This PR fixes it:

master:
screenshot 2016-12-07 at 5 47 18 pm

pr:
screenshot 2016-12-07 at 5 47 50 pm

PR #10115, opened 2016-12-09
fix(mdTooltip): AngularJS 1.3.20 test failures.

### LINKED ISSUE #10114: fix(mdTooltip): AngularJS 1.3.20 test failures.
**Actual Behavior**:
 -  `What is the issue? *` Several errors are occurring in the tooltip test spec with AngularJS 1.3.20.
 -  `What is the expected behavior?` No errors!

**Angular Versions**: *
 -  `Angular Version:` 1.3.20

bradrich:
The test failures were occurring due to the tooltip using the panel API. All requests to destroy the scope need to also be accompanied by requests to destroy the panel.

Fixes #10114

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -297,6 +297,7 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
       parent.one('$destroy', onElementDestroy);
       scope.$on('$destroy', function() {
         setVisible(false);
+        panelRef && panelRef.destroy();
         element.remove();
         attributeObserver && attributeObserver.disconnect();
       });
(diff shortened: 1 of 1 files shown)


src/components/tooltip/tooltip.js
```diff
@@ -311,6 +312,7 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
       }
 
       function onElementDestroy() {
+        panelRef && panelRef.destroy();
```
ThomasBurleson: Why is this needed when you have L300 ?

bradrich: L300 occurs when the tooltip's scope is destroyed. L315 occurs when the parent element's scope is destroyed.

bradrich: You are right about this @ThomasBurleson. It should only be on L300. I will update this asap.

PR #10124, opened 2016-12-13
update fab toolbar so the background element has -webkit-transform scale for safari

renrizzolo:
fix for the fab toolbar background to show (scale) on webkit (safari osx/ios)


clshortfuse:
`-webkit-transform` is no longer needed as of Safari 9.0 and Chrome 36.0. We do not support older versions.

 We strongly recommend filing an issue first, so we can approach and discuss problem before resorting to code changes.

ThomasBurleson:
Closing as invalid.

PR #10136, opened 2016-12-14
fix(tabs): remove manual pagination sizing algorithm

### LINKED ISSUE #9429: tabs issue with 1.1.0
md-pagination-wrapper width is 1px less from what it should be generated, and this makes last tab to disappear from the bar
untithjkhkhkled-1

when i set 1px more it works ok, this issue happen with all may tabs after updating to 1.1.0 from rc5

clshortfuse:
Since tabs now use flexbox for sizing, manually sizing is no longer needed.
This solves issues related tabs inside flexible containers like dialogs.

* Remove updatePagingWidth() function
* Revert width to default values

Fixes #9429

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -606,21 +606,8 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
    * Updates whether or not pagination should be displayed.
    */
   function updatePagination () {
-    updatePagingWidth();
     ctrl.maxTabWidth = getMaxTabWidth();
     ctrl.shouldPaginate = shouldPaginate();
-  }
-
-  /**
-   * Sets or clears fixed width for md-pagination-wrapper depending on if tabs should stretch.
-   */
-  function updatePagingWidth() {
-    var elements = getElements();
-    if (shouldStretchTabs()) {
-      angular.element(elements.paging).css('width', '');
-    } else {
-      angular.element(elements.paging).css('width', calcPagingWidth() + 'px');
-    }
   }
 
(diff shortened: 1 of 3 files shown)


clshortfuse:
@ErinCoughlan this should solve your issue

Just to clarify, this changes the width of the `paginator`, not the actual tab buttons. 

Tested with http://codepen.io/shortfuse/pen/LbXVWP

- [ ] IE11
- [ ] Edge
- [ ] Safari
- [ ] Safari Mobile
- [ ] Chrome
- [ ] Firefox

ThomasBurleson:
@clshortfuse - nice work.

bvahdat:
@clshortfuse @ErinCoughlan please see #10386

Splaktar:
This appears to have broken the display of tabs when pagination is needed on IE11: https://github.com/angular/material/issues/10406#issuecomment-382279969

src/components/tabs/js/tabsController.js
```diff
@@ -606,23 +606,10 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
 
-  /**
-   * Sets or clears fixed width for md-pagination-wrapper depending on if tabs should stretch.
-   */
-  function updatePagingWidth() {
-    var elements = getElements();
-    if (shouldStretchTabs()) {
-      angular.element(elements.paging).css('width', '');
-    } else {
-      angular.element(elements.paging).css('width', calcPagingWidth() + 'px');
```
Splaktar: This was the only call to `calcPagingWidth()` but that function wasn't removed as is currently dead code.

PR #10143, opened 2016-12-15
feat(tabs): add the ability to align tabs to the end

tuncelmert:
Basically has the same functionality with `md-center-tabs` but aligning
the tabs right

* Define boolean attribute `md-align-right`
* Add .md-align-right class if enabled
* Add css rule for `.md-align-right`

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -66,7 +66,8 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
     defineBooleanAttribute('noDisconnect');
     defineBooleanAttribute('autoselect');
     defineBooleanAttribute('noSelectClick');
-    defineBooleanAttribute('centerTabs', handleCenterTabs, false);
+    defineBooleanAttribute('centerTabs', handleCenterTabs, false);    
+    defineBooleanAttribute('alignRight', handleAlignRight, false);
     defineBooleanAttribute('enableDisconnect');
 
     // Define public properties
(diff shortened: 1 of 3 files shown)


tuncelmert:
I signed the CLA

tuncelmert:
@EladBezalel I've made the necessary changes.

src/components/tabs/tabs.scss
```diff
@@ -167,6 +167,11 @@ md-pagination-wrapper {
     width: auto;
     margin: 0 auto;
   }
+  &.md-align-end {
+    position: relative;
+    width: auto;
+    margin: 0 0 0 auto;
```
EladBezalel: in rtl it would still align to the right no? 
can you please check that?

tuncelmert: Yes @EladBezalel , it would. Also `direction` does not change the default alignment of the tabs, only the order is affected. So there wouldn't be any conflict in `rtl`.

tuncelmert: P.S.: Tabs do not work well with rtl. I'm guessing that won't be a good option to go with.

EladBezalel: We do have a pr for that.. well i guess we will have to take of it later.

ThomasBurleson:
@EladBezalel - status of this PR?

EladBezalel:
I should manual test it, overall lgtm

src/components/tabs/js/tabsController.js
```diff
@@ -547,6 +554,9 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
   function shouldCenterTabs () {
     return ctrl.centerTabs && !ctrl.shouldPaginate;
   }
+    function shouldAlignEnd () {
+    return ctrl.alignEnd && !ctrl.shouldPaginate;
+  }
```
Splaktar: These indents need to be fixed please.

Splaktar:
Moving to the backlog as the request for rebase and squash of commits hasn't been addressed yet.

Splaktar:
I'm going to close this as there does not appear to be significant interest in getting this enhancement merged in.

PR #10147, opened 2016-12-16
Update script.js

Bradbev:
Removed controller: from the pre-rendered dialog config.  Controllers on pre-rendered dialogs are not used, having it here is confusing.


Bradbev:
I signed it!

ThomasBurleson:

PR #10159, opened 2016-12-20
fix(tooltip): properly interpolate tooltip text to prevent possible XSS

jelbourn:
Replaces #9606


ThomasBurleson:
@jelbourn - lgtm.

src/components/tooltip/tooltip.js
```diff
@@ -104,8 +104,9 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
 
     function addAriaLabel(override) {
```
ThomasBurleson: @jelbourn  - Per the PR ancestor (that was closed), we should add comments about *what* is `override` ?

PR #10163, opened 2016-12-21
fix(mdTooltip): Tooltip position and watcher issue.

### LINKED ISSUE #10157: tooltip aria watchers
Pullrequest 7822 https://github.com/angular/material/pull/7822/files was doing such a good work getting rid of many watchers and then there were added the following code that again adds thousand of watchers for my application: 

```
// Updates the aria-label when the element text changes. This watch
// doesn't need to be set up if the element doesn't have any data
// bindings.
if (element.text().indexOf($interpolate.startSymbol()) > -1) {
    scope.$watch(function() {
      return element.text().trim();
    }, addAriaLabel);
}

```

### LINKED ISSUE #10162: fix(mdTooltip): Dynamic position changes not removing position class
**Actual Behavior**:
 -  `What is the issue? *` When the position of a tooltip is changed dynamically, the new tooltip still has the same position class as the original tooltip.
 -  `What is the expected behavior?` The new position class should be shown and the original class removed.

 -  `Details:` N/A

**Angular Versions**: *
 -  `Angular Version:` 1.X.X

**Additional Information**:
 -  `Browser Type: *` All
 -  `OS: *` All
 -  `Stack Traces:` N/A

bradrich:
Fixes #10162 - When the position was changed dynamically the resulting tooltip's panel would retain the position classes on its panelEl. Now when creating and recreating the tooltip with a different position class due to a position change, the panel will grab that tracked panel and update its configuration object so that the new position class can be used.

Fixes #10157 - There was a `mdDirection` watcher that was deleted due to a rebase/timing issue when moving to the `mdPanel` API. Reference to the previous PR: https://github.com/angular/material/pull/7822/files

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -910,9 +910,12 @@ MdPanelService.prototype.create = function(config) {
   config = config || {};
 
   // If the passed-in config contains an ID and the ID is within _trackedPanels,
-  // return the tracked panel.
+  // return the tracked panel after updating its config with the passed in
+  // config.
   if (angular.isDefined(config.id) && this._trackedPanels[config.id]) {
-    return this._trackedPanels[config.id];
+    var trackedPanel = this._trackedPanels[config.id];
+    angular.extend(trackedPanel.config, config);
+    return trackedPanel;
   }
 
   this._config = {
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -267,34 +267,17 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
     }
 
     function configureWatchers() {
-      if (element[0] && 'MutationObserver' in $window) {
-        var attributeObserver = new MutationObserver(function(mutations) {
-          mutations.forEach(function(mutation) {
-            switch (mutation.attributeName) {
-              case "md-visible":
-                scope.visibleWatcher = scope.visibleWatcher || scope.$watch('mdVisible', onVisibleChanged);
-                break;
-              case "md-direction":
-                $mdUtil.nextTick(updatePosition);
-                break;
-            }
-          });
-        });
+      if ( !buildMutationObserver() ) {
+        // Build watcher only if attribute is being used.
 
-        attributeObserver.observe(element[0], {
-          attributes: true
-        });
-
-        // Build watcher only if mdVisible is being used.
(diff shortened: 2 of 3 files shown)


bradrich:
For tests to succeed, it desperately needs #10115.

ThomasBurleson:
@bradrich - I refactored the source to `MdTooltipDirective::configureWatchers()`. Please review and confirm correctness. Thx.

ThomasBurleson:
@bradrich tooltip.js#L325 especially needs confirmation.

bradrich:
I will take a look this evening. Thank you for your attention. 





On Sun, Jan 1, 2017 at 2:54 PM -0500, "Thomas Burleson" <notifications@github.com> wrote:












@bradrich tooltip.js#L325 especially needs confirmation.

ThomasBurleson:
Please merge this [fixed] PR with #10115.

clshortfuse:
@ThomasBurleson I wrote the original watcher code so I'll comment on the changes. For L325, we want to use an Angular Watcher instead. The change here would mean the MutationObserver would only catch string literal changes. The logic is, if the developer decides to use `mdVisible` we attach the a watcher. We use MutationObserver to track its usage. The `visibleWatcher` was made to avoid creating multiple watchers. `mdDirection` only accepts string literals so that's fine.

For reference, this was the original PR to introduce these changes: https://github.com/angular/material/pull/7822/files

bradrich:
@ThomasBurleson In this particular setup, would this MutationObserver for mdVisible be able to handle expressions? It seems that this would not handle those. I understand that MutationObservers are used for performance reasons, but in this case would it be better to abandon it and only use it for mdDirection?

ThomasBurleson:
@clshortfuse, @bradrich 
I think my changes are the right direction, but missing the logic you outlined above.
I will be online tomorrow AM if you want to pair-up on the logic.

ThomasBurleson:
Rejecting this PR. Will resolve with PRs #10115 and a new PR.
Attn: @bradrich

Splaktar:
@ThomasBurleson you mention "a new PR" will resolve the problems in https://github.com/angular/material/issues/10157 which were part of the focus of this PR. Do you know if that PR was ever created and merged? I don't see any links or trace to indicate that occurred.

I'm trying to close the loop on https://github.com/angular/material/issues/10157 and verify that a performance regression issue is indeed resolved or if it needs attention.

PR #10183, opened 2016-12-27
BugFix to $$interimElement

Shrulik:
This affects mdDialog at the very least. I'm guessing this is somewhat of a typo, where someone forgot to add a paramter name to the resolve function. What happens is that another reponse value is captured, which is the promise that $mdDialog.show returns, so the result is we resolve a promise with itself, which Angular detects and and throws, so kudos to them.

Other weird things I noticed, but didn't look into :
1. The hideElement call on 529, doesn't call rejectAll on rejected promise. I would guess it should. 
2. if isCancelled is true, there is a call to rejectAll with the success result of the call of hideElement. Seems a bit weird.

### DIFF
--- a/src/core/services/interimElement/interimElement.js
+++ b/src/core/services/interimElement/interimElement.js
@@ -526,20 +526,19 @@ function InterimElementProvider() {
 
           if ( options.$destroy === true ) {
 
-            return hideElement(options.element, options).then(function(){
-              (isCancelled && rejectAll(response)) || resolveAll(response);
-            });
+            return hideElement(options.element, options).then(hideSuccessful);
 
           } else {
             $q.when(showAction).finally(function() {
-              hideElement(options.element, options).then(function() {
-                isCancelled ? rejectAll(response) : resolveAll(response);
-              }, rejectAll);
+              hideElement(hideSuccessful, rejectAll);
             });
 
             return self.deferred.promise;
           }
 
+          function hideSuccessful(hideResult){
+              (isCancelled && rejectAll(hideResult)) || resolveAll(hideResult);            
+          }
 
           /**
            * The `show()` returns a promise that will be resolved when the interim


ThomasBurleson:
@DevVersion - can you review ? 
@Shrulik  - can you provide 2 or more unit tests for this ?

src/core/services/interimElement/interimElement.js
```diff
@@ -526,20 +526,19 @@ function InterimElementProvider() {
...
-              hideElement(options.element, options).then(function() {
-                isCancelled ? rejectAll(response) : resolveAll(response);
-              }, rejectAll);
+              hideElement(hideSuccessful, rejectAll);
             });
 
             return self.deferred.promise;
           }
 
+          function hideSuccessful(hideResult){
```
DevVersion: Nit: Can you add space between the `hideResult) {`

src/core/services/interimElement/interimElement.js
```diff
@@ -526,20 +526,19 @@ function InterimElementProvider() {
...
+              (isCancelled && rejectAll(hideResult)) || resolveAll(hideResult);
```
DevVersion: Isn't the `hideResult` different from the previous one? Previously it always passed over the provided `response`.

DevVersion:
@Shrulik Thanks for the PR - I'm really worried about this change. Can you file an issue for the problem you are trying to solve?

Shrulik:
@DevVersion - Hey, I apologize for the waste of time, but you shouldn't merge this PR. I wrongly assumed this was a sort of a typo, but from the failing unit tests it seems this was the intended behavior, with the problem this intended behavior is wrong according to Angular $q/Promise spec. See : https://github.com/angular/angular.js/blob/v1.6.1/src/ng/q.js#L406

Which means fixing this would require actually understanding the code. The scenario where I stumbled upon this was simply trying to chain a promise to the result of $mdDialog.show().  
This happened to me when I was writing something small and internal for a client, so it was easy to workaround and I couldn't justify exploring this bug more extensivly. 

I will try to look into this over a free weekend or at least get you a failling unit test.

DevVersion:
@Shrulik No, you didn't waste our time ;)

> So you basically say that [this code](https://github.com/angular/angular.js/blob/v1.6.1/src/ng/q.js#L406) will throw an error in ngMaterial when using Angular 1.6?

Shrulik:
I would like to cut my speculating losses and will only confirm it happens on Angular 1.5.8, but the logic is much older. 
https://github.com/angular/angular.js/commit/1339c11e36d9b9d23b26b98e90b523ab99901cfe

DevVersion:
@Shrulik Interesting. Never noticed such an error while using things like the `$mdDialog` or `$mdBottomsheet`.

Splaktar:
Thank you for this contribution!

Splaktar:
I am going to close this as there is not a CodePen to demonstrate the issue and it does not appear to be of high priority (no associated issue or other users who have reported this). Additionally, mdDialog is in use across many projects and they have not run into this issue.

Will reopen if a CodePen and failing test can be provided.

Issue #3990, opened 2015-08-02
Angular Material breaks form ng-submit via enter/go button on iOS


wlingke:
If you have a form with ng-submit, pressing enter on a iOS device (with both chrome and safari), will not trigger the ng-submit function. This works as expected with angularjs alone but with the addition of angular material, it no longer works.

Working (without ngMaterial)
http://embed.plnkr.co/Cs3QTjaN6GdnfL1vOKNl/preview

Not working (with ngMaterial)
http://embed.plnkr.co/F0qVeaIxNnBlLHgVKQ5Z/preview

julienpa:
I think I have the same issue when trying to submit a form with Angular (1.4.3) and Angular Material (0.10.1) by pressing enter: the ng-submit is not triggered, whereas it works well when there is a simple `md-button` inside the form

mackelito:
+1.. 
Same issue here, using angular 1.4.5, Materials 0.10.1

aogut:
+1
it worked for me on 0.11.2 but broke on 1.0

jamalx31:
+1

tonyawad88:
+1

topherfangio:
I definitely see the issue within the Plunkr; however, when I extracted it into a simple HTML file on my own iPad to reproduce, the issue disappears. I am running iOS 9, so perhaps that is part of the difference?

Note that the Plunkr fails on Chrome on desktop as well as iOS.

Can someone double-check my code below and let me know if there is something specific you are doing to reproduce, or if it's only a particular version of iOS, because it appears to be working to me?

_P.S. I've added a comment for testing against master as well to see if it was fixed, but I can't reproduce with either of them loaded._

```html
<!doctype html>
<html>
  <head>
    <script src="http://ajax.googleapis.com/ajax/libs/angularjs/1.3.15/angular.min.js"></script>
    <script src="http://ajax.googleapis.com/ajax/libs/angularjs/1.3.15/angular-animate.min.js"></script>
    <script src="https://cdnjs.cloudflare.com/ajax/libs/angular-material/0.10.0/angular-material.js"></script>
    <!--
    <script src="https://material.angularjs.org/HEAD/angular-material.min.js"></script>
    //-->
  </head>

  <body ng-app="app">
    <div ng-controller="Ctrl">

      <form name="testForm" ng-submit="submit(test)">
        <input type="text" ng-model="test">
        <input type="submit">
      </form>

    </div>

    <script typ="text/javascript">
[...]

PR #10189, opened 2016-12-28
fix(mdGesture): fix form submit via enter/go button on iOS (#3990)

radotzki:
Fixes [#3990](https://github.com/angular/material/issues/3990) - when a user clicks enter/go on an iOS device, the form's submit function is not triggered.

### DIFF
--- a/src/core/services/gesture/gesture.js
+++ b/src/core/services/gesture/gesture.js
@@ -552,8 +552,10 @@ function attachToDocument( $mdGesture, $$MdGestureHandler ) {
 
   function clickHijacker(ev) {
     var isKeyClick = ev.clientX === 0 && ev.clientY === 0;
+    var isSubmitEvent = ev.target && ev.target.type === 'submit';
     if (!isKeyClick && !ev.$material && !ev.isIonicTap
-      && !isInputEventFromLabelClick(ev)) {
+      && !isInputEventFromLabelClick(ev)
+      && !isSubmitEvent) {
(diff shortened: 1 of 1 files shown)


EladBezalel:
@DevVersion please review

DevVersion:
Looks valid :+1: But definitely would like to see a small test for this.

radotzki:
@DevVersion I tried to implement test for this, however this is relevant only to iOS/Android devices. According to the [source](https://github.com/angular/material/blob/master/src/core/services/gesture/gesture.js#L83) you hijack clicks only if the user agent is iOS or Android. Also, I saw that you are not using a launcher for these browsers so the test will never run.

In addition I notice that you have 2 tests that depended on $mdGesture.$$hijackClicks ([this](https://github.com/angular/material/blob/master/src/core/services/gesture/gesture.spec.js#L171) and [this](https://github.com/angular/material/blob/master/src/core/services/gesture/gesture.spec.js#L195)).

How can I manipulate $mdGesture and test this case?

DevVersion:
Good point. I'm also wondering where `$$hijackClicks` is set.. Looks like those tests never run (regardless of being mobile or not).

radotzki:
@DevVersion @EladBezalel what do you say? How can we continue from here?

DevVersion:
Yeah LGTM. In the future we are willing to drop `$mdGesture` and use HammerJS instead.

PR #10215, opened 2017-01-04
feat(panel): Panel provider

### LINKED ISSUE #10006: feat(panel): Panel configuration provider

### LINKED ISSUE #10162: fix(mdTooltip): Dynamic position changes not removing position class

bradrich:
Fixes #10006 - Allows users to be able to add the $mdPanelProvider to the config block
of their application. They can use this provider to create one or more
configurations for upcoming panel elements.

Fixes #10162 - Alters the configuration of a panel with a user-defined id that has been
saved when that panel is re-created. This will update the configuration properties such
as panelClass, etc.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -7,11 +7,137 @@ angular
     'material.core',
     'material.components.backdrop'
   ])
-  .service('$mdPanel', MdPanelService);
+  .provider('$mdPanel', MdPanelProvider);
 
 
 /*****************************************************************************
  *                            PUBLIC DOCUMENTATION                           *
+ *****************************************************************************/
+
+
+/**
+ * @ngdoc service
+ * @name $mdPanelProvider
+ * @module material.components.panel
+ *
+ * @description
+ * `$mdPanelProvider` allows users to create configuration presets that will be
+ * stored within a cached presets object. When the configuration is needed, the
+ * user can request the preset by passing it as the first parameter in the
+ * `$mdPanel.create` or `$mdPanel.open` methods.
+ *
+ * @usage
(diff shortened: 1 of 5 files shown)


src/components/panel/panel.js
```diff
@@ -789,27 +915,113 @@ angular
...
+  _presets[name] = preset;
+}
+
+
+/**
+ * Gets all of the presets.
+ * @return {!Object}
+ */
+function getAllPresets() {
+  return _presets;
```
ThomasBurleson: This should be a clone of the presets so the API is readonly.

bradrich:
@ThomasBurleson We are GTG!

ThomasBurleson:
lgtm

PR #10218, opened 2017-01-04
fix(list): empty aria-label attributes for list-items with interpolation

DevVersion:
* The `md-list-item` component determines the aria-label from the content if no `aria-label` is specified.
  Right now the `md-list-item` is not able to determine the label from interpolated content and shows a lot of $mdAria warnings **(See the warnings issue [here](https://material.angularjs.org/HEAD/demo/list))**

* Fixes that the `md-list-item` complains twice about a missing `aria-label`, because the label was expected on the list and on the underlaying button element. **(See the issue [here](http://codepen.io/DevVersion/pen/mRyJpP))**

* Fixes a potential issue in the `$mdAriaProvider` where the provider class instance is applied to the service.

@ThomasBurleson This should go into 1.1.2, because otherwise developers would have a lot of warnings in their applications.


src/core/services/aria/aria.js
```diff
@@ -30,18 +30,15 @@ angular
...
+  var config = {
+    /** Whether we should show ARIA warnings in the console if labels are missing on the element */
+    showWarnings: true
+  };
 
   return {
     disableWarnings: disableWarnings,
-    $get: function($$rAF, $log, $window, $interpolate) {
-      return MdAriaService.apply(self, arguments);
+    $get: function($$rAF, $log, $q, $window, $interpolate) {
```
ThomasBurleson: Why did you inject `$q` when I do not see it being used [in these new changes].

src/core/services/aria/aria.js
```diff
@@ -30,18 +30,15 @@ angular
...
 function MdAriaProvider() {
 
-  var self = this;
-
-  /**
-   * Whether we should show ARIA warnings in the console, if labels are missing on the element
-   * By default the warnings are enabled
-   */
-  self.showWarnings = true;
+  var config = {
```
ThomasBurleson: Do we have an specs for these Provider changes ?

DevVersion: The changes won't be noticeable from outside, so testing this would be unnecessary. Also we still didn't confirm that Angular instantiates the provider with a new instance or by just calling the function.

ThomasBurleson:
LGTM

ThomasBurleson:
@jelbourn  - Sample Console errors **BEFORE** this PR is merged:

screen shot 2017-01-06 at 8 17 08 am

PR #10219, opened 2017-01-04
fix(navBar): automatically add aria-label for navBarItem

### LINKED ISSUE #10110: md-nav-item should be able to input aria-label
md-nav-item needs aria label so we can use md-icon inside the md-nav-item block

for now, there's no way to put aria-label in the anchor block generated from md-nav-item

clshortfuse:
NavBarItems should populate aria-label attribute automatically

* Inject $mdAria into navBarItem controller
* Use text content from element to populate aria-label during postLink
* Create new spec tests

Fixes #10110


ThomasBurleson:
LGTM

PR #10221, opened 2017-01-04
Revert "refactor(list): simplify list component scss"

kara:
Reverts angular/material#9194. This PR intentionally changed the spacing of lists and subheaders. This caused a number of screen diff breakages, some of which were the intended changes and some of which were layout breakages due to updated CSS. Due to these visual changes, we need to revert this for now.


DevVersion:
It was actually intentional that stuff like the `dense` lists are now having different spacing (and are now conform with the specs).

The subheaders were also affected because those had a height depending on the text, instead of the exact `48px`

PR #10224, opened 2017-01-05
fix(chips): failing unit tests against Angular 1.3

crisbeto:
Fixes a couple of chips tests which were failing against Angular 1.3.x due to a timeout that wasn't being flushed.


ThomasBurleson:
lgtm

PR #10229, opened 2017-01-06
chore: run tests against angular snapshot

DevVersion:
* Now runs unit tests against Angular snapshot
* Exported the version download script from the `test-versions.sh` script from Jenkins (allows us to test against snapshot)


ThomasBurleson:
@DevVersion  - very Nice.

PR #10243, opened 2017-01-10
fix(mdTooltip): Tooltip parent aria label override

### LINKED ISSUE #10242: fix(mdTooltip): Tooltip parent aria label override
**Actual Behavior**:
 -  `What is the issue? *` The tooltip, when using interpolation to fill its text, overrides the parent `aria-label` even when the parent already has a preset `aria-label`.
 -  `What is the expected behavior?` The tooltip should not override an `aria-label` on a parent element that is already set.

 -  `CodePen Demo which shows your issue:` http://codepen.io/erincoughlan/pen/YNXYQe
 -  `Details:` N/A

**Angular Versions**: *
 -  `Angular Version:` 1.5.5

**Additional Information**:
 -  `Browser Type: *` All
 -  `OS: *` All
 -  `Stack Traces:` N/A

Ping: @ErinCoughlan @clshortfuse

bradrich:
Fixes #10242 - The tooltip now does not override or update the parent element `aria-label` unless the parent's `aria-label` does not exist, or it has been previously set by the tooltip. It determines if it has been set by the tooltip by placing the `aria-labelledby` attribute on the parent element and setting it to `tooltip`.

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -75,6 +75,7 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
 
   function linkFunc(scope, element, attr) {
     // Set constants.
+    var tooltipId = 'tooltip-' + $mdUtil.nextUid();
     var parent = $mdUtil.getParentWithPointerEvents(element);
     var debouncedOnResize = $$rAF.throttle(updatePosition);
     var mouseActive = false;
@@ -102,12 +103,24 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
       }
     }
 
-    function addAriaLabel(override) {
-      if (override || !parent.attr('aria-label')) {
-        // Only interpolate the text from the HTML element because otherwise the custom text
-        // could be interpolated twice and cause XSS violations.
-        var interpolatedText = override || $interpolate(element.text().trim())(parent.scope);
+    function addAriaLabel(labelText) {
+      // Only interpolate the text from the HTML element because otherwise the custom text could
+      // be interpolated twice and cause XSS violations.
+      var interpolatedText = labelText || $interpolate(element.text().trim())(parent.scope);
+
+      // Only add the `aria-label` to the parent if there isn't already one, if there isn't an
+      // already present `aria-labelledby`, or if the previous `aria-label` was added by the
+      // tooltip directive.
+      if (
+        (!parent.attr('aria-label') && !parent.attr('aria-labelledby')) ||
+        parent.attr('aria-labelledby') === tooltipId
+      ) {
         parent.attr('aria-label', interpolatedText);
+
+        // Set the `aria-labelledby` attribute if it has not already been set.
+        if (!parent.attr('aria-labelledby')) {
+          parent.attr('aria-labelledby', tooltipId);
+        }
       }
     }
 
@@ -360,7 +373,6 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
       }
 
       if (!panelRef) {
-        var id = 'tooltip-' + $mdUtil.nextUid();
         var attachTo = angular.element(document.body);
         var panelAnimation = $mdPanel.newPanelAnimation()
             .openFrom(parent)
@@ -371,7 +383,7 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
             });
 
         var panelConfig = {
-          id: id,
+          id: tooltipId,
           attachTo: attachTo,
           contentElement: element,
           propagateContainerEvents: true,
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -75,7 +75,7 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
 
   function linkFunc(scope, element, attr) {
     // Set constants.
-    var tooltipId = 'tooltip-' + $mdUtil.nextUid();
+    var tooltipId = 'md-tooltip-' + $mdUtil.nextUid();
     var parent = $mdUtil.getParentWithPointerEvents(element);
     var debouncedOnResize = $$rAF.throttle(updatePosition);
     var mouseActive = false;
--- a/src/components/tooltip/tooltip.spec.js
+++ b/src/components/tooltip/tooltip.spec.js
@@ -82,6 +82,34 @@ describe('MdTooltip Component', function() {
     );
 
     expect(element.attr('aria-label')).toEqual('Tooltip');
+    expect(element.attr('aria-labelledby')).toBeDefined();
+  });
+
+  it('should not label the parent if it already has a label', function() {
+    buildTooltip(
+      '<md-button aria-label="Button Label">' +
+        '<md-tooltip md-visible="testModel.isVisible">' +
+          'Tooltip' +
+        '</md-tooltip>' +
+      '</md-button>'
+    );
+
+    expect(element.attr('aria-label')).toEqual('Button Label');
+    expect(element.attr('aria-labelledby')).toBeUndefined();
+  });
+
+  it('should not label the parent if it has already been labelled by ' +
+      'another element.', function() {
+    buildTooltip(
+      '<md-button aria-labelledby="button-1">' +
+        '<md-tooltip md-visible="testModel.isVisible">' +
+          'Tooltip' +
+        '</md-tooltip>' +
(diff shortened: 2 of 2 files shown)


src/components/tooltip/tooltip.js
```diff
@@ -102,12 +102,20 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
...
+      var interpolatedText = labelText || $interpolate(element.text().trim())(parent.scope);
+
+      // Only add the `aria-label` to the parent if there isn't already one, or if the previous
+      // `aria-label` was added by the tooltip directive.
+      if (!parent.attr('aria-label') || parent.attr('aria-labelledby') === 'tooltip') {
         parent.attr('aria-label', interpolatedText);
+
+        // Set the `aria-labelledby` attribute if it has not already been set.
+        if (!parent.attr('aria-labelledby')) {
+          parent.attr('aria-labelledby', 'tooltip');
```
ErinCoughlan: I thought aria-labelledby took the id of the element which does the labeling. So you should also be adding an id somewhere to the component (unless I'm missing one already added).

"Tooltip" seems too generic to be an id because then only one tooltip can exist on the page, and it's not prefixed with md- so could conflict with elements already on the page. I think using the panel id could work since those are generated each time, but I'd consider the prefix as well, similar to class names and component names.

src/components/tooltip/tooltip.js
```diff
@@ -102,12 +103,20 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
...
-        // could be interpolated twice and cause XSS violations.
-        var interpolatedText = override || $interpolate(element.text().trim())(parent.scope);
+    function addAriaLabel(labelText) {
+      // Only interpolate the text from the HTML element because otherwise the custom text could
+      // be interpolated twice and cause XSS violations.
+      var interpolatedText = labelText || $interpolate(element.text().trim())(parent.scope);
+
+      // Only add the `aria-label` to the parent if there isn't already one, or if the previous
+      // `aria-label` was added by the tooltip directive.
+      if (!parent.attr('aria-label') || parent.attr('aria-labelledby') === tooltipId) {
```
clshortfuse: We should abort the function if `aria-label` or `aria-labelledy` are already in use. 

For example : `<md-button aria-labelledby="myOtherElementId">` 

would have its label changed.

src/components/tooltip/tooltip.js
```diff
@@ -102,12 +103,20 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
...
         parent.attr('aria-label', interpolatedText);
```
clshortfuse: If we are going to use `aria-labelledby`, we don't need to use `aria-label`. Therefore, we don't have to interpolate the text at all.

bradrich:
@clshortfuse This pertains to the usage of `aria-labelledby` vs. `aria-label`. Right now, the mdPanel API either creates an ID for each panel that it builds or uses the ID that the user passes it. If the user passes an ID, then the panel does not recreate that panel if that panel is closed, removed from the DOM, and then reopened. However, the mdPanel API is not adding the ID to the panel element.

I agree with you that we should ditch the `aria-label` and its accompanying interpolation as it saves us a lot of headache, but with the mdPanel API not adding the panel's associated ID to the DOM, `aria-labelledby` is only being used to make sure that we only override `aria-label`s created by the tooltip and not by another element.

I will be creating another issue that outlines the mdPanel API ID issue and we can discuss it further there, but I think that we can move forward with the tooltip fix as it is right now.

Ping @ErinCoughlan

src/components/tooltip/tooltip.spec.js
```diff
@@ -84,6 +84,31 @@ describe('MdTooltip Component', function() {
     expect(element.attr('aria-label')).toEqual('Tooltip');
   });
 
+  it('should not label the parent if it already has a label', function() {
```
clshortfuse: Your tests should also check the value of `aria-labelledby`

src/components/tooltip/tooltip.js
```diff
@@ -102,12 +103,24 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
...
+      // already present `aria-labelledby`, or if the previous `aria-label` was added by the
+      // tooltip directive.
+      if (
+        (!parent.attr('aria-label') && !parent.attr('aria-labelledby')) ||
+        parent.attr('aria-labelledby') === tooltipId
+      ) {
         parent.attr('aria-label', interpolatedText);
+
+        // Set the `aria-labelledby` attribute if it has not already been set.
+        if (!parent.attr('aria-labelledby')) {
```
clshortfuse: There's some redundancy here. We should use a variable to avoid checking the same thing twice. (IE: `hasAriaLabelledBy`)

bradrich: The second test for the `aria-labelledby` attribute (on line 121) is so that I will know if I need to add the attribute. This would not be redundant, as it is possible to get into this `if` statement has been labelled by the tooltip already and is capable of being overridden. You just wouldn't want to attempt to put the `aria-labelledby` attribute on the element again...

src/components/tooltip/tooltip.spec.js
bradrich: I do. Check line 99.

src/components/tooltip/tooltip.spec.js
```diff
@@ -84,6 +84,31 @@ describe('MdTooltip Component', function() {
...
+        '<md-tooltip md-visible="testModel.isVisible">' +
+          'Tooltip' +
+        '</md-tooltip>' +
+      '</md-button>'
+    );
+
+    expect(element.attr('aria-label')).toEqual('Button Label');
+  });
+
+  it('should not label the parent if it has already been labelled by ' +
```
bradrich: @clshortfuse Test for the presence of the `aria-labelledby` attribute on the parent element.

bradrich:
@clshortfuse Please re-review after comments.

src/components/tooltip/tooltip.spec.js
clshortfuse: There is no `expect(element.attr('aria-labelledby'))`

src/components/tooltip/tooltip.spec.js
```diff
@@ -92,47 +120,86 @@ describe('MdTooltip Component', function() {
...
 
-      $rootScope.$apply(function() {
-        $rootScope.testModel.ariaTest = 'test {{1+1}}';
-      });
+    $rootScope.$apply(function() {
+      $rootScope.testModel.ariaTest = 'test {{1+1}}';
+    });
 
-      expect(element.attr('aria-label')).toBe('test {{1+1}}');
+    expect(element.attr('aria-label')).toBe('test {{1+1}}');
```
clshortfuse: If `aria-labelledby` is being used, `aria-label` should not be used. 

See: https://www.w3.org/TR/wai-aria/states_and_properties#aria-labelledby

src/components/tooltip/tooltip.js
clshortfuse: You are running the same comparison in L115 and L121: `!parent.attr('aria-labelledby')`

It's redundant and a stored variable can used instead.

src/components/tooltip/tooltip.spec.js
bradrich: @clshortfuse I agree with you. However, the panel API is not technically putting the tracked ID on the panel DOM element. This would make the `aria-labelledby` attribute not function properly. @ErinCoughlan and I are deliberating on this issue and will be creating a new issue for it soon.

src/components/tooltip/tooltip.js
bradrich: @clshortfuse I don't think that a stored variable will reduce the bytes used nor increase readability significantly. I think that we are good to go on this fix. Can you approve all changes please?

src/components/tooltip/tooltip.js
```diff
@@ -75,6 +75,7 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
 
   function linkFunc(scope, element, attr) {
     // Set constants.
+    var tooltipId = 'tooltip-' + $mdUtil.nextUid();
```
DevVersion: How about prefixing the id with `md-`? (As same as in the `datepicker` [here](https://github.com/angular/material/blob/342ee53363e06a46958541eac980e5ff3d62029d/src/components/datepicker/js/datepickerDirective.js#L331))

ThomasBurleson:
Closed with new PR #10464

Issue #10240, opened 2017-01-09
input: ng-messages do not work if animation is disabled


nikclayton:
**Actual Behavior**:

With a form like:

```html
<form name="f">
  <md-input-container class="md-block">
    <label>Some text</label>
    <input type="text" name="i" placeholder="Enter something" required />
    <div ng-messages="f.i.$error">
      <div ng-message="required">This field is required.</div>
    </div>
  </md-input-container>
</form>
```

Everything works -- tab in to the field, tab out, and the field behaves as expected. Specifically:

1. The placeholder text "Enter something" appears in red.
2. The form field is underlined in red.
3. The text "This field is required." appears underneath the field.

However, if your app turns off animation by, for example, injecting a $animate instance and calling $animate.enabled(false), (1) and (2) still occur, but (3) no longer happens -- the ng-message does not appear.

This is a regression from 1.0, where it did work.

**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:` http://codepen.io/anon/pen/dNYKwr
 -  `Details:` That codepen is identical to the one at https://material.angularjs.org/latest/demo/input (the "Errors Advanced" section). All I've done is change the controller to inject `$animate`, set `$scope.showHints = false`, and then called `$animate.enabled(false);`.

If you change that to `$animate.enabled(true)` the errors will start appearing again.

**Angular Versions**: *
 -  `Angular Version:` 1.5.x
 -  `Angular Material Version:` 1.1.x

**Additional Information**:
 -  `Browser Type: *` Chrome.
 -  `Browser Version: *` 55.0.2883.87
 -  `OS: *` Ubuntu 14.04
 -  `Stack Traces:` N/A

nikclayton:
Suspect that https://github.com/angular/material/commit/cf7f21aa6c633d9f0bd22013e9a79f99b82b16af is the culprit, which unconditionally uses $animate() to add new classes instead of checking to see if animation is enabled.

clshortfuse:
@nikclayton Correct. I'm the process of moving the animations to pure CSS. It just needs more testing but I'll probably have it ready for today. 

The reason why we moved to $animate was due to a flicker bug. We can now move back to CSS because of the newer ng-enter-prepare class that was added since.

topherfangio:
I've removed the `has: Pull Request` label because the PR didn't account for many instances. I think we should just check if animation is enabled and if not, call `element.addClass()`. I will look into this soon.

[...]

Issue #9723, opened 2016-09-29
input-container: not showing ng-messages error by default even when md-auto-hide="false" is applied


vinhnghi223:
**Actual Behavior**:
-  `What is the issue? *`: Under `md-input-container`, when applied `md-auto-hide="false"` to`ng-messages`, error message(s) is not shown by default.
-  `What is the expected behavior?`: If the input is invalid, error message(s) should be shown by default when `md-auto-hide="false"` is applied on `ng-messages`.

**CodePen** (or steps to reproduce the issue): *
-  `CodePen Demo which shows your issue:` http://codepen.io/vinhnghi223/pen/JRyoXJ ( I used the same example from Angular Material Input Errors demo: https://material.angularjs.org/latest/demo/input) (The `Hourly Rate (USD)` example)
-  `Details:` I realized that the reason why the example on Angular Material Input Errors demo works is each tag, where `ng-message` resides, has an inline styles of  `style="opacity: 1; margin-top: 0px;"`.

**Angular Versions**: *
-  `Angular Version:` 1.5.5
-  `Angular Material Version:` 1.1.1

**Additional Information**:
-  `Browser Type: *` Any
-  `Browser Version: *` Any
-  `OS: *` Any
-  `Stack Traces:` Any

vinhnghi223:
Currently my quick and dirty quick in our project is:

Replace these lines of code:
https://github.com/angular/material/blob/master/src/components/input/input.scss#L262-L268

with these lines of code:

```
&.md-input-invalid,
  &.md-input-has-value,
  &.md-input-focused,
  &.nd-input-focused {
    .md-input-message-animation {
      &:not(.ng-animate) {
        opacity: 0;
        margin-top: -100px;
      }
    }
  }
```

That way the error message will then be visible from the start. So it mitigates the issue. Still the label is colored into red. I guess to make that possible some logic inside `inputTextareaDirective `should be changed.

mikila85:
+1

bruno-serfe:
Hi there, Is this still an issue?

Splaktar:
Fixed in `1.1.5` via https://github.com/angular/material/commit/0151b4bed9d90aa942734591615878f5fcc0fe52.

Issue #9543, opened 2016-09-07
input-container: ng-messages are no longer hidden on form reset as of 1.1.1


tom94zoe:
**Actual Behavior**:
-  I tried to hide ng-message with the following code, since Angular Material 1.1.1 the message wasn't hidden on reset. 

``` html
<md-input-container flex style="margin-top: 0;margin-bottom: 5px">
        <label translate>maintenanceMessage.description</label>
        <input md-maxlength="40" required name="description" md-no-asterisk
               ng-model="maintenanceMessageCtrl.newMaintenanceMessage.description" flex>
        <div ng-messages="maintenanceMessageForm.description.$error" ng-show="maintenanceMessageForm.description.$touched">
          <div ng-message="required" translate>maintenanceMessage.description.requiredMessage</div>
          <div ng-message="md-maxlength" translate>maintenanceMessage.description.maxLengthMessage</div>
        </div>
</md-input-container>
```

``` js
ctrl.resetFormAndHideDialog = function () {
      ctrl.newMaintenanceMessage = {};
      $scope.maintenanceMessageForm.$setUntouched();
      $scope.maintenanceMessageForm.$setPristine();
     };
```

**Angular Versions**: *
-  `Angular Version:1.5.8
-  `Angular Material Version:`1.1.1

![angular-material-ng-message](https://cloud.githubusercontent.com/assets/17591511/18313018/cd280370-750c-11e6-964d-939f1714fdb6.png)

mjanthrax:
yea me too, i'm experiencing that kind of issue. so i decided to go back to Angular Material Version 1.1.0rc5

mjanthrax:
hope someone can help us here. cause i really need the new browser color feature of new update in angular material

tom94zoe:
That would be fine, but I think this bug will be fixed with 1.1.3. So we have to wait some weeks.

ThomasBurleson:
@tom94zoe - can you provide a CodePen for use to verify fixes?

rhinoceraptor:
@ThomasBurleson Here is a codepen I made showing the bug:

[https://codepen.io/rhinocerapt0r/pen/ORBQJZ?editors=1011](https://codepen.io/rhinocerapt0r/pen/ORBQJZ?editors=1011)

clshortfuse:
[...]

PR #10246, opened 2017-01-10
fix(input): correct initial animation state of messages

### LINKED ISSUE #6767: md-input-container not showing error ng-messages

clshortfuse:
With commit cf7f21a, animations were moved to input.js in response to an ng-enter
flicker issue with Angular 1.4.x. While the flicker was fixed, new animation bugs
arised. Angular 1.4.x added and backported a ng-enter-prepare to avoid this bug.

* Use `.ng-enter-prepare` to avoid flicker on Angular 1.4+
* Use `.ng-enter:not(.ng-enter-active)` to prepare animations on 1.3 and below
* Update spec tests to use computedStyle

Fixes #6767, #9543, #9723, #10240

Related: angular/angular.js#13408

### DIFF
--- a/src/components/input/input-animations.spec.js
+++ b/src/components/input/input-animations.spec.js
@@ -1,27 +1,32 @@
 describe('md-input-container animations', function() {
-  var $rootScope, $compile, $animateCss, $material, $$mdInput,
-    el, pageScope, invalidAnimation, messagesAnimation, messageAnimation,
-    cssTransitionsDisabled = false, lastAnimateCall;
+  var $rootScope, $compile, $material, $$mdInput, $window, $animate, $rootElement, $document, $timeout,
+    el, root, body, pageScope, computedStyle;
 
   // Load our modules
   beforeEach(module('ngAnimate', 'ngMessages', 'material.components.input', 'material.components.checkbox'));
 
   // Run pre-test setup
-  beforeEach(decorateAnimateCss);
   beforeEach(injectGlobals);
   beforeEach(setupVariables);
 
   // Run after-test teardown
   afterEach(teardown);
 
-  it('set the proper styles when showing messages on an input', function() {
+  it('set the proper styles when showing messages on an input', performInputAnimationTests);
+  it('set the proper styles when showing messages on an input with animations disabled', function() {
+    $animate.enabled(false);
+    performInputAnimationTests();
+    $animate.enabled(true);
+  });
+
+  function performInputAnimationTests() {
     compile(
       '<form name="testForm">' +
       '  <md-input-container>' +
       '    <input name="foo" ng-model="foo" required ng-pattern="/^1234$/" />' +
       '    <div class="errors" ng-messages="testForm.foo.$error">' +
-      '      <div ng-message="required">required</div>' +
-      '      <div ng-message="pattern">pattern</div>' +
+      '      <div ng-message="required" style="transition: 0s none">required</div>' +
+      '      <div ng-message="pattern" style="transition: 0s none">pattern</div>' +
       '    </div>' +
       '  </md-input-container>' +
       '</form>'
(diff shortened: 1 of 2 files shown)


src/components/input/input-animations.spec.js
```diff
@@ -20,169 +18,69 @@ describe('md-input-container animations', function() {
       '  <md-input-container>' +
       '    <input name="foo" ng-model="foo" required ng-pattern="/^1234$/" />' +
       '    <div class="errors" ng-messages="testForm.foo.$error">' +
-      '      <div ng-message="required">required</div>' +
-      '      <div ng-message="pattern">pattern</div>' +
+      '      <div ng-message="required" style="transition:0s none">required</div>' +
+      '      <div ng-message="pattern" style="transition:0s none">pattern</div>' +
```
topherfangio: Why are these in the tests? Are we going to need this in live code?

src/components/input/input.scss
```diff
@@ -210,22 +210,46 @@ md-input-container {
...
 
-    &.ng-enter {
-      // Upon entering the DOM, messages should be hidden
-      .md-input-message-animation {
-        opacity: 0;
-        margin-top: -100px;
-      }
+  .md-input-message-animation {
+    // Enter animation
+    // Pre-animation state is transparent and off target
```
topherfangio: What exactly do you mean by "off target" here? Can you elaborate or use a different terminology?

src/components/input/input.scss
```diff
@@ -254,26 +278,11 @@ md-input-container {
     .md-auto-hide {
       .md-input-message-animation {
         opacity: 0;
-        margin-top: -100px;
+        margin-top: -1em;
```
topherfangio: What if the message is larger than `1em`? Won't this just "jump" in part-way through?

src/components/input/input-animations.spec.js
```diff
@@ -1,70 +1,73 @@
...
-    messageAnimation.enter(getError(), doneSpy);
     flush();
-
-    expectError(getError(), 'pattern');
-    expect(doneSpy).toHaveBeenCalled();
+    errors = getError();
+    expectError(errors, 'pattern');
     expect(container).not.toHaveClass('md-input-invalid');
-    expect(lastAnimateCall).toEqual({element: getError(), options: {}});
+    computedStyle = $window.getComputedStyle(errors[0]);
```
topherfangio: I feel like these should be inside an integration test, not inside the Karma tests. Is there a reason you removed the `lastAnimateCall` way of testing?

src/components/input/input-animations.spec.js
```diff
@@ -1,70 +1,73 @@
...
+  function performInputAnimationTests() {
     compile(
       '<form name="testForm">' +
       '  <md-input-container>' +
       '    <input name="foo" ng-model="foo" required ng-pattern="/^1234$/" />' +
       '    <div class="errors" ng-messages="testForm.foo.$error">' +
-      '      <div ng-message="required">required</div>' +
-      '      <div ng-message="pattern">pattern</div>' +
+      '      <div ng-message="required" style="transition: 0s none">required</div>' +
+      '      <div ng-message="pattern" style="transition: 0s none">pattern</div>' +
```
topherfangio: I'm also still confused by these additions to the test. Why are the necessary in the test but not in user-facing code? If they are required in user-facing code, shouldn't we fix that by adding them?

src/components/input/input-animations.spec.js
clshortfuse: The previous tests weren't actually testing the component's animation procedure. 

For example:

`messageAnimation.leave(patternError, doneSpy);` was called and then `  expect(lastAnimateCall.options.event).toEqual('leave');`. It's a assertion, but didn't actually test whether the component was doing it automatically.

Actually testing the element `opacity` and `margin values` at each state of interaction ensures the object appears as it should.

src/components/input/input-animations.spec.js
clshortfuse: This is the same method of testing as with the original code which had this block:

```
// Make sure any transitions happen immediately; NOTE: this is REQUIRED for the above
// tests to pass without using window.setTimeout to wait for the animations
if (cssTransitionsDisabled) {
  element.css('transition', '0s none');
}
```

I just moved it inline since I dropped animation call tracing in favor or computed style checking

topherfangio: 👍

src/components/input/input-animations.spec.js
topherfangio: As long as the tests pass I guess I'm ok with it.

clshortfuse: I think I should add more tests like md-auto-hide, ngIf, and ngShow attributes to better test each case.

mikila85:
PULL??

PR #10263, opened 2017-01-16
feat(autocomplete): add inline autocomplete options

### LINKED ISSUE #9584: autocomplete: keyboard navigation does not update input box
**Actual Behavior**:
-  `What is the issue? *` When navigating through the dropdown items, the input box is not updated
-  `What is the expected behavior?` Input box should be updated in some way while navigation

-  `CodePen Demo which shows your issue:` https://material.angularjs.org/latest/demo/autocomplete
-  `Details:` Compare with Chrome, Firefox, Safari, etc browsers with their autocomplete implemenations in input boxes and address bar

**Angular Versions**: *
-  `Angular Version:` 1.5.5

---

I'm filing a proper issue for this but, a PR has already been built for this feature in #8904 

In my opinion, there are two methods, "replace text" and "append text".

Autocomplete UX Research: https://docs.google.com/document/d/1t60A2_Qni3IOSZDRqwwum3r5EVPjANJAkvFuMoqGfQA/

clshortfuse:
Creates an option for adding inline options for input text by tracking keyboard
events and dropdown navigation

* Add `md-inline-option` option with possible values of `append` or `replace`
* Update docs to show inline options
* Create a new model called `inputText` representing the input box text
* Do not change `searchText` unless the user changes the input box or accepts
  a dropdown value

Note: This creates added complexity to autocomplete, since the user changes
the `inputText` model, and then the `searchText` is updated. Essentially,
this will decouple the text that exists inside the input box from the text
from which queries are run against.

Autocomplete UX Research: https://docs.google.com/document/d/1t60A2_Qni3IOSZDRqwwum3r5EVPjANJAkvFuMoqGfQA/

Fixes #9584

### DIFF
--- a/src/components/autocomplete/autocomplete.spec.js
+++ b/src/components/autocomplete/autocomplete.spec.js
@@ -2771,4 +2771,130 @@ describe('<md-autocomplete>', function() {
     }));
   });
 
+  describe('inline options', function() {
+    it('should replace input text on item select', inject(function($timeout, $mdConstant, $material) {
+      var scope = createScope();
+      var template = '\
+          <md-autocomplete\
+              md-inline-option="replace"\
+              md-selected-item="selectedItem"\
+              md-search-text="searchText"\
+              md-items="item in match(searchText)"\
+              md-item-text="item.display"\
+              placeholder="placeholder">\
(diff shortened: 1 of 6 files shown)


ThomasBurleson:
Rejected since Issue #9584 was closed.

PR #10273, opened 2017-01-18
fix(select): don't override initial model value

crisbeto:
Fixes an issue, introduced in #9945, that caused the select to override the initial model value.

CC @jelbourn


crisbeto:
Looks like this fails against Angular 1.5, but works against 1.6. Will investigate.

crisbeto:
Fixed. It was an issue with the unit test.

PR #10284, opened 2017-01-19
fix(tooltip): always resolve expressions against the correct scope

jelbourn:
Also stop putting things directly on `$rootScope` in the tooltip tests.


ThomasBurleson:
@EladBezalel, @topherfangio - plz review.

Issue #5492, opened 2015-11-02
Fill-layout is having no height on Safari & Android webview


Wardormeur:
On Android default browser and Safari as it's also Webkit, layout-fill is using a `height:100%` which is resulting in a layout not having any height. When using  `height:auto ` it seems to work fine.
http://codepen.io/anon/pen/NGBxeE
In this example, the dialog 's line are overlapping each others because of one of them having layout-fill

mmiszy:
Angular Material 1.0.0 has been released. What does it mean for this issue and for 1.0-rc8 milestone?
I'd say it's quite critical because the framework is basically unusable on iOS now.

jimdubbs:
Any ETA or workaround for this? Its causing me alot of problems.

workfel:
Waiting the issue solved override the class. And its good
```css
  .layout-fill {
    margin: 0 !important;
    width: 100% !important;
    min-height: 100% !important;
    height: auto;
    }`

gadzhimari:
Still not working. I tried to set min-height: auto, min-height: 100%, but won't work. Here the markup https://codeshare.io/1Z3UZ

ThomasBurleson:
This issue is closed as part of our ‘Surge Focus on Material 2' efforts.
      For details, see our forum posting @ http://bit.ly/1UhZyWs.

kkulikowski:
I think I found solution for it. At least in all my cases it was working.

eknowles:
@kkulikowski care to share?

kkulikowski:
Aloha im on vacation now, when ill be  back ill give a sign. I made pull
request

eknowles:
It's okay, I found it attached to the issue, couldn't get your fix to work - will try tomorrow!

rbournissent:
I had a similar issue and fixed it adding `flex` to the element using `layout-fill`

Splaktar:
No longer reproducible in Safari `Version 13.1 (15609.1.20.111.8)`.

PR #10289, opened 2017-01-20
layout: add flex-grow to .layout-fill class due to issues with safari

kkulikowski:
If many cases when you use md-content and layout-fill Safari doesn't show element 100% height. Adding flex-grow: 1 fixes the issue.
Fixes f.e. #5492

### DIFF
--- a/src/core/services/layout/layout.scss
+++ b/src/core/services/layout/layout.scss
@@ -26,6 +26,7 @@
     width: 100%;
     min-height: 100%;
     height: 100%;
+    flex-grow: 1;
   }
 }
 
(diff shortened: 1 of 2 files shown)


kkulikowski:
I signed it!

Splaktar:
Thank you for your PR and your first contribution to AngularJS Material! I'm sorry that it took so long to take another look at your PR.

The linked issue doesn't appear to have any problems in Safari `Version 13` and most Android browsers are now Chrome or another modern browser.

If you can provide a reproduction and OS/device/browser specs to reproduce it and show that this resolves that issue, then I would be happy to re-open this and investigate. But for now, I'm going to close this as not reproducible.

Issue #10187, opened 2016-12-28
md-slider is not working in chrome 55 version in Android Phones


manakbhard:
**Actual Behavior**:
Drag of slider is not working properly only click is working

**CodePen** (or steps to reproduce the issue): *

Open the demo link of angular material in Chrome 55 on Android Phone
https://material.angularjs.org/latest/demo/slider

**Angular Versions**: *
 -  `Angular Version:` 1.5.8
 -  `Angular Material Version:`0.11.4

Note :- these are versions of my current app it is not working for even 1.1.1 version of angular material

**Additional Information**:
 -  `Browser Type: *`Chrome
 -  `Browser Version: *` 55
 -  `OS: *`Android
 -  `Stack Traces:`

----
Shortcut to create a [new CodePen Demo](http://codepen.io/team/AngularMaterial/pen/bEGJdd).
Note: `*` indicates required information. Without this information, your issue may be auto-closed.

> Do not modify the titles or questions. Simply add your responses to the ends of the questions.
  Add more lines if needed.

praveengupta29:
+1

bbialas:
+1

clshortfuse:
Might be because fastClick. Will investigate

clshortfuse:
Not working properly on ChromeOS either. Might be all touch systems.

bbialas:
Any new ideas?

mohammadrafigh:
it's related to #10145

mustaqahmed:
As I commented on the md-slider issue above, this bug is an existing crack in handling PointerEvents in Angular. That's why Edge on a Windows touch device shows the same bug.

Adding a <code>touch-action: none</code> style to the slider element fixes the problem.

@clshortfuse: perhaps adding a <code>touch-action: none</code> for .md-slider-wrapper is enough.

salsaverde123:
You're fucking awesome, You saved my live. Thank You. "No all super heroes wear capes "

Issue #10145, opened 2016-12-15
md-swipe not working correctly on chrome for android


wolfet:
**Actual Behavior**:
 -  `What is the issue? *` 
On the latest version of Chrome for Android, md-swipe is not working as expected. md-swipe-left is triggered from a swipe in any direction. md-swipe-right, md-swipe-up, and md-swipe-down do not respond.
 -  `What is the expected behavior?` 
md-swipe-left should respond only to left swipes, and other md-swipe-* directives should respond to their respective directions.
**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:` http://codepen.io/twolfe/pen/pNQBrb
 -  `Details:` This works on chrome desktop, but not chrome for Android version 55.0.2883.91. Open the codepen, http://codepen.io/twolfe/full/pNQBrb/, on an Android phone with the latest Chrome update, or with Chrome DevTools device mode, to see the issue. This issue is showing up on my Pixel XL.

**Angular Versions**: *
 -  `Angular Version:` 1.5.5
 -  `Angular Material Version:` 1.1.1

**Additional Information**:
 -  `Browser Type: *` Chrome for Android
 -  `Browser Version: *`55.0.2883.91
 -  `OS: *`Android 7.1.1
 -  `Stack Traces:`

RoniPa:
This can also be repeated in the Chrome DevTools Device Mode

LohithAnand:
I encountered the same issue on my hybrid Cordova app. Downgrading the Android System Web view from the play store is a workaround for hybrid apps which use web views for now. 
I hope the issue will be resolved soon.

mdtjn:
Hi Team, its driving my users crazy. Need some quick fix

imromec:
@LohithAnand, how to owngrade the Android System Web view from the play store?

LohithAnand:
@imromec Locate the Android System Webview in play store and click on the uninstall button. This will rollback to the version of Webview shipped with the OS.

SlavaSlimmy:
Same problem on main site demo in the Chrome DevTools Device Mode: https://material.angularjs.org/latest/demo/swipe

mohammadrafigh:
+1

motabass:
+1
Any workarounds for now?

mohammadrafigh:
let me mention that issue also appears on Firefox when testing in device mode with touch enabled. it's kind of a normal to high priority issue IMO but there is no attention for it yet...

[...]

PR #10314, opened 2017-01-27
fix(gestures): slider and swipe touch

### LINKED ISSUE #10294: md-slider not working in chrome on windows touch devices
**Actual Behavior**

Drag of slider is not working properly only click is working. The slider can be dragged for a short time but is lost after a few pixels. When the slider is configured as discrete the slider always jumps to the beginning when dragged. Might be related to #10187

**CodePen (or steps to reproduce the issue)**

Open the demo link of angular material in Chrome 55 on a windows touch device (e.g. Surface, WinPad etc.)
https://material.angularjs.org/latest/demo/slider

**Angular Versions**

Angular Version: 1.5.8
Angular Material Version: 1.1.1

**Additional Information**

Browser Type: Chrome
Browser Version: 55.0.2883.87 m
OS: Windows 8 & 10 (on touch device)

EladBezalel:
- slider was based on mdGesture service that set `touch-action: pan-x` in case the gesture was horizontal;
according the [google pointer-events documentation](https://developers.google.com/web/updates/2016/10/pointer-events):
```
pan-x: The browser is only allowed to perform the horizontal scroll default action.
```
that means we should had enabled `pan-y` and vice-versa instead.

- swipe directive didn't had any `touch-action` css styles, added `touch-action: none` on the directive element on link

fixes #10294, #10187, #10145

### DIFF
--- a/src/core/services/gesture/gesture.js
+++ b/src/core/services/gesture/gesture.js
@@ -248,7 +248,7 @@ function MdGesture($$MdGestureHandler, $$rAF, $timeout) {
         if (touchActionProperty) {
           // We check for horizontal to be false, because otherwise we would overwrite the default opts.
           this.oldTouchAction = element[0].style[touchActionProperty];
-          element[0].style[touchActionProperty] = options.horizontal === false ? 'pan-y' : 'pan-x';
+          element[0].style[touchActionProperty] = options.horizontal ? 'pan-y' : 'pan-x';
         }
       },
       onCleanup: function(element) {
--- a/src/components/swipe/swipe.js
+++ b/src/components/swipe/swipe.js
@@ -84,6 +84,8 @@ function getDirective(name) {
   function DirectiveFactory($parse) {
       return { restrict: 'A', link: postLink };
       function postLink(scope, element, attr) {
+        element.css('touch-action', 'none');
+
         var fn = $parse(attr[directiveName]);
         element.on(eventName, function(ev) {
           scope.$applyAsync(function() { fn(scope, { $event: ev }); });


src/core/services/gesture/gesture.js
```diff
@@ -248,7 +248,7 @@ function MdGesture($$MdGestureHandler, $$rAF, $timeout) {
         if (touchActionProperty) {
           // We check for horizontal to be false, because otherwise we would overwrite the default opts.
           this.oldTouchAction = element[0].style[touchActionProperty];
-          element[0].style[touchActionProperty] = options.horizontal === false ? 'pan-y' : 'pan-x';
+          element[0].style[touchActionProperty] = options.horizontal ? 'pan-y' : 'pan-x';
```
DevVersion: I think I initially wrote that lines and I think it was intentional to only allow `pan-x` for horizontal gestures and `pan-y` for vertical gestures.

Can you explain more why this needs to be the other way around?

| State | Value | Description |
| ----- | ------ | --- |
| horizontal | pan-x | Only recognize X-Axis movement s |
| vertical |  pan-y | Only recognize Y-Axis movements |

jelbourn:
Tried this with Chrome 55 and Edge on a MS Surface and everything was peachy. Thanks! LGTM

src/core/services/gesture/gesture.js
EladBezalel: Cause it's the other way around:

 | Value | Description |
| ------ | --- |
| pan-x | The browser is only allowed to perform the horizontal scroll default action. |
|  pan-y | The browser is only allowed to perform the vertical scroll default action. |

There for we need

| State | Value |
| ----- | ------ |
| horizontal | pan-y |
| vertical |  pan-x |

crownofapollo:
This now break my application; I have an md-tabs with md-dynamic-height inside a layout-full and an md-content. Vertical scrolling in the md-content no longer works.

xcafebabe:
Same scenario described by @crownofapollo . Vertical srolling in md-content no longer works.

EladBezalel:
@xcafebabe i will look into it asap

xcafebabe:
Hi @EladBezalel,

I will send you two codepens with same code, turn on Device mode in Chrome and swipe up and down.

With 1.1.2: swipe up-down works.
http://codepen.io/xcafebabe/pen/zNaJBq

With 1.1.3: swipe up-down does not work
http://codepen.io/xcafebabe/pen/NdzLbJ

Tested with
Jessie Debian, Chrome 56.0.2924.87 (64-bit) and without using touch screen, but I'm also experiencing same behavior in an Android 6 mobile device (LG G3), either using google chrome app or an android webview, both version 55.0.2883.91.

Nevertheless thank for the great work you are doing! :)

Paul-Reed:
@EladBezalel Thanks for looking at this, but I'm also unable to swipe up-down too - Android 6 (HTC1)
Issue seems to be confined to Chrome browser, as works fine on Firefox...

PR #10330, opened 2017-01-31
chore: fix response from docs.json

kara:
During the last release, the docs site navigation was broken for a bit because the response from `docs.json` wasn't parsed properly. As a result, the versions were `[]` and didn't display.  Seems that the resp object wasn't updated when we switched from `success()` to `then()`.


ThomasBurleson:
@kara - lgtm

Issue #10351, opened 2017-02-02
md-list-item should support rel attribute since it supports target


pr4v33n:
**Actual Behavior**:
 -  `What is the issue? *`
md-list-item doesn't copy rel attribute.
https://developers.google.com/web/tools/lighthouse/audits/noopener
 -  `What is the expected behavior?`
It should copy rel attribute.

**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:`
 -  `Details:`

**Angular Versions**: *
 -  `Angular Version:` 1.6.1
 -  `Angular Material Version:` 1.1.3

**Additional Information**:
 -  `Browser Type: *`
 -  `Browser Version: *`
 -  `OS: *`
 -  `Stack Traces:`

----
Shortcut to create a [new CodePen Demo](http://codepen.io/team/AngularMaterial/pen/bEGJdd).
Note: `*` indicates required information. Without this information, your issue may be auto-closed.

> Do not modify the titles or questions. Simply add your responses to the ends of the questions.
  Add more lines if needed.

PR #10352, opened 2017-02-02
fix(list): add `rel` to copied attributes for `md-list-item` buttons …

pr4v33n:
…#10351

This is to add support for rel attribute to be copied to generated links.
See https://developers.google.com/web/tools/lighthouse/audits/noopener for more details.


DevVersion:
Thanks for the PR. I think we should add a test for this new change?

ThomasBurleson:
@pr4v33n - please add at least 2 unit tests for this change.

Issue #10380, opened 2017-02-08
md-select: prototypical inheritance and duplicate values


philipyun:
**Actual Behavior**:
 -  `What is the issue? *` When using prototypical inheritance and property defining on Object, the name of the value assigned to object cannot be used as an md-option.  When used, it cannot be selected.  The name will show up in the drop down list but will not select properly
 -  `What is the expected behavior?` Should be able to select anything in md-option list, as long as the ng-value is not repeated in the md-option list.

**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:` http://codepen.io/philipyun/pen/apaBpL
 -  `Details:`

**Angular Versions**: *
 -  `Angular Version:` 1.6.1
 -  `Angular Material Version:` 1.1.3

**Additional Information**:
 -  `Browser Type: *` Chrome/Firefox/Safari
 -  `Browser Version: *` latest
 -  `OS: *` Mac OS Sierra
 -  `Stack Traces:`

```
Error: Duplicate md-option values are not allowed in a select. Duplicate value 'inherited' found.
    at SelectMenuController.self.addOption (https://cdnjs.cloudflare.com/ajax/libs/angular-material/1.1.3/angular-material.js:19479:15)
    at setOptionValue (https://cdnjs.cloudflare.com/ajax/libs/angular-material/1.1.3/angular-material.js:19651:18)
    at Scope.$digest (https://ajax.googleapis.com/ajax/libs/angularjs/1.6.1/angular.js:17814:23)
    at Scope.$apply (https://ajax.googleapis.com/ajax/libs/angularjs/1.6.1/angular.js:18080:24)
    at bootstrapApply (https://ajax.googleapis.com/ajax/libs/angularjs/1.6.1/angular.js:1841:15)
    at Object.invoke (https://ajax.googleapis.com/ajax/libs/angularjs/1.6.1/angular.js:4842:19)
    at doBootstrap (https://ajax.googleapis.com/ajax/libs/angularjs/1.6.1/angular.js:1839:14)
    at bootstrap (https://ajax.googleapis.com/ajax/libs/angularjs/1.6.1/angular.js:1859:12)
    at angularInit (https://ajax.googleapis.com/ajax/libs/angularjs/1.6.1/angular.js:1744:5)
    at https://ajax.googleapis.com/ajax/libs/angularjs/1.6.1/angular.js:32977:5
```

[...]

PR #10381, opened 2017-02-08
fix(md-select) prototypical inheritance and md-option compatibility

philipyun:
change self.options and self.selected semantics to Map structure
When Object gets a new a property associated to it, the name of that property is no longer usable
 by md-option
Checks in the md-select controller check existence of repeated objects in the md-options list on a
 global scale
By converting the semantics of self.options and self.selected to a Map structure we isolate the
 namespaces of the md-options list and the global scheme of the project

 No Breaking changes.  Closes issue #10380


philipyun:
If a more definitive explanation on this pull request is needed, please look at issue #10380

Splaktar:
@philipyun thank you very much for your contribution!

This PR breaks a number of the unit tests. Can you please review the broken tests and resolve the issues?

PR #10399, opened 2017-02-13
mdTabs: Opt reflows / repaints for invisible contents

vladimir-barsuchenko:
**Actual Behavior**:
 -  `What is the issue? *` When transition is performed on `mdTabs' parent` (except `transform, opacity`) the huge amount of reflows / repaints are fired on each frame. Thus it produces bad UX if each tab contains a huge DOM structure and there are a lot of tabs
 -  `What is the expected behavior?` Do not take into account tab's content if it's invisible by settings `display: none` to it.

In a nutshell:

  - By default only selected tab is visible (doesn't have `display: none`). The rest have `display: none`
  - By changing selected tab we: 
      - remove css style `display: none` from content of the tab which should be **selected**
      - set `display: none` for content of the tab that should be **deselected** only after slide animation is ended. If animation is terminated clear timeout
  - To achieve this `tab.scope.select()` & `tab.scope.deselect()` functions are decorated.   Original `deselect()` and `select()` are called when element is still visible. Because user's code could possibly contain DOM logic

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -97,13 +97,96 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
       // (IE/Edge) lose them and start throwing "Invalid calling object" errors, when we
       // compile the element contents down in `compileElement`.
       elements = getElements();
+      // Should be called before updateHeightFromContent() because it has an impact on
+      // calculating height properly
+      initTabsDisplay(elements);
       updateHeightFromContent();
       adjustOffset();
       updateInkBarStyles();
-      ctrl.tabs[ ctrl.selectedIndex ] && ctrl.tabs[ ctrl.selectedIndex ].scope.select();
       loaded = true;
       updatePagination();
     });
+  }
+
+  function initTabsDisplay(elements) {
+    angular.forEach(elements.contents, function (content) {
+      angular.element(content).css({display: 'none'});
+    });
+    angular.forEach(ctrl.tabs, function (tab) {
+      initTabDisplayCallbacks(tab);
+    });
+
+    ctrl.tabs[ ctrl.selectedIndex ] && ctrl.tabs[ ctrl.selectedIndex ].scope.select();
+  }
+
+  function getContentByTabId(tabId) {
+    return $element[0]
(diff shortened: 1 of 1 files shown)


levgaas:
I signed it!

levgaas:
Latest success build didn't use Firefox v.50
Please check because Travis CI is now using Firefox v.50 instead of v.49. 

So builds are failed with error:
`Firefox 50.0.0 (Windows 10 0.0.0) md-input-container directive with ng-messages should select the input value on focus FAILED`

Splaktar:
Thank you very much for your contribution! I've resolved the build issues.

Is there an issue that is associated with this PR? If not, can you please provide a CodePen that demonstrates the issue that you are fixing?

levgaas:
@Splaktar 
Hi. Excuse me for a very very late response =( 

Nope there is no issue associated. Basically it's not a fix. But an improvement.

While I've been working with a complex project I've noticed that having a lot of DOM under tabs could cause perf impact if we animate tabs container. 

In my case I had flex sidenav with tabs. And one of the tab had Slack-like chat. There were like 5 tabs overall.

So my proposition was if we don't see the content of the tab (in other words if tab is not active) we cut it out from reflow / repaints flow by applying `display: none;`.
And in my case it helped a lot.

Of course I could have used something like `ng-if`. But I also wanted to save cute animation tabs have and it'd awesome to have it out of box.

With that all said you can't merge it =) Since after some time I've found some bugs in the implementation which I fixed later on but didn't reflect here. Something related to "initializing tabs on bootstrapping an app"

Splaktar:
No problem. I know it took a long time just to have someone respond to your PR.

It's a known limitation of `md-tabs` that lots of DOM in any of the tabs will cause performance issues. `md-tabs` is specifically designed to not be used with tabs that contain a lot of DOM elements.

For cases where you need to tab between large chunks of DOM, then [md-nav-bar](https://material.angularjs.org/latest/demo/navBar) is recommended. It loses a couple features (animations, tab pagination) in order to provide good performance when there is a lot of content in each tab.

Splaktar:
I'm going to close this with the above advice that `md-nav-bar` should be used for any tabs that have significant amounts of DOM in them.

PR #10412, opened 2017-02-16
Allow users to configure where to append the styles

shahraship:
A lot of people use ShadowDom api now. In this case, the styles by default appending to head isn't ideal. Ideal scenario is if users get an option where to append the styles too.


ThomasBurleson:
You are missing unit tests and a sample demo showing how this will be used.
Rejecting until these are provided.

PR #10464, opened 2017-03-07
fix(mdTooltip): Tooltip parent aria label override

### LINKED ISSUE #10242: fix(mdTooltip): Tooltip parent aria label override

ThomasBurleson:
The tooltip now does not override or update the parent
element `aria-label` unless the parent's `aria-label` does not exist,
or it has been previously set by the tooltip. It determines if it
has been set by the tooltip by placing the `aria-labelledby` attribute
on the parent element and setting it to `tooltip`.

Thx to @bradrich  for original PR.

Fixes #10242.

### DIFF
--- a/src/components/tooltip/tooltip.js
+++ b/src/components/tooltip/tooltip.js
@@ -75,6 +75,7 @@ function MdTooltipDirective($timeout, $window, $$rAF, $document, $interpolate,
 
   function linkFunc(scope, element, attr) {
     // Set constants.
+    var tooltipId = 'md-tooltip-' + $mdUtil.nextUid();
     var parent = $mdUtil.getParentWithPointerEvents(element);
     var debouncedOnResize = $$rAF.throttle(updatePosition);
     var mouseActive = false;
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
FYI - CLA warning is invalid.

ThomasBurleson:
LGTM

PR #10469, opened 2017-03-07
fix(compiler): respect preAssignBindingsEnabled state

### LINKED ISSUE #10016: feat(mdCompiler): support `preAssignBindingsEnabled()`
**Actual Behavior**:
 -  `What is the issue? *` `mdCompiler` [always pre-assigns bindings][1] on controller instances.
 -  `What is the expected behavior?` `mdCompiler` should respect the `preAssignBindingsEnabled()` flag (if present).
    This new configuration option was introduced with angular/angular.js#15095. Although it is not directly related to `mdCompiler`, people might expect the option to be respected by `ngMaterial` as well (possibly as a result of not realizing `ngMaterial` uses its own compiler for certain things).

**Angular Versions**: *
 -  `Angular Version:` 1.5.9+, 1.6.+

[1]: https://github.com/angular/material/blob/f6dedff67dac455a91ac0a957b902f9274492e6a/src/core/services/compiler/compiler.js#L253-L255

DevVersion:
* The `$mdCompiler` now respects the `preAssignBindingsEnabled` state when using AngularJS 1.6 or higher.

**BREAKING CHANGE**

Developers that are upgrading from AngularJS 1.5 to AngularJS 1.6 may run into issues with their custom controllers.

All custom controllers that are are being used inside of Material components, like in the `$mdDialog` or `$mdToast`, can be affected.

> Limited to controllers that are used in combination with the `bindToController` and `locals` option.

Angular Material starts respecting the [`preAssignBindingsEnabled`](angular/angular.js#15352) state when using AngularJS 1.6+

This will mean that bindings are no longer initialized at `constructor` time.

```js
$mdDialog.show({
  locals: {
    myVar: true
  },
  controller: MyController,
  bindToController: true
}

function MyController() {
  // No locals from Angular Material are set yet. e.g myVar is undefined.
}

MyController.prototype.$onInit = function() {
  // Bindings are now set in the $onInit lifecycle hook.
}
```

To provide consistency, Angular Material components with custom controllers now work with the `$onInit` lifecycle hook (**no** other lifecycle hooks)

Fixes #10016

### DIFF
--- a/src/core/services/compiler/compiler.js
+++ b/src/core/services/compiler/compiler.js
@@ -6,7 +6,22 @@
  */
 angular
   .module('material.core')
-  .service('$mdCompiler', MdCompilerService);
+  .service('$mdCompiler', MdCompilerService)
+  .provider('$$mdPreAssignBindings', PreAssignBindingsProvider);
+
+/**
+ * Provider that is used to report the preAssignBindingsEnabled state to the $mdCompiler service.
+ */
+function PreAssignBindingsProvider($compileProvider) {
+  // To avoid that the preAssignBindings state causes issues when upgrading from AngularJS 1.5 to
+  // AngularJS 1.6, AngularJS Material will only respect the preAssignBindingsEnabled state in 1.6
+  var respectPreAssignState = angular.version.major === 1 && angular.version.minor === 6;
+  var fallbackValue = !!$compileProvider.preAssignBindingsEnabled;
+
+  this.$get = function() {
+    return respectPreAssignState ? $compileProvider.preAssignBindingsEnabled() : fallbackValue;
+  };
+}
 
 /**
  * @ngdoc service
@@ -81,7 +96,9 @@ angular
  * </hljs>
  *
  */
-function MdCompilerService($q, $templateRequest, $injector, $compile, $controller) {
+function MdCompilerService($q, $templateRequest, $injector, $compile, $controller,
+                           $$mdPreAssignBindings) {
+
   /** @private @const {!angular.$q} */
   this.$q = $q;
 
@@ -96,6 +113,9 @@ function MdCompilerService($q, $templateRequest, $injector, $compile, $controlle
 
   /** @private @const {!angular.$controller} */
   this.$controller = $controller;
+
+  /** @private @const {boolean} */
+  this.preAssignBindingsEnabled = $$mdPreAssignBindings;
 }
 
 /**
@@ -244,17 +264,12 @@ MdCompilerService.prototype._compileElement = function(locals, element, options)
     // Instantiate controller if the developer provided one.
     if (options.controller) {
 
-      var injectLocals = angular.extend(locals, {
+      var injectLocals = angular.extend({}, locals, {
         $element: element
       });
 
-      var invokeCtrl = self.$controller(options.controller, injectLocals, true, options.controllerAs);
-
-      if (options.bindToController) {
-        angular.extend(invokeCtrl.instance, locals);
-      }
-
-      var ctrl = invokeCtrl();
+      // Create the specified controller instance.
+      var ctrl = self._createController(options, injectLocals, locals);
 
       // Unique identifier for Angular Route ngView controllers.
       element.data('$ngControllerController', ctrl);
@@ -270,6 +285,34 @@ MdCompilerService.prototype._compileElement = function(locals, element, options)
 
   return compileData;
 
+};
+
+/**
+ * Creates and instantiates a new controller with the specified options.
+ * @param {!Object} options Options that include the controller
+ * @param {!Object} injectLocals Locals to to be provided in the controller DI.
+ * @param {!Object} locals Locals to be injected to the controller.
+ * @returns {!Object} Created controller instance.
+ * @private
+ */
+MdCompilerService.prototype._createController = function(options, injectLocals, locals) {
+  var invokeCtrl = this.$controller(options.controller, injectLocals, true, options.controllerAs);
+
+  if (this.preAssignBindingsEnabled && options.bindToController) {
+    angular.extend(invokeCtrl.instance, locals);
+  }
+
+  // Instantiate and initialize the specified controller.
+  var ctrl = invokeCtrl();
+
+  if (!this.preAssignBindingsEnabled && options.bindToController) {
+    angular.extend(invokeCtrl.instance, locals);
+  }
+
+  // Call the $onInit hook if it's present on the controller.
+  angular.isFunction(ctrl.$onInit) && ctrl.$onInit();
+
+  return ctrl;
 };
 
 /**
--- a/src/core/services/compiler/compiler.spec.js
+++ b/src/core/services/compiler/compiler.spec.js
@@ -12,6 +12,13 @@ describe('$mdCompiler service', function() {
     return compileData;
   }
 
+  function setPreAssignBindings(preAssignBindingsEnabled) {
+    module(function($compileProvider) {
+      $compileProvider.preAssignBindingsEnabled(preAssignBindingsEnabled);
+    });
+  }
+
+
   describe('setup', function() {
 
     it('element should use templateUrl', inject(function($templateCache) {
@@ -159,25 +166,75 @@ describe('$mdCompiler service', function() {
         expect(scope.myControllerAs).toBe(data.element.controller());
       }));
 
-      it('should work with bindToController', inject(function($rootScope) {
-        var called = false;
-        var data = compile({
-          template: 'hello',
-          controller: function($scope) {
-            expect(this.name).toBe('Bob');
-            expect($scope.$apply).toBeTruthy(); // test DI working properly
-            called = true;
-          },
-          controllerAs: 'ctrl',
-          bindToController: true,
-          locals: { name: 'Bob' }
-        });
-        var scope = $rootScope.$new();
-        data.link(scope);
-        expect(scope.ctrl.name).toBe('Bob');
-        expect(called).toBe(true);
-      }));
+    });
 
+  });
+
+  describe('with preAssignBindings', function() {
+
+    function compileAndLink(options, preAssignBindings) {
+      var compileData;
+
+      inject(function($mdCompiler, $rootScope) {
+        $mdCompiler.preAssignBindingsEnabled = preAssignBindings;
+        $mdCompiler.compile(options).then(function(data) {
+          data.link($rootScope);
+          compileData = data;
+        });
+
+        $rootScope.$apply();
+      });
+
+      return compileData;
+    }
+
+    it('enabled should assign bindings at instantiation', function() {
+      var isInstantiated = false;
+
+      function TestController($scope) {
+        isInstantiated = true;
+        expect($scope.$apply).toBeTruthy();
+        expect(this.name).toBe('Bob');
+      }
+
+      compileAndLink({
+        template: 'hello',
+        controller: TestController,
+        controllerAs: 'ctrl',
+        bindToController: true,
+        locals: { name: 'Bob' }
+      }, true);
+
+      expect(isInstantiated).toBe(true);
+    });
+
+    it('disabled should assign bindings after constructor', function() {
+      setPreAssignBindings(false);
+
+      var isInstantiated = false;
+
+      function TestController($scope) {
+        isInstantiated = true;
+        expect($scope.$apply).toBeTruthy();
+        expect(this.name).toBeUndefined();
+      }
+
+      TestController.prototype.$onInit = function() {
+        expect(this.name).toBe('Bob');
+      };
+
+      spyOn(TestController.prototype, '$onInit').and.callThrough();
+
+      compileAndLink({
+        template: 'hello',
+        controller: TestController,
+        controllerAs: 'ctrl',
+        bindToController: true,
+        locals: { name: 'Bob' }
+      }, false);
+
+      expect(TestController.prototype.$onInit).toHaveBeenCalledTimes(1);
+      expect(isInstantiated).toBe(true);
     });
 
   });
--- a/package.json
+++ b/package.json
@@ -61,7 +61,7 @@
     "jquery": "^3.0.0",
     "jshint": "^2.9.2",
     "jshint-summary": "^0.4.0",
-    "karma": "^1.0.0",
+    "karma": "1.4.1",
     "karma-chrome-launcher": "^2.0.0",
     "karma-firefox-launcher": "^1.0.0",
     "karma-jasmine": "^1.0.2",
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -604,7 +604,7 @@ function MdDialogProvider($$interimElementProvider) {
     });
 
   /* @ngInject */
-  function advancedDialogOptions($mdDialog, $mdConstant) {
+  function advancedDialogOptions() {
     return {
       template: [
         '<md-dialog md-theme="{{ dialog.theme || dialog.defaultTheme }}" aria-label="{{ dialog.ariaLabel }}" ng-class="dialog.css">',
@@ -631,27 +631,37 @@ function MdDialogProvider($$interimElementProvider) {
         '  </md-dialog-actions>',
         '</md-dialog>'
       ].join('').replace(/\s\s+/g, ''),
-      controller: function mdDialogCtrl() {
-        var isPrompt = this.$type == 'prompt';
-
-        if (isPrompt && this.initialValue) {
-          this.result = this.initialValue;
-        }
-
-        this.hide = function() {
-          $mdDialog.hide(isPrompt ? this.result : true);
-        };
-        this.abort = function() {
-          $mdDialog.cancel();
-        };
-        this.keypress = function($event) {
-          if ($event.keyCode === $mdConstant.KEY_CODE.ENTER) {
-            $mdDialog.hide(this.result);
-          }
-        };
-      },
+      controller: MdDialogController,
       controllerAs: 'dialog',
       bindToController: true,
+    };
+  }
+
+  /**
+   * Controller for the md-dialog interim elements
+   * @ngInject
+   */
+  function MdDialogController($mdDialog, $mdConstant) {
+    // For compatibility with AngularJS 1.6+, we should always use the $onInit hook in
+    // interimElements. The $mdCompiler simulates the $onInit hook for all versions.
+    this.$onInit = function() {
+      var isPrompt = this.$type == 'prompt';
+
+      if (isPrompt && this.initialValue) {
+        this.result = this.initialValue;
+      }
+
+      this.hide = function() {
+        $mdDialog.hide(isPrompt ? this.result : true);
+      };
+      this.abort = function() {
+        $mdDialog.cancel();
+      };
+      this.keypress = function($event) {
+        if ($event.keyCode === $mdConstant.KEY_CODE.ENTER) {
+          $mdDialog.hide(this.result);
+        }
+      };
     };
   }
 
--- a/src/components/toast/toast.js
+++ b/src/components/toast/toast.js
@@ -321,24 +321,7 @@ function MdToastProvider($$interimElementProvider) {
             '    </md-button>' +
             '  </div>' +
             '</md-toast>',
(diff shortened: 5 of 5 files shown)


config/karma.conf.js
```diff
@@ -75,7 +75,13 @@ module.exports = function(config) {
     client: {
       // Do not clear the context as this can cause reload failures with Jasmine
       clearContext:false
+    },
+
+    browserConsoleLogOptions: {
```
ThomasBurleson: What does this do/provide ?

ThomasBurleson:
@DevVersion - Think this one of my favorite PRs for Q1 2017. :1st_place_medal:

mgol:
The Karma not logging issue is a regression: https://github.com/karma-runner/karma/issues/2582. How about locking Karma to 1.4 for now as it's not clear for now how they will resolve the issue and removing the workaround?

src/core/services/compiler/compiler.js
```diff
@@ -6,7 +6,20 @@
...
+
+/**
+ * Function that is called on the configuration phase of AngularJS.
+ * It is used to determine if bindings should be pre-assigned or not.
+ */
+function configureCompiler($compileProvider, $provide) {
+  $provide.constant('$$mdPreAssignBindings',
+    // For Angular versions that do not support `preAssignBindingsEnabled`, always pre-assign the
+    // bindings.
+    $compileProvider.preAssignBindingsEnabled ? $compileProvider.preAssignBindingsEnabled() : true
```
mgol: Are we sure the `$compileProvider.preAssignBindingsEnabled()` output is set in stone at this point?

Also, note that AngularJS 1.7 will remove `$compileProvider.preAssignBindingsEnabled` and the bindings so this code is not future-proof.

If the method is not present, it's safer to check the library version and set `true` only if it's older than 1.6.

gkalpak: > Are we sure the `$compileProvider.preAssignBindingsEnabled()` output is set in stone at this point?

As a matter of fact, it is never set in stone :grin: (you can change it any time). Not that this is a common usecase. 

```js
.provider('$$mdPreAssignBindings', function ($compileProvider) {
  var hardcodedValue = (angular.version.major === 1 && angular.version.minor < 6) ? true : false;
  var getValue = $compileProvider.preAssignBindingsEnabled ?
    function () { $compileProvider.preAssignBindingsEnabled(); } :
    function () { return hardcodedValue; };

  this.$get = function () { return getValue; };  
})
```

src/core/services/compiler/compiler.js
```diff
@@ -248,13 +266,8 @@ MdCompilerService.prototype._compileElement = function(locals, element, options)
...
 
-      var invokeCtrl = self.$controller(options.controller, injectLocals, true, options.controllerAs);
-
-      if (options.bindToController) {
-        angular.extend(invokeCtrl.instance, locals);
-      }
-
-      var ctrl = invokeCtrl();
+      // Create the specified controller instance.
+      var ctrl = self._createController(options, injectLocals);
```
gkalpak: Not directly related to this PR, but it seems that you are accidentally putting `$element` into `locals`. Essentially, `locals` and `injectLocals` are the same object, but this doesn't seem intentional (and indeed assigning `$element` to the controller instance seems strange).

Was it intentional?

src/core/services/compiler/compiler.spec.js
```diff
@@ -163,21 +163,27 @@ describe('$mdCompiler service', function() {
...
-      }));
 
+
+        function TestController($scope) {
+          called = true;
+          expect($scope.$apply).toBeTruthy(); // Ensure that DI works properly
+        }
+
+        TestController.prototype.$onInit = function() {
+          expect(this.name).toBe('Bob');
```
gkalpak: There is no verification that `$onInit()` is actually called.

src/core/services/compiler/compiler.js
ThomasBurleson: Uugh.

src/core/services/compiler/compiler.spec.js
ThomasBurleson: What do you mean.

gkalpak: For example, for the constructor you check that `called` has been set to true, so you know that the constructor has been called (and the expectations it contains verified).
There is nothing similar for `$onInit()`.

You could add `expect('foo').toBe('bar')` inside `$onInit()` and the test would not fail if `$onInit()` is never called. This means that the test may not catch some future regression.

src/core/services/compiler/compiler.js
mgol: @gkalpak your implementation is good, I'd just drop the ` ? true : false` part as it's redundant here.

src/core/services/compiler/compiler.spec.js
mgol: That's why I like QUnit's `assert.expect(number)` so much but Jasmine has no equivalent.

A proper test here should also probably spy on `$onInit` (with `calledThrough()`) to ensure it's not skipped.

src/core/services/compiler/compiler.js
DevVersion: Good point.

DevVersion:
@gkalpak @mgol @ThomasBurleson Updated the PR.

src/core/services/compiler/compiler.js
gkalpak: I do this `true/false` thing too often :smiley:

src/core/services/compiler/compiler.spec.js
```diff
@@ -159,31 +166,71 @@ describe('$mdCompiler service', function() {
...
+        controllerAs: 'ctrl',
+        bindToController: true,
+        locals: { name: 'Bob' }
+      });
+
+      expect(isInstantiated).toBe(true);
+    });
+
+    it('disabled should assigns bindings after constructor', function() {
+      configureCompiler(false);
```
ThomasBurleson: let's use:

```js
var shouldPreAssign = false;
var isIntantiated = false;

configureCompiler(shouldPreAssign);
```

DevVersion: What advantage does this have? I think the spec name and function parameter already explains that this changes the `preAssignBindings` behavior.

ThomasBurleson: I had to search for how it was disabled. the `configureCompiler(false)` was not intuitive.

DevVersion: Ah I understand. How is `setPreAssignBindings`?

ThomasBurleson:
@DevVersion - please squash your commits.

src/core/services/compiler/compiler.js
```diff
@@ -6,7 +6,20 @@
...
+
+/**
+ * Provider that is used to report the preAssignBindingsEnabled state to the $mdCompiler service.
+ */
+function PreAssignBindingsProvider($compileProvider) {
+  var fallbackValue = angular.version.major === 1 && angular.version.minor < 6;
+  var hasPreAssignState = !!$compileProvider.preAssignBindingsEnabled;
+
+  this.$get = function() {
+    return hasPreAssignState ? $compileProvider.preAssignBindingsEnabled() : fallbackValue;
```
gkalpak: This will not catch changes that will happend at runtime (which is theoretically possible, although highly improbable).

src/core/services/compiler/compiler.spec.js
```diff
@@ -12,6 +12,13 @@ describe('$mdCompiler service', function() {
     return compileData;
   }
 
+  function setPreAssignBindings(preAssignBindingsEnabled) {
+    module(function($compileProvider) {
+      $compileProvider.preAssignBindingsEnabled(preAssignBindingsEnabled);
+    })
```
gkalpak: Missing semicolon 😱

src/core/services/compiler/compiler.js
```diff
@@ -248,13 +266,8 @@ MdCompilerService.prototype._compileElement = function(locals, element, options)
...
 
-      var invokeCtrl = self.$controller(options.controller, injectLocals, true, options.controllerAs);
-
-      if (options.bindToController) {
-        angular.extend(invokeCtrl.instance, locals);
-      }
-
-      var ctrl = invokeCtrl();
+      // Create the specified controller instance.
+      var ctrl = self._createController(options, injectLocals, locals);
```
gkalpak: Currently, `injectLocals` and `locals` are the same object. I assume this a bug, but not directly related to (or introduced by) this PR.

src/core/services/compiler/compiler.spec.js
```diff
@@ -159,25 +166,71 @@ describe('$mdCompiler service', function() {
...
+      var data = compile(options);
+
+      inject(function($rootScope) {
+        data.link($rootScope);
+      });
+
+      return data;
+    }
+
+    it('enabled should assigns bindings at instantiation', function() {
```
gkalpak: _assings_ --> _assign_

src/core/services/compiler/compiler.spec.js
```diff
@@ -159,25 +166,71 @@ describe('$mdCompiler service', function() {
...
+        controller: TestController,
+        controllerAs: 'ctrl',
+        bindToController: true,
+        locals: { name: 'Bob' }
+      });
+
+      expect(isInstantiated).toBe(true);
+    });
+
+    it('disabled should assigns bindings after constructor', function() {
```
gkalpak: _assigns_ --> _assign_

src/core/services/compiler/compiler.js
DevVersion: Yeah I intentionally did it that way. If users change it at runtime then I don't think this should be respected. Providers should be set at **config** phase.

src/core/services/compiler/compiler.js
DevVersion: In what way are those similar? `injectLocals` contains the `$element` DI token.

gkalpak: In the way that `angular.extend(locals, {$element: ...})` extends `locals` with `$element` and then returns `locals`. So, `injectLocals === locals`. I assume the intended behavior was `injectLocals = angular.extend({$element: ...}, locals)` or `injectLocals = angular.extend({}, locals, {$element: ...})`.

ThomasBurleson:
@DevVersion - please squash your commits.

DevVersion:
@ThomasBurleson Done.

mgol:
@DevVersion did you see my comment regarding the Karma update?

DevVersion:
@mgol Done. @ThomasBurleson Made the changes we discussed on our meeting.

ThomasBurleson:
@mgol, @gkalpak - last review before presubmit ?

DevVersion:
I just saw that I need to make one adjustment for the tests. Ignore the failing test for review.

src/core/services/compiler/compiler.js
```diff
@@ -6,7 +6,21 @@
...
+  .service('$mdCompiler', MdCompilerService)
+  .provider('$$mdPreAssignBindings', PreAssignBindingsProvider);
+
+/**
+ * Provider that is used to report the preAssignBindingsEnabled state to the $mdCompiler service.
+ */
+function PreAssignBindingsProvider($compileProvider) {
+  // To avoid breaking changes in Material, AngularJS Material will only respect the
+  // preAssignBindingsEnabled state in AngularJS 1.6.X and higher.
+  var respectPreAssignState = angular.version.major === 1 && angular.version.minor > 5;
```
mgol: This will break with AngularJS `~1.5.10` if the developer sets the `preAssignBindingsEnabled` flag to `false`.

What are the breaking changes mentioned in a comment?

src/core/services/compiler/compiler.js
```diff
@@ -6,7 +6,21 @@
...
+
+  this.$get = function() {
+    return respectPreAssignState ? $compileProvider.preAssignBindingsEnabled() : true;
```
mgol: This will break in AngularJS 1.7+. We had a thread about it with @gkalpak suggesting an implementation that would work, what did happen with that? :)

DevVersion: The implementation did not cover the changes we discussed in our meeting. Also we don't want to re-check the `preAssignBindingsEnabled` state for each `$mdCompiler` call.

I changed it now to work as we discussed on our meeting
* 1.5: Bindings will be always pre-assigned as same as with Material before.
* 1.6: Angular Material will start respecting the `preAssignBindingsEnabled` state
* 1.7: Compiler will always sets preAssignBindingsEnabled to false.

src/core/services/compiler/compiler.js
DevVersion: Why will it break with `1.5.10`? I thought it would rather break with `1.7` as the function will be removed.

src/core/services/compiler/compiler.js
```diff
@@ -6,7 +6,22 @@
...
+  .provider('$$mdPreAssignBindings', PreAssignBindingsProvider);
+
+/**
+ * Provider that is used to report the preAssignBindingsEnabled state to the $mdCompiler service.
+ */
+function PreAssignBindingsProvider($compileProvider) {
+  // To avoid that the preAssignBindings state causes issues when upgrading from AngularJS 1.5 to
+  // AngularJS 1.6, AngularJS Material will only respect the preAssignBindingsEnabled state in 1.6
+  var respectPreAssignState = angular.version.major === 1 && angular.version.minor === 6;
+  var fallbackValue = !!$compileProvider.preAssignBindingsEnabled;
```
gkalpak: As @mgol has pointed out, this will break in the future (most probably 1.7+). The plan is to remove the `preAssingBindingsEnabled` method and not support pre-assignement neither by default, nor as an opt-in (this is the PR btw: angular/angular.js#15782).

Base on what we have discussed in the last meeting, you need something like this:

```js
var defaultPreAssignValue = angular.version.major === 1 && angular.version.minor < 6;
var respectPreAssignState = angular.version.major === 1 && angular.version.minor > 5 &&
                            $compileProvider.preAssignBindingsEnabled;

this.$get = function() {
  return respectPreAssingState ? $compileProvider.preAssignBindingsEnabled() : defaultPreAssignState;
};
```

gkalpak: BTW, the current implementation will also break in 1.5.0-1.5.9 (where there is no `$compileProvider.preAssignBindingsEnabled` method).

src/core/services/compiler/compiler.js
```diff
@@ -6,7 +6,22 @@
...
+  .service('$mdCompiler', MdCompilerService)
+  .provider('$$mdPreAssignBindings', PreAssignBindingsProvider);
+
+/**
+ * Provider that is used to report the preAssignBindingsEnabled state to the $mdCompiler service.
+ */
+function PreAssignBindingsProvider($compileProvider) {
+  // To avoid that the preAssignBindings state causes issues when upgrading from AngularJS 1.5 to
+  // AngularJS 1.6, AngularJS Material will only respect the preAssignBindingsEnabled state in 1.6
+  var respectPreAssignState = angular.version.major === 1 && angular.version.minor === 6;
```
mgol: About not respecting the flag in 1.5: it's about how `MdCompilerService#_createController` behaves so it will not break 1.5, it will just still preassign bindings in 1.5.10 even if the developer sets `preAssignBindingsEnabled` to `false`, I see it now.

We could have respected the flag in 1.5.10 if your PR from a while ago accounting for this flag wasn't merged before this one as it wouldn't be breaking then - no one would be using 1.5.10 with Material & this flag set to `false` as this configuration was broken.

I assume Material needs to ignore this flag because Material 1.1.3 works fine in 1.5.10 with this flag set to `false`? (Does it?)

mgol: BTW, just in case we'd have to revert the removal and defer it to 1.8, it would be safer to check for `angular.version.minor > 5` instead of `angular.version.minor === 6`

src/core/services/compiler/compiler.js
mgol: @gkalpak 
> BTW, the current implementation will also break in 1.5.0-1.5.9 (where there is no `$compileProvider.preAssignBindingsEnabled` method).

How so? `fallbackValue` will then be `false` but it won't be used anyway.

src/core/services/compiler/compiler.js
mgol: BTW, regarding the argument for not respecting the flag in `~1.5.10` but respecting in `~1.6` - if someone was using Material with AngularJS 1.6 and it worked for them as the components they used worked fine, this change will break them.

Maybe it would be safer to release this as 1.2.0, not 1.1.5 and then not ignoring the flag if it exists?

What do you think? @DevVersion @gkalpak @ThomasBurleson

src/core/services/compiler/compiler.js
gkalpak: `fallbackValue` will be `false` and it _will_ be used. `respectPreAssignState` would be `false`, so the (incorrect) fallback value will be used.

<sub>(To be clear, by "current implementation" I mean the implementation currently used in this PR.)</sub>

DevVersion: Well, I thought that `preAssignBindingsEnabled` has been added in 1.5.X and changed in 1.6 to `false` by default and will be removed in 1.7.

src/core/services/compiler/compiler.js
gkalpak: > if someone was using Material with AngularJS 1.6 and it worked for them as the components they used worked fine, this change may break them.

🤢 True.
It sounds like there is indeed no (reaconable) way to make this non-breaking (without making it opt-in - e.g. with an `mdCompiler` flag) 😞

src/core/services/compiler/compiler.js
mgol: @gkalpak Ah, you're right.

@DevVersion The flag has been added in 1.5.9 which was released after a long wait and even after 1.6.0 was released. It's highly likely most of 1.5.x users will not have this flag available.

mgol: The fallback value should be defined based on AngularJS version, not the presence of the `$compileProvider.preAssignBindingsEnabled` flag.

Besides, as far as I understand, Material still supports AngularJS 1.3+, doesn't it?

jelbourn:
@mmalerba can you presubmit this (but don't merge it if it passes)?

mmalerba:
@jelbourn latest is good, v1_1 has ~10 failures (this is with merging this PR into master at the commit v1_1 is currently synced to)

mgol:
Since there are edge cases that would break when this PR lands in its current state, I think we could do the following to be able to not wait for 1.2 with landing it:

1. Add a new public flag (sth like `$mdCompilerProvider.respectPreAssignBindingsState(true|false)`) that would make Material respect the value of AngularJS' `$compilerProvider.preAssignBindingsEnabled()`.
2. Make this flag `false` by default, turn it to `true` by default in `1.2.0`, remove in `1.3.0`. Optionally, the flag could be made a noop in Material `1.2.0` (and removed in `1.3.0`) if we didn't want to keep this logic that long.
3. Change the logic in the current PR to:
3.1. If `$mdCompilerProvider.respectPreAssignBindingsState() === false`: always preassign bindings.
3.2. If `$mdCompilerProvider.respectPreAssignBindingsState() === true`:
3.2.1. If `$compilerProvider.preAssignBindingsEnabled` exist, preassign bindings if and only if it returns `true`.
3.2.2. If it doesn't exist then:
3.2.2.1. With AngularJS <1.6: preassign bindings.
3.2.2.2. With AngularJS >=1.6: don't preassign bindings.

What do you think?

ThomasBurleson:
@mmalerba - Can you elaborate. I admit that I do not understand your response. Is this PR invalid for a presubmit? If yes, then why ?

ThomasBurleson:
@mgol - let's talk about your ideas offline in a call. Thx

mmalerba:
@ThomasBurleson there were 10 failures when I checked, which is a relatively small number and may include some that were just flakes. It can probably be merged with some shepherding from the caretaker

mgol:
This PR evolved into PR #10726; PTAL.

DevVersion:
@mgol Thanks for continuing this. Closing my PR here.

PR #10472, opened 2017-03-08
fix(checkbox): click handling when ngCheckbox is set

### LINKED ISSUE #10468: ngChecked Bug with md-checkbox
I'm experiencing different behavior with an md-checkbox than with a vanilla checkbox and expected they would work the same. The [master/slave example from the ngChecked page](https://docs.angularjs.org/api/ng/directive/ngChecked) does not work when the checkboxes are md-checkbox but does work when <input type="checkbox" ...> is used.

Here is a [Plunk with the example from the ngChecked documentation but with the ngMaterial dependencies added](https://plnkr.co/edit/gccAaC56xl9HHRm8GeID?p=preview). It behaves as expected — the 'slave' checkbox can be toggled whether or not the 'master' checkbox is selected.

Now here is a [Plunk with the same example but both instances of <input type="checkbox" ...>  have been changed to <md-checkbox> elements](https://plnkr.co/edit/xFd6V4BulOHKwnPrAKb5?p=preview) (with closing tags). **Note that the 'slave' checkbox does not toggle its state when clicked, irrespective of whether the 'master' is selected or not.**

busiq:
When md-checkbox has ng-checked set, checkbox should still be clickable, and work as regular checkbox. 
The only case I see right now where `attr.checked` should be used as `$viewValue`  is when `ng-click` && `ng-checked` is provided - we can then assume `ng-checked` could be changed by `ng-click` callback. This behavior is used in `demoSyncing` for example.

Fixes #10468

### DIFF
--- a/src/components/checkbox/checkbox.js
+++ b/src/components/checkbox/checkbox.js
@@ -184,7 +184,7 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
 
         scope.$apply(function() {
           // Toggle the checkbox value...
-          var viewValue = attr.ngChecked ? attr.checked : !ngModelCtrl.$viewValue;
+          var viewValue = attr.ngChecked && attr.ngClick ? attr.checked : !ngModelCtrl.$viewValue;
 
           ngModelCtrl.$setViewValue(viewValue, ev && ev.type);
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
Thx @busiq.

PR #10474, opened 2017-03-08
[MdPanel] Fix offscreen positioning bug

ErinCoughlan:
@jelbourn @tinayuangao I think this should fix the positioning bug. Will run the presubmit and verify.


ErinCoughlan:
I signed it!

PR #10481, opened 2017-03-10
fix(chips): validate input before adding chip on enter

ianchi:
userInputNgModelCtrl is tested for validity before adding chip only when
adding on blur, but not on enter (or separator key).
Add validation consistent validation on both events.

Signed-off-by: Adrian Panella <ianchi74@outlook.com>

### DIFF
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -267,7 +267,7 @@ MdChipsCtrl.prototype.inputKeydown = function(event) {
     event.preventDefault();
 
     // Only append the chip and reset the chip buffer if the max chips limit isn't reached.
-    if (this.hasMaxChipsReached()) return;
+    if (!this.shouldAdd()) return;
 
     this.appendChip(chipBuffer.trim());
     this.resetChipBuffer();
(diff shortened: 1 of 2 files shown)


ianchi:
I signed it!

ThomasBurleson:

src/components/chips/js/chipsController.js
```diff
@@ -267,7 +267,7 @@ MdChipsCtrl.prototype.inputKeydown = function(event) {
     event.preventDefault();
 
     // Only append the chip and reset the chip buffer if the max chips limit isn't reached.
-    if (this.hasMaxChipsReached()) return;
+    if (!this.shouldAdd()) return;
```
Splaktar: Can you please update the comment here to be accurate?

src/components/chips/js/chipsController.js
```diff
@@ -770,7 +770,7 @@ MdChipsCtrl.prototype.configureAutocomplete = function(ctrl) {
  * Whether the current chip buffer should be added on input blur or not.
  * @returns {boolean}
  */
-MdChipsCtrl.prototype.shouldAddOnBlur = function() {
+MdChipsCtrl.prototype.shouldAdd = function(onBlur) {
```
Splaktar: Can you please add the JSDoc for the `onBlur` param?

src/components/chips/js/chipsController.js
```diff
@@ -770,7 +770,7 @@ MdChipsCtrl.prototype.configureAutocomplete = function(ctrl) {
  * Whether the current chip buffer should be added on input blur or not.
```
Splaktar: Can you please update this comment to be accurate?

Splaktar:
Is it also possible that you could provide a demo via CodePen of this issue? It doesn't appear to be something that is reproducible on the docs site.

PR #10488, opened 2017-03-13
ng-messages node must be outside the md-autocomplete element

EightArmCode:
My ng-messages were not visible with the element placed inside the md-autocomplete element. It works perfectly when placed just outside of it. If that is correct, then this example would benefit from this update.


EightArmCode:
I signed it!

PR #10502, opened 2017-03-16
Fix casing and add clarity to documentaton on mdPanelRef

bryanrideshark:
This adds more documentation so that users of $mdPanel are able to easily inject the correct reference to the created panel in their controller.

The example source code has 'mdPanelRef' as the name of its import, but the rendered HTML in the docs site, converts this to 'MdPanelRef'.

This might be indicative of a larger problem with the dgeni process; Either way, I think the lines I added to the documentation should be enough to remove confusion.

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -111,6 +111,7 @@ angular
  *   }
  *
  *   function PanelMenuCtrl(mdPanelRef) {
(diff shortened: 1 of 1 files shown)


bryanrideshark:
I signed it. My employer requires a work Github account and a private Github account to be separate from one another for some reason.

Issue #10144, opened 2016-12-15
mdDatepicker calendar opens up to wrong view


ppham27:
**Actual Behavior**:
 -  `What is the issue? *` When I click on the datepicker, the calendar view shows the wrong month.
 -  `What is the expected behavior?` It should open up to the currently selected date or the current date.

**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:` https://material.angularjs.org/latest/demo/datepicker
 -  `Details:` Here's a video, https://youtu.be/N3EK1ByvhV0

**Angular Versions**: *
 -  `Angular Version:` 1.5.9
 -  `Angular Material Version:` 1.1.1

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *` 55
 -  `OS: *` Ubuntu 12.04
 -  `Stack Traces:` N/A

I should note that this problem doesn't occur in OS X. It also only occurs on the Angular Material Demo page, but works fine in the CodePen.

jsgoupil:
I see the same problem with Angular Material 1.0.6. It actually scrolls to April 2017 for me.

xxmuaddib:
I have created a plnkr with a similiar problem. md-datepicker doesn't even render.

- Angular Version: 1.6.0
- Angular Material Version: 1.1.1
- Google Chrome Version: 54

[https://plnkr.co/edit/4YBK4Eb8LVNeQbH5wIac](https://plnkr.co/edit/4YBK4Eb8LVNeQbH5wIac)

rhinoceraptor:
I have this problem with:

- Angular 1.5.9
- Angular Material 1.1.1
- Chrome 55
- OS: macOS 10.12.1

DAquilina:
Confirming same problem:

- Angular 1.5.8
- Angular Material 1.1.1
- Chrome 55.0.2883.87 (64-bit)
- Ubuntu 16.04.1 LTS

Note that this issue does not seem to occur in Firefox.

julioponce:
Same issue with:

Angular: 1.5.8
Material: 1.1.0
Browser: Chrome 55.0.2883.95 (64-bits)
Mac OS X El Capitan 10.11.6

Dinosys:
Same issue with:

Angular: 1.5.8
Material: 1.1.1
Browser: Chrome 55.0.2883.87
OS: Windows 10

Adding: 
[...]

PR #10506, opened 2017-03-17
fix(virtualRepeat): Fix datepicker scroll to wrong selected date

ernsheong:
Fixes https://github.com/angular/material/issues/10144, or rather the related issue that remains unfixed (scroll to wrong position). Appreciate @ppham27 taking a look to make sure that original scenario in https://github.com/angular/material/issues/10144 remains fixed.

Datepicker scrolls to wrong date when a min date is set. 
Issue reproducible in Chrome here:
https://material.angularjs.org/latest/demo/datepicker "Custom calendar trigger" (Click "Open")


ernsheong:
Superceded by https://github.com/angular/material/pull/10511

PR #10517, opened 2017-03-20
chore(babel): added babel to support es6

EladBezalel:


Frank3K:
What about [babel-preset-env](https://github.com/babel/babel-preset-env)? It seems like that is [the recommendation](https://twitter.com/left_pad/status/837443603379204096) now.

EladBezalel:
@Frank3K I certainly thought of it but i realized it shouldn't be env based..
But now i'm thinking of it again and maybe we should do 
```
["env", {
      "targets": {
        "browsers": ["last 2 versions", "safari >= 7"]
      }
    }]
```

Frank3K:
I think you should use the list that is used by autoprefixer (in scripts/gulp-utils.js).

EladBezalel:
@DevVersion what do you think?

DevVersion:
@EladBezalel I think that the `babel-preset-env` would work. Also like the idea of sharing the browsers array with the one from the autoprefixer.

Frank3K:
Note that you can add the browserlist to package.json, or set it in a separate config file as described [here](https://github.com/ai/browserslist/blob/master/README.md). Autoprefixer and babel-env-preset will automatically pick it up.

EladBezalel:
@Frank3K @DevVersion please review

gulp/util.js
```diff
@@ -48,6 +49,9 @@ function buildJs () {
 
   var jsBuildStream = gulp.src( jsFiles )
       .pipe(filterNonCodeFiles())
+      .pipe(babel({
+          presets: [["env", {targets: {browsers: config.browsers}}]]
```
Frank3K: I think `uglify: true` should be added here. [Readme](https://github.com/babel/babel-preset-env#targetsuglify) and [PR](https://github.com/babel/babel-preset-env/pull/178).

EladBezalel: sounds like `useBuiltIns` would work better

ThomasBurleson:
This is look good. Let's finish with final changes and one last review.

gulp/util.js
Frank3K: If I understand it correctly `useBuiltIns` alone won't solve the minification issues. I'm not sure what uglify + `useBuiltIns` does.

The option `useBuiltIns`seems to require a module loader (see [explanation] (https://github.com/babel/babel-preset-env/releases/tag/v1.0.0)).

ThomasBurleson:
Closing as not valid for stabilization.

PR #10520, opened 2017-03-20
fix(datepicker): Fix validation in IE.

### LINKED ISSUE #9994: datepicker: validation behaves differently in IE

topherfangio:
Adds in validation for the input as a last check as IE interpreted the date '11/14/2016d' as valid which is incorrect.

_Note: This PR supersedes #10015 by adding a test._

Fixes #9994. Closes #10015.

### DIFF
--- a/src/components/datepicker/js/datepickerDirective.js
+++ b/src/components/datepicker/js/datepickerDirective.js
@@ -607,7 +607,29 @@
       this.ngModelCtrl.$setValidity('valid', date == null);
     }
 
+    var input = this.inputElement.value;
+    var parsedDate = this.locale.parseDate(input);
+
+    if (!this.isInputValid(input, parsedDate) && this.ngModelCtrl.$valid) {
+      this.ngModelCtrl.$setValidity('valid', date == null);
+    }
+
     angular.element(this.inputContainer).toggleClass(INVALID_CLASS, !this.ngModelCtrl.$valid);
+  };
+
+  /**
+   * Check to see if the input is valid as the validation should fail if the model is invalid
+   *
+   * @param {String} inputString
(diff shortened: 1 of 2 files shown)


richardhughes:
Hey @topherfangio, Looks good to me, I've just noticed that Google Bot has put a message on this to get confirmation.

I couldn't get the test runner to work in the end, but that was down to my old development machine.

topherfangio:
@richardhughes Thanks! I've marked it as `cla: yes` since you approved it. Thomas should review it when he has a chance, but he and much of the team is at ngConf right now, so it will likely happen after they return.

ThomasBurleson:
@topherfangio - can you squash your commits.

Splaktar:
Closing this in favor of https://github.com/angular/material/pull/11110 since the request to squash commits was not resolved.

PR #10524, opened 2017-03-21
fix(input): Fix placeholder text being read twice by screen readers.

### LINKED ISSUE #10510: placeholder text is copied to aria-label, resulting in the text read twice
**Actual Behavior**:
 -  The screen reader reads out "Address 2 with hint address 2,  edit text", which reads the placeholder text twice.

 - The reason is that the placeholder text is copied to aria-label if there is no aria-label present (https://github.com/angular/material/blob/master/src/components/input/input.js#L343)

 - By removing the redundant aria-label, the screen reader reads out "Address 2,  edit text", which should be as expected.

**Demo** (or steps to reproduce the issue): *
 - Please listen to the screen reader when the 'Address 2' input get focus in angular material demo page:
https://material.angularjs.org/latest/demo/input

**AngularJS Versions**: *
 -  `AngularJS Version:`1.5.5
 -  `AngularJS Material Version:`1.1.3

----

  Add more lines if needed.

du6:
Fixes #10510

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -339,8 +339,9 @@ function inputTextareaDirective($mdUtil, $window, $mdAria, $timeout, $mdGesture)
     var errorsSpacer = angular.element('<div class="md-errors-spacer">');
     element.after(errorsSpacer);
 
-    if (!containerCtrl.label) {
-      $mdAria.expect(element, 'aria-label', attr.placeholder);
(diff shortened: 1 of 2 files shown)


du6:
I signed it!

ThomasBurleson:
@du6 - still see it as unsigned.

ThomasBurleson:
@crisbeto - can you review?

PR #10545, opened 2017-03-30
Update 03_configuring_a_theme.md

memee:
It was not true that 'white' was in 'grey' pallete. For reference look here: https://material.io/guidelines/style/color.html#color-color-palette


norkunas:
@memee and how it's not true https://github.com/angular/material/blob/master/src/core/services/theming/theme.palette.js#L333 ?

PR #10596, opened 2017-04-18
fix(tooltip): resolve expressions against correct scope

kara:
Reinstates scope fix from https://github.com/angular/material/commit/685b9029d699c486ef27f9d00a0c49b90a363552, which was accidentally reverted in https://github.com/angular/material/pull/10464.  Blocking 1.1 release due to Google test failures.

cc: @ThomasBurleson

### DIFF
--- a/src/components/tooltip/tooltip.spec.js
+++ b/src/components/tooltip/tooltip.spec.js
@@ -14,7 +14,8 @@ describe('MdTooltip Component', function() {
   // Test filter for ensuring tooltip expressions are evaluated against the correct scope.
   angular.module('fullNameFilter', []).filter('fullName', function() {
     return function(user) {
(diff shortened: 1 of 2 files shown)


src/components/tooltip/tooltip.spec.js
```diff
@@ -14,7 +14,8 @@ describe('MdTooltip Component', function() {
   // Test filter for ensuring tooltip expressions are evaluated against the correct scope.
   angular.module('fullNameFilter', []).filter('fullName', function() {
     return function(user) {
-      return user ? user.name.first + ' ' + user.name.last : "";
+        // Intentionally reference user without checking whether it is defined. Do not change!
```
jelbourn: nit: `reference` > `dereference`

jelbourn:
LGTM

PR #10537, opened 2017-03-28
fix(virtualRepeat): Fix datepicker scroll to wrong current date

ernsheong:
Fixes https://github.com/angular/material/issues/10144

Datepicker scrolls to wrong date in Chrome.
Issue reproducible in Chrome here: https://material.angularjs.org/latest/demo/datepicker -> “Custom calendar trigger" (Click "Open" a few times and it will reoccur)
Video of bug: https://youtu.be/4Tj0k6b8rVg

PR #10600, opened 2017-04-19
fix(tooltip): use different attribute to track setting of aria-label

kara:
We recently started setting `aria-labelledby` on parent elements to track if the `aria-label` had already been set by the tooltip. It seems we should use a different attribute as a marker because `aria-labelledby` will take precedence over `aria-label` when set on the same element. In this case, it will point to an element that only exists on the DOM when the tooltip is showing, which hurts accessibility. This is currently blocking 1.1.4 release due to Google breakages.

cc: @ThomasBurleson


jelbourn:
LGTM

ernsheong:
@kara can you help with https://github.com/angular/material/pull/10537 as well?

ernsheong:
@kara can you help with the PR here https://github.com/angular/material/pull/10537 as well? thanks

PR #10603, opened 2017-04-20
update(autoComplete): allow clear button even if directive is disabled

achuinard:
I am attempting to use md-autocomplete and 'lock' in a users selection once an autocomplete is selected. I ng-disable the directive accordingly. However, I still want the user to be able to clear the autocomplete, even if it is disabled - I simply don't want them to be able to edit the existing text. This pull request will still allow the clear button to be shown even if it is disabled. If the developer doesn't want a clear button when the control is disabled, they can simply update their boolean expression inside of the md-clear-button attribute.


achuinard:
I signed it!

PR #10650, opened 2017-05-09
fix(panel): bounds not unfolded for positioning

### LINKED ISSUE #10543: md-tooltip positioning problem with md-panel change
**Actual Behavior**:
 -  `What is the issue? *` The second time the tooltip appears it is not positioned correct when $compileProvider.debugInfoEnabled(false)
 -  `What is the expected behavior?` tooltip should stay on top

 -  `CodePen Demo which shows your issue:` http://codepen.io/anon/pen/mWQJve?editors=1111
 -  `Details:` when $compileProvider.debugInfoEnabled(false) it does fail; with $compileProvider.debugInfoEnabled(true) it works as expected

**AngularJS Versions**: *
 -  `AngularJS Version:` 1.5.5
 -  `AngularJS Material Version:` 1.1.3

**Additional Information**:
 -  `Browser Type: *` Chrome + Firefox
 -  `Browser Version: *` Version 56.0.2924.76 & 52.0.1 (64-bit)
 -  `OS: *` kubuntu
 -  `Stack Traces:` 

i was searching for hours. The new "small" version update changed a lot of basic things. i rly was suprised. when md-tooltip switched to use md-panel it stoped working correctly. I found out that it has sth to do with debugInfoEnabled(false). But why should this affect the tooltip/panel ?

is there any hotfix or workaround for it? just simply enabling debugInfo is not a solution.

----

  Add more lines if needed.

Best,
Florian

oliversalzburg:
When Angular debug information is disabled, operation is sped up enough so that the bounds of the panel are not always unfolded when they are checked during positioning.
This change introduces a delay of one digest cycle to let the element unfold.

Closes #10543

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -1376,6 +1376,9 @@ function MdPanelRef(config, $injector) {
   /** @private @const {!angular.$window} */
   this._$window = $injector.get('$window');
 
+  /** @private @const {!angular.$timeout} */
+  this._$timeout = $injector.get('$timeout');
+
   /** @private @const {!Function} */
   this._$$rAF = $injector.get('$$rAF');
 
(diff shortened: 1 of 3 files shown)


oliversalzburg:
Please tell me if this is the way to go before I update all those tests :)

EladBezalel:
@oliversalzburg just to make sure - you did tested it manually to see that it's working right?

IMO that's the best solution atm and that should do the trick.
please go ahead and fix the failing tests!

Thanks again!

oliversalzburg:
@EladBezalel Good that you asked :P Because I didn't and I forgot to actually *inject* `$timeout`.

Well, I would have found it eventually when checking out the tests, but I'll finish it up proper now. As the missing injection is only the cause for a few failures.

oliversalzburg:
@EladBezalel Sorry, I think I messed that up. I went a [different route](10651) now. Works much more reliably.

ernsheong:
@EladBezalel any idea when the next release would come? 1.1.5 deadline was 18th May... seems very cumbersome to make a custom release on own fork due to the build process. Would really help us in our project at work if you all can provide a release asap 👍

PR #10683, opened 2017-05-23
chore(chips): remove test for input type=number

jelbourn:
Because `selectionStart` an `selectionEnd` cannot be used with `<input type="number">`, there is no real way to capture this behavior in a unit test.


jelbourn:
Don't know how that "breaking change" label got on there...

ThomasBurleson:
Closing as not-needed.

PR #10684, opened 2017-05-23
chore(chips): fix chrome test for selection

DevVersion:
Fixes a test in chrome that failed because the test tried to set a programmatic selection on a number input.

Chrome throws a DOMException if something tries to set a text selection on a number input.


ThomasBurleson:
LGTM

PR #10688, opened 2017-05-25
fix(navbar): remove redundant roles from the list inside the nav element

michaldan100:
Remove redundant roles (listbox and option) from the list inside the nav element. According to https://www.w3.org/TR/wai-aria-practices/#aria_lh_navigation and the navigation landmark example, these roles are not needed.

### DIFF
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -31,7 +31,7 @@ angular.module('material.components.navBar', ['material.core'])
  * Alternatively, the user could simply watch currentNavItem for changes.
  *
  * Accessibility functionality is implemented as a site navigator with a
- * listbox, according to
+ * navigation, according to
  * https://www.w3.org/TR/wai-aria-practices/#Site_Navigator_Tabbed_Style
  *
  * @param {string=} mdSelectedNavItem The name of the current tab; this must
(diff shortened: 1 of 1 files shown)


michaldan100:
I signed it!

2017-05-25 11:41 GMT+03:00 googlebot <notifications@github.com>:

ThomasBurleson:
@jelbourn  - can you quick review this one?

Splaktar:
@michaldan100 Thank you very much for your contribution!

it looks like the CLA issue is caused because you authored the commit using your `ca.com` email address but that address isn't associated to your @michaldan100 GitHub account. Can you please sort that out or re-submit the PR with the commit coming from the email address associated with your GitHub account?

Splaktar:
Thank you for your contribution. Please follow the [commit message guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format).

jelbourn:
Even though the roles are redundant per the spec, it's possible that they may have been added as a workaround for a screen-reader bug. I don't see any harm in leaving them. Thoughts @Splaktar?

Splaktar:
@jelbourn I traced the origin of both of these roles back to the [original PR](https://github.com/angular/material/pull/7781) for navbar. They don't appear to be related to any kind of bug fix or code review feedback.

jelbourn:
I still wouldn't want to remove them without thorough screen-reader testing, though, since the extra roles don't hurt, but they could be helping in a way that was just never recorded.

Splaktar:
OK, we're going to decline this PR at this time since there does not appear to be any negative impact to the current approach and making this change risks a regression.

If you feel strongly about this and would like to perform thorough screen-reader testing on the navbar component and provide us with the results, then please do and we'll re-evaluate this at that time.

PR #10706, opened 2017-05-30
fix(tabs): Accessibility and keyboard interaction fixes

### LINKED ISSUE #10075: md-tab: Dummy tabs not properly hidden from a screen reader [Accessibility]
**Actual Behavior**:
 -  `What is the issue? *`
When using md-tab, the dummy tabs that are visually hidden with the "md-visually-hidden" class are still visible and accessible by a screen reader (JAWS). This results in 2 mirror copies of the list of tabs appearing when using arrow-keys to browse the page.
Dummy tabs, if visually hidden, should be hidden from a screen reader as well using the 'aria-hidden' attribute. Only one set of tabs should be visible by a screen reader when using the arrow-keys to browse.
N/A (issue present in current demos on Angular Material website)
 -  `Details:`
Here is the offending <div> that shows up in the code when the md-tabs are rendered:
<div class="_md-visually-hidden md-dummy-wrapper"> 

This should also have the attribute of 'aria-hidden="true"'. I confirmed that adding this attribute to this particular <div> wrapper fixes the issue.
**Angular Versions**: *
 -  `Angular Version:`
1.5.3
1.1.0-rc2
**Additional Information**:
 -  `Browser Type: *`
Firefox and IE
 -  `Browser Version: *`
Firefox latest version and IE11
 -  `OS: *`
Windows 7/64 bit
 -  `Stack Traces:`
N/A
----

  Add more lines if needed.

jeyoshimi:
This commit does a few things related to accessibility:
1) Hides md-dummy-tabs completely from screen readers with aria-hidden.
With this change, screen readers do not hear the dummy tabs read out
when using linear navigation. Instead of focusing the dummy tabs, the
visible tabs are focused. There was a comment that mentions focusing the
dummy tabs as a way to avoid animation issues; however, I could not
replicate the mentioned issues.
2) Fixes the ARIA roles so users can hear "tab 1 of 3" instead of just
"tab" by moving role="tablist" to the direct parent element (which
contains the items marked role="tab"). (VoiceOver in particular)
3) Fixes keyboard interaction to match recommended ARIA practices by
focusing on the active tab when using TAB navigation, then resolving the
focus/active tab discrepancy on pressing TAB within the tablist. This
prevents the behavior where tabbing away from the tablist then back
results in the previously focused tab getting focus instead of the
active tab.
4) Note that there is a SCSS change added in to retain current behavior
when the active tab is added to the tab order; however, this should
likely be reconsidered as it actively overrides default focus outline
behavior.
5) Removes unsupported ARIA attribute (aria-activedescendant) from
md-tabs-canvas and removes md-tab-id from dummy tabs to prevent
duplicate setting of aria-controls.
6) Moved id to tab item from dummy tab so the aria-labelledby on the tab
content (role="tabpanel") items is set as per ARIA recommendation.

Tests were updated and fleshed out to check that the active tab is the
only tab allowed in the tab order and that focused/active tabs are
reconciled on tabbing out of the tablist.

Closes #10075.

### DIFF
--- a/src/components/tabs/tabs.spec.js
+++ b/src/components/tabs/tabs.spec.js
@@ -129,7 +129,7 @@ describe('<md-tabs>', function () {
 
     }));
 
-    it('should select tab on space or enter', inject(function ($mdConstant) {
+    it('should select tab on space or enter', inject(function ($document, $mdConstant) {
       var tabs     = setup('<md-tabs>' +
                            '<md-tab></md-tab>' +
                            '<md-tab></md-tab>' +
@@ -140,10 +140,15 @@ describe('<md-tabs>', function () {
       triggerKeydown(tabs.find('md-tabs-canvas').eq(0), $mdConstant.KEY_CODE.RIGHT_ARROW);
       triggerKeydown(tabs.find('md-tabs-canvas').eq(0), $mdConstant.KEY_CODE.ENTER);
       expect(tabItems.eq(1)).toBeActiveTab();
+      expect(tabItems.eq(1).attr('tabindex')).toBe('0');
 
       triggerKeydown(tabs.find('md-tabs-canvas').eq(0), $mdConstant.KEY_CODE.LEFT_ARROW);
       triggerKeydown(tabs.find('md-tabs-canvas').eq(0), $mdConstant.KEY_CODE.SPACE);
       expect(tabItems.eq(0)).toBeActiveTab();
+      // Active tab should be added to the tab order as per ARIA practices.
+      expect(tabItems.eq(0).attr('tabindex')).toBe('0');
+      // Deactivated tab should be removed from the tab order.
+      expect(tabItems.eq(1).attr('tabindex')).toBe('-1');
     }));
 
     it('should bind to selected', function () {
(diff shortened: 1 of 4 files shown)


jeyoshimi:
Tagging @jelbourn as requested. :)

jelbourn:
@crisbeto could you also take a quick look at this?

src/components/tabs/tabs.spec.js
```diff
@@ -176,30 +178,45 @@ describe('<md-tabs>', function () {
...
 
       expect(tabItems.eq(0).attr('aria-disabled')).toBe('false');
-      expect(dummyTabs.eq(0).attr('aria-disabled')).toBe('false');
       expect(tabItems.eq(1).attr('aria-disabled')).toBe('true');
-      expect(dummyTabs.eq(1).attr('aria-disabled')).toBe('true');
     });
 
   });
 
+  describe('leaving tablist', function() {
```
crisbeto: nit: I don't think that this needs to be in an extra `describe`.

jeyoshimi: Moved it into the above describe. :)

PR #10708, opened 2017-05-31
Don't assume that the scrollMask is present when closing the md-dialog

janvandenberg:
Fixed an issue where md-dialog contains md-autocomplete elements. These elements remove the scroll mask before the dialog is closed, resulting in an exception when closing the dialog. This a symptom fix. I don't assume that the mask is still there before I try to delete it. But now the dialog closed properly and no errors.


janvandenberg:
I signed it!

ThomasBurleson:
@janvandenberg - lgtm

PR #10726, opened 2017-06-08
feat(compiler): respect preAssignBindingsEnabled state

### LINKED ISSUE #10016: feat(mdCompiler): support `preAssignBindingsEnabled()`

### LINKED ISSUE #10377: Controller with prerendered $mdDialog never called.
**Actual Behavior**: A controller can be assigned to a prerendered dialog but it will never be used.
 -  `What is the issue? *` Buttons calling functions of the controller have no effect.
 -  `What is the expected behavior?` I expected a controller assigned to a prerendered dialog to work the same as with a compiled dialog, or something in the documentation to warn that this is the case.

**CodePen** (or steps to reproduce the issue): * Create a simple dialog that will be compiled with a controller and any number of functions. Assign any one of these functions to a button on the dialog. Then convert the dialog to be an element in the main page (prerendered) and try the button again.
 -  `Details:`

**Angular Versions**: *
 -  `Angular Version:` v1.6.1
 -  `Angular Material Version:` v1.1.3

**Additional Information**:
 -  `Browser Type: *` Google Chrome
 -  `Browser Version: *` 54.0.2840.99
 -  `OS: *` Windows 7 Enterprise
 -  `Stack Traces:`

----

  Add more lines if needed.

mgol:
1. The `$mdCompiler` is now able to respect the `preAssignBindingsEnabled` state
when using AngularJS 1.5.10 or higher; to enable, invoke:

    $mdCompilerProvider.respectPreAssignBindingsEnabled(true);

2. `$mdCompiler` now understands the `$onInit` lifecycle hooks in controllers.
Note that no other AngularJS 1.5+ lifecycle hooks are supported currently.

Invoking:

    $mdCompilerProvider.respectPreAssignBindingsEnabled(true);

will make bindings in Material custom components like `$mdDialog` or `$mdToast`
only available in controller constructors if they are available in constructors
of all other AngularJS controllers. By default this will happen in AngularJS
1.6 or newer. This can also be achieved with AngularJS >=1.5.10 <1.6.0 if:

    $compilerProvider.preAssignBindingsEnabled(false);

is invoked.

Example:

```js
$mdDialog.show({
  locals: {
    myVar: true
  },
  controller: MyController,
  bindToController: true
}

function MyController() {
  // No locals from Angular Material are set yet. e.g myVar is undefined.
}

MyController.prototype.$onInit = function() {
  // Bindings are now set in the $onInit lifecycle hook.
}
```

Huge thanks to Paul Gschwendtner (@DevVersion) for the initial draft version of
this feature.

Fixes #10016
Fixed #10377 
Ref #10469

~~NOTE: I'm marking this PR as WIP as I still want to test it manually with some real code to ensure there's nothing left to do. I also need to think if there's something else I need to test here.~~


src/core/services/compiler/compiler.js
```diff
@@ -81,249 +87,384 @@ angular
...
+   * the `$compileProvider.preAssignBindingsEnabled()` value.
+   *
+   * @deprecated
+   * sinceVersion="1.2.0"
+   * removeVersion="1.3.0"
+   *
+   * This method and the option to assign the bindings before calling the controller's constructor
+   * will be removed in v1.3.0.
+   */
+  var respectPreAssignBindingsEnabled = false;
```
mgol: Add a TODO comment about the default changing to `true` in Material 1.2.

src/core/services/compiler/compiler.js
```diff
@@ -81,249 +87,384 @@ angular
...
+   */
+  this.preAssignBindingsEnabled = function() {
+    if (!respectPreAssignBindingsEnabled) {
+      // respectPreAssignBindingsEnabled === false
+      return true;
+    }
 
-  /** @private @const {!angular.$compile} */
-  this.$compile = $compile;
+    // respectPreAssignBindingsEnabled === true
```
mgol: Add a comment that we're ignoring the AngularJS setting in this case.

src/core/services/compiler/compiler.spec.js
```diff
@@ -159,27 +160,135 @@ describe('$mdCompiler service', function() {
...
-          template: 'hello',
-          controller: function($scope) {
+    });
+
+  });
+
+  [
+      {respectPreAssignBindingsEnabled: true},
+      {respectPreAssignBindingsEnabled: false},
+      {respectPreAssignBindingsEnabled: '"default"', equivalentTo: false}
```
mgol: Add a comment about the Material 1.2 change.

src/core/services/compiler/compiler.spec.js
```diff
@@ -159,27 +160,135 @@ describe('$mdCompiler service', function() {
...
+    var respectPreAssignBindingsEnabled = angular.isDefined(options.equivalentTo) ?
+      options.equivalentTo :
+      realRespectPreAssignBindingsEnabled;
+
+    describe('with respectPreAssignBindingsEnabled set to ' + realRespectPreAssignBindingsEnabled, function() {
+      var preAssignBindingsEnabledInAngularJS = angular.version.minor < 6;
+
+      beforeEach(function() {
+        module(function($mdCompilerProvider) {
+          console.log(angular.version, preAssignBindingsEnabledInAngularJS);
```
mgol: remove it

src/core/services/compiler/compiler.spec.js
```diff
@@ -159,27 +160,135 @@ describe('$mdCompiler service', function() {
...
+
+        return compileData;
+      }
+
+      // Bindings are not preassigned only if we respect the AngularJS config and they're
+      // disabled there. This logic will change in Material 1.2.0.
+      if (respectPreAssignBindingsEnabled && !preAssignBindingsEnabledInAngularJS) {
+        it('$mdCompileProvider.preAssignBindingsEnabled should report bindings are not pre-assigned', function() {
+          module(function($mdCompilerProvider) {
+            expect($mdCompilerProvider.preAssignBindingsEnabled()).toBe(false);
```
mgol: It should be `true`; why did the tests pass? I need to check that.

src/core/services/compiler/compiler.spec.js
```diff
@@ -159,27 +160,135 @@ describe('$mdCompiler service', function() {
...
-      it('should work with bindToController', inject(function($rootScope) {
-        var called = false;
-        var data = compile({
-          template: 'hello',
-          controller: function($scope) {
+    });
+
+  });
+
+  [
```
mgol: Add a test asserting that if `bindToController` is set to `false` the `$onInit` method is called anyway.

src/core/services/compiler/compiler.js
```diff
@@ -81,249 +87,384 @@ angular
...
+   *
+   * The default value is `false` but will change to `true` in AngularJS Material 1.2.
+   *
+   * It is recommended to set this flag to `true` in AngularJS Material 1.1.x; the only reason it's not set that way
+   * by default is backwards compatibility. Not setting the flag to `true` when
+   * `$compileProvider.preAssignBindingsEnabled()` is set to `false` (i.e. default behavior in AngularJS 1.6.0 or newer)
+   * makes it hard to unit test Material Dialog/Toast controllers using the `$controller` helper as it always follows
+   * the `$compileProvider.preAssignBindingsEnabled()` value.
+   *
+   * @deprecated
```
mgol: Remove the deprecated note (here & in other places if present).

src/core/services/compiler/compiler.js
```diff
@@ -81,249 +87,384 @@ angular
...
+   * Material & user-created components; their behavior can be checked via `$compileProvider.preAssignBindingsEnabled()`
+   * in AngularJS `>=1.5.10 <1.7.0`.
+   *
+   * @deprecated
+   * sinceVersion="1.2.0"
+   * removeVersion="1.3.0"
+   *
+   * This method will be removed in v1.3.0.
+   */
+  this.preAssignBindingsEnabled = function() {
```
mgol: Maybe we don't need this method to be public until someone needs it?

mgol: Also, maybe change the name to indicate it's a getter? Like `getPreAssignBindingsEnabled`.

src/core/services/compiler/compiler.js
```diff
@@ -81,249 +87,384 @@ angular
...
+   * @description
+   * Call this method to enable/disable whether Material-specific (dialogs/toasts) controllers respect the AngularJS
+   * `$compile.preAssignBindingsEnabled` flag. Note that this doesn't affect directives/components created via
+   * regular AngularJS methods which constitute most Material & user-created components.
+   *
+   * @see [AngularJS documentation for `$compile.preAssignBindingsEnabled`
+   * ](https://code.angularjs.org/1.6.4/docs/api/ng/provider/$compileProvider#preAssignBindingsEnabled)
+   * for more information.
+   *
+   * If disabled (false), the compiler assigns the value of each of the bindings to the
```
ThomasBurleson: @gkalpak - @mgol  and I reviewed this PR yesterday... he is making slight improvements within the next 2 days. Standy by...

gkalpak: 😞 OK 😞 
But I'll be back 😈

ThomasBurleson: Of course. I expect nothing less.

src/components/dialog/dialog.js
```diff
@@ -631,30 +631,40 @@ function MdDialogProvider($$interimElementProvider) {
...
 
+  /**
+   * Controller for the md-dialog interim elements
+   * @ngInject
+   */
+  function MdDialogController($mdDialog, $mdConstant) {
+    // For compatibility with AngularJS 1.6+, we should always use the $onInit hook in
+    // interimElements. The $mdCompiler simulates the $onInit hook for all versions.
+    this.$onInit = function() {
+      var isPrompt = this.$type == 'prompt';
```
EladBezalel: Please use ===

src/components/dialog/dialog.js
```diff
@@ -631,30 +631,40 @@ function MdDialogProvider($$interimElementProvider) {
...
+
+      if (isPrompt && this.initialValue) {
+        this.result = this.initialValue;
+      }
+
+      this.hide = function() {
+        $mdDialog.hide(isPrompt ? this.result : true);
```
EladBezalel: i think hide returns a promise so let's return it so it can be chainable

src/components/dialog/dialog.js
```diff
@@ -631,30 +631,40 @@ function MdDialogProvider($$interimElementProvider) {
...
+      };
+      this.abort = function() {
+        $mdDialog.cancel();
```
EladBezalel: return here as well

src/components/dialog/dialog.js
```diff
@@ -631,30 +631,40 @@ function MdDialogProvider($$interimElementProvider) {
...
+      };
+      this.keypress = function($event) {
+        if ($event.keyCode === $mdConstant.KEY_CODE.ENTER) {
+          $mdDialog.hide(this.result);
```
EladBezalel: return here as well 

and outside the if
```
return $q.resolve();
```

src/components/toast/toast.js
```diff
@@ -354,6 +337,33 @@ function MdToastProvider($$interimElementProvider) {
...
+          self.highlightClass
+        ]
+      }
+
+      $scope.$watch(function() { return activeToastContent; }, function() {
+        self.content = activeToastContent;
+      });
+
+      this.resolve = function() {
+        $mdToast.hide( ACTION_RESOLVE );
```
EladBezalel: return here as well

src/core/services/compiler/compiler.spec.js
```diff
@@ -159,27 +160,135 @@ describe('$mdCompiler service', function() {
...
+      realRespectPreAssignBindingsEnabled;
+
+    describe('with respectPreAssignBindingsEnabled set to ' + realRespectPreAssignBindingsEnabled, function() {
+      var preAssignBindingsEnabledInAngularJS = angular.version.minor < 6;
+
+      beforeEach(function() {
+        module(function($mdCompilerProvider) {
+          console.log(angular.version, preAssignBindingsEnabledInAngularJS);
+          // Don't set the value so that the default state can be tested.
+          if (realRespectPreAssignBindingsEnabled === true || realRespectPreAssignBindingsEnabled === false) {
```
EladBezalel: cant we do `angular.isDefined(realRespectPreAssignBindingsEnabled)` instead?

src/components/dialog/dialog.js
mgol: This is exactly the same code that was here before. I'd rather not change unrelated things in this PR.

The same applies to your other comments.

If you ignore whitespace-only changes (via appending `?w=1` to the URL: https://github.com/angular/material/pull/10726/files?w=1) you will only see what really changed in the logic.

src/core/services/compiler/compiler.spec.js
mgol: Sure

src/components/dialog/dialog.js
EladBezalel: Yeah i understand.. thanks!

src/core/services/compiler/compiler.spec.js
mgol: I wonder why I didn't think about that having used exactly this method in source. :)

EladBezalel: That's why it stroke me

gkalpak: Because you want to differentiate `true`/`false` from `"default"`. `angular.isDefined()` won't work here.

src/core/services/compiler/compiler.spec.js
```diff
@@ -159,27 +160,135 @@ describe('$mdCompiler service', function() {
...
+            locals: {name: 'Bob'}
+          });
+
+          expect(TestController.prototype.$onInit).toHaveBeenCalledTimes(1);
+          expect(isInstantiated).toBe(true);
+        });
+      } else {
+        it('$mdCompileProvider.preAssignBindingsEnabled should report bindings are pre-assigned', function() {
+          module(function($mdCompilerProvider) {
+            expect($mdCompilerProvider.preAssignBindingsEnabled()).toBe(false);
```
mgol: It should be `true`; why did the tests pass? I need to check that.

src/core/services/compiler/compiler.spec.js
mgol: It shouldn't; I commented on a wrong line; fixed.

Nice catch!

src/core/services/compiler/compiler.spec.js
mgol: Ah, that's why I didn't think about that. :D

EladBezalel: What about
```
typeof(realRespectPreAssignBindingsEnabled) === 'boolean'
```

gkalpak: That would work. As would `realRespectPreAssignBindingsEnabled !== 'default'`.

src/core/services/compiler/compiler.js
mgol: @gkalpak What's wrong here? I think current text is OK and your proposed change is wrong...

mgol:
@ThomasBurleson @EladBezalel @gkalpak PR updated, comments addressed.

mgol:
I've tested this with one of my company's internal apps and at first sight everything seemed to work with `$mdCompilerProvider#respectPreAssignBindingsEnabled(true)` (including tests with some minor modifications).

I've also run tests on default settings and everything seemed to work without any modifications.

There's one last thing to resolve - @jbedard & @gkalpak mentioned that exposing `$mdCompilerProvider.preAssignBindingsEnabled()` to the public is not necessary, it may be an internal function. Let me know what you think, @ThomasBurleson; I can apply such a change. The rest should be ready.

src/components/dialog/dialog.js
ThomasBurleson:
```js
this.hide = function() {
 return $mdDialog.hide(isPrompt ? this.result : true);
}
```

src/components/toast/toast.js
ThomasBurleson:
```js
this.resolve = function() {
   return $mdToast.hide( ACTION_RESOLVE );
}
```

src/core/services/compiler/compiler.js
gkalpak: Yes, it seems correct now. Not sure what happened :grin:

jelbourn:
Looks like this succeeds on google presubmit

PR #10735, opened 2017-06-13
fix(dialog): generate `aria-label` with dialog title (if exists) when `.ariaLabel()` is not specified

### LINKED ISSUE #10582: mdDialog generates unhelpful aria-label
**Actual Behavior**:
 -  `What is the issue? *`

By default, `$mdDialog` generates an `aria-label` for confirm/alert dialogs by shortening the title. For JAWS at least, this is counterproductive because JAWS reads something like "This is an dot dot dot this is an alert title" instead of just "This is an alert title".

There's also no way to suppress this; if I do `.ariaLabel('')`, it still generates the label.

If I don't specify an aria-label, I just want the screen reader to read the contents of the dialog.

http://codepen.io/anon/pen/jmNZpG

alopezsanchez:
Fixes #10582

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -1083,9 +1083,15 @@ function MdDialogProvider($$interimElementProvider) {
       }
       else {
         $mdAria.expectAsync(element, 'aria-label', function() {
-          var words = dialogContent.text().split(/\s+/);
-          if (words.length > 3) words = words.slice(0, 3).concat('...');
-          return words.join(' ');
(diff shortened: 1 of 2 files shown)


ThomasBurleson:
@alopezsanchez - very nice.

alopezsanchez:
Hello @ThomasBurleson. What `needs: presubmit` label means?

ThomasBurleson:
@alopezsanchez - internally Google team processes where we perform strict testing and validate the PR.

Issue #10668, opened 2017-05-16
requestAnimationFrame still run when set md-progress-circular ng-show='false'


feichao:
**Actual Behavior**:
 -  `requestAnimationFrame is still running when set md-progress-circular ng-show='false', and the usage of cpu retains 19% `
 -  `when set ng-show='false' or ng-hide='true', the requestAnimationFrame should pause`

**AngularJS Versions**: *
 -  `AngularJS Version: 1.5.8`
 -  `AngularJS Material Version: v1.1.1`

**Additional Information**:
 -  `Browser Type: chrome`
 -  `Browser Version: 58`
 -  `OS: Mac OS`

ThomasBurleson:
We will not fix this in AngularJS Material. You are welcome to provide a Pull Request to resolve this for Material 1.1.x.

Splaktar:
It would be helpful if you could provide a CodePen that demonstrates this. Here is a starter http://codepen.io/team/AngularMaterial/pen/RrbXyW

Splaktar:
No demo provided. If you are still seeing this, please submit a new issue using the issue template and providing a CodePen reproduction.

PR #10745, opened 2017-06-16
fix(progressCircular): requestAnimationFrame is still running when us…

feichao:
fix(progressCircular): requestAnimationFrame is still running when using ng-show or ng-hide [#10668](https://github.com/angular/material/issues/10668)

### DIFF
--- a/src/components/progressCircular/js/progressCircularDirective.js
+++ b/src/components/progressCircular/js/progressCircularDirective.js
@@ -211,6 +211,23 @@ function MdProgressCircularDirective($window, $mdProgressCircular, $mdTheming,
 
     });
 
+    
+    // When using ng-show or ng-hide, 
+    // the animation won't stop because requestAnimationFrame is still running.
+    // Do this to stop requestAnimationFrame can save some usage of system resources.
+    scope.$watch(function() {
+      var style = $window.getComputedStyle(node);
+      return style && style.display;
+    }, function(newValue, oldValue) {
+      if(newValue && newValue !== oldValue) {
+        if(newValue === 'none') {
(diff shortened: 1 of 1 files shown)


feichao:
I signed it!

ThomasBurleson:
@feichao - lgtm, thx.

clshortfuse:
Looks rather expensive to evaluate DOM attributes on every Angular tick. MutationObserver would give better performance. 

Also, using `ng-if` instead of `ng-show` would solve the underlying issue, so a recommendation could be added in the documentation. Another alternative would be to only use the watcher if `ng-show` or `ng-hide` are being used.

Splaktar:
@clshortfuse any ideas on how to use the watcher only when `ng-show` or `ng-hide` are being used?

@feichao can you please add some tests for this?
For example: Verify that the `md-mode-indeterminate` is applied after the element is displayed via toggling `ng-show` and verify that the `md-mode-indeterminate` is removed after the element is hidden via toggling `ng-hide`.

clshortfuse:
@Splaktar Sure, you can use a `MutationObserver` to watch for the `ng-show` or `ng-hide` class. If the value exists then a watcher should be created to handle it's computed value.

You can see a similar implementation created with `md-visible` with `md-tooltip` in https://github.com/angular/material/pull/7822

Splaktar:
@feichao can you please update your implementation to only apply the watcher when `ng-show` or `ng-hide` are present as described in https://github.com/angular/material/pull/10745#issuecomment-345852087?

Since this whole PR is about performance, it would be best to use this more optimally performing approach.

Issue #10751, opened 2017-06-20
tabs: input inside md-tab-label does not handle spaces well


barzik:
**Actual Behavior**:
 -  Insert input to md-tab-label repository
 -  You cannot insert spaces to the input.
 - It is happening because of [line 396 in tabsController.js](https://github.com/angular/material/blob/master/src/components/tabs/js/tabsController.js#L306)
Here is a [pull request](https://github.com/angular/material/pull/10752), tell me if it is )the desired path:

**Steps to reproduce the issue):**
 -  https://codepen.io/barzik/pen/RgpvRP
 -  `Details:` Just try to insert space to the input. It is important to me because the my usecase is allowing some users to edit the code.

**AngularJS Versions**: *
 -  `AngularJS Version: 1.6`
 -  `AngularJS Material Version: 1.1.4`

**Additional Information**:
 -  `Browser Type: All`
 -  `Browser Version: *`
 -  `OS: *`
 -  `Stack Traces:`

----
Shortcut to create a [new CodePen Demo](http://codepen.io/team/AngularMaterial/pen/bEGJdd).
Note: `*` indicates required information. Without this information, your issue may be auto-closed.

> Do not modify the titles or questions. Simply add your responses to the ends of the questions.
  Add more lines if needed.

andrewseguin:
An API change seems unnecessary for this since you can just prevent the input's event from propagating upward to the header.

alimcharaniya:
Did you find a solution to enter spaces here? I have the same issue.

Splaktar:
Re-opening this until we can provide a CodePen that demonstrates @andrewseguin's guidance in https://github.com/angular/material/issues/10751#issuecomment-351131181.

Splaktar:
@alimcharaniya OK, here's the solution via [CodePen](https://codepen.io/Splaktar/pen/gzOObw?editors=1010#0). As hinted at by @andrewseguin, you can catch the keydown event on the input and avoid it getting propagated up to `md-tabs` via `event.stopPropagation()`.

PR #10752, opened 2017-06-20
fix(md-tab): Adding no keypress to allow input inside md-tab 

barzik:
## Current situation
Keypress preventdefault is causing issues with some use cases, like input inside `md-tabs`. See issue [10751](https://github.com/angular/material/issues/10751)

## Before
`keydown` was activated all times.

## AFTER
`md-no-keypress` allow users to choose if to use keypress in `md-tabs` default is false.

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -67,6 +67,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
     defineBooleanAttribute('noDisconnect');
     defineBooleanAttribute('autoselect');
     defineBooleanAttribute('noSelectClick');
+    defineBooleanAttribute('noKeypress');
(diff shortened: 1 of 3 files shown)


barzik:
Adding my other mail to the CLA

ThomasBurleson:
lgtm

src/components/tabs/js/tabsController.js
```diff
@@ -300,10 +301,11 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
   // Event handlers / actions
 
   /**
-   * Handle user keyboard interactions
+   * Handle user keyboard interactions if `md-no-keypress` are false.
```
Splaktar: Shouldn't it be "is false"?

andrewseguin:
This does not seem like the right approach - an additional API to prevent keyboard events from being handled is not necessary. If you do not want the header to handle these events, just stop the event from being propagated to the header.

PR #10759, opened 2017-06-21
fix/refactor(autocomplete): use real <label> instead of placeholder

jeyoshimi:
This affects the autocomplete version with "placeholder" specified
(not the md-floating-label version). This means that it also affects the chips component.

BREAKING CHANGE (STYLE): The text is no longer a placeholder, so styling that
previously used the webkit/moz placeholder attributes will no longer
apply. Now, styles should be applied to ".md-autocomplete-placeholder". This is actually slightly cleaner for cross-browser style consistency.

Before this change, placeholder and aria-label are identical, resulting
in some screenreaders (like JAWS) reading what seems like duplicate
text. After this change, a is used instead of both the
aria-label and the placeholder, resulting in only a single reading of
the single specified label.

Also change the roles of the autocomplete list to be more compliant with
ARIA practices by adding role="listbox" and role="option". Remove
aria-haspopup="true" because role="combobox" elements have aria-haspopup
set to true by default.

CSS positioning is used to ensure that the layering order is as follows
(from bottom to top):
autocomplete container
placeholder
text input field
This way, the placeholder text is on top of the white background, but
behind the transparent text input, allowing us to see it, yet still be
able to click on the text field.

Some style changes were combined/duplicates were removed from the chips.scss.
From the history/blame, it didn't seem like intentional cascading, but I couldn't tell,
so please do sanity check me.

### DIFF
--- a/src/components/autocomplete/autocomplete.scss
+++ b/src/components/autocomplete/autocomplete.scss
@@ -110,9 +110,18 @@ md-autocomplete {
     padding: 0 15px;
     line-height: 40px;
     height: 40px;
+    position: absolute;
+  }
+  // Dimensions of md-autocomplete-placeholder should match md-autocomplete.
+  .md-autocomplete-placeholder {
+    color: #757575;
+    font-size: 14px;
+    line-height: 40px;
+    padding: 0 15px;
   }
   .md-show-clear-button button {
-    position: relative;
+    position: absolute;
+    right: 0;
     line-height: 20px;
     text-align: center;
     width: $md-autocomplete-clear-size;
@@ -123,7 +132,7 @@ md-autocomplete {
     padding: 0;
     font-size: 12px;
     background: transparent;
-    margin: auto 5px;
+    margin: 5px;
     &:after {
       content: '';
       position: absolute;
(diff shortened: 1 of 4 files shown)


jeyoshimi:
@jelbourn I know this library is largely in maintenance mode, so I'm a bit hesitant about the change, but I'll leave it to your discretion! o7

src/components/autocomplete/autocomplete.scss
```diff
@@ -110,9 +110,18 @@ md-autocomplete {
     padding: 0 15px;
     line-height: 40px;
     height: 40px;
+    position: absolute;
+  }
+  // Dimensions of md-autocomplete-placeholder should match md-autocomplete.
+  .md-autocomplete-placeholder {
+    color: #757575;
```
jelbourn: The color for autocomplete needs to be defined as part of the theme, otherwise this will be off for dark themes.

jeyoshimi:
Closing pull request; this change breaks a bunch of projects and will not be submitted in order to preserve compatibility for angularjs (as it is technically in maintenance mode)

Issue #9908, opened 2016-10-24
select: improves with a11y


ThomasBurleson:
-  The aria-label should not include the selected value.
-  The element for the placeholder value should have role=”option”, aria-posinset=”(relative_position_in_list)”, and aria-sizeof=”(number_of_items)”. The latter two ARIA attributes are to trick the screen reader into thinking that there is more than one option.
-  Drop the aria-owns.
-  Hide the place holder when the option list is shown.
-  Put role=”presentation” on the option list container.

EladBezalel:
regarding `aria-posinset="(relative_position_in_list)"`- what about multi select?

npbenjohnson:
This is also a problem in relation to #10737 because screen readers ignore aria-label if aria-labelledby is present. I have a different solution from the suggestion working locally. I just pointed aria-describedby at the summary container and added a hidden "Selected Value(s): " span. This makes md-select-text and md-select-html the information for screen readers too, which I would debate is probably the correct behavior.

Splaktar:
Closing as many of these were fixed in 1.1.5 (https://github.com/angular/material/pull/10760) and others have open issues that specifically reference just one of these bullets.

PR #10760, opened 2017-06-22
fix(select): A11y fixes allowing screen readers (VoiceOver) to read list item position

jeyoshimi:
1. Add role="presentation" to the container divs so assistive technology
(primarily VoiceOver) to accurately count and announce "item 3 of 7".
https://www.w3.org/TR/wai-aria/roles#presentation

2. Only add aria-owns when the md-select element does not contain its
menu element in the DOM. This means when the select is opened, as
opening the select means the element, previously nestled within
md-select, gets appended to the end of the DOM.

3. Dynamically hide md-select-value with aria-hidden to get an accurate
item count for the md-select. This can't just be hidden all the time
because VoiceOver does not allow linear navigation into selects with no
visible options.

4. Move autoFocus call to after the select menu element is fully rendered
in order to properly focus the first item in the select.

Addresses some points in #9908.

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -202,9 +202,11 @@ function SelectDirective($mdSelect, $mdUtil, $mdConstant, $mdTheming, $mdAria, $
     }
 
     // There's got to be an md-content inside. If there's not one, let's add it.
-    if (!element.find('md-content').length) {
+    var mdContentEl = element.find('md-content');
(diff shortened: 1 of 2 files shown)


jeyoshimi:
@jelbourn A couple more accessibility fixes for the md-select :)

ThomasBurleson:
lgtm

PR #10786, opened 2017-07-05
fix(tabs): allow right and left arrows to cycle through tabs

kbae00:
As per the ARIA spec on tabs
(https://www.w3.org/TR/wai-aria-practices-1.1/#tabpanel, see "Keyboard
Interaction" section), LEFT ARROW should move the focus to the last tab
if it's currently on the first tab. RIGHT ARROW should move the focus to
the first tab if it's currently on the last tab. Previously, LEFT ARROW
stopped moving focus once it hit the first tab and RIGHT ARROW stopped
moving focus once it hit the last tab.

### DIFF
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -684,6 +684,9 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
     for (newIndex = index + inc;
          ctrl.tabs[ newIndex ] && ctrl.tabs[ newIndex ].scope.disabled;
          newIndex += inc) {}
+
+    newIndex = (index + inc + ctrl.tabs.length) % ctrl.tabs.length;
+
     if (ctrl.tabs[ newIndex ]) {
       ctrl[ key ] = newIndex;
     }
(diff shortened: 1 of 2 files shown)


kbae00:
@jelbourn

src/components/tabs/js/tabsController.js
```diff
@@ -684,6 +684,17 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+
+    // Handle cycling through the tabs
+    var lastIndex = ctrl.tabs.length - 1;
+    // If the left arrow is pressed from the first tab
+    if (inc === -1 && index === 0) {
+      newIndex = lastIndex;
+    // If the right arrow is pressed from the last tab
+    } else if (inc === 1 && index === lastIndex) {
+      newIndex = 0;
+    }
```
jelbourn: This can be shortened to
```js
newIndex = (index + inc + ctrl.tabs.length) % ctrl.tabs.length;
```

src/components/tabs/js/tabsController.js
```diff
@@ -684,6 +684,17 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
...
+
```
jelbourn: You removes the unit test checks for no-op on the ends, but didn't add new checks for the wrapping behavior.

src/components/tabs/js/tabsController.js
kbae00: Done.

src/components/tabs/js/tabsController.js
kbae00: Done.

Issue #10784, opened 2017-07-05
md-datepicker sets the date to day before if daylight savings time is on


relair:
**Actual Behavior**:
 -  `What is the issue? *` When having a model data as new Date("2001-01-01T00:00:00Z") and changing the date to 10th of Jun 2001 the myDate.toGMTString() is  Sat, 09 Jun 2001 23:00:00 GMT- the day before. 
 -  `What is the expected behavior?` Expected string would be  Sun, 10 Jun 2001 00:00:00 GMT

**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:` https://codepen.io/anon/pen/yXEQZm
 -  `Details:` It happens with daylight savings time enabled:
My windows culture is (UTC+00:00) Edinburgh with Adjust for daylight savings time on (works fine without)
It seems the the model date has some impact as well as I saw it not always happen, but it does happen every time for new Date("2001-01-01T00:00:00Z")

**AngularJS Versions**: *
 -  `AngularJS Version:` 1.6.4
 -  `AngularJS Material Version:` v1.1.3

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *` Version 59.0.3071.115 (Official Build) (64-bit)
 -  `OS: *` Windows 10 Home
 -  `Stack Traces:`
![datepickerbug](https://user-images.githubusercontent.com/13729116/27866820-b09b4a1c-618f-11e7-81b1-2a4ec27b982d.png)
----
Shortcut to create a [new CodePen Demo](http://codepen.io/team/AngularMaterial/pen/bEGJdd).
Note: `*` indicates required information. Without this information, your issue may be auto-closed.

> Do not modify the titles or questions. Simply add your responses to the ends of the questions.
  Add more lines if needed.

relair:
After some debugging I have found some methods that change the date to midnight, that seem quite redundant because date gets changed to a string for validation by angular input[type="date"] validation and it is angular what parses the string back to date.

The way I can easily fix this is by adding ng-model-options="{timezone: 'UTC'}" to the date picker.

I think that md-datepicker should work with UTC by default because usually the time is irrelevant when picking date and it leads to lots of confusion when you select a date, send it to api only to date be a different day because of timezone offset.

relair:
[...]

PR #10788, opened 2017-07-06
fix(dateUtil): change date functions to use UTC as base for date calculations

relair:
Changing date util functions to use UTC as base for date calculation to enable proper support of `ng-model-options="{timezone: 'UTC'}"` and different time zones.

Together with setting `ng-model-options="{timezone: 'UTC'}"` it fixes the issues with date time zones, without it it should still work as intended for purposes of passing date string for validation by angular.

This pull request is a follow up of https://github.com/angular/material/issues/10784.

### DIFF
--- a/src/components/datepicker/js/dateUtil.js
+++ b/src/components/datepicker/js/dateUtil.js
@@ -38,7 +38,7 @@
      * @returns {Date}
      */
     function getFirstDateOfMonth(date) {
-      return new Date(date.getFullYear(), date.getMonth(), 1);
+      return new Date(Date.UTC(date.getFullYear(), date.getMonth(), 1));
     }
 
     /**
@@ -47,7 +47,7 @@
      * @returns {number}
      */
     function getNumberOfDaysInMonth(date) {
-      return new Date(date.getFullYear(), date.getMonth() + 1, 0).getDate();
+      return new Date(Date.UTC(date.getFullYear(), date.getMonth() + 1, 0)).getDate();
     }
 
     /**
@@ -56,7 +56,7 @@
      * @returns {Date}
      */
     function getDateInNextMonth(date) {
-      return new Date(date.getFullYear(), date.getMonth() + 1, 1);
+      return new Date(Date.UTC(date.getFullYear(), date.getMonth() + 1, 1));
     }
 
     /**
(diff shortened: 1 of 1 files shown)


relair:
We've signed it.

SocVi100:
Could any collaborator merge this pull request, please?

relair:
We've signed it. I guess corporate takes longer to process, but agreement is listed for my account now.

Splaktar:
Thank you for your first contribution to AngularJS Material! 

Splaktar:
It's not clear to me that this is the right approach. Will this work for a datepicker with `ng-model-options="{timezone: '-0500'}"` set?

Would something closer to https://github.com/angular/material/pull/9410 for md-calendar be more appropriate (minus some of the integration with the `input` element)?

jelbourn:
This change seems risky to me; the datepicker was always intended to work in local time, so changing it now may have unexpected consequences

relair:
The date selected internally by date picker is getting changed to a string for validation by angular input[type="date"] so the hour or timezone doesn't matter at that point. 
Problem was because of DST the internal date and the resulting string were wrong. By using UTC internally it makes the string value consistent with the selected date. 
The logic for ng-model-options is pretty much done by input[type="date"] later on and this change shouldn't make difference for it (just for initial date string creation).
It would be good to have unit test replicating this bug, but unfortunately it requires system time to be set to DST time zone, so it will depend on the settings of the machine it is being run on.
I know it may seem risky, but without it we have a bug for sure, and all you need to do is to debug it and see how it works to know if this change makes sense or not.

Splaktar:
We've made changes to support `ng-model-options="{timezone: 'UTC'}"` in the datepicker and calendar (see https://github.com/angular/material/commit/e24d09c68ce255ed82cac68361c8ee3dfeeb56e4 and https://github.com/angular/material/commit/2a01746c4fbea1bbd2ca82808a6e9f0f979a2968) in `1.1.24`.

In that work, we continued to stick with local time as the internal time structure, to reduce risk, but it does require us to do some extra conversions. We don't plan to re-write things to make UTC be the internal time.

Thank you for the input on this and helping to move this feature forward!

Issue #10595, opened 2017-04-18
md-menu-bar not working properly with  ng-repeat 


Ashish142:
**Actual Behavior**:
 -  `What is the issue? *`
 -  `What is the expected behavior?`

**CodePen** (or steps to reproduce the issue): *
 -  `CodePen Demo which shows your issue:`
 -  `Details:`

**AngularJS Versions**: 
 -  `AngularJS Version:`1.5.3
 -  `AngularJS Material Version:`1.1.1

**Additional Information**:
 -  `Browser Type: *`
 -  `Browser Version: *`
 -  `OS: *`
 -  `Stack Traces:`

----
Shortcut to create a [new CodePen Demo](http://codepen.io/team/AngularMaterial/pen/bEGJdd).
Note: `*` indicates required information. Without this information, your issue may be auto-closed.

> Do not modify the titles or questions. Simply add your responses to the ends of the questions.
  Add more lines if needed.

Ashish142:
i am using md-menu-bar for display categories , and in the category there are four levels 
by clicking on main first level will come, by clicking on any one in the response of first level second level will come , this is going up to four levels , i have checked md-menu-bar with static data up to four levels then its working fine, but as i am using it with dynamic data its not working properly , i am able to click on main categories only for first time , and if click on second time on main categories then it will disappears.
please help me to resolve 

thanks

Ashish142:
@gmoothart @mprobst @wesalvaro @mprobst

wesalvaro:
Please provide a demo when filing an issue like this.

Splaktar:
Thank you for taking the time to submit this issue. However it does not follow our required issue template. This gives the team the information needed to investigate the bug or prioritize the feature.

Specifically, the CodePen reproduction is required. It was requested again on June 22nd, but there has been no movement here. Closing it.

PR #10802, opened 2017-07-18
fix(menu): fix dynamically created nested menu

kerzsi:
… (#10595)

When create cascade/nested menu dynamically, and switch between menus, it closes every submenu expect the root.
This happens because from angular 1.6 it is necessary to bind properties at init.

### DIFF
--- a/src/components/menu/js/menuController.js
+++ b/src/components/menu/js/menuController.js
@@ -14,13 +14,12 @@ function MenuController($mdMenu, $attrs, $element, $scope, $mdUtil, $timeout, $r
   var self = this;
   var triggerElement;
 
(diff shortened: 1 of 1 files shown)


kerzsi:
I signed it!

kerzsi:
[clabot:check]

igyorgyi:
👍

Splaktar:
@kerzsi or @Ashish142 can either of you provide a CodePen that demonstrates the issue that this PR is trying to solve?

Splaktar:
I put together https://github.com/angular/material/pull/11094 to try and reproduce this and I was not able to reproduce the error of see an improvement due to the change made in this PR. I tested with both AngularJS 1.5.5 and 1.6.7.

PR #10807, opened 2017-07-19
	fix(dialog):prevent dialogs from hiding live regions

### LINKED ISSUE #10804: Accessibility: Autocomplete results are not announced when autocomplete is inside a dialog
**Actual Behavior**:
 -  `What is the issue? *` When navigating an autocomplete list within an dialog, list values are not announced by screen reader.
 -  `What is the expected behavior?` As the user scrolls through items in the list, the values are read aloud.  

> “If an item is focused within the list, this should be announced next, followed by an indication of the item's position with the list if the screen reader supports this. As focus moves within the list, the screen reader should announce the relevant list items.” - [MDN Using the Listbox Role](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_listbox_role)

 -  `Steps to Reproduce`
    +  Activate your screen reader
    +  [Navigate to the Angular Material demo Page for mdAutocomplete](https://material.angularjs.org/latest/demo/autocomplete)
    + Open the example labelled "Inside Dialog"
    + Use your keyboard's directional keys to navigate through the items, notice that the values are not announced aloud
    + Repeat the process for any of the other autocomplete examples, notice the values *are* announced aloud.

 -  `Details:`

When a dialog is created that contains an autocomplete, the autocomplete first appends a live region [(div[role=status][aria-live=polite])](https://github.com/angular/material/blob/master/src/core/services/liveAnnouncer/live-announcer.js#L77) to the body of the document. [The dialog then hides everything in the DOM that is not an ancestor or child of the dialog](https://github.com/angular/material/blob/master/src/components/dialog/dialog.js#L1152), which unfortunately includes the new live region the new autocomplete just appended to the body. 

Thus, autocompletes within dialogs have live regions (`$mdLiveAnnouncer._liveElement`) with aria-hidden=true and that is why the autocomplete values are not read aloud.

**AngularJS Versions**: *
 -  `AngularJS Version:` 1.5.5

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *` Version 59.0.3071.115 (Official Build) (64-bit)
 -  `OS: *` Mac OS X 10.12.5
 -  `Stack Traces:`

----

  Add more lines if needed.

@jelbourn

0ddfell0w:
Edit:
change how dialogs hide background elements to avoid hiding live regions, ensuring that an autocomplete inside a dialog will not have its live region aria-hidden

Fixes #10804

***********
Out of date:
add moveLiveRegion fn that moves an autocomplete's live region off document.body
and makes it a child of md-autocomplete. this prevents dialogs from
mistakenly hiding an autocomplete's live region

Fixes #10804

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -1142,7 +1142,9 @@ function MdDialogProvider($$interimElementProvider) {
           for (var i = 0; i < children.length; i++) {
             // skip over child if it is an ascendant of the dialog
             // or a script or style tag
-            if (element !== children[i] && !isNodeOneOf(children[i], ['SCRIPT', 'STYLE'])) {
+            if (element !== children[i] &&
+             !isNodeOneOf(children[i], ['SCRIPT', 'STYLE']) &&
+             !children[i].hasAttribute('aria-live')) {
               children[i].setAttribute('aria-hidden', isHidden);
             }
           }
--- a/src/components/dialog/dialog.spec.js
+++ b/src/components/dialog/dialog.spec.js
@@ -1708,6 +1708,23 @@ describe('$mdDialog', function() {
       expect(sibling.attr('aria-hidden')).toBe('true');
     }));
 
+    it('should not apply aria-hidden to live region siblings', inject(function($mdDialog, $rootScope, $timeout) {
+
+      var template = '<md-dialog aria-label="Some Other Thing">Hello</md-dialog>';
+      var parent = angular.element('<div>');
+      parent.append('<div aria-live="polite"></div>')
+
+      $mdDialog.show({
+        template: template,
+        parent: parent
+      });
+
+      runAnimation();
+
+      var liveRegion = angular.element(parent[0].querySelector('[aria-live]'));
+      expect(liveRegion.attr('aria-hidden')).toBe(undefined);
+    }));
+
     it('should trap focus inside of the dialog', function() {
       var template = '<md-dialog>Hello <input></md-dialog>';
       var parent = document.createElement('div');


0ddfell0w:
@jelbourn

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -196,6 +197,16 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
   }
 
   /**
+   * Moves the live region within the autocomplete to ensure the live region is
+   * is not hidden from assistive technologies when the autocomplete appears
+   * within a dialog.
+   */
+  function moveLiveRegion () {
+    angular.element($mdLiveAnnouncer._liveElement).detach();
+    elements.$.main.append($mdLiveAnnouncer._liveElement)
```
jelbourn: The `$mdLiveAnnouncer` is a global singleton; it does not belong to the autocomplete. With this, each autocomplete will "steal" the aria-live element when it's created. Said element will also be destroyed if whichever autocomplete stole it most recently is destroyed, breaking all subsequent announcing. 

The easiest way to fix this issue is probably to have the dialog specifically ignore the aria-live element when marking everything as `aria-hidden`.

0ddfell0w:
@jelbourn thanks for explaining, didn't know services were singletons. took your suggestion, which works as well.  🙏 👀

src/components/dialog/dialog.js
```diff
@@ -1142,7 +1142,9 @@ function MdDialogProvider($$interimElementProvider) {
           for (var i = 0; i < children.length; i++) {
             // skip over child if it is an ascendant of the dialog
             // or a script or style tag
-            if (element !== children[i] && !isNodeOneOf(children[i], ['SCRIPT', 'STYLE'])) {
+            if (element !== children[i] &&
+             !isNodeOneOf(children[i], ['SCRIPT', 'STYLE']) &&
+             !children[i].hasAttribute('aria-live')) {
```
jelbourn: Can you also add a unit test for this change?

Here's an existing test for this behavior: https://github.com/angular/material/blob/e96293a7010e63e381336f4e0928057eb7880bd1/src/components/dialog/dialog.spec.js#L1690

You should be able to follow that to create a new test

0ddfell0w: 👍

PR #10818, opened 2017-07-27
fix(theming): don't assume selector corresponds to expression

### LINKED ISSUE #10793: colorType buckets of style rules are confused, causing Angular expressions in the generated CSS
**Actual Behavior**:
`What is the issue? *`
Generated CSS rules contain un-evaluated expressions, because the theme generation is confused in the way rules are assigned to colorType buckets. 

Rules are parsed per colorType: https://github.com/angular/material/blob/e1bb097ce19c8644c425724cf6bd8a4da97fefa8/src/core/services/theming/theming.js#L1073

These colorType buckets are created based on the contents of the rules, first looking for each color type /.md-$colorType/, then for each just /$colorType/, then default. The order matters: first, that looks for `primary`, then `accent`, then `warn`.
https://github.com/angular/material/blob/e1bb097ce19c8644c425724cf6bd8a4da97fefa8/src/core/services/theming/theming.js#L973

In this parseRules, only that colorType's expressions is replaced (the `{{$colorType-color-0.26}}`-like matches, built in hueRegex), so a rule in the wrong bucket will keep its expression as-is:
https://github.com/angular/material/blob/e1bb097ce19c8644c425724cf6bd8a4da97fefa8/src/core/services/theming/theming.js#L888

And finally, here's a rule that will get assigned to the `primary` bucket because it has a .md-primary inside, which will therefore keep the {{warn-*}} expression intact during treatment, and that will end up as-is in the DOM:
https://github.com/angular/material/blob/72d068593b7f353d32f8308870d2b0d1c7c759c9/src/components/radioButton/radio-button-theme.scss#L90

There might be more, but I have only seen {{warn-color-0.26}} in style so far. I'm not sure what is wrong, the rules or the generation or sorting method. I found the issue by instrumenting $parse, so my demo works this way too, but it's also an issue if the generated code isn't in the angular-enabled DOM.

No angular expressions in css.

 -  `CodePen Demo which shows your issue:` https://codepen.io/anon/pen/zzMdGb
 -  `Details:`

**AngularJS Versions**: *
 -  `AngularJS Version:` I tried 1.6 and 1.5, likely works with any
 -  `AngularJS Material Version:` both 1.0 and 1.1.4 at least.

**Additional Information**:
 -  `Browser Type: *` Any
 -  `Browser Version: *` Any
 -  `OS: *` Any
 -  `Stack Traces:` See links above.

----

  Add more lines if needed.

jelbourn:
Before this change, the theming code assume that if you have a selector
like "md-primary", all of the expressions therein will only be in terms
of "primary". This was not the case, causing one style to leave an
unevaluated expression in the css output.

This change *also* deletes the troublesome style because it seemed to be
incorrect anyway.

Fixes #10793


ThomasBurleson:
@EladBezalel - as our ngM1 Theming dev, can you review these changes and validate?

EladBezalel:
LGTM

PR #10845, opened 2017-08-11
fix(compiler): removed `@private` to fix dgeni failures

EladBezalel:


mgol:
I added the `@private` keyword so that those methods are not publicly documented as they're supposed to be private.

Is there another way to achieve that?

ThomasBurleson:
@mgol - AFAIR, we simply remove the `@ngdoc` notations. This keeps the method private but maintains the internal docs with returns/params and description.

ThomasBurleson:
@EladBezalel - CI still failing on this one... any ideas why?

mgol:
OK, in that case we'll need you, @EladBezalel to remove the `@ngdoc` pragmas.

mgol:
@ThomasBurleson did you close it on purpose?

ThomasBurleson:
@mgol - No, that was a click-mistake.

EladBezalel:
@ThomasBurleson could seems to pass, I'll remove the @ngdoc notations

ThomasBurleson:
@EladBezalel - as soon as the ngdocs are removed, we can test the docs build and mark this a presubmit.

PR #10869, opened 2017-08-24
fix(list): don't strip download attribute from list items

alphatwit:
Clickable list items containing the `ng-href` attribute currently can't be downloaded on click.


alphatwit:
I signed it!

alphatwit:
cla: yes, robot

PR #10824, opened 2017-07-31
chore(eslint): Configure ESLint Fixes #9261

graingert:

PR #10826, opened 2017-08-01
fix(all): remove all direct use of browser globals

graingert:
use injectable $window and $document instead.

* This removes the private provider $$mdMetaProvider because it used to access `document` before it was available. Meta headers can only be changed via the $$mdMeta service.
* This updates and deprecates $mdThemingProvider.enableBrowserColor() to queue up color changes and dispatch them in the actual service when $$mdMeta is available.
* This adds support for defining panel presets with an injectable function for the case where you want to
define a preset that sets `attachTo: angular.element($document[0].body)` inside a config function.
* This replaces all alert() calls with `$mdDialog.show($mdDialog.alert()...);` calls

This depends on eslint to find the misuse of global variables: https://github.com/angular/material/pull/10824

What are $window and $document?
============================
According to the AngularJS documentation: https://docs.angularjs.org/api/ng/service/$window
> A reference to the browser's window object. While window is globally available in JavaScript, it causes testability problems, because it is a global variable. In **AngularJS we always refer to it through the $window service**, so it may be overridden, removed or mocked for testing.

$document is similar

Why is this needed?
================

* Before $window, window and document, $document were being used **interchangeably**, this PR ensures **consistency**.
* using an injectable $window and $document allows ...
  * ... multiple angular apps to co-exist without destructively altering each other.
  * ... users of angular-material to patch out those values in tests.
  * ... angular-material to work in non-browser environments without those globals: extensions, webworkers and node.
  *  ... material-tools to work without virtual_context https://github.com/angular/material-tools/issues/31

PR #10886, opened 2017-09-05
No browser globals 2

graingert:
use injectable $window and $document instead.

* This removes the private provider $$mdMetaProvider because it used to access `document` before it was available. Meta headers can only be changed via the $$mdMeta service.
* This updates and deprecates $mdThemingProvider.enableBrowserColor() to queue up color changes and dispatch them in the actual service when $$mdMeta is available.
* This adds support for defining panel presets with an injectable function for the case where you want to
define a preset that sets `attachTo: angular.element($document[0].body)` inside a config function.
* This replaces all alert() calls with `$mdDialog.show($mdDialog.alert()...);` calls

This depends on eslint to find the misuse of global variables: https://github.com/angular/material/pull/10824

What are $window and $document?
============================
According to the AngularJS documentation: https://docs.angularjs.org/api/ng/service/$window
> A reference to the browser's window object. While window is globally available in JavaScript, it causes testability problems, because it is a global variable. In **AngularJS we always refer to it through the $window service**, so it may be overridden, removed or mocked for testing.

$document is similar

Why is this needed?
================

* Before $window, window and document, $document were being used **interchangeably**, this PR ensures **consistency**.
* using an injectable $window and $document allows ...
  * ... multiple angular apps to co-exist without destructively altering each other.
  * ... users of angular-material to patch out those values in tests.
  * ... angular-material to work in non-browser environments without those globals: extensions, webworkers and node.
  *  ... material-tools to work without virtual_context https://github.com/angular/material-tools/issues/31


graingert:
This is a temporary PR while waiting for https://github.com/angular/material/pull/10826 to re-open

Issue #8563, opened 2016-05-26
md-autocomplete with md-no-cache and an ajax request goes into an infinite loop wnen second character is typed


alan-hobbart:
**Actual behavior**:
-  What is the issue? *
md-autocomplete with md-no-cache and an ajax request goes into an infinite loop wnen second character is typed

-  What is the expected behavior?
single ajax search for each key press, not an infinite loop of ajax requests afther second character

**CodePen** or Steps to reproduce the issue: *
-  CodePen Demo which shows your issue :
-  Details:
 <md-autocomplete required md-input-name="entityCode"
                         md-selected-item="$ctrl.ngModel"
                         md-search-text="$ctrl.searchText"
                         md-items="item in $ctrl.querySearch($ctrl.searchText)"
                         md-item-text="item.name"
                         md-floating-label="Search for Entity"
                         md-selected-item-change="$ctrl.selectedItemChange(item)"
                         md-no-cache="true"
                         md-delay="300">
            <md-item-template>
                <span md-highlight-text="$ctrl.searchText" md-highlight-flags="i">{{item.name}} ({{item.code}})</span>
            </md-item-template>
            <div ng-messages="searchText.$error">
                <div ng-message="required">This field is required</div>
            </div>
            <md-not-found>
                No matches found.
            </md-not-found>
        </md-autocomplete>

 function querySearch(query) {
            var familyEntityTypesUrlParam = vm.familyEntityTypes.join(',');
            var url = 'api/Entities/Search/' + query + '/' +
                familyEntityTypesUrlParam + '/' +
                config.apiDataReadySearchMaxResults;
            var encodedUrl = encodeURI(url);
            
            console.log('search was run for' + url);

            return dataservice.getData({
                dataUrl: encodedUrl,
                useDataReadyApi: true
            });
        }

**Angular Versions**: *
-  Angular Version: 1.5
-  Angular Material Version: 1.08

**Additional Information**:
-  Browser Type: *
-  Browser Version: *
-  OS: *
-  Stack Traces:

----
Shortcut to create a [new CodePen Demo](http://codepen.io/team/AngularMaterial/pen/bEGJdd).
Note: `*` indicates required information. Without this information, your issue may be auto-closed.

> Do not modify the titles or questions. Simply add your responses to the ends of the questions.
  Add more lines if needed.

ThomasBurleson:
This issue is closed as part of our ‘Surge Focus on Material 2' efforts.
[...]

PR #10916, opened 2017-09-26
fix(autocomplete): infinite $digest loop

chunyenHuang:
I have the same issue with ticket #8563.
```
Error: $rootScope:infdig Infinite $digest Loop for angular 1.6.6 and angular material 1.1.4+
```
It turns out the keydown event fires multiple times. Adding a variable to stop the loop if searchText is not updated.


chunyenHuang:
I signed it!

Splaktar:
This change breaks many of the existing tests, please review the test failures and determine if you can fix the issue without breaking existing tests and risking regressions.

Splaktar:

PR #10955, opened 2017-10-27
update(mdSelect) add ng-change in doc

KimAlexander:
Just doc update


Splaktar:
Please update the commit message to be `docs(select): add ng-change`.

Splaktar:
Thank you for your contribution! It looks like this update was already made via another PR.

PR #10965, opened 2017-11-02
Quick update to documentation `npm run docs:watch`

davestacey:
docs:watch script exists in package.json, so `npm run docs:watch` vs `npm run watch`


davestacey:
I signed it!

PR #10976, opened 2017-11-08
Make ES6 classes work as dialog controllers

trevorade:
If the controller is a function, `invokeCtrl.instance === ctrl`

If the controller is an ES6 Class, `invokeCtrl.instance` is some other object separate from `ctrl`.

Either way, `ctrl` is the correct reference in all cases for post-binding locals.  `invokeCtrl.instance` is the correct reference for pre-binding locals (which only works for function controllers anyways).


jelbourn:
Fixing in #10977

Issue #9125, opened 2016-07-25
slider: position does not update correctly when both bound value and bound max are changed


damianmerrick:
**Actual Behavior**:
-  `What is the issue? *` When I change both the bound value in an md-slider and the bound maximum, the slider position is not always updated correctly. When both value and max are increased, the position is incorrect until the value is changed again. However, when both are decreased, it seems to work ok.
-  `What is the expected behavior?` The slider position should be correct according to both the new value and the new maximum.

**CodePen** (or steps to reproduce the issue): *
-  `CodePen Demo which shows your issue:`Please see this [Plunk](http://plnkr.co/edit/oEvSSsBonXU0kmZg63go?p=preview). 
-  `Details:` Change "Half value" from 2 to 3 as described in the HTML and observe the position change in the slider. Compare the position to what is obtained by changing the value alone with the numerical input.

**Angular Versions**: *
-  `Angular Version:` 1.5.3
-  `Angular Material Version:` 1.1.0-rc4

**Additional Information**:
-  `Browser Type: *` Google Chrome
-  `Browser Version: *` 51.0.2704.103 m
-  `OS: *` Windows 8.1 x64
-  `Stack Traces:` N/A

xiagu:
I encountered this recently myself, and have a theory as to what exactly is going on:

When both the bound value and the bound max are changed in the same digest cycle, the value is updated first, [such that min <= value <= max](https://github.com/angular/material/blob/82fdd2c7c1cd2eb3b84608e2245a353b561c1c28/src/components/slider/slider.js#L413). Then the max is updated.

If the bound value is increased above the old max value, it will be capped at that. You can see this in the plunk linked above, where changing "Half value" from 2 to 3 will leave the slider displaying 5, the max when half-value is 2. (It should display 6.)

Doing the changes in different digest cycles (max first, then value) avoids the problem. That's why the $timeout workaround works.

Splaktar:
I reviewed the PR. If we can get the minor updates made soon, then this should be able to get into the next release.

PR #10980, opened 2017-11-09
fix(slider): fixed incorrect update of model and min/max values in same digest cycle

sviams:
Implemented check for out of sync state between model and view value in Slider, which occurred when both model and either min and/or max values were updated within the same digest cycle. This is due to the $ngModelCtrl formatters kicking in and validating the view value before the Slider directive $observer on the min and max attributes could fire.

Fixes angular/material#9125

### DIFF
--- a/src/components/slider/slider.spec.js
+++ b/src/components/slider/slider.spec.js
@@ -136,6 +136,162 @@ describe('md-slider', function() {
     expect(pageScope.model).toBe(100);
   });
 
+  describe('when raising max and model value equally beyond previous max simultaneously', function() {
+
+    var slider = null;
+
+    beforeEach(function() {
+      slider = setup('min="0" max="{{max}}" ng-model="model"');
+      pageScope.max = 5;
+      pageScope.model = 5;
+      pageScope.$apply();
+      expect(slider.attr('aria-valuenow')).toEqual('5');
+      expect(slider.attr('aria-valuemax')).toEqual('5');
+      pageScope.model = 6;
+      pageScope.max = 6;
+      pageScope.$apply();
+    });
+
+    it('should have updated max correctly', function () {
+      expect(slider.attr('aria-valuemax')).toEqual('6');
+    });
+
+    it('should have updated value correctly', function () {
+      expect(slider.attr('aria-valuenow')).toEqual('6');
+    });
+
+  });
+
+  describe('when raising max and model value beyond previous max simultaneously', function() {
+
+    var slider = null;
+
+    beforeEach(function() {
+      slider = setup('min="0" max="{{max}}" ng-model="model"');
+      pageScope.max = 4;
+      pageScope.model = 3;
+      pageScope.$apply();
+      expect(slider.attr('aria-valuenow')).toEqual('3');
+      expect(slider.attr('aria-valuemax')).toEqual('4');
+      pageScope.model = 6;
+      pageScope.max = 7;
+      pageScope.$apply();
+    });
+
+    it('should have updated max correctly', function () {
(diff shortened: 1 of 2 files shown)


sviams:
I signed it!

Splaktar:
Thank you for your contribution! Please update your commit message to follow the [commit message guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format). This is important for your changes to properly show up in the [changelog](https://github.com/angular/material/blob/master/CHANGELOG.md).

For the last line, `Fixes #9125` should be sufficient.

src/components/slider/slider.spec.js
```diff
@@ -136,6 +136,162 @@ describe('md-slider', function() {
     expect(pageScope.model).toBe(100);
   });
 
+  describe('when updating max and model value equally beyond previous max simultaneously', function() {
+
+    var slider = null;
+
+    beforeEach(function() {
+      slider = setup('min="0" max="{{max}}" ng-model="model"')
```
Splaktar: Please add semicolon at the end.

src/components/slider/slider.spec.js
```diff
@@ -136,6 +136,162 @@ describe('md-slider', function() {
...
+    });
+
+  });
+
+  describe('when updating max and model value below previous max simultaneously', function() {
+
+    var slider = null;
+
+    beforeEach(function() {
+      slider = setup('min="0" max="{{max}}" ng-model="model"')
```
Splaktar: Please add semicolon at the end.

src/components/slider/slider.spec.js
```diff
@@ -136,6 +136,162 @@ describe('md-slider', function() {
...
+    });
+
+  });
+
+  describe('when updating min and model value equally below previous min simultaneously', function() {
+
+    var slider = null;
+
+    beforeEach(function() {
+      slider = setup('min="{{min}}" max="10" ng-model="model"')
```
Splaktar: Please add semicolon at the end.

src/components/slider/slider.spec.js
```diff
@@ -136,6 +136,162 @@ describe('md-slider', function() {
...
+    });
+
+  });
+
+  describe('when updating min and model value above previous min simultaneously', function() {
+
+    var slider = null;
+
+    beforeEach(function() {
+      slider = setup('min="{{min}}" max="10" ng-model="model"')
```
Splaktar: Please add semicolon at the end.

src/components/slider/slider.spec.js
```diff
@@ -136,6 +136,162 @@ describe('md-slider', function() {
...
+      pageScope.min = 5;
+      pageScope.model = 6;
+      pageScope.$apply();
+      expect(slider.attr('aria-valuenow')).toEqual('6');
+      expect(slider.attr('aria-valuemin')).toEqual('5');
+      pageScope.model = 5;
```
Splaktar: This seems to be checking "when updating min and model value equal to previous min simultaneously" rather than "when updating min and model value above previous min simultaneously". Since the previous min was `5` and the new model value is `5`. Am I reading this wrong? Was this intended?

Splaktar:
Oh and one more thing, sorry for taking so long to get to reviewing this PR.

src/components/slider/slider.spec.js
sviams: I agree, this was not entirely clear so I updated both the title and expectations of the test to better match the intention.

Splaktar:
Thanks for the quick turnaround! Sending to presubmit testing now.

PR #10990, opened 2017-11-20
fix(dialog): fix prompt closing on ENTER key when input is required and empty

### LINKED ISSUE #10953: md-dialog: prompt dialogs w/ required=true submit dialog on ENTER key of input when input empty
I can resolve `undefined` by keypress Enter, it should be check if required.
<pre>
this.keypress = function($event) {
  if ($event.keyCode === $mdConstant.KEY_CODE.ENTER) {
    $mdDialog.hide(this.result); // here
  }
};
</pre>
https://github.com/angular/material/blob/2015ae8c9098d17c6ee9dc46859d62026071eae7/src/components/dialog/dialog.js#L647

eknowles:
When using the `required` method in `$mdDialog.prompt()` the user is able to hide the dialog by hitting the `ENTER` key even when the input is empty, defeating the implication of a required input. This PR adds a check on keypress to ensure if the dialog is prompt **and** is required that a value is present.

* Adds missing documentation for `required` method.
* Adds missing test for `required` method.

Fixes #10953 

codepen with issue reproducible https://codepen.io/eknowles/pen/JOpgrq

### DIFF
--- a/src/components/dialog/dialog.js
+++ b/src/components/dialog/dialog.js
@@ -472,6 +472,7 @@ function MdDialogDirective($$rAF, $mdTheming, $mdDialog) {
  * - $mdDialogPreset#htmlContent(string) - Sets the prompt message as HTML. Requires ngSanitize
  *     module to be loaded. HTML is not run through Angular's compiler.
  * - $mdDialogPreset#placeholder(string) - Sets the placeholder text for the input.
+ * - $mdDialogPreset#required(boolean) - Sets the input required value.
  * - $mdDialogPreset#initialValue(string) - Sets the initial value for the prompt input.
  * - $mdDialogPreset#ok(string) - Sets the prompt "Okay" button text.
  * - $mdDialogPreset#cancel(string) - Sets the prompt "Cancel" button text.
@@ -658,7 +659,9 @@ function MdDialogProvider($$interimElementProvider) {
         $mdDialog.cancel();
       };
       this.keypress = function($event) {
-        if ($event.keyCode === $mdConstant.KEY_CODE.ENTER) {
+        var invalidPrompt = isPrompt && this.required && !this.result;
+
+        if ($event.keyCode === $mdConstant.KEY_CODE.ENTER && !invalidPrompt) {
           $mdDialog.hide(this.result);
         }
       };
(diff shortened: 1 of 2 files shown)


eknowles:
I signed it!

eknowles:
I'd be happy to add a demo to the documentation for this if anyone thinks it would benefit

Splaktar:
@eknowles a simple CodePen to show the problematic behavior would be a good first step since the original issue did not include that.

Thank you very much for your contribution.

Splaktar:
@eknowles as mentioned in https://github.com/angular/material/issues/10953#issuecomment-345939302, a CodePen isn't needed as the Dialog demos in the docs demonstrate this issue clearly.

eknowles:
No worries, if needs be, I've got a codepen here showing it https://codepen.io/eknowles/pen/JOpgrq

src/components/dialog/dialog.spec.js
```diff
@@ -781,6 +781,34 @@ describe('$mdDialog', function() {
 
       expect(response).toBe('responsetext');
     }));
+
+    it('should not submit after ENTER key when input is empty and prompt is required', inject(function($mdDialog, $rootScope, $timeout, $mdConstant) {
```
Splaktar: This is good, but I think that we should also have a test that: when required is set and the input has a value that hide is called and the result is correct.

src/components/dialog/dialog.js
```diff
@@ -658,7 +659,9 @@ function MdDialogProvider($$interimElementProvider) {
         $mdDialog.cancel();
       };
       this.keypress = function($event) {
-        if ($event.keyCode === $mdConstant.KEY_CODE.ENTER) {
+        var invalidPrompt = isPrompt && this.required && !this.result;
```
Splaktar: I don't think that this check is accurate enough. The input could be asking for the number of children and 0 would be a valid response. However `!this.result` would be false and flag the result as invalid in that case. `angular.isDefined(this.result)` is a safer approach and is used in many other components.

eknowles: agreed

eknowles:
@Splaktar I've added that missing test case and fix the value check change, sorry for the delay!

PR #10202, opened 2017-01-03
feat(navBar): add disabled attribute

maxjoehnk:
* Add disabled attribute to navitem
* Update demo to showcase ng-disabled and disabled usage
* Add tests
* Ignore navbar specific button styling when disabled

fixes #9667

PR #10992, opened 2017-11-21
feat(navBar): add disabled attribute

### LINKED ISSUE #9667: `md-nav-item` ignores `disabled`-attribute
**Actual Behavior**:
-  `What is the issue? *` `md-nav-item` ignores `disabled`-attribute
-  `What is the expected behavior?` shown in the demo as manual disabled

-  `CodePen Demo which shows your issue:` http://codepen.io/anon/pen/ozBBya
-  `Details:` notice the difference between the two tabs "Attempt to disable" and "Manual disabled"

**Angular Versions**: *
-  `Angular Version:` 1.5.5

**Additional Information**:
-  `Browser Type: *` tested on desktop, but assume all
-  `Browser Version: *` tested firefox and chrome, but assume all
-  `OS: *`tested on Linux, but assume all
-  `Stack Traces:` no

currensy:
Resubmitting @maxjoehnk PR #10202 
* Originally developed by: @maxjoehnk - made a small change, not using $mdUtils which was used by @maxjoehnk in his version. I'm just resubmitting
* Add disabled attribute to navitem
* Update demo to showcase ng-disabled and disabled usage
* Add tests

My Additions: 
* disabled style (reduced opacity and not-allowed cursor) 

fixes #9667

### DIFF
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -388,7 +388,7 @@ MdNavBarController.prototype.onKeydown = function(e) {
 /**
  * @ngInject
  */
-function MdNavItem($mdAria, $$rAF) {
+function MdNavItem($mdAria, $$rAF, $mdUtil, $window) {
   return {
     restrict: 'E',
     require: ['mdNavItem', '^mdNavBar'],
@@ -429,6 +429,7 @@ function MdNavItem($mdAria, $$rAF) {
           '<md-button class="_md-nav-button md-accent" ' +
             'ng-class="ctrl.getNgClassMap()" ' +
             'ng-blur="ctrl.setFocused(false)" ' +
+            'ng-disabled="ctrl.disabled"' +
             'tabindex="-1" ' +
             navigationOptions +
             navigationAttribute + '>' +
@@ -454,11 +455,11 @@ function MdNavItem($mdAria, $$rAF) {
       // When accessing the element's contents synchronously, they
       // may not be defined yet because of transclusion. There is a higher
       // chance that it will be accessible if we wait one frame.
+      var disconnect;
       $$rAF(function() {
         var mdNavItem = controllers[0];
         var mdNavBar = controllers[1];
         var navButton = angular.element(element[0].querySelector('._md-nav-button'));
-
         if (!mdNavItem.name) {
           mdNavItem.name = angular.element(element[0]
               .querySelector('._md-nav-button-text')).text().trim();
@@ -469,8 +470,30 @@ function MdNavItem($mdAria, $$rAF) {
           scope.$apply();
         });
 
+        if('MutationObserver' in $window) {
+          var config = {attributes: true, attributeFilter: ['disabled']};
+          var targetNode = element[0];
+          var mutationCallback = function(mutationList) {
+            $mdUtil.nextTick(function() {
+              mdNavItem.disabled = $mdUtil.parseAttributeBoolean(attrs[mutationList[0].attributeName], false);
+            });
+          };
+          var observer = new MutationObserver(mutationCallback);
+          observer.observe(targetNode, config);
+          disconnect = observer.disconnect.bind(observer);
+        } else {
+          attrs.$observe('disabled', function (value) {
+            mdNavItem.disabled = $mdUtil.parseAttributeBoolean(value, false);
+          });
+        }
+
+
         $mdAria.expectWithText(element, 'aria-label');
       });
+
+      scope.$on('destroy', function() {
+        disconnect();
+      })
     }
   };
 }
--- a/src/components/navBar/navBar.scss
+++ b/src/components/navBar/navBar.scss
@@ -38,6 +38,7 @@ $md-nav-bar-height: 48px;
   &:hover {
     background-color: inherit;
   }
+
 }
 
 md-nav-ink-bar {
--- a/src/components/navBar/demoBasicUsage/index.html
+++ b/src/components/navBar/demoBasicUsage/index.html
@@ -7,11 +7,14 @@
       <md-nav-item md-nav-click="goto('page1')" name="page1">
         Page One
       </md-nav-item>
-      <md-nav-item md-nav-click="goto('page2')" name="page2">
+      <md-nav-item md-nav-click="goto('page2')" name="page2" ng-disabled="secondTabDisabled">
         Page Two
       </md-nav-item>
       <md-nav-item md-nav-click="goto('page3')" name="page3">
         Page Three
+      </md-nav-item>
+      <md-nav-item md-nav-click="goto('page4')" name="page4" disabled>
+        Page Four
       </md-nav-item>
       <!-- these require actual routing with ui-router or ng-route, so they
       won't work in the demo
@@ -28,6 +31,10 @@
     </div>
 
     <md-checkbox ng-model="disableInkBar">Disable Ink Bar</md-checkbox>
+
+    <md-checkbox ng-model="secondTabDisabled" aria-label="Disable Second Tab" style="margin: 5px;">
+      Disable Second Tab
+    </md-checkbox>
 
   </md-content>
 </div>
--- a/src/components/navBar/navBar.spec.js
+++ b/src/components/navBar/navBar.spec.js
@@ -176,6 +176,40 @@ describe('mdNavBar', function() {
               .toBe('{"reload":true,"notify":true}');
         });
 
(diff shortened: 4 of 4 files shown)


Splaktar:
Restarted the build in Travis and everything looks good now. Thank you!

src/components/navBar/navBar.js
```diff
@@ -469,6 +470,10 @@ function MdNavItem($mdAria, $$rAF) {
           scope.$apply();
         });
 
+          attrs.$observe('disabled', function (value) {
+              mdNavItem.disabled = value.length === 0 ? true : !!value;
```
Splaktar: This conversion should use the `parseAttributeBoolean()` function from `$mdUtil` to be consistent with other directives like [this one](https://github.com/angular/material/blob/9f198c93af2e883d25bbce9fadf2a3d4ac5f4283/src/components/autocomplete/js/autocompleteController.js#L211).

currensy:
@Splaktar changes made.

src/components/navBar/navBar.scss
```diff
@@ -38,6 +38,11 @@ $md-nav-bar-height: 48px;
   &:hover {
     background-color: inherit;
   }
+
+  &[disabled] {
+    cursor: not-allowed;
+    opacity: 0.5;
```
clshortfuse: Changing `opacity` is bad here. It will cause a device to be rendered in 3D/Vector by most browsers instead the general 2D rendering.

You should apply the styling change to `navBar-theme.scss` instead, and perhaps use `foreground-3` for disabled items (I don't exactly remember what the right color should be).

src/components/navBar/navBar.js
```diff
@@ -469,6 +470,10 @@ function MdNavItem($mdAria, $$rAF) {
           scope.$apply();
         });
 
+          attrs.$observe('disabled', function (value) {
```
clshortfuse: This seems unnecessary since ngDisabled does this automatically for us. It's built into AngularJS already. I don't think think we need to touch any of the JS at all.

https://docs.angularjs.org/api/ng/directive/ngDisabled

currensy: I didn't quite understand what you mean, how can we disable the inner button element?

clshortfuse: Ah, I see what you mean. 

We have to be careful with adding features like this. The problem is introducing these checks will add on to the digest loop with watchers/observers. This would cause an unexpected strain on the CPU when users upgrade, especially on something like nav-item.

Using a MutationObserver on the `ctrl` element's `disabled` attribute would be more performant. If the browser doesn't support it, then we can fall back to watchers/observers.

Splaktar: @currensy are you still able to look into this?

I know that there aren't many examples of doing this in the AngularJS Material code base, but here is one: [tabsDummyWrapperDirective.js](https://github.com/angular/material/blob/master/src/components/tabs/js/tabsDummyWrapperDirective.js#L24-L52).

It's not an exact match for what needs to be done here, but hopefully it can put you on the right track?

currensy:
```javascript
if('MutationObserver' in $window) {
          var config = {attributes: true, attributeFilter: ['disabled']};
          var targetNode = element[0];
          var mutationCallback = function(mutationList) {
            mdNavItem.disabled = $mdUtil.parseAttributeBoolean(attrs[mutationList[0].attributeName], false);
            scope.$apply();
          };
          var observer = new MutationObserver(mutationCallback);
          observer.observe(targetNode, config);
          disconnect = observer.disconnect.bind(observer);
        } else {
          attrs.$observe('disabled', function (value) {
            mdNavItem.disabled = $mdUtil.parseAttributeBoolean(value, false);
          });
        }
```
-------------
The demo seems to be working fine but navBar's disable support tests are failing. Did I miss something?

Splaktar: What do the test failures look like?

currensy: Expected undefined to be true.
	@src/components/navBar/navBar.spec.js:190:11

Expected undefined to be false.
	@src/components/navBar/navBar.spec.js:204:11
	Expected undefined to be true.
	@src/components/navBar/navBar.spec.js:207:11

Splaktar: You may also want to refer to the use of MutationObserver in [tooltip.js](https://github.com/angular/material/blob/master/src/components/tooltip/tooltip.js#L157-L193).

In your code, I'm not sure that `mutationList[0]` is always the right mutation? In the tooltip example it goes through all of the mutations looking for `mutation.attributeName === 'disabled'`.

Splaktar: Should the target node just be `element` instead of `element[0]`?

currensy: I guess the attributeFilter is supposed to ensure that? - Exactly.

I tried to filter it manually anyway, but the result stays the same. It looks like the evaluation in the test happens prior to the mdNavItem.disabled assignment, some sort of race condition...

currensy: element is a wrapped node while the observer expects to get [Node](https://developer.mozilla.org/en-US/docs/Web/API/Node)

Splaktar: Maybe instead of `scope.$apply();`?
```
        $mdUtil.nextTick(function() {
          mdNavItem.disabled = $mdUtil.parseAttributeBoolean(attrs[mutationList[0].attributeName], false);
        });
```

currensy: I've actually made this change (looks better than calling $apply imo) - no luck with tests though

Splaktar: Ah yeah, it wouldn't want the jQuery wrapped element. If you can't figure it out, you can push up your changes and I can try to debug it in the next few days.

Splaktar:
Can you please squash your commits and revert the change to `package-lock.json`?

Splaktar:
This gist has some guidance on how to squash commits: https://gist.github.com/patik/b8a9dc5cd356f9f6f980#combining-the-commits

currensy:
@Splaktar - Not sure how to play it now that i've merged master to my forked branch.

currensy:
Do I need to reset package-lock.json to it's state before my latest commit on 01/30?

Splaktar:
Yes, please. That will avoid any merge conflicts.

currensy:
I think it's fine now.

currensy:
Remember we tried to addressed these failing tests 12 days ago? [https://github.com/angular/material/pull/10992#discussion_r164677146](url) 

I didn't have the chance to debug it yet, I'll try to make some time

Splaktar:
Hrm, I don't recall the tests failing on this PR other than some bogus CI issues (https://github.com/angular/material/pull/10992#issuecomment-346140013) that were solved when I re-ran the builds.

currensy:
Look at previous comments, the tests were failing locally you asked for me to push the changes so you could take a look at it :)

image

Splaktar:
Oh, now I remember! Thanks. I'll put this on my TODO list to take a look at.

currensy:
```javascript 
it('should set the disabled attribute', function () {
          create('<md-nav-bar>' +
              '  <md-nav-item md-nav-href="#1" name="tab1">' +
              '    tab1' +
              '  </md-nav-item>' +
              '  <md-nav-item md-nav-href="#2" name="tab2" disabled>' +
              '    tab2' +
              '  </md-nav-item>' +
              '</md-nav-bar>');
          $timeout(function(){
            var tabCtrl = getTabCtrl('tab2');
            expect(tabCtrl.disabled).toBe(true);
          });

      });
```

wrapping expect function with $timeout seems to solve the problem. I know this technique is sometimes used to initiate new digest cycle, is it appropriate to use it in tests?

clshortfuse:
@currensy Probably because you'd have to wait for a digest for the changes to take effect. Most spec tests use `scope.$apply();`, I believe.

currensy:
@clshortfuse Tests are failing with just $scope.$apply() - by the time expect is executed tabCtrl.disabled is still undefined.

Splaktar:
I think that the `$timeout` approach will be OK as it is used for other tests where a `MutationObserver` is involved like the various `updates pagination and ink styles` tests.

currensy:
Good catch. This happens because I've used MutationObserver to watch for changes on disabled attribute. It's callback doesn't fire on load therefore the disabled property on the controller is never initialized. Would you file an issue or should I?

ferc:
I created a new issue: #11172

PR #10994, opened 2017-11-22
chore: add script to generate a script to schedule presubmit runs

jelbourn:
This will make it easier for us to schedule presubmit runs during off-peak hours

### DIFF
--- a/scripts/write-presubmit-scheduler.js
+++ b/scripts/write-presubmit-scheduler.js
@@ -1,0 +1,83 @@
+#!/usr/bin/env node
+
+const path = require('path');
+const fs = require('fs');
+const GitHubApi = require('github');
+const github = new GitHubApi();
+
+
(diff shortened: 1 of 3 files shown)


scripts/write-presubmit-scheduler.js
```diff
@@ -0,0 +1,83 @@
...
+const path = require('path');
+const fs = require('fs');
+const GitHubApi = require('github');
+const github = new GitHubApi();
+
+
+/** CONFIGURATION: change these things if you want to tweak how the runs are made. */
+
+/** Path to the local ng-material. By default based on the location of this script. */
+const localRepo = path.resolve(__dirname, '..',);
```
Splaktar: The second command seems to be unnecessary.

jelbourn: Done

PR #11020, opened 2017-12-15
fix(util): nodes.item is not a function

johanbyren:
Fixed an error where Firefox 57.0.1 threw this error: `Error: nodes.item is not a function`

image

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -146,7 +146,7 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
 
       var results = [];
       for (var i = 0; i < nodes.length; ++i) {
-        results.push(nodes.item(i));
+        results.push(nodes[i]);
       }
       return results;
     },


johanbyren:
I signed it!

Splaktar:
Thank you for your contribution. Please follow the [commit message guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format).

Splaktar:
Does this only occur on Firefox? What version of AngularJS and AngularJS Material does this occur on? Do you have a CodePen that reproduces the issue? It would be preferred that an issue be submitted with those details before sending a PR in the future.

johanbyren:
I have tried it on Opera, Chrome, Safari and it only occurs on Firefox. I had a friend try it on Edge and it seems to throw error there as well: 
image

I have AngularJS 1.5.9 and Angular Material 1.1.3

Splaktar:
OK, thank you for the information. How can we reproduce this issue? We need some reproduction steps at minimum if you can't provide a CodePen example.

Splaktar:
We can't merge this without clear reproduction steps (ideally a CodePen demo) so that we can verify the issue and verify that the fix addresses the issue. Please provide that info or we'll have to close this issue.

mmpowers:
We only started hitting this problem on Firefox when the webcomponentsjs polyfill is present.  The workaround was to inject $mdUtil and patch this function directly.  (Note the test for whether nodes.item exists before accessing it as an array.)  

    $mdUtil.nodesToArray = function(nodes) {
      nodes = nodes || [];
      var results = [];
      for (var i = 0; i < nodes.length; ++i) {
        if (nodes.item) {
          results.push(nodes.item(i));
        } else {
          results.push(nodes[i]);
        }
      }
      return results;
    };

Splaktar:
Unfortunately, we can't accept this change without unit tests and a reproduction.

PR #11031, opened 2017-12-22
fix(gridList): demo example responsiveness broken on Safari

wangsha:
Tested Safari Version: 11.0.2
Machine: Macbook Pro

Left: Safari, grid not responsive to browser resizing
Right: Chrome, behave as expected
screen shot 2017-12-22 at 3 51 07 pm

### DIFF
--- a/src/components/gridList/demoBasicUsage/styles.css
+++ b/src/components/gridList/demoBasicUsage/styles.css
@@ -5,9 +5,3 @@ md-grid-list { margin: 8px; }
 .blue { background: #84ffff; }
 .purple { background: #b388ff; }
 .red { background: #ff8a80; }
-
-
-
-md-grid-tile {
-  transition: all 400ms ease-out 50ms;
-}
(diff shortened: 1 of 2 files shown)


Splaktar:
Thank you very much for the contribution!

Do you have any more information about why this `transition: all 400ms ease-out 50ms;` style is not handled properly by Safari?

There is nothing listed on https://caniuse.com/#feat=css-transitions related to Safari issues with `transition`.

I can confirm that this is an issue for me as well on Safari Version 11.0.2 (13604.4.7.1.3) on macOS 10.13.2.

Splaktar:
Can you please update your commits to squash them into a single commit and then follow the [commit message guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format)?

Splaktar:
Re-ran the CI builds and they now pass.

Splaktar:
It would be really helpful to open an issue that describes the problem here in more detail. Ideally in the future, that would happen before submitting a PR.

wangsha:
Thanks @Splaktar  for the feedback. Created an issue #11043 and squashed all commits.

PR #11044, opened 2018-01-02
fix($mdGesture):reAdd isAndroid & isIos to $get

Ardeshir81:
This was removed as part of this commit : https://github.com/angular/material/pull/9803
I think it was observed that isAndroid and isIos is not used anywhere else in material, so it was removed.
But the normal usages of $mdGesture would be broken after this commit , I mean a developer can't simply inject `$mdGesture` and use it as :
```javascript
if($mdGesture.isAndroid):{
  this.suggestDownloadOurAndroidApp()
}
```

in our use case, after updating from material 1.1.1 to 1.1.5, a part of our app broke.

### DIFF
--- a/src/core/services/gesture/gesture.js
+++ b/src/core/services/gesture/gesture.js
@@ -87,6 +87,8 @@ function MdGesture($$MdGestureHandler, $$rAF, $timeout) {
   var self = {
     handler: addHandler,
     register: register,
+    isAndroid: isAndroid,
+    isIos: isIos,
     // On mobile w/out jQuery, we normally intercept clicks. Should we skip that?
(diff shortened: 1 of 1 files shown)


Ardeshir81:
Alright dear bot! I signed it!

Splaktar:
It does appear that the linked PR did remove 2 APIs which could have been used by users of AngularJS Material. 

That said, the method currently used by `mdGesture` to determine iOS vs Android is not following the latest best practices. In Angular Material, the CDK exposes a more comprehensive set of checks: https://github.com/angular/material2/blob/master/src/cdk/platform/platform.ts.

It seems like we should add these APIs back in for the 1.1.x releases and consider removing them sometime post-1.2.0 if there is still an interest in removing them.

PR #11064, opened 2018-01-12
build: configure ESLint coverage for gulp, scripts, docs, config

Splaktar:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Feature
[x] Code style update (formatting, local variables)
[x] Refactoring (no functional changes, no api changes)
[x] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Currently none of the repository is linted with ESLint except for `src/`. This means that all of the scripts and code in `config/ docs/ gulp/ scripts/` is not covered. This code contains quite a wide range of bad practices. Additionally, there is an informal mix of ES5 and ES6 in the repository.

configure parts of the repo for ES5 (src, docs/app/js)
configure parts of the repo for ES6 (scripts, gulp, config, root)
remove unused committers.json
clean up ESLint errors
some minor ES6 updates and refactoring outside of `src/`

Issue Number: 
N/A

## What is the new behavior?
Add linting coverage for the vast majority of the repository.
Explicitly configure different sections of the repo for ES5 or ES6 and run linting to verify that no crossover violations occur (breaking browser support).

```
[ ] Yes
[x] No
```

## Other information
N/A


Splaktar:
@graingert As you implemented the initial ESLint support and have another ESLint PR in flight, can you please take a look at this PR and suggest any possible improvements or best practices for configuring ESLint.

I put this together while reviewing all of the tooling and scripting in the project so that I could better understand it prior to putting out the 1.1.6 release. 

- [x] Run all of the gulp tasks successfully
- [x] Manually test and verify that the docs site functioned and looks as expected
- [x] Run the release script in dryRun mode and verify that it completes properly
- [x] Run `npm run lint` and verify that there are no failures

docs/app/js/.eslintrc.json
```diff
@@ -0,0 +1,274 @@
...
+  },
+  "env": {
+    "browser": true
+  },
+  "globals": {
+    "angular": true,
+    "CryptoJS": true,
+    "hljs": true
+  },
+  "overrides": [
```
graingert: eg this overrides list can add an object with `files: ["docs/"]` configured for 'docs'.

docs/app/js/highlight-angular.js
```diff
@@ -53,6 +53,7 @@ angular.module('docsApp')
 
           var highlightedCode = hljs.highlight(attr.language || attr.lang, lines.join('\n'), true);
           highlightedCode.value = highlightedCode.value
+            // eslint-disable-next-line no-div-regex
             .replace(/=<span class="hljs-value">""<\/span>/gi, '')
```
graingert: should be possible to do `\=` here

docs/config/index.js
```diff
@@ -42,6 +42,7 @@ module.exports = new Package('angular-md', [
 
   computeIdsProcessor.idTemplates.push({
     docTypes: ['content'],
+    // eslint-disable-next-line no-template-curly-in-string
     idTemplate: 'content-${fileInfo.relativePath.replace("/","-")}',
```
graingert:
```js
idTemplate: { getId: ({ fileInfo}) => `content-${fileInfo.relativePath.replace("/","-")}` }
```

src/.eslintrc.json
```diff
@@ -0,0 +1,273 @@
+{
+  "extends": "eslint:recommended",
```
graingert: same here, use 'overrides'

scripts/write-presubmit-scheduler.js
```diff
@@ -19,6 +19,7 @@ const logDir = '/tmp/v1-pr-presubmit-logs';
  * The default path is stored in an environment variable because it references an internal-Google
  * location.
  */
+// eslint-disable-next-line no-process-env
```
graingert: ignore this in the 'scripts' overrides.

updateVersionPicker.js
```diff
@@ -47,23 +44,20 @@
```
graingert: just use 

```js
`cp -r ${docs.latest}` latest`,
```

updateVersionPicker.js
```diff
@@ -47,23 +44,20 @@
 
   //-- utility methods
 
-  function fill(str) {
-    return str.replace(/\{\{[^\}]+\}\}/g, function (match) {
+  function fill (str) {
+    return str.replace(/{{[^}]+}}/g, function (match) {
+      // eslint-disable-next-line no-eval
       return eval(match.substr(2, match.length - 4));
```
graingert: instead of eval just use template string functions.

graingert: actually this is only used for `cp -r ${docs.latest} latest`,

.eslintignore
```diff
@@ -0,0 +1,3 @@
+src/core/services/compiler/compiler.spec.js
```
graingert: you could probably do these with the 'overrides' feature.

graingert:
@Splaktar good idea to upgrade all the util scripts to modern js. Looks like there was a bit of a misunderstanding on how .eslintrc inheritance/overrides operates.

Splaktar:
Yep, thank you very much for the review. I had a feeling that it may have some inheritance features like TSLint, I just didn't have time to research every part of how it worked.

docs/app/js/.eslintrc.json
Splaktar: I see that `overrides` is an array... oh and it has a seperate `rules`, `globals`, and `env` that applies to that set of `files`. Oh very nice. Then I can do it all in that one file, excellent. I'll work on merging things together. Thank you!

.eslintignore
Splaktar: I tried to add them to an `excludedFiles` array, but that didn't work. Most of them cause ESLint to actually throw a parsing error if the file is evaluated at all. Adding them to the ignore file avoided this issue.

scripts/write-presubmit-scheduler.js
Splaktar: I actually like that rule and want to eventually make changes to get it passing. So I don't want to disable it that widely.

docs/app/js/highlight-angular.js
Splaktar: That gives me `no-useless-escape` error. I think that I am just going to disable `no-div-regex` globally.

updateVersionPicker.js
Splaktar: Done.

.eslintrc.json
```diff
@@ -1,271 +1,312 @@
 {
-    "extends": "eslint:recommended",
```
graingert: I think it would be convenient to not change the indenting in this PR so there's fewer rebase issues with my document/window PR, and so it's easier to review changes here.

gulp/config.js
```diff
@@ -1,5 +1,6 @@
 var args = require('minimist')(process.argv.slice(2));
-var VERSION = args.version || require('../package.json').version;
+var version = require('../package.json').version;
+var VERSION = args.version || version;
```
DevVersion: This is confusing and ambiguous. I'd prefer just inlining this.

```js
var version = args.version || require('../package.json').version;
```

DevVersion: Also if you update it like that in the `const.js` file, you can just import the `VERSION` export here. This removes duplicate logic.

gulp/tasks/karma-fast.js
```diff
@@ -30,14 +29,15 @@ exports.task = function (done) {
 
   gutil.log('Running unit tests on unminified source.');
 
-  karma = new Server(karmaConfig, captureError(clearEnv,clearEnv));
+  var karma = new Server(karmaConfig, captureError(clearEnv,clearEnv));
```
DevVersion: Is it intended to use `var` here? It kind of feels inconsistent to me (why not `const`)

graingert: it's odd that it didn't complain.

gulp/config.js
graingert: that breaks the top level require only rule.

graingert: and you can't use any of the wacy conditional require stuff with esm, so it's good to require top level imports only.

maybe use:
```js
const pkg = require('../package.json');

const VERSION = args.version || pkg.version;
```

DevVersion: Yeah, the important thing is just, that the `version` and `VERSION` names are way too similar and should be somehow differentiated.

.eslintrc.json
Splaktar: You have a good point there. Can you please update your PR to use 2 spaces instead of 4? That should minimize most of it. I'm not sure if/when, we'll be able to get that in as it doesn't have non-zero risk. I would rather not keep this file using inconsistent 4 space formatting for an extended period of time.

graingert: > I'm not sure if/when, we'll be able to get that in as it doesn't have non-zero risk

as it has non-zero risk?

gulp/tasks/karma-fast.js
Splaktar: I updated `scripts/` but I didn't update all of the gulp tasks yet...

graingert: probably worth running the safe transforms

.eslintrc.json
graingert: I'd be happy with the indent change happening in this PR as long as it's in one atomic commit at the end

graingert: It's currently impossible to review

gulp/tasks/karma-fast.js
Splaktar: Yep, done. Pushing soon.

.eslintrc.json
Splaktar: OK, I understand. That's the hard thing with the original formatting not being the same as the project. I guess it's using CLANG for formatting atm (which I don't actually use). I'm not sure why `.editorconfig` isn't used other than Google is used to CLANG. Either way, I don't have your original formatting settings. I'll set it back to 4 spaces and see how that looks.

Splaktar: Pushed. Yeah, the diff looks much better now. Please take a look.

package.json
```diff
@@ -85,6 +85,6 @@
     "docs:watch": "gulp watch site --dev",
     "test:fast": "gulp karma-fast",
     "test:full": "gulp karma",
-    "lint": "eslint src"
+    "lint": "eslint src docs gulp config scripts"
```
graingert: could probably be `eslint .`

Splaktar: Would that try to lint `dist/`?

graingert: I believe eslint also uses `.gitignore ` by default.

graingert: ah, it looks like you can't set two ignore files. Probably worth just adding dist to .eslintignore.

graingert: https://github.com/eslint/eslint/issues/9794

Splaktar: Updated and pushed.

.eslintignore
graingert: `test/` doesn't seem to have that much wrong with it. I suspect they can be added to the `overrides.2.files` array.

Splaktar: I started into cleaning those up and getting them passing, but there were a number of things that I am not comfortable changing at this time. I'd like to keep that as part of a separate PR since those files are actually used to test the library and changing them would make this PR no longer 'merge safe'.

graingert:
Makes sense.

Splaktar:
I ran into some issues testing the build-contributors task. After debugging, it uncovered https://github.com/mgechev/github-contributors-list/issues/12. I opened https://github.com/mgechev/github-contributors-list/pull/13 to fix the issue.

This should be good to go now.

Issue #11124, opened 2018-02-16
md-chips: ng-required is not validating


IMM0rtalis:
### Bug, feature request, or proposal: 
Feature: md-chips should validate against ng-required.

### What is the expected behavior?
Since md-chips is an element which gets input from the user it should get proper validation. md-max-chips was added but the basic required is not working with the current implementation. 

### What is the current behavior?
ng-required has no effect.

##### [CodePen Demo](http://codepen.io/team/AngularMaterial/pen/bEGJdd) which shows your issue: 
https://codepen.io/anon/pen/gvXJee

### What is the use-case or motivation for changing an existing behavior?
Elements which are accepting user input should have validation working.

### Is there anything else we should know? 
Already discussed here but sadly closed during clean-up-time. #8126

PR #11125, opened 2018-02-16
feat(md-chips): added validation for ng-required

IMM0rtalis:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[x] Other... Please describe: In my opinion it is more a bug than a feature because this should be a default behavior
```

ng-required not working with md-chips

Issue Number: #11124

## What is the new behavior?
ng-required works with md-chips

```
[ ] Yes
[x] No (I would say no)
```

## Other information
Pull-Request is mainly a follow-up from #8034. I had the same idea in how to add this but found this not followed/merged pull-request by @Emeegeemee .
So nearly everything of the code got reviewed once by @DevVersion & @crisbeto . I just changed the naming a bit.

### DIFF
--- a/src/components/chips/demoValidation/index.html
+++ b/src/components/chips/demoValidation/index.html
@@ -1,0 +1,38 @@
+<div ng-controller="ChipsValidationCtrl as ctrl" layout="column" ng-cloak>
+  <md-content class="md-padding" layout="column">
+
+    <h2 class="md-title">Required</h2>
+    <form name="fruitForm">
+      <md-chips name="fruits"
+                ng-model="ctrl.selectedFruit"
+                ng-required="true"
+                placeholder="Add a fruit">
+      </md-chips>
+      <div class="md-chips-messages"
+           ng-show="fruitForm.fruits.$dirty || fruitForm.$submitted"
+           ng-messages="fruitForm.fruits.$error">
+        <div ng-message="required">At least one fruit is required</div>
+      </div>
+    </form>
+
+
+    <h2 class="md-title">Max Chips</h2>
+    <form name="vegetableForm">
+      <md-chips name="vegetables"
+                ng-model="ctrl.selectedVegetables"
+                placeholder="Add a vegetable"
+                md-max-chips="5">
+      </md-chips>
+      <div class="md-chips-messages"
+           ng-show="vegetableForm.vegetables.$dirty || vegetableForm.$submitted"
+           ng-messages="vegetableForm.vegetables.$error">
+        <div ng-message="md-max-chips">You reached the maximum number of vegetables</div>
+      </div>
+    </form>
+
+    <p class="note">
+      Be aware that error messages for chips are not styled by default since they are not part of <code>md-input-container</code>.
+    </p>
+
+  </md-content>
+</div>
--- a/src/components/chips/demoValidation/script.js
+++ b/src/components/chips/demoValidation/script.js
@@ -1,0 +1,11 @@
+(function () {
+  'use strict';
+  angular
+    .module('chipsValidationDemo', ['ngMaterial', 'ngMessages'])
+    .controller('ChipsValidationCtrl', ValidationCtrl);
+
+  function ValidationCtrl () {
+    this.selectedFruit = [];
+    this.selectedVegetables = [];
+  }
+})();
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -160,6 +160,13 @@ function MdChipsCtrl ($scope, $attrs, $mdConstant, $log, $element, $timeout, $md
    */
   this.chipAppendDelay = DEFAULT_CHIP_APPEND_DELAY;
 
+  /**
+   * Collection of functions to call to un-register watchers
+   *
+   * @type {Array}
+   */
+  this.deRegister = [];
+
   this.init();
 }
 
@@ -170,18 +177,32 @@ MdChipsCtrl.prototype.init = function() {
   var ctrl = this;
 
   // Set the wrapper ID
-  ctrl.wrapperId = '_md-chips-wrapper-' + ctrl.$mdUtil.nextUid();
+  this.wrapperId = '_md-chips-wrapper-' + this.$mdUtil.nextUid();
 
   // Setup a watcher which manages the role and aria-owns attributes
-  ctrl.$scope.$watchCollection('$mdChipsCtrl.items', function() {
-    // Make sure our input and wrapper have the correct ARIA attributes
-    ctrl.setupInputAria();
-    ctrl.setupWrapperAria();
-  });
+  this.deRegister.push(
+    this.$scope.$watchCollection('$mdChipsCtrl.items', function() {
+      // Make sure our input and wrapper have the correct ARIA attributes
+      ctrl.setupInputAria();
+      ctrl.setupWrapperAria();
+    })
+  );
 
-  ctrl.$attrs.$observe('mdChipAppendDelay', function(newValue) {
-    ctrl.chipAppendDelay = parseInt(newValue) || DEFAULT_CHIP_APPEND_DELAY;
-  });
+  this.deRegister.push(
+    this.$attrs.$observe('mdChipAppendDelay', function(newValue) {
+      ctrl.chipAppendDelay = parseInt(newValue) || DEFAULT_CHIP_APPEND_DELAY;
+    })
+  );
+};
+
+/**
+ * Destructor for cleanup
+ */
+MdChipsCtrl.prototype.$onDestroy = function $onDestroy() {
+  var $destroyFn;
+  while (($destroyFn = this.deRegister.pop())) {
+    $destroyFn.call(this);
+  }
 };
 
 /**
@@ -560,6 +581,7 @@ MdChipsCtrl.prototype.hasMaxChipsReached = function() {
  */
 MdChipsCtrl.prototype.validateModel = function() {
   this.ngModelCtrl.$setValidity('md-max-chips', !this.hasMaxChipsReached());
+  this.ngModelCtrl.$validate(); // rerun any registered validators
 };
 
 /**
@@ -685,6 +707,12 @@ MdChipsCtrl.prototype.configureNgModel = function(ngModelCtrl) {
   this.ngModelCtrl = ngModelCtrl;
 
   var self = this;
+
+  // in chips the meaning of $isEmpty changes
+  ngModelCtrl.$isEmpty = function(value) {
+    return !value || value.length === 0;
+  };
+
   ngModelCtrl.$render = function() {
     // model is updated. do something.
     self.items = self.ngModelCtrl.$viewValue;
@@ -717,6 +745,43 @@ MdChipsCtrl.prototype.onInputBlur = function () {
 };
 
 /**
+ * Configure event bindings on input element.
+ * @param inputElement
+ */
+MdChipsCtrl.prototype.configureInput = function configureInput(inputElement) {
+  // Find the NgModelCtrl for the input element
+  var ngModelCtrl = inputElement.controller('ngModel');
+  var ctrl = this;
+
+  if (ngModelCtrl) {
+
+    // sync touched-state from inner input to chips-element
+    this.deRegister.push(
+      this.$scope.$watch(
+        function() {
+          return ngModelCtrl.$touched;
+        },
+        function(isTouched) {
+          isTouched && ctrl.ngModelCtrl.$setTouched();
+        }
+      )
+    );
+
+    // sync dirty-state from inner input to chips-element
+    this.deRegister.push(
+      this.$scope.$watch(
+        function() {
+          return ngModelCtrl.$dirty;
+        },
+        function(isDirty) {
+          isDirty && ctrl.ngModelCtrl.$setDirty();
+        }
+      )
+    );
+  }
+};
+
+/**
  * Configure event bindings on a user-provided input element.
  * @param inputElement
  */
@@ -743,7 +808,7 @@ MdChipsCtrl.prototype.configureUserInput = function(inputElement) {
       .attr({ tabindex: 0 })
       .on('keydown', function(event) { scopeApplyFn(event, ctrl.inputKeydown) })
       .on('focus', function(event) { scopeApplyFn(event, ctrl.onInputFocus) })
-      .on('blur', function(event) { scopeApplyFn(event, ctrl.onInputBlur) })
+      .on('blur', function(event) { scopeApplyFn(event, ctrl.onInputBlur) });
 };
 
 MdChipsCtrl.prototype.configureAutocomplete = function(ctrl) {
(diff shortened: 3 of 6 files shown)


src/components/chips/demoValidation/index.html
```diff
@@ -0,0 +1,27 @@
...
+                ng-model="ctrl.selectedVegetables"
+                placeholder="Add a vegetable"
+                md-max-chips="5">
+      </md-chips>
+      <div ng-messages="vegetableForm.vegetables.$error">
+        <div ng-message="md-max-chips">You reached the maximum number of vegetables</div>
+      </div>
+    </form>
+  </md-content>
+</div>
```
Splaktar: Blank line at end of file please.

src/components/chips/demoValidation/script.js
```diff
@@ -0,0 +1,11 @@
...
+  'use strict';
+  angular
+    .module('chipsValidationDemo', ['ngMaterial', 'ngMessages'])
+    .controller('ChipsValidationCtrl', ValidationCtrl);
+
+  function ValidationCtrl () {
+    this.selectedFruit = [];
+    this.selectedVegetables = [];
+  }
+})();
```
Splaktar: Blank line at end of file please.

src/components/chips/demoValidation/index.html
```diff
@@ -0,0 +1,27 @@
...
+    <h2 class="md-title">Required</h2>
+    <form name="fruitForm">
+      <md-chips name="fruit"
+                ng-model="ctrl.selectedFruit"
+                ng-required="true"
+                placeholder="Add a fruit">
+      </md-chips>
+      <div ng-messages="fruitForm.fruit.$error">
+        <div ng-message="required">At least one fruit is required</div>
+      </div>
```
Splaktar: These don't show up red because the `chips-theme.scss` is lacking styles for `md-invalid`. Can you please review the `input-theme.scss` and implement something similar for chips as part of this PR?

Splaktar: `input.scss` also uses a `md-input-message-animation` class to animate the message in and to set a smaller line height and font size (among other things). Can we please add those styles to `chips.scss` so that they will be similar when used in a form together?

Splaktar: The error message is showing up even when the input is pristine and not dirty. For input, this isn't the case. Can you please look into this?

IMM0rtalis: Ok, I found out what md-input-container is doing: It adds a special animation-class `md-input-messages-animation` (https://github.com/angular/material/blob/v1.1.7/src/components/input/input.js#L840) and does some stuff with it. I see two ways here. A: make the demos with md-chips inside md-input-container and assure that this have the same look&feel. B: Add a chips-special-class and add styles there. 
But B is not really following the ways it is made atm. ng-messages is just styled inside certain containers. So just A is left, but are chips belonging to md-input-container? - Is there any rule which forbids it?

IMM0rtalis:
@Splaktar you labeled this PR and the related issue with "bug". Is it a bug or feature in your opinion? - because then I should adjust the commit-message-prefix.

Splaktar:
Yeah, I think you are right. I'll updated this to `enhancement`. Using `feat` in the commit message is good.

src/components/chips/demoValidation/index.html
Splaktar: Yeah, I can see how we wouldn't want to style `ng-messages` that aren't contained in an AngularJS Material component. I've asked @DevVersion for some input and created a [CodePen](https://codepen.io/Splaktar/pen/QQmwop) to demonstrate how `ng-messages` are styled with chips today (with `md-max-chips`) for reference since there is no example in the demos.

DevVersion:
I think just implementing the logic for supporting the `required` validator is fine here. That's the main intention of the PR and the design of error messages for chips is more a separate issue in my opinion.

For now, we could just have some custom CSS in the demo and maybe show a small caption under the chips demo that explains that there is no default CSS for error messages yet.

Placing chips inside of the  `<md-input-container>` element makes sense, because the chips can be also considered as an input. Since making the input container more abstract will likely result in breaking changes, I'd vote for another container like `<md-form-field>`.

The `<md-form-field>` and `<md-input-container>` would then share somehow the CSS for the error messages. e.g. using a class like `.md-error-messages-container`. The actual directives for animations could then be also updated to work in either `form-field` or `input-container`.

As you see, this turns out to be a whole new issue, so I'd say that this should be handled in a follow-up.

Splaktar:
@DevVersion thank you very much. That makes a lot of sense.

@IMM0rtalis can you please look at https://github.com/angular/material/pull/11125#discussion_r168939210 and then add the styles to the demo along with a note about developers needing to style the `ng-messages` for chips? This note could be similar to the note in the Static Chips Demo.

IMM0rtalis:
@Splaktar I added the note and example-styles and conditional ng-show of error-messages in the demo.
While doing this I ran into problems with the dirty & touched-state not bubbled up from the input so I added a watcher. 
In general I figured out that no watcher was deregistered in the chipsController so I also added the destruction of some $watchers for onDestroy.

salimane:
@IMM0rtalis is there a way to use this branch in a project ? 

I tried without success : 

```json
"devDependencies": {
    "angular-material-source": "git+https://github.com/IMM0rtalis/material.git#c60f055728e1102afa7bb5e9d6e18df021000941"
}
```

IMM0rtalis:
@salimane not sure, but until the next version with this included is published you could try a hacky way (we had to do it for now)

```javascript
/**
 * Fix MdChipsController
 *  - fixes problem with not validating required-field (fix in v1.1.8)
 *
 * @see https://github.com/angular/material/pull/11125
 */
(function() {
    function updateMdChipsController(MdChipsCtrl) {

        /**
         * Updates the validity properties for the ngModel.
         */
            this.ngModelCtrl.$setValidity('md-max-chips', !this.hasMaxChipsReached());
            this.ngModelCtrl.$validate(); // rerun any registered validators
        };

        /**
         * Configures the required interactions with the ngModel Controller.
         * Specifically, set {@code this.items} to the {@code NgModelCtrl#$viewVale}.
         * @param ngModelCtrl
         */
        MdChipsCtrl.prototype.configureNgModel = function(ngModelCtrl) {
            this.ngModelCtrl = ngModelCtrl;

            var self = this;

            // in chips the meaning of $isEmpty changes
            ngModelCtrl.$isEmpty = function(value) {
                return !value || value.length === 0;
            };

            ngModelCtrl.$render = function() {
                // model is updated. do something.
                self.items = self.ngModelCtrl.$viewValue;
            };
        };
    }

    var invokes = angular.module('material.components.chips')._invokeQueue,
        i;

    for(i=0; i < invokes.length; ++i) {
        if(invokes[i][0] === "$controllerProvider" && invokes[i][2][0] === "MdChipsCtrl") {
            updateMdChipsController(invokes[i][2][1]);
        }
    }

})();
```

salimane:
thanks @IMM0rtalis .  what is the connection between the current repo and this one https://github.com/angular/bower-material .
do we need the same changes over there ?

Thanks

Splaktar:
@salimane bower-material is the repo where the releases live. This is what gets installed by bower or NPM. Everytime we merge something into master (including making a release), a new, installable build is generated in bower-material and available to Bower and NPM.

src/components/chips/js/chipsController.js
```diff
@@ -170,18 +177,40 @@ MdChipsCtrl.prototype.init = function() {
   var ctrl = this;
 
   // Set the wrapper ID
-  ctrl.wrapperId = '_md-chips-wrapper-' + ctrl.$mdUtil.nextUid();
+  this.wrapperId = '_md-chips-wrapper-' + this.$mdUtil.nextUid();
```
Splaktar: What's the benefit of the changes on this line?

salimane:
@Splaktar thanks for reply. 

Do you think there is way to have a branch in https://github.com/angular/bower-material that will use the changes in the current pull-request ? 

Thanks

src/components/chips/js/chipsController.js
```diff
@@ -170,18 +177,40 @@ MdChipsCtrl.prototype.init = function() {
...
+  this.deRegister.push(
+    this.$scope.$watchCollection('$mdChipsCtrl.items', function() {
+      // Make sure our input and wrapper have the correct ARIA attributes
+      ctrl.setupInputAria();
+      ctrl.setupWrapperAria();
+    })
+  );
+
+  this.deRegister.push(
+    this.$scope.$watch('$mdChipsCtrl.items', function() {
```
Splaktar: The comment here is the same as above. What's the reason that this new `$watch` is needed?

Splaktar:
@salimane not until we get this merged.

src/components/chips/js/chipsController.js
IMM0rtalis: Actually I do not see a point in using a helper-var instead of `this` if `this` is usable in a place. In my opinion it reduces the readability of code. And if the helper-var for encapsulated inner functions is called `ctrl`, `self` or `that` is in my opinion too much a matter of taste (otherwise I would have replaced `ctrl` with more widely used helpers in this repository)

src/components/chips/js/chipsController.js
IMM0rtalis: You are just pointing out the replacement of `ctrl` with `this`, right? Then see comment above^^

src/components/chips/demoValidation/script.js
IMM0rtalis: Is this fixed @Splaktar ? Because in the code it seems like, but just the one comment of you got hidden, this here not.

src/components/chips/js/chipsController.js
Splaktar: Ah OK. 👍

src/components/chips/js/chipsController.js
Splaktar: Nope, I'm asking about the two comments "Make sure our input and wrapper have the correct ARIA attributes" and the existing `$watchCollection` versus the new `$watch`.

IMM0rtalis: Just added the unwatcher-handling and replaced `ctrl` with `this`
screenshot_20180226_101737

IMM0rtalis: Wait... this is duplicated. Fix it asap

src/components/chips/demoValidation/script.js
Splaktar: Yep, looks good now.

Splaktar:
I asked @DevVersion to take one last look at this. Then I think that we can send it to presubmit.

src/components/chips/js/chipsController.js
```diff
@@ -170,18 +177,32 @@ MdChipsCtrl.prototype.init = function() {
...
+};
+
+/**
+ * Destructor for cleanup
+ */
+MdChipsCtrl.prototype.$onDestroy = function $onDestroy() {
+  var $destroyFn;
+  while (($destroyFn = this.deRegister.pop())) {
+    $destroyFn.call(this);
+  }
```
DevVersion: Why is this made so complicated? What about just `this.unregisterFunctions.forEach(fn => fn());`?

src/components/chips/js/chipsController.js
```diff
@@ -716,6 +744,43 @@ MdChipsCtrl.prototype.onInputBlur = function () {
...
+  var ngModelCtrl = inputElement.controller('ngModel');
+  var ctrl = this;
+
+  if (ngModelCtrl) {
+
+    // sync touched-state from inner input to chips-element
+    this.deRegister.push(
+      this.$scope.$watch(
+        function() {
+          return ngModelCtrl.$touched;
```
DevVersion: You can combine this single watch into two. e.g.

```js
this.$scope.$watch(() => {
  return { touched: inputNgModelCtrl.$touched, dirty: inputNgModelCtrl.$dirty };
}, XXXX);
```

src/components/chips/js/chipsController.js
IMM0rtalis: Because it is ES6, right? Is this repo require es6 ?

src/components/chips/js/chipsController.js
IMM0rtalis: But dont I need $watchCollection for this? And then separate afterwards anyway?

src/components/chips/js/chipsController.js
Splaktar: ES6 is fine in the build scripts and docs site, but not in the production library code (i.e. anything in `src/`).

DevVersion: Yeah, using the `ES5` `forEach` method should be still a benefit.

src/components/chips/js/chipsController.js
DevVersion: Don't think a `$watchCollection` will work here, because it requires us to watch properties on the `$scope`. Using `$watch` and just having a different watch condition should work.

https://stackoverflow.com/questions/17550947/how-to-watch-multiple-variables

IMM0rtalis: If I do as suggested I run in an endless digest-cycle with karma. What speaks against having to simple watcher?

Splaktar: Let's leave it as is, thanks for giving it a try.

IMM0rtalis:
@tinayuangao or @jelbourn or @andrewseguin or @josephperrott or @mmalerba 
When I understoof correct you are the ones who can/should merge it, right? - Since here is no activity for like 8-10 days I thought it maybe got lost between all the issues. Can sb pls merge it?

Splaktar:
@IMM0rtalis there is no need to mention all of the caretakers. The issue can't be merged until it passes internal presubmit testing.

@tinayuangao started the presubmits on this on 2/27, but I haven't received the results from that yet. I've pinged her for that information.

Thank you for the reminder.

Splaktar:
Presubmits failed. The failures are being investigated. I'll provide more details when they are available.

IMM0rtalis:
Ahh ok. I thought they are the ones who should merge it (https://github.com/angular/material/blob/master/docs/guides/COMMIT_LEVELS.md) and since there was no update...  But yeah, they are often not passed I have the feeling :D

Splaktar:
Yep, they will merge it when presubmits get sorted out and any internal failures are fixed (by them). They rotate shifts each week, so mentioning them all isn't ideal. I know who is on shift each week and I work with them to get the merges run through presubmit and merged. This process starts after I apply the `pr: merge ready` tag.

We recently updated many of the tag descriptions (new GitHub feature) to help make this process a bit more clear. I'll try to document a bit more of the process in the commit levels guide when I have some time.

jelbourn:
Passes presubmit

Splaktar:
This appears to have caused a regression related to adding chips onBlur. We're planning to revert it in https://github.com/angular/material/pull/11215 and do more investigation/testing.

PR #11127, opened 2018-02-18
fix(menu): close menu on tab

### LINKED ISSUE #11123: menu: should close when pressing tab
Currently, when you tab from an open menu it moves through the items and then escapes the panel into the rest of the page. The menu should actually just close upon pressing tab and the focus should move to the next focusable element (the menu items shouldn't be tab stops). `MatMenu` does the right thing for this and can be an example.

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Pressing tab inside of a menu panel (`md-menu-content`) can cause the focus to escape the panel and move to other parts of the page (with the menu panel still open).

Issue Number: 
Fixes #11123

## What is the new behavior?
Use only arrow keys to navigate menus (not tabs).
Resolve issue where tab can escape the panel by closing the menu when the `TAB` keyCode is detected.

```
[ ] Yes
[x] No
```

## Other information
N/A

### DIFF
--- a/src/components/menu/js/menuServiceProvider.js
+++ b/src/components/menu/js/menuServiceProvider.js
@@ -292,6 +292,12 @@ function MenuProvider($$interimElementProvider) {
               opts.mdMenuCtrl.close(false, { closeAll: true });
               handled = true;
               break;
+            case $mdConstant.KEY_CODE.TAB:
+              opts.mdMenuCtrl.close(false, { closeAll: true });
+              // Don't prevent default or stop propagation on this event as we want tab
+              // to move the focus to the next focusable element on the page.
+              handled = false;
+              break;
             case $mdConstant.KEY_CODE.UP_ARROW:
               if (!focusMenuItem(ev, opts.menuContentEl, opts, -1) && !opts.nestLevel) {
                 opts.mdMenuCtrl.triggerContainerProxy(ev);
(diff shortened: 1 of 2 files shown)


src/components/menu/js/menuServiceProvider.js
```diff
@@ -289,6 +289,7 @@ function MenuProvider($$interimElementProvider) {
           var handled;
           switch (ev.keyCode) {
             case $mdConstant.KEY_CODE.ESCAPE:
+            case $mdConstant.KEY_CODE.TAB:
               opts.mdMenuCtrl.close(false, { closeAll: true });
               handled = true;
```
crisbeto: Looking at the logic below, when an event is marked as `handled`, it'll prevent the default action. This shouldn't be the case `TAB`, because it'll prevent people from moving forward.

jelbourn: Yeah, good catch.

Splaktar:
Feedback addressed. Ready for presubmit.

src/components/menu/js/menuServiceProvider.js
```diff
@@ -292,6 +292,10 @@ function MenuProvider($$interimElementProvider) {
...
               break;
+            case $mdConstant.KEY_CODE.TAB:
+              opts.mdMenuCtrl.close(false, { closeAll: true });
+              handled = false;
```
jelbourn: Add comment that explains why `handled = false` here.

Splaktar: Done.

Issue #10082, opened 2016-12-02
mdMaxlength is overriding 'required' validator for blank spaces.


christrude:
**Actual Behavior**:
 -  When an input has the **mdMaxlength** directive and **required** attribute set, mdMaxlength forces **ng-trim** to be set to false. This allows the user to enter all space characters and the input will pass validation.
 -  /angular/material/blob/master/src/components/input/input.js - mdMaxlengthDirective - line 650:
```
      // Stop model from trimming. This makes it so whitespace
      // over the maxlength still counts as invalid.
      attr.$set('ngTrim', 'false'); 
```
   I kind of expect **mdMaxlength** to not care about ending whitespace to be part of it's calculation, especially if it gets trimmed anyways.

**Fiddle**
The following fiddle shows the issue caused with **ng-trim** added to a **required** input. Just add spaces to the input to see the **required** error go from true to false, when it should still be true
http://jsfiddle.net/ad4ghLf7/

**Angular Versions**: *
 -  Angular Version: 1.5.8
 -  Angular Material Version: 1.1.1

**Additional Information**:
 -  Browser Type: Chrome
 -  Browser Version: latest
 -  OS: macOS Sierra

christrude:
Did the work myself and created a pull request! https://github.com/angular/material/pull/10083

Splaktar:
The Fiddle here just shows how `ng-trim` works in AngularJS. But it doesn't demonstrate the issue described above with AngularJS Material and `md-maxlength`.

Splaktar:
The Errors demo in the Input demos of the docs reproduces this issue in the first input if you clear it and add spaces. Here's a matching [CodePen](https://codepen.io/Splaktar/pen/Jpebxy).

PR #11136, opened 2018-02-27
 feat(mdMaxlength): support use with required and ng-trim

Splaktar:
## PR Checklist

## PR Type
```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

When `md-maxlength` is used on an `input` that has the `required` attribute set, the required validation will accept blank spaces in the input as valid by default. Trying to override this with `ng-trim="true"` is ignored.

Issue Number: 
#10082

## What is the new behavior?
By default, blank spaces in required inputs cause an invalid state. If this is undesirable, it can be overridden by adding `ng-trim="false"` to the input.

```
[ ] Yes
[x] No
```

## Other information
This is based on an abandoned PR (https://github.com/angular/material/pull/10083).


Splaktar:
CLA consent here: https://github.com/angular/material/pull/10083#issuecomment-264527006

Splaktar:
I can't get the test to pass which is supposed to verify this functionality. In manual testing (demos), I can clearly see the functionality working as expected. But I can't get the required error to trigger in the test.

Splaktar:
The test issues have been resolved. This is ready for review.

src/components/input/input.js
```diff
@@ -679,16 +694,17 @@ function mdMaxlengthDirective($animate, $mdUtil) {
       });
     });
 
-    function renderCharCount(value) {
-      // If we have not been initialized or appended to the body yet; do not render
+    function renderCharCount() {
+      // If we have not been initialized or appended to the body yet; do not render.
       if (!charCountEl || !charCountEl.parent) {
```
gkalpak: Not directly related to this PR, but this check seems incorrect. `charCountEl.parent` will always be truthy if `charCountEl` is truthy. Maybe the intention was `!charCountEl.parent()`?

src/components/input/input.js
```diff
@@ -679,16 +694,17 @@ function mdMaxlengthDirective($animate, $mdUtil) {
...
+        return;
       }
 
+      var elementVal = element.val();
+      elementVal = ngTrim && angular.isString(elementVal) ? elementVal.trim() : elementVal;
       // Force the value into a string since it may be a number,
       // which does not have a length property.
-      charCountEl.text(String(element.val() || value || '').length + ' / ' + maxlength);
-      return value;
+      charCountEl.text(String(elementVal || '').length + ' / ' + maxlength);
```
gkalpak: If the intention is to have this working for numeric inputs as well (which might be the case, since [this comment][1] was removed), it will not work correctly for `0` :grin:

[1]: https://github.com/angular/material/pull/11136/files#diff-da06475ed96460fbcb944d6f1f31748aL193

src/components/input/input.js
```diff
@@ -679,16 +694,17 @@ function mdMaxlengthDirective($animate, $mdUtil) {
...
-      // If we have not been initialized or appended to the body yet; do not render
+    function renderCharCount() {
+      // If we have not been initialized or appended to the body yet; do not render.
       if (!charCountEl || !charCountEl.parent) {
-        return value;
+        return;
       }
 
+      var elementVal = element.val();
+      elementVal = ngTrim && angular.isString(elementVal) ? elementVal.trim() : elementVal;
```
gkalpak: Nit: AngularJS ignores `ngTrim` for inputs of type `password` (and ever trim the value). In such a case, there will be an inconsistency between the value reported by `mdMaxlength` and the `$modelValue` 😛

src/components/input/input.js
```diff
@@ -650,7 +651,21 @@ function mdMaxlengthDirective($animate, $mdUtil) {
       // Using the $validators for triggering the update works very well.
       renderCharCount();
 
-      return ( modelValue || element.val() || viewValue || '' ).length <= maxlength;
+      // Force the value into a string since it may be a number,
+      // which does not have a length property.
+      return String(modelValue || element.val() || viewValue || '').length <= maxlength;
```
gkalpak: 1. This will not work correctly if `element.val()` is the number `0`.

2. What about an empty `modelValue`? If the intention is to validate on the `viewValue` (which would be is a reasonable approach; this is what AngularJS moves towards), then why take a look at the `modelValue` at all.

3. What about `ngTrim`? The char count would take that into account (and display the correct count, but the validator might report something that does not match the char count.

src/components/input/input.js
```diff
@@ -650,7 +651,21 @@ function mdMaxlengthDirective($animate, $mdUtil) {
...
+    ngModelCtrl.$validators['required'] = function(modelValue, viewValue) {
+      if (!attr.required) {
+        return true;
+      }
+
+      var elementVal = modelValue || element.val() || viewValue || '';
+      elementVal = ngTrim && angular.isString(elementVal) ? elementVal.trim() : elementVal;
+      // Force the value into a string since it may be a number,
+      // which does not have a length property.
+      return angular.isDefined(elementVal) && String(elementVal).length;
```
gkalpak: Nit: At this point `elementVal` can't be undefined. At worse, it'll be the empty string :grin:

src/components/input/input.js
```diff
@@ -650,7 +651,21 @@ function mdMaxlengthDirective($animate, $mdUtil) {
       // Using the $validators for triggering the update works very well.
       renderCharCount();
 
-      return ( modelValue || element.val() || viewValue || '' ).length <= maxlength;
+      // Force the value into a string since it may be a number,
+      // which does not have a length property.
+      return String(modelValue || element.val() || viewValue || '').length <= maxlength;
+    };
+
+    ngModelCtrl.$validators['required'] = function(modelValue, viewValue) {
```
gkalpak: This means that the default (or a custom) `required` validator will be overwritten 😱
So, for example `ctrl.$isEmpty()` (which is a standard way to define the behavior for the `required` validator), will be ignored.

This should at least be mentioned as a limitation in the docs. But it feels a little weird 😕
Is it really necessary? Why do you even need this here?

src/components/input/input.js
```diff
@@ -650,7 +651,21 @@ function mdMaxlengthDirective($animate, $mdUtil) {
       // Using the $validators for triggering the update works very well.
       renderCharCount();
```
gkalpak: It might be preferrable to call [ngModelCtrl.$validate()][1] instead, since that will take care of updating input/form `$error`s, validation classes, etc

[1]: https://code.angularjs.org/1.6.9/docs/api/ng/type/ngModel.NgModelController#$validate

src/components/input/input.js
Splaktar: Ah yeah, I saw that but didn't imagine the case where a password input would be used with `mdMaxlength`. I think it makes sense to support that. Thanks!

src/components/input/input.js
Splaktar: I was trying to handle the `0` case, but I can see from your comments above that this isn't the place to do it. Thanks.

src/components/input/input.js
Splaktar: This was required to get the tests to pass. It wasn't required for the actual validation and use case to work in manual testing. I think that I'll take a look at your comment above about just using the viewValue to see if that's already trimmed as I would like to avoid this if possible.

Splaktar:
Thanks a ton @gkalpak! That's exactly the kind of feedback that I was looking for! I'll re-visit and make changes.

gkalpak:
:+1: Feel free to ping me for another review :smiley:

Splaktar:
@gkalpak can you please take another look? 

`viewValue` did not have trim already applied in my debugging. 
I believe that I was able to properly override the `ngModelCtrl.$isEmpty()` function instead of the required validator.
The last few checks that didn't work for `0` should be fixed.

src/components/input/input.js
```diff
@@ -650,7 +651,23 @@ function mdMaxlengthDirective($animate, $mdUtil) {
...
+      return String(elementVal).length <= maxlength;
+    };
+
+    ngModelCtrl.$isEmpty = function(value) {
+      var elementVal = value;
+      if (elementVal === undefined || elementVal === null) {
+        elementVal = '';
+      }
+      elementVal = ngTrim && !isPasswordInput && angular.isString(elementVal) ? elementVal.trim() : elementVal;
+      return String(elementVal).length === 0;
```
gkalpak: Nit: You can make a re-usable function for calculating the length of a value (which is used in the validator, `$isEmpty()` and `renderCharCount()`):

```js
function caclulateLength(value) {
  if (value == null) {
    value = '';
  } else {
    value = ngTrim && !isPasswordInput && angular.isString(value) ? value.trim() : value;
  }
  return String(value).length;
}
```

src/components/input/input.spec.js
```diff
@@ -393,6 +393,79 @@ describe('md-input-container directive', function() {
...
+      expect(pageScope.form.foo.$error['md-maxlength']).toBeFalsy();
+
+      pageScope.$apply('foo = "  "');
+      expect(input.hasClass('ng-invalid')).toBe(true);
+      expect(input.hasClass('ng-invalid-required')).toBe(true);
+      expect(pageScope.form.foo.$error['required']).toBeTruthy();
+      expect(pageScope.form.foo.$error['md-maxlength']).toBeFalsy();
+    });
+
+    it('should not trim spaces for required password inputs by default', function() {
```
gkalpak: Nit: This is not just "by default". It is the only posible behavior (i.e. it is not possible to override it).

src/components/input/input.spec.js
```diff
@@ -393,6 +393,79 @@ describe('md-input-container directive', function() {
...
+
+      // Flush any pending $mdUtil.nextTick calls
+      $timeout.flush();
+
+      expect(pageScope.form.foo.$error['required']).toBeTruthy();
+      expect(pageScope.form.foo.$error['md-maxlength']).toBeFalsy();
+
+      pageScope.$apply('foo = " "');
+      expect(pageScope.form.foo.$error['required']).toBeFalsy();
+    });
```
gkalpak: OOC, why not verify `pageScope.form.foo.$error['md-maxlength']` too? (Possibly using `foo = "  "` as well to make it truthy.)

src/components/input/input.js
Splaktar: I had a feeling that this was needed. Thanks for confirming it. Much cleaner 😃

PR #11149, opened 2018-03-04
docs(datepicker): follow Material Design guidelines for messages

feloy:
## PR Checklist
Please check that your PR fulfills the following requirements:

- [x] The commit message follows [our guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format)
- [n/a] Tests for the changes have been added (for bug fixes / features)
- [n/a] Docs have been added / updated (for bug fixes / features)

## PR Type

Documentation changes 

```
[ ] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[x] Documentation content changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: #11007 

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information

Closes #11007

PR #11150, opened 2018-03-04
fix(input): make md-maxlength validation happen on initialization

### LINKED ISSUE #10320: md-maxlength validation doesn't happen on initialization
**Actual Behavior**:
 -  `What is the issue? *` Using md-maxlength="MAX" on an input field that initializes to some value of length > "MAX", the md-maxlength validation does not happen, and the input field is considered valid.
 -  `What is the expected behavior?` The input field to be considered invalid on initialization.

 -  `CodePen Demo which shows your issue:` http://codepen.io/jessika/pen/MJvxKE
 -  `Details:` I created two input fields, one using md-maxlength for character limit validation, and one using ng-maxlength. Both have the same initial values and same character limit. You should be able to see that the field using md-maxlength is still valid and the field using ng-maxlength is marked invalid. I would expect that both fields be invalid.

**Angular Versions**: *
 -  `Angular Version:` 1.5.5
 -  `Angular Material Version:` 1.1.1, 1.1.0

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *` 56.0.2924.76 (64-bit)
 -  `OS: *` Ubuntu
 -  `Stack Traces:` N/A

----

  Add more lines if needed.

feloy:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

When an input has an md-maxlength attribute and an initial value longer than this maxlength, no error is triggered until the input value is changed.

Issue Number: #10320 

## What is the new behavior?

When an input has an md-maxlength attribute and an initial value longer than this maxlength, an error is triggered at initialization. 

Note that the error message is not shown at initialization but only when the focus leave the input (without content changes), which seems to be the correct behavior (in a similar way, for 'required' input with empty initial value, no message is displayed at initialization).

```
[ ] Yes
[x] No
```

## Other information

Closes #10320

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -634,10 +634,24 @@ function mdMaxlengthDirective($animate, $mdUtil) {
   };
 
   function postLink(scope, element, attr, ctrls) {
-    var maxlength;
+    var maxlength = parseInt(attr.mdMaxlength);
+    if (isNaN(maxlength)) maxlength = -1;
     var ngModelCtrl = ctrls[0];
     var containerCtrl = ctrls[1];
     var charCountEl, errorsSpacer;
+
+
+    ngModelCtrl.$validators['md-maxlength'] = function(modelValue, viewValue) {
+      if (!angular.isNumber(maxlength) || maxlength < 0) {
+        return true;
+      }
+
+      // We always update the char count, when the modelValue has changed.
+      // Using the $validators for triggering the update works very well.
+      renderCharCount();
+
+      return ( modelValue || element.val() || viewValue || '' ).length <= maxlength;
+    };
 
     // Wait until the next tick to ensure that the input has setup the errors spacer where we will
     // append our counter
@@ -663,23 +677,11 @@ function mdMaxlengthDirective($animate, $mdUtil) {
           $animate.leave(charCountEl);
         }
       });
-
-      ngModelCtrl.$validators['md-maxlength'] = function(modelValue, viewValue) {
-        if (!angular.isNumber(maxlength) || maxlength < 0) {
-          return true;
-        }
-
-        // We always update the char count, when the modelValue has changed.
-        // Using the $validators for triggering the update works very well.
-        renderCharCount();
-
-        return ( modelValue || element.val() || viewValue || '' ).length <= maxlength;
-      };
     });
 
     function renderCharCount(value) {
-      // If we have not been appended to the body yet; do not render
-      if (!charCountEl.parent) {
+      // If we have not been initialized or appended to the body yet; do not render
+      if (!charCountEl || !charCountEl.parent) {
         return value;
       }
 
--- a/src/components/input/input.spec.js
+++ b/src/components/input/input.spec.js
@@ -257,6 +257,40 @@ describe('md-input-container directive', function() {
       return angular.element(el[0].querySelector('.md-char-counter'));
     }
 
+    it('should error with a constant and incorrect initial value', function() {
+      var el = $compile(
+        '<form name="form">' +
+        '  <md-input-container>' +
+        '    <input md-maxlength="2" ng-model="foo" name="foo">' +
+        '  </md-input-container>' +
+        '</form>')(pageScope);
+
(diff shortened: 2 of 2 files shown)


Splaktar:
Same as https://github.com/angular/material/pull/11149#issuecomment-370247813, Please add "Closes #10320" to the footer of your commit message as detailed in the [commit message guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format). Just adding it the PR description is not enough as that won't trigger the issue to be closed on merge or show up in the changelog properly.

src/components/input/input.js
```diff
@@ -634,11 +634,25 @@ function mdMaxlengthDirective($animate, $mdUtil) {
   };
 
   function postLink(scope, element, attr, ctrls) {
-    var maxlength;
+    var maxlength = Number.parseInt(attr.mdMaxlength);
+    if (isNaN(maxlength)) maxlength = -1;
```
Splaktar: Should this use `!angular.isNumber(maxlength)` to be consistent with line 645, or is there a reason that `isNaN` better here?

src/components/input/input.js
```diff
@@ -663,23 +677,11 @@ function mdMaxlengthDirective($animate, $mdUtil) {
...
-        renderCharCount();
-
-        return ( modelValue || element.val() || viewValue || '' ).length <= maxlength;
-      };
     });
 
     function renderCharCount(value) {
       // If we have not been appended to the body yet; do not render
-      if (!charCountEl.parent) {
+      if (!charCountEl || !charCountEl.parent) {
```
Splaktar: Is `||` correct here? Should this be `&&`?

src/components/input/input.js
feloy: No, that's not equivalent here, because angular.isNumber(NaN) returns true but in this case, we do not want to keep NaN.

src/components/input/input.js
feloy: We now have to check that charCountEl is not undefined or charCountEl.text will fail in the following lines in some cases (because charCountEl is initialized at nextTick and renderCharCount can now be called before nextTick, from validator).

Splaktar: OK, can you please update the comment as well?

src/components/input/input.js
Splaktar: Ah yes, I thought that `angular.isNumber()` was smarter than it actually is.

Splaktar:
Btw, amending commits is the best way to keep your changes in a single commit when updates are needed. Then force push to your branch is the best way to sync those updates to the branch.

Splaktar:
Looks great! Thank you for your patience. I've sent this over to the caretaker to have presubmit tests run.

Splaktar:
This is currently triggering some presubmit failures that appear to be flakes (i.e. from flaky tests, not actual real failures). This is being looked at internally by Google. No work is needed on your side at this time.

Splaktar:
I re-ran the failing build steps in Travis and everything is green now 😃

jelbourn:
This fails google presubmit because `Number.parseInt` is not supported in IE11

src/components/input/input.js
```diff
@@ -634,11 +634,25 @@ function mdMaxlengthDirective($animate, $mdUtil) {
   };
 
   function postLink(scope, element, attr, ctrls) {
-    var maxlength;
+    var maxlength = Number.parseInt(attr.mdMaxlength);
```
Splaktar: As pointed out by Jeremy, `Number.parseInt` is not supported in IE11. So we'll need to use something different here.

It looks like the global [parseInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/parseInt) should work?
MDN also points out that [Number.parseInt](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/parseInt) has no IE support.

feloy: I've used the global parseInt instead

Issue #11108, opened 2018-02-06
select(trackBy): preset object triggers ngChange


zendu:
This issue is very similar to what is mentioned in issue #4118.

I have mdSelect with ngModel which is already populated and ngModelOptions using trackBy option. mdSelect has ngChange handler to trigger some more model updates when user really changes selection.
Options are rendered using ngRepeat and use ngValue to select the object.

Problem: ngChange on mdSelect is triggered just by the preset object bound to the model.

https://codepen.io/zendu/pen/VQjBzG

AngularJS: v1.6.4
Material:   v1.1.6

free-easy:
I'm facing the same problem.

In my case I have ngModel which could be already populated, whereas the array of ngRepeat is loaded only when `md-select` opens.

In order to show the populated value, I insert it into the array of ngRepeat so that one `<md-option` gets created and obviously ngModel matches with this option, so the value of ngModel is shown on `md-select`.

The problems is raised when I open `md-select`. Data is loaded from server and in the meanwhile I've cleaned the array of ngRepeat from the fake item I inserted before.

Now, I've set `ng-model-options="{ "trackBy": "$value._id" }` (yeah, I'm using mongodb) and among the loaded items there might be an object with the same `_id` of the one in ngModel.
I have the match again, but the two objects might be little different. Once the match is established the value of ngModel is set to the values of the matched object. **They're conceptually the same, but e.g. in my case there are different $$hashKey properties added to both of them. So ng-change is triggered** 

I think the problem is here.
https://github.com/angular/material/blob/a4cc1869e63c6fb80e78a0c875329878cb3e34db/src/components/select/select.js#L896

When using trackBy, instead of using `angular.equals` on the whole objects, a comparison between the property specified by trackBy should be performed.

Splaktar:
Also of interest is that removing ` ng-model-options="{trackBy: '$value.value'}"` from the CodePen in the OP causes the problem to go away. So this behavior is correct until `trackBy` is used.

@free-easy thank you for looking into this further.

We have a [test for this](https://github.com/angular/material/blob/master/src/components/select/select.spec.js#L152-L166) but I think that it's currently giving a false positive. This is because our [other tests](https://github.com/angular/material/blob/master/src/components/select/select.spec.js#L767-L784) for select and `ng-change` were failing and got disabled a few years ago.

Is anyone interested in fixing these tests, the `usingTrackBy` equality check, and opening a PR?

free-easy:
[...]

PR #11156, opened 2018-03-10
fix(select): when using trackBy, trigger ng-change only when tracked property is different

free-easy:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

When using `ng-model-options="{ 'trackBy': '$value.someProperty' }"` on md-select, `ng-change` is triggered when the selected option's object is not equal to the one of the model (the comparison is done with `angular.equals` which compares recursively all the properties of the object).  
Issue Number: #11108

## What is the new behavior?
`ng-change` is triggered when the tracked property of the model is different from the selected option.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/select/demoTrackBy/index.html
+++ b/src/components/select/demoTrackBy/index.html
@@ -1,0 +1,52 @@
+<div layout="column" layout-align="center center" class="md-padding" ng-cloak>
+  <div layout="row" layout-align="space-between">
+    <div ng-controller="AppCtrl as ctrl" layout="column" flex="40">
+      <div>
+        <h1 class="md-title">Without trackBy</h1>
+        <div layout="row">
+          <md-input-container>
+            <label>Items</label>
+            <md-select ng-model="ctrl.selectedItem"
+              ng-change="ctrl.modelHasChanged = true">
+              <md-option ng-repeat="item in ctrl.items" ng-value="item">
+                {{ item.name }}
+              </md-option>
+            </md-select>
+          </md-input-container>
+        </div>
+      </div>
+      <div layout="column">
+        <h5>Initial model</h5>
+        <code><pre>{{ ::ctrl.selectedItem | json }}</pre></code>
+        <h5>Current model</h5>
+        <code><pre>{{ ctrl.selectedItem | json }}</pre></code>
+        <span ng-show="ctrl.modelHasChanged">Model has changed</span>
+      </div>
+    </div>
+
+    <div ng-controller="AppCtrl as ctrl" layout="column" flex="40">
+      <div>
+        <h1 class="md-title">With trackBy</h1>
+        <div layout="row">
+          <md-input-container>
+            <label>Items</label>
+            <md-select ng-model="ctrl.selectedItem"
+              ng-change="ctrl.modelHasChanged = true"
+              ng-model-options="{ trackBy: '$value.id' }">
+              <md-option ng-repeat="item in ctrl.items" ng-value="item">
+                {{ item.name }}
+              </md-option>
+            </md-select>
+          </md-input-container>
+        </div>
+      </div>
+      <div layout="column">
+        <h5>Initial model</h5>
+        <code><pre>{{ ::ctrl.selectedItem | json }}</pre></code>
+        <h5>Current model</h5>
+        <code><pre>{{ ctrl.selectedItem | json }}</pre></code>
+        <span ng-show="ctrl.modelHasChanged">Model has changed</span>
+      </div>
+    </div>
+  </div>
+</div>
--- a/src/components/select/demoTrackBy/script.js
+++ b/src/components/select/demoTrackBy/script.js
@@ -1,0 +1,27 @@
+(function() {
+  'use strict';
+  angular
+    .module('selectDemoTrackBy', ['ngMaterial', 'ngMessages'])
+    .controller('AppCtrl', function() {
+      this.selectedItem = {
+        id: '5a61e00',
+        name: 'Bob',
+        randomAddedProperty: 123
+      };
+
+      this.items = [
+        {
+          id: '5a61e00',
+          name: 'Bob',
+        },
+        {
+          id: '5a61e01',
+          name: 'Max',
+        },
+        {
+          id: '5a61e02',
+          name: 'Alice',
+        },
+      ];
+    });
+})();
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -259,12 +259,13 @@ function SelectDirective($mdSelect, $mdUtil, $mdConstant, $mdTheming, $mdAria, $
 
     // Use everything that's left inside element.contents() as the contents of the menu
     var multipleContent = isMultiple ? 'multiple' : '';
+    var ngModelOptions = attr.ngModelOptions ? $mdUtil.supplant('ng-model-options="{0}"', [attr.ngModelOptions]) : '';
     var selectTemplate = '' +
       '<div class="md-select-menu-container" aria-hidden="true" role="presentation">' +
-      '<md-select-menu role="presentation" {0}>{1}</md-select-menu>' +
+      '<md-select-menu role="presentation" {0} {1}>{2}</md-select-menu>' +
       '</div>';
 
-    selectTemplate = $mdUtil.supplant(selectTemplate, [multipleContent, element.html()]);
+    selectTemplate = $mdUtil.supplant(selectTemplate, [multipleContent, ngModelOptions,  element.html()]);
     element.empty().append(valueEl);
     element.append(selectTemplate);
 
@@ -764,28 +765,40 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
         return !self.options[self.hashGetter($viewValue)];
       };
 
-      // Allow users to provide `ng-model="foo" ng-model-options="{trackBy: 'foo.id'}"` so
+      // Allow users to provide `ng-model="foo" ng-model-options="{trackBy: '$value.id'}"` so
       // that we can properly compare objects set on the model to the available options
-      var trackByOption = $mdUtil.getModelOption(ngModel, 'trackBy');
+      //
+      // If the user doesn't provide a trackBy, we automatically generate an id for every
+      // value passed in with the getId function
+      if ($attrs.ngModelOptions) {
+        self.hashGetter = function(value) {
+          var ngModelOptions = $parse($attrs.ngModelOptions)($scope);
+          var trackByOption = ngModelOptions && ngModelOptions.trackBy;
 
-      if (trackByOption) {
-        var trackByLocals = {};
-        var trackByParsed = $parse(trackByOption);
-        self.hashGetter = function(value, valueScope) {
-          trackByLocals.$value = value;
-          return trackByParsed(valueScope || $scope, trackByLocals);
-        };
-        // If the user doesn't provide a trackBy, we automatically generate an id for every
-        // value passed in
-      } else {
-        self.hashGetter = function getHashValue(value) {
-          if (angular.isObject(value)) {
-            return 'object_' + (value.$$mdSelectId || (value.$$mdSelectId = ++selectNextId));
+          if (trackByOption) {
+            return $parse(trackByOption)($scope, { $value: value });
+          } else if (angular.isObject(value)) {
+            return getId(value);
           }
           return value;
         };
+      } else {
+        self.hashGetter = getId;
       }
       self.setMultiple(self.isMultiple);
+
+      /**
+       * If the value is an object, get the unique, incremental id of the value.
+       * If it's not an object, the value will be converted to a string and then returned.
+       * @param value
+       * @returns {string}
+       */
+      function getId(value) {
+        if (angular.isObject(value) && !angular.isArray(value)) {
+          return 'object_' + (value.$$mdSelectId || (value.$$mdSelectId = ++selectNextId));
+        }
+        return value + '';
+      }
     };
 
     self.selectedLabels = function(opts) {
@@ -888,14 +901,40 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
           values.push(self.selected[hashKey]);
         }
       }
-      var usingTrackBy = $mdUtil.getModelOption(self.ngModel, 'trackBy');
 
       var newVal = self.isMultiple ? values : values[0];
       var prevVal = self.ngModel.$modelValue;
 
-      if (usingTrackBy ? !angular.equals(prevVal, newVal) : (prevVal + '') !== newVal) {
+      if (!equals(prevVal, newVal)) {
         self.ngModel.$setViewValue(newVal);
         self.ngModel.$render();
+      }
+
+      function equals(prevVal, newVal) {
+        if (self.isMultiple) {
+          if (!angular.isArray(prevVal)) {
+            // newVal is always an array when self.isMultiple is true
+            // thus, if prevVal is not an array they are different
+            return false;
+          } else if (prevVal.length !== newVal.length) {
+            // they are different if they have different length
+            return false;
+          } else {
+            // if they have the same length, then they are different
+            // if an item in the newVal array can't be found in the prevVal
+            var prevValHashes = prevVal.map(function(prevValItem) {
+              return self.hashGetter(prevValItem);
+            });
+            return newVal.every(function(newValItem) {
+              var newValItemHash = self.hashGetter(newValItem);
+              return prevValHashes.some(function(prevValHash) {
+                return prevValHash === newValItemHash;
+              });
+            });
+          }
+        } else {
+          return self.hashGetter(prevVal) === self.hashGetter(newVal);
+        }
       }
     };
 
--- a/src/components/select/select.spec.js
+++ b/src/components/select/select.spec.js
@@ -152,10 +152,46 @@ describe('<md-select>', function() {
     it('should not trigger ng-change without a change when using trackBy', function() {
       var changed = false;
       $rootScope.onChange = function() { changed = true; };
-      $rootScope.val = { id: 1, name: 'Bob' };
+
+      // Since we're tracking by id, ng-change shouldn't be triggered
+      // when we have two objects that are not strictly equivalent (one has a 'randomAddedProperty')
+      // but that have the same tracked field
+      $rootScope.val = { id: 1, name: 'Bob', randomAddedProperty: 'random' };
 
       var opts = [ { id: 1, name: 'Bob' }, { id: 2, name: 'Alice' } ];
       var select = setupSelect('ng-model="$root.val" ng-change="onChange()" ng-model-options="{trackBy: \'$value.id\'}"', opts);
+      expect(changed).toBe(false);
+
+      openSelect(select);
+      clickOption(select, 1);
+      $material.flushInterimElement();
+      expect($rootScope.val.id).toBe(2);
+      expect(changed).toBe(true);
+    });
+
+    it('should support trackBy to be updated', function() {
+      var changed = false;
+      $rootScope.onChange = function() { changed = true; };
+      $rootScope.useTrackBy = false;
+      $rootScope.trackByOption = '$value.id';
+
+      var opts = [ { id: 1, name: 'Bob' }, { id: 2, name: 'Alice' } ];
+      $rootScope.val = opts[0];
+      var select = setupSelect('ng-model="$root.val"' +
+        'ng-change="onChange()"' +
+        'ng-model-options="{ trackBy: $root.useTrackBy ? $root.trackByOption : undefined }"', opts);
+      expect(changed).toBe(false);
+
+      $rootScope.$apply(function() {
+        $rootScope.useTrackBy = true;
+        // Since we're tracking by id, ng-change shouldn't be triggered
+        // when we have two objects that are not strictly equivalent (one has a 'randomAddedProperty')
+        // but that have the same tracked field
+        $rootScope.val = { id: 1, name: 'Bob', randomAddedProperty: 'random' };
+      });
+      openSelect(select);
+      clickOption(select, 0);
+      $material.flushInterimElement();
       expect(changed).toBe(false);
 
       openSelect(select);
--- a/src/components/select/demoTrackBy/style.css
+++ b/src/components/select/demoTrackBy/style.css
@@ -1,0 +1,4 @@
+code {
+  display: block;
+  padding: 8px;
+} 


src/components/select/select.js
```diff
@@ -893,10 +893,33 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
...
+          if (!angular.isArray(prevVal)) {
+            // newVal is always an array when self.isMultiple is true
+            // thus, if prevVal is not an array they are different
+            return false;
+          } else if (prevVal.length !== newVal.length) {
+            // they are different if they have different length
+            return false;
+          } else {
+            // if they have the same length, then they are different
+            // if a item in the newVal array can't be found in the prevVal
```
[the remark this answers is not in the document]
free-easy: I've fixed

free-easy:
@Splaktar Do you have any clue on why is the travis ci build failing?

Splaktar:
TravisCI had a significant outage last night around that time. I've restarted the tests.

Splaktar:
I need to improve the commit message guidelines, sorry for the confusion. I'll try to add some examples to them.

I believe that the footer needs to use the format `Closes #11108` and not `Closes issue #11108`. This is how the feature of GitHub works.

free-easy:
I've changed the commit message.

But the travis ci build seems to be still failing.

Splaktar:
It also looks like there is an issue with the AngularJS snapshot builds. I'll check with the team.

Splaktar:
It's possible that their snapshot somehow got corrupted by the TravisCI outage. I'll let you know when I find out more.

free-easy:
Which would be the best way to integrate this fix into my project before the next version of angular material is released?

Splaktar:
Once we get it merged, there will be a build with a hash on it that you will be able to install directly via NPM.

src/components/select/demoTrackBy/index.html
```diff
@@ -0,0 +1,59 @@
...
+              <md-option ng-value="item" ng-repeat="item in items">
+                {{ item.name }}
+              </md-option>
+            </md-select>
+          </md-input-container>
+        </div>
+      </div>
+      <div layout="column">
+        <h5>Initial model</h5>
+        <code style="display: block; padding: 8px;">{{ ::selectedItem | json}}</code>
```
gkalpak: Missing space before `}}` :grin:
(Here and below.)

src/components/select/select.js
```diff
@@ -893,10 +893,33 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
...
+            // thus, if prevVal is not an array they are different
+            return false;
+          } else if (prevVal.length !== newVal.length) {
+            // they are different if they have different length
+            return false;
+          } else {
+            // if they have the same length, then they are different
+            // if an item in the newVal array can't be found in the prevVal
+            return newVal.every(function(newValItem) {
+              return !!prevVal.find(function(prevValItem) {
```
gkalpak: Not sure what the supported browsers are, but make sure they all implement `.find()` (IE does not for example).

src/components/select/select.js
```diff
@@ -893,10 +893,33 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
...
+                return self.hashGetter(prevValItem) === self.hashGetter(newValItem);
```
gkalpak: This could be made a little more efficient.
Now, you compute hashes `O(prevVal.length * newVal.length)`, while you only need to compute each hash once (`O(prevVal.length + newVal.length)`).

src/components/select/select.js
```diff
@@ -893,10 +893,33 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
...
+            // newVal is always an array when self.isMultiple is true
+            // thus, if prevVal is not an array they are different
+            return false;
+          } else if (prevVal.length !== newVal.length) {
+            // they are different if they have different length
+            return false;
+          } else {
+            // if they have the same length, then they are different
+            // if an item in the newVal array can't be found in the prevVal
+            return newVal.every(function(newValItem) {
```
gkalpak: Just wanted to point out that this assumes that there are no items in `prevVal` with the same hash.
(This is probably fine, because the `trackBy` expression is supposed to _uniquely_ identify each item.)

src/components/select/select.js
free-easy: Gosh! I've always thought `find` was supported since IE 9...

free-easy:
@gkalpak I should have fixed all.

src/components/select/demoTrackBy/index.html
```diff
@@ -0,0 +1,61 @@
...
+              <md-option ng-value="item" ng-repeat="item in items">
+                {{ item.name }}
+              </md-option>
+            </md-select>
+          </md-input-container>
+        </div>
+      </div>
+      <div layout="column">
+        <h5>Initial model</h5>
+        <code style="display: block; padding: 8px;"><pre>{{ ::selectedItem | json:2 }}</pre></code>
```
gkalpak: You shouldn't need the `2`, since it is the default.

src/components/select/select.spec.js
```diff
@@ -152,7 +152,11 @@ describe('<md-select>', function() {
     it('should not trigger ng-change without a change when using trackBy', function() {
       var changed = false;
       $rootScope.onChange = function() { changed = true; };
-      $rootScope.val = { id: 1, name: 'Bob' };
+
+      // since we're tracking by id, ng-change shouldn't be triggered
+      // when we have two objects that are not strictly equivalent (one has a "randomAddedProperty")
+      // but that have the same id
+      $rootScope.val = { id: 1, name: 'Bob', randomAddedProperty: "random" };
```
gkalpak: Nit: Inconsistent quoting (`"random"` --> `'random'`) 😇

free-easy:
@gkalpak With this push all the quoting inconsistencies should be fixed. At least I hope so. :joy:

mmalerba:
There appears to be 1 project with some failures due to this PR

free-easy:
Can you please give me more information?

mmalerba:
Will need to investigate further, its not clear at a glance why it's failing. Could just be that their tests need to be fixed.

Splaktar:
Andrew looked into these presubmit failures this week, but got stuck when debugging an internal app. He's going to have @josephperrott continue the investigation next week.

free-easy:
Thank you for your efforts.

A possible problem that I'm aware of is that with this fix, if the tracked property of the current model does match with one of a md-option the value of the model won't be set to the one of the md-option. I consider this a feature, since when using trackBy, even if only the tracked property matches, the two objects are the same . But maybe this might lead to some bug to your internal app.

Splaktar:
That might be part of it. The other info that I got was "a select value is not being set for some reason, which causes the failures".

src/components/select/select.js
```diff
@@ -893,10 +893,37 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
       var newVal = self.isMultiple ? values : values[0];
       var prevVal = self.ngModel.$modelValue;
 
-      if (usingTrackBy ? !angular.equals(prevVal, newVal) : (prevVal + '') !== newVal) {
+      if (usingTrackBy ? !equals(prevVal, newVal) : (prevVal + '') !== newVal) {
```
josephperrott: Because the new `equals` function relies on `self.hashGetter`, the `self.hashGetter` method needs to be redefined (potentially), depending on if `usingTrackBy` has changed value.


For instance, if on initialization, `$mdUtil.getModelOption(self.ngModel, 'trackBy')` evaluates to `false` then later when running `self.refreshViewValue`, `$mdUtil.getModelOption(self.ngModel, 'trackBy')` evaluates to `true`, the `self.hashGetter` function that is called is not the one that is expected.


This was not an issue in the past, as the `angular.equals` does not rely on the `hashGetter` method that is defined.  This issue could also come to light in any other usage of `self.hasGetter` throughout the object.

gkalpak: Ooops! If changing `trackBy` after initialization is supported, it will get a little messy :grin:
The last `usingTrackBy` value should be kept track of and checked everywhere before using `self.hashGetter()` (afaict).

Splaktar: I guess that we also need to add a test for changing `trackBy` on `md-select`. I never imagined a use case where you would want to change `trackBy` over time.

free-easy: Something like this would make sense?

```js
self.hashGetter = function(value) {
  var trackByOption = $mdUtil.getModelOption(ngModel, 'trackBy');

  if (trackByOption) {
    return $parse(trackByOption)($scope, { $value: value });

    // If the user doesn't provide a trackBy, we automatically generate an id for every
    // value passed in
  } else if (angular.isObject(value)) {
      return 'object_' + (value.$$mdSelectId || (value.$$mdSelectId = ++selectNextId));
  }
  return value;
};
```
Basically I evaluate `ngModelOption.trackBy` each time.

Clearly it's not so performant as it was.

Splaktar: I was trying to figure out if there is a way to watch/observe changes to `ng-model-options` and I found this:

From the AngularJS docs:
> The ngModelOptions expression is only evaluated once when the directive is linked; it is not watched for changes. However, it is possible to override the options on a single ngModel.NgModelController instance with [NgModelController#$overrideModelOptions](https://docs.angularjs.org/api/ng/type/ngModel.NgModelController#$overrideModelOptions).

The linked docs specify
> Note: this function only affects the options set on the `ngModelController`, and not the options on the `ngModelOptions` directive from which they might have been obtained initially.

That should be OK if we're watching the controller (as we are now) but it means that watching the `ng-model-options` attribute directive for changes would be a bad idea.

@josephperrott can you confirm that the failures are being caused by calls to `NgModelController#$overrideModelOptions` in that app?

@gkalpak do you see any more performant alternatives than the suggestion above?

gkalpak: No, I don't see any great alternative. I would still check `trackBy` outside of `hashGetter`, because that gets called multiple times inside certain functions. I would check it at the top of functions that call `hashGetter()`.

Note that AngularJS generally does not check for changes in `modelOptions`, but `trackBy` is an `ngMaterial` specific option (and there is nothing stopping `ngMaterial` from responding to `modelOptions` changes anyway.

BTW, I doubt `$overrideModelOptions()` is relevant here.

josephperrott: The failures I have found do something to this effect:

```html
<md-select ng-model="modelName ng-model-options="{
    trackBy: 'ctrl.hasTracker() ? ctrl.tracker() : undefined'}">
  <md-option ng-value="something" ng-repeat="something in ctrl.somethings">
    {{something}}
  </md-option>
</md-select>
```

As @gkalpak noted, `$overrideModelOptions()` isn't really in play here.

Splaktar: @free-easy OK, let's try your approach in https://github.com/angular/material/pull/11156#discussion_r177416656 and see if it gets through presubmit testing. I agree that things get a bit too messy when having to scatter calls around different functions.

If it becomes a performance problem that users notice, we'll drop support for changing `trackBy` after initialization in 1.2.0.

free-easy: @Splaktar It seems that it doesn't work anyway since angular.js doesn't update ngModelOptions.

```js
ng-model-options="{ trackBy: 'ctrl.hasTracker() ? ctrl.tracker() : undefined' }"
```
The ternary expression seems to be evaluated only once.

Splaktar: @free-easy are you saying that `$mdUtil.getModelOption(self.ngModel, 'trackBy')` mentioned in https://github.com/angular/material/pull/11156#discussion_r177183110 can't currently change after initialization as described there?

free-easy: Exactly

free-easy: Exactly

free-easy:
@Splaktar 

Could you please review the last commit?
 
The last commit should support `ngModelOptions` to be updated. I've worked around the problem of `ngModelOptions` not updating by parsing (with $parse) the expression instead of just calling `$mdUtil.getModelOption()`.

There is a test case failing due to my modifications, though.
Before
```js
if (usingTrackBy ? !equals(prevVal, newVal) : (prevVal + '') !== newVal) {
```
With this commit
```js
if (!equals(prevVal, newVal))
```
I did it because, since `trackBy` availability is handled by `self.hashGetter`, it's redudant
to do another check in the `self.refreshViewValue` function. And even when not using trackBy  `equals` should be enough, since it uses `self.hashGetter` internally and if values are not objects `self.hashGetter` will just return them.

The removal of  `(prevVal + '') !== newVal` causes the test to fail. But I can't figure out the sense of this line.
Before this pull request it was like
```js
if (usingTrackBy ? !angular.equals(prevVal, newVal) : (prevVal + '') !== newVal) {
```
which means "If not using trackBy, compare the values, whathever they are, by first converting the first to string, which could lead to a object being first converted to `"[object Object]"` and this doesn't make sense.

free-easy:
The failing test is this
https://github.com/angular/material/blob/aaf75cc84250fabbdd4fd790075f733c972ec1fb/src/components/select/select.spec.js#L840

What should `should not override the initial model value` means?

Splaktar:
The failure is `Expected '2' to be 2, 'Expected value not to have been overwritten.'.` So it set the initial model to `2`, but after initialization, it was overridden to `'2'`. I.e. a `string`.

The code was basically expecting to run in two different modes:
1. If using `trackBy` then assume the value is an `object`
2. If not using `trackBy` then assume the value is a `string`

By not converting the `self.ngModel.$modelValue` to a `string` for the comparison, it is causing the view value to be updated. Previously, it converted the model value to a `string` and then it was `===` to the view value and the view value update wasn't triggered.

You could probably keep that equals check the same if you did
```javascript
var prevVal = usingTrackBy ? self.ngModel.$modelValue : self.ngModel.$modelValue + '';
```
But it looks like `usingTrackBy` is no longer defined in this scope.

As long as you convert the view value (prevVal) to a string before doing an equality check when not using `trackBy` then the test will pass. I have some ideas, but I'll let you take a look at this and see what you come up with.

free-easy:
@Splaktar I've squashed.

src/components/select/demoTrackBy/index.html
```diff
@@ -0,0 +1,61 @@
+<div layout="column" layout-align="center center" class="md-padding" ng-cloak>
+  <div layout="row" layout-align="space-between">
+    <div ng-controller="AppCtrl" layout="column" flex="40">
```
Splaktar: Can you use Controller as here like in the demoBasicUsage?

src/components/select/demoTrackBy/index.html
```diff
@@ -0,0 +1,61 @@
...
+      <div layout="column">
+        <h5>Initial model</h5>
+        <code style="display: block; padding: 8px;"><pre>{{ ::selectedItem | json }}</pre></code>
+        <h5>Current model</h5>
+        <code style="display: block; padding: 8px;"><pre>{{ selectedItem | json }}</pre></code>
+        <span ng-show="modelHasChanged">Model has changed</span>
+      </div>
+    </div>
+
+    <div ng-controller="AppCtrl" layout="column" flex="40">
```
Splaktar: Can you use Controller as here like in the demoBasicUsage?

src/components/select/demoTrackBy/script.js
```diff
@@ -0,0 +1,24 @@
...
+.controller('AppCtrl', function($scope) {
+  $scope.selectedItem = {
+    id: '5a61e00595764e18abc2ff01',
+    name: 'Bob',
+    randomAddedProperty: 123
+  };
+
+  $scope.items = [
+    {
+      id: '5a61e00595764e18abc2ff01',
```
Splaktar: Since this is just a demo, can we shorten these `id`s to 7 characters?

src/components/select/select.spec.js
```diff
@@ -11,6 +11,7 @@ describe('<md-select>', function() {
       $compile = $injector.get('$compile');
       $timeout = $injector.get('$timeout');
       $material = $injector.get('$material');
+      $mdUtil = $injector.get('$mdUtil');
```
Splaktar: This doesn't appear to be used anymore? Can you please remove it.

src/components/select/select.spec.js
```diff
@@ -152,7 +153,11 @@ describe('<md-select>', function() {
     it('should not trigger ng-change without a change when using trackBy', function() {
       var changed = false;
       $rootScope.onChange = function() { changed = true; };
-      $rootScope.val = { id: 1, name: 'Bob' };
+
+      // since we're tracking by id, ng-change shouldn't be triggered
```
Splaktar: Start with a capital please. 😄

src/components/select/select.spec.js
```diff
@@ -152,7 +153,11 @@ describe('<md-select>', function() {
...
+      // when we have two objects that are not strictly equivalent (one has a 'randomAddedProperty')
+      // but that have the same id
```
Splaktar: Let's go with "but that have the same tracked field." Since developers may not always be tracking by the `id` field.

src/components/select/select.spec.js
```diff
@@ -165,6 +170,38 @@ describe('<md-select>', function() {
...
+      var select = setupSelect('ng-model="$root.val"' +
+        'ng-change="onChange()"' +
+        'ng-model-options="{ trackBy: $root.useTrackBy ? $root.trackByOption : undefined }"', opts);
+      expect(changed).toBe(false);
+
+      $rootScope.$apply(function() {
+        $rootScope.useTrackBy = true;
+        // since we're tracking by id, ng-change shouldn't be triggered
+        // when we have two objects that are not strictly equivalent (one has a 'randomAddedProperty')
+        // but that have the same id
```
Splaktar: Same thing as above 😃

src/components/select/select.js
```diff
@@ -764,28 +765,40 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
...
-        self.hashGetter = function getHashValue(value) {
-          if (angular.isObject(value)) {
-            return 'object_' + (value.$$mdSelectId || (value.$$mdSelectId = ++selectNextId));
+      //
+      // If the user doesn't provide a trackBy, we automatically generate an id for every
+      // value passed in with the getId function
+      if ($attrs.ngModelOptions) {
+        self.hashGetter = function(value) {
+          // Allow users to provide `ng-model="foo" ng-model-options="{trackBy: '$value.id'}"` so
+          // that we can properly compare objects set on the model to the available options
```
Splaktar: Since this comment is a duplicate of the one a few lines above, can we remove it?

src/components/select/select.js
```diff
@@ -764,28 +765,40 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
...
         };
+      } else {
+        self.hashGetter = getId;
       }
       self.setMultiple(self.isMultiple);
+
+      /*
+       * Get the unique, incremental id of the value, if it's an object.
+       * If it's not an object, the value will be converted to string and then returned
+       */
```
Splaktar: Let's update this comment and add some JSDoc here:
```javascript
      /**
       * If the value is an object, get the unique, incremental id of the value.
       * If it's not an object, the value will be converted to a string and then returned.
       * @param value
       * @returns {string}
       */
```

free-easy:
@Splaktar All the requested changes should have been addressed.

src/components/select/demoTrackBy/script.js
```diff
@@ -0,0 +1,23 @@
+angular.module('selectDemoTrackBy', ['ngMaterial', 'ngMessages'])
```
Splaktar: Sorry, I forgot to add this to the last review, can you please wrap this in an IIFE like the demoBasicUsage?

free-easy:
@Splaktar Sorry for my oversight. It should be ok now.

mmalerba:
I'm seeing several unit test failures now. It looks like maybe that have a component with some `matchType` property (wired up to a `md-select`) and then in there unit tests they're doing:

```
ctrl.matchType = ...;
scope.$apply();
```

and finding that `matchType` isn't what they expect after. Maybe the `md-select` is overwriting it during the `scope.$apply`?

free-easy:
Where is `ctrl.matchType` used?
Is it used as value for `trackBy`?
Can you give me the md-select HTML too?

mmalerba:
The html looks like this:

```html
<md-input-container ng-if="...">
    <md-select ng-model="$ctrl.matchType"
               ng-change="$ctrl.setMatchType()"
               required>
      <md-option ng-repeat="option in ..."
                 ng-value="::option"
                 ng-selected="option.value == $ctrl.matchType.value && option.complement == $ctrl.matchType.complement"
                 md-no-ink>
        {{::option.name}}
      </md-option>
    </md-select>
</md-input-container>
```

free-easy:
That's really strange,
since all the unit tests for md-select are passing.

Splaktar:
Is there any chance that we can get a CodePen reproduction of this presubmit failure?

mmalerba:
I don't think so, it will have to be debugged by a Googler

Splaktar:
It looks like `$ctrl.matchType` is an `object` and not a `string`.
Is it just `{value: string, complement: string}`?

@free-easy since they aren't using `trackBy`, perhaps there was a change in how the default hashing is done and that could be causing the model to not be set with the same value as before this change? Can you please check to see if we have a unit test for this (I would imagine that we must)?

Splaktar:
It looks like it would be related to the new `getId` function or some other place that we assume that if the value is an object, then `trackBy` would be used (but in this case it is not).

Splaktar:
@jelbourn this is another PR that has been blocked since April due to unknown Google internal presubmit failures. Can you please provide some feedback so that this fix can move forward?

Splaktar:
The only information that I have on these presubmit failures is that they are failing on internal Google e2e tests that will require further investigation and debugging by a caretaker.

The failures look like this ` - Expected '' to contain 'crud_test_connection_1529682048366'.` which is not overly helpful for debugging externally.

Bumping this to the next release again.

free-easy:
I beg apologise for the long absence, but I was busy with work and exams.
If there are more informations, please let me know.
I'm available to continue with this PR.

Splaktar:
@free-easy sorry for the delay in getting feedback here, I am working on getting more information.

Can you please respond to my comments in https://github.com/angular/material/pull/11156#issuecomment-383234310? Thank you.

free-easy:
@Splaktar Apparantly it seems that we don't have any test case for "which hash function is being used in which case" and "is the hash function working as expected?". I'm not sure whether we should test it at all, since AFAIK unit tests are supposed to test interface and not implementation (Ok, I'm totally unsure of what I'm saying).
Anyway, as you can see, I've tried to do as less modifications as possible to the hash function. Apart from the one used when using trackBy, which, since I had to support trackBy values to be updated, had to be unavoidably changed a lot, the logic of the hash function when not using trackBy (which is indeed the case the Google app fails in its test) should be pretty much the same.
Could it be that their code was based on the previous wrong implementation of `md-select` which used to use `usingTrackBy ? !angular.equals(prevVal, newVal) : (prevVal + '') !== newVal`?

Splaktar:
It looks as though the internal issues have been resolved and this may be able to get into `1.1.19`.

Issue #3857, opened 2015-07-22
feat(contact-chips): add support for ng-change


izeau:
Hi guys, [here's a codepen](http://codepen.io/anon/pen/OVowEo?editors=101) showing the issue.

Basically, to trigger a function when the `<md-contact-chips>` updates, we (seemingly) have to use `$scope.$watch`, which I'm trying to avoid thorough the codebase. Can we have it working (or a similar attribute such as `md-contact-change`)?

rgolea:
Try with `$scope.$watchCollection` or if not pass in a 3rd parameter as `true`:
```
$scope.$watchCollection('whatever', fn)
//or
$scope.$watch('whatever', fn, true);
```

izeau:
Thanks for the heads up! But I’d really like it as an attribute.

rgolea:
Well... You can always do a directive. The thing is that the actual `ng-change` is based on the view change:
(Code from source)
```
var ngChangeDirective = valueFn({
  restrict: 'A',
  require: 'ngModel',
  link: function(scope, element, attr, ctrl) {
    ctrl.$viewChangeListeners.push(function() {
      scope.$eval(attr.ngChange);
    });
  }
});
```
You can always do something like this:
```
.directive('hardChange', function(){
    return {
        restrict: A,
        require: 'ngModel',
        link: function(scope, elem, attr, ctrl){
            scope.$watchCollection(function(){
                 return ctrl;
            }, function(newVal, oldVal){
              if(newVal === oldVal) return;
              scope.$eval(attr.hardChange)
            });
       }
    };
});

izeau:
Yeah sure but that looks to me like a feature Material could offer by itself!

rgolea:
Do a fork and do a PR yourself if you want...

typotter:
What we need to do here, is update the $viewValue of the ngModelCtrl on md-chips when the collection changes. This should not be too hard, and should make ng-change work for you.

beazergood:
Hi peeps. I'm a bit of a noob when it comes to Github and how all this works.  I've read this as it's not released yet and 1.1.0 is the target release for this functionality?... That correct?

aleccool213:
any chance of this ever being done?

avilaj:
Has this been fixed? we need this in order to subscribe to ngModel $viewChangeListeners.

jmwilkinson:
Why was this closed? This issue isn't fixed.

rhclayto:
This is still an issue.

[...]

PR #11166, opened 2018-03-13
feat(chips): trigger ng-change on chip addition/removal

### LINKED ISSUE #11161: feat(chips): add support for ng-change
`ng-change` is not triggered when I add a chip or remove it.

`ng-change` should be triggered.

`ng-change` is not triggered.

Try to insert some chip and you'll notice that ng-change is not triggered.

Well, I think that any component using ng-model, should also support ng-change, since
internally it should call `ngModelController.$setViewValue`.

Sure, `md-chips` supplies `md-on-add` and `md-on-remove`, which should suffice, but in my case
I'm using the `ngModelOptions` `getterSetter: true` and the setter, just as ng-change, seems to not be called.

AngularJs: 1.5.5 (the default one of the CodePen that you supply
Angular Material: 1.1.7

free-easy:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

`ng-change` is not supported for `md-chips` or `md-contact-chips`.

Issue Number: 
#11161 #3857 

## What is the new behavior?
`ng-change` is supported on `md-chips`. It will be triggered on chip addition/removal.

```
[ ] Yes
[x] No
```

## Other information
Closes #11161 and #3857

### DIFF
--- a/src/components/chips/demoBasicUsage/script.js
+++ b/src/components/chips/demoBasicUsage/script.js
@@ -14,6 +14,7 @@
 
     // Lists of fruit names and Vegetable objects
     self.fruitNames = ['Apple', 'Banana', 'Orange'];
+    self.ngChangeFruitNames = angular.copy(self.fruitNames);
     self.roFruitNames = angular.copy(self.fruitNames);
     self.editableFruitNames = angular.copy(self.fruitNames);
 
@@ -38,6 +39,10 @@
         name: chip,
         type: 'unknown'
       };
+    };
+
+    self.onModelChange = function(newModel) {
+      alert('The model has changed');
     };
   }
 })();
--- a/src/components/chips/demoContactChips/script.js
+++ b/src/components/chips/demoContactChips/script.js
@@ -25,6 +25,7 @@
 
     self.querySearch = querySearch;
     self.delayedQuerySearch = delayedQuerySearch;
+    self.onModelChange = onModelChange;
 
     /**
      * Search for contacts; use a random delay to simulate a remote call
@@ -49,7 +50,7 @@
             resolve( self.querySearch(criteria) );
 
             refreshDebounce();
-          }, Math.random() * 500, true)
+          }, Math.random() * 500, true);
         });
       }
 
@@ -82,6 +83,10 @@
         return (contact._lowername.indexOf(lowercaseQuery) != -1);
       };
 
+    }
+
+    function onModelChange(model) {
+      alert('The model has changed');
     }
 
     function loadContacts() {
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -328,7 +328,7 @@ MdChipsCtrl.prototype.getCursorPosition = function(element) {
 MdChipsCtrl.prototype.updateChipContents = function(chipIndex, chipContents){
   if(chipIndex >= 0 && chipIndex < this.items.length) {
     this.items[chipIndex] = chipContents;
-    this.ngModelCtrl.$setDirty();
+    this.updateNgModel();
   }
 };
 
@@ -483,9 +483,7 @@ MdChipsCtrl.prototype.appendChip = function(newChip) {
   var length = this.items.push(newChip);
   var index = length - 1;
 
-  // Update model validation
-  this.ngModelCtrl.$setDirty();
-  this.validateModel();
+  this.updateNgModel();
 
   // If they provide the md-on-add attribute, notify them of the chip addition
   if (this.useOnAdd && this.onAdd) {
@@ -584,6 +582,13 @@ MdChipsCtrl.prototype.validateModel = function() {
   this.ngModelCtrl.$validate(); // rerun any registered validators
 };
 
+MdChipsCtrl.prototype.updateNgModel = function() {
+  this.ngModelCtrl.$setViewValue(this.items.slice());
+  // TODO add the md-max-chips validator to this.ngModelCtrl.validators so that
+  // the validation will be performed automatically on $viewValue change
+  this.validateModel();
+};
+
 /**
  * Removes the chip at the given index.
  * @param index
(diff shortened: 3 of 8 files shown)


Splaktar:
Sorry for more commit message confusion. `Closes #11161 and #3857` needs to be `Closes #11161. Closes #3857` for the GitHub parsing to work properly.

free-easy:
@Splaktar Why do we need a test of `ng-change` for `md-contact-chips`? Since `md-contact-chips` is just a wrapper of `md-chips`, I think it's redudant to test on both the components.

Splaktar:
OK, that test may not be needed, as long as we can see it working in the demo. It still seems somewhat useful to have though to make sure that some change to `md-contact-chips` does not break the `ng-change` functionality.

src/components/chips/js/contactChipsDirective.js
```diff
@@ -122,6 +124,7 @@ function MdContactChips($mdTheming, $mdUtil) {
       contactImage: '@mdContactImage',
       contactEmail: '@mdContactEmail',
       contacts: '=ngModel',
+      ngChange: "&",
```
IMM0rtalis: Why double quote here when all the other bindings have single quote?

src/components/chips/js/chipsController.js
```diff
@@ -329,6 +329,7 @@ MdChipsCtrl.prototype.updateChipContents = function(chipIndex, chipContents){
   if(chipIndex >= 0 && chipIndex < this.items.length) {
     this.items[chipIndex] = chipContents;
     this.ngModelCtrl.$setDirty();
+    this.ngModelCtrl.$setViewValue(this.items.slice());
```
IMM0rtalis: Since this line is repeated 3-times - why not wrap it into a helper-function similar to validateModel() ?

src/components/chips/js/contactChipsDirective.js
free-easy: Opss, it just a matter of habit. In the project I'm working on we use double quotes. I'll fix.

src/components/chips/js/chipsController.js
free-easy: It seems there is already a function called `validateModel` and I think that `updateModel` describes better what does this statement do.

free-easy: But do we really need an helper function for just a single line of instruction?

On the other hand, if you meant to wrap into a function the two lines that are repeated

```
this.ngModelCtrl.$setDirty();
this.ngModelCtrl.$setViewValue(this.items.slice());
```
then I think we can remove `this.ngModelCtrl.$setDirty()`, since AFAIK `$setViewValue()` should already set the model as $dirty

src/components/chips/js/contactChipsDirective.js
free-easy: In the meantime I've fixed this.

src/components/chips/js/chipsController.js
IMM0rtalis: oh, I maybe choosed the wrong words. I meant something like the validateModel(), not name it this way.
But it is 3-times calling a function `$setViewValue` and chooses the a copy (`slice()`) from the inner value-cache (`this.items`). Maybe there is a small change there so the developer has to update now 3-lines instead of one. 
This is just an suggestion to wrap this in an extra function because I like to reduce repeated code and make it easier to maintain. But every developer has different workflows. 

PS.: if this helper-function could also reduce other duplications it is even better.

free-easy:
@IMM0rtalis 
How do you like the changes I made?
In order to make the unit tests pass, I needed to call `this.validateModel` in `updateNgModel` as well, which
is kinda redudant since by calling `this.ngModelCtrl.$setViewValue` the validation is automatically run.
I've added a TODO, since the possible minor refactor would be better done in a different pull-request.

src/components/chips/demoBasicUsage/script.js
```diff
@@ -39,5 +40,9 @@
         type: 'unknown'
       };
     };
+
+    self.onModelChange = function(newModel) {
+      alert("The model has changed");
```
Splaktar: Can you please use single quotes here?

src/components/chips/demoContactChips/script.js
```diff
@@ -84,6 +85,10 @@
 
     }
 
+    function onModelChange(model) {
+      alert("The model has changed");
```
Splaktar: Can you please use single quotes here?

src/components/chips/demoBasicUsage/script.js
free-easy: Fixed

src/components/chips/demoContactChips/script.js
free-easy: Fixed

src/components/chips/demoBasicUsage/script.js
```diff
@@ -39,5 +40,9 @@
         type: 'unknown'
       };
     };
+
+    self.onModelChange = function(newModel) {
+      alert('The model has changed');
```
Splaktar: Let's use the model via `alert('The model has changed to ' + newModel);`

Splaktar:
Testing done. Submitting to presubmit.

Splaktar:
I'm not sure what the details are yet, but there may be some issue with these changes that may require reverting them. Just a heads up. More details may be available in https://github.com/angular/material/pull/11214 at some point.

src/components/chips/js/chipsController.js
```diff
@@ -584,16 +582,21 @@ MdChipsCtrl.prototype.validateModel = function() {
   this.ngModelCtrl.$validate(); // rerun any registered validators
 };
 
+MdChipsCtrl.prototype.updateNgModel = function() {
+  this.ngModelCtrl.$setViewValue(this.items.slice());
```
Splaktar: This line seems to be related to https://github.com/angular/material/issues/11304 and https://github.com/angular/material/issues/11301.

PR #11169, opened 2018-03-16
fix(toast): update style direction to flex-direction

ypislon:
In the component md-toast the property "direction" is set to "row". It seems like it's intended to use "flex-direction", as "row" is not a valid value for the property direction, which should indicate the direction of text. 

This doesn't change the behaviour of md-toast, just cleaning up a small typo :)

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

The css property "direction" is set to the value of "row" in the md-toast component.

## What is the new behavior?

In the component md-toast the property "direction" is set to "row". It seems like it's intended to use "flex-direction" instead of "direction", as "row" is not a valid value for the property direction, which should indicate the direction of text. 

This doesn't change the behaviour of md-toast, just cleaning up a small typo :)

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/toast/toast.scss
+++ b/src/components/toast/toast.scss
@@ -26,7 +26,7 @@ md-toast {
 
   .md-toast-content {
     display: flex;
-    direction: row;
+    flex-direction: row;
     align-items: center;
 
     max-height: 7 * $toast-height;


ypislon:
I signed the CLA.

Splaktar:
Yep, looks like only `rtl` and `ltr` are [valid values](https://developer.mozilla.org/en-US/docs/Web/CSS/direction) other than `initial/inherit/unset`.

Splaktar:
Thank you for your contribution! Please update your commit message to follow the [commit message guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format). This is important for your changes to properly show up in the [changelog](https://github.com/angular/material/blob/master/CHANGELOG.md).

For example: `fix(toast): update style direction to flex-direction`

Splaktar:
This looks good to me once the commit message gets updated. Thank you!

ypislon:
Commit message is updated!

Issue #11202, opened 2018-03-30
md-autocomplete panel doesn't inherit dark theme


rudzikdawid:
screen of issue:
![light automplete in dark theme](https://i.imgur.com/eZLh0oi.png)

in `material1` implementation there is render issue. **light autocomplete panel in dark theme**.
`material2` has proper implementation dark autocomplete panel in dark theme.
https://material.angular.io/components/autocomplete/examples

automplete in dark theme should look like:
![dark automplete in dark theme](https://i.imgur.com/QGIzMfQ.png)

PR: https://github.com/angular/material/pull/11203
fixed scss styles:
```scss
md-autocomplete.md-THEME_NAME-theme {
  background: '{{background-hue-1}}';
  &[disabled]:not([md-floating-label]) {
    background: '{{background-hue-2}}';

 input {
    color: '{{foreground-1}}';
  }
}


.md-autocomplete-suggestions-container.md-THEME_NAME-theme {
  background: '{{background-hue-1}}';
  li {
    color: '{{foreground-1}}';
    
    &:hover,
    &.selected {
      background: '{{background-300-0.2}}';
    }

  }
}
```

PR #11198, opened 2018-03-28
fix(select): options panel theme supports dark mode

Splaktar:
## PR Checklist
Please check that your PR fulfills the following requirements:

- [x] The commit message follows [our guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format)
- [ ] Tests for the changes have been added (for bug fixes / features)
- [ ] Docs have been added / updated (for bug fixes / features)

## PR Type
What kind of change does this PR introduce?

```
[x] Bugfix
[ ] Feature
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Documentation content changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?
The select's option panel is always using light mode hues even when the theme is dark mode.

Issue Number: 
Fixes #3379.

## What is the new behavior?
The select's option panel use dark mode hues when the theme is set to dark mode.

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information
I'm going to try and presubmit this for 1.1.9, but I understand that this might trigger some screenshot failures in Google apps. If that happens, we'll need to evaluate pushing this to 1.2.0.

Dark Before:
screen shot 2018-03-28 at 5 59 25 pm

Dark After with Item 1 selected:
screen shot 2018-04-07 at 4 05 18 am

Light Before:
screen shot 2018-04-07 at 4 04 37 am

Light After:
screen shot 2018-04-07 at 4 07 55 am

PR #11203, opened 2018-03-30
fix(autocomplete): autocomplete panel theme supports dark mode

rudzikdawid:
## PR Checklist

- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type

```
[x] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

The autocomplete panel is always using light mode hues even when the theme is dark mode.

Issue Number: 
#11202

## What is the new behavior?
The autocomplete panel use dark mode hues when the theme is set to dark mode.

```
[ ] Yes
[x] No
```

## Other information
Before:

After:

### DIFF
--- a/src/components/autocomplete/autocomplete-theme.scss
+++ b/src/components/autocomplete/autocomplete-theme.scss
@@ -1,7 +1,7 @@
 md-autocomplete.md-THEME_NAME-theme {
-  background: '{{background-A100}}';
+  background: '{{background-hue-1}}';
   &[disabled]:not([md-floating-label]) {
-    background: '{{background-100}}';
+    background: '{{background-hue-2}}';
   }
   button {
     md-icon {
@@ -13,17 +13,20 @@ md-autocomplete.md-THEME_NAME-theme {
       background: '{{background-600-0.3}}';
     }
   }
+  input {
+    color: '{{foreground-1}}';
+  }
 }
 .md-autocomplete-suggestions-container.md-THEME_NAME-theme {
-  background: '{{background-A100}}';
+  background: '{{background-hue-1}}';
   li {
-    color: '{{background-900}}';
+    color: '{{foreground-1}}';
     .highlight {
       color: '{{background-600}}';
     }
     &:hover,
     &.selected {
-      background: '{{background-200}}';
+      background: '{{background-500-0.18}}';
     }
   }
 }


Splaktar:
The focus background color in light mode has gotten significantly lighter. It should stay the same or at minimum be visually indistinguishable.

Before:
screen shot 2018-04-04 at 10 45 43 pm
After:
screen shot 2018-04-04 at 10 45 55 pm

Before:
screen shot 2018-04-04 at 10 46 48 pm
After:
screen shot 2018-04-04 at 10 47 03 pm

rudzikdawid:
You have right 
hover `background: '{{background-300-0.2}}';` wasn't perfect

so i tried your solutions from https://github.com/angular/material/pull/11198
but then i relised that yours: `background-color: '{{background-hue-2}}'` also have problems, mostly in dark mode

after that i tried hover `background: '{{background-500-0.17}}';`
and it seems ok

before:


after:


dark mode:

rudzikdawid:
I think that `background-500-0.20` seems to be a little bit darker than master build, esspecially in light theme. `background-500-0.17` looks mutch better.

Splaktar:
Yep, as mentioned in https://github.com/angular/material/pull/11198#issuecomment-379476633, let's use `0.18`.

rudzikdawid:
done

Splaktar:
According to https://material.io/guidelines/components/text-fields.html#text-fields-states, it appears that a disabled dark mode text field (`md-autocomplete` with `md-floating-label`) should be have the default background color (i.e. the same as the content area). This seems to be displayed properly.

According to https://material.io/guidelines/components/text-fields.html#text-fields-text-field-boxes (I think that this maps closest to our autocomplete design w/o `md-floating-label`), it appears that the disabled dark mode should have a background color (`#262626`) that is slightly lighter than the default background (`#1D1D1D`) and slightly darker than the non-disabled background (`#2F2F2F`). Currently in this branch, it is darker than the default background in dark mode.
screen shot 2018-04-09 at 4 41 17 am

Splaktar:
I don't see any combination that we have available atm that meets the requirements listed in my last post. I think that we should go with the following for the time being:
```scss
  &[disabled]:not([md-floating-label]) {
    background: '{{background-default}}';
  }
```
It still provides a difference between enabled/disabled styling, but avoids applying a darker background color than the default background.
screen shot 2018-04-09 at 5 02 31 am

Splaktar:
But then again, testing light mode `background-hue-2` gives us the same styling as previously, so if we use the `background-default`, it could cause problems. So let's keep it like you have it with `background-hue-2` as I would rather have the inconsistency on the dark mode side rather than light mode.

rudzikdawid:
i see this issue also on master build. So maybe we should open another issue/PR for that.
This issue was opened to prepare fix for dark theme.

Splaktar:
I think that we can send this for presubmit testing now.

rudzikdawid:
Now we have backward compatible in light mode with `master` build.
I think this is important because issue #11202 concerns a dark theme support

Issue #11218, opened 2018-04-09
autocomplete(chips): deleting chips breaks ability to add or render new chips


Splaktar:
### Bug, feature request, or proposal:
Bug

### What is the expected behavior?
Deleting chips should not break the ability to add or render other chips.

### What is the current behavior?
We support multi value with separator comma, enter, space and tab.
It works good as long as we don't delete any chip. As soon as we delete the chip, nothing works. 
We made sure model looks good even after deleting but somehow `md-item-template` doesn't get rendered.

##### CodePen Demo which shows your issue: TBD
```html
<md-chips
      flex="100"
      md-autocomplete-snap
      class="..."
      readonly="..."
      ng-model="..."
      ng-if="..."
      md-transform-chip="..."
      md-on-remove="...">
</md-chips>
```
### What is the use-case or motivation for changing an existing behavior?
It worked as expected in 1.1.7.

### Which versions of AngularJS, Material, OS, and browsers are affected?
AngularJS Material 1.1.7
Others TBD

### Is there anything else we should know? Stack Traces, Screenshots, etc.
Verified that this is fixed by reverting https://github.com/angular/material/pull/11166 in https://github.com/angular/material/pull/11214.

Splaktar:
@jelbourn the issue mentions `md-item-template`, but I assume that it means `md-chip-template`?
Do you have any details on what that template looks like?

It also appears that the custom logic within the `md-transform-chip` and `md-on-remove` expressions could be causing the problem here.

I tested this on the [1.1.8 docs](https://material.angularjs.org/latest/demo/chips) and I cannot reproduce any problems related to deleting and adding more chips.

Splaktar:
@jelbourn Here is a [Demo](https://codepen.io/Splaktar/pen/oqmNdr?editors=1010) where `md-on-remove` can cause `md-transform-chip` to start returning `null` and blocking chip creation. Is this what is happening? Is this behavior (chip being removed when it should be added) what the team is seeing?

hetal0713:
Hi Splaktar,

@jelbourn reported issue on behalf of me. See my comments below

@jelbourn the issue mentions md-item-template, but I assume that it means md-chip-template?
Do you have any details on what that template looks like?
>> yes its chip template. Below is template (some info changed from original template)

<md-chip-template
        username-chip
        class="xyz"
        ng-class="{'invalidTag': $chip.invalid}">
      <span>
         {{$chip.name}}
      </span>
      <md-tooltip>
[...]

Issue #11217, opened 2018-04-09
chips: adding chip on blur no longer working when required is also used


Splaktar:
### Bug, feature request, or proposal:
Bug

### What is the expected behavior?
Clicking outside of the chips with input in the chips should create a new chip.
Adding chips and removing chips and adding more chips should work.

### What is the current behavior?
We have multi value supported with separator keys of enter, space and tab. 
It works good when we have input like (url1 url2) and hit tab to move out of the field
It doesn't transform into chips when we have input like (url1 url2) and click anywhere outside of the field (this used to work before this sync)
After adding chips, if we remove the chips and retry with any input, it doesn't work. Somehow it makes the model as "undefined"

### What is the use-case or motivation for changing an existing behavior?
It used to work in 1.1.7.

### Which versions of AngularJS, Material, OS, and browsers are affected?
AngularJS Material 1.1.7
Others TBD

### Is there anything else we should know? Stack Traces, Screenshots, etc.
Verified that this is fixed by reverting https://github.com/angular/material/pull/11125 in https://github.com/angular/material/pull/11215.

Splaktar:
@jelbourn can you confirm that they have `md-add-on-blur="true"` on the `md-chips`?
Are they using `ng-required`?

IMM0rtalis:
I‘ll try to find a fix/solution the next 48hours

Splaktar:
@jelbourn It seems like the custom logic in the `md-transform-chip` expression may be causing this issue. If it starts returning `null`, this will block chips from being created.

Splaktar:
I tested on the [docs for HEAD](https://material.angularjs.org/HEAD/demo/chips) and I could not reproduce these issues there. But those docs don't have an example of using `md-add-on-blur` with `required`.

So I created a [demo](https://codepen.io/Splaktar/pen/NYoWEv?editors=1010#0) that uses `md-add-on-blur` with `required` and there does appear to be an issue. Add on blur does not work for the first chip either with the tab key or a mouse click.

Splaktar:
@hetal0713 we were able to isolate and reproduce this issue. We're working on a test and a fix atm. In the meantime, the code that caused this has been reverted from `master`.

hetal0713:
Ok great. Thank you for the update

Splaktar:
[...]

PR #11219, opened 2018-04-09
revert(chips): ng-required and ng-change

Splaktar:
## PR Checklist

- [] The commit message follows [our guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format)
- [ ] Tests for the changes have been added (for bug fixes / features)

## PR Type
```
[ ] Bugfix
[ ] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[x] Other... Please describe: Reverts previous PRs that introduced regressions
```

**md-chips (standalone)**
We have multi value supported with separator keys of enter, space and tab. 
It works good when we have input like (url1 url2) and hit tab to move out of the field
It doesn't transform into chips when we have input like (url1 url2) and click anywhere outside of the field (this used to work before this sync)
After adding chips, if we remove the chips and retry with any input, it doesn't work. Somehow it makes the model as "undefined"

**md-chips with md-autocomplete**
We support multi value with separator comma, enter, space and tab.
It works good as long as we don't delete any chip. As soon as we delete the chip, nothing works. 
We made sure model looks good even after deleting but somehow md-item-template doesn't get rendered (edited)

Issue Number: 
https://github.com/angular/material/issues/11217 and https://github.com/angular/material/issues/11218

## What is the new behavior?
`md-add-on-blur` works again and deleting chips does not break the ability to add or render new chips.

```
[x] Yes
[ ] No
```
It removes two features that were introduced in 1.1.8. New PRs will need to be submitted to add them again with tests that cover these issues.

## Other information
TBD


IMM0rtalis:
Would it not make more sense to wait 1-2 days before reverting? I it is still an improvement which needs a small bugfix and an additional test-case in my opinion.

PR #11225, opened 2018-04-11
feat(md-chips): re-added validation for ng-required

Splaktar:
## PR Checklist

## PR Type
```
[x] Bugfix
[x] Feature
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

#11125 had to be reverted due to regression #11217.

Issue Number: 
#11217

## What is the new behavior?
This includes #11125 plus the regression fix for #11217.

```
[ ] Yes
[x] No
```

## Other information
N/A

### DIFF
--- a/src/components/chips/chips.spec.js
+++ b/src/components/chips/chips.spec.js
@@ -1055,6 +1055,59 @@ describe('<md-chips>', function() {
 
       });
 
+      describe('ng-required', function() {
+        beforeEach(function() {
+            // Clear default items to test the required chips functionality
+            scope.items = [];
+        });
+
+        it('should set the required error when chips is compiled with an empty array', function() {
+          var template =
+              '<form name="form">' +
+              '<md-chips name="chips" ng-required="true" ng-model="items"></md-chips>' +
+              '</form>';
+
+          var element = buildChips(template);
+          element.scope().$apply();
+
+          expect(scope.form.chips.$error['required']).toBe(true);
+        });
+
+        it('should unset the required error when the first chip is added', function() {
+          var template =
+              '<form name="form">' +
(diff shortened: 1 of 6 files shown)


Splaktar:
@IMM0rtalis can you please take a look and confirm that this is OK?

IMM0rtalis:
@googlebot I confirm.

IMM0rtalis:
@Splaktar the fix of yours works for me.

But I do not really get why we validate the model in `shouldAddOnBlur`. The model is not even set here, right? So it validates the model before adding (so validates the existing ones). If I am right (and do not oversee sth) I think the logic is wrong. 
Only thing which could be validated here would be a copy of the model which then includes the new element and validate then. If valid -> add, if not, do not add.

Splaktar:
I actually tried to remove the call to `validateModel()` from `shouldAddOnBlur()` and just check for `!hasMaxChipsReached()`, but that caused a couple tests to fail. Part of this is because `appendChip()` already calls `validateModel()`. So for the add on blur case, `validateModel()` is always called twice. 

One of the existing tests just directly calls `$setValidity('is-valid', false)` and doesn't depend on the `md-max-chips` validity. I'm unclear exactly what the use case could be there.

From what I can tell the valid model check is only there so that a chip doesn't get added when `md-max-chips` has been reached. But we have tests that go beyond that and it's unclear why, but I'm not comfortable removing those tests as users may be relying on that behavior.

IMM0rtalis:
I also just looked into the tests and I still think fixing the blur-function would be a good improvement. But this would be out of scope for this fix anyway.

PR #11302, opened 2018-05-29
fix(ripple): cleanup ripple even when the touch is very quick on iOS

### LINKED ISSUE #11069: ripple: ink ripple not always removed when animation is complete on iOS

The ink ripple is not always removed on mobile browsers

Note: this is different from the simulated hover on mobile browsers; The simulated hover is also present but that can be overridden in css if one doesn't want it. But the stale ink ripple is a real bug.

This seems to be related to synthesized mousedown events; When I add 

    if (!event.$material) return;

to InkRippleCtrl.prototype.handleMousedown then I cannot reproduce it any more. Though that alone is probably too simple a fix since $mdGesture.isHijackingClicks need not be turned on.

Ink ripples disappear eventually.

Sometimes, an ink ripple remains. Not initially, but eventually when combining long and short clicks.

* Open codepen in chrome mobile emulator
* Long-click on button (>1s)
* Wait (>1s)
* Short-click on button
* Wait
* Short-click on button 

Now there is a 
````
<div class="md-ripple-container" style="background-color: rgba(33, 33, 33, 0.1);"><div class="md-ripple md-ripple-placed md-ripple-scaled md-ripple-active" style="left: 29px; top: 30px; background: rgb(33, 33, 33); width: 132.378px; height: 132.378px; border-color: rgb(33, 33, 33);"></div></div>
````
in the dom that does not go away

AngularJS material 1.1.5
Android chrome
iOS Safari
Chrome mobile emulator

### PR DESCRIPTION
**Pull request by jpike88**
## PR Checklist

## PR Type

- [x] Bugfix

Issue Number:  #11069
http://streaka.testing.videos.s3.amazonaws.com/screencast_2018-05-29_14-38-08.mp4

## What is the new behavior?
Issue doesn't occur anymore, seems fine

- [x] No

## Other information

### DIFF
--- a/src/core/services/ripple/ripple.js
+++ b/src/core/services/ripple/ripple.js
@@ -280,7 +280,9 @@ InkRippleCtrl.prototype.handleMousedown = function (event) {
  * mouseup, touchend or mouseleave event)
  */
 InkRippleCtrl.prototype.handleMouseup = function () {
-  autoCleanup(this, this.clearRipples);
+  this.$timeout(function () {
+    autoCleanup(this, this.clearRipples);
+  }.bind(this));
 };
 
 /**

### CODE REVIEW

**Comment by jpike88** [comment]
I signed it!

File: src/core/services/ripple/ripple.js [hunk 1]
```diff
@@ -280,7 +280,9 @@ InkRippleCtrl.prototype.handleMousedown = function (event) {
  * mouseup, touchend or mouseleave event)
  */
 InkRippleCtrl.prototype.handleMouseup = function () {
-  autoCleanup(this, this.clearRipples);
+  setTimeout(function () {
```
Splaktar: Please use `this.$timeout` instead of `setTimeout` directly.

PR #11310, opened 2018-06-01
fix(chips): regression where chips model gets out of sync with view

### LINKED ISSUE #11301: chips: ng-change with read-only array-model causes problems
ng-change introduced new assignment to model-value instead of just adding/removing the elements of the array. Therefore model-values which just have a getter and no setter are not working anymore.

_In some cases it is just wanted to have a property with no setter to not touch the reference to the array but let the values be changed._

[CodePen Demo 1.1.8](https://codepen.io/anon/pen/yEBoWE) where it still works. 
[CodePen Demo 1.1.9](https://codepen.io/anon/pen/gKYxVg) where it broke. 

See codepen + look at [The codeline which broke the working code](https://github.com/angular/material/commit/0563f241e3b4b9916e5a572c863178ec201f8f70#diff-441d4815c4f57bf9fcf5e56ca16b5801R587).

Code does still work after updating to 1.1.9. 

Completely dies, latets after removing a chip.

Make it work again without changing the application logic because of a material update

- AngularJS Material: 1.1.9
 
Introduced with #11237

### LINKED ISSUE #11304: chips: when the model(array) is modified the chips view is not updated

When using `md-on-add` in Angular Material 1.1.9 to remove a chip with a JS function, the view is not updated but the model is.

1. Open the CodePen and enter brocoli in fruit chips list.
2. The brocoli is supposed to go in vegetable chips list.
3. If you change the version to Angular Material 1.1.8 it will work properly.

To update the chips view when the chip is removed programmatically.

The chips view is not updated and does not match the model.

### What is the use-case or motivation for changing an existing behaviour?
Was working with AM 1.1.8 and below.

- AngularJS: 1.6.9
- AngularJS Material: 1.1.9
 
I think is caused by #11237

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

There was a breaking change in 1.1.9 related to the enhancement to support `ng-change` on `md-chips`. This was introduced in #11166 where the passed in model and the `items` Array is replaced via `Array.slice`. This appears to have led to the issues listed below and may be related to https://github.com/angular/material/issues/11218 as well.

Issue Number: 
Fixes #11304. Fixes #11301.

## What is the new behavior?
- fix issue where view and model get out of sync
- manually call `ngChange` as it won't be triggered by `ngModel` for chips
- add `ng-change` test to contact-chips
- refinements to chips and contact-chips docs
- remove references to `filter-selected` on contact-chips
- it was disabled 2 years ago
- remove use of `angular.lowercase` which is removed in AngularJS 1.7

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/chips/js/chipsController.js
+++ b/src/components/chips/js/chipsController.js
@@ -21,9 +21,11 @@ angular
  * @param $element
  * @param $timeout
  * @param $mdUtil
+ * @param $exceptionHandler
  * @constructor
  */
-function MdChipsCtrl ($scope, $attrs, $mdConstant, $log, $element, $timeout, $mdUtil) {
+function MdChipsCtrl ($scope, $attrs, $mdConstant, $log, $element, $timeout, $mdUtil,
+                      $exceptionHandler) {
   /** @type {$timeout} **/
   this.$timeout = $timeout;
 
@@ -41,6 +43,9 @@ function MdChipsCtrl ($scope, $attrs, $mdConstant, $log, $element, $timeout, $md
 
   /** @type {$log} */
   this.$log = $log;
+
+  /** @type {$exceptionHandler} */
+  this.$exceptionHandler = $exceptionHandler;
 
   /** @type {$element} */
   this.$element = $element;
@@ -142,10 +147,10 @@ function MdChipsCtrl ($scope, $attrs, $mdConstant, $log, $element, $timeout, $md
   this.contentIds = [];
 
   /**
-   * The index of the chip that should have it's tabindex property set to 0 so it is selectable
+   * The index of the chip that should have it's `tabindex` property set to `0` so it is selectable
    * via the keyboard.
    *
-   * @type {int}
+   * @type {number}
    */
   this.ariaTabIndex = null;
 
@@ -326,9 +331,9 @@ MdChipsCtrl.prototype.getCursorPosition = function(element) {
  * @param chipContents
  */
 MdChipsCtrl.prototype.updateChipContents = function(chipIndex, chipContents){
-  if(chipIndex >= 0 && chipIndex < this.items.length) {
+  if (chipIndex >= 0 && chipIndex < this.items.length) {
     this.items[chipIndex] = chipContents;
-    this.updateNgModel();
+    this.updateNgModel(true);
   }
 };
 
@@ -454,8 +459,7 @@ MdChipsCtrl.prototype.getAdjacentChipIndex = function(index) {
 /**
  * Append the contents of the buffer to the chip list. This method will first
  * call out to the md-transform-chip method, if provided.
- *
- * @param newChip
+ * @param {string} newChip chip buffer contents that will be used to create the new chip
  */
 MdChipsCtrl.prototype.appendChip = function(newChip) {
   this.shouldFocusLastChip = true;
@@ -486,7 +490,7 @@ MdChipsCtrl.prototype.appendChip = function(newChip) {
 
   this.updateNgModel();
 
-  // If they provide the md-on-add attribute, notify them of the chip addition
+  // If the md-on-add attribute is specified, send a chip addition event
   if (this.useOnAdd && this.onAdd) {
     this.onAdd({ '$chip': newChip, '$index': index });
   }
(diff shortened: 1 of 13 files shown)


src/components/chips/js/chipsController.js
```diff
@@ -577,23 +577,34 @@ MdChipsCtrl.prototype.hasMaxChipsReached = function() {
...
-  // TODO add the md-max-chips validator to this.ngModelCtrl.validators so that
-  // the validation will be performed automatically on $viewValue change
-  this.validateModel();
+/**
+ * Function to handle updating the model, validation, and change notification when a chip
+ * is added, removed, or changed.
+ * @param {boolean=} skipValidation true to skip calling validateModel()
+ */
+MdChipsCtrl.prototype.updateNgModel = function(skipValidation) {
+  this.ngModelCtrl.$setDirty();
```
gkalpak: Why not call `$setViewValue()`? This will break expected `ngModel` behavior.

src/components/chips/js/chipsController.js
```diff
@@ -577,23 +577,34 @@ MdChipsCtrl.prototype.hasMaxChipsReached = function() {
...
+  if (!skipValidation) {
+    this.validateModel();
+  }
+  if (this.ngChange) {
+    this.ngChange(this.items);
```
gkalpak: AFAICT, this will not be called correctly. Since you are defining a scope binding (via `ngChange: '&'`), you need to pass an object that maps arguments to values. For example, if you have `ng-change="onChange(items)"`, then you need to call it with `this.ngChange({items: this.items})`.

src/components/chips/js/chipsController.js
Splaktar: With what value? Calling `$setViewValue(this.items.slice())` is what caused these regressions (issues linked in OP). Calling `$setViewValue(this.items)` doesn't trigger `ngChange`, but I guess I could try that again now that we're doing it manually.

Splaktar: `this.ngModelCtrl.$setViewValue(this.items);` does seem to work now with the manual triggering of `ngChange`. Will update.

gkalpak: I see. So, yeah, calling `$setViewValue(this.items)` might not work (since it uses `===` to [compare to previous value][1]) and thus will not perform the expected operations (update classes, validate, call `$viewChangeListeners`).

As a workaround, maybe you could call `$viewChangeListeners` yourself, which I think it a better approach than manually calling `ngChange` ([relevant AngularJS code][2]):

```js
angular.forEach(ngModelCtrl.$viewChangeListeners, function(listener) {
  try {
    listener();
  } catch (e) {
    $exceptionHandler(e);
  }
});
```

[1]: https://github.com/angular/angular.js/blob/459997b482ef9597ce3505d16f7ebc3b9a14b9cb/src/ng/directive/ngModel.js#L713
[2]: https://github.com/angular/angular.js/blob/459997b482ef9597ce3505d16f7ebc3b9a14b9cb/src/ng/directive/ngModel.js#L781-L788

src/components/chips/js/chipsController.js
gkalpak: The key has to be the same as what they used as variable name. Essentially, a `&` binding allows you to evaluate an expression on the parent scope (of an isolate scope) and pass certain data. For security reasons, AngularJS creates and passes a wrapper to the isolate scope. So, the isolate scope does not get direct access to the expression (most often a method). What it gets is a wrapper function that takes a map of argument names to values and calls the expression on the parent scope with these values.

For example, having a `parentMethod: '&'` binding and using the component as `parent-method="doStuff(arg1, arg2)"`, allows the component's isolate scope to call `this.parentMethod({arg1: 'value1', arg2: 'value2'})`, which will eventually call `doStuff('value1', 'value2')` on the parent scope.

So, you need to know what arguments are used in the expression.

##
Usually, we don't have this issue with `ngChange`, because there is no isolate scope. In cases like this (where you have an isolate scope and you can't just `.$eval(attrs.ngChange)`), a common approach is to specify (i.e. document) that in the `ngChange` expression the chips will be available as `$chips` or `$items` (or whatever name you choose).

Then, people can use `ng-change="onChipsChange($chips)"` and you can call it with `this.ngChange({$chips: this.items})` (and their `onChipsChange()` method will be called with `this.items` as argument).

src/components/chips/js/chipsController.js
Splaktar: Excellent, thank you. That seems to work and resolves the argument naming issue.

src/components/chips/js/chipsController.js
Splaktar: Thank you very much for the explanation! That is quite helpful.

It looks like this is not needed due to the suggested change in https://github.com/angular/material/pull/11310#discussion_r192338790.

jelbourn:
I confirmed that that internal app w/ the chips problem works with this.

PR #11320, opened 2018-06-07
fix(compiler): remove dependency on AngularJS private API

### LINKED ISSUE #11319: compiler: refactor to support AngularJS 1.7.1
Proposal

#### Demo which demonstrates the issue: 
https://travis-ci.org/angular/material/jobs/388688438 is just one of many CI unit test jobs which have failed for the AngularJS `SNAPSHOT` view since https://github.com/angular/angular.js/pull/16580 was merged.

AngularJS Material unit tests in `master` pass against the AngularJS `SNAPSHOT` version.

54 failing tests can be seen in all recent jobs like https://travis-ci.org/angular/material/jobs/388688438.

We were fairly well prepared for the changes in https://github.com/angular/angular.js/pull/15782 that went into AngularJS 1.7.0, but the follow on breaking changes in https://github.com/angular/angular.js/pull/16580 that are part of the unreleased AngularJS 1.7.1 remove some private and undocumented APIs that AngularJS Material depended upon.

One example is the use of the third and fourth arguments in `$controller` which were added to fix [an issue with ES6](https://github.com/angular/material/pull/10977) that we resolved in 1.1.5. You can see some of the comments that explain this here: https://github.com/angular/material/blob/0046467372a309414a6de5ac51d8ae99604ea49e/src/core/services/compiler/compiler.js#L452-L477.

- AngularJS: 1.7.1
- AngularJS Material: All
- OS: All
- Browsers: All
 
We cover some of the details of `$compileProvider.preAssignBindingsEnabled()` in our updated docs for 1.1.9 here: https://material.angularjs.org/latest/api/service/$mdCompilerProvider#mdcompilerprovider-respectpreassignbindingsenabled-respected

Splaktar:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Enhancement
[x] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[x] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

`MdCompilerService._createController` uses two private and undocumented AngularJS APIs (on `$controller`).
Some of the docs for `respectPreAssignBindingsEnabled` is confusing or incomplete.

Issue Number: 
Closes #11319

## What is the new behavior?

- remove one of two uses of private/undocumented arguments to $controller
- mark setting `respectPreAssignBindingsEnabled(false)` as deprecated
- updated documentation
- add AngularJS 1.5.x back to TravisCI
- update AngularJS devDependencies to `^1.7.2`

```
[ ] Yes
[x] No
```


src/core/services/compiler/compiler.js
```diff
@@ -124,7 +124,7 @@ function MdCompilerProvider($compileProvider) {
    * - `>=1.7` - the compiler calls the constructor first before assigning bindings and
    *    `$compileProvider.preAssignBindingsEnabled()` no longer exists.
    *
-   * The default value is `false` but will change to `true` in AngularJS Material 1.2.
+   * The default value is `false` in to AngularJS 1.6 and earlier but `true` in AngularJS 1.7.
```
mgol: `in to AngularJS 1.6` -> `in AngularJS 1.6`

mgol: Also, the info that the default will change to `true` (for all AngularJS versions) in Material 1.2 was useful as well, I think it should stay here.

src/core/services/compiler/compiler.js
```diff
@@ -133,9 +133,17 @@ function MdCompilerProvider($compileProvider) {
    * Material Dialog/Panel/Toast/BottomSheet controllers using the `$controller` helper
    * as it always follows the `$compileProvider.preAssignBindingsEnabled()` value.
    */
-  // TODO change it to `true` in Material 1.2.
-  var respectPreAssignBindingsEnabled = false;
+  var respectPreAssignBindingsEnabled;
+  if (angular.version.major === 1 && angular.version.minor === 7) {
```
mgol: `=== 7` -> `>= 7`.

src/core/services/compiler/compiler.js
```diff
@@ -133,9 +133,17 @@ function MdCompilerProvider($compileProvider) {
...
+    respectPreAssignBindingsEnabled = true;
+  } else {
+    respectPreAssignBindingsEnabled = false;
+  }
   this.respectPreAssignBindingsEnabled = function(respected) {
+    if (!respected && angular.version.major === 1 && angular.version.minor === 7) {
```
mgol: `=== 7` -> `>= 7`

src/core/services/compiler/compiler.js
```diff
@@ -133,9 +133,17 @@ function MdCompilerProvider($compileProvider) {
...
+      throw new Error(
+        'Disabling respectPreAssignBindingsEnabled is not supported in AngularJS 1.7.');
```
mgol: `AngularJS 1.7` -> `AngularJS >= 1.7`.

src/core/services/compiler/compiler.js
mgol: The `!respected` check means you'll catch a getter as well so this API won't work at all in 1.7. The error message suggests you meant to add this error for the setter, not a getter. The getter is probably fine.

IOW, this should be moved to inside the `if (angular.isDefined(respected))` block.

src/core/services/compiler/compiler.js
```diff
@@ -444,27 +452,33 @@ function MdCompilerProvider($compileProvider) {
...
-    // controller is defined as a function, `invokeCtrl.instance` is the *same instance* as
-    // `invokeCtrl()`. However, then the controller is an ES6 class, `invokeCtrl.instance` is a
-    // *different instance* from `invokeCtrl()`.
-    var invokeCtrl = this.$controller(options.controller, injectLocals, true, options.controllerAs);
-
-    if (getPreAssignBindingsEnabled() && options.bindToController) {
-      angular.extend(invokeCtrl.instance, locals);
-    }
+    var ctrl;
+    if (angular.version.major === 1 && angular.version.minor === 7) {
```
mgol: `=== 7` -> `>= 7`.

src/core/services/compiler/compiler.spec.js
```diff
@@ -164,12 +164,24 @@ describe('$mdCompiler service', function() {
 
   });
 
-  [
+  var bindingStatesToTest;
+  if (angular.version.major === 1 && angular.version.minor === 7) {
```
mgol: `=== 7` -> `>= 7`.

src/core/services/compiler/compiler.js
```diff
@@ -133,9 +133,17 @@ function MdCompilerProvider($compileProvider) {
    * Material Dialog/Panel/Toast/BottomSheet controllers using the `$controller` helper
    * as it always follows the `$compileProvider.preAssignBindingsEnabled()` value.
    */
-  // TODO change it to `true` in Material 1.2.
-  var respectPreAssignBindingsEnabled = false;
+  var respectPreAssignBindingsEnabled;
+  if (angular.version.major === 1 && angular.version.minor === 7) {
+    respectPreAssignBindingsEnabled = true;
+  } else {
+    respectPreAssignBindingsEnabled = false;
```
mgol: You can shorten that if you want:
```js
var respectPreAssignBindingsEnabled = angular.version.major === 1 && angular.version.minor >= 7;
```

src/core/services/compiler/compiler.js
gkalpak: Unfortunately, you need to check for 1.7.1 here. In 1.7.0 the `options.controllerAs` should be the 4th argument 😒
(It is still OK if you don't support pre-assigning in 1.7.0 so that material bahaves consistently with all 1.7+ versions, but you need to call `$controller` differently on 1.7.0 vs 1.7.1.)

mgol: `controllerAs` is an internal parameter, isn't it? Is there a way to stop relying on it?

gkalpak: It should be possible. It is only used for assigning the controller instance onto `locals.$scope`. It could be done manually with `injectLocals.$scope[options.controllerAs] = <instance>`.

mgol: That seems like a better way to do it.

Splaktar:
I believe that all of the feedback has been addressed. Test still seem to look good. PTAL.

src/core/services/compiler/compiler.js
```diff
@@ -444,27 +448,38 @@ function MdCompilerProvider($compileProvider) {
...
-    // gets the controller instance instead returning of the instance directly. When the
-    // controller is defined as a function, `invokeCtrl.instance` is the *same instance* as
-    // `invokeCtrl()`. However, then the controller is an ES6 class, `invokeCtrl.instance` is a
-    // *different instance* from `invokeCtrl()`.
-    var invokeCtrl = this.$controller(options.controller, injectLocals, true, options.controllerAs);
-
-    if (getPreAssignBindingsEnabled() && options.bindToController) {
-      angular.extend(invokeCtrl.instance, locals);
+    var ctrl;
+    if ((angular.version.major === 1 && angular.version.minor === 7 && angular.version.dot >= 1) ||
```
mgol: I don't think you have to check for `>= 1.7.1` now that internal API usage is completely removed (i.e. the `controllerAs` parameter is no longer passed).

This can be changed to just `angular.version.major === 1 && angular.version.minor >= 7` as in other places.

Am I right, @gkalpak?

src/core/services/compiler/compiler.js
```diff
@@ -124,7 +124,8 @@ function MdCompilerProvider($compileProvider) {
    * - `>=1.7` - the compiler calls the constructor first before assigning bindings and
    *    `$compileProvider.preAssignBindingsEnabled()` no longer exists.
    *
-   * The default value is `false` but will change to `true` in AngularJS Material 1.2.
+   * The default value is `false` in AngularJS 1.6 and earlier but `true` in AngularJS 1.7.
+   * It is planned to change this to always default to `true` in AngularJS Material 1.2.
```
mgol: Saying that the default is `true` in AngularJS 1.7 suggests it can be changed, especially that the docs nowhere say that you can't change it.

Also, `AngularJS 1.7` -> `AngularJS 1.7 or later`.

How about something like:
```
In AngularJS 1.7 or later the value is hardcoded to `true`. In AngularJS 1.6 and earlier
it's `false` by default. It is planned to change this to always default to `true` in
AngularJS Material 1.2.
```

src/core/services/compiler/compiler.spec.js
```diff
@@ -164,12 +164,24 @@ describe('$mdCompiler service', function() {
 
   });
 
-  [
+  var bindingStatesToTest;
+  if (angular.version.major === 1 && angular.version.minor >= 7) {
```
mgol: That's not 100% necessary but a test checking the `respectPreAssignBindingsEnabled` setter throws in 1.7 would be awesome.

src/core/services/compiler/compiler.js
gkalpak: Right 👍

src/core/services/compiler/compiler.spec.js
Splaktar: I gave it a shot, but I'm not sure how to test an exception in a module `config` function and calling it outside of that isn't allowed (not a function).

Splaktar:
I updated the docs a good bit more for clarity and consistency. I also addressed the feedback (other than the unit test). PTAL.

src/core/services/compiler/compiler.js
```diff
@@ -112,31 +113,45 @@ function MdCompilerProvider($compileProvider) {
...
    *
    * If enabled (`true`) the behavior depends on the AngularJS version used:
    *
-   * - `<1.5.10` - bindings are pre-assigned.
-   * - `>=1.5.10 <1.7` - behaves like set to whatever `$compileProvider.preAssignBindingsEnabled()` reports.
-   *    If the `preAssignBindingsEnabled` flag wasn't set manually, it defaults to pre-assigning bindings
-   *    with AngularJS `1.5.x` and to calling the constructor first with AngularJS `1.6.x`.
-   * - `>=1.7` - the compiler calls the constructor first before assigning bindings and
+   * - `<1.5.10`
+   *  - Bindings are pre-assigned.
```
mgol: Oh, I didn't know you can nest sub-lists like that!

src/core/services/compiler/compiler.js
```diff
@@ -444,27 +459,37 @@ function MdCompilerProvider($compileProvider) {
...
-    var invokeCtrl = this.$controller(options.controller, injectLocals, true, options.controllerAs);
-
-    if (getPreAssignBindingsEnabled() && options.bindToController) {
-      angular.extend(invokeCtrl.instance, locals);
+    var ctrl;
+    if (angular.version.major === 1 && angular.version.minor >= 7) {
+      ctrl = this.$controller(options.controller, injectLocals);
+    } else {
+      // The third argument to $controller is considered private and is undocumented
+      // in AngularJS prior to 1.7.1:
```
mgol: A nit but it seems to suggest it was undocumented prior to AngularJS 1.7.1 and now it is which is not the intended meaning. How about:
```js
// The third argument to $controller is considered private in AngularJS prior
// to 1.7.1 and is undocumented:
```

src/core/services/compiler/compiler.js
```diff
@@ -444,27 +459,37 @@ function MdCompilerProvider($compileProvider) {
...
+      // https://github.com/angular/angular.js/blob/v1.6.10/src/ng/controller.js#L102-L109
```
mgol: It's good that you changed it from a volatile `master` reference to a concrete tag. 👍

src/core/services/compiler/compiler.spec.js
mgol: This may be doable, the main problem is it's not possible to call `module` (which is used in tests to access the config phase) after `inject` was called at least once. You'd have to move `inject` from `beforeEach` of `it` blocks so that at the beginning of a new `it` you can still call:
```js
module(function ($mdCompilerProvider) {
  expect(function () {
    $compileProvider.preAssignBindingsEnabled(false);
  }).toThrow();
});
```
I'm not 100% sure if an error thrown during the config phase doesn't break the test infrastructure but maybe it doesn't.

gkalpak: You can test that it throws with something like:

```js
it('should throw ...', () => {
  module('material.core', $mdCompilerProvider =>
    $mdCompilerProvider.respectPreAssignBindingsEnabled(false));
  expect(inject).toThrowError(regexThatMatchesTheErrorMessage);
```

src/core/services/compiler/compiler.js
Splaktar: Yeah, that was a bit confusing. I've separated it into two sentences.

Splaktar:
Since AngularJS 1.7.1 was released today, I wrote up a warning and some guidance in the AngularJS Material forums: https://groups.google.com/forum/?pli=1#!topic/ngmaterial/escIayUJTwE

Splaktar:
@gkalpak I just wanted to confirm that this change is still desirable with PR https://github.com/angular/angular.js/pull/16595 reverting the removal of some of these private APIs?

It looks like there is just one mention of 1.7.1 which can probably be modified. Other than that, I think that this is still valid?

Splaktar:
Oh it looks like the revert in PR https://github.com/angular/angular.js/pull/16595 might be going into AngularJS 1.7.2. So we may still need to keep in the comment about 1.7.1.

mgol:
> I just wanted to confirm that this change is still desirable with PR angular/angular.js#16595 reverting the removal of some of these private APIs?

@Splaktar It's all up to you, really. This decision means that if you want you can support `respectPreAssignBindings` with AngularJS 1.7 (with the exception of 1.7.1). If it's OK to **not** support that w/ AngularJS 1.7, though, then the changes from this PR are still good, especially that they make Material stop relying on private APIs for newer AngularJS versions.

Splaktar:
I think that moving off of the private APIs and removing the complexity of `respectPreAssignBindings` (in 1.7+) is the way to go.

gkalpak:
Since you can't get rid of `respectPreAssignBindings()` in 1.1.x, maybe it is worth keeping the current behavior (to avoid breaking people) and removing support for pre-assigning (as a breaking change) in 1.2.x.

Splaktar:
Right, I don't plan to actually remove the API in 1.1.x, but I would like to keep the behavior that is in this PR and is described in the docs:
> `>=1.7` - the compiler calls the constructor first before assigning bindings and $compileProvider.preAssignBindingsEnabled() no longer exists.

gkalpak:
> `>=1.7` - the compiler calls the constructor first before assigning bindings and `$compileProvider.preAssignBindingsEnabled()` no longer exists.

This can still break an app, though.

Splaktar:
Right, but it's a breaking change in AngularJS 1.7, so they can keep AngularJS on 1.6.x or they can refactor their app.

If they only upgrade AngularJS Material (and not AngularJS), then they shouldn't see anything break.

Am I missing something here?

gkalpak:
> Right, but it's a breaking change in AngularJS 1.7, so they can keep AngularJS on 1.6.x or they can refactor their app.

This is debatable :grin: AngularJS removed `preAssignBidnings` and that is a breaking change in AngularJS. `ngMaterial` (through its own `$mdCompiler`) also supported `preAssignBindings` for specific components (that were instantiated via `$mdCompiler`).

There are two ways to look at things:

1. The fact that `ngMaterial` used some private AngularJS APIs to achieve this is an implementation detail (as far the user is concerned). Seen that way, it could be considered a breaing change in `ngMaterial` to stop pre-assigning bindings in dialogs, etc.

2. On the other hand, the fact that `ngMaterial` has a separate `preAssignBindings` mechanism could also be seen as an implementation detail. In that regard, one could say `ngMaterial` should always follow the AngularJS behavior (and respect `$compileProvider.preAssingBindingsEnabled()` - if present).

I think (2) would be better (aligning `ngMaterial` with AngularJS), but this hasn't been the case 😞
`ngMaterial` used to always pre-assign bindings and ignore `preAssignBindingsEnabled()` and then supported `respectPreAssignBindingsEnabled()`, which imply that pre-assigning bindings in `ngMaterial` is decoupled from AngularJS' pre-assigning.

It is a grey area :smiley:

##
Possible scenario:
- I am using ngMaterial with AngularJS 1.6.x with `preAssignBindingsEnabled(false)`.
- I have fixed all my component that go through `$compile` to work without pre-assigning bindings.
- My `ngMaterial` components (e.g. dialogs), still depend on pre-assigning bindings.
- I upgrade to AngularJS 1.7.x - everything works.
- I upgrade `ngMaterial` to a version that removes pre-assigning bindings - app breaks.

(Again, this is very open to interpretation. E.g. based on peerDependency ranges, one shouldn't be using 1.7.x yet, etc.)

jpike88:
It was said in a different PR, but _preAssignBidnings_ has been reverted and is now back in AngularJS 1.7.2

gkalpak:
@jpike88, to be clear, _preAssignBindings_ have not been reverted in AngularJS. I.e. AngularJS 1.7.x _does not_ support pre-assigning bindings on components and directives that are compiled through the built-in `$compile` service.

What was reverted, were some internal, private APIs (no longer used by AngularJS itself), that are used by AngularJS Material to implement pre-assigning bindings in their `$mdCompiler` service. This means, that pre-assigning is only possible for components/directive that go through `ngMaterial`'s `$mdCompiler` (such as `mdDialog`s, etc.).

Splaktar:
As @mgol mentioned, there is no current version of AngularJS Material that officially supports AngularJS 1.7.x. AngularJS Material 1.1.10 will be the first version that supports it.

Our [current documentation](https://material.angularjs.org/latest/api/service/$mdCompilerProvider#mdcompilerprovider-respectpreassignbindingsenabled-respected) (1.1.9) for `$mdCompilerProvider.respectPreAssignBindingsEnabled(true)` says the following:
> >=1.7 - the compiler calls the constructor first before assigning bindings and $compileProvider.preAssignBindingsEnabled() no longer exists.

That is the recommended configuration, but it only covers half of the options. If `$mdCompilerProvider.respectPreAssignBindingsEnabled(false)` then
> If disabled (`false`), the compiler assigns the value of each of the bindings to the properties of the controller object before the constructor of this object is called.

So it seems like we need to support this `$mdCompilerProvider.respectPreAssignBindingsEnabled(false)`, at least until 1.2.0. We should probably mark it deprecated in this update though.

mgol:
> So it seems like we need to support this `$mdCompilerProvider.respectPreAssignBindingsEnabled(false)`, at least until 1.2.0. We should probably mark it deprecated in this update though.

Deprecating it is tricky as this means you're discouraging people from using it - and not using it means having the default of `false` which is not desirable...

I guess you can deprecate setting it to `false` but since that's the default I don't imagine a lot of people specifying it...

Splaktar:
@mgol good point, we can't use `@deprecated` on the API. I'll just add to the existing description:
> If disabled (`false`), the compiler assigns the value of each of the bindings to the properties of the controller object before the constructor of this object is called.

With the following:
> The ability to disable this settings is **deprecated** and will be removed in AngularJS Material 1.2.0.

Splaktar:
PR updated, description in OP updated, commit msg updated. PTAL.

src/core/services/compiler/compiler.js
```diff
@@ -444,27 +454,33 @@ function MdCompilerProvider($compileProvider) {
...
+    // TODO remove the use of this third argument in AngularJS Material 1.2.0.
     // Passing `true` as the third argument causes `$controller` to return a function that
-    // gets the controller instance instead returning of the instance directly. When the
+    // gets the controller instance instead of returning the instance directly. When the
     // controller is defined as a function, `invokeCtrl.instance` is the *same instance* as
-    // `invokeCtrl()`. However, then the controller is an ES6 class, `invokeCtrl.instance` is a
+    // `invokeCtrl()`. However, when the controller is an ES6 class, `invokeCtrl.instance` is a
     // *different instance* from `invokeCtrl()`.
-    var invokeCtrl = this.$controller(options.controller, injectLocals, true, options.controllerAs);
+    var invokeCtrl = this.$controller(options.controller, injectLocals, true);
```
mgol: Could you switch the logic to not pass the third parameter if `getPreAssignBindingsEnabled() === false`? This would be safer as in the recommended usage no private API would be used.

Splaktar: Good call. I think it makes the 1.2.0 update more straightforward as well.

Splaktar: PTAL and let me know if I missed anything.

src/core/services/compiler/compiler.js
```diff
@@ -444,29 +454,42 @@ function MdCompilerProvider($compileProvider) {
...
     }
 
-    // Instantiate and initialize the specified controller.
-    var ctrl = invokeCtrl();
+    if (options.controllerAs) {
+      injectLocals.$scope[options.controllerAs] = ctrl;
+    }
 
-    if (!getPreAssignBindingsEnabled() && options.bindToController) {
+    if (!preAssignBindingsEnabled && options.bindToController) {
```
gkalpak: Nit: this could be moved inside the the `else` block above.

PR #11330, opened 2018-06-18
fix(layout): allow flex-offset to be a child of layout-margin

### LINKED ISSUE #11328: flex-ma

https://codepen.io/anon/pen/VdMbQB

1. Create an element with `<element layout="row" layout-margin layout-padding>`
2. Fill that element with a `<element flex-offset="75" flex="25">

The element with `flex-offset` should be offset to the right.

The CSS class `.layout-margin` overwrites the **margin-left** declaration from `.flex-offset-*`, so it doesn't get offset to the right.

`flex-offset` does not work if it has a parent with `layout-margin`, so changing this behavior would allow a user to use both directives.

- AngularJS: all
- AngularJS Material: * - 1.1.9
- OS: All (Tested on OSX)
- Browsers: All (tested on FF 61)
 
Changing the flex-offset mixin https://github.com/angular/material/blob/master/src/core/style/layout.scss#L48 fixed the issue for me. (also line 65 and 81), adding in **.layout-margin** before .flex-offset makes this selector more specific and stops it from being overwritten.

```scss
 $offsets : '.offset-#{$suffix}, .flex-offset-#{$suffix}, .layout-margin .flex-offset-#{$suffix}, .layout-margin .offset-#{$suffix}';
```

bersLucas:
Adds .layout-margin to the layout-offset selector for more specificity

Fixes #11328

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

The CSS class `.layout-margin` overwrites the margin-left declaration from `.flex-offset-*`, so it doesn't get offset to the right.

Issue Number:  #11328

## What is the new behavior?

`.flex-offset` now includes `.layout-margin` in it's CSS selector to increase specificity and not have it's margin overwritten by `.layout-margin` when it is a child of `.layout-margin`

```
[ ] Yes
[x] No
```


bersLucas:
I signed it!

src/core/style/layout.scss
```diff
@@ -45,7 +45,7 @@
       @if $s != '' { $suffix : '#{$s}-#{$i * 5}'; }
       @else        { $suffix : '#{$i * 5}';       }
 
-      $offsets : '.offset-#{$suffix}, .flex-offset-#{$suffix}';
+       $offsets : '.offset-#{$suffix}, .flex-offset-#{$suffix}, .layout-margin .flex-offset-#{$suffix}, .layout-margin .offset-#{$suffix}';
```
Splaktar: It looks like the extra space at the start of the line breaks the alignment of the code here? Can you please remove that one space?

PR #11340, opened 2018-06-25
fix(css): remove bloat from autoprefixing of unsupported browsers

### LINKED ISSUE #9396: docs: improve browser support documentation
**Actual Behavior**:
-  `What is the issue? *` README says: "Angular Material is targeted for all browsers with versions n-1; where n is the current browser version."
-  `What is the expected behavior?` It should spell out supported IE versions explicitly. It's not clear what "n-1 where n is the current browser version" means with respect to IE as IE, while version 11 still receives security fixes, is not developed anymore and has been succeeded by Microsoft Edge. This sentence could both be understood as that IE 10+ is supported (as current IE is 11) or as if no IE was supported (since IE has been superseded by Edge and last 2 Edge versions are 13 and 14). I also think it'd be better to spell out concrete browser versions that this project supports; otherwise someone can find a niche browser (like Konqueror with the KHTML engine) and complain we don't support it's latest or previous version.

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

We've long claimed support for n-1 browser versions, but that hasn't been kept up to date. As a result, we've got our autoprefixer settings generating CSS for Safari 8, 9, 10.0, and some dead browsers like Blackberry, Opera Mobile 12.1, and IE 10. This significantly bloats our CSS.
With our old settings we only supported 86.6% of the global browser market.

Issue Number: 
Fixes #9396

## What is the new behavior?
- Unminified CSS reduced by 135 KB
- We now support 90.93% of the global browser market
- use recommended configuration for autoprefix and browserslist
- use a more accurate screenshot of browser support
- update README with link to browserslist settings in package.json
- update README with link to browserl.ist page that shows our config
- add browser support info to AngularJS Material home page

```
[ ] Yes
[x] No
```

## Other information
Users of Safari 8, 9, 10.0, Blackberry, Opera Mobile 12.1, and IE 10 may consider this a breaking change, but this follows our past browser support guidance so it should not be considered breaking. That said, I am still going to push this to 1.1.11 instead of putting it into 1.1.10 as we have some other significant CSS changes in that release that I don't want to mix with these.

screen shot 2018-06-25 at 4 49 21 pm
screen shot 2018-06-25 at 6 46 03 pm
screen shot 2018-06-25 at 6 46 28 pm

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,10 +18,15 @@ Quick Links:
 *  [Installing](#installing)
 
 
-Please note that using AngularJS Material requires the use of **[AngularJS](https://angularjs.org/) 1.4.x** or higher.
-AngularJS Material is targeted for the browser versions shown below in the green boxes:
+Please note that using AngularJS Material requires the use of **[AngularJS](https://angularjs.org/)
+1.4.x** or higher.
 
-![ngm1_browser_support](https://user-images.githubusercontent.com/3506071/35176284-1419c42c-fd46-11e7-9381-d93e5c5db39a.png)
+AngularJS Material is targeted for the browser versions defined in the `broswerslist` field
+of our [package.json](package.json). Below is an screenshot from 
+[browserl.ist](http://browserl.ist/?q=%3E+0.5%25%2C+last+2+versions%2C+Firefox+ESR%2C+not+ie+%3C%3D+10%2C+not+ie_mob+%3C%3D+10%2C+not+bb+%3C%3D+10%2C+not+op_mob+%3C%3D+12.1)
+that provides a visual representation of this configuration:
+
+![AngularJS Material Browser Support](https://user-images.githubusercontent.com/3506071/41875080-d3096d7a-7897-11e8-8838-2bf7473c9502.png)
 
 ## <a name="demos"></a> Online Documentation and Demos
 
--- a/docs/app/css/style.css
+++ b/docs/app/css/style.css
@@ -884,6 +884,66 @@ table.md-css-table td p {
   padding-bottom: 10px;
 }
 
+table.custom-table {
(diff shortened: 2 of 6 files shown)


mgol:
If you expand the table to specific browser versions that will be constantly getting out of date and, in turn, confuse people. I'd collapse all "last 2 versions" browsers into one row and add a restriction that only ones with certain usage (> 0.5%) are included; "last 2 iOS versions" is different from "iOS 11 & 11.3" as the latter is fixed and the former changes over time.

jelbourn:
I agree that it would be better to avoid using specific versions unless it's a special case (like IE11)

Splaktar:
I want to include specific versions for certain browsers where their versioning strategy is inconsistent or odd. Many people ask for an actual translation of what our browser policy is in specific versions and in some cases, it's too confusing to explain it without just listing the version numbers. I've collapsed many of the browsers into shared rows as suggested.

mgol:
@Splaktar In that case, perhaps it'd be nice to have the versions of those browsers generated in the build process? Then, once a new browserslist comes out and is included in AngularJS Material, the table would update automatically. Otherwise it will keep getting out of date.

Splaktar:
`angular-material.min.css`
- Before 398 KB
- After 360 KB

Reduced size of the minified CSS - **38 KB**

Splaktar:
@mgol let's handle the automatic updates in a separate issue as I think that has a different priority than getting the initial documentation in place.

mgol:
Cool, makes sense.

PR #11315, opened 2018-06-06
fix(md-svg-icon): id mangling on cache destroys xhref references

ystreibel:
update all kind of reference in svg when clone ids for caching

Fixes #8689

## PR Checklist
Please check that your PR fulfills the following requirements:
- [X] The commit message follows [our guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format)
- [X] Tests for the changes have been added or this is not a bug fix / enhancement
- [X] Docs have been added, updated, or were not required

## PR Type
What kind of change does this PR introduce?

```
[X] Bugfix
[ ] Enhancement
[ ] Documentation content changes
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: #8689 

## What is the new behavior?
Updates references in svg tree attributes when corresponding to the caching ids.

## Does this PR introduce a breaking change?
```
[ ] Yes
[X] No
```

PR #11342, opened 2018-06-26
fix(icon): stop breaking id references when caching icon ids

### LINKED ISSUE #8689: svg-icon: id mangling on cache destroys xhref references
When a svg with internal def ID references is created in an ephemeral element (dialog, bottomsheet, etc), every subsequent cached load causes the id references to fail lookup (as the original svg is no longer present to reference).

Raw SVG:

``` svg
<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="100" height="100">
<defs>
 <g id="bg" fill="#000000"><path d="M50,100C22.43,100,0,77.57,0,50C0,22.43,22.43,0,50,0c27.57,0,50,22.43,50,50C100,77.57,77.57,100,50,100z M50,2C23.533,2,2,23.533,2,50c0,26.468,21.533,48,48,48c26.468,0,48-21.532,48-48C98,23.533,76.468,2,50,2z"/></g> 
</defs>
<use x="0" y="0" class="bg" xlink:href="#bg"/>
</svg>
```

After material caching:

``` svg
<svg viewBox="0 0 100 100" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" width="100%" height="100%" fit="" preserveAspectRatio="xMidYMid meet" focusable="false">
<defs>
 <g id="bg_cache53" fill="#000000"><path d="M50,100C22.43,100,0,77.57,0,50C0,22.43,22.43,0,50,0c27.57,0,50,22.43,50,50C100,77.57,77.57,100,50,100z M50,2C23.533,2,2,23.533,2,50c0,26.468,21.533,48,48,48c26.468,0,48-21.532,48-48C98,23.533,76.468,2,50,2z"></path></g> 
</defs>
<use x="0" y="0" class="bg" xlink:href="#bg"></use>
</svg>
```

Note that the xlink:href tags now point to an id that no longer exists.

**Angular Versions**:
- Angular: 1.5.5
- Material: 1.1.0-rc.5

Specifically, it's this bit of code:
https://github.com/angular/material/blob/master/src/components/icon/js/iconService.js#L450,L456

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

SVGs with embedded `id`s can be broken when they are read out of the icon cache.

Issue Number: 
Fixes #8689

## What is the new behavior?
SVGs with embedded `id`s can be read out of the icon cache without breaking because the `id` references are now updated in addition to the `id`s themselves.

```
[ ] Yes
[x] No
```

## Other information
Thanks to @ystreibel for starting this work in https://github.com/angular/material/pull/11315!

### DIFF
--- a/src/components/icon/icon.spec.js
+++ b/src/components/icon/icon.spec.js
@@ -419,7 +419,11 @@ describe('MdIcon service', function() {
     $scope = $rootScope;
 
     $templateCache.put('android.svg'    , '<svg><g id="android"></g></svg>');
-    $templateCache.put('angular-logo.svg','<svg><g id="angular"></g><defs><filter id="shadow"></filter></defs><path filter="url(#shadow)"></path></svg>');
+    $templateCache.put('angular-logo.svg',
+      '<svg><g id="angular"></g><defs><filter id="shadow"></filter>' +
+      '<g id="bg" fill="#000000"><path d="M10 10"/></g></defs>' +
+      '<path filter="url(#shadow)"></path><use x="0" y="0" xlink:href="#bg"></use>' +
+      '</svg>');
     $templateCache.put('social.svg'     , '<svg><g id="s1"></g><g id="s2"></g></svg>');
     $templateCache.put('symbol.svg'     , '<svg><symbol id="s1"></symbol><symbol id="s2" viewBox="0 0 32 32"></symbol></svg>');
     $templateCache.put('core.svg'       , '<svg><g id="c1"></g><g id="c2" class="core"></g></svg>');
@@ -596,9 +600,13 @@ describe('MdIcon service', function() {
         $scope.$digest();
 
         $mdIcon('angular-logo.svg').then(function(el) {
-          expect(el.querySelector('defs').firstChild.id).toMatch(/.+_cache[0-9]+/g);
-          expect(el.querySelector('path').attributes.filter.value.split(/url\(#(.*)\)$/g)[1]).toMatch(/.+_cache[0-9]+/g);
-          expect(el.querySelector('defs').firstChild.id === el.querySelector('path').attributes.filter.value.split(/url\(#(.*)\)$/g)[1]);
+          expect(el.querySelector('defs').firstChild.id).toMatch(/.+_cache\d+/g);
+          expect(el.querySelectorAll('path')[1].attributes.filter.value.split(/url\(#(.*)\)$/g)[1])
+            .toMatch(/.+_cache[0-9]+/g);
+          expect(el.querySelectorAll('path')[1].attributes.filter.value.split(/url\(#(.*)\)$/g)[1])
+            .toEqual(el.querySelector('defs').firstChild.id);
+          expect(el.querySelector('use').attributes['xlink:href'].value.split(/#(.*)/)[1]).toEqual(
+            el.querySelector('defs').children[1].id);
         });
 
         $scope.$digest();
--- a/src/components/icon/icon.spec.js
+++ b/src/components/icon/icon.spec.js
@@ -419,6 +419,7 @@ describe('MdIcon service', function() {
     $scope = $rootScope;
 
     $templateCache.put('android.svg'    , '<svg><g id="android"></g></svg>');
+    $templateCache.put('angular-logo.svg','<svg><g id="angular"></g><defs><filter id="shadow"></filter></defs><path filter="url(#shadow)"></path></svg>');
     $templateCache.put('social.svg'     , '<svg><g id="s1"></g><g id="s2"></g></svg>');
     $templateCache.put('symbol.svg'     , '<svg><symbol id="s1"></symbol><symbol id="s2" viewBox="0 0 32 32"></symbol></svg>');
     $templateCache.put('core.svg'       , '<svg><g id="c1"></g><g id="c2" class="core"></g></svg>');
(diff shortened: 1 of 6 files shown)


src/components/icon/icon.spec.js
```diff
@@ -587,6 +592,25 @@ describe('MdIcon service', function() {
...
       });
+
+      it('should suffix duplicated ids and refs', function() {
+        // Just request the icon to be stored in the cache.
+        $mdIcon('angular-logo.svg');
+
+        $scope.$digest();
+
+        $mdIcon('angular-logo.svg').then(function(el) {
+          expect(el.querySelector('defs').firstChild.id).toMatch(/.+_cache[0-9]+/g);
```
jelbourn: `[0-9]` is equivalent to `\d`

src/components/icon/js/iconService.js
```diff
@@ -447,38 +447,58 @@ function MdIconService(config, $templateRequest, $q, $log, $mdUtil, $sce) {
...
     var useDefault = angular.isUndefined(alias) || !(alias && alias.length);
-    if (useDefault) return config.defaultFontSet;
+    if (useDefault) {
+      return config.defaultFontSet;
+    }
 
     var result = alias;
     angular.forEach(config.fontSets, function(it) {
-      if (it.alias == alias) result = it.fontSet || result;
+      if (it.alias === alias) {
```
jelbourn: Rename `it` while you're here? I can't tell what it was ever supposed to signify

src/components/icon/js/iconService.js
```diff
@@ -447,38 +447,58 @@ function MdIconService(config, $templateRequest, $q, $log, $mdUtil, $sce) {
...
-    angular.forEach(clone.querySelectorAll('[id]'), function(item) {
-      item.id += cacheSuffix;
+    // For each cached icon, we need to modify the id attributes and references.
+    // This is needed because SVG ids are treated as normal DOM ids and should not be duplicated on
+    // the page.
+    if (clone.id) {
+      clone.id += cacheSuffix;
+    }
+
+    var addCacheSuffixToId = function(match, p1, p2, p3) {
```
jelbourn: Could this be moved out so that it's always around instead of being redefined every time `transformClone` is called?

src/components/icon/js/iconService.js
```diff
@@ -447,38 +447,58 @@ function MdIconService(config, $templateRequest, $q, $log, $mdUtil, $sce) {
...
+      return [p1, p2, cacheSuffix, p3].join('');
+    };
+    angular.forEach(clone.querySelectorAll('[id]'), function(descendantElem) {
+      descendantElem.id += cacheSuffix;
     });
+    // Inject the cacheSuffix into all instances of url(id) and xlink:href="#id".
```
jelbourn: Can you add a comment that explains why there's no xss risk?

ystreibel:
Hi everyone! You could use my commit for this PR!

src/components/icon/js/iconService.js
Splaktar: It depends on `cacheSuffix` which has to be generated with a new `Uid` every time `transformClone` is called. I don't think that I want to pull `cacheSuffix` out to be service-wide since it's only used in this function.

src/components/icon/js/iconService.js
Splaktar: Is there some way that https://github.com/angular/material/blob/master/src/core/util/util.js#L1-L7 could be vulnerable to XSS? Since that is what we'll be coercing to a `string`, sticking into `innerHTML`, and then calling `++` on.

jelbourn: I pretty much meant a comment that says just that.

Guard also also still useful in cases like this, though, in case someone else down the line changes the way unique ids are generated

Splaktar: OK I added a comment just below this.

I'm not clear what you meant by "Guard also also still useful". Perhaps that I should add a comment about this in `$mdUtil` as well?

jelbourn: I meant something like asserting that the ID matches a certain pattern and throwing if not. Not required, but something to keep in mind.

Splaktar: Good idea. Guard added!

Splaktar:
It looks like @coennijhuis' [CodePen demo](https://codepen.io/anon/pen/jpxBZx) for this shows that we are no longer applying unique ids for each SVG as we need to. I'll investigate this some more.

Splaktar:
This is an [updated version of the CodePen](https://codepen.io/Splaktar/pen/QBrmop?editors=1010#0) which uses the code from this branch.

coennijhuis:
I do not see a change in behavior. The updated CodePen also has 5 identical IDs. It looks like the IconService has a caching mechanism that is not 100%. 

I believe the icon is cached in the following method (line 794 in IconService.js):

function cacheIcon(id) {

    return function updateCache(icon) {
      iconCache[id] = isIcon(icon) ? icon : new Icon(icon, config[id]);

      return iconCache[id].clone();
    };
  }

Shouldn't there be a call to transformClone in this method?

coennijhuis:
@Splaktar: is there any progress regarding this issue? Or is there anything I can do to help?

Splaktar:
@coennijhuis sorry for the delay. I've been busy digging into P1, a11y, and regression issues. I plan to revisit this again soon.

Splaktar:
@coennijhuis if you would like to investigate this and propose another PR, I would be happy to help you review, test, and get it merged. I am currently focused on P1 and a11y issues.

EmielH:
@Splaktar, @coennijhuis and me made a new PR for this issue: #11465. Could you take a look at this PR?

Splaktar:
@jelbourn I think that this has some presubmit failures at one point, but it's not clear what they were. I've since rebased it and made some minor changes. Can you please provide some details behind any presubmit failures if they still exist?

This PR fixes the mangling of embedded ids and dependencies within SVGs. 
PR https://github.com/angular/material/pull/11465 fixes the issue of multiple SVGs on the page having the same ids (mentioned above and tracked in #11395).

josephperrott:
Confirmed CLA manually.

Splaktar:
This is causing 1 presubmit test failure that needs to be investigated and feedback provided so that a fix can be considered.

Splaktar:
It looks like our move to less verbose RegEx, in place of very verbose `querySelectorAll` strings, is causing a [performance problem with large SVG files](https://github.com/angular/material/issues/11651).

PR #11352, opened 2018-06-29
fix(input): remove unnecessary warnings when ng-messages not provided

### LINKED ISSUE #10461: input: ng-messages animation errors
screen shot 2017-03-06 at 3 44 25 pm

screen shot 2017-03-06 at 3 44 11 pm

Visible with ng 1.5.x and 1.6.x

marosoft:
There is no requirement to include the messages in case of an error.
Highlighting the field can be enough.

Fixes #10461

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

At the moment warning are displayed in the console when ng-messages are not specified for the input field.

Issue Number:
Relates to #10461 
Relates to #9468

## What is the new behavior?
The warnings are not generated as it is not required to include the error messages.
The original fix for #9468 is still in place. The other changes introduced together with that fix were removed.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/input/input.js
+++ b/src/components/input/input.js
@@ -26,8 +26,6 @@ if (window._mdMocksIncluded) {
     return {
       // special accessor to internals... useful for testing
       messages: {
-        show        : showInputMessages,
-        hide        : hideInputMessages,
         getElement  : getMessagesElement
       }
     };
@@ -687,7 +685,7 @@ function mdMaxlengthDirective($animate, $mdUtil) {
       attr.$observe('ngTrim', function (value) {
         ngTrim = angular.isDefined(value) ? $mdUtil.parseAttributeBoolean(value) : true;
       });
- 
+
       scope.$watch(attr.mdMaxlength, function(value) {
         if (angular.isNumber(value) && value > 0) {
           if (!charCountEl.parent().length) {
@@ -945,10 +943,10 @@ function ngMessageDirective($mdUtil) {
   }
 }
 
-var $$AnimateRunner, $animateCss, $mdUtil, $log;
+var $$AnimateRunner, $animateCss, $mdUtil;
 
-function mdInputInvalidMessagesAnimation($$AnimateRunner, $animateCss, $mdUtil, $log) {
-  saveSharedServices($$AnimateRunner, $animateCss, $mdUtil, $log);
+function mdInputInvalidMessagesAnimation($$AnimateRunner, $animateCss, $mdUtil) {
+  saveSharedServices($$AnimateRunner, $animateCss, $mdUtil);
 
   return {
     addClass: function(element, className, done) {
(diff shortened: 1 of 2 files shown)


src/components/input/input.js
```diff
@@ -1007,15 +1005,8 @@ function ngMessageAnimation($$AnimateRunner, $animateCss, $mdUtil, $log) {
 function showInputMessages(element, done) {
   var animators = [], animator;
   var messages = getMessagesElement(element);
-  var children = messages.children();
-
-  if (messages.length == 0 || children.length == 0) {
-    $log.warn('mdInput messages show animation called on invalid messages element: ', element);
-    done();
-    return;
-  }
```
Splaktar: I can understand removing the warning, but this seems to also be removing the `done();` call and the early `return`. I think that we still want those things for performance and correctness. We don't need to trigger the animations if there are no messages to animate in.

src/components/input/input.js
```diff
@@ -1027,15 +1018,8 @@ function showInputMessages(element, done) {
 function hideInputMessages(element, done) {
   var animators = [], animator;
   var messages = getMessagesElement(element);
-  var children = messages.children();
-
-  if (messages.length == 0 || children.length == 0) {
-    $log.warn('mdInput messages hide animation called on invalid messages element: ', element);
-    done();
-    return;
-  }
```
Splaktar: Same here. We don't need to animate the messages out if there are no messages.

src/components/input/input.js
marosoft: This makes sense, thanks. I will implement those changes.

As the `animators` array is empty it should not cause any performance issues. However for the correctness it makes sense to handle it here rather than depend on the behaviour of the called `all` function.

I will fix it as soon as possible.

src/components/input/input.js
```diff
@@ -1007,12 +1005,11 @@ function ngMessageAnimation($$AnimateRunner, $animateCss, $mdUtil, $log) {
 function showInputMessages(element, done) {
   var animators = [], animator;
   var messages = getMessagesElement(element);
-  var children = messages.children();
+  var children = messages.children();
```
Splaktar: Can you please remove the added whitespace at the end of this line?

src/components/input/input.js
```diff
@@ -1027,12 +1024,11 @@ function showInputMessages(element, done) {
 function hideInputMessages(element, done) {
   var animators = [], animator;
   var messages = getMessagesElement(element);
-  var children = messages.children();
+  var children = messages.children();
```
Splaktar: Can you please remove the added whitespace at the end of this line?

marosoft:
@Splaktar just wanted to highlight that I squashed the commits

PR #11385, opened 2018-07-29
feat(select): support customizing the separation of selected options in multiple mode

cobuscoetsee:
Update select.js multiselect function to give control over the separator.
User can add `multiselect-separator="<br/>"`  as attribute

## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Issue Number: 

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -74,7 +74,8 @@ angular.module('material.components.select', [
  * @param {boolean=} multiple When present, allows for more than one option to be selected.
  *  The model is an array with the selected choices. **Note:** This attribute is only evaluated
  *  once; it is not watched.
- * @param {string=} multiple-separator Determines how multiple values will be seperated, defaults to ', '.
+ * @param {string=} selection-separator-char Options available is limited to `comma`(Default),`semi-colon`, `pipe`.
+ * @param {string=} selection-separator-spacing Options available is limited to `after`(Default), `before`, `none`, `break`.
  * @param {expression=} md-on-close Expression to be evaluated when the select is closed.
  * @param {expression=} md-on-open Expression to be evaluated when opening the select.
  * Will hide the select options and show a spinner until the evaluated promise resolves.
@@ -132,6 +133,18 @@ angular.module('material.components.select', [
  *       <md-select-header>
  *         <span> Neighborhoods - </span>
  *       </md-select-header>
+ *       <md-option ng-value="opt" ng-repeat="opt in neighborhoods2">{{ opt }}</md-option>
+ *     </md-select>
+ *   </md-input-container>
+ * </hljs>
+ * Multiselect with a custom separator
+ *
+ * When a developer needs to add a seperator with multi select, adding to the attribute `selection-separator-char` or
+ `selection-separator-spacing` with options. The example below demonstrate a new line for after each selected item.
+ *
+ * <hljs lang="html">
+ *   <md-input-container>
+ *     <md-select ng-model="someModel" selection-separator-char="none" selection-separator-spacing="break">
  *       <md-option ng-value="opt" ng-repeat="opt in neighborhoods2">{{ opt }}</md-option>
  *     </md-select>
  *   </md-input-container>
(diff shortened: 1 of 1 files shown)


cobuscoetsee:
I signed it!

src/components/select/select.js
```diff
@@ -259,7 +260,9 @@ function SelectDirective($mdSelect, $mdUtil, $mdConstant, $mdTheming, $mdAria, $
     }
 
     var isMultiple = $mdUtil.parseAttributeBoolean(attr.multiple);
-
+    //Defaults ','
+    var multipleSeparator = attr.multipleSeparator || ', ';
```
Splaktar: I don't think that we need to support an infinite list of options here. Plus doing so seems to expose us to XSS (see below).

There is no guidance from [the MD spec](https://material.io/design/components/text-fields.html#anatomy) on this currently.

Let's create a finite list of options that can be passed into the attribute like the following

- comma (default)
- semi-colon
- `|` whatever that is called...
- space
- plus

I'm open to expanding this list if any other ideas come to mind. However, it won't be too hard to add to it later if we need to.

That should probably be called `selection-separator-char`.

Then for `selection-separator-spacing`, we should probably allow the following

- after (default)
- before
- none
- break - where this would ignore the `selection-separator-char`

I think that these 2 sets of options give us a ton of flexibility without causing us too much of a maintenance burden or exposing us to XSS.

src/components/select/select.js
```diff
@@ -678,6 +681,7 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
   function SelectMenuController($scope, $attrs, $element) {
     var self = this;
     self.isMultiple = angular.isDefined($attrs.multiple);
+    self.multipleSeparator = $attrs.multipleSeparator || ', ';
```
Splaktar: This looks like the right place to initialize this variable. The one in the `SelectDirective` doesn't seem to be used and can be removed?

src/components/select/select.js
```diff
@@ -826,7 +830,7 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
         }
 
         // Ensure there are no duplicates; see https://github.com/angular/material/issues/9442
-        return $mdUtil.uniq(selectedOptionEls.map(mapFn)).join(', ');
+        return $mdUtil.uniq(selectedOptionEls.map(mapFn)).join(self.multipleSeparator);
```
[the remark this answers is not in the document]
cobuscoetsee: Thanks will update as suggested. :)

cobuscoetsee:
Updated code let me know what you think? See example here.

src/components/select/select.js
```diff
@@ -259,15 +273,39 @@ function SelectDirective($mdSelect, $mdUtil, $mdConstant, $mdTheming, $mdAria, $
...
+        default:
+            selectionSeparator = selectionSeparatorChar+' ';
+    }
+   
     // Use everything that's left inside element.contents() as the contents of the menu
     var multipleContent = isMultiple ? 'multiple' : '';
     var selectTemplate = '' +
       '<div class="md-select-menu-container" aria-hidden="true" role="presentation">' +
-      '<md-select-menu role="presentation" {0}>{1}</md-select-menu>' +
+      '<md-select-menu role="presentation" {0} selection-separator="{1}">{2}</md-select-menu>' +
```
Splaktar: I would rather not have to apply this again here. Was `$attrs.selectionSeparator` not defined on https://github.com/angular/material/pull/11385/files#diff-d7b463f733c2e3fbc783f38073561b74R719 without this?

src/components/select/select.js
```diff
@@ -259,15 +273,39 @@ function SelectDirective($mdSelect, $mdUtil, $mdConstant, $mdTheming, $mdAria, $
...
+      'comma' : ',',
+      'semicolon' : ';',
+      'pipe' : '|',
+      'none' : ''
+    };
+    var selectionSeparatorChar = selectionSeparatorCharOptions[attr.selectionSeparatorChar || 'comma'];
+    //Set the different spacing options with the selectionSeparatorChar
+    var selectionSeparator = ''; 
+    switch(attr.selectionSeparatorSpacing) {
+        case 'before':
```
Splaktar: Please update all indentation to be 2 spaces instead of 4.

src/components/select/select.js
```diff
@@ -259,15 +273,39 @@ function SelectDirective($mdSelect, $mdUtil, $mdConstant, $mdTheming, $mdAria, $
     }
 
     var isMultiple = $mdUtil.parseAttributeBoolean(attr.multiple);
-
+       
+    //Set the separation options in a list
```
Splaktar: Please add a space after the start of all comments, I.e. `// Set`

src/components/select/select.js
```diff
@@ -135,6 +137,18 @@ angular.module('material.components.select', [
  *     </md-select>
  *   </md-input-container>
  * </hljs>
+ * Multiselect with a custom separator
+ *
+ * When a developer needs to add a seperator with multi select, adding to the attribute `selection-separator-char` or
+ `selection-separator-spacing` with options. The example below demonstrate a new line for after each selected item.
```
Splaktar: "The example below demonstrates how to use a new line after each selected item."

src/components/select/select.js
```diff
@@ -135,6 +137,18 @@ angular.module('material.components.select', [
...
+ *
+ * <hljs lang="html">
+ *   <md-input-container>
+ *     <md-select ng-model="someModel" selection-separator-char="none" selection-separator-spacing="break">
```
Splaktar: For the line break example, let's just omit the `selection-separator-char="none"` and verify that the code works in that case. We'll probably want a unit test for that.

src/components/select/select.js
```diff
@@ -135,6 +137,18 @@ angular.module('material.components.select', [
...
+ *       <md-option ng-value="opt" ng-repeat="opt in neighborhoods2">{{ opt }}</md-option>
+ *     </md-select>
+ *   </md-input-container>
+ * </hljs>
```
Splaktar: We should add another example that uses `selection-separator-char="semicolon"` and `selection-separator-spacing="after"`.

src/components/select/select.js
```diff
@@ -74,6 +74,8 @@ angular.module('material.components.select', [
  * @param {boolean=} multiple When present, allows for more than one option to be selected.
  *  The model is an array with the selected choices. **Note:** This attribute is only evaluated
  *  once; it is not watched.
+ * @param {string=} selection-separator-char Options available is limited to `comma`(Default),`semi-colon`, `pipe`.
```
Splaktar: `semicolon` instead of `semi-colon` here.

src/components/select/select.js
```diff
@@ -74,6 +74,8 @@ angular.module('material.components.select', [
...
+ * @param {string=} selection-separator-spacing Options available is limited to `after`(Default), `before`, `none`, `break`.
```
Splaktar: Let's display both of these sets of options in a bulleted list like we do [here](https://material.angularjs.org/latest/api/directive/mdMenu#attributes) using [this code](https://github.com/cobuscoetsee/material/blob/8db610681bf06b610b6ade12129deb37564bac27/src/components/menu/js/menuDirective.js#L172-L185).

Splaktar:
Closing as stale. We are also no longer accepting features or enhancements now that we're in LTS>

PR #11390, opened 2018-07-31
feat(panel): allow transform to be animated on an offset panel

### LINKED ISSUE #9905: feat(panel): allow transform to be animated on elements with an offset

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[x] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Previously, if a panel had a position with an offset (e.g. `$mdPanel.newPanelPosition().center()`), the positioning would break any `transform` animations on the panel. This was due to the fact that `mdPanel` uses inline `transform` styles to do the offsetting. 

Issue Number: 
Relates to #9641.
Fixes #9905.

## What is the new behavior?
These changes introduce a wrapper around the panel (`.md-panel-inner-wrapper`), which will handle all of the positioning, allowing for any animations to be applied to the `.md-panel` itself.

```
[ ] Yes
[ ] No
```

## Other information
Moving this over from https://github.com/angular/material/pull/9941 so that rebasing can happen and some additional clean up can be completed. This already went through a couple rounds of reviews in that PR.

TODO
- [x] Fix tooltips are no longer styled properly with these changes
- [x] Do some more manual testing
- [x] Review for breaking changes

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -1306,10 +1306,10 @@ MdPanelService.prototype._closeFirstOpenedPanel = function(groupName) {
 
 
 /**
- * Wraps the users template in two elements, md-panel-outer-wrapper, which
- * covers the entire attachTo element, and md-panel, which contains only the
- * template. This allows the panel control over positioning, animations,
- * and similar properties.
+ * Wraps the user's template in three elements:
+ * - md-panel-outer-wrapper - covers the entire `attachTo` element.
+ * - md-panel-inner-wrapper - handles the positioning.
+ * - md-panel - contains the user's content and deals with the animations.
  * @param {string} origTemplate The original template.
  * @returns {string} The wrapped template.
  * @private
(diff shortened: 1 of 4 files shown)


Splaktar:
@crisbeto can you please confirm to make @googlebot happy?

crisbeto:
I don't think it's down to confirmation, but rather that there are two people on the commit.

Splaktar:
@crisbeto right, that causes the warning. Part of the process is just for you to post here
> We need to confirm that all authors are ok with their commits being contributed to this project. Please have them confirm that here in the pull request.

When the Googler goes to merge this, they just need to check that you are OK with your commits being checked into the project. I'm sure that's not a problem in this case 😁

crisbeto:
I signed it!

src/components/panel/panel.scss
```diff
@@ -31,7 +26,7 @@
 
 .md-panel {
   opacity: 0;
-  position: fixed;
+  position: relative;
```
Splaktar, line 29: This is the change that breaks `md-tooltip`. With this, it causes the `height` style to no longer have any effect

https://github.com/angular/material/blob/e3f977214010aa8217c79dfc732e504d83a654bf/src/components/tooltip/tooltip.scss#L20

I can work around this manually by adding `padding: 5px 8px 6px;` to the `md-tooltip`, but that's not something that I can programatically map `$tooltip-height-sm` or `$tooltip-height-lg` into.

crisbeto: Disclaimer that I was testing this against the docs site without any of these changes so the effect might not be the same.

I think the problem is that `md-tooltip` is a custom element which defaults to `display: inline`. It should work like before if you set `display: block` or `display: inline-block` on `.md-tooltip`.

src/components/panel/panel.js
```diff
@@ -1321,26 +1321,32 @@ MdPanelService.prototype._wrapTemplate = function(origTemplate) {
   // height and width for positioning.
   return '' +
       '<div class="md-panel-outer-wrapper">' +
-      '  <div class="md-panel _md-panel-offscreen">' + template + '</div>' +
+        '<div class="md-panel-inner-wrapper" style="left: -9999px;">' +
```
Splaktar, line 1324: This inline style (`style="left: -9999px;"`) triggers a CSP violation for `unsafe-inline`.

src/components/panel/panel.js
```diff
@@ -1321,26 +1321,32 @@ MdPanelService.prototype._wrapTemplate = function(origTemplate) {
...
+ * proper control over positoning and animations.
  * @param {!angular.JQLite} contentElement Element to be wrapped.
  * @return {!angular.JQLite} Wrapper element.
  * @private
  */
 MdPanelService.prototype._wrapContentElement = function(contentElement) {
-  var wrapper = angular.element('<div class="md-panel-outer-wrapper">');
+  var outerWrapper = angular.element(
+    '<div class="md-panel-outer-wrapper">' +
+      '<div class="md-panel-inner-wrapper" style="left: -9999px;"></div>' +
```
Splaktar, line 1342: This inline style (`style="left: -9999px;"`) triggers a CSP violation for `unsafe-inline`.

PR #11392, opened 2018-08-01
fix(tabs): show a visual indication of tab focus

### LINKED ISSUE #11384: tabs: visual indicator of focus is missing
#### CodePen and steps to reproduce the issue:
When you go through the page using tab key (for example, tabs demo) you can see no indicator of that the md-tabs element have recieved focus. But if you use arrow key to navigate through md-tabs, visual indicator appears, and keeps working properly. 

1. Go to Tabs Demo Page https://material.angularjs.org/1.1.5/demo/tabs
1. Press tab long enough, to get to ONE/TWO/THREE tabs under Dynamic Height example
1. See that there is no visual indicator of focus
1. Use left and write arrows
1. See that now visual indicator of focus appeared

Visual indicator appears immediately that the focus is set on md-tabs element. 

Visual focus is not present than it should be. 

Accessibility guideline: "Confirm that a visual indicator is present for each element that currently has keyboard focus."

- AngularJS: The one used in AngularJS Material demos
- AngularJS Material: 1.1.10 (could see the issue in current demo)
- OS: Mac OS
- Browsers: Chrome, Safari
 
[Screenshot with right and wrong version](https://docs.google.com/a/google.com/presentation/d/e/2PACX-1vS3NMxkq9Ob6N0coSZ3Mp6wSsGOXJSFUhYTaoZGMA0ou_0ipr-HojsFBk2UFroU6_sunKbvY53sIfjG/pub?start=false&loop=false&delayms=3000)

codymikol:
the visual highlighting of focused tabs was not functioning

Fixes #11384

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Currently when tabbing to an md-tab, there is no visual indication of focus.

This was happening because the md-dummy tab used to track has-focus on $mdTabsCtrl with ng-focus, but this was removed in [this](https://github.com/angular/material/commit/072f8328391c884b7ebc1ef49f4dc3e6c02e636c) commit

Issue Number: #11384

## What is the new behavior?

Now when tabbing to an md-tab the proper focused styling will be applied.

There seems to be two things required to fix this issue.

Firstly, ng-focus and ng-blur on md-dummy-tab are used to toggle the $mdTabsCtrl.hasFocus flag. This is important because the md-focused style is applied depending on the truthiness of the tabsController hasFocus method. This hasFocus method is partly dependent on $mdTabsCtrl.hasFocus being true.

Secondly, the tabindex of the md-tabs-canvas is set to 0. I'm really not exactly sure why this is necessary as I would expect the tabIndex to default to 0 and it is otherwise not set from what I can tell. If this is not set explicitly to 0, ng-focus will not be triggered when tabbing to the md-tabs. This means that the $mdTabsCtrl.redirectFocus method will not be called and focusing of the tab will not happen. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/tabs/js/tabsDirective.js
+++ b/src/components/tabs/js/tabsDirective.js
@@ -146,6 +146,7 @@ function MdTabs ($$mdSvgRegistry) {
             '<md-icon md-svg-src="'+ $$mdSvgRegistry.mdTabsArrow +'"></md-icon> ' +
           '</md-next-button> ' +
           '<md-tabs-canvas ' +
+              'tabindex="{{ $mdTabsCtrl.hasFocus ? -1 : 0 }}" ' +
               'ng-focus="$mdTabsCtrl.redirectFocus()" ' +
               'ng-class="{ ' +
                   '\'md-paginated\': $mdTabsCtrl.shouldPaginate, ' +
@@ -184,6 +185,8 @@ function MdTabs ($$mdSvgRegistry) {
               '<md-dummy-tab ' +
                   'class="md-tab" ' +
                   'tabindex="-1" ' +
+                  'ng-focus="$mdTabsCtrl.hasFocus = true" ' +
+                  'ng-blur="$mdTabsCtrl.hasFocus = false" ' +
                   'ng-repeat="tab in $mdTabsCtrl.tabs" ' +
                   'md-tabs-template="::tab.label" ' +
                   'md-scope="::tab.parent"></md-dummy-tab> ' +


src/components/tabs/js/tabsDirective.js
```diff
@@ -146,6 +146,7 @@ function MdTabs ($$mdSvgRegistry) {
             '<md-icon md-svg-src="'+ $$mdSvgRegistry.mdTabsArrow +'"></md-icon> ' +
           '</md-next-button> ' +
           '<md-tabs-canvas ' +
+              'tabindex="0" ' +
```
Splaktar: Please go back to what the old commit did `'tabindex="{{ $mdTabsCtrl.hasFocus ? -1 : 0 }}" ' +`. This allows for shift+tab to escape from the tabs control. With it staying `0` all of the time, shift+tab gets stuck and can't get out of the tabs control.

codymikol:
You're definitely right, changed 👍

PR #11393, opened 2018-08-01
fix(list): account for IE11 bug with flexbox and min-height

### LINKED ISSUE #11177: list-item: layout alignment breaks in IE 11

Bug

md-list-item should consistently center their content across all browsers.

In IE 11 if you add a click event to an md-list-item the content will no longer be centered vertically but rather align itself to the top of the list item.

On the official demos for 1.1.8 https://material.angularjs.org/latest/demo/list

If you look at the "Basic Usage" "Normal" section in chrome you will see that the list items text content is centered veritically

screen shot 2018-03-20 at 3 47 45 pm

But in IE 11 you can see that the content is not centered

screen shot 2018-03-20 at 3 49 18 pm

Browser consistency. It also looks bad and probably doesn't follow the material design spec.

All 1.1.x versions have this issue. The browser is IE 11.

This only happens after a click event is added to the list item, without one it is styled properly.

codymikol:
when using an md-3-line list item, use the mixin workaround for ie11 flexbug https://github.com/philipwalton/flexbugs#flexbug-3

Fixes #11177

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Currently in IE11, md-list-item content vertical alignment can break if the md-list-item has a click event applied to it.

Issue Number: #11177

## What is the new behavior?

Now the alignment will be correct in IE11. This is accomplished by wrapping the applied min-height in the ie11-min-height-flexbug mixin. 

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/list/list.scss
+++ b/src/components/list/list.scss
@@ -163,7 +163,7 @@ md-list-item {
       .md-list-item-inner {
         // The list item content should fill the complete width.
         width: 100%;
-        min-height: inherit;
+        @include ie11-min-height-flexbug(inherit);
       }
 
     }


Splaktar:
Please update the commit message title to be `fix(list): account for IE11 bug with flexbox and min-height`.

Splaktar:
Oh and thank you for reporting and fixing this!

codymikol:
Commit title is fixed, thanks for reviewing so quickly 🎉

PR #11407, opened 2018-08-14
feat(autocomplete): add support for aria-labelledby

Splaktar:
## PR Checklist
Please check that your PR fulfills the following requirements:
- [x] The commit message follows [our guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format)
- [ ] Tests for the changes have been added or this is not a bug fix / enhancement
- [x] Docs have been added, updated, or were not required

## PR Type
What kind of change does this PR introduce?

```
[ ] Bugfix
[x] Enhancement
[ ] Documentation content changes
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?
applying `aria-labelledby` to `md-autocomplete` has no effect because the attribute is not forwarded on to the `input`.

Issue Number: 
Fixes #10815

## What is the new behavior?
add support for `aria-labelledby`

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

PR #11412, opened 2018-08-15
feat(autocomplete): add support for aria-labelledby

### LINKED ISSUE #10815: feat(autocomplete): add support for aria-labelledby
**Actual Behavior**:
 -  `What is the issue? *`
The aria-label and placeholder text are the same in situations where no floatingLabel is defined with md-autocomplete. This means that some screen readers will read off the placeholder text twice (once for the placeholder, and once for the aria-label).
For example:
`<md-autocomplete placeholder="What is your favorite city?"/>`
will create
`<input aria-label="What is your favorite city?" placeholder="What is your favorite city?">`

There should always be a visible label for a combobox, thus there should not be a situation where the placeholder text acts as the label. However, given the constraint that this implementation is already in production, this is the proposed change:
Add an md-aria-labelledby attribute to md-autcomplete. If defined, then return an <input> element that has an aria-labelledby attribute in the place of the aria-label attribute that contains the id of the visible label.
For example:
`<label id="cityLabel">What is your favorite city?</label>`
`<md-autocomplete placeholder="Enter city name" md-aria-labelledby="cityLabel"/>`
would create
`<input ... aria-labelledby="cityLabel" placeholder="Enter city name" >`
Now, screen readers will always read "What is your favorite city?" as the label for the combobox, and some will read the placeholder text (which is extraneous information).
**Solution that will not work:** Removing the aria-label from the <input> element given the current implementation. There always needs to be a label for this combobox, and one cannot rely on a placeholder as a label since many assistive technologies will not read the placeholder. The solution outlined above is a workaround because replacing the aria-label with an aria-labelledby in all cases will be a breaking change for some users.

https://material.angularjs.org/1.1.4/demo/autocomplete#basic-usage
Using Voiceover, navigate to the autocomplete combobox in the Basic Usage section. Notice that "What is your favorite UX state?" is read twice.

**AngularJS Versions**: *
 -  `AngularJS Version:`

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *` Version 58.0.3029.110 (64-bit)
 -  `OS: *` macOS Sieera Version 10.12.5
 -  `Stack Traces:`
 -  `Screen Reader:` Voiceover (not reproducible with ChromevoxNext)

----

Splaktar:
## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

applying `aria-labelledby` to `md-autocomplete` has no effect because the attribute is not forwarded on to the `input`.

Issue Number: 
Fixes #10815

## What is the new behavior?
add support for `aria-labelledby`

```
[ ] Yes
[x] No
```

## Other information
It looks like the previous PR https://github.com/angular/material/pull/11407 had a problem with https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_text_01.

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -96,6 +96,14 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       if ($scope.ariaDescribedBy) {
         elements.input.setAttribute('aria-describedby', $scope.ariaDescribedBy);
       }
+      if (!$scope.floatingLabel) {
+        if ($scope.ariaLabelledBy) {
+          elements.input.setAttribute('aria-labelledby', $scope.ariaLabelledBy);
+        } else {
+          // If no aria-labelledby references are defined, then just label using the placeholder.
+          elements.input.setAttribute('aria-label', $scope.placeholder);
+        }
+      }
     });
   }
 
--- a/src/components/autocomplete/js/autocompleteDirective.js
+++ b/src/components/autocomplete/js/autocompleteDirective.js
@@ -19,7 +19,8 @@ angular
  *     no matches were found.  You can do this by wrapping your template in `md-item-template` and
  *     adding a tag for `md-not-found`.  An example of this is shown below.
  *
- * To reset the displayed value you must clear both values for `md-search-text` and `md-selected-item`.
+ * To reset the displayed value you must clear both values for `md-search-text` and
+ * `md-selected-item`.
  *
  * ### Validation
  *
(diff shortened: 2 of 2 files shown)


Splaktar:
@jelbourn I am actually going to change this once more after doing some further testing on a larger chips a11y PR. I think that applying `input-aria-labelledby` is more accurate. If we keep with just `aria-labelledby`, it seems to cause some duplication with screen readers.

I'll send over the chips PR for your initial review later today.

Splaktar:
@jelbourn OK, this PR is updated. It also adds an `input-aria-label` similar to `md-chips`. This is useful since the placeholder and label are not always the same.

This PR also updates `aria-describedby` to be `input-aria-describedby` for `md-autocomplete`.

It's possible that this could solve the a11y error that was seen by Miles' presubmit. However, the changes were made for general correctness after doing some additional screen reader testing with chips. 

It's still possible to apply an `aria-label` to the `md-autocomplete` and then a different `input-aria-label` which would apply that string to the `aria-label` of the embedded `input`.

Splaktar:
Feedback on presubmit failures: This PR "is failing an accessibility check in one project where an element that formerly had a label now doesn't. Need to check that the change isn't overwriting an aria-label with nothing."

I'll try to reproduce this in a unit test.
- [ ] check that existing aria-labels are not lost/overwritten with nothing

Splaktar:
I added a bunch of tests to verify that an existing `aria-label` on the `md-autocomplete` would not be overwritten. I was not able to reproduce the problem described by Jeremy in the presubmit for 1 project. I made one change that might address the issue, but if it fails presubmit again, I'll need a more detailed description of the template that is failing the test and exactly where the `aria-label` is being applied.

mmalerba:
I'm seeing a couple of these errors:

```
Error: AX_TEXT_01 (Controls and media elements should have labels) failed on the following element:
#input-0
See https://github.com/GoogleChrome/accessibility-developer-tools/wiki/Audit-Rules#ax_text_01 for more information.
```

Splaktar:
@mmalerba I'll need more detailed information. As mentioned above, I've written up a lot of test cases for this. Please let me know the use case that is failing so that I can add another unit test that covers it.

Splaktar:
Bumped this to P1 as getting this in is blocking some work on P1 issues.

Splaktar:
Rebased.

Splaktar:
Rebased ✅

Splaktar:
Rebased ✔️

Issue #10850, opened 2017-08-14
chips(readonly): chips can be selected/focused via keyboard


EamonKavanagh:
**Actual Behavior**:
Setting `readonly="true"` on an `md-chip` component causes tab/arrow navigation to function incorrectly. In an example with 5 chips, the expected behavior would be pressing the right arrow would cycle through the chips in left to right order before wrapping back to the start.

**CodePen** (or steps to reproduce the issue): * https://material.angularjs.org/latest/demo/chips
In the same example with 5 chips, the actual behavior is:

1) First right arrow key press focuses on the first chip 
2) Second right arrow key press 'selects' the second chip without focusing on it (no visual indication of selection but selection can be verified using ChromeVox) 
3) Third right arrow key press 'selects' the first chip, with no visual indication as in 2) 
4) Forth right arrow key press focuses on first chip as in 1), cycling back to the start.

Pressing the left arrow keys does the above 4 steps in reverse so the middle chip is not accessible.

**Versions**: *
AngularJS: 1.6.4-local+sha.617b36117
AngularJS Material: 1.1.10

**Additional Information**:
Chrome 59.0.3071.115

Splaktar:
Thank you for reporting this. I can reproduce this on the latest docs site (1.1.10).

marosoft:
Actually what should be a default behaviour here? Should you be able to navigate through the chips if those are in readonly mode or not?

There are three lines of code in the chipsDirective.js file trying to block the navigation. However it is not blocked in the chipsController.js (left & right arrow key).

Removing the checks `!$mdChipsCtrl.readonly && ` in those lines fixes the problem with the navigation (allowing it for the readonly mode):
* https://github.com/angular/material/blob/master/src/components/chips/js/chipsDirective.js#L211
* https://github.com/angular/material/blob/master/src/components/chips/js/chipsDirective.js#L212
* https://github.com/angular/material/blob/master/src/components/chips/js/chipsDirective.js#L381

If this is what is expected, I can create a pull request for it.

Splaktar:
For a11y, I thought that we might need to allow focusing and selecting of readonly chips as that might be the only way that a screen reader would be able to see what chips exist.

However, after doing some testing in Voiceover tonight, that isn't the case. The screen reader has a way to go through each chip using screen reader specific shortcuts (not arrows or tab).

[...]

PR #11422, opened 2018-08-20
fix(chips): improve screen reader support

### LINKED ISSUE #2618: chips: accessibility improvements
There are a few issues in mdChips that make them hard to use with a screen reader, including:
- The input needs a label enforced - right now it's empty if no placeholder
- When the input is focused, it should announce the label, not "Press delete to remove this chip"
- Chips need to be uniquely labeled along with the delete hint text. Ex: when a chip is focused, it should say: "Apple. Press delete to remove this chip".
- Need to verify chips can be removed from touch devices with screen readers running.

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[x] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

- static chips are not accessible
- contact chips have very poor accessibility
- chips have numerous accessibility issues
  - the index # of chips read to the user is always wrong and confusing
  - there is no feedback when a chip is removed
  - there is no way to provide different container instructions for an empty set of chips
  - when a chip is selected, there are no instructions for how to delete it
  - the delete buttons on chips are in the button accessibility tree, but they are confusing since they only say that pressing delete will remove the chip, but they don't indicate which chip they are associated with

Issue Number: 
Fixes #2618

## What is the new behavior?
- make static chips accessible
- add support for specifying input-aria-labelledby
- add support for specifying input-aria-describedby
- add a container-empty-hint API for use when there are no chips
- add support for screen readers announcing chip removal and addition
- add a md-added-message API to customize the chip creation announcement
- add a md-removed-message API to customize the chip removal announcement
- move chips-container-hint and delete-button-label to aria-labels
- fixed use of aria-posinset which is 1 indexed, not 0 indexed
- when a chip is removable, the label will include the delete-hint
- add the same screen reader support to md-contact-chips as well
- change autocomplete's aria-described by to input-aria-describedby
  - fixes a conflict with duplicating messages in screen readers

```
[ ] Yes
[x] No
```

## Other information
You can test these changes out here:
https://angularjs-material.firebaseapp.com/demo/autocomplete
https://angularjs-material.firebaseapp.com/demo/chips

This has been tested against: 
- [x] VoiceOver for macOS.
- [x] ChromeVox on ChromeOS 69
- [ ] Talkback for Android
- [ ] JAWS
- [x] NVDA

I plan to do some more testing of this via Talkback for Android. 
I could use some help with testing on JAWS and NVDA.


Splaktar:
@marcysutton I know you are busy with a lot of other things atm. But if some spare time comes up, I'd really appreciate a review of these changes or some feedback on the chips and autocomplete demo pages if you only find time to look at this after it's been merged.

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -491,7 +491,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
           ctrl.matches = [];
 
           setLoading(false);
-          reportMessages(false, ReportType.Count);
+          reportMessages(true, ReportType.Count);
```
Splaktar: This stops the screen reader from interrupting the message about a chip being added if you have a long enough `md-chip-append-delay`. With the default delay, the chip added message is still interrupted on VoiceOver, but that's an issue for another time.

src/components/autocomplete/js/autocompleteDirective.js
```diff
@@ -430,7 +430,6 @@ function MdAutocomplete ($$mdSvgRegistry) {
                 aria-autocomplete="list"\
                 role="combobox"\
                 aria-haspopup="true"\
-                aria-activedescendant=""\
```
Splaktar: The lighthouse a11y audit was complaining about these invalid aria values. They don't seem to be required by the spec and also don't seem to be used anywhere.

Splaktar:
Added links to the OP to allow testing of these changes with a screen reader.

Splaktar:
Static chips don't work as well on Chromevox as on VoiceOver. I opened https://bugs.chromium.org/p/chromium/issues/detail?id=876571 to report this issue.

Splaktar:
Did some testing on NVDA today. I'm still getting the hang of it, but here were my observations:

1. The `md-chip-append-delay` feature that was added for a11y in 1.1.2 seems to be the wrong approach and causes a lot of problems in NVDA, [VoiceOver](https://github.com/angular/material/pull/11422#discussion_r211432631), and Chromevox. I'd really like to remove it, but perhaps we can just deprecate it for now and change the default to `0`. It reads the chip + the "Press delete to remove this chip instruction". I'll investigate these top two issues together.
1. To address the previous issue, we should use `$mdLiveAnnouncer` to announce when a chip is created (this PR adds this announcement when a chip is removed). This would only announce the chip contents (not instructions) and then a message like "created". This needs to be customizable like the `md-remove-message`.
1. NVDA announces "blank <chip contents>" often, I.e. "blank Apple Press delete to remove this chip". I'm not sure why this is happening. Also at the end of reading the announcements for a chips input, it announces "blank".
1. Sometimes when navigating readonly chips via the keyboard, the first chip will get selected and announced properly, but then the 2nd chip does not get selected and gets announced. Then things just get worse and more confused from there. Something seems to be broken with NVDA's ability to go through the `listbox`. This is an existing issue being tracked in https://github.com/angular/material/issues/10850.
1. Navigation seems to work better for static chips, but it seems to require the up/down arrow keys instead of left/right.

Splaktar:
Regarding NVDA announcing "blank" when it isn't helpful, this is apparently a long known issue with NVDA. Here are some related open issues:
https://github.com/nvaccess/nvda/issues/2522
- This issue wants `<br>` to be handled more like `<p>` where NVDA doesn't announce "blank".
https://github.com/nvaccess/nvda/issues/8747
- This is a request to add an option to disable announcing "blank" for blank lines.
https://github.com/nvaccess/nvda/issues/4904
- This is a request to disable announcing "blank" in browse mode.
https://github.com/nvaccess/nvda/issues/1674
- This one is specifically an issue where any kind of whitespace is read as "blank". 

I will try to minimize the impacts of these issues by reducing/eliminating whitespace within chip content and chips input elements.

Splaktar:
I was testing this on macOS with Chrome 69 (stable) today and I almost had a heartattack that none of the aria-live regions worked!

Then I found https://bugs.chromium.org/p/chromium/issues/detail?id=875843 today. Basically `aria-live` regions were broken on VoiceOver in Chrome 69. Fixes coming in Chrome Canary (today/tomorrow), but that means about 2 months where macOS Chrome users will not have working Live Regions!

Note that it does work properly for me in Chrome 71 (Canary) today.

Splaktar:
I made one tweak to make NVDA announce "blank" in one less instance.

I also removed the deprecation and `aria-hidden` on the chip delete buttons. This is needed for Talkback for Android and VoiceOver on iOS (i.e. touch screen readers).

I also added the `md-added-message` and announcement when a chip is added as mentioned in `2.` above.

Splaktar:
Please run this through presubmit and let me know if there are any issues, but please don't merge it yet if it passes.

Splaktar:
Most Lighthouse a11y issues have been resolved other than:
- [ ] Issues with aria attributes not matching their roles

Splaktar:
Fixed issue where `aria-setsize` needs to be declared on each `option` or `listitem` rather than on the `listbox` or `list` as defined by https://www.w3.org/TR/wai-aria/#aria-setsize.

Splaktar:
https://bugs.chromium.org/p/chromium/issues/detail?id=876571 mentioned above, was fixed and merged. However, I am unsure which version of Chrome it is included in.

src/components/autocomplete/js/autocompleteDirective.js
Splaktar: This comes back in https://github.com/angular/material/pull/11441 but when there is no selected item, the attribute is removed as required by the spec.

Splaktar:
Rebased ✅

PR #11424, opened 2018-08-24
fix(toast): improve a11y support for $mdToast.simple(). improve docs

### LINKED ISSUE #349: toast: needs to support non-visual & keyboard users

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Actions within toasts are not announced at all for screen readers.
It's almost impossible to activate an action in a toast from a screen reader.
There are no instructions for screen reader users on how to dismiss a toast.

Issue Number: 
Fixes #349

## What is the new behavior?
move the role="alert" up a level
 - makes action button visible to screen readers
add support for defining an actionKey to assign a hot key to an action
- this enables Control-actionKey to activate the action
add support for defining a dismissHint for screen readers
add support for defining an actionHint for screen readers
align custom toast demo with Material Design guidelines
enhance custom toast demo to demonstrate an accessible custom toast

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/toast/toast.js
+++ b/src/components/toast/toast.js
@@ -156,9 +156,30 @@ function MdToastDirective($mdToast) {
   *        <td>`.action(string)`</td>
   *        <td>
   *          Adds an action button. <br/>
-  *          If clicked, the promise (returned from `show()`)
-  *          will resolve with the value `'ok'`; otherwise, it is resolved with `true` after a `hideDelay`
-  *          timeout
+  *          If clicked, the promise (returned from `show()`) will resolve with the value `'ok'`;
+  *          otherwise, it is resolved with `true` after a `hideDelay` timeout.
+  *        </td>
+  *      </tr>
+  *      <tr>
+  *        <td>`.actionKey(string)`</td>
+  *        <td>
+  *          Adds a hotkey for the action button. <br/>
+  *          If the `actionKey` and Control are pressed, the toast's action will be triggered.<br>
+  *          Defaults to the first character of the action if not defined.
+  *        </td>
+  *      </tr>
+  *      <tr>
+  *        <td>`.actionHint(string)`</td>
+  *        <td>
+  *          Text that a screen reader will announce to let the user know how to activate the
+  *          action. <br>Defaults to: "Press Control-"`actionKey`" to " followed by the action.
+  *        </td>
+  *      </tr>
+  *      <tr>
+  *        <td>`.dismissHint(string)`</td>
+  *        <td>
+  *          Text that a screen reader will announce to let the user know how to dismiss the toast.
+  *          <br>Defaults to: "Press Escape to dismiss."
   *        </td>
   *      </tr>
   *      <tr>
@@ -232,9 +253,11 @@ function MdToastDirective($mdToast) {
   *   - `scope` - `{object=}`: the scope to link the template / controller to. If none is specified, it will create a new child scope.
   *     This scope will be destroyed when the toast is removed unless `preserveScope` is set to true.
   *   - `preserveScope` - `{boolean=}`: whether to preserve the scope when the element is removed. Default is false
-  *   - `hideDelay` - `{number=}`: How many milliseconds the toast should stay
-  *     active before automatically closing.  Set to 0 or false to have the toast stay open until
-  *     closed manually. Default: 3000.
+  *   - `hideDelay` - `{number=}`: The number of milliseconds the toast should stay active before
+  *     automatically closing. Set to `0` or `false` to have the toast stay open until closed
+  *     manually via an action in the toast, a hotkey, or a swipe gesture. For accessibility, toasts
+  *     should not automatically close when they contain an action.<br>
+  *     Defaults to: `3000`.
   *   - `position` - `{string=}`: Sets the position of the toast. <br/>
   *     Available: any combination of `'bottom'`, `'left'`, `'top'`, `'right'`, `'end'` and `'start'`.
   *     The properties `'end'` and `'start'` are dynamic and can be used for RTL support.<br/>
@@ -247,7 +270,7 @@ function MdToastDirective($mdToast) {
   *     be used as names of values to inject into the controller. For example,
   *     `locals: {three: 3}` would inject `three` into the controller with the value
   *     of 3.
-  *   - `bindToController` - `bool`: bind the locals to the controller, instead of passing them in.
+  *   - `bindToController` - `{boolean=}`: bind the locals to the controller, instead of passing them in.
   *   - `resolve` - `{object=}`: Similar to locals, except it takes promises as values
   *     and the toast will not open until the promises resolve.
   *   - `controllerAs` - `{string=}`: An alias to assign the controller to on the scope.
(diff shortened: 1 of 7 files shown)


src/components/toast/toast.js
```diff
@@ -156,9 +156,30 @@ function MdToastDirective($mdToast) {
...
-  *          timeout
+  *          If clicked, the promise (returned from `show()`) will resolve with the value `'ok'`;
+  *          otherwise, it is resolved with `true` after a `hideDelay` timeout.
+  *        </td>
+  *      </tr>
+  *      <tr>
+  *        <td>`.actionKey(string)`</td>
+  *        <td>
+  *          Adds a hot key for the action button. <br/>
+  *          If the `actionKey` and Control are pressed, the toast's action will be triggered.<br>
```
jelbourn: Always having this be based on `control` seems a little too specific. The user might want to do `ctrl + shift` or even just a single key. Gmail, for example, uses just `z` for undo. For `MatSnackbar`, we just document that the user should add their own keyboard shortcut.

src/components/toast/toast.js
```diff
@@ -306,17 +328,22 @@ function MdToastProvider($$interimElementProvider) {
...
-      methods: ['textContent', 'content', 'action', 'highlightAction', 'highlightClass', 'theme', 'parent' ],
+      methods: ['textContent', 'content', 'action', 'actionKey', 'actionHint', 'highlightAction',
+                'highlightClass', 'theme', 'parent', 'dismissHint' ],
       options: /* @ngInject */ function($mdToast, $mdTheming) {
         return {
           template:
             '<md-toast md-theme="{{ toast.theme }}" ng-class="{\'md-capsule\': toast.capsule}">' +
-            '  <div class="md-toast-content">' +
-            '    <span class="md-toast-text" role="alert" aria-relevant="all" aria-atomic="true">' +
+            '  <div class="md-toast-content" role="alert" aria-relevant="all">' +
```
jelbourn: Using `role="alert"` isn't really the best pattern for snackbars. `alert` is a more specific type of dialog, which are inherently interuptive, while snackbars are meant to be non-interuptive. 

The recommend approach is to use an `aria-live` region (polite) for the text of the notification (which is what the MatSnackbar does)

src/components/toast/toast.js
Splaktar: I agree. However this only applies to the "simple" toasts. 

If developers need more flexible options, then the custom toast demo was updated to demonstrate how to setup custom keyboard shortcuts.

src/components/toast/toast.js
Splaktar: I looked into using `$mdLiveAnnouncer` which does enable the use of `role="status"` live regions with `aria-live="polite"` but the current implementation of it does not work well with interim elements like toasts.

I'm not 100% sure, but I believe for the current implementation to work properly, the `$mdLiveAnnouncer` needs to be injected in a component that is longer lived. This might work fine if other components are using `$mdLiveAnnouncer` and you add a toast, but it doesn't seem to work if you only add the toast w/o any other components to trigger instantiation earlier.

The [description of the `alert` role](https://www.w3.org/TR/wai-aria-1.1/#alert) does seem to align fairly well with the behavior of snackbars and toasts. 

The [`alertdialog` role](https://www.w3.org/TR/wai-aria-1.1/#alertdialog) does not, as it better fits a modal dialog.

jelbourn: Ah, you're right, I mixed up `alert` and `alertdialog`. In that case you should be able to swap `role="alert"` with `aria-live="polite"` and get the same treatment. Not going through the live announcer isn't ideal, but AFAIK it should work most of the time.

Splaktar: If I swap to [`role="status"`](https://www.w3.org/TR/wai-aria-1.1/#status) then it should default to `aria-live="polite"`. However, it seems like a snackbar with an interactive button / hotkey would better fit the `alert` role versus the `status` role. 

The other downside here is that `role="status"` is broken in Chromevox until Chrome M70.

It looks like `role="alert" aria-live="polite"` together may be the best fit. I'll do some more testing.

Splaktar: OK, adding both isn't a good idea as it causes double reading of the messages. 

However, `polite` in most cases (with our current defaults and demos) causes the toast to be read after the toast times out and disappears. This means that by the time the user heard the instructions for interacting, the toast would be gone.

To mitigate this, I've added instructions to the `hideDelay` docs to recommend setting it to `0` when the toast includes an action. I would like to do this automatically for simple toasts, but I will avoid it for now since it might cause some g3 tests to fail. I've also updated the demos to use `hideDelay(0)` when they have an action.

src/components/toast/toast.js
Splaktar: I'm also happy to extend this API to accept values for the `metaKey`, `ctrlKey`, `shiftKey`, and `altKey` should the request arise.

Do you think that we should just go ahead and support those now?

jelbourn: I don't feel super strongly either way.

src/components/toast/toast.js
jelbourn: Yeah, I don't believe `role="status"` is widely supported. 

The Material Design team also says that snackbars shouldn't auto-dismiss when they have an action, so that would indeed be the right behavior.

src/components/toast/toast.js
```diff
@@ -306,17 +330,22 @@ function MdToastProvider($$interimElementProvider) {
...
-            '  <div class="md-toast-content">' +
-            '    <span class="md-toast-text" role="alert" aria-relevant="all" aria-atomic="true">' +
+            '  <div class="md-toast-content" aria-live="polite" aria-relevant="all">' +
+            '    <span class="md-toast-text">' +
             '      {{ toast.content }}' +
             '    </span>' +
+            '    <span class="md-visually-hidden">{{ toast.dismissHint }}</span>' +
+            '    <span class="md-visually-hidden" ng-if="toast.action">' +
+            '      {{ toast.actionHint }}' +
+            '    </span>' +
```
Splaktar: The new `dismissHint` and `actionHint` `span` elements are returned by tests that do `.find('span').text()` like `expect(parent.find('span').text().trim()).toBe('Do something');`. So the tests will now get the all of the contents (even the contents of the two new `md-visually-hidden` `span`s). This broke two AngularJS Material tests, but it appears to be breaking some tests within Google as well.

Can anyone think of any way to work around this that doesn't mean changing all of the tests?

This is in an Aria Live Region, so I don't think that something like `aria-describedby` would be a valid workaround.

I wonder if wrapping the new elements in a `div` or changing them to `div`s would work. It seems like it may break screen readers but I will test it.

- [ ] Test wrapping `dismissHint` and `actionHint` `span` elements in `divs`
- [ ] Test changing  `dismissHint` and `actionHint` `span` elements to `divs`

Splaktar: Jeremy indicated that it should actually be easy to fix the 2 tests within Google that are broken by this.

PR #11428, opened 2018-08-27
feat(theming): add ability to specify hues as options to defineTheme

### LINKED ISSUE #10335: theming: add ability to specify hues as options to defineTheme
**Actual Behavior**:
 -  `What is the issue? *`

When using defineTheme to generate theme on the fly, we can only indicate the "name" of a palette. It is not possible to inject hues options, as we could do it before with : 
` $mdThemingProvider.theme(name).primaryPalette(palettes.primaryPalette.name, {
                'default': palettes.primaryPalette.main,
                'hue-1': palettes.primaryPalette.hue1,
                'hue-2': palettes.primaryPalette.hue2,
                'hue-3': palettes.primaryPalette.hue3
            })`
We are stuck now with the default hues, if we use defineTheme.
As a conclusion, today, we have to stick to the provider methods.

We should be able to pass a whole set of palette options containing name & hues, in order to dynamically set the hues of the relevant palette.

No code pen would show anything. The present signature of defineTheme only takes: 
`defineTheme(themeName, paletteName) `   (String, String)
A different signature could be: 
`defineTheme(themeName, paletteOptions)` (String, Objet)
paletteOptions could be an object like this:
`{
'name': 'paletteName',
'options': {
                'default': '500',
                'hue-1': '100',
                'hue-2':'200',
                'hue-3': '800'
            }
}`
 -  `Details:`

**Angular Versions**: *
 -  `Angular Version:` 1.5.11
 -  `Angular Material Version:` 1.1.2 & 1.1.3

**Additional Information**:
 -  `Browser Type: *`    All
 -  `Browser Version: *` All
 -  `OS: *` All
 -  `Stack Traces:`

----

  Add more lines if needed.

xyng:
* extend test of `$mdTheming` to include one palette with and one without hue.
* add docs regarding new `paletteHues` options for `defineTheme` method of `$mdTheming`.

Fixes #10335

## PR Checklist

## PR Type

```
[ ] Bugfix
[x] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Currently, there is no possibility to apply hue-options to palettes when using `defineTheme`. This is described thoroughly in the following issue and Google Group.
https://groups.google.com/forum/#!topic/ngmaterial/d-D3MNFl6wg
Issue Number: #10335

## What is the new behavior?
`paletteHues`options have been added to the options object. These are completely optional.

```
[ ] Yes
[x] No
```

## Other information
I opted to add an option for each palette instead of changing each palette option to be either of type string or object, as that would be more likely to break or be confusing.
However I'm willing to implement it that way if you desire.

### DIFF
--- a/src/core/services/theming/theming.js
+++ b/src/core/services/theming/theming.js
@@ -574,6 +574,8 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
    * @description
    * Service that makes an element apply theming related <b>classes</b> to itself.
    *
+   * For more information on the hue objects, their default values, as well as valid hue values, please visit <a ng-href="Theming/03_configuring_a_theme#specifying-custom-hues-for-color-intentions">the custom hues section of Configuring a Theme</a>.
+   *
    * <hljs lang="js">
    * // Example component directive that we want to apply theming classes to.
    * app.directive('myFancyDirective', function($mdTheming) {
@@ -585,7 +587,27 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
    *
    *       $mdTheming.defineTheme('myTheme', {
    *         primary: 'blue',
+   *         primaryHues: {
+   *           default: '500',
+   *           hue-1: '300',
+   *           hue-2: '900',
+   *           hue-3: 'A100'
+   *         },
    *         accent: 'pink',
+   *         accentHues: {
+   *           default: '600',
+   *           hue-1: '300',
+   *           hue-2: '200',
+   *           hue-3: 'A500'
+   *         },
+   *         warn: 'red',
+   *         // It's not necessary to specify all hues in the object.
+   *         warnHues: {
+   *           default: '200',
+   *           hue-3: 'A100'
+   *         },
+   *         // It's not necessary to specify custom hues at all.
+   *         background: 'grey',
    *         dark: true
    *       });
    *       // Your directive's custom code here.
@@ -670,9 +692,13 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
    * @param {object} options Theme definition options
    * Options are:<br/>
    * - `primary` - `{string}`: The name of the primary palette to use in the theme.<br/>
+   * - `primaryHues` - `{object=}`: Override hues for primary palette.<br/>
    * - `accent` - `{string}`: The name of the accent palette to use in the theme.<br/>
+   * - `accentHues` - `{object=}`: Override hues for accent palette.<br/>
    * - `warn` - `{string}`: The name of the warn palette to use in the theme.<br/>
+   * - `warnHues` - `{object=}`: Override hues for warn palette.<br/>
    * - `background` - `{string}`: The name of the background palette to use in the theme.<br/>
+   * - `backgroundHues` - `{object=}`: Override hues for background palette.<br/>
    * - `dark` - `{boolean}`: Indicates if it's a dark theme.<br/>
    * @returns {Promise<string>} A resolved promise with the new theme name.
    */
@@ -711,16 +737,16 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
       var theme = registerTheme(name);
 
       if (options.primary) {
-        theme.primaryPalette(options.primary);
+        theme.primaryPalette(options.primary, options.primaryHues);
       }
       if (options.accent) {
-        theme.accentPalette(options.accent);
+        theme.accentPalette(options.accent, options.accentHues);
       }
       if (options.warn) {
-        theme.warnPalette(options.warn);
+        theme.warnPalette(options.warn, options.warnHues);
       }
       if (options.background) {
-        theme.backgroundPalette(options.background);
+        theme.backgroundPalette(options.background, options.backgroundHues);
       }
       if (options.dark){
         theme.dark();
--- a/src/core/services/theming/theming.spec.js
+++ b/src/core/services/theming/theming.spec.js
@@ -760,6 +760,35 @@ describe('$mdTheming service', function() {
     expect($mdTheming.THEMES.hasOwnProperty('test')).toBeTruthy();
   }));
 
+  it('supports setting palette options when registering theme on the fly', inject(function ($mdTheming) {
+    expect($mdTheming.THEMES.hasOwnProperty('testHues')).toBeFalsy();
+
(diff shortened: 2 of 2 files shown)


xyng:
We signed the cla. Google Group is: https://groups.google.com/forum/#!forum/interweber

src/core/services/theming/theming.js
```diff
@@ -670,9 +670,13 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
    * @param {object} options Theme definition options
    * Options are:<br/>
    * - `primary` - `{string}`: The name of the primary palette to use in the theme.<br/>
+   * - `primaryHues` - `{object} (Optional)`: Overwrite hues for primary palette.<br/>
```
Splaktar: Please use `=` at the end of the type to indicate that it is optional, i.e. `{object=}`.

Let's use "Override" instead of "Overwrite".

src/core/services/theming/theming.js
```diff
@@ -670,9 +670,13 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
...
    * - `accent` - `{string}`: The name of the accent palette to use in the theme.<br/>
+   * - `accentHues` - `{object} (Optional)`: Overwrite hues for accent palette.<br/>
    * - `warn` - `{string}`: The name of the warn palette to use in the theme.<br/>
+   * - `warnHues` - `{object}`: Overwrite hues for warn palette.<br/>
```
Splaktar: Can you please add an example to the main [`$mdTheming` docs](https://github.com/angular/material/blob/v1.1.10/src/core/services/theming/theming.js#L570-L597) where these new parameters are used?

Without that, it's not clear here what this `object` would consist of.

src/core/services/theming/theming.spec.js
```diff
@@ -752,8 +752,10 @@ describe('$mdTheming service', function() {
   it('supports registering theme on the fly', inject(function ($mdTheming) {
     expect($mdTheming.THEMES.hasOwnProperty('test')).toBeFalsy();
 
+    // optional <color>Hue option for each color, as in normal theme definition.
     $mdTheming.defineTheme('test', {
       primary: 'red',
+      primaryHue: 300,
```
Splaktar: This should be `primaryHues` right? The value should also be an `object`?

src/core/services/theming/theming.spec.js
```diff
@@ -752,8 +752,10 @@ describe('$mdTheming service', function() {
   it('supports registering theme on the fly', inject(function ($mdTheming) {
     expect($mdTheming.THEMES.hasOwnProperty('test')).toBeFalsy();
 
+    // optional <color>Hue option for each color, as in normal theme definition.
```
Splaktar: After you make the other updates, please clarify this comment or remove it.

src/core/services/theming/theming.spec.js
Splaktar: We should probably have a test that verifies that the correct hue was defined/used.

src/core/services/theming/theming.js
```diff
@@ -585,6 +585,9 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
    *
    *       $mdTheming.defineTheme('myTheme', {
    *         primary: 'blue',
+   *         primaryHues: {
```
xyng: I'm not sure if this doc makes things clear. The `setPalette` function seems to have no docs itself.
Any suggestions?

src/core/services/theming/theming.spec.js
xyng: I have little experience with writing tests. I'll look into it and see what I can do.

src/core/services/theming/theming.spec.js
```diff
@@ -754,10 +754,14 @@ describe('$mdTheming service', function() {
...
     $mdTheming.defineTheme('test', {
       primary: 'red',
+      primaryHues: {
+        default: '300'
+      },
       warn: 'yellow'
     });
 
     expect($mdTheming.THEMES.hasOwnProperty('test')).toBeTruthy();
+    expect($mdTheming.THEMES.test.colors.primary.hues.default).toBe('300');
```
xyng: I suppose this should be enough to test the hue overrides. Suggestions?

Splaktar: The assertion looks like it will work.

However, it would be helpful if, instead of changing this test, you could add a new test that was similar but set and verified each of the different `xHues` fields.

Splaktar:
Hopefully those comments were helpful. Please don't hesitate to ask for clarification or any other information that you may need.

Thank you for your contributions!

src/core/services/theming/theming.spec.js
xyng: I did a new test case, following your suggestion. I agree that this is a better way of testing it.

src/core/services/theming/theming.js
xyng: Thank you very much for this thorough information. I really appreciate your feedback.
I tried to implement it as good as possible - though I'm not the best at writing docs.

Sadly, I'm running out of time for this. So if you should have other suggestions regarding documentation I'd like to ask you to make these changes directly, if that is possible.

src/core/services/theming/theming.js
```diff
@@ -574,6 +574,8 @@ function ThemingProvider($mdColorPalette, $$mdMetaProvider) {
    * @description
    * Service that makes an element apply theming related <b>classes</b> to itself.
    *
+   * For more Information on the `xHues` objects, their default values, as well as valid hue values, please visit <a ng-href="Theming/03_configuring_a_theme#specifying-custom-hues-for-color-intentions">the custom hues section of Configuring a Theme</a>.
```
Splaktar: OK, out of respect for your limited time, I will hold off on some feedback, but I do want to fix one thing in this line.

Can you please change `Information on the `xHues` objects` to `information on the hue objects`?

src/core/services/theming/theming.js
Splaktar: Unfortunately, it's non-trivial for me to make changes to your PR, but I do think that this is about ready to wrap up. Thank you for your cooperation.

src/core/services/theming/theming.js
xyng: Done :-)

PR #11432, opened 2018-08-30
fix(tabs): min-width not respected on IE11. fix right/left padding

### LINKED ISSUE #10406: md-tabs: text turns to dots on small screen on IE11
I need tabs to be streched always and on small screans to, but words are turning in to dots or half of a word turns to dots. I thought maybe words are too big, but then i tried to resize the window and then it turns to normal words.

My code:
```
<md-tabs class="main-tabs" md-selected="vm.selectedIndex"
             md-center-tabs md-border-bottom md-autoselect md-no-ink
             md-no-pagination="false"
             md-stretch-tabs="auto"
             style="width: auto !important; margin-top: 10px;">
        <md-tab ng-click="openLeftMenu()" id="tab1">
            <md-tab-label>{{translates.top_menu_meniu}}</md-tab-label>
        </md-tab>
        <md-tab ui-sref="map" id="tab2">
            <md-tab-label>{{translates.top_menu_map}}</md-tab-label>
        </md-tab>
        <md-tab ui-sref="list" id="tab3">
            <md-tab-label>{{translates.top_menu_list}}</md-tab-label>
        </md-tab>
    </md-tabs>
```

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

- Tab items have right and left padding of `24px` which is `12px` more on each side than specified by [the MD spec](https://material.io/archive/guidelines/components/tabs.html#tabs-specs).
- Tabs on IE11 are truncated by pagination due to a calculated width of `0px` and the only width being the `48px` of padding.
- There are 2 functions in the code that are not used anymore.
- Many of the tab styles use `translate3d` which causes performance issues and bugs on IE11. We don't actually translate anything in the `z` dimension.

Issue Number: 
Fixes #10406

## What is the new behavior?
- tab-items should have 12px right/left padding instead of 24px
    - this is to align with [the spec](https://material.io/archive/guidelines/components/tabs.html#tabs-specs) and more of the label visible on mobile
- remove unused calcPagingWidth()
- convert all translate3d to translate for better IE11 support

```
[ ] Yes
[x] No
```

## Other information
The breaking changes to make `min-width` align with the MD spec have been pulled out of this PR and put into PR https://github.com/angular/material/pull/11438 for 1.2.0.

Please note that this is fixing a **regression** from a breaking change that was in 1.1.2.

Before (IE on bottom, Chrome on top):
chrome-and-ie-before-fix

After in IE
tabs-fixed-ie11-without-min-width-fix
tabs-pagination-ie11-fixed-without-min-width-fix

After in Chrome
tabs-pagination-chrome-without-min-width-fix

And here is IE11 rendering with a small width (sub `600px`)
tabs-pagination-ie11-xs-fixed
tabs-fixed-ie11-xs

### DIFF
--- a/src/components/tabs/tabs.scss
+++ b/src/components/tabs/tabs.scss
@@ -76,7 +76,7 @@ md-tabs-wrapper {
   display: block;
   position: relative;
   // transform is needed for iOS Safari to prevent content from disappearing on scroll
-  transform: translate3d(0, 0, 0);
+  transform: translate(0, 0);
   md-prev-button, md-next-button {
     height: 100%;
     width: $tabs-paginator-width;
@@ -103,7 +103,7 @@ md-tabs-wrapper {
       position: absolute;
       top: 50%;
       left: 50%;
-      transform: translate3d(-50%, -50%, 0);
+      transform: translate(-50%, -50%);
     }
 
     // For RTL tabs, rotate the buttons
@@ -121,7 +121,7 @@ md-tabs-wrapper {
 
     // In regular mode, we need to flip the chevron icon to point the other way
     md-icon {
-      transform: translate3d(-50%, -50%, 0) rotate(180deg);
+      transform: translate(-50%, -50%) rotate(180deg);
     }
   }
   &.md-stretch-tabs {
@@ -166,8 +166,9 @@ md-pagination-wrapper {
   display: flex;
   transition: transform $swift-ease-in-out-duration $swift-ease-in-out-timing-function;
   position: absolute;
+  width: 999999px;
   @include rtl-prop(left, right, 0, auto);
-  transform: translate3d(0, 0, 0);
+  transform: translate(0, 0);
   &.md-center-tabs {
     position: relative;
     justify-content: center;
@@ -194,7 +195,7 @@ md-tab-content {
   transition: transform $swift-ease-in-out-duration $swift-ease-in-out-timing-function;
   overflow: auto;
   // transform is needed for iOS Safari to prevent content from disappearing on scroll
-  transform: translate3d(0, 0, 0);
+  transform: translate(0, 0);
   &.md-no-scroll {
     bottom: auto;
     overflow: hidden;
--- a/src/components/tabs/js/tabsController.js
+++ b/src/components/tabs/js/tabsController.js
@@ -236,7 +236,7 @@ function MdTabsController ($scope, $element, $window, $mdConstant, $mdTabInkRipp
     // Fix double-negative which can happen with RTL support
     newValue = newValue.replace('--', '');
 
-    angular.element(elements.paging).css($mdConstant.CSS.TRANSFORM, 'translate3d(' + newValue + ', 0, 0)');
+    angular.element(elements.paging).css($mdConstant.CSS.TRANSFORM, 'translate(' + newValue + ', 0)');
     $scope.$broadcast('$mdTabsPaginationChanged');
   }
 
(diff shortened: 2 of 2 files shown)


Splaktar:
@jelbourn please review and let me know if you think that increasing the `min-width` to `160px` on `gt-xs` devices will cause too many problems in Google. 

This change aligns with the spec and with Angular Material (other than it using `24px` padding on right/left of tab items). Note that it looks like the 2018 MD spec actually suggests `16px` padding, so that may be the right value for Angular Material.

Splaktar:
The `min-width` change on `gt-xs` devices causes 100+ tests to fail in g3. So I'll need to pull that part of this PR into a separate PR for 1.2.0.

Splaktar:
Removed fix for `min-width` to align with the MD spec. Put those changes into PR https://github.com/angular/material/pull/11438 for 1.2.0.

Updated OP screenshots to reflect these changes.

mmalerba:
@Splaktar did you forget to push the commit removing the `min-width`?

Splaktar:
@mmalerba no, that was pulled out to https://github.com/angular/material/pull/11438/commits/b7e14149304ded4ffb216ab7e2d773e2bddf9630 which is part of PR https://github.com/angular/material/pull/11438.

Is this still failing presubmit? If so, what are the issues?

mmalerba:
There's still a lot of screenshot diffs, I suspect this is the culprit:

```css
md-tab-item {
  min-width: 72px;
}
```

Splaktar:
@mmalerba Unfortunately that is the key line of the whole fix here. Without that, having many tabs and tab pagination is broken in IE11.

Just removing that one line causes:
tabs-pagination-ie11-broken

Splaktar:
I looked into using some more specific selectors, but those don't work as it's the `min-width` setting here that triggers the successful pagination calculations. Without it, the tabs on IE11 get a calculated width of `1.41px` instead of `72px`.

Are there apps that are using tabs with widths smaller than `72px`? Is there some `min-width` value that we can use to avoid breaking (even if it won't comply with the Material Design spec)?

src/components/tabs/tabs.scss
```diff
@@ -259,7 +262,7 @@ md-tab {
   font-size: 14px;
   text-align: center;
   line-height: $tabs-header-height - 24;
-  padding: 12px 24px;
+  padding: 12px;
```
Splaktar: @mmalerba this is one other change that is likely to cause screenshot diffs. The result is that there is more room in the tab to show the text and less chance of it being truncated. This also better aligns with the spec.

Are you seeing screenshot diffs that are actually breaking changes or are they just slightly shifted text within the tabs?

mmalerba: The changes looked mostly ok, but due to the number of screen diffs I'm not really sure we can make the change in google3

Splaktar: I tried removing this padding change, but it doesn't work without it. Those two CSS changes are the key to fixing this P2 Regression that was introduced in 1.1.2.

Here's what it looks like w/o the padding change:
tabs-pagination-ie11-broken2

Splaktar:
OK, it looks like we need to go back to the `width: 999999px;` hack that was removed in https://github.com/angular/material/commit/5b799e2.

After that change is brought back and the other `padding` and `min-width` changes are removed, this is the result:
Chrome:
tabs-fixed-chrome-9999

IE11 desktop
tabs-fixed-ie11-9999
tabs-dynamic-fixed-ie11-9999

IE11 on XS device
tabs-dynamic-fixed-ie11-9999-xs
tabs-fixed-ie11-9999-xs

Splaktar:
I've pushed that latest code and this should be ready for another round of presubmits.

Splaktar:
This latest iteration "causes a slight layout change, which seems pretty minimal, but is screenshot test breaking." It looks like it will need approval from about 5 apps to get in.

After about 5 iterations now, I think this is the least disruptive we can get on this regression fix. Considering it breaks IE11 completely, I think that we need to get this in.

meankiddeli:
Try setting the CSS property for it. `.md-tabs{ overflow : visible; }`

PR #11448, opened 2018-09-17
fix(select): perform full cleanup of the select drop-down after the c…

### LINKED ISSUE #11447: select: wrong drop-down position after the resize

Bug

#### [The official demo](https://material.angularjs.org/HEAD/demo/select#validations) demonstrates the issue:

After the opening and closing a drop-down in `Validations` section, the drop-down will appear in a wrong place after a resize or orientation change.

1. Go to [validations](https://material.angularjs.org/HEAD/demo/select#validations) section.

2. Open and close `Favorite Color` drop-down:

3. Reduce the size of the window, so the select also changes its size.

4. Open `Favorite Color` drop-down one more time:

The select drop-down works incorrectly on the stretchable forms or on the orientation change.

- AngularJS: all
- AngularJS Material: <=1.1.10 (all at the moment)
- OS: all
- Browsers: all
 
It needs proper cleanup here: https://github.com/angular/material/blob/master/src/components/select/select.js#L1253

dokmic:
fix(select): perform full cleanup of the select drop-down after the close

it affects on the drop-down size if resize is happening after using the select

Fixes #11447

## PR Checklist
- [ ] The commit message follows [our guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format)

## PR Type
```
[ ] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Issue Number: #11447 

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -1250,7 +1250,15 @@ function SelectProvider($$interimElementProvider) {
         element
           .removeClass('md-active')
           .attr('aria-hidden', 'true')
-          .css('display', 'none');
+          .css({
+            'display': 'none',
+            'top': '',
+            'right': '',
+            'bottom': '',
+            'left': '',
+            'font-size': '',
+            'min-width': ''
+          });
         element.parent().find('md-select-value').removeAttr('aria-hidden');
 
         announceClosed(opts);


src/components/select/select.js
```diff
@@ -1250,7 +1250,15 @@ function SelectProvider($$interimElementProvider) {
...
           .attr('aria-hidden', 'true')
-          .css('display', 'none');
+          .css({
+            'display': 'none',
+            'top': '',
+            'right': '',
+            'bottom': '',
+            'left': '',
+            'font-size': '',
+            'min-width': ''
```
Splaktar: Thankfully it looks like it results in these styles being cleared rather than having them set to empty strings.

Splaktar:
Thank you very much for your first contribution to AngularJS Material! It looks great to me. 

We'll send this to run some presubmit tests in Google and then merge it if successful.

PR #11453, opened 2018-09-21
fix(util,select): extra scrollbar when select dropdown is open

### LINKED ISSUE #11227: select: causes backdrop and extra scrollbar on IE

Bug

The expected behavior is the select box to open without overlay everything 

When the select box opens, a gray background appear and the box overlay everything and creates a scroll. I notice it creates a position fixed in the body and a width of 100%.

## CodePen and steps to reproduce the issue:

I don't have a codepen example , but your website if open on IE11 reproduces the same issue when we try to open the select box, i will leave a screenshot

untitled

https://material.angularjs.org/latest/demo/select , when it opens on IE11 it got the wrong behavior 

IE11, angular-material": "^1.1.6",

This is a screenshot of my page when i open the selectbox to choose an option, the background is filled whit a color and the select box is all deformed.

_ie11

Thank you !

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

The `-ms-overflow-y: scroll` style is applied to the `html` element when a `md-select` is open. This causes an extra scrollbar.

Issue Number: 
Fixes #11227

## What is the new behavior?
Don't apply `-ms-overflow-y: scroll` to the `html` element. No extra scrollbar is created.

```
[ ] Yes
[x] No
```

## Other information
Tested on Windows against IE11, Firefox, and Chrome.

Before:
select-ie11-extra-scrollbar

After:
select-ie11-extra-scrollbar-fixed

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -302,13 +302,6 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
           body.style.overflow = 'hidden';
         }
 
-        // This should be applied after the manipulation to the body, because
-        // adding a scrollbar can potentially resize it, causing the measurement
-        // to change.
-        if (hasVerticalScrollbar) {
-          documentElement.style.overflowY = 'scroll';
-        }
-
         return function restoreScroll() {
           // Reset the inline style CSS to the previous.
           body.style.cssText = prevBodyStyle;


codymikol:
This is actually something that's been really annoying in our app, thanks for fixing! :chicken:

Splaktar:
This is causing 17 failures. Screenshots are showing an application-level scrollbar showing up where there wasn't one before.

I need additional information on how to reproduce this so that I can investigate. We may need to add another check instead of removing the `if (hasVerticalScrollbar) {` block. But I need more information in order to debug it.

mcalmus:
This patch is making the screen jump for me now when I open a select because the page scrollbar goes away. Version 1.1.10 works fine.

PR #11441, opened 2018-09-11
fix(autocomplete): Chromevox indicates selected option on focus

Splaktar:
## PR Checklist
Please check that your PR fulfills the following requirements:
- [x] The commit message follows [our guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format)
- [ ] Tests for the changes have been added or this is not a bug fix / enhancement
- [x] Docs have been added, updated, or were not required

## PR Type
What kind of change does this PR introduce?

```
[x] Bugfix
[ ] Enhancement
[ ] Documentation content changes
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?
When selecting a value from an auto-complete, Chromevox reads it out as "{md-autocomplete aria-label} {Value} collapsed combo box". It's unclear to the user which part of the text read is the combo-box-name, and which is the value.

Issue Number: 
Fixes #10838

## What is the new behavior?
With the updates Chromevox will announce the following:
"{md-autocomplete aria-label} {Value} selected Collapsed Combo box"

If required:
"{md-autocomplete aria-label} {Value} selected Collapsed Required Combo box"

- implement WAI-ARIA 1.0 recommendation for autocomplete
- avoid changing listbox selection or announcing changes when using arrow keys when a selection has been made and the listbox is closed
- set aria-setsize and aria-posinset for dropdown options

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

## Other information
Tested on Chromevox for ChromeOS 69 and VoiceOver for macOS High Sierra and Mojave.

PR #11486, opened 2018-10-14
fix(nav-bar): improve screen reader support

### LINKED ISSUE #11485: nav-bar: Screen readers can't properly interact with a button inside an option inside a listbox.

Bug

Sorry, due to the inaccessibility of CodePen's editors, I cannot actually provide a codePen example for this. What I've done is change the ARIA roles to my proposed roles via the Firefox dev tools to prove that these work better with a screen reader.

1. Open any example that uses the md-nav-bar and md-nav-item elements.
2. Try to interact with them using the NVDA screen reader and current Firefox on Windows.

NVDA should see the buttons and be able to select each.

NVDA sees the listbox, but nothing in it. Only if I press Enter while the virtual (browse mode) cursor is on the list box, do I get focus to the buttons and can select each. The reason is that a button inside an option inside a listbox is totally non-standard and not expected by any current design pattern.

Better accessibility for blind screen reader users.

- AngularJS: Version used by Material.
- AngularJS Material: Current Master.
- OS: Windows 10
- Browsers: Firefox, Chrome.
 
The design pattern from the WAI-ARIA Authoring Practices no longer exists and thus should be re-thought. The closest pattern is:

- Change the role on the md-nav-bar from "listbox" to "tablist".
- Change the role of the li inside the md-nav-item template from "option" to "presentation".
- Add the role "tab" to the button in the button template for the li element.
- Set aria-selected on the button directly instead of the li element, since that is now presentational.

That way, the widget becomes a tab list with true tabs as children in the accessibility layer, which is a totally valid pattern and supported by screen readers. The one piece that is missing, since it is not part of this compound widget, is the correllation to a tabpanel object.

The alternative would be to use toolbar and buttons that are turned into toggle buttons, or alternatively, radio buttons of which only one can be selected at a time, but since visually these often look like tabs anyway, using the tablist and tab roles is the closest match.

MarcoZehe:
… reader users

* make the nav-bar component a tablist.
* make the nav-item's li element presentational and the button a tab.
* apply the aria-label, if provided, to the button directly, not the li element.

Fixes #11485

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Issue Number: #11485 

Currently, the nav-bar element is exposed as a listbox accessible to screen reader users. Its nav-item child consists of a li element which currently has role "option" (as appropriate for a listbox child), and the button inside that li is exposed as is.

Screen readers do have a hard time interacting with this construct, since they do not expect an option to contain a button.

## What is the new behavior?

The new behavior is that the nav-bar element is exposed as a tablist, and the button inside the li of the nav-item as a tab accessible. In addition, the aria-label that was originally applied to the nav-item's li element is now applied to the buton, since the li in the tablist/tab pattern is to be marked presentational. aria-label must therefore go onto the element that has role "tab".

No changes besides semantic ARIA exposure has been changed, the keyboard behavior is the same as well as any other behavior this component exposes. Web developers needn't change anything, either.

```
[ ] Yes
[x] No
```

## Other information

The tests have been adjusted to look for the aria-label on the tab directly, not the parent.

### DIFF
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -31,10 +31,10 @@ angular.module('material.components.navBar', ['material.core'])
  * Alternatively, the user could simply watch the value of `md-selected-nav-item`
  * (`currentNavItem` in the below example) for changes.
  *
- * Accessibility functionality is implemented as a site navigator with a
- * listbox, according to the
- * <a href="https://www.w3.org/TR/2016/WD-wai-aria-practices-1.1-20160317/#Site_Navigator_Tabbed_Style">
- *   WAI-ARIA Authoring Practices 1.1 Working Draft from March 2016</a>.
+ * Accessibility functionality is implemented as a
+ * <a href="https://www.w3.org/TR/wai-aria-1.0/complete#tablist">
+ *   tablist</a> with
+ * <a href="https://www.w3.org/TR/wai-aria-1.0/complete#tab">tabs</a>.
  * We've kept the `role="navigation"` on the `<nav>`, for backwards compatibility, even though
  *  it is not required in the
  * <a href="https://www.w3.org/TR/wai-aria-practices/#aria_lh_navigation">
@@ -43,7 +43,7 @@ angular.module('material.components.navBar', ['material.core'])
  * @param {string=} md-selected-nav-item The name of the current tab; this must
  *     match the `name` attribute of `<md-nav-item>`.
  * @param {boolean=} md-no-ink-bar If set to true, the ink bar will be hidden.
- * @param {string=} nav-bar-aria-label An `aria-label` applied to the `md-nav-bar`'s listbox
+ * @param {string=} nav-bar-aria-label An `aria-label` applied to the `md-nav-bar`'s tablist
  * for accessibility.
  *
  * @usage
@@ -90,7 +90,10 @@ angular.module('material.components.navBar', ['material.core'])
  * Exactly one of the `md-nav-click`, `md-nav-href`, or `md-nav-sref` attributes are required
  * to be specified.
  *
- * @param {string=} aria-label Adds alternative text for accessibility.
+ * @param {string=} nav-item-aria-label Allows setting or overriding the label that is announced by
+ *     a screen reader for the nav item's button. If this is not set, the nav item's transcluded
+ *     content will be announced. Make sure to set this if the nav item's transcluded content does
+ *     not include descriptive text, for example only an icon.
  * @param {expression=} md-nav-click Expression which will be evaluated when the
  *     link is clicked to change the page. Renders as an `ng-click`.
  * @param {string=} md-nav-href url to transition to when this link is clicked.
@@ -127,7 +130,7 @@ function MdNavBar($mdAria, $mdTheming) {
     template:
       '<div class="md-nav-bar">' +
         '<nav role="navigation">' +
-          '<ul class="_md-nav-bar-list" ng-transclude role="listbox" ' +
+          '<ul class="_md-nav-bar-list" ng-transclude role="tablist" ' +
             'tabindex="0" ' +
             'ng-focus="ctrl.onFocus()" ' +
             'ng-keydown="ctrl.onKeydown($event)" ' +
@@ -444,6 +447,9 @@ function MdNavItem($mdAria, $$rAF, $mdUtil, $window) {
             'ng-blur="ctrl.setFocused(false)" ' +
             'ng-disabled="ctrl.disabled" ' +
             'tabindex="-1" ' +
+            'role="tab" ' +
+            'ng-attr-aria-label="{{ctrl.navItemAriaLabel ? ctrl.navItemAriaLabel : undefined}}" ' +
+            'aria-selected="{{ctrl.isSelected()}}" ' +
             navigationOptions +
             navigationAttribute + '>' +
             '<span ng-transclude class="_md-nav-button-text"></span>' +
@@ -452,8 +458,7 @@ function MdNavItem($mdAria, $$rAF, $mdUtil, $window) {
 
       return '' +
         '<li class="md-nav-item" ' +
-          'role="option" ' +
-          'aria-selected="{{ctrl.isSelected()}}">' +
+          'role="presentation">' +
           (buttonTemplate || '') +
         '</li>';
     },
@@ -463,6 +468,7 @@ function MdNavItem($mdAria, $$rAF, $mdUtil, $window) {
       'mdNavSref': '@?',
       'srefOpts': '=?',
       'name': '@',
+      'navItemAriaLabel': '@?',
     },
     link: function(scope, element, attrs, controllers) {
       var disconnect;
@@ -504,7 +510,9 @@ function MdNavItem($mdAria, $$rAF, $mdUtil, $window) {
           });
         }
 
-        $mdAria.expectWithText(element, 'aria-label');
+        if (!mdNavItem.navItemAriaLabel) {
+          $mdAria.expectWithText(navButton, 'aria-label');
+        }
       });
 
       scope.$on('destroy', function() {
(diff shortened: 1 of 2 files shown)


Splaktar:
The build failure looks like a fluke. I have restarted it.

Splaktar:
It looks like the test failures are not a fluke as they keep happening and I can reproduce them on my local machine. Are you able to get all of the tests to pass on your local machine with `gulp karma`?

Splaktar:
I've opened https://github.com/angular/material/issues/11487 to track the test failures on `master`.

src/components/navBar/navBar.js
```diff
@@ -457,6 +459,7 @@ function MdNavItem($mdAria, $$rAF, $mdUtil, $window) {
       'mdNavSref': '@?',
       'srefOpts': '=?',
       'name': '@',
+      'navItemAriaLabel': '@?',
```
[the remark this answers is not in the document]
MarcoZehe: Hm, what do you think needs updating? It only says that this adds alternative text for accessibility, and that is still a valid statement. It only internally adds it in a different spot.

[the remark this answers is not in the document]
MarcoZehe: OK, I've made it explicit that the aria-label is applied to the `<md-button>` element.

src/components/navBar/navBar.js
```diff
@@ -90,7 +90,7 @@ angular.module('material.components.navBar', ['material.core'])
  * Exactly one of the `md-nav-click`, `md-nav-href`, or `md-nav-sref` attributes are required
  * to be specified.
  *
- * @param {string=} aria-label Adds alternative text for accessibility.
+ * @param {string=} aria-label Adds alternative text for accessibility to the `<md-button>` element.
```
Splaktar: I think that my main point was that we need to rename the API from `aria-label` to `nav-item-aria-label`.

For the text, let's use "Allows setting or overriding the label that is announced by a screen reader for the nav item's button. If this is not set, the nav item's transcluded content will be announced. Make sure to set this if the nav item's transcluded content does not include descriptive text, for example only an icon."

MarcoZehe:
@Splaktar Thanks for the review! I also changed the documentation further up to point to the tablist and tab specifications instead of the outdated site navigator pattern.

src/components/navBar/navBar.js
```diff
@@ -498,7 +504,9 @@ function MdNavItem($mdAria, $$rAF, $mdUtil, $window) {
           });
         }
 
-        $mdAria.expectWithText(element, 'aria-label');
+        if (!ctrl.navItemAriaLabel) {
```
Splaktar: `ctrl` is not defined at this point. `mdNavItem` should be used instead.

src/components/navBar/navBar.js
```diff
@@ -127,7 +130,7 @@ function MdNavBar($mdAria, $mdTheming) {
     template:
       '<div class="md-nav-bar">' +
         '<nav role="navigation">' +
-          '<ul class="_md-nav-bar-list" ng-transclude role="listbox" ' +
+          '<ul class="_md-nav-bar-list" ng-transclude role="tablist" ' +
```
Splaktar: The docs on line 46 still refer to `listbox` and should be updated to tablist.

Splaktar: Similarly, as you noted in a comment, the docs on lines 34-41, need to be updated to point to the applicable tablist specs.

src/components/navBar/navBar.js
```diff
@@ -438,6 +441,9 @@ function MdNavItem($mdAria, $$rAF, $mdUtil, $window) {
             'ng-blur="ctrl.setFocused(false)" ' +
             'ng-disabled="ctrl.disabled" ' +
             'tabindex="-1" ' +
+            'role="tab" ' +
+            'aria-label="{{ctrl.navItemAriaLabel}}" ' +
```
Splaktar: We don't want to set an empty `aria-label` here as that will be flagged as an invalid aria attribute and may cause issues with certain screen readers.

If no `nav-item-aria-label` is defined, no `aria-label` attribute should be applied. We can do this using the following:
```
'ng-attr-aria-label="{{ctrl.navItemAriaLabel ? ctrl.navItemAriaLabel : undefined}}" ' +
```
I have a few examples of this in https://github.com/angular/material/pull/11441/files#diff-dfc7d490c731a45e66ad22d8efb7456dR358.

src/components/navBar/navBar.js
```diff
@@ -498,7 +504,9 @@ function MdNavItem($mdAria, $$rAF, $mdUtil, $window) {
           });
         }
 
-        $mdAria.expectWithText(element, 'aria-label');
+        if (!ctrl.navItemAriaLabel) {
+          $mdAria.expectWithText(element, 'aria-label');
```
Splaktar: This is still checking for the `aria-label` on the `li` element. Replacing `element` with `navButton` here seems to properly check the `md-button._md-nav-button` element.

src/components/navBar/navBar.spec.js
```diff
@@ -364,13 +364,13 @@ describe('mdNavBar', function() {
...
-      expect(getTab('tab1').parent().attr('aria-label')).toBe('tab1');
-      expect(getTab('tab2').parent().attr('aria-label')).toBe('tab2');
+      expect(getTab('tab1').attr('aria-label')).toBe('tab1');
+      expect(getTab('tab2').attr('aria-label')).toBe('tab2');
     });
 
     it('does not change aria-label on nav items', function() {
       createTabs();
-      expect(getTab('tab3').parent().attr('aria-label')).toBe('foo');
+      expect(getTab('tab3').attr('aria-label')).toBe('foo');
```
Splaktar: To ensure that we don't break any apps, let's continue to check that the `aria-label` API still works the way it did before:
```js
    it('does not change aria-label on nav items', function() {
      createTabs();
      expect(getTab('tab3').parent().attr('aria-label')).toBe('foo');
    });
```
Then let's add a new check for the `nav-item-aria-label`
```js
    it('does not change nav-item-aria-label on nav item buttons', function() {
      createTabs();
      expect(getTab('tab4').attr('aria-label')).toBe('foo');
    });
```
This will also require adding another tab to `createTabs()` around line 47:
```js
        '  <md-nav-item md-nav-href="#4" name="tab4" nav-item-aria-label="foo">' +
        '    tab4' +
        '  </md-nav-item>' +
```

Splaktar:
I've tracked down the CI failure and the disconnection issue to [this test](https://github.com/angular/material/blob/master/src/components/navBar/navBar.spec.js#L124-L132).

MarcoZehe:
> I've opened PR #11488 to fix the test issues here. Sorry that you had to run into them!

Thank you for this, and for your patience and guidance with my PR! I think I've addressed all your comments and will push an updated revision in a minute.

Splaktar:
@MarcoZehe OK, PR #11488 has been merged. Can you please rebase this and then force push to your branch? That should hopefully get all of the tests passing in TravisCI. Thank you!

MarcoZehe:
> 
> 
> @MarcoZehe OK, PR #11488 has been merged. Can you please rebase this and then force push to your branch? That should hopefully get all of the tests passing in TravisCI. Thank you!

Yes, all passed! \-O-/

Splaktar:
The first line of your commit is a bit too long. Can you please update it to just be "fix(nav-bar): improve screen reader support"? Thanks!

MarcoZehe:
> 
> 
> The first line of your commit is a bit too long. Can you please update it to just be "fix(nav-bar): improve screen reader support"? Thanks!

Done. :) Thank you again for your patience and guidance!

Splaktar:
Looks good, thank you! I'll let you know if we run into any issues in the "presubmit" process. This will run the change through all of the relevant tests in Google to make sure that it passes unit tests, screenshot tests, and a11y tests.

Splaktar:
I've submitted PR https://github.com/angular/material/pull/11494 which should improve the a11y state of nav-bar quite a bit more. I tested it with the changes in this PR and it worked well and doesn't seem to conflict.

PR #11491, opened 2018-10-18
fix(select): don't change form state when adding to a form

### LINKED ISSUE #11490: select(multiple): adding to a form causes the form to be marked pristine
Bug

The codepen creates a dynamic form which uses a ng-repeat with a filter tied to an input.   Inside the ng-repeat is a list of multi select elements, which will be added and removed as the filter changes.

1) Make a change to one of the selects in the form
2) NOTICE, is dirty is set to true
3) Type into the filter to reduce the number of controls shown
4) NOTICE, is dirty is still true
5) clear the filter causing selects to be added
6) NOTICE, is dirty is set to false

The form should remain dirty if it was already dirty and vice versa,
added or removing elements should not change the form state

Adding a md-select with the multiple option causes a side effect on the form setting it to pristine

Forms which are used to manage list of items typically use some type of filter to easily find and make modifications.  The issue prevents the use of these controls within a ng-repeat that uses a filter, or any other use that dynamically adds fields to a form.

- AngularJS: 1.7.5
- AngularJS Material:  1.1.10
- OS: tested on a mac
- Browsers: tested on Chrome, 69
 
From my understanding of this reading some comments and tracing around commits and issues.  It seems that at one point a related issue was created where a form would be marked dirty  when a md-select/multiple  was added.  It looks like Instead of fixing the root cause a hack was added to modify the form to set pristine on creation of this element. After that, similar issues like mine have been reported however instead of correcting the issue they built on top of the other hack to try to limit its, using the assumption that if a model was set on it it already that it must be a clean form (obviously a very bad assumption).    
see issue: https://github.com/angular/material/issues/6556

I currently have a fix for this which delays the initial render for the multiple variant.  Which allows the element to be created without triggering a dirty change on the form.  I still consider my update a hack but its a hack without side effects.

cstephe:
* Remove code which causes side effect on form elements
* Add a delay to the initial render of a multiple select so that it doesn't set the element to dirty on initialization.

Fixes #11490

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Issue Number: #11490
The select/multiple has code it in which causes a side-effect on the containing form by explicitly setting the form state to pristine anytime one is added with a set model value.  This errant code was introduced originally to try to fix an issue where during initialization the control thought it was being updated and would implicitly mark the form dirty.  This is obviously a deal breaker when using this control in dynamic forms.

## What is the new behavior?
Now the control initializes so that it does not think it is being updated and the "corrective" side-effect has been removed.  Adding and removing these controls from a dynamic form no longer changes the form state.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/select/select.js
+++ b/src/components/select/select.js
@@ -291,15 +291,6 @@ function SelectDirective($mdSelect, $mdUtil, $mdConstant, $mdTheming, $mdAria, $
       findSelectContainer();
       $mdTheming(element);
 
-      if (formCtrl && angular.isDefined(attr.multiple)) {
-        $mdUtil.nextTick(function() {
-          var hasModelValue = ngModelCtrl.$modelValue || ngModelCtrl.$viewValue;
-          if (hasModelValue) {
-            formCtrl.$setPristine();
-          }
-        });
-      }
-
       var originalRender = ngModelCtrl.$render;
       ngModelCtrl.$render = function() {
         originalRender();
@@ -673,13 +664,26 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
       if (deregisterCollectionWatch) deregisterCollectionWatch();
 
       if (self.isMultiple) {
+        // We want to delay the render method so that the directive has a chance to load before
+        // rendering, this prevents the control being marked as dirty onload.
+        var loaded = false;
+        var delayedRender = function(val) {
+          if (!loaded) {
+            $mdUtil.nextTick(function () {
+              renderMultiple(val);
+              loaded = true;
+            });
+          } else {
+            renderMultiple(val);
+          }
+        };
         ngModel.$validators['md-multiple'] = validateArray;
-        ngModel.$render = renderMultiple;
+        ngModel.$render = delayedRender;
 
         // watchCollection on the model because by default ngModel only watches the model's
-        // reference. This allowed the developer to also push and pop from their array.
+        // reference. This allows the developer to also push and pop from their array.
         $scope.$watchCollection(self.modelBinding, function(value) {
-          if (validateArray(value)) renderMultiple(value);
+          if (validateArray(value)) delayedRender(value);
         });
 
         ngModel.$isEmpty = function(value) {
--- a/src/components/select/select.spec.js
+++ b/src/components/select/select.spec.js
@@ -507,7 +507,7 @@ describe('<md-select>', function() {
 
       $rootScope.$digest();
       $rootScope.$digest();
-
+      $timeout.flush();
       expect(label.text()).toBe('One, Three');
       expect(label.hasClass('md-select-placeholder')).toBe(false);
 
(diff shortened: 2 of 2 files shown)


cstephe:
I already have a CLA, any help on resolving that check point?

src/components/select/select.js
```diff
@@ -673,13 +664,26 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
       if (deregisterCollectionWatch) deregisterCollectionWatch();
 
       if (self.isMultiple) {
+        // We want to delay the render method so that the directive has a chance to load before
+        // rendering, this prevents the control being marked as dirty onload.
+        var loaded = false;
+        var delayedRender = function(val){
```
Splaktar: Please add a space here `){` -> `) {`

src/components/select/select.js
```diff
@@ -673,13 +664,26 @@ function SelectMenuDirective($parse, $mdUtil, $mdConstant, $mdTheming) {
...
+          } else {
+            renderMultiple(val);
+          }
+        };
         ngModel.$validators['md-multiple'] = validateArray;
-        ngModel.$render = renderMultiple;
+        ngModel.$render = delayedRender;
 
         // watchCollection on the model because by default ngModel only watches the model's
         // reference. This allowed the developer to also push and pop from their array.
```
Splaktar: Can you update this to "This allows the developer to also push and pop from their array." while you are in here please?

src/components/select/select.spec.js
```diff
@@ -1087,6 +1087,37 @@ describe('<md-select>', function() {
...
+            $timeout.flush();
+            expect(sut.find('md-select').length).toBe(1);
+            expect($rootScope.testForm.$pristine).toBeFalsy();
+            $rootScope.filterTerm = "";
+            $rootScope.$apply();
+            $timeout.flush();
+            expect(sut.find('md-select').length).toBe(2);
+            expect($rootScope.testForm.$dirty).toBeTruthy();
+            expect($rootScope.testForm.$pristine).toBeFalsy();
+
```
Splaktar: Please remove blank line.

src/components/select/select.spec.js
```diff
@@ -1087,6 +1087,37 @@ describe('<md-select>', function() {
         expect($rootScope.testForm.$pristine).toBe(true);
       });
 
+        it('should keep the form dirty form is already dirty', function() {
```
Splaktar: I think that "should not change a dirty form to pristine" is more clear.

src/components/select/select.spec.js
```diff
@@ -1087,6 +1087,37 @@ describe('<md-select>', function() {
...
+            $rootScope.rows = [[2], [4,3]];
+            $rootScope.filterTerm;
+            $rootScope.opts = [1, 2, 3, 4];
+            var sut = $compile('<form name="testForm">' +
```
Splaktar: Can we rename `sut` to `select` please? It's not clear what that acronym stands for, if it is an acronym.

src/components/select/select.js
cstephe: cool

src/components/select/select.js
cstephe: np

src/components/select/select.spec.js
cstephe: will do

src/components/select/select.spec.js
cstephe: ok

src/components/select/select.spec.js
cstephe: yeah I'll update it,  it stood for subject under test,

cstephe:
@Splaktar I'm not sure if i'm supposed to resolve the conversations for the PR comments.  I did one, then I thought, maybe thats for you to resolve.

PR #11500, opened 2018-11-01
fix(autocomplete): updateSize incorrectly sets the size to zero

### LINKED ISSUE #10834: autocomplete: suggestions don't show after $md-resize raised while something is selected
**Actual Behavior**:
 -  `What is the issue? *` Suggestions don't show after clicking X or by clicking on input.
 -  `What is the expected behavior?` Suggestions should show after clicking X or by clicking on input.

 -  `CodePen Demo which shows your issue:` https://codepen.io/anon/pen/Vzmyem (this is not changed basic demo from main angular material webpage)
 -  `Details:` Choose something from autocomplete. Then resize browser window. Then click X. After that suggestions will not be displayed like normally.

**AngularJS Versions**: *
 -  `AngularJS Version:` 1.5.7

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *` 59.0.3071.115
 -  `OS: *` Windows

My fix which works, but I don't know if it is proper fix:
```
VirtualRepeatContainerController.prototype.updateSize = function() {
  // If the original size is already determined, we can skip the update.
  if (this.originalSize) return;

  var size = this.isHorizontal()
      ? this.$element[0].clientWidth
      : this.$element[0].clientHeight;
  
  if (size) {
    this.size = size;
  }

  // Recheck the scroll position after updating the size. This resolves
  // problems that can result if the scroll position was measured while the
  // element was display: none or detached from the document.
  this.handleScroll_();

  this.repeater && this.repeater.containerUpdated();
};
```

This is fix for this line: https://github.com/angular/material/blob/master/src/components/virtualRepeat/virtual-repeater.js#L225

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

In some cases, like resizing the page, updateSize incorrectly sets the size to zero.
This can break the ability to show the dropdown on focus.

Issue Number: 
Fixes #10834

## What is the new behavior?
Check that the size value isn't zero before setting the size.

```
[ ] Yes
[x] No
```

## Other information
Manually tested as I wasn't sure how to resize the browser in a unit test.

### DIFF
--- a/src/components/virtualRepeat/virtual-repeater.js
+++ b/src/components/virtualRepeat/virtual-repeater.js
@@ -223,9 +223,13 @@ VirtualRepeatContainerController.prototype.updateSize = function() {
   // If the original size is already determined, we can skip the update.
   if (this.originalSize) return;
 
-  this.size = this.isHorizontal()
+  var size = this.isHorizontal()
       ? this.$element[0].clientWidth
       : this.$element[0].clientHeight;
+
+  if (size) {
+    this.setSize_(size);
+  }
 
   // Recheck the scroll position after updating the size. This resolves
   // problems that can result if the scroll position was measured while the


andrewseguin:
Reverting this change because it broke some internal tests. It seems that the behavior is now changed so that re-evaluating the size changes the element height through setSize.

Splaktar:
Fixed the regression caused by this issue in https://github.com/angular/material/pull/11518.

PR #11517, opened 2018-11-15
fix(slider): enable page up/down and home/end

### LINKED ISSUE #11515: md-slider is not up to date with WAI-ARIA spec

Bug 

This can be reproduced at [material demo page](https://material.angularjs.org/latest/demo/slider). Basically up/down arrows, and home/end keys are not properly handled per spec.

The slider directive is fully complaint with WAI-ARIA spec here: https://www.w3.org/TR/wai-aria-practices-1.1/#slider

Left/right arrows work, but the rest of the keyboard shortcuts do not work.

To be complaint with ARIA spec

- AngularJS: All
- AngularJS Material: All
- OS: All
- Screen reader: All
- Browsers: All

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

- keyboard interaction with page up/down and home/end are not supported
- keyboard up/down/left/right are orientation-specific. This means that up/down don't work in horizontal mode.

Issue Number: 
Fixes #11515

## What is the new behavior?
- fix up/down/left/right to not be orientation-specific
- based on https://www.w3.org/TR/wai-aria-practices-1.1/#slider

```
[ ] Yes
[x] No
```


src/components/slider/slider.js
```diff
@@ -382,29 +382,63 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
...
+          changeAmount = -step;
+          break;
+        case keyCodes.UP_ARROW:
+        case keyCodes.RIGHT_ARROW:
+          ev.preventDefault();
+          changeAmount = step;
+          break;
+        case keyCodes.PAGE_DOWN:
+          ev.preventDefault();
+          changeAmount = -step * 4;
```
josephperrott: Why is the larger step, `step * 4`?  4 seems to be an arbitrary value.

Splaktar: In the ARIA practices examples they jump 10 steps:
https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/slider-1.html
https://www.w3.org/TR/wai-aria-practices-1.1/examples/slider/slider-2.html

But the spec only says:
> Page Up (Optional): Increment the slider by an amount larger than the step change made by Up Arrow.

We could make this configurable, but I would prefer to do that in a separate issue and PR.

Splaktar: We also support moving 4 steps by pressing the meta/alt/ctrl key along with the arrow key. For now this step size is the same as paging.

Do you think that they should be different?
- arrow + modifying = 4 steps
- page up/down = 10 steps?

Splaktar: The arrow + modifier is pre-existing behavior: https://github.com/angular/material/pull/11517/files#diff-19f59514041130aaa9c83ecc89c23536R425

Splaktar: I've made the change to 4 and 10 steps as mentioned above. I've also updated the docs.

PR #11564, opened 2018-12-19
fix(radio-group): style focused but unchecked radio buttons

### LINKED ISSUE #3643: No focus style for md-radio-group when no button is selected
Focus is shown on a radio group with at least one selected item (as an ink ripple around the selected item), but not on a radio group with no selected items. This is contrary to the spec:

https://www.google.com/design/spec/components/selection-controls.html#selection-controls-radio-button

To reproduce:

0. Go to the [radio button demo page](https://material.angularjs.org/latest/#/demo/material.components.radioButton)
0. Remove all the radio buttons from the second demo using the "remove" button
0. Add back some items using the "add" button.
0. Tab into the radio group and notice that there's no way to tell it's focused.

### LINKED ISSUE #8339: md-radio-group has no focus indicator
There is no indication of focus for a radio group without a selected button.

Without a focus indicator, a user won't know when they are in a radio group and can choose a radio button with the arrow keys.

Other focusable elements such as checkboxes, inputs and buttons have focus indication.

This can be observed with the following code pen:
http://codepen.io/eegz/pen/RaEVom

1. Select a radio button in the first group, then tab to the second radio group and back.
2. Observe that the first radio group has affordance because the selected radio button has a focus effect.
3. Observe that the second radio group has no focus affordance.

Angular: 1.4.8
Angular Material 1.0.8

### LINKED ISSUE #11563: radio-group: no indication of focus when no option selected
## Bug, enhancement request, or proposal: Bug

1. In the attached code pen focus the first input box
1. Tab to the md-radio-group

I expect that there will be some indication of focus. In the docs there is a defined "focused state" for radio inputs. 

The current behavior is that there is no indication of focus.

I am currently implementing a questionnaire that receives questions one by one. I would like to focus controls as they become available so users can select an option and hit enter to submit the form. This is not obvious if radio groups don't indicate focus.

- AngularJS: Latest
- AngularJS Material: Latest 
- OS: All
- Browsers: All 
 
This is the archived material docs showing the "off / focused" radio buttons

image

I think the best solution to this problem while retaining a11y is to change how the arrow keys modify radio selections.

Currently focusing a radio group allows you to use the arrow keys to change the selected radio button. If we were to instead have the arrow keys change which radio button is focused and allow the enter key to change the selection. We could set the either the first radio button to focused given that no selection is made, or we could focus the selected radio button given that a selection has been made.

Now when focusing a radio group for the first time the first radio button would go into a focused state. 

image

If a selection has already been made that radio would be selected + focused

image

I see a couple of closed tickets from long ago that sort of bring up this issue in a lot less detail, but they seem to have been closed without much explanation.

https://github.com/angular/material/issues/8339

https://github.com/angular/material/issues/3643

Now that the project is very much focused on a11y I am hoping this can be reconsidered. I am willing to spend some of my own time on this if the team decides this is an acceptable change.

Thanks!

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Based on the MD spec (image below), we fail to provide a decoration for focused but unselected radio buttons (whether disabled or enabled).

Issue Number: 
Fixes #11563. Fixes #8339. Fixes #3643.

## What is the new behavior?
Better align with the MD spec by decorating the focused but unselected radio buttons (whether disabled or enabled).
radio-buttons

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/radioButton/radio-button-theme.scss
+++ b/src/components/radioButton/radio-button-theme.scss
@@ -89,6 +89,12 @@ md-radio-group.md-THEME_NAME-theme {
   }
 }
 
+md-radio-group.md-THEME_NAME-theme.md-focused.ng-empty>md-radio-button:first-child {
+  .md-container:before {
+    background-color: '{{foreground-3-0.26}}';
+  }
+}
+
 md-radio-group.md-THEME_NAME-theme.md-focused:not(:empty) {
   .md-checked .md-container:before {
     background-color: '{{accent-color-0.26}}';
--- a/src/components/radioButton/radio-button.scss
+++ b/src/components/radioButton/radio-button.scss
@@ -149,8 +149,16 @@ md-radio-group {
     outline: none;
   }
 
-  &.md-focused {
+  &.md-focused.ng-not-empty {
     .md-checked .md-container:before {
+      left: -8px;
+      top: -8px;
+      right: -8px;
+      bottom: -8px;
+    }
+  }
+  &.md-focused.ng-empty>md-radio-button:first-child {
+    .md-container:before {
       left: -8px;
       top: -8px;
       right: -8px;


Splaktar:
@codymikol can you please take a look at this?

You can test it some by using this for your CSS (https://drive.google.com/uc?export=download&id=1Vw_o1NPFcKMqTc09ocKQsEPWn42sDvW9).

codymikol:
I tried that CSS without any success. I also did a clean build from your commit to be sure and it seems the issue is still present. 

In the conversation over here https://github.com/angular/material/issues/11563

You mentioned that with VoiceOver software you are able to tab focus onto md-radio-buttons themselves. Are you able to tab over each of these md-radio-buttons?

 In my codepen when tabbing from the first input I am able to set focus to the md-radio-group, but not any of the md-radio-buttons inside. Another tab will set focus to the next input in the form.

Splaktar:
Sorry, the theme CSS is applied by the JS, so you need that one too. Here's an updated [CodePen](https://codepen.io/Splaktar/pen/bOgNLd?editors=1010) using it. You can get the links to the .js and .css from its settings.

No there is no tabbing over the radio buttons. That behavior is the same, you tab to the radio group, then use the arrow keys to move around the radio-buttons.

codymikol:
Yeah, that definitely fixes it. Awesome job, thanks for such a quick fix! 🍻

codymikol:
Thanks @mmalerba 🐔

PR #11570, opened 2018-12-29
build: make the build output deterministic and reproducible

### LINKED ISSUE #11502: build: make build output deterministic
Currently the generated JS and CSS files have the output emitted in a different order every time they're run. When we sync the code into Google, this causes noise in the diff. It would make things a tiny bit smoother if the output was deterministically ordered.

Splaktar:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Enhancement
[x] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Consecutive builds, without any changes, result in bundle content with different ordering. This is even true in a single build where the `dist/angular-material.js` isn't identical to the `dist/docs/angular-material.js`. This makes comparing releases harder and it makes syncing into g3 more time consuming.

- passing globs to `gulp.src` can cause ordering to not be preserved.
this is because `gulp.src` is async
- gulp-concat tries to preserve the order passed into `gulp.src`.
if the globbing is too vague, then there isn't enough info for ordering

Issue Number: 
Fixes #11502

## What is the new behavior?
All of the build output (JS, SCSS, CSS) should be identical if the build is run multiple times without any changes to the code. 

```
[ ] Yes
[x] No
```

## Other information
- more statically defining the paths like this isn't ideal but it is acceptable atm since we aren't adding new components.

I looked at sorting the file paths manually or using `gulp-sort` but neither solution was flexible enough to order some of our internal dependencies properly. This approach seems to work fine w/o adding `gulp-order`, but if problems pop up in the future, that is another package that might help.

Note that the root cause of this lies in the intersection of `gulp.src` in Gulp core and `gulp-concat`. This is a well established problem that the Gulp team has mostly decided to punt into user land or third party packages so that they can maintain speed and not need to support crazy levels of customization.

### DIFF
--- a/gulp/util.js
+++ b/gulp/util.js
@@ -40,34 +40,38 @@ exports.args = args;
 /**
  * Builds the entire component library javascript.
  */
-function buildJs () {
-  const jsFiles = config.jsBaseFiles.concat([path.join(config.paths, '*.js')]);
+function buildJs() {
+  const jsFiles = config.jsCoreFiles;
+  config.componentPaths.forEach(component => {
+    jsFiles.push(path.join(component, '*.js'));
+    jsFiles.push(path.join(component, '**/*.js'));
+  });
 
   gutil.log("building js files...");
 
   const jsBuildStream = gulp.src(jsFiles)
-      .pipe(filterNonCodeFiles())
-      .pipe(utils.buildNgMaterialDefinition())
-      .pipe(plumber())
-      .pipe(ngAnnotate())
-      .pipe(utils.addJsWrapper(true));
+  .pipe(filterNonCodeFiles())
+  .pipe(utils.buildNgMaterialDefinition())
+  .pipe(plumber())
+  .pipe(ngAnnotate())
+  .pipe(utils.addJsWrapper(true));
 
   const jsProcess = series(jsBuildStream, themeBuildStream())
-      .pipe(concat('angular-material.js'))
-      .pipe(BUILD_MODE.transform())
-      .pipe(insert.prepend(config.banner))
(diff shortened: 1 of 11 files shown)


jelbourn:
Caretaker: this PR should be synced on its own. It should be a complete no-op, but on the off chance that something *does* break, we'll know it's from this and not another PR.

gulp/util.js
```diff
@@ -224,14 +228,18 @@ function filterNonCodeFiles() {
...
-      .pipe(concat('default-theme.scss'))
-      .pipe(utils.hoistScssVariables())
-      .pipe(sass())
-      .pipe(dedupeCss())
-      // The PostCSS orderedValues plugin modifies the theme color expressions.
-      .pipe(minifyCss({ orderedValues: false }))
-      .pipe(utils.cssToNgConstant('material.core', '$MD_THEME_CSS'));
+  var paths = config.themeBaseFiles;
+  config.componentPaths.forEach(component => paths.push(path.join(component, '*-theme.scss')));
+  paths.push(config.themeCore);
```
Splaktar: This fixes the issue where some apps had a white background instead of light gray in the presubmits.

Splaktar:
It looks like the module builds of the CSS (closure or js) are injecting tons of extra theme template styles into component CSS like `icon.css` and also `icon-default-theme.css`. Both have increased from ~400 bytes to ~190KB!

This PR tried to ignore the module based builds, but it's clear that I need to go through there and make some tweaks + verify.

Splaktar:
Duplicate CSS in module-based builds has been resolved. This is ready for presubmit again.

PR #11575, opened 2019-01-03
fix(autocomplete): show dropdown on top only when there is room

### LINKED ISSUE #10859: autocomplete: shows dropdown on top even when there is no room
**Actual Behavior**:
 -  `What is the issue? *`
Sometimes the autocomplete dropdown will be shown at the `top` of an input, partially out of the screen, but it should be `bottom` position in this case.

 -  `CodePen Demo which shows your issue:` https://codepen.io/dumchikov/pen/zdRrYO

#### Steps to reproduce
   - The `md-autocomplete` should not have `md-dropdown-position` (position should be determined automatically) 
   -  The `md-autocomplete`'s input should be placed below the screen, so the window should be scrolled to see the input.
   - Scroll the window until the `md-autocomplete`'s input gets the top of the window
   - Focus the input 
   - The autocomplete dropdown will be shown at the top of input partially out of the screen, but should be at the input's bottom

**AngularJS Versions**: *
 -  `AngularJS Version:` 1.5.5

**Additional Information**:
 -  `Browser Type: *` Chrome, FF, IE
 -  `Browser Version: *` Chrome - 60.0.3112.90, FF - 54.0.1, IE - 11
 -  `OS: *` Window 7

 Another issue reproduced on FF and IE only - after closing the dropdown the window is scrolled to the window top.

marosoft:
Don't use the bounding rectangles of a document body to determine
the position of the dropdown.
Top and bottom values of the snap element are relative to the viewport,
which is enough to decide about the position of the dropdown.

Fixes #10859

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Dropdown is shown on top even if there is not enough space for it.
Issue Number: #10859

## What is the new behavior?
A proper checks are made to determine if the dropdown should be shown on top or under.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -140,7 +140,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
 
     // Automatically determine dropdown placement based on available space in viewport.
     if (!position) {
-      position = (top > bot && root.height - top - MENU_PADDING < dropdownHeight) ? 'top' : 'bottom';
+      position = (vrect.top + MENU_PADDING > dropdownHeight) ? 'top' : 'bottom';
     }
     // Adjust the width to account for the padding provided by `md-input-container`
     if ($attrs.mdFloatingLabel) {


Splaktar:
We're currently investigating a regression that "might" be caused by this. PR https://github.com/angular/material/pull/11626 reverts this change and is being tested to see if it solves the issue.

PR #11581, opened 2019-01-07
fix(sidenav): allow to hide side nav depending on the media query

### LINKED ISSUE #4595: Sidenav component does not automatically handle entering hidden state.

marosoft:
Current behavior is not correct

Fixes #4595

## PR Checklist

## PR Type

```
[x] Bugfix
[x] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

At the moment it is not possible to hide side nav without breaking the view - the overlays stays visible.
Issue Number: #4595

## What is the new behavior?
A new attribute allows define when the side nav should hide depending on the media query.

```
[ ] Yes
[x] No
```


src/components/sidenav/sidenav.js
```diff
@@ -244,16 +245,19 @@ function SidenavFocusDirective() {
  * @param {expression=} md-is-locked-open When this expression evaluates to true,
  * the sidenav 'locks open': it falls into the content's flow instead
  * of appearing over it. This overrides the `md-is-open` attribute.
```
Splaktar: In the MD Spec's [Surface Behaviors](https://material.io/archive/guidelines/layout/responsive-ui.html#responsive-ui-surface-behaviors) when this value is `true`, it maps to `Permanent`. The default when this is `false` is `Temporary`. If you want to implement the `Persistant` behavior, then you need to allow the value of `md-is-locked-open` to be toggled like in this [CodePen example](https://codepen.io/Splaktar/pen/aPjBpV?editors=1010#0).

We should link to this part of the spec in the main SideNav docs (above) and mention these behavior names. We also need to have demos of each type of behavior (could be as part of 1 or 2 demos).

src/components/sidenav/sidenav.js
```diff
@@ -244,16 +245,19 @@ function SidenavFocusDirective() {
...
+ * @param {expression=} md-is-locked-closed When this expression evaluates to true,
+ * the sidenav 'locks closed': hides sidenav.
```
Splaktar: I'm still trying to get a handle on what behavior this API really has and how to name it.

Based on the naming, `md-is-locked-closed="true"` should make it so that you can't open the sidenav at all, even via `toggle()`. But in this [CodePen](https://codepen.io/Splaktar/pen/rorWzw) you can see that is not the case. This doesn't mean that your behavior is incorrect, it just means that we have the wrong naming of this API here.

Splaktar:
@marosoft let's revisit this one for `1.1.23`?

src/components/sidenav/sidenav.js
marosoft: @Splaktar what if I change it to `md-is-hidden`, `is-hidden-when` or `md-hide-when`?
Basically by specifying the condition you determine when the sidenav has to become closed after it was opened.

Splaktar:
It would be nice if we had a `md-is-persistant` API that influenced the behavior of the `md-is-open` expression.

So if `md-is-persistant` is `false` or not defined, then `md-is-open` expression would open and close a `Temporary` overlay. 

If `md-is-persistant` is `true`, then the`md-is-open` expression would open and close a `Persistant` overlay.

Issue #11592, opened 2019-01-10
ci: migrate from TravisCI to CircleCI


Splaktar:
### What is the expected behavior?
That we migrate over to CircleCI.

### What is the current behavior?
We run most all of our CI on TravisCI. In general, it has been working fine other than some outages. We recently setup Firefox and Chrome to run in Headless mode and removed our use of Saucelabs.

### What is the use-case or motivation for changing an existing behavior?
Our paid plan with TravisCI will end on January 31st, 2019. This will lead to longer queue times for builds. The Angular team has moved over to CircleCI for angular/angular, angular/material, and some of the other repos. We need to do the same in order to benefit from faster builds, paid support, etc.

### Which versions of AngularJS, Material, OS, and browsers are affected?

- AngularJS: 1.7.5
- AngularJS Material: 1.1.12
- OS: macOS
- Browsers: Chrome, Firefox
 
## TODO
- [x] Configure CircleCI to run the tests against AngularJS 1.5, 1.6, 1.7, and snapshot for `master` and PRs
- [x] Enable CircleCI for `angular/material`
- [x] Define `ANGULARJS_MATERIAL_BOWER_TOKEN` and `ANGULARJS_MATERIAL_DOCS_SITE_TOKEN ` Github Push Tokens as CircleCI environment variables
- [x] Disable TravisCI completely
- [x] Enable CircleCI to deploy to bower-material and update the `HEAD` docs site on commits to `master`

Splaktar:
CircleCI is configured and appears to be working correctly for PRs and commits to `master`.

PR #11596, opened 2019-01-14
chore(ci): enable CircleCI

Splaktar:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Enhancement
[ ] Build related changes
[x] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

We only run CI on Travis.

Issue Number: 
Relates to #11592

## What is the new behavior?
We run CI on CircleCI now too.

We'll disable Travis in future PRs. See #11592 for a list of related tasks.

```
[ ] Yes
[x] No
```


scripts/circleci/update-snapshot-and-docs.sh
```diff
@@ -0,0 +1,28 @@
...
+
+function init {
+  # If --git-push-dryrun or --verbose are set, be sure to export them
+  # so they are set in all the other scripts too
+  export GIT_PUSH_DRYRUN=$GIT_PUSH_DRYRUN
+  export VERBOSE=$VERBOSE
+}
+
+function run {
+  if [[ "$CIRCLE_PULL_REQUEST" != "false" ]]; then
```
Splaktar: Running this file isn't enabled yet. I want to do some more testing and disable this in Travis before enabling it. I am not sure that this `CIRCLE_PULL_REQUEST` env var will work similar to `TRAVIS_PULL_REQUEST`.

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+var_8: &rebuild_node_sass
+  run: npm rebuild node-sass
+
+var_9: &install_gulp
+  run: sudo npm i -g gulp@^3
+
+var_10: &store_artifacts
+  store_artifacts:
+    path: dist
+    destination: /angularjs_material/material_build_output
```
devversion: Do we need this?

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+# in sync. Unlike in Travis, we don't need to manually purge the node modules if stale because
+# CircleCI automatically discards the cache if the checksum of the lock file has changed.
+var_7: &npm_install
+  run: npm ci
+
+var_8: &rebuild_node_sass
+  run: npm rebuild node-sass
+
+var_9: &install_gulp
+  run: sudo npm i -g gulp@^3
```
devversion: We should just add `gulp` as a dev dependency and use it using from NPM.

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+      - *install_gulp
+      - run: ./scripts/circleci/run-tests.sh
+      - *store_test_results
+
+# ----------------------------------------------------------------------------------------
+# Workflow definitions. A workflow usually groups multiple jobs together. This is useful if
+# one job depends on another.
+#
+# NOTE: When updating this configuration section, make sure to update GitHub robot
+#       config to match the new workflow jobs.
```
devversion: Are you using the Github robot?

scripts/circleci/update-snapshot-and-docs.sh
devversion: It should work similar to that. Though you need to either filter out branches or have a branch check here, because otherwise commits to upstream branches would cause this assertion to be true.

.circleci/config.yml
```diff
@@ -0,0 +1,182 @@
...
+
+var_9: &install_gulp
+  run: sudo npm i -g gulp@^3
+
+var_10: &store_artifacts
+  store_artifacts:
+    path: dist
+    destination: /angularjs_material/material_build_output
+
+var_11: &store_test_results
```
devversion: `store_junit_test_results`

.circleci/config.yml
Splaktar: It's not doing anything useful at this point in time, no.

.circleci/config.yml
Splaktar: Yeah, this was just easier than adding a bunch of new NPM `script` entries, but I can do that instead.

.circleci/config.yml
Splaktar: Nope, will remove this comment.

Splaktar:
Yep, https://github.com/angular/material/issues/11592 details the various follow-ups and steps.

Issue #11591, opened 2019-01-10
nav-bar: clicking a tab causes focus to move to a previously selected tab


rohit1018:
#### Detailed Reproduction Steps: 
1. Click on below  md-nav-bar items
https://material.angularjs.org/latest/demo/navBar
### What is the expected behavior?
md-nav-bar focus remove on previously selected nav-item

### What is the current behavior?
md-nav-bar focus shows on previously selected nav-item

### Which versions of AngularJS, Material, OS, and browsers are affected?

- AngularJS: 1.5.11
- AngularJS MateriaC: 1.1.12
- OS: Mac
- Browsers: Chrome
 
### Is there anything else we should know? Stack Traces, Screenshots, etc.
<img width="904" alt="screen shot" src="https://user-images.githubusercontent.com/28265981/50940432-f4e28000-144e-11e9-8f0c-cf745e96f171.png">

https://material.angularjs.org/latest/demo/navBar

codymikol:
Your codepen is not valid, but this is reproducible in the docs here https://material.angularjs.org/latest/demo/navBar

codymikol:
Looks like this is a regression introduced in 1.1.11

rohit1018:
> I've Isolated this to a unit test, now working on a fix.
> 
> This is actually a difficult issue, it looks like the focus handling is fired immediately, but there are a bunch of timeouts on updating the actual state of the tabs so we have to wait for that to complete before setting focus.
> 
> It also looks like the updating functionality is called inside of a timeout and then internally it also calls a bunch of timeouts.
> 
> EDIT: This completely fixes the issue and I absolutely cannot see anything wrong with this
> 
> ![image](https://user-images.githubusercontent.com/13606342/50945686-df2b8580-1463-11e9-8769-2bc39d71d7cf.png)
> 
> 20 timeouts didn't seem to do the trick, but 30 works great.
> 
> (Obviously this is a joke, please don't do this)
> 
> Still working on an elegant solution, hopefully I'll have more luck tomorrow.

Okay thank you

Splaktar:
Thank you for taking the time to submit this issue. However it does not follow our **required** issue template. Providing a CodePen demo or link to a demo in the docs **along with** reproduction steps is required. This gives the team the information needed to investigate the bug or prioritize the feature.

Please provide reproduction steps as I do not see an issue with the existing demo. Note that pressing the arrow keys only moves the focus, it does not change the tab selection. You need to press the space bar for the tab to be selected. 

[...]

PR #11598, opened 2019-01-15
fix(navbar): fix navbar focus on click

codymikol:
add a $timeout that will allow enough time for the click event to
complete so the focus event can properly evaluate

Fixes: #11591

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Currently clicking on a navbar tab will select it, but pull focus to the previously selected tab.

Issue Number: #11591

## What is the new behavior?

Now the click event will evaluate before the focus event, updating the state which is necessary to set the focus styling.

```
[ ] Yes
[x] No
```

## Other information

I spent a lot of time spinning my wheels on this, it turns out that there is just a significant delay from when the focus event is fired and when the click event is fired.

### DIFF
--- a/src/components/navBar/navBar.js
+++ b/src/components/navBar/navBar.js
@@ -493,10 +493,11 @@ MdNavBarController.prototype.onKeydown = function(e) {
  * @param $$rAF
  * @param $mdUtil
  * @param $window
+ * @param $timeout
  * @constructor
  * @ngInject
  */
-function MdNavItem($mdAria, $$rAF, $mdUtil, $window) {
+function MdNavItem($mdAria, $$rAF, $mdUtil, $window, $timeout) {
   return {
     restrict: 'E',
     require: ['mdNavItem', '^mdNavBar'],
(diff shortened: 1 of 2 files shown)


codymikol:
Looks like the project needs a full lint, travis failed because of something in autocomplete.

codymikol:
I've also realized that this fix does not work 100% of the time. It looks like sometimes it can take more than 100ms between the focus and click events. @Splaktar do you know of any other instances in the project that have to work around this quirk?

codymikol:
It seems to work perfectly with a 200ms delay, but it does feel a bit "wrong" to rely on a timeout to work around this behavior 🤷‍♂️

Splaktar:
The lint issue is fixed in `master`. If you rebase it will be resolved.

I will have to dig into this a bit deeper next week. I agree with your assessment that we should find a better way than just a longer timeout, if possible.

Splaktar:
I think that I found a better way in PR https://github.com/angular/material/pull/11600. Please review and try it out if possible and let me know what you think.

codymikol:
Closing as you've come up with a much better solution 🐔

PR #11608, opened 2019-01-23
chore(ci): enable CircleCI snapshot and doc publishing

Splaktar:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Enhancement
[ ] Build related changes
[x] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

CircleCI is not enabled to create snapshots, add tags, or update the docs on commits to `master`.

Issue Number: 
Relates to #11592

## What is the new behavior?
After a commit to master, CircleCI runs lint and tests against different AngularJS versions. If those are successful and the commit was to the `master` branch, then it snapshots, commits to `angular/code.material.angularjs.org`, and deploys the updated docs site. It also commits and tags the code snapshot in `angular/bower-material`.

```
[ ] Yes
[x] No
```

## Other information
The list of tasks related to this and their order can be found in https://github.com/angular/material/issues/11592.

**Caretaker note:** This can't be merged until the GitHub tokens get set up in the CircleCI environment variables as described in #11592.


.circleci/config.yml
```diff
@@ -137,6 +145,23 @@ jobs:
...
       - *store_junit_test_results
 
+  # ------------------------------------------------------------------------------------------
+  # Jobs that snapshot the `master` branch and update the docs on commits to master
+  # ------------------------------------------------------------------------------------------
+  bower_material_snapshot:
+    <<: *job_defaults
+    steps:
+      - *checkout_code
+      - *restore_cache
```
devversion: Missing npm install?

.circleci/config.yml
```diff
@@ -137,6 +145,23 @@ jobs:
...
+      - run: ./scripts/bower-material-release.sh
+
+  update_and_snapshot_docs:
+    <<: *job_defaults
+    steps:
+      - *checkout_code
+      - *restore_cache
```
devversion: Missing NPM install?

.circleci/config.yml
Splaktar: The thought was that we don't need to do `npm ci` here since the cache includes `node_modules/` from the build that is required before this job.

devversion: That's a fair point, but usually we just run `yarn`/`npm` again in order ensure that the modules are properly installed (these calls are a noop if the integrity is verified)

Also technically the cache is not always guaranteed to be stored as part of the same cache key. e.g. something changes the lock file in previous builds

Splaktar:
@andrewseguin has set up the GitHub tokens in the CircleCI environment variables as described in #11592.

PR #11625, opened 2019-01-31
fix(autocomplete): tap outside options panel on iOS does not close panel

### LINKED ISSUE #9581: autocomplete: suggestion box not closing on blur
Reopening #7419 as suggested by @EladBezalel 

**Actual Behavior**:
Inputs are not blurred when tapping outside of a focussed input element.

The autocomplete suggestion box is not closing when the user taps outside the autocomplete to close the suggestion box without selecting a value.

This only happens on touch devices. Tested on Android/Chrome 52.0.2743.98  and iPhone/Safari 9. 

When using the "Done" on iOS keyboard the suggestion box closes. 
Hiding the keyboard on Android did not close the suggestion box.

**Expected Behavior**:
The autocomplete suggestion box should close when the user blurs the control by clicking/touching outside of it.

**Additional Info**

According to the comments in #7419 it worked in 0.11.4.

The issue might origin from function `mouseInputHijacker` and a quick-and-dirty fix for the autocomplete may be applying a css rule:

``` css
.md-scroll-mask {
  display: none;
}
```

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Clicking outside of the options dropdown panel for `md-autocomplete` does not dismiss the panel on iOS.

Issue Number: 
Fixes #9581.

## What is the new behavior?
Clicking outside of the options dropdown panel for `md-autocomplete` dismisses the panel on iOS. It is a similar approach to what the `md-datepicker` uses.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -27,6 +27,15 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       debouncedOnResize    = $mdUtil.debounce(onWindowResize),
       mode                 = MODE_VIRTUAL; // default
 
+  /**
+   * The root document element. This is used for attaching a top-level click handler to
+   * close the options panel when a click outside said panel occurs. We use `documentElement`
+   * instead of body because, when scrolling is disabled, some browsers consider the body element
+   * to be completely off the screen and propagate events directly to the html element.
+   * @type {!angular.JQLite}
+   */
+  ctrl.documentElement = angular.element(document.documentElement);
+
   // Public Exported Variables with handlers
   defineProperty('hidden', handleHiddenChange, true);
 
@@ -356,8 +365,10 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
       if (elements) {
         $mdUtil.disableScrollAround(elements.ul);
         enableWrapScroll = disableElementScrollEvents(angular.element(elements.wrap));
+        ctrl.documentElement.on('click', handleClickOutside);
       }
     } else if (hidden && !oldHidden) {
+      ctrl.documentElement.off('click', handleClickOutside);
       $mdUtil.enableScrolling();
 
       if (enableWrapScroll) {
@@ -365,6 +376,15 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
         enableWrapScroll = null;
       }
     }
+  }
+
+  /**
+   * Handling click events that bubble up to the document is required for closing the dropdown
+   * panel on click outside of the panel on iOS.
+   * @param {Event} $event
+   */
+  function handleClickOutside($event) {
+    ctrl.hidden = true;
   }
 
   /**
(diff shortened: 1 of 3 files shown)


AdriVanHoudt:
Is there any way you can land this in the next patch release, this completely breaks forms on iOS :/
If there is something I can do to speed this up let me know!

AdriVanHoudt:
I tested it and this fix doesn't seem to do it.
If instead of adding the click handler to the document I add it to the .md-scroll-mask, it seems to behave as expected

Splaktar:
It has to go through the presubmit process and so far that hasn't gone through without an issues. There hasn't been time yet to review what needs to change (if anything) to get it through presubmit, but we're not holding `1.1.13` for this as we need to get some regression fixes released ASAP.

Splaktar:
@AdriVanHoudt I took that approach initially with this fix, but it had some unintended, negative side effects. 

Can you please provide reproduction steps and a CodePen demo for the case where this fix isn't working for you? Also please provide the device and OS version.

AdriVanHoudt:
I'll see if I can get a minimal case to work/break, I tested it directly in our application.
The device I tested on is an iPhone 6 with iOS 12.1.4

What is the easiest way to test this branch in a CodePen?

Splaktar:
Here's [the issue](https://github.com/angular/material/issues/9581#issuecomment-457780774) that I hit when trying to click the `scroll-mask`.

Uploading your `angular-material.js` to Google Drive and then getting a sharable link with `view` permission can make the file available for use in CodePen via a URL like `https://drive.google.com/uc?export=download&amp;id=1NRoxKfRH_ka9I063vGIwXi_RvxKV7F__` where the part after `id=` is the same as that of the link from the Drive UI (`https://drive.google.com/open?id=1NRoxKfRH_ka9I063vGIwXi_RvxKV7F__`). You just need to use the `uc?export=download&amp;` API to make it work with CodePen.

Here's [a CodePen](https://codepen.io/Splaktar/pen/XOwQMz) from the issue using a build from this PR.

AdriVanHoudt:
That's a nice way to host code :D

To clarify: I tested this with md-chips with the md-autocomplete as a custom input. I didn't think about mentioning this earlier sorry.
I found out that the demo on https://material.angularjs.org/latest/demo/chips for the md-chips with the custom md-autocomplete input has the same problem so I forked that and added this version to it.
https://codepen.io/anon/pen/pGXpYw
It still has some weird behaviour, not sure what is happening. Also Safari devtools + iPhone is not the greatest debugging exp

Splaktar:
I tested this fix on iOS with the chips autocomplete in the docs and it fixed the issue. In 1.1.19, tapping away to close the panel doesn't work. With the changes in this branch, it seemed to work fine.

AdriVanHoudt:
Thanks for this 🔥

Splaktar:
It looks like there is an issue with this change. I'm going to be investigating it to see if we can solve it rather than reverting this PR.

Splaktar:
I can reproduce the issue in https://material.angularjs.org/HEAD/demo/autocomplete on macOS and Chrome 75.

Here's the behavior, after opening an autocomplete options panel by clicking the input with the mouse, when the mouse first moves over the edge of the panel, the panel flickers (hides and then reappears).

Splaktar:
PR https://github.com/angular/material/pull/11757 posted to address the regression.

PR #11626, opened 2019-01-31
Revert "fix(autocomplete): show dropdown on top only when there is room"

mmalerba:
This reverts commit a37f7ccecf225bcb3cf96774a7a7b97091cbab62.

Creating this PR to investigate failure in google3 possibly caused by the reverted change


Splaktar:
OK, please let me know if there is anything that I can do to help.

mmalerba:
If it does turn out to be the cause I'll merge this and we can work on fixing the issue in a follow-up

Splaktar:
Related to https://github.com/angular/material/pull/11575.

PR #11604, opened 2019-01-22
fix(icon): restore SVG Element correctly in IE11

rreitmann:
the fix for IE11 introduced by #11545 removes the outer SVG element tag.
correctly restore the element by rebuilding it.

fixes #11603

## PR Checklist
Please check that your PR fulfills the following requirements:
- [x] The commit message follows [our guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format)
- [x] Tests for the changes have been added or this is not a bug fix / enhancement
- [x] Docs have been added, updated, or were not required

## PR Type
What kind of change does this PR introduce?

```
[x] Bugfix
[ ] Enhancement
[ ] Documentation content changes
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: #11603 

## Does this PR introduce a breaking change?
```
[ ] Yes
[x] No
```

PR #11635, opened 2019-02-09
fix(icon): SVG elements are not loaded from the cache properly on IE11

### LINKED ISSUE #11603: icon: SVG icons disappear in IE11 when reused from internal cache
Since commit https://github.com/angular/material/commit/47527f245acc8382a69a42a07989fd5f646e9744 SVG icons which are reused from the cache (md-icons internal one) are not displayed anymore in IE11 because `transformClone` removes the `<svg>`-element and adds only the children to the dom (see https://github.com/angular/material/commit/47527f245acc8382a69a42a07989fd5f646e9744#diff-8a475bebd83d2830320cb84bceeade2cR505)

Just go to the docs (https://material.angularjs.org/latest/demo/icon) with IE11. Initially the SVG on the "Edit on Codepen"- and "View Source"-buttons will be visible. Click on a different demo and they will be invisible. 

- AngularJS: 1.7.6 
- AngularJS Material: 1.1.11 and 1.1.12
- OS: Windows
- Browsers: IE11

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

SVG elements are not loaded from the cache properly on IE11. The outer `<svg>` tag is lost.

Issue Number: 
Fixes #11603. Relates to #11545. Closes #11604.

## What is the new behavior?
SVG elements are loaded from the cache properly on IE11 including the `<svg>` tag and the element's attributes.

```
[ ] Yes
[x] No
```

## Other information
We can't just use `outerHTML` here either due to that [not being supported for SVG elements in IE](https://stackoverflow.com/questions/29888050/outerhtml-is-not-working-in-ie) as well.

### DIFF
--- a/src/core/util/util.js
+++ b/src/core/util/util.js
@@ -864,21 +864,38 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
     },
 
     /**
-     * Function to get innerHTML of SVG and Symbol elements in IE11
+     * Gets the inner HTML content of the given HTMLElement.
+     * Only intended for use with SVG or Symbol elements in IE11.
      * @param {Element} element
-     * @returns {string} the innerHTML of the element passed in
+     * @returns {string} the inner HTML of the element passed in
      */
     getInnerHTML: function(element) {
+      // For SVG or Symbol elements, innerHTML returns `undefined` in IE.
+      // Reference: https://stackoverflow.com/q/28129956/633107
+      // The XMLSerializer API is supported on IE11 and is the recommended workaround.
       var serializer = new XMLSerializer();
 
       return Array.prototype.map.call(element.childNodes, function (child) {
         return serializer.serializeToString(child);
       }).join('');
+    },
+
+    /**
+     * Gets the outer HTML content of the given HTMLElement.
+     * Only intended for use with SVG or Symbol elements in IE11.
+     * @param {Element} element
+     * @returns {string} the outer HTML of the element passed in
+     */
+    getOuterHTML: function(element) {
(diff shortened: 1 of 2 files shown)


src/core/util/util.js
```diff
@@ -874,6 +874,16 @@ function UtilFactory($document, $timeout, $compile, $rootScope, $$mdAnimate, $in
...
         return serializer.serializeToString(child);
       }).join('');
+    },
+
+    /**
+     * Function to get outerHTML of SVG and Symbol elements in IE11
+     * @param {Element} element
+     * @returns {string} the outerHTML of the element passed in
+     */
+    getOuterHTML: function(element) {
```
jelbourn: Why not just use the `outerHTML` property?
https://developer.mozilla.org/en-US/docs/Web/API/Element/outerHTML

Splaktar: It's not supported by IE11 for SVG elements. It returns `undefined`.

Splaktar: It's not mentioned in https://caniuse.com/#feat=xml-serializer, but I've left feedback that it should be added to the Known Issues.

Splaktar: Reference: https://stackoverflow.com/questions/29888050/outerhtml-is-not-working-in-ie.
Also manually tested.

jelbourn: Got it. I would just add an line comment here that adds all of that context to the code. The function description should just be `Gets the outer HTML content of the given HTMLElement`

PR #11640, opened 2019-02-15
fix(checkbox): enter submits form when submit button is disabled

### LINKED ISSUE #11639: checkbox: should fire click event on submit button if one exists
## Bug

1. Focus the native checkbox
1. Press enter and verify that the form was not submitted (validation text doesn't display)
1. Focus the `md-checkbox`
1. Press enter and see that the form was submitted (validation text displays)

Examples of the native behavior
- https://jsbin.com/matuvoj/1/edit?html,js,output
- https://jsbin.com/leforagipo/1/edit?html,js,output

When the `ENTER` key is pressed, checkbox should search the form for all submit `button`s and `input`s and send a `MouseClick` event to the first submit button that is not disabled.

Verify that this [CodePen](https://codepen.io/Splaktar/pen/ZVjWRJ?editors=1010#0) still works.

Checkbox emits a native `submit` event when the `ENTER` key is pressed.

Checkbox keyboard behavior should match the native behavior on Chrome.

- AngularJS: 1.7.6
- AngularJS Material: 1.1.13
- OS: macOS
- Browsers: Chrome
 
@codymikol heads up that this is related to PR #11584 and issue https://github.com/angular/material/issues/11583. I will try to make a fix for this today.

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Checkbox emits a native `submit` event when the `ENTER` key is pressed.

Issue Number: 
Fixes #11639

## What is the new behavior?
- match the behavior of the native input type checkbox
- When the `ENTER` key is pressed, `md-checkbox` searches the `form` for all submit `button`s and `input`s and send a `MouseClick` event to the first one that is enabled.

```
[ ] Yes
[x] No
```

## Other information
I created tests for all of the following use cases
- [x] if there is an enabled `input type="submit"` with a `ng-click` before a `button type="submit"`, verify that the click is fired on the `input`
- [x] if there is an enabled `button type="submit"` with a `ng-click` before a `input type="submit"`, verify that the click is fired on the `button`
- [x] if the form has `ng-submit` and an enabled `input type="submit"`, verify that the form gets submitted
- [x] if the form has `ng-submit` and an enabled `button type="submit"`, verify that the form gets submitted
- [x] if the form has an enabled `input type="submit"` with `ng-click`, verify that the form gets submitted
- [x] if the form has an enabled `button type="submit"` with `ng-click`, verify that the form gets submitted
- [x] Variations where `input type="submit"` and `button type="submit"` are disabled but a second one is enabled and gets the click event
- [x] negative tests where the `input type="submit"` and `button type="submit"` are disabled, verify that the form does not get submitted

### DIFF
--- a/src/components/checkbox/checkbox.js
+++ b/src/components/checkbox/checkbox.js
@@ -169,8 +169,13 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
         }
       }
 
+      /**
+       * @param {KeyboardEvent} ev 'keypress' event to handle
+       */
       function keypressHandler(ev) {
         var keyCode = ev.which || ev.keyCode;
+        var submit, form;
+
         ev.preventDefault();
         switch (keyCode) {
           case $mdConstant.KEY_CODE.SPACE:
@@ -178,10 +183,17 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
             listener(ev);
             break;
           case $mdConstant.KEY_CODE.ENTER:
-            var form = $mdUtil.getClosest(ev.target, 'form');
-            // We have to use a native event here as the form directive does not use jqlite
+            // Match the behavior of the native <input type="checkbox">.
+            // When the enter key is pressed while focusing a native checkbox inside a form,
+            // the browser will trigger a `click` on the first non-disabled submit button/input
+            // in the form. Note that this is different from text inputs, which
+            // will directly submit the form without needing a submit button/input to be present.
+            form = $mdUtil.getClosest(ev.target, 'form');
             if (form) {
-              form.dispatchEvent(new Event('submit'));
+              submit = form.querySelector('button[type="submit"]:enabled, input[type="submit"]:enabled');
+              if (submit) {
+                submit.click();
+              }
             }
             break;
         }
--- a/src/components/checkbox/checkbox.spec.js
+++ b/src/components/checkbox/checkbox.spec.js
@@ -2,7 +2,7 @@
 describe('mdCheckbox', function() {
   var CHECKED_CSS = 'md-checked';
   var INDETERMINATE_CSS = 'md-indeterminate';
-  var $compile, $log, pageScope, $mdConstant;
+  var $compile, $log, pageScope, $mdConstant, $window;
 
   beforeEach(module('ngAria', 'material.components.checkbox'));
 
@@ -10,9 +10,10 @@ describe('mdCheckbox', function() {
     $compile = $injector.get('$compile');
     $log = $injector.get('$log');
     $mdConstant = $injector.get('$mdConstant');
+    $window = $injector.get('$window');
 
     var $rootScope = $injector.get('$rootScope');
-    pageScope = $rootScope.$new();
+    pageScope = $rootScope.$new(false);
   }));
 
   function compileAndLink(template, opt_scope) {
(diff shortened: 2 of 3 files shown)


src/components/checkbox/checkbox.js
```diff
@@ -169,19 +169,29 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
-            // We have to use a native event here as the form directive does not use jqlite
+            // Match the behavior of the native input type="checkbox".
+            // It's important that MouseEvent is used here versus Event in order to match the native
+            // behavior and properly trigger ng-submit.
+            form = $mdUtil.getClosest(ev.target, 'form');
             if (form) {
-              form.dispatchEvent(new Event('submit'));
+              submit = form.querySelector('button[type="submit"]:enabled, input[type="submit"]:enabled');
+              if (submit) {
+                submit.dispatchEvent(new MouseEvent('click'));
```
jelbourn: Why not just call `submit.click()`?

Splaktar: Good call. Done.

src/components/checkbox/checkbox.js
```diff
@@ -169,19 +169,29 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
+            // Match the behavior of the native input type="checkbox".
+            // It's important that MouseEvent is used here versus Event in order to match the native
+            // behavior and properly trigger ng-submit.
+            form = $mdUtil.getClosest(ev.target, 'form');
             if (form) {
-              form.dispatchEvent(new Event('submit'));
+              submit = form.querySelector('button[type="submit"]:enabled, input[type="submit"]:enabled');
+              if (submit) {
+                submit.click();
+              }
```
codymikol: I would imagine this would work similarly to inputs.
If a submit button is found, fire a click event onto it, otherwise just submit the form.

This behavior is currently present on inputs: [Code Pen](https://codepen.io/codymikol/pen/OdaQYy?editors=1010#0)

I leverage this kind of behavior in my application. In some circumstances I have an outer controller that doesn't know about it's contents, but has the ability to fire submit events to child forms via a directive. I also want the user to be able to use the keyboard inside of the form to submit.

There is an attribute "form" which allows you to specify the form for a submit button outside of a form, but it is not supported in IE 11.

src/components/checkbox/checkbox.spec.js
```diff
@@ -298,14 +298,14 @@ describe('mdCheckbox', function() {
...
+    it('should not submit a parent form when ENTER is pressed', function () {
       var form = compileAndLink('<form><md-checkbox ng-checked="value"></md-checkbox></form>');
       angular.element(form[0].getElementsByTagName('md-checkbox')[0]).triggerHandler({
         type: 'keypress',
         keyCode: $mdConstant.KEY_CODE.ENTER
       });
       pageScope.$apply();
-      expect(form[0].classList.contains('ng-submitted')).toBe(true);
+      expect(form[0].classList.contains('ng-submitted')).toBe(false);
     });
```
codymikol: You might want to add a test to prove that a checkbox is submitted when expected as well.

src/components/checkbox/checkbox.js
codymikol: I do see that there is definitely native inconsistency with this. Inputs will submit regardless of submit button presence, but checkboxes require one. However, I don't see any drawback to allowing this kind of behavior in absence of a submit button.

Splaktar: Well the primary thing that we have to fix is that `Enter` on checkboxes should not submit the form if the submit button/input is disabled.

I forked your [CodePen](https://codepen.io/Splaktar/pen/NooaYZ?editors=1010#0) and added a native `input`. It submits on `Enter` without a submit button/input *only* if it's the only input in the form. If there are 2 `inputs`, the `Enter` key does not submit without a submit button/input.

Some of this odd behavior is captured in the [AngularJS Form Directive docs](https://docs.angularjs.org/api/ng/directive/form#submitting-a-form-and-preventing-the-default-action).

I suppose we could theoretically submit the form on `Enter` for a checkbox if no submit button or input is found, but is inconsistent with [native behavior](https://jsbin.com/lajutej/edit?html,js,output). So it's likely not a good idea. If you comment out the input in this JSBin, `Enter` on a lone `checkbox` does not submit the form.

src/components/checkbox/checkbox.js
```diff
@@ -169,19 +169,29 @@ function MdCheckboxDirective(inputDirective, $mdAria, $mdConstant, $mdTheming, $
...
           case $mdConstant.KEY_CODE.SPACE:
             element.addClass('md-focused');
             listener(ev);
             break;
           case $mdConstant.KEY_CODE.ENTER:
-            var form = $mdUtil.getClosest(ev.target, 'form');
-            // We have to use a native event here as the form directive does not use jqlite
+            // Match the behavior of the native input type="checkbox".
+            // It's important that MouseEvent is used here versus Event in order to match the native
+            // behavior and properly trigger ng-submit.
```
jelbourn: I think the `Event` part of this comment is no longer necessary. I would, however, expand the comment like:
```ts
// Match the behavior of the native <input type="checkbox">.
// When the enter key is pressed while focusing a native checkbox
// inside a form, the browser will trigger a `click` on the first non-disabled
// submit button in the form. Note that this is different from text inputs, which
// will directly submit the form without needing a submit button to be present.
```

src/components/checkbox/checkbox.spec.js
```diff
@@ -298,15 +299,137 @@ describe('mdCheckbox', function() {
       expect(isChecked(checkbox)).toBe(false);
     });
 
-    it('should submit a parent form when ENTER is pressed', function () {
-      var form = compileAndLink('<form><md-checkbox ng-checked="value"></md-checkbox></form>');
-      angular.element(form[0].getElementsByTagName('md-checkbox')[0]).triggerHandler({
-        type: 'keypress',
-        keyCode: $mdConstant.KEY_CODE.ENTER
+    it('should not submit a parent form when ENTER is pressed and there is no submit button/input',
```
jelbourn: There should also be a case that tests that the checkbox triggers the *first* non-disabled submit button when there are multiple.

Splaktar: Added.

src/components/checkbox/checkbox.js
Splaktar: Good catch, updated.

Issue #11493, opened 2018-10-23
dialog: memory leak when injecting $mdDialog


kimtuck:
## Bug:

https://codepen.io/kimtuck/pen/yRRbjb

#### Detailed Reproduction Steps: 
 
1. Run the contents of the pen using karma.  Monitor the memory usage of chrome while the tests are running.  You will see the memory creep up dramatically, until around 2.1GB chrome crashes.

### What is the expected behavior?
Memory remains more or less constant, because all we're doing here is injecting `$mdDialog`, which is a service (a singleton).

### What is the current behavior?
Memory is dramatically consumed

### What is the use-case or motivation for changing an existing behavior?
Memory leak will lead to application crashes or poor performance.  In a set of unit tests that inject `$mdDialog` in many places (as our application does) chrome crashes during the unit test run, leading to a failed unit test.

### Which versions of AngularJS, Material, OS, and browsers are affected?
See the package.json in the codepen.  Relevant lines from package.json:
    "angular": "^1.7.2",
    "angular-animate": "^1.7.2",
    "angular-aria": "^1.7.2",
    "angular-messages": "^1.7.2",
    "angular-material": "^1.1.10",
- OS: windows 10
- Browsers: Chrome
 
### Is there anything else we should know? Stack Traces, Screenshots, etc.

This is the (partial) code for MdDialogProvider.  If you remove the `.setDefaults()` and the `.addPresets()` then there is no memory leak.  If you include the `.setDefaults()` and remove the `.addPresets()` then the memory leak occurs.

So, the memory leak is in the `.setDefaults()` call.  

Note that on each test invocation, the MdDialogProvider method executes; and thus the `setDefaults()` method runs on each test invocation.
```js
function MdDialogProvider($$interimElementProvider) {
  // Elements to capture and redirect focus when the user presses tab at the dialog boundary.
  MdDialogController['$inject'] = ["$mdDialog", "$mdConstant"];
  dialogDefaultOptions['$inject'] = ["$mdDialog", "$mdAria", "$mdUtil", "$mdConstant", "$animate", "$document", "$window", "$rootElement", "$log", "$injector", "$mdTheming", "$interpolate", "$mdInteraction"];
  var topFocusTrap, bottomFocusTrap;

  return $$interimElementProvider('$mdDialog')
    .setDefaults({
      methods: ['disableParentScroll', 'hasBackdrop', 'clickOutsideToClose', 'escapeToClose',
[...]

PR #11641, opened 2019-02-19
fix($mdInteraction): clean up events on $rootScope destroy

codymikol:
this prevents unit tests from leaking memory

Fixes: #11493

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Currently the $mdInteraction service registers when invoked. This causes a memory leak in unit tests because nothing is in place to clean up the events.

Issue Number: #11493

## What is the new behavior?

Now on the $destroy event of rootScope the events registered will be unregistered.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/core/services/interaction/interaction.js
+++ b/src/core/services/interaction/interaction.js
@@ -33,15 +33,20 @@ angular
  * </hljs>
  *
  */
-function MdInteractionService($timeout, $mdUtil) {
+function MdInteractionService($timeout, $mdUtil, $rootScope) {
   this.$timeout = $timeout;
   this.$mdUtil = $mdUtil;
+  this.$rootScope = $rootScope;
 
+  // IE browsers can also trigger pointer events, which also leads to an interaction.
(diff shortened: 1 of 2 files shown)


codymikol:
I have another version that squishes the init and deregister methods into one that is passed 'on' or 'off', but thought this might be overly clever and less easily understood. It does have the have the benefit of having one place where events are manipulated so you don't have to worry about keeping both the init and deregister methods in sync. Let me know if that is preferred. 🐔

Splaktar:
In the commit message, can you please change `Fixes: #11493` to `Fixes #11493`?

codymikol:
> In the commit message, can you please change `Fixes: #11493` to `Fixes #11493`?

Changed 🐔 

> This is great! Thank you!
>
> Is there any way to verify this fix in a unit test? Maybe destroy the the rootScope and then check if the handlers are still there?

I added unit tests that see if the lastInteractionType is modified after an event is fired. 🐔

PR #11652, opened 2019-02-23
WIP: fix(toast): sanitize auto wrapped custom toast templates

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

There was an issue filed in g3 about using [innerHTML here](https://github.com/angular/material/blob/v1.1.13/src/components/toast/toast.js#L437) in `md-toast` for custom toasts.

Issue Number: 
Related to #6494. Related to #6259.

## What is the new behavior?
Protect against a possible XSS vector by sanitizing all elements inside of the template's outer `<md-toast><sanitize everything here></md-toast>` element. This could be multiple DOM elements, comments, etc.

```
[x] Yes
[ ] No
```
This may require `ngSanitize` in some apps that previously didn't use it. In those cases, the apps would break with an `$sce` exception.

## Other information
This uses a similar approach to [md-select's approach](https://github.com/angular/material/blob/ecf1705b948ece78531a675f8960f90ec4fda163/src/components/select/select.js#L338-L341) to sanitizing text that can contain HTML.

### DIFF
--- a/src/components/toast/toast.js
+++ b/src/components/toast/toast.js
@@ -416,7 +416,7 @@ function MdToastProvider($$interimElementProvider) {
   }
 
   /* @ngInject */
-  function toastDefaultOptions($animate, $mdToast, $mdUtil, $mdMedia, $document) {
+  function toastDefaultOptions($animate, $mdToast, $mdUtil, $mdMedia, $document, $sce) {
     var SWIPE_EVENTS = '$md.swipeleft $md.swiperight $md.swipeup $md.swipedown';
     return {
       onShow: onShow,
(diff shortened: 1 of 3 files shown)


src/components/toast/toast.js
```diff
@@ -450,15 +451,41 @@ function MdToastProvider($$interimElementProvider) {
...
-          // We have to return the innerHTMl, because we do not want to have the `md-template`
+          // We have to return the innerHTML, because we do not want to have the `md-template`
           // element to be the root element of our interimElement.
           return templateRoot.innerHTML;
+        } else if (template) {
+          // We can't sanitize the contents of these templates as it would affect our simple
+          // templates as well as custom templates. It would remove attributes like
+          // aria-live and aria-relevant. It would also remove md-buttons.
+          // TODO Should we go in and call trustAsHtml on those templates farther up the chain?
+          return template;
```
Splaktar: @mmalerba there's another edge case here where a custom template may specify `autoWrap = false`. I tested and verified that a `<script>` could be injected here (but I wasn't able to get it to execute). 

I just wanted to confirm that you thought it was worth while to go and try to trust the templates passed in here from `$mdToast.simple()` and `$mdToast.showSimple()`.

bersLucas: Browser will (hopefully) stop injected stuff from being executed, but adding stuff like `` will inject and execute

PR #11671, opened 2019-03-10
docs(scroll): scroll to top on route change

### LINKED ISSUE #11669: docs: content pane should scroll to top when left menu item is clicked
This is a bug on the docs website.
https://material.angularjs.org

There is no code pen, as this issue is reproducible in the docs website.

1.  Go to https://material.angularjs.org
2. Open the Theming panel.
3. Click on "Introduction and Terms"
4. Scroll content to bottom on the right panel.
5. Click on "Declarative Syntax" on the sidebar menu.
6. The scrollbar in the content panel stays at the bottom.

The scroll bar should reset to 0.

The scroll bar stays at the same place as it was before selecting the new menu item

It is frustrating to scroll to top every time I select a new menu item.

- AngularJS: The docs seem to use AngularJS v1.7.6
- AngularJS Material: Latest docs: https://material.angularjs.org/latest/Theming/02_declarative_syntax
- OS: Windows
- Browsers: Chrome 72

gaurav5430:
scroll to top on every route/location change. Previously, for some of
the route changes when clicking on the side bar menu, the scroll stayed in place.

Fixes #11669

## PR Checklist

## PR Type

- [ ] Bugfix
- [ ] Enhancement
- [ ] Documentation content changes
- [ ] Code style update (formatting, local variables)
- [ ] Refactoring (no functional changes, no api changes)
- [ ] Build related changes
- [ ] CI related changes
- [ ] Infrastructure changes
- [x] Other... Please describe:

Bugfix/Enhancement for docs.

Issue Number: #11669 

## What is the new behavior?
Content pane scrolls to top on clicking a link from sidebar menu item

- [ ] Yes
- [x] No

### DIFF
--- a/docs/app/js/app.js
+++ b/docs/app/js/app.js
@@ -491,7 +491,7 @@ function(SERVICES, COMPONENTS, DEMOS, PAGES, $location, $rootScope, $http, $wind
   }
 }])
 
-.directive('menuLink', function() {
+.directive('menuLink', ['scrollCache', function(scrollCache) {
   return {
     scope: {
       section: '='
@@ -508,10 +508,12 @@ function(SERVICES, COMPONENTS, DEMOS, PAGES, $location, $rootScope, $http, $wind
         // set flag to be used later when
         // $locationChangeSuccess calls openPage()
         controller.autoFocusContent = true;
+        // set flag to be used later when $routeChangeStart saves scroll position
+        scrollCache.linkClicked = true;
       };
     }
   };
-})
+}])
 
 .directive('menuToggle', ['$mdUtil', '$animateCss', '$$rAF', function($mdUtil, $animateCss, $$rAF) {
   return {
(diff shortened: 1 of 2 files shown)


gaurav5430:
ok, i will check that out.

Are you suggesting that when the user clicks on browser back button, the page should scroll to where the user left it ?

1. Reading Button docs,
2. scroll to bottom,
3. Click on Checkbox menu-link

this should scroll to top,

4. Now, click browser back button.
5. Button page should load and it should scroll to bottom ?

gaurav5430:
I think we can work with a similar kind of approach as mentioned in your linked stackoverflow answer.
we would just not base it on templateUrl, but the route + templateUrl, as we do use the same template for different pages, 

One problem i can see with this is when we press back, and scroll to the previous value, some of the content might still be loading (or the browser may be parsing some template to load it on the page) and would change the height of the content panel, so our solution would not take the user to their previously scrolled position (but may be somewhere close by.

Splaktar:
> Are you suggesting that when the user clicks on browser back button, the page should scroll to where the user left it ?

Yes

Splaktar:
> so our solution would not take the user to their previously scrolled position (but maybe somewhere close by).

This seems to be similar to what we have today. If the page that it is going back to has a different height than the current page, it seems to not take you back to the same spot. Let's give it a shot and then when you have that pushed up, I can test and tweak it a little to see if we can work around that issue.

Thank you for investigating.

gaurav5430:
I was able to tweak the directive for our use case.
Need more inputs about the functionality.

What do you want to happen when the user clicks on a menu item in the sidebar ?
scroll to zero, or scroll to previous position ?

ex: 
Click on Button, - scroll 0
Scroll to x 
Click on Chips - scroll 0

Click on Button - scroll 0 or x ?

would you expect the scroll to go back to 0, or stay at x

gaurav5430:
( I guess i could use $location.path() instead of getting the path from $$route )

Splaktar:
> What do you want to happen when the user clicks on a menu item in the sidebar ?

Scroll to zero (top)

docs/app/js/app.js
```diff
@@ -881,4 +882,37 @@ function($rootScope, $scope, component, demos, $templateRequest) {
...
+ * This helps in restoring the scroll when the user visits the same page again
+ */
+.directive('cacheScrollPosition', ['$route', '$mdUtil', '$timeout', '$location', '$anchorScroll',
+function($route, $mdUtil, $timeout, $location, $anchorScroll) {
+  var scrollPosCache = {};
+  var mainContentArea = document.querySelector("[role='main']");
+  var scrollContentEl = mainContentArea.querySelector('md-content[md-scroll-y]');
+  return function(scope) {
+
+      scope.$on('$routeChangeStart', function() {
```
Splaktar: Please reformat this new block to use 2 space indentation.

Splaktar: You probably want to run `npm run lint` to check for any other issues as well if you don't have ESLint enabled in your IDE.

gaurav5430: Sure, I do have eslint enabled in my editor and it warns me on other rules, but not in space/tabs. Anyway would run limiting before commit.

gaurav5430:
Consider this scenario:

1. You are on button docs.
2. You scroll to X
3. You click on Toolbar docs link.
Scrolls to zero.
4. You click on button docs link
Scrolls to zero 
5. Go back twice on your browser.
6. You are now on button docs, where would you expect the scroll to be?

If you expect it to be x, we might need to look for  a more sophisticated solution

Splaktar:
I think that scroll being at 0 (top) is fine in that case. We don't need a stack of history that goes back 10x or something for each route.

gaurav5430:
added the discussed functionality.
We explicitly set a flag when a menu-link is clicked, and use that to set scroll to zero for that route.

This means, scroll positions are cached by default, but are overridden if menu-link is clicked.
So, the scroll position on back button is also cached.

We do have a problem though, we only check for explicit menu-link clicks, but we do not override the scroll positions when any other link is clicked, for e.g links in Layout Container and Layout Options.
So those clicks would still go to the cached scroll positions.

I am conflicted about whether as a user I would like to scroll to top in that case as well.
But if we do, then with our current approach, it would mean we would need to wrap all our internal links in a directive and set the same flag as menu-link to override the cached value.

Splaktar:
Those link clicks in something like Layout Container would still go to the top if there was no cached scroll position?

If that's the case, then I think that we're in good shape.

I'll try to take a longer look at the latest changes and do some testing myself next week. Thank you for your work on this!

docs/app/js/app.js
```diff
@@ -881,4 +883,59 @@ function($rootScope, $scope, component, demos, $templateRequest) {
...
+  }
+
+  return {
+    getScroll: getScroll,
+    setScroll: setScroll,
+    linkClicked: linkClicked
+  };
+})
+/** This directive caches the scroll position for each route + templateUrl combination
+ * This helps in restoring the scroll when the user visits the same page again
```
Splaktar: Let's make this a little more accurate.

"This helps in restoring the scroll when the user uses back or forward navigation to return to a page."

docs/app/js/app.js
```diff
@@ -881,4 +883,59 @@ function($rootScope, $scope, component, demos, $templateRequest) {
...
+ */
+.directive('cacheScrollPosition', ['$route', '$mdUtil', '$timeout', '$location', '$anchorScroll', 'scrollCache',
+function($route, $mdUtil, $timeout, $location, $anchorScroll, scrollCache) {
+  var mainContentArea = document.querySelector("[role='main']");
+  var scrollContentEl = mainContentArea.querySelector('md-content[md-scroll-y]');
+  return function(scope) {
+    scope.$on('$routeChangeStart', function(event, next, current) {
+        // store scroll position for the current path + template
+        if ($route.current) {
+          scrollCache.setScroll(current.loadedTemplateUrl+":"+current.$$route.originalPath, scrollContentEl.scrollTop);
```
Splaktar: Please wrap lines at 100 characters.

gaurav5430:
Great!
I will make the formatting changes (will copy/paste your formatted code and see if there are any other formatting issues)

Do I need to squash the commits and amend the message as well ?
Do you want a single commit for the functionality ?

Splaktar:
Yes, please squash your commits. In the future, you can use amend and force push to avoid the need to squash.

gaurav5430:
Squashed commits by rebasing, Force pushed.
Now you should see a single commit with the latest date.

PR #11673, opened 2019-03-12
fix(colors): handle null or undefined in md-colors

### LINKED ISSUE #11672: colors: add support for null or undefined values
## Bug or maybe enhancement request:
`md-colors` directive throws error when property values evaluate to null or undefined.
Handling null or undefined is particularly useful when expression is controlled by some nullable user input.

Currently it not logs an error in the console, but also fails to correctly clean previous styles when going from some value to undefined or null.

1. Select a color from color selection control
2. Select **none**

`md-colors` should gracefully handle `null` or `undefined`. It should not log error in the console, and also it should clean previous styles when going from a valid color to `null`/`undefined`

It logs lots of errors in the console and doesn't clean previous styles properly.

All use cases when values are not static strings and are dynamically evaluated and can may be nullable.

- AngularJS:  any (not relevant)
- AngularJS Material: 1.1.13 (latest at the time)
- OS: any
- Browsers: any
 
No

alirezamirian:
If md-colors expression evaluates to an object like `{ backgroundColor: undefined }`,
md-colors should handle it and remove backgroundColor css property.

fixes #11672 

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Described in #11672 

Issue Number: #11672 

## What is the new behavior?
Described in #11672 

```
[ ] Yes
[ ] No
[x] Not sure, but probably no.
```

## Other information
Nothing

### DIFF
--- a/src/components/colors/colors.js
+++ b/src/components/colors/colors.js
@@ -121,6 +121,9 @@
      * @returns rgba color string
      */
     function parseColor(color, contrast) {
+      if(!color) {
+        return '';
+      }
       contrast = contrast || false;
       var rgbValues = $mdTheming.PALETTES[color.palette][color.hue];
 
@@ -171,6 +174,9 @@
      * For the evaluated expression, extract the color parts into a hash map
      */
     function extractColorOptions(expression) {
+      if(!expression) {
+        return undefined;
+      }
       var parts = expression.split('-');
       var hasTheme = angular.isDefined($mdTheming.THEMES[parts[0]]);
       var theme = hasTheme ? parts.splice(0, 1)[0] : $mdTheming.defaultTheme();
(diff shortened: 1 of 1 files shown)


src/components/colors/colors.js
```diff
@@ -171,6 +174,9 @@
      * For the evaluated expression, extract the color parts into a hash map
      */
     function extractColorOptions(expression) {
+      if(!expression) {
+        return undefined;
```
Splaktar: If this is going to start returning `undefined`, then we need to make sure that callers like `interpolateColors()` on lines 154 and 156 don't call `parseColor(undefined)`.

PR #11681, opened 2019-03-15
fix(panel): append panel animation transforms after existing transforms

Noglen:
append the transform for panel animations after any existing transforms.  Previously, the animation
transform was appened first, resulting in the destination location being incorrect.

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

When a panel has a centered location (or other transform applied) combined with an OpenFrom/CloseTo, the panel animates to/from the wrong location

Issue Number:
N/A

## What is the new behavior?

The panel animates to/from the correct location

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/panel/panel.js
+++ b/src/components/panel/panel.js
@@ -3371,7 +3371,7 @@ MdPanelAnimation.prototype.animateOpen = function(panelEl) {
 
       var openScale = animator.calculateZoomToOrigin(
               panelEl, this._openFrom) || '';
-      openFrom = animator.toTransformCss(openScale + ' ' + panelTransform);
+      openFrom = animator.toTransformCss(panelTransform + ' ' + openScale);
       break;
 
     case MdPanelAnimation.animation.FADE:
@@ -3436,7 +3436,7 @@ MdPanelAnimation.prototype.animateClose = function(panelEl) {
 
       var closeScale = animator.calculateZoomToOrigin(
               panelEl, this._closeTo) || '';
-      closeTo = animator.toTransformCss(closeScale + ' ' + panelTransform);
+      closeTo = animator.toTransformCss(panelTransform + ' ' + closeScale);
       break;
 
     case MdPanelAnimation.animation.FADE:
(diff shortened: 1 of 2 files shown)


Splaktar:
Since there is no associated issue and thus no associated demo, can you please provide a CodePen demo that reproduces the issue being fixed here? Or point to a demo on the docs site that reproduces the issue?

Noglen:
Here's a CodePen that shows the issue.  Its the panel animations from the doco - but with the panel position set to .center() instead of .right().top();

[https://codepen.io/noglen/pen/qwBMOx](https://codepen.io/noglen/pen/qwBMOx)

Splaktar:
For future testing reference
- [My updated CodePen](https://codepen.io/Splaktar/pen/BEaOOo) that shows the problem
- [My CodePen with the fix from this branch](https://codepen.io/Splaktar/pen/ZZEMma?editors=1010)

Splaktar:
@Noglen I just wanted to remind you that I'm still waiting on a response to my previous comment in https://github.com/angular/material/pull/11681#pullrequestreview-220922781.

Noglen:
> @Noglen I just wanted to remind you that I'm still waiting on a response to my previous comment in [#11681 (review)](https://github.com/angular/material/pull/11681#pullrequestreview-220922781).

Sorry about that.

I've had another look over it gain and can't repo the problem with `slide`, so i'm gussing it was something else I had in my project that was causing the issue.

Do you think I should revert back the change for `slide`, or leave the change so that it is consistant with `scale`?

I'll try and get some unit tests added in today aswell.

Splaktar:
Yeah, let's revert the changes to slide until we can isolate a reproduction. That should reduce the risk that this causes any issues or has problems with the presubmit process.

OK, thank you for looking into adding the unit tests!

Noglen:
reverted the change to `slide`

unit tests were a bit tricky since it's the browser's rending of the CSS that causes the problem with the position (since the order of operations in transform actually matters), not the actual position calc in code  - so ended up just checking that the scale transform is added after the existing transform

PR #11690, opened 2019-03-26
docs(migration): add reference document for migration

Splaktar:
## PR Checklist

## PR Type

```
[ ] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Reference documentation for migration to Angular Material, CDK, and Flex Layout does not exist.

Issue Number: 
N/A

## What is the new behavior?
- Reference documentation for migration to Angular Material, CDK, and Flex Layout exists
- highlight.js updated from 2014 version to latest 2019 version with support for Sass (`scss`) and TypeScript

```
[ ] Yes
[x] No
```

## Other information
Captura de Pantalla 2019-04-16 a la(s) 19 39 16

Preview here: https://angularjsmaterial.dev/migration


docs/content/migration.md
```diff
@@ -0,0 +1,535 @@
...
+The Angular CDK's [Layout][121] module provides:
+- A set of `Breakpoints` that match the [Material Design][122] responsive layout grid breakpoints
+- A `BreakpointObserver` Service that provides
+    - the ability to check media query matching state via `isMatched()`
+    - the ability to subscribe to a stream of events based on changes to an array of media queries
+      that are passed into `observe()`
+- A `MediaMatcher` Service that provides access to the low level, native [`MediaQueryList`][123]
+
+If you are building responsive components or need responsive features in your application with a
+minimal increase in bundle size, then the CDK's Layout module is an option you should consider.
```
jelbourn: I would omit this paragraph here and instead move to the introduction of this section something like
```markdown
The CDK's layout capabilities offer a lightweight alternative (or complement) to larger, full-blown layout systems.
```

docs/content/migration.md
```diff
@@ -0,0 +1,535 @@
...
+
+### Typography
+
+AngularJS Material [typography classes][107] are generally identical to the Angular Material
```
jelbourn: "are generally identical to" -> "align with"

docs/content/migration.md
```diff
@@ -0,0 +1,535 @@
...
+- Utility Sass mixins and functions for use in [styling your custom components][175]
+  
+## Comparison of Features and Migration Tips
+
+Angular Material introduces a number of [new components](#new-components) and features that do not
+exist in AngularJS Material. Angular Material also makes greater use of semantic HTML and native
+elements to improve accessibility.
+
+However, some directives and services have not been implemented in Angular Material. These are
+indicated by an `N/A` in the New Docs column of the feature tables found below.
```
jelbourn: I would omit this paragraph and just write out "Not implemented" for the notes for those components (and probably replace "N/A" with just a dash).

docs/content/migration.md
```diff
@@ -0,0 +1,535 @@
...
+Angular Material introduces a number of [new components](#new-components) and features that do not
+exist in AngularJS Material. Angular Material also makes greater use of semantic HTML and native
+elements to improve accessibility.
+
+However, some directives and services have not been implemented in Angular Material. These are
+indicated by an `N/A` in the New Docs column of the feature tables found below. 
+
+The Notes column, in each row of the tables, includes information like the new name of the feature
+or an alternative approach to implementing the feature. If the notes are blank, then the name of
+the old feature and the new feature are nearly identical (the APIs are likely to be different).
```
jelbourn: I would omit this paragraph and use the header "Special notes" to signal that it's fine for it to be empty

docs/content/migration.md
```diff
@@ -0,0 +1,535 @@
...
+
+### Directives
+Most directives changed to components. These make for straight forward conversions, for instance
```
jelbourn: The fact that everything in AngularJS implements its components as "directives" is just an implementation detail. I would put this more like
```markdown
Most components use the same name between the two libraries, changing the `md-` prefix to `mat-`. For example, `<md-card>` becomes `<mat-card>`.
```

docs/content/migration.md
```diff
@@ -0,0 +1,535 @@
...
+`<md-card>` changed to `<mat-card>`. Some element directives changed to attribute directives, for
+instance `<md-subheader>` changed to `matSubheader`.
+
+Here is a reference table where you can review and compare the AngularJS Material and
+Angular Material documentation. AngularJS Material uses the `md-` and `$md` prefixes while Angular
+Material uses the `mat-`, `mat`, and `Mat` prefixes. The `md-` prefixes have been omitted to make
+this table more readable.
```
jelbourn: I would replace this paragraph with just the header `#### Component comparison reference` and then just tweak the column headers to be ``Old (`md-`)`` and ``New (`mat-`)``

docs/content/migration.md
```diff
@@ -0,0 +1,535 @@
...
+| Directive        |     Docs     | Notes                                                  |
+|------------------|--------------|--------------------------------------------------------|
+| badge            |   [Docs][37] |                                                        |
+| button-toggle    |   [Docs][15] |                                                        |
+| table            |   [Docs][28] | Non-themed CDK and Material versions available.        |
+| drag-drop        |   [Docs][39] | CDK                                                    |
+| expansion-panel  |   [Docs][32] |                                                        |
+| paginator        |   [Docs][29] |                                                        |
+| sort-header      |   [Docs][30] |                                                        |
+| stepper          |   [Docs][33] | Non-themed CDK and Material versions available.        |
```
jelbourn: "Non-themed" -> "Unstyled"

Splaktar: OK, will update. I went with that first, but my spell checker complained.

Splaktar:
Just FYI that I plan to address the review feedback on Monday unless there is an urgent need for this to be wrapped up this week.

docs/content/migration.md
```diff
@@ -0,0 +1,535 @@
...
+- A comparison of AngularJS Material to Angular Material directives and services
+
+### Angular CDK
+
+Some of the features of AngularJS Material have been made more generic and moved into the [Angular
+CDK][cdk]. 
+
+These include:
+
+- [`$mdPanel`][109] becomes the CDK's [`Overlay`][110]
```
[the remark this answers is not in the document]
Splaktar: This summarizes the tables below to put this all in one place and make it readable for those who don't want to look at tables.

docs/content/migration.md
```diff
@@ -0,0 +1,535 @@
...
+  separate [section of this guide](#layout)
+
+### Compile-time vs. Run-time theming
+
+AngularJS Material offers powerful [Theming Features][106]. Many of these features do not exist in
+Angular Material. Additionally, while most theming was configured in JavaScript for
+AngularJS Material, Angular Material uses [Sass][170] features like variables and mixins to
+configure theming.
+
+#### AngularJS Material Compile-time Theming
```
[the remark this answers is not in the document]
Splaktar: Yes, especially for many advanced users who utilize https://github.com/angular/material-tools

docs/content/migration.md
```diff
@@ -0,0 +1,535 @@
...
+  copying one of the [provided palettes][167], modifying it, and then using it in your custom theme
+  creation
+- [Custom theme creation][164] including definition of color intentions and whether it is a dark or
+  light theme
+- [Generation of multiple themes][165] at build time
+- [Theme customization for a set of component types][166] (i.e. `mat-button` and `mat-checkbox`)
+- Use Angular Material themes to [style application or library components][168]
+
+Unsupported features:
+- There is no built-in, automatic way to change the theme of a specific Angular Material
```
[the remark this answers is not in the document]
Splaktar: I'd love to just state 
```
- [Theme customization of a specific set of Angular Material elements based on CSS selectors][link?]
```
But I'm not sure where this is documented? I wasn't able to find it anywhere that I searched.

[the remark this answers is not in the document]
Splaktar: Opened https://github.com/angular/material2/issues/15760 to request that be documented.

docs/content/migration.md
```diff
@@ -0,0 +1,535 @@
...
+- Dynamically apply `theme-palette-hue-opacity` color combinations as CSS styles to your own
+  application-specific components
+
+
+### <a name="layout"></a> Changes to Layout Features
+
+The AngularJS Material [layout][108] system includes includes attributes and classes like `layout`,
+`flex`, `show`, `hide`, etc.
+
+#### Angular Layout
```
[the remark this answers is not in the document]
Splaktar: They are in the process of re-branding and have already started changing their official docs (https://github.com/angular/flex-layout/wiki) to Angular Layout.

docs/content/migration.md
```diff
@@ -0,0 +1,535 @@
...
+#### Run-time Theming in an Angular Material app
+
+Features include:
+- Angular's [Meta Service][147] can be used to dynamically modify `<meta>` tags that affect browser
+header and status bar colors. The [AngularJS Material docs][149] for this feature have links to
+learn more about the specific `<meta>` tags.
+
+Unsupported features:
+- Changing a component's hues like AngularJS Material's [`md-hue-1` class][169]
+- Definition and generation of themes at run-time
```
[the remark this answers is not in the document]
Splaktar: Is there some documentation that I can point them to?

docs/content/migration.md
Splaktar: I'll tweak this a bit, but the suggested approach seems to omit (or leave the reader to assume) quite a bit of information.

docs/content/migration.md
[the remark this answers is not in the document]
Splaktar: For now I've just added
```
- Theme customization of a specific set of Angular Material elements based on CSS selectors
```
with no link or code example, as this part of the document doesn't really seem like the place for something like that. If we wanted to add another section to the end of this doc with some code examples, that might work, but it seems better documented in material.angular.io.

docs/content/migration.md
```diff
@@ -0,0 +1,503 @@
...
+- [Generation of multiple themes][165] at build time
+- [Theme customization for a set of component types][166] (i.e. `mat-button` and `mat-checkbox`)
+- Theme customization of a specific set of Angular Material elements based on CSS selectors using
+  Sass
+- Use Angular Material themes to [style application or library components][168]
+
+#### Run-time Theming in an Angular Material app
+
+Features include:
+- Angular's [Meta service][147] can be used to dynamically modify `<meta>` tags that affect browser
```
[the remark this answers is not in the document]
Splaktar: Yes, this link points to the framework docs. This is only here because in AngularJS Material this is a feature of theming and included in the theming docs. Users doing to migration will want to know how to migrate this feature.

docs/content/migration.md
```diff
@@ -0,0 +1,542 @@
...
+| $mdBottomSheet   |   [Docs][42] |   [Docs][38] | `MatBottomSheet`                                       |
+| $mdColors        |   [Docs][125]|   -          | [`mat-color` mixin][74] supports static theme color lookups.|
+| $mdCompiler      |   [Docs][126]|   -          | Not implemented                                        |
+| $mdCompilerProvider| [Docs][127]|   -          | Not implemented                                        |
+| $mdDateLocaleProvider|[Docs][128]|  [Docs][129]| `MomentDateAdapter`                                    |
+| $mdDialog        |   [Docs][48] |   [Docs][22] | `MatDialog`                                            |
+| $mdGestureProvider|  [Docs][130]|   -          | Not implemented                                        |
+| $mdIcon          |   [Docs][131]|   [Docs][132]| `svgIcon`                                              |
+| $mdIconProvider  |   [Docs][133]|   [Docs][134]| `MatIconRegistry`                                      |
+| $mdInkRipple     |   [Docs][58] |   [Docs][19] | `matRippleTrigger`                                     |
```
devversion: Is it right to mention the `matRippleTrigger` input here? Sounds a bit confusing to me. How about referring to the `MatRipple` directive instead and linking to the guide?

e.g. https://material.angular.io/components/ripple/overview#manual-ripples

Splaktar: The thought was that `$mdInkRipple` attaches ripples to an element and I had thought that `matRippleTrigger` was the way to do that. Re-reading the docs, it looks like the `matRippleTrigger` reference should be removed. Thanks.

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+AngularJS Material uses a service, [`$mdTheming`][143], to define themes and generate CSS styles at
+run-time.
+
+Defining and generating theme CSS at run-time is not supported by Angular Material. However,
+loading additional theme CSS into your application (or modifying an existing theme's `link` node
+to [swap theme CSS files][181]) at run-time is supported.
+
+Your application can choose to generate additional themes at compile-time or build an API to
+dynamically generate theme CSS on-demand. In either case, it is possible to [lazy load][165] the
+theme CSS at run-time.
```
Splaktar: There aren't any official Angular CLI docs (reported via https://github.com/angular/angular/issues/29822) that cover doing this but I found [this StackOverflow answer](https://stackoverflow.com/a/54855112/633107) the provides some, but not all, of the guidance needed.

Splaktar:
All of the TODO items have been handled. I'll leave this open to review feedback for another day before adding the `pr: merge ready` label.

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+| $mdGestureProvider|  [Docs][130]|   -          | Not implemented                                        |
+| $mdIcon          |   [Docs][131]|   [Docs][132]| `svgIcon`                                              |
+| $mdIconProvider  |   [Docs][133]|   [Docs][134]| `MatIconRegistry`                                      |
+| $mdInkRipple     |   [Docs][58] |   [Docs][19] |                                                        |
+| $mdInkRippleProvider|[Docs][135]|   [Docs][136]| `MAT_RIPPLE_GLOBAL_OPTIONS`                            |
+| $mdInteraction   |   [Docs][137]|   -          | Not implemented                                        |
+| $mdLiveAnnouncer |   [Docs][111]|   [Docs][112]| CDK `LiveAnnouncer`                                    |
+| $mdMedia         |   [Docs][137]|   [Docs][138]| CDK `BreakpointObserver.isMatched()` or Flex Layout's [MediaObserver service][119]|
+| $mdPanel         |   [Docs][109]|   [Docs][110]| CDK `Overlay`                                          |
+| $mdPanelProvider |   [Docs][139]|   -          | Not implemented                                        |
```
jelbourn: Isn't this just the configuration API for `$mdPanel`, which is not just captured as part of `Overlay`?

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+| $mdProgressCircularProvider|[Docs][140]|-      | Not implemented                                        |
```
jelbourn: I wouldn't say this is "Not implemented"; the strokeWidth is just an input on `mat-progress-spinner` and the size is controlled via CSS.

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+| $mdSidenav       |   [Docs][141]|   -          | Not implemented                                        |
```
jelbourn: I wouldn't say this is unimplemented so much as no longer necessary with the `MatSidenav` component

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+| $mdSticky        |   [Docs][142]|   -          | Not implemented                                        |
```
jelbourn: I'd mention for this one that the native `position: sticky` is a superior API for all browsers except IE11

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+| progress-circular  |   [Docs][56] |   [Docs][11] | `mat-progress-spinner`                                 |
+| radio-button       |   [Docs][57] |   [Docs][4]  |                                                        |
+| radio-group        |   [Docs][92] |   [Docs][93] |                                                        |
+| select             |   [Docs][59] |   [Docs][23] |                                                        |
+| select-on-focus    |   [Docs][94] |   -          | Not implemented                                        |
+| sidenav            |   [Docs][60] |   [Docs][6]  |                                                        |
+| sidenav-focus      |   [Docs][95] |   -          | [autofocus][96] only supports focusing the first focusable element.|
+| switch             |   [Docs][61] |   [Docs][14] | `mat-slide-toggle`                                     |
+| slider             |   [Docs][62] |   [Docs][16] |                                                        |
+| slider-container   |   [Docs][97] |   -          | Not implemented                                        |
```
jelbourn: I kind of want to say here "Replaced by literally putting whatever you want next to mat-slider"

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+| tooltip            |   [Docs][67] |   [Docs][18] |                                                        |
+| truncate           |   [Docs][103]|   -          | Not implemented                                        |
+| virtual-repeat     |   [Docs][68] |   [Docs][40] | `cdk-virtual-scroll-viewport` and `*cdkVirtualFor`     |
+| whiteframe         |   [Docs][104]|  [Guide][105]| Based on classes and mixins                            |
+
+### Services
+
+| Service          | Old          | New          | Special Notes                                          |
+|------------------|--------------|--------------|--------------------------------------------------------|
+| $mdAriaProvider  |   [Docs][124]|   -          | Not implemented                                        |
```
jelbourn: It would be good to say here something like "Use a tool like aXe or Chrome DevTool's Accessibility audit instead"

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+  `fxHide.gt-xs`, `[ngClass.sm]`, `[ngStyle.xs]`, and more.
+- Both APIs include [CSS Grid][65] features in addition to the [CSS Flexbox][flexbox] features from
+  AngularJS Material.
+- The ability to define [Custom Breakpoints][118]
+- A [MediaObserver service][119] that allows subscribing to `MediaQuery` activation changes
+
+Since AngularJS Material's inception, browser support for [CSS Flexbox][flexbox] has [broadly
+improved][177] and [CSS Grid][65] has become [widely available][178]. Application developers should
+carefully consider the the costs of additional run-time dependencies versus native alternatives
+when choosing an approach to layout. This library adds a nontrivial payload cost.
```
jelbourn: Andrew brought this up:

What do you think about adding a bit more here like
```markdown
Many of the AngularJS Material layout APIs map to straightforward, vanilla CSS using Flexbox
and media queries. We recommend checking out [the MDN docs](https://developer.mozilla.org/en-US/docs/Learn/CSS/CSS_layout/Flexbox),
[this scotch.io guide](https://scotch.io/tutorials/a-visual-guide-to-css3-flexbox-properties),
and [Flexbox Froggy](https://flexboxfroggy.com).

```

docs/content/migration.md
Splaktar: `$mdPanelProvider` allows developers to define configuration presets that will be stored within a cached presets object. It doesn't actually provide any configuration other than accepting a subset of the options for `$mdPanel.create()`. 

I wasn't able to find any idea of a "preset" or caching within the Overlay docs.

docs/content/migration.md
Splaktar: This is used to specify the default options for the `md-progress-circular` directive. I wasn't able to find any docs for configuring the default options for `mat-progress-spinner`.

docs/content/migration.md
Splaktar: This is just a convenience to allow something like `$mdSidenav('right').toggle()` and `$mdSidenav('right').toggle()` for use from components without the need to have an actual reference to the specific `md-sidenav` instance.

I wasn't able to find any service that provides this convenience in Angular Material. You need to manually grab a reference to the instance and then stash it in an application-wide service for later use.

Some of the features implemented by these services are still possible via custom code in Angular Material. Should I add a section here that explains how to accomplish some of these use cases in Angular Material?

docs/content/migration.md
Splaktar: Right, similar to https://github.com/angular/material/pull/11690#discussion_r274240311, there is no matching API, but we could add a section to explain how to accomplish some of these use cases in Angular Material? 

Certainly the answer for a number of these is that such an API is not needed because there is a better native or Angular-specific way to accomplish the same thing.

docs/content/migration.md
Splaktar: Yeah, I think that we can certainly flesh this out. I think that providing some links to learning resources will be good as well since many AngularJS Material developers have spent years not needing to learn the specifics of CSS Flexbox and CSS Grid.

docs/content/migration.md
Splaktar: Yeah, I think that we need another section below that explains these in more detail. Then we can change this `Not implemented` for many of these to `See below`.

docs/content/migration.md
devversion: I think in Angular the idiomatic way to do this is by accessing the template reference variables or through `ViewChild`. 

I would consider it an anti-pattern to have a global registry of sidenav's in Angular, so maybe saying this is not implemented is wrong. It's just implemented in a Angular idiomatic way.

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+
+[See the full documentation](https://material.angular.io/guide/theming) for details on themes,
+including examples of multiple themes.
+
+##### Dynamically applying themes
+
+AngularJS Material's [`md-theme`][155] and [`md-theme-watch`][156] directives support data binding
+like `md-theme="{{ dynamicTheme }}" md-theme-watch`.
+
+With Angular Material, you can make use of Angular's [ngClass][159] API to dynamically apply classes
```
kseamon: I'd emphasize a normal class binding over ngClass rather than the other way around

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+Most components use the same name between the two libraries, changing the `md-` prefix to `mat-`.
+For example, `<md-card>` becomes `<mat-card>`. Some element directives changed to attribute
+directives, for instance `<md-subheader>` changed to `matSubheader`.
+
+#### Component comparison reference 
+
+| Directive          | Old (`md-`)  | New (`mat-`) | Special Notes                                          |
+|--------------------|--------------|--------------|--------------------------------------------------------|
+| autocomplete       |   [Docs][41] |   [Docs][24] |                                                        |
+| autofocus          |   [Docs][69] |   -          | Not implemented                                        |
```
kseamon: cdkFocusInitial (Not a directive but works in numerous scenarios such as dialogs for specifying the initially focused element)

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+| button             |   [Docs][43] |   [Docs][1]  |                                                        |
+| calendar           |   [Docs][70] |   -          | Not implemented                                        |
```
kseamon: Maybe note that there is a Date picker but it is currently tied to inputs

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+| whiteframe         |   [Docs][104]|  [Guide][105]| Based on classes and mixins                            |
+
+### Services
+
+| Service          | Old          | New          | Special Notes                                          |
+|------------------|--------------|--------------|--------------------------------------------------------|
+| $mdAriaProvider  |   [Docs][124]|   -          | Not implemented                                        |
+| $mdBottomSheet   |   [Docs][42] |   [Docs][38] | `MatBottomSheet`                                       |
+| $mdColors        |   [Docs][125]|   -          | [`mat-color` mixin][74] supports static theme color lookups.|
+| $mdCompiler      |   [Docs][126]|   -          | Not implemented                                        |
```
kseamon: Maybe also worth noting that this sort of thing has no equivalent in Angular

docs/content/migration.md
```diff
@@ -0,0 +1,545 @@
...
+These are new components found in Angular Material and the Angular CDK that do not exist in
+AngularJS Material.
+
+The `mat-` and `cdk-` prefixes have been omitted to make this table more readable.
+
+| Directive        | New          | Special Notes                                          |
+|------------------|--------------|--------------------------------------------------------|
+| badge            |   [Docs][37] |                                                        |
+| button-toggle    |   [Docs][15] |                                                        |
+| table            |   [Docs][28] | Unstyled CDK and Material versions available.          |
```
kseamon: alphabetize?

docs/content/migration.md
Splaktar: I agree in general and considered that, but the `ngClass` example is the closest to the functionality of `md-theme="{{ dynamicTheme }}" md-theme-watch` unless there is a `[class.xxx]="yyy"` syntax that I'm forgetting which allows changing between multiple classes.

docs/content/migration.md
Splaktar: Ah yeah, https://material.angular.io/cdk/a11y/overview#regions seems like it would work. I don't think that I saw it because it [wasn't in the API docs](https://github.com/angular/material2/issues/15799).

docs/content/migration.md
Splaktar: I even found a `mat-calendar` component, but it's not documented and I don't believe that it can be used stand-alone. Note added. Thanks.

docs/content/migration.md
```diff
@@ -0,0 +1,575 @@
...
+
+```scss
+// Define a CSS class for a special button treatment.
+.super-special-button {
+  @include mat-button-theme($super-special-theme);
+}
+```
+```html
+<!-- Apply this treatment the same as you would with any other CSS class. -->
+<button mat-button class="super-special-button">Save</button>
```
jelbourn: I realized after writing this that the class actually has to be on a parent element

PR #11724, opened 2019-05-16
fix(select): edge case where the model value is unset w/ AngularJS 1.7.8

### LINKED ISSUE #11679: select: required unsets model value with AngularJS 1.7.8
Bug

https://codepen.io/jazdw/pen/VRyVYa?editors=1011

1. Run the codepen
2. Observe the error on the form and that $ctrl.value is now undefined
3. Also see that "Test value" remains selected in the drop down and has an asterisk next to it

It should not delete the value from the model, input should not have an error.

Value is deleted from the model, input has an error.

- AngularJS: 1.7.8
- AngularJS Material: 1.1.13
- OS: Windows 10
- Browsers: Chrome (latest), Firefox (latest)
 
Seems like behavior changes over the last few released of AngularJS have exposed this problem:
https://github.com/angular/angular.js/commit/69262239632027b373258e75c670b89132ad9edb
https://github.com/angular/angular.js/commit/5ad4f5562c37b1cb575e3e5fddd96e9dd10408e2
https://github.com/angular/angular.js/commit/a4c7bdccd76c39c30e33f6215da9a00cc8acde2c

The cause of the problem seems to be the $validate() method being called by the AngularJS required directive (via attr.$observe on required attribute) before the options are added to the mdSelect controller. When $validate() is called the mdSelect $isEmpty implementation returns true as the options are not populated, this results in the ngModelController writing undefined to the scope.

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

when md-select is used inside of an AngularJS component
- where the component sets an initial value
- and that value is bound to the md-select's ng-model
- and the options are using the value attribute (instead of ng-value)
- and the md-select has ng-required that binds to an expression
- and that expression is updated by $scope.$applyAsync in the component
then the value could end up being reset to undefined.

Issue Number: 
Fixes #11679

## What is the new behavior?
- with this fix, the md-select's value remains set
- add an example of testing an AngularJS Component

```
[ ] Yes
[x] No
```


Splaktar:
@Narretz and @gkalpak PTAL.

src/components/select/select.spec.js
```diff
@@ -1486,6 +1506,29 @@ describe('<md-select>', function() {
     });
   });
 
+  // Only test with custom components on AngularJS 1.5+
+  if (angular.version.major === 1 && angular.version.minor >= 5) {
+    describe('with custom components', function() {
+
+      if (angular.version.major === 1 && angular.version.minor >= 7) {
```
gkalpak: The `angular.version.major` check is redundant (since it has been checked above).
At the moment, there is only one test, so the two `if` blocks could be merged.

src/components/select/select.spec.js
```diff
@@ -1486,6 +1506,29 @@ describe('<md-select>', function() {
...
+        // The $flushPendingTasks API is new in AngularJS 1.7 and is required to reproduce the issue
```
gkalpak: FWIW, `$flushPendingTasks` is an alias for `$timeout#flush()` (which works the same on all 1.5+ versions).

src/components/select/select.spec.js
```diff
@@ -1486,6 +1506,29 @@ describe('<md-select>', function() {
...
+          inject(function($flushPendingTasks) {
+            var nodes = $compile('<form name="testForm">' +
+              ' <required-form-field-component ng-ref="comp"></required-form-field-component>' +
+              '</form>')($rootScope);
+
+            $rootScope.$digest();
+            $flushPendingTasks();
+
+            expect($rootScope.comp.value).toBe('test');
+            expect($rootScope.testForm.value.$error.required).not.toBe(true);
```
gkalpak: It is generally safer to assert the actual value: `expect(...).toBe(undefined)` (or whatever the value should be).
(Because for example, it might not be true, but be an object (as happens on forms for example), etc.)

src/components/select/select.spec.js
```diff
@@ -1486,6 +1506,27 @@ describe('<md-select>', function() {
...
 
+  // Only test with custom components on AngularJS 1.5+
+  if (angular.version.major === 1 && angular.version.minor >= 5) {
+    describe('with custom components', function() {
+
+      // The ng-ref API is new in AngularJS 1.7 and is required to reproduce the issue in this test.
+      if (angular.version.minor >= 7) {
+        it('should re-validate when the options and required value change', function() {
+          var nodes = $compile('<form name="testForm">' +
+            ' <required-form-field-component ng-ref="comp"></required-form-field-component>' +
```
Splaktar: @gkalpak thank you for the very informative review! Do you know of any alternative to `ng-ref` here?

gkalpak: No, no direct alternatives. You could create a simple directive to do that, but it would be overkill here imo.
The cleanest/easiest way (for the test) is probably: 

```js
var ctrl = nodes.find('required-form-field-component').controller('requiredFormFieldComponent');
```

Splaktar: That works, many thanks!

PR #11761, opened 2019-07-15
fix(select): overhaul screen reader support

### LINKED ISSUE #10748: select(multiple): aria-checked and aria-selected broken
**Actual Behavior**:
 -  `What is the issue? *`
`md-select` multiple values are always "checked" because of ngAria, and selected states don't clearly denote the control's intended usage.

Only checked values should be checked. There is a workaround, adding `aria-checked` to `md-option` with a correct binding will prevent the ng-aria behavior. Ideally `md-select` would fix the ngAria behavior internally.

The distinction between selected and checked needs to be more clearly defined and expressed. According to specs, selection should be managed with `aria-activedescendant` unless the multiple select actually supports selecting (not checking) multiple values. `aria-selected` would be appropriate for a control where holding ctrl highlights (selects) multiple values, and hitting space checks them.

The issue is ambiguous because best practices for multiselects follow a pattern that is completely incompatible with `md-select` (`md-select` is actually more of a persistent-selectable tree-view than a multiselect based on examples I've seen). Expected behavior is a best-guess based on how I interpret selected and checked when both are present while using a screen reader.

Example of the confusion in actual vs expected read values by NVDA **for an unchecked item**:
Actual:
Foo item 1 of 3 checked not-selected
Expected:
Foo item 1 of 3 not checked

### LINKED ISSUE #10967: select: does not work well with VoiceOver
The component does not act as one whole form control; we may want to explore changing this from `role="listbox"` to `"combobox"` to get better results with VoiceOver specifically. Will be tricky to not mess up NVDA and JAWS at the same time.

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

- `md-select` does not work well with VoiceOver on iOS or macOS
- When using `md-select` with the `multiple` attribute, options are always "checked" because of ngAria.

Issue Number: 
Fixes #10748. Fixes #10967.

## What is the new behavior?
- move to [WAI-ARIA's Collapsible Dropdown Listbox practice](https://www.w3.org/TR/2019/NOTE-wai-aria-practices-1.1-20190207/examples/listbox/listbox-collapsible.html)
  - don't apply aria-required to md-select
  - as it isn't compatible with the button role
  - the md-content element is now the listbox
      - has the appropriate attributes and a unique id
      - and receives focus when the pop-up panel opens
      - aria-owns now points to this listbox so that indexes work
  - option focus is handled via `aria-activedescendant`
  - remove `aria-expanded` when collapsed
  - remove `aria-disabled` attribute when not disabled
-  manually remove `aria-checked` set by ngAria due to ngValue usage
-  apply `md-focused` class to the option with focus
- improve `ng-multiple` implementation
  - account for `multiple` attribute on `md-select-menu`
- remove unused `deregisterCollectionWatch()`
- fix overloaded variable names
- don't set aria-selected="false" on options in single selection mode
- stop labels and values from being announced multiple times
- add JSDoc/Closure Compiler details and types
- refinements for VoiceOver users
- clean up watchers, observers, and event handlers on $destroy
- fix a case where the initial selection model could contain two values
  - for the empty option, i.e. "" and "None"
  - deselection was only clearing the first one in single selection mode
- reduce duplicated code for focusing option nodes
- improve keyboard option scrolling behavior
- eliminate duplicate call to `autoFocus()`

```
[ ] Yes
[x] No
```

## Other information
- [Deque Quick Reference Guide: VoiceOver for iOS](https://dequeuniversity.com/assets/pdf/screenreaders/voiceover-ios-guide.pdf)
- [Accessibility testing with Android Talkback](https://developer.paciellogroup.com/blog/2015/10/accessibility-testing-with-android-talkback/)
- [VoiceOver iOS Gesture/Keyboard Commands](http://pauljadam.com/demos/voiceovercommands.html)

### Completed Testing and Issues Identified
- [x] Finalize testing on VoiceOver in Chrome
- [x] Finalize testing on VoiceOver in Safari
- [x] Finalize testing on NVDA
- [x] Escape doesn't set the model back to the previous value
- [x] User defined `aria-labelledby` gets overwritten
- [x] `aria-hidden` on `md-select-value` is cleared when pop-up is closed, resulting in label being read twice when the `md-select` has no selection
- [x] Test in VoiceOver for iOS on Safari
  - [x] Test with [VoiceOver rotor](https://github.com/angular/material/issues/10967#issuecomment-354295026)
    - This seems to be working well as using the swipe right/left now cycles through the options as intended. 
    - It should be noted that `md-select` elements show up under the Buttons Rotor instead of Form Controls. This seems to be working as intended.
  - [x] Test [exiting the pop-up panel in multiple selection mode](https://github.com/angular/material/issues/10967#issuecomment-356022949)
    - [ ] This issue with being difficult to exit a multiple selection select exists with Angular Material as well. The "Two finger scrub" to dismiss alerts/dialogs or go back a page seems to always go back a page when used with AngularJS/Angular Material Autocomplete and Select. I tried changing the `md-select-menu-container` to `role="dialog"` and adding `"aria-haspopup="dialog"` to the `md-select`, but that didn't help and the Z (two finger scrub) always went back to the previous page. This is being tracked in https://github.com/angular/material/issues/11791.
- [x] Test on ChromeVoX
  - [ ] ChromeVox doesn't handle `aria-lablelledby` well and reads the `md-select-value` first and then the `aria-label` followed by the `md-select-value` again. Ideally, the label should be read and then the value. This will need to be handled in separate issue after consulting with the a11y team and possibly opening a bug against ChromeVox.
- [x] Autocomplete options' hover styling doesn't extend fully to the left edge
- [x] Test with TalkBack on Chrome for Android. I tested and saw similar behavior on Firefox for Android as well.
  - [x] Single selection appears to be working well. "Checked" is no longer announced in single selection mode.
  - [x] Multiple selection can be dismissed by a separated two finger tap.
  - [ ] Options in multiple selection cannot be selected (checked). This works for Angular Material, but not here. This isn't a regression as this has always been a problem with `md-select` on Android (back at least to `1.1.5`). This is being tracked in https://github.com/angular/material/issues/11770. This is likely a bug with TalkBack. Need to try to reproduce with a WAI-ARIA example and then find out where to submit the defect against TalkBack or Chrome for Android.

### DIFF
--- a/docs/app/css/style.css
+++ b/docs/app/css/style.css
@@ -155,13 +155,13 @@ ul {
   margin: 0;
   padding: 0;
 }
-ul li:not(.md-nav-item) {
+ul:not(.md-autocomplete-suggestions) li:not(.md-nav-item) {
   margin-left: 16px;
   padding: 0;
   margin-top: 3px;
   list-style-position: inside;
 }
-ul li:not(.md-nav-item):first-child {
+ul:not(.md-autocomplete-suggestions) li:not(.md-nav-item):first-child {
   margin-top: 0;
 }
 /************
--- a/src/components/select/demoBasicUsage/index.html
+++ b/src/components/select/demoBasicUsage/index.html
@@ -30,7 +30,8 @@
           <label>State</label>
           <md-select ng-model="ctrl.userState">
             <md-option><em>None</em></md-option>
-            <md-option ng-repeat="state in ctrl.states" ng-value="state.abbrev" ng-disabled="$index === 1">
+            <md-option ng-repeat="state in ctrl.states" ng-value="state.abbrev"
+                       ng-disabled="$index === 1">
               {{state.abbrev}}
             </md-option>
           </md-select>
@@ -60,7 +61,8 @@
 
       <div layout="row" layout-align="space-between center">
         <span>What armor do you wear?</span>
-        <md-select ng-model="armor" placeholder="Armor" class="md-no-underline" required md-no-asterisk="false">
+        <md-select ng-model="armor" placeholder="Armor" class="md-no-underline" required
+                   md-no-asterisk="false">
           <md-option value="cloth">Cloth</md-option>
           <md-option value="leather">Leather</md-option>
           <md-option value="chain">Chainmail</md-option>
--- a/src/components/select/demoOptionGroups/index.html
+++ b/src/components/select/demoOptionGroups/index.html
@@ -9,7 +9,7 @@
         </md-select>
       </md-input-container>
       <md-input-container>
-        <label>Topping</label>
+        <label>Toppings</label>
         <md-select ng-model="selectedToppings" multiple>
           <md-optgroup label="Meats">
             <md-option ng-value="topping.name" ng-repeat="topping in toppings | filter: {category: 'meat' }">{{topping.name}}</md-option>
--- a/src/components/select/select-theme.scss
+++ b/src/components/select/select-theme.scss
@@ -154,19 +154,22 @@ md-select-menu.md-THEME_NAME-theme {
 
       &:not([disabled]) {
         &:focus,
-        &:hover {
+        &:hover,
+        &.md-focused {
           background-color: '{{background-500-0.18}}'
         }
       }
 
       &[selected] {
         color: '{{primary-500}}';
-        &:focus {
+        &:focus,
+        &.md-focused {
           color: '{{primary-600}}';
         }
         &.md-accent {
           color: '{{accent-color}}';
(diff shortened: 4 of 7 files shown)


mmalerba:
This has ~50 targets failing in g3, will need to look into it more when I get a chance

Splaktar:
Rebased.

Splaktar:
Rebased.

PR #11801, opened 2019-09-26
fix(slider): vertical slider in a scrolled container sets value to zero on all clicks

Thaina:
## PR Checklist
Please check that your PR fulfills the following requirements:
- [ ] The commit message follows [our guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format)
- [ ] Tests for the changes have been added or this is not a bug fix / enhancement
- [ ] Docs have been added, updated, or were not required

## PR Type
What kind of change does this PR introduce?

```
[ x ] Bugfix
[ x ] Enhancement
[ ] Documentation content changes
[ ] Code style update (formatting, local variables)
[ ] Refactoring (no functional changes, no api changes)
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

## What is the current behavior?

Issue Number: https://github.com/angular/material/issues/11800

## Does this PR introduce a breaking change?
```
[ ] Yes
[ ] No
```

PR #11804, opened 2019-09-27
fix(slider): clicking vertical slider sets value to 0

### LINKED ISSUE #11800: slider(vertical): clicking slider sets value to 0 instead of value clicked when the scroll bar is not at the top
I made a page with vertical `md-slider` inside table. Whenever I scroll down the page, the slider then cannot be clicked or dragged, it will reset it value to minimum like it cannot calculate its value

It seem like it use the page's scroll position to calculate its knob position or something. How can I fix this?

As the picture below. I scroll this page at the top. And then click on the green slider. The value become `61.5` normally

image

But when I just scroll down the page a bit, I try to click the same position on slider. It then become 0 just by clicked

image

Thaina:
Fix #11800 alternative to Fix #11801

## PR Type

```
[ x ] Bugfix
[ x ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Issue Number: #11800

## What is the new behavior?
Correction for scroll

```
[ ] Yes
[ ] No
```

### DIFF
--- a/src/components/slider/slider.js
+++ b/src/components/slider/slider.js
@@ -643,15 +643,15 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
      * @returns {number}
      */
     function positionToPercent(position) {
-      var offset = vertical ? sliderDimensions.top : sliderDimensions.left;
+      var offset = vertical ? (sliderDimensions.top + window.scrollY) : (sliderDimensions.left + window.scrollX);
       var size = vertical ? sliderDimensions.height : sliderDimensions.width;
       var calc = (position - offset) / size;
 
-      if (!vertical && $mdUtil.isRtl(attr)) {
+      if (vertical || (!vertical && $mdUtil.isRtl(attr))) {
         calc = 1 - calc;
       }
 
-      return Math.max(0, Math.min(1, vertical ? 1 - calc : calc));
+      return Math.min(Math.max(0, calc), 1);
     }
 
     /**
--- a/src/components/slider/slider.js
+++ b/src/components/slider/slider.js
@@ -644,7 +644,7 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
      * @returns {number}
      */
     function positionToPercent(position,wnd) {
-      if(!wnd)
+      if (!(wnd && Number.isFinite(wnd.scrollX) && Number.isFinite(wnd.scrollY)))
         wnd = window;
       var offset = vertical ? (sliderDimensions.top + wnd.scrollY) : (sliderDimensions.left + wnd.scrollX);
       var size = vertical ? sliderDimensions.height : sliderDimensions.width;
--- a/src/components/slider/slider.js
+++ b/src/components/slider/slider.js
@@ -640,10 +640,13 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
     /**
      * Convert position on slider to percentage value of offset from beginning...
      * @param position
+     * @param {{ scrollX:number,scrollY:number }} wnd
      * @returns {number}
      */
-    function positionToPercent(position) {
-      var offset = vertical ? (sliderDimensions.top + window.scrollY) : (sliderDimensions.left + window.scrollX);
+    function positionToPercent(position,wnd) {
+      if(!wnd)
+        wnd = window;
+      var offset = vertical ? (sliderDimensions.top + wnd.scrollY) : (sliderDimensions.left + wnd.scrollX);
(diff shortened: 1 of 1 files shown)


Thaina:
@Splaktar How could we made unittest for something that require visually interaction and environment variable on rendered browser like scrollbar position though?

src/components/slider/slider.js
```diff
@@ -643,15 +643,15 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
      * @returns {number}
      */
     function positionToPercent(position) {
-      var offset = vertical ? sliderDimensions.top : sliderDimensions.left;
+      var offset = vertical ? (sliderDimensions.top + window.scrollY) : (sliderDimensions.left + window.scrollX);
```
Splaktar: So the idea is that this makes the offset relative to the current viewport (that is scrolled into view) rather than the entire page?

src/components/slider/slider.js
```diff
@@ -643,15 +643,15 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
...
       var size = vertical ? sliderDimensions.height : sliderDimensions.width;
       var calc = (position - offset) / size;
 
-      if (!vertical && $mdUtil.isRtl(attr)) {
+      if (vertical || (!vertical && $mdUtil.isRtl(attr))) {
```
Splaktar: Can you help me understand why this change is needed? Is this just moving some of the logic from line 654 up to this line?

src/components/slider/slider.js
```diff
@@ -643,15 +643,15 @@ function SliderDirective($$rAF, $window, $mdAria, $mdUtil, $mdConstant, $mdThemi
...
         calc = 1 - calc;
       }
 
-      return Math.max(0, Math.min(1, vertical ? 1 - calc : calc));
+      return Math.min(Math.max(0, calc), 1);
```
Splaktar: Would calling the above `clamp(calc)` function make sense here now?

Splaktar:
> @Splaktar How could we made unittest for something that require visually interaction and environment variable on rendered browser like scrollbar position though?

That would require an end to end (e2e) test which we don't currently support in this repo.

However, we can still create unit tests for the `positionToPercent()` function without needing to send any clicks or calculate scroll if we pass in the `scrollX` and `scrollY` as parameters.

src/components/slider/slider.js
Thaina: @Splaktar No it's opposite. `sliderDimensions.top` and `sliderDimensions.left` is actually relative to current viewport. So we need to `+ scrollY` to make it become relative to the entire page instead. And it would match with `pointer.x`

src/components/slider/slider.js
Thaina: @Splaktar Because it was the exactly same logic. `x = 1 - x`. It just 2 condition for this logic. `If the slider is vertical` OR `if the slider is horizontal and RTL` it then inverse the value. So I prefer to have same code in the same `if` block

Thaina: Maybe we could use exclusive or `^`?

```
if (vertical ^ $mdUtil.isRtl(attr))
```

I felt unfamiliar with XOR in js though

src/components/slider/slider.js
Thaina: @Splaktar I don't know that we have `clamp` function in this file. The original one also use `Math.max(0, Math.min(1, value))`. I just prefer that if we made a `minmax` in place of `clamp` we should order it in the same manner as `clamp`

Thaina:
> However, we can still create unit tests for the `positionToPercent()` function without needing to send any clicks or calculate scroll if we pass in the `scrollX` and `scrollY` as parameters rather than pulling them from `window` in the function.

@Splaktar could we simulate `window.scroll` like this though?

```js
//// unittest
var tmp = window.scrollY;
window.scrollY = $simValue;
var percent = positionToPercent(position);
window.scrollY = tmp;
```

src/components/slider/slider.js
Splaktar: No, this is better than XOR. Thank you for the explanation.

Splaktar:
> @Splaktar could we simulate window.scroll like this though?

Probably not in a reliable way. 

If we change the function to `function positionToPercent(position, x, y) {` then we could make it easy to unit test without depending upon the browser's size and ability to scroll the actual browser in a unit test.

Thaina:
Maybe like this?

```js
//// unittest
var percent0 = positionToPercent(position);
var percent1 = positionToPercent(position,{ scrollX : testX,scrollY : testY });
```

Splaktar:
The alternative implementation of this was [merged](https://github.com/angular/material/pull/11801) and released in `1.1.23`.

PR #11843, opened 2020-01-15
fix(toast): toast defaults to wrong values when passed an empty position

oliversalzburg:
If the user specifies no position at all, the toast will open from the top, while the class name will reflect a toast opening from the bottom.

By chaning the ternary as suggested, we end up with the correct default case while maintaining all configured cases as-is.

## PR Checklist

## PR Type

```
[X] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Go to https://material.angularjs.org/latest/demo/toast and uncheck all 4 checkboxes, then trigger a toast. The toast will appear on the top, the bottom FAB will make space for the toast.

## What is the new behavior?

The toast will appear on the top, the top FAB will make space for the toast.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/components/toast/toast.js
+++ b/src/components/toast/toast.js
@@ -541,6 +541,11 @@ function MdToastProvider($$interimElementProvider) {
         return 'md-toast-open-bottom';
       }
 
+      // If no position was provided, use the default.
+      if (!position) {
+        return 'md-toast-open-bottom';
+      }
+
       return 'md-toast-open-' + (position.indexOf('top') > -1 ? 'top' : 'bottom');
     }
 


Splaktar:
From https://material.angularjs.org/latest/api/service/$mdToast#mdtoast-show-optionsorpreset:

> position - `{string=}`: Sets the position of the toast.
Available: any combination of 'bottom', 'left', 'top', 'right', 'end', and 'start'. The properties 'end' and 'start' are dynamic and can be used for RTL support.
>
> Default combination: 'bottom left'.

I managed to fix [the example](https://codepen.io/Splaktar/pen/rNaEbWW?editors=1011) by not passing `""` into `.position()` as that is an invalid value. However, the code should probably handle the empty string input by using the default `"bottom left"` instead of having it somehow result in `"top left"`.

oliversalzburg:
Good point. I will submit an updated version next week.

Splaktar:
Sounds good.

oliversalzburg:
@Splaktar Sorry for the delay. It should be better now.

Splaktar:
I tested this updated fix and it doesn't appear to work.

After some investigation, I discovered that there was actually a different place in the code that needed to change. I've opened PR https://github.com/angular/material/pull/11904 to solve this.

PR #11844, opened 2020-01-17
feat: Declare a DevContainer for easy onboarding

oliversalzburg:
DevContainers are a feature of VS Code: https://code.visualstudio.com/docs/remote/containers

They allow us to declare a development environment that is independent from the developers workstation. This allows us to easily work in a Node <12 environment, regardless of the Node version installed on the developers workstation.

## PR Checklist

## PR Type
```
[ ] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[X] Other... Please describe: Developer feature
```

AngularJS Material requires NodeJS <12 for development. Developers may already be using NodeJS 12+.

## What is the new behavior?

Developers will get the ability to easily work on AngularJS Material in a defined environment, given that they meet the requirements to use the VS Code DevContainers feature.

```
[ ] Yes
[X] No
```

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,0 +1,132 @@
+# Based on the example NodeJS DevContainer
+FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:10
+
+# The image referenced above includes a non-root user with sudo access. Add
+# the "remoteUser" property to devcontainer.json to use it. On Linux, the container
+# user's GID/UIDs will be updated to match your local UID/GID when using the image
+# or dockerFile property. Update USER_UID/USER_GID below if you are using the
+# dockerComposeFile property or want the image itself to start with different ID
+# values. See https://aka.ms/vscode-remote/containers/non-root-user for details.
+ARG USER_UID=1000
+ARG USER_GID=$USER_UID
+ARG USERNAME=node
+
+# Uses parts from the CircleCI Node container to ensure alignment with test environment
+# https://github.com/CircleCI-Public/circleci-dockerfiles/blob/master/node/images/10.18.0-stretch/Dockerfile
+# https://github.com/CircleCI-Public/circleci-dockerfiles/blob/master/node/images/10.18.0-stretch/browsers/Dockerfile
+# The parts have been cleaned up to introduce consistency and remove all the
+# useless parts there were likely copied from StackOverflow answers.
+
+# Make apt non-interactive
+RUN echo 'APT::Get::Assume-Yes "true";' > /etc/apt/apt.conf.d/90circleci \
+	&& echo 'DPkg::Options "--force-confnew";' >> /etc/apt/apt.conf.d/90circleci
+
+ENV DEBIAN_FRONTEND=noninteractive
+
+# Debian Jessie is EOL'd and original repos don't work.
+# Switch to the archive mirror until we can get people to
+# switch to Stretch.
+RUN if grep -q Debian /etc/os-release && grep -q jessie /etc/os-release; then \
+	rm /etc/apt/sources.list && \
+	echo "deb http://archive.debian.org/debian/ jessie main" >> /etc/apt/sources.list && \
+	echo "deb http://security.debian.org/debian-security jessie/updates main" >> /etc/apt/sources.list \
+	; fi
+
+# Make sure PATH includes ~/.local/bin
+# https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=839155
+# This only works for root. The circleci user is done near the end of this Dockerfile
+RUN echo 'PATH="$HOME/.local/bin:$PATH"' >> /etc/profile.d/user-local-path.sh
+
+# man directory is missing in some base images
+# https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=863199
+RUN apt-get update && \
+	mkdir -p /usr/share/man/man1 && \
+	apt-get install -y git mercurial xvfb apt locales sudo openssh-client ca-certificates tar gzip parallel net-tools netcat unzip zip bzip2 gnupg curl wget make
+
+# Set timezone to UTC by default
+RUN ln -sf /usr/share/zoneinfo/Etc/UTC /etc/localtime
+
+# Use Unicode
+RUN locale-gen C.UTF-8 || true
+ENV LANG=C.UTF-8
+
+# Install jq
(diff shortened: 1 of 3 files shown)


Splaktar:
I haven't heard of this feature of VSCode, but it seems useful. 

We should probably support a full set of VS Code options/extensions similar to Angular and [Angular Components](https://github.com/angular/components/tree/fbae483caef90b2120b032c7f297ca34884bb375/.vscode).

However, we would need to add some Markdown documentation for both contributing with VS Code and separately, using the DevContainer.

For now, I'm going to target this at the `1.2.0` release where we will have a bit more flexibility for these types of changes.

.devcontainer/Dockerfile
```diff
@@ -0,0 +1,28 @@
+# Based on the example NodeJS DevContainer
+FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:0.43-10
```
Splaktar: I tracked down https://github.com/microsoft/vscode-dev-containers/tree/master/containers/javascript-node-10, but I wasn't able to find where this `0.43` release is recorded or where its changelog can be found.

oliversalzburg: Does this help? https://github.com/microsoft/vscode-dev-containers/tree/v0.43.0/containers/javascript-node-10

I just used https://github.com/microsoft/vscode-remote-try-node/blob/master/.devcontainer/Dockerfile and checked if there was a Node 10 label for the image and it worked.

Splaktar: OK, yeah it looks like there is a more recent `v0.101.0` release (https://github.com/microsoft/vscode-dev-containers/tree/v0.101.0/containers/javascript-node-10). It looks like they are updating these versions pretty quickly. Does something like `*-10` work to get the latest NodeJS 10 container?

oliversalzburg: I'm not sure. I'm pretty new to the concept myself and thought this project might be a good candidate to use it in, because of the NodeJS version requirement. The feature is still under development. I guess that's the reason for the frequent updates.

Too lax version ranges sometimes cause future problems, because a newer version is pulled that was not verified to work. That's why I often prefer to use a known-good version, even if it's a bit outdated.

oliversalzburg: Regardless, it doesn't work:
> [2020-02-14T11:46:17.459Z] [PID 6416] [1146 ms] Step 1/5 : FROM mcr.microsoft.com/vscode/devcontainers/javascript-node:*-10
invalid reference format

oliversalzburg: It also appears to me that all devcontainer images are built through the same project/pipeline. So the 0.101.0 is actually the version for the parent project. The NodeJS container has only seen a handful of changes since 0.43.0: https://github.com/microsoft/vscode-dev-containers/commits/v0.101.0/containers/javascript-node-10/.devcontainer/Dockerfile

I will probably update this PR to 0.101.0 anyway though, as I like the changes that were made in the meanwhile.

marosoft: Looking at this [web page](https://hub.docker.com/_/microsoft-vscode-devcontainers?tab=description) I found that you should be able to use `mcr.microsoft.com/vscode/devcontainers/javascript-node:10`

I checked it and it worked for me

oliversalzburg: @marosoft Thanks for the heads-up. Would you prefer I use that tag over a fixed version? I'm still slightly concerned with the instability a non-fixed version tag might introduce.

Splaktar: In general, I prefer fixed versions, but perhaps, since this is fairly early on, we should go with the `:10` and document how to fix the version if you run into a version that doesn't work. We should also mention in the docs for the person to open an issue. Then we can evaluate over time if there is a need to fix the version due to breaking changes/issues.

oliversalzburg:
@Splaktar That seems reasonable.

The last time I checked that guide, I got the impression that it was abandoned, because they mention `gulp server`, which isn't a valid gulp task.

Should I add a new section about DevContainer usage at the bottom and then maybe create a new ticket to review and correct the guide?

oliversalzburg:
I added some documentaiton to the guide, although a lot of it is just describing the DevContainer feature in general. I tried to not duplicate too much. Users who use the extension or plan to use it, will likely educate themselves through the official documentation to an extent that they don't require a lot of introduction into the specifics in this project, as there are almost none.

I've also since added the default LiveReload port of gulp-connect to the forwarded ports, which now enabled the livereload feature as well when using the DevContainer.

Splaktar:
> The last time I checked that guide, I got the impression that it was abandoned, because they mention `gulp server`, which isn't a valid gulp task.

It's not abandoned and issues should be filed if it's not accurate. I've gone through and refreshed most of these, but this one does look like it needs a lot of attention. Thank you for the heads up.

> Should I add a new section about DevContainer usage at the bottom and then maybe create a new ticket to review and correct the guide?

Yes, please.

docs/guides/BUILD.md
```diff
@@ -199,6 +199,16 @@ After running `gulp server` to start a *LiveReload* server in your project root:
 * Open Dev Tools and debug...
 
 
+### <a name="devcontainer"></a> VS Code DevContainer
+
+When using [VS Code](https://code.visualstudio.com/) as a development environment, you can [develop inside a container](https://code.visualstudio.com/docs/remote/containers) to allow for a predefined environment in which AngularJS Material code is being built.
+
+To use the DevContainer environment defined for AngularJS Material, just open the project in VS Code and you should be prompted to use the container automatically, if you have [Remote Development extension pack](https://marketplace.visualstudio.com/items?itemName=ms-vscode-remote.vscode-remote-extensionpack) installed.
+
+When you first try to open the project in the container, VS Code will build the container (this can take a while) and then make our source code available to it. When we build the documentation in the container and serve it through `gulp site`, the port will automatically be forwarded and the documentation will be available on your local development machine on http://localhost:8080.
```
Splaktar:
```suggestion
When you first try to open the project in the container, VS Code will build the container (this can take a while) and then make our source code available to it. When we build the documentation in the container and serve it through `npm run docs:watch`, the port will automatically be forwarded and the documentation will be available on your local development machine on http://localhost:8080.
```

Splaktar: We don't really want to refer directly to `gulp` commands anymore. I'll open a PR soon to clean up this guide.

Splaktar: PR here: https://github.com/angular/material/pull/11880

oliversalzburg:
I integrated a lot of the CircleCI stuff and cleaned it up. Full test suite can be run in the container now.

PR #11846, opened 2020-01-21
fix(menu-bar): z-index not restored on menu close

### LINKED ISSUE #11235: menuBar: failure to remove md-has-open-menu selector
### Bug:
When md-menu is used inside md-toolbar after clicking on some nested menu option, the md-toolbar still contains md-has-open-menu which causes z-index of toolbar to be increased all the time.

md-toolbar should remove md-has-open-menu after closing the md-menu.

md-toolbar contains md-has-open-menu even after closing the md-menu.

1. Select on _View_ menu.
2. Click on _Mode --> Edit_.
3. Now look at the md-toolbar, it will still contain .md-has-open-menu.

The z-index caused because of md-has-open-menu breaks UI for some other elements(for ex: md-sidenav ui breaks, since its z-index is only 60 while that has 100)
screenshot from 2018-04-16 19-53-58

The issue was there from v1.1.1 till now v1.1.8

screenshot from 2018-04-16 18-56-34

oliversalzburg:
When an mdMenuBar contains a menu and that menu is opened, the menu bar will receives a CSS class that ultimately affects its z-order. When the menu is closed, the CSS class is supposed to be removed from the menu bar.

When the menu has multiple levels, the menu hierarchy can not be correctly resolved upon menu closure. This results in the CSS class not being removed from the menu bar.

I assumed that this failure is mostly due to the attempt to detect menu structure through DOM traversal, which will not work in this case as the menu hierarchy is not directly reflected in the DOM structure (menus are attached to the body instead).

Fixes #11235

## PR Checklist

## PR Type
```
[X] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Issue Number: #11235

## What is the new behavior?
I store a reference to the original menu bar controller to establish the hierarchy when handling the menu closure. This way the CSS class is properly removed from the menu bar and Z order is restored.

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/components/menu/js/menuDirective.js
+++ b/src/components/menu/js/menuDirective.js
@@ -258,6 +258,7 @@ function MenuDirective($mdUtil) {
   function link(scope, element, attr, ctrls) {
     var mdMenuCtrl = ctrls[0];
     var isInMenuBar = !!ctrls[1];
+    var mdMenuBarCtrl = ctrls[1];
     // Move everything into a md-menu-container and pass it to the controller
     var menuContainer = angular.element('<div class="_md md-open-menu-container md-whiteframe-z2"></div>');
     var menuContents = element.children()[1];
@@ -275,7 +276,6 @@ function MenuDirective($mdUtil) {
 
     element.append(menuContainer);
     menuContainer[0].style.display = 'none';
-    mdMenuCtrl.init(menuContainer, { isInMenuBar: isInMenuBar });
-
+    mdMenuCtrl.init(menuContainer, { isInMenuBar: isInMenuBar, mdMenuBarCtrl: mdMenuBarCtrl });
   }
 }
--- a/src/components/menu/js/menuController.js
+++ b/src/components/menu/js/menuController.js
@@ -29,6 +29,7 @@ function MenuController($mdMenu, $attrs, $element, $scope, $mdUtil, $timeout, $r
     triggerElement.setAttribute('aria-expanded', 'false');
 
     this.isInMenuBar = opts.isInMenuBar;
+    this.mdMenuBarCtrl = opts.mdMenuBarCtrl;
     this.nestedMenus = $mdUtil.nodesToArray(menuContainer[0].querySelectorAll('.md-nested-menu'));
 
     menuContainer.on('$mdInterimElementRemove', function() {
--- a/src/components/menuBar/js/menuBarController.js
+++ b/src/components/menuBar/js/menuBarController.js
@@ -51,7 +51,8 @@ MenuBarController.prototype.init = function() {
       el[0].classList.remove('md-open');
     }
 
-    if ($element[0].contains(el[0])) {
+    var ctrl = angular.element(el[0]).controller('mdMenu');
+    if (ctrl.isInMenuBar && ctrl.mdMenuBarCtrl === self) {
       var parentMenu = el[0];
       while (parentMenu && rootMenus.indexOf(parentMenu) == -1) {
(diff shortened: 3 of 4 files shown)


src/components/menu/js/menuDirective.js
```diff
@@ -258,6 +258,7 @@ function MenuDirective($mdUtil) {
   function link(scope, element, attr, ctrls) {
     var mdMenuCtrl = ctrls[0];
     var isInMenuBar = !!ctrls[1];
+    var mdMenuBarCtrl = ctrls[1];
```
Splaktar: What's the value when not in a menu bar? `null`? `undefined`? If that's the case, then I suppose this is fine in that case, but I want to be sure that it's not something else.

src/components/menu/js/menuDirective.js
```diff
@@ -275,7 +276,7 @@ function MenuDirective($mdUtil) {
 
     element.append(menuContainer);
     menuContainer[0].style.display = 'none';
-    mdMenuCtrl.init(menuContainer, { isInMenuBar: isInMenuBar });
+    mdMenuCtrl.init(menuContainer, { isInMenuBar: isInMenuBar, mdMenuBarCtrl : mdMenuBarCtrl });
```
Splaktar:
```suggestion
    mdMenuCtrl.init(menuContainer, { isInMenuBar: isInMenuBar, mdMenuBarCtrl: mdMenuBarCtrl });
```

oliversalzburg:
Thanks for the quick review! I'll get back to it later tonight.

src/components/menu/js/menuDirective.js
oliversalzburg: It appears to be `null` in that case.

src/components/menu/js/menuDirective.js
oliversalzburg: Thanks for the suggestion. I'll make sure to apply this in my next change.

oliversalzburg:
I can't get ChromeHeadless to work in the DevContainer I'm using. I can't properly write the tests like this. I might try a couple of pushes and see what I get out of CircleCI, but I might have to temporarily downgrade some machine to Node 10 to finish this later.

oliversalzburg:
Alright, those were though to write, but I think I got something acceptable.

In the `should close when clicking on a nested menu item` test, I also wanted to spy on the `close()` of the nested menu controller, but I have no idea how to retrieve it. I opted to skip that check. If you feel like it should be added, please help me out with that controller reference.

In general, I didn't add tests specifically for the existence of the class, but focused on the general closing behavior that was already under test by existing tests. These tests included assertions for the removal of the `md-has-open-menu` class. So it seemed to make most sense to just extend this area of the test suite to cover more deeply nested menus.

Splaktar:
> I might have to temporarily downgrade some machine to Node 10 to finish this later.

I'm not sure if you're on Windows or not, but NVM is one of a few options for quickly switching your NodeJS installation. The most mature one is macOS, but there is a separate version for Windows as well (plus another option or two).

oliversalzburg:
Yeah, I worked with nvm/nvs/... in the past (and I'm on Windows using bash). It always ended up causing more problems than it solved.

I opted to create a quick Hyper-V Windows Development VM. Worked well enough. Thanks for the suggestion though :)

oliversalzburg:
Should I rewrite the commit message to use `fix(menu-bar):` as well?

Splaktar:
That would be helpful, thank you.

PR #11857, opened 2020-02-14
chore: Introduce Editorconfig

oliversalzburg:
Any .editorconfig was set to be ignored in cf92799314ff5483fbc588241dbaf72ba0a7a97d.
This change introduces such a configuration file to allow contributors to use the correct code style more easily.

I'm not sure why the file was set to be ignored. IMHO it appears to be beneficial to place it in the repository to make it clear which indentation style is expected.

When I open the project in VS Code, it always attempts to indent using tabs, which is very frustrating. Obviously, I can just place my `.editorconfig` file from this PR into the project and resolve the issue for myself, but having the style defined right from the beginning for all developers would decrease onboarding friction.

## PR Checklist
- [ ] The commit message follows [our guidelines](https://github.com/angular/material/blob/master/.github/CONTRIBUTING.md#-commit-message-format)

## PR Type

```
[ ] Bugfix
[ ] Enhancement
[X] Code style update (formatting, local variables)
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[X] Other... Please describe:
```

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,11 @@
+root = true
+
+[*]
+charset = utf-8
+indent_style = space
+indent_size = 2
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.md]
+trim_trailing_whitespace = false
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,5 @@ dist
 /.idea/
 .nvmrc
 tmp/
-.editorconfig
 bower-material/
 code.material.angularjs.org/


.editorconfig
```diff
@@ -0,0 +1,10 @@
+root = true
+
+[*]
+end_of_line = lf
```
Splaktar:
```suggestion
charset = utf-8
indent_style = space
indent_size = 2
insert_final_newline = true
trim_trailing_whitespace = true
```

.editorconfig
```diff
@@ -0,0 +1,10 @@
...
+
+[*.{html,js}]
+charset = utf-8
+indent_size = 2
+indent_style = space
+insert_final_newline = true
```
Splaktar:
```suggestion

[*.md]
trim_trailing_whitespace = false
```

.editorconfig
```diff
@@ -0,0 +1,10 @@
+root = true
+
+[*]
+end_of_line = lf
+
+[*.{html,js}]
```
Splaktar: I think that this particular subset of rules can be removed given the above rules for `[*]`.

.editorconfig
oliversalzburg: The diffs on GitHub were a bit confusing. I hope I got all of it right.

PR #11858, opened 2020-02-14
fix(compiler): $onDestroy hook not called

### LINKED ISSUE #11847: mdCompiler: $onDestroy lifecycle hook never called
## Bug

1. Use `$mdPanel` to instantiate a controller instance for a dialog.
1. The `$onInit` lifecycle hook is invoked by the compiler as expected.
1. Destroy the panel.

The `$onDestroy` lifecycle hook is invoked.

It is not.

These are standard AngularJS lifecycle hooks that should be expected to work on all controller instances. _Especially_ since `$onInit` is working as expected.

- AngularJS: 1.7
- AngularJS Material: 1.1.20
- OS: Windows
- Browsers: Chrome

### Related issues
- #8153
- #8845
- #10567

oliversalzburg:
## PR Checklist

## PR Type

```
[X] Bugfix
[X] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Even though the `$onInit` lifecycle hook is fully supported, the `$onDestroy` hook isn't. This is unexpected for developers.

## What is the new behavior?
This change introduces a listener for the `$destroy` event on the linked scope. See https://docs.angularjs.org/api/ng/type/$rootScope.Scope#event-$destroy

This implementation closely resembles the AngularJS internal implementation of the lifecycle hook invokation behavior: https://github.com/angular/angular.js/blob/2b28c540ad7ebf4a9c3a6f108a9cb5b673d3712d/src/ng/compile.js#L3298-L3302

Fixes #11847

```
[ ] Yes
[X] No
```

### DIFF
--- a/src/core/services/compiler/compiler.js
+++ b/src/core/services/compiler/compiler.js
@@ -436,6 +436,15 @@ function MdCompilerProvider($compileProvider) {
         // Create the specified controller instance.
         var ctrl = self._createController(options, injectLocals, locals);
 
+        // Registering extra $destroy listeners should be avoided.
+        // Only register the listener if the controller implements a $onDestroy hook.
+        if (angular.isFunction(ctrl.$onDestroy)) {
+          scope.$on('$destroy', function() {
+            // Call the $onDestroy hook if it's present on the controller.
+            angular.isFunction(ctrl.$onDestroy) && ctrl.$onDestroy();
+          });
+        }
+
         // Unique identifier for AngularJS Route ngView controllers.
         element.data('$ngControllerController', ctrl);
         element.children().data('$ngControllerController', ctrl);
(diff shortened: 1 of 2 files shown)


src/core/services/compiler/compiler.js
```diff
@@ -436,6 +436,15 @@ function MdCompilerProvider($compileProvider) {
         // Create the specified controller instance.
         var ctrl = self._createController(options, injectLocals, locals);
 
+        // Registering excisse $destroy listeners should be avoided.
```
Splaktar:
```suggestion
        // Registering extra $destroy listeners should be avoided.
```

src/core/services/compiler/compiler.spec.js
```diff
@@ -482,4 +482,61 @@ describe('$mdCompiler service', function() {
     });
   });
 
+  describe('AngularJS 1.6 lifecycle hooks', function() {
```
Splaktar, line 485:
```suggestion
  describe('AngularJS 1.6+ lifecycle hooks', function() {
```

PR #11891, opened 2020-04-22
fix(select): menuController not defined edge case in focusOptionNode

### LINKED ISSUE #11885: select: focusOptionNode throws exception when parent container is removed from DOM during panel close
## Bug

Work in progress: https://stackblitz.com/edit/angularjs-material-11885?file=home%2Fhome.html

1. The `md-select` needs to be in a section of the page that is hidden (or has its content hidden) by an `ng-if`.
1. Click on an element that causes the `ng-if` to show the `md-select`.
1. Click on the `md-select` to open the options panel.
1. Click on an element that causes both
  1. the `ng-if` to hide the `md-select`
  1. the `md-select` to close its options panel

The `md-select` is hidden and the options panel closes without any exceptions or issues.

`SelectProvider.focusOptionNode()` throws an exception when trying to call `SelectMenuController.setActiveDescendant()`
https://github.com/angular/material/blob/a8954dfaded6b5ae36aa35eaba3f33602f1e9da3/src/components/select/select.js#L1593
Because `menuController` is `undefined`. This is passed in via `SelectProvider.autoFocus()`:
https://github.com/angular/material/blob/a8954dfaded6b5ae36aa35eaba3f33602f1e9da3/src/components/select/select.js#L1616-L1617

In addition to throwing the exception, there are reports that the whole app may freeze.

Fix an edge case.

- AngularJS: 1.7.9
- AngularJS Material: 1.1.21-rc.0
- OS: macOS
- Browsers: Chrome 80
 
Related to PR https://github.com/angular/material/pull/11761.

Reported by @codymikol.

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

As reported in #11885, `SelectProvider.focusOptionNode()` can throw an exception when trying to call `SelectMenuController.setActiveDescendant()` in edge cases where the `SelectMenuController` is no longer defined or its lookup failed.

Issue Number: 
Fixes #11885

## What is the new behavior?
- Handle edge cases where the `selectMenuController` has been destroyed or is an `{}` rather than a Controller
- remove unused `selectController` argument to `focusOptionNode()`

```
[ ] Yes
[x] No
```
The API changes are for unreleased, private APIs only, which were introduced in `1.1.22-rc.0`.

## Other information
Not able to reproduce the issue (in or out of a test). This just adds some safety checks that should make the reported exception not throw.

### DIFF
--- a/src/components/select/demoBasicUsage/script.js
+++ b/src/components/select/demoBasicUsage/script.js
@@ -1,11 +1,12 @@
 (function () {
   'use strict';
   angular
-      .module('selectDemoBasic', ['ngMaterial'])
-      .controller('AppCtrl', function() {
-        this.userState = '';
-        this.states = ('AL AK AZ AR CA CO CT DE FL GA HI ID IL IN IA KS KY LA ME MD MA MI MN MS ' +
-            'MO MT NE NV NH NJ NM NY NC ND OH OK OR PA RI SC SD TN TX UT VT VA WA WV WI ' +
-            'WY').split(' ').map(function (state) { return { abbrev: state }; });
-      });
+  .module('selectDemoBasic', ['ngMaterial'])
+  .controller('AppCtrl', function() {
+    var ctrl = this;
+    ctrl.userState = '';
+    ctrl.states = ('AL AK AZ AR CA CO CT DE FL GA HI ID IL IN IA KS KY LA ME MD MA MI MN MS ' +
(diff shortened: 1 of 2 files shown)


Splaktar:
@codymikol please take a look (PTAL) and let me know if you think this will resolve the issue. If you can test out the fix locally, even better.

PR #11934, opened 2020-06-08
fix(menu-bar): close just nested menu when escape key is pressed

### LINKED ISSUE #11678: menu-bar: Focus is lost after pressing "esc" while being in nested menu
## Bug:
If you pressing "ecs" while being in a nested menu, focus is lost. This is bad for accessibility users, which rely on focus heavily to navigate. Instead focus should be returned on previous menu item or on the menu button. 

1. Go to [Menu Bar basic usage demo](https://material.angularjs.org/latest/demo/menuBar)
1. Using the keyboard, navigate to File > New >  Presentation
Screenshot from 2019-03-14 14-07-27
1. Press "ecs". 
1. Focus is lost.
[WAI-ARIA menu practices for accessibility](https://www.w3.org/TR/wai-aria-practices/#menu) states that on
Escape: Close the menu that contains focus and return focus to the element or context, e.g., menu button or parent menuitem, from which the menu was opened.

Focus is lost.

Accessibility users (screen reader users, keyboard users, switch users) use focus heavily to navigate. Then focus is suddenly lost it's a frustrating experience. 

- AngularJS Material: Latest Release 1.1.13
- Browsers: Chrome

marosoft:
focus stays with the parent menu when nested one is closed instead of closing all the menus

Fixes #11678

## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

At the moment when the user presses escape key while the nested menu is open, it causes all the menus to become closed and the focus is lost.

Issue Number: #11678

## What is the new behavior?
When the escape key is pressed, only the current nested menu is closed and the focus goes to the parent menu.

```
[ ] Yes
[x] No
```

### DIFF
--- a/src/components/menuBar/menu-bar.spec.js
+++ b/src/components/menuBar/menu-bar.spec.js
@@ -226,6 +226,39 @@ describe('material.components.menuBar', function() {
           menuBar.remove();
         });
 
+        it('closes only current sub-menu with escape key', inject(function($mdConstant) {
+          menuBar = setup();
+          ctrl = menuBar.controller('mdMenuBar');
+          menus = menuBar[0].querySelectorAll('md-menu md-menu');
+
+          angular.element(document.body).append(menuBar);
+
+          expect(getNumberOfOpenMenus()).toBe(0);
+
+          // Open the menu-bar menu
+          ctrl.focusMenu(1);
+          ctrl.openFocusedMenu();
+          waitForMenuOpen();
+
+          expect(getNumberOfOpenMenus()).toBe(1);
+
+          // Open the first nested menu
+          openSubMenu(0);
+          waitForMenuOpen();
+          expect(getOpenSubMenu().text().trim()).toBe('Sub 1 - Content');
+
+          expect(getNumberOfOpenMenus()).toBe(2);
+
+          // Close the first nested menu with escape key
+          pressKey(getOpenSubMenu(), $mdConstant.KEY_CODE.ESCAPE);
+          waitForMenuClose();
+
+          // Just main menu should be visible
+          expect(getNumberOfOpenMenus()).toBe(1);
+
+          menuBar.remove();
+        }));
+
         function openSubMenu(index) {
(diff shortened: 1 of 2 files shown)


src/components/menuBar/menu-bar.spec.js
```diff
@@ -279,6 +279,108 @@ describe('material.components.menuBar', function() {
           return el;
         }
       });
+
+      describe('nested menus - close menu with escape key', function() {
```
Splaktar, line 283: Why is this nested `describe` needed?

src/components/menuBar/menu-bar.spec.js
```diff
@@ -279,6 +279,108 @@ describe('material.components.menuBar', function() {
...
+          pressKey(getOpenSubMenu(), $mdConstant.KEY_CODE.ESCAPE);
+          waitForMenuClose();
+
+          // Just main menu should be visible
+          expect(getNumberOfOpenMenus()).toBe(1);
+
+          menuBar.remove();
+        }));
+
+        function openSubMenu(index) {
```
Splaktar, line 319: These seem to be duplicated from the parent `describe` block? Except for `getNumberOfOpenMenus ()`.

Can this all be done within that same `describe` for `nested menus`?

src/components/menuBar/menu-bar.spec.js
```diff
@@ -279,6 +279,108 @@ describe('material.components.menuBar', function() {
           return el;
         }
       });
+
+      describe('nested menus - close menu with escape key', function() {
+        var menuBar, menus, subMenuOpen, ctrl;
+
+        it('opens consecutive nested menus', inject(function($mdConstant) {
```
Splaktar, line 286: This name doesn't seem representative of what's being tested. It seems like the `describe` text is what should be here?

Also it should probably be `closes only the top level menu with the escape key`.

PR #11978, opened 2020-07-24
fix(interim-element): onComplete called before stack updated

justin-caldicott:
## PR Checklist

## PR Type
```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Calling `$mdDialog.hide() ` within the `onComplete` callback passed to `$mdDialog.show()` will not hide the dialog just shown.

In our case, we have a progress dialog which shows briefly, and sometimes the operation completes quickly and before the dialog has finished showing. We can't call hide() at that point (though that would be nice) and cancel the show, but we can track a "pendingHide" and close the dialog as soon as it has finished showing.

But the current behaviour makes this very hard, because the show hasn't fully completed - it hasn't made it into the `showingInterims` array / stack yet.

## What is the new behavior?

Calling `$mdDialog.hide() ` within the `onComplete` callback passed to `$mdDialog.show()` will hide the dialog just shown, because the show is fully completed and the dialog is tracked in the showingInterims array.

```
[ ] Yes
[x] No
```

## Other information

I'm assuming the existing tests will cover this change, though I didn't add a regression test. (It would be more tricky as I'm not familiar with the codebase)


justin-caldicott:
@googlebot I signed it!

Splaktar:
Can you please provide a CodePen or StackBlitz demo of the issue so that I can confirm that this PR fixes it? Without a test that reproduces the problem, we currently have no way to know if this code change resolves the issue.

src/core/services/interimElement/interimElement.js
```diff
@@ -696,8 +699,6 @@ function InterimElementProvider() {
         function showElement(element, options, controller) {
           // Trigger onShowing callback before the `show()` starts
           var notifyShowing = options.onShowing || angular.noop;
-          // Trigger onComplete callback when the `show()` finishes
-          var notifyComplete = options.onComplete || angular.noop;
```
Splaktar, line 700: This is here because it is meant to be used for animations. It is run as soon as the `$animate.enter` Promise resolves. This enables the ability to trigger an animation that comes right after the transitionIn animation is complete.

justin-caldicott:
Closing for now as I haven't the time to make the case for this properly. I'll use my own fork for now and reopen later time permitting.

PR #12015, opened 2020-09-21
fix(select): in popup, hover and focus background colors are the same

### LINKED ISSUE #9851: select: can have 2 options with a selected (grey) background
Hi,

With Angular Materials 1.1, it seems that both the md-option with the focus and the one with mouse-over have a grey background. You can view that on the demos page:
https://material.angularjs.org/latest/demo/select

So clicking on a select you can move one of the grey background elements with the keyboard arrows and the other by hovering the mouse, leading to 2 different elements having a grey background (both can differ from the currently selected option which has a blue text color).

I couldn't find it reported and it felt a bit confusing to me.
So I just wanted to confirm if it was intended behavior (trying to please both people navigating with the mouse and people navigating with the keyboard) or something you intended to modify in a future release.

Thanks.

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

- In the `md-select` popup, hover and focus background colors are the same. This can cause confusion when navigating with both keyboard and mouse on desktop.

Issue Number: 
Fixes #9851

## What is the new behavior?
- use a similar approach to the one used by MDC. I.e. hover is 8% less opaque than focus.
  - the opacity values are slightly different because
    they use `#000` and we use `rgba(0, 0, 0, 0.87)`

```
[ ] Yes
[x] No
```

## Other information
No tests as this is a CSS-only change.

### DIFF
--- a/src/components/select/select-theme.scss
+++ b/src/components/select/select-theme.scss
@@ -153,8 +153,10 @@ md-select-menu.md-THEME_NAME-theme {
       }
 
       &:not([disabled]) {
+        &:hover {
+          background-color: '{{background-500-0.10}}'
+        }
         &:focus,
-        &:hover,
         &.md-focused {
           background-color: '{{background-500-0.18}}'
         }


wagnermaciel (approved):
lgtm

PR #12030, opened 2020-10-11
fix(datepicker): time is modified before even selecting a date

### LINKED ISSUE #12026: datepicker: modifies time
## Bug

# Demo and steps to reproduce the issue
[This blank StackBlitz demo](https://stackblitz.com/edit/angularjs-material-blank?file=app%2Fapp.template.html) can be used to create a reproduction that demonstrates your issue.

**Demo URL (required)***:  
[Here](https://jsfiddle.net/frcoeb4n/) is a live demo.

#### Detailed Reproduction Steps
1. Notice the date strings have some time values
1. Click or focus the datepicker
1. Notice that the date strings have all of their time values set to midnight

### Explain the expected behavior

The `ng-model` value must to remain the predefined time as far as possible.

However, regarding the limits (`md-min-date` and `md-max-date`), they have to be treat as read-only data, not altered.

### Explain the current behavior

In the last updates, when date picker is interacted (clicked, calendar showed, focused...), time of scope values are changed. It happens both to `ng-model` and to limits (`md-min-date` and `md-max-date`). It is very traumatic when you are using these values to control datetime by another time input.

### List the affected versions of AngularJS, Material, OS, and browsers

- AngularJS: 1.7.2
- AngularJS Material: 1.2.1
- OS: macOS
- Browsers: Chrome 85
- Screen Readers: 
 
#### Stack Trace

N/A

### LINKED ISSUE #12028: datepicker: with UTC timezone set doesn't handle time offset +2:00
## Bug

# Demo and steps to reproduce the issue
1. [Demo page with user interaction](https://stackblitz.com/edit/angularjs-material-datepicker-bug-demo?file=app.js)
1. [Demo page during load](https://codepen.io/nicklastromb/pen/OJXLqeL?editable=true%3Dhttps%3A%2F%2Fmaterial.angularjs.org%2F)

#### Detailed Reproduction Steps (With user interaction)
1. Create a controller with a md-datepicker in the template
1. Set the `ng-model-options` to `{timezone:'UTC'}`
1. Load the view and open the datepicker. The date will be incorrectly displayed in the datepicker and the date will have an updated time offset.
1. The time for the date is now 00:00 (but with correct timezone) instead of the correct timezone adjustment

See demo url (1)

#### Detailed Reproduction Steps (Without interaction)

This step is a bit harder to reproduce as it seems to depend on what is loaded in the view. I've been able to reproduce this in the example for timezone settings at https://material.angularjs.org/latest/demo/datepicker, I've copied the demo to codepen where it seems to appear as long as the calendar is loaded together with the datepicker. I've also experienced this issue without any calendar being loaded in the view.

1. Create a view with a calendar and a datepicker component
1. Set the `ng-model-options` to `{timezone:'UTC'}`
1. Load the view, the datepicker timestamp should now be "00:00" instead of the correct offset

Any selection of a date will update the time offset properly.
Removing the calendar will prevent the problem from occurring. 

See demo url (2)

### Explain the expected behavior
Time for the date should not be affected when opening or loading the datepicker.

### Explain the current behavior
When setting an initial date the time part of the date will be affected when either opening up the datepicker and closing it or during the initial load of the datepicker in some scenarios. This is not expected bahavior.

### Discuss the use-case or motivation for changing the existing behavior

The current behavior is unexpected and may lead to an invalid date being selected when using the datepicker component.

### List the affected versions of AngularJS, Material, OS, and browsers

- AngularJS: 1.8.1, 1.8.0
- AngularJS Material: 1.2.1, 1.1.26
- OS: Windows 10
- Browsers: Chrome 85.0.4183.12
 
#### Screenshots

**When opening the datepicker**
2020-10-06-184523-RiI1B7KOub

**During first load**

2020-10-06-182557-chrome_jFJj1KV2jd

After a date is selected after first load the time offset is updated correctly

image

Splaktar:
## PR Checklist

## PR Type

```
[x] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

- time is modified before even selecting a date

Issue Number: 
Fixes #12028. Closes #12026.

## What is the new behavior?
- time is not modified before selecting a date
- `updateErrorState()` was mutating the model incorrectly and clearing the
  time to midnight when the datepicker is first interacted with, without
  selecting a date

```
[ ] Yes
[x] No
```


wagnermaciel (approved):
lgtm

PR #12064, opened 2020-12-16
fix(autocomplete): prevent flashing of invalid state

### LINKED ISSUE #10975: autocomplete: md-require-match flashes ng-message error on selection
**Actual Behavior**: md-require-match briefly shows an error message on mouse down event.

 -  `What is the issue? * I am using md-require-match inside an md-autocomplete and, once autocomplete's item are shown,  I select one of them with my mouse and, for a millisecond, error message for md-require-match is showed. 
Please note that this message is **not** showed if I hit an item with enter key.

 -  `What is the expected behavior?` No error message should be displayed when an item is selected from the list.

 -  `CodePen Demo which shows your issue:` https://codepen.io/anon/pen/KyNZxB
 -  `Details:` The above code is simply the demo on angular material web site. 
Steps to reproduce: 

- click "_Favourite States_"
- search for '_a_' to get list of available states .
- left click with your mouse ( don't leave the mouse button for clarity)
- error message 'Please select an existing state' is showed

**AngularJS Versions**: *
 -  `AngularJS Version:`1.5.5
 -  `AngularJS Material Version:` v1.1.5

**Additional Information**:
 -  `Browser Type: *` Chrome
 -  `Browser Version: *`61.0.3163.100
 -  `OS: *` Windows 10
 -  `Stack Traces:`

----

  Add more lines if needed.

shishkinilya:
## PR Checklist
Please check your PR fulfills the following requirements:

## PR Type

```
[X] Bugfix
[ ] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

When user click on suggestion item, input element lost focus, and then programmatically get focused again, in between these moments invalid state flashes

Also described in https://github.com/angular/material/issues/10975

Fixes #10975

## What is the new behavior?
Invalid state doesn't flash

```
[ ] Yes
[X] No
```

## Other information
Perhaps you can help me, my branch fails the test in autocomplete.spec.js "shows the md-not-found template even if we have lost focus", and to be honest, I don't understand the meaning of this test. Why should the md-not-found element must be visible in case of focus loss? Plus, this does not match the actual behavior of the component, if the input element loses focus, the md-not-found element is also hidden.

### DIFF
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -601,6 +601,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     if (!noBlur) {
       ctrl.hidden = shouldHide();
       evalAttr('ngBlur', { $event: $event });
+    } else {
+      $event.stopImmediatePropagation();
     }
   }
 
--- a/src/components/autocomplete/js/autocompleteController.js
+++ b/src/components/autocomplete/js/autocompleteController.js
@@ -594,6 +594,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
 
   /**
    * Handles input blur event, determines if the dropdown should hide.
+   * @param {Event=} $event
    */
   function blur($event) {
     hasFocus = false;
@@ -601,7 +602,7 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     if (!noBlur) {
       ctrl.hidden = shouldHide();
       evalAttr('ngBlur', { $event: $event });
-    } else {
+    } else if (angular.isObject($event)) {
       $event.stopImmediatePropagation();
     }
   }


src/components/autocomplete/js/autocompleteController.js
```diff
@@ -601,6 +601,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
     if (!noBlur) {
       ctrl.hidden = shouldHide();
       evalAttr('ngBlur', { $event: $event });
+    } else {
```
Splaktar, line 604: We should add
```
@param {Event=} $event
```
to the JSDoc to mark it as optional since this function is most often called without an argument.

This is also why one of the tests is failing with
```js
TypeError: Cannot read property 'stopImmediatePropagation' of undefined
    at MdAutocompleteCtrl.blur (src/components/autocomplete/js/autocompleteController.js:605:14)
```
and
```js
TypeError: $event is undefined
blur@src/components/autocomplete/js/autocompleteController.js:605:7
```

src/components/autocomplete/js/autocompleteController.js
```diff
@@ -601,6 +601,8 @@ function MdAutocompleteCtrl ($scope, $element, $mdUtil, $mdConstant, $mdTheming,
...
+      $event.stopImmediatePropagation();
```
Splaktar, line 605:
```suggestion
    } else if (angular.isObject($event)) {
      $event.stopImmediatePropagation();
    }
```
```

**Review by Splaktar — CHANGES_REQUESTED** [review]
Thank you for investigating this issue!

File: src/components/autocomplete/js/autocompleteController.js (hunk 2 above)
Comment (shishkinilya) [reply src/components/autocomplete/js/autocompleteController.js]: Thank you! I made a commit with your suggestion, but I don't quite understand this test `shows the md-not-found template even if we have lost focus` in autocomplete.spec.js, why after `controller.blur();` we expecting that notFoundTemplate should be visible?
```
controller.blur();
expect(controller.notFoundVisible()).toBe(true);
```

shishkinilya: And one more question, should I add some tests for my changes?

Splaktar: You can see the reasoning behind the test using Git Blame which points to this commit:
https://github.com/angular/material/commit/52a519ead04f2faeba227fbc51a7214253cbf2ba

It explains that the main use case here is actually clicking on the not found item in the dropdown. It shouldn't dismiss the panel. The test could certainly have a bit better description.

Splaktar: I don't think that we need a test for this. I was just able to verify this manually in the demo on the docs site.

Splaktar (approved):
LGTM. I verified this manually using the autocomplete floating label demo.

src/components/autocomplete/js/autocompleteController.js
shishkinilya: Ok, thanks for the explanation

mmalerba (approved):
LGTM

PR #12080, opened 2021-02-24
feat(menu): render backdrop to parent

atfzls:
- attach the backdrop in menu to `options.backdropParent`

backdrop in menu now attaches to `options.backdropParent` if passed, previously it was always attached to `document.body`

## PR Checklist
Please check your PR fulfills the following requirements:

## PR Type
```
[ ] Bugfix
[x] Enhancement
[ ] Build related changes
[ ] CI related changes
[ ] Infrastructure changes
[ ] Other... Please describe:
```

Backdrop is always rendered to `document.body`.

## What is the new behavior?
Backdrop now renders to `options.backdropParent` if provided.

```
[] Yes
[x] No
```

Menu backdrop was always rendered to `document.body` even if we pass `options.parent`(which is used to decide where to attach the menu not backdrop).
This PR adds a new options `options.backdropParent` which defines an element to attach the backdrop to.

## Other information
We needed this option in our app so that backdrop is rendered to a specific element, not the body.
With this change, we'll be using `parent` property to render the menu to a specific element and `backdropParent` property to render backdrop to a specific element.

Appending the backdrop or menu directly to body resulted in forced reflows. So, this options helps in improving the performance of our web app.

### DIFF
--- a/src/components/menu/js/menuServiceProvider.js
+++ b/src/components/menu/js/menuServiceProvider.js
@@ -59,7 +59,7 @@ function MenuProvider($$interimElementProvider) {
       if (options.hasBackdrop) {
         options.backdrop = $mdUtil.createBackdrop(scope, "md-menu-backdrop md-click-catcher");
 
-        $animate.enter(options.backdrop, $document[0].body);
+        $animate.enter(options.backdrop, options.backdropParent || $document[0].body);
       }
 
       /**
--- a/src/components/menu/menu.spec.js
+++ b/src/components/menu/menu.spec.js
@@ -397,6 +397,34 @@ describe('material.components.menu', function() {
       $log = $injector.get('$log');
     }));
 
+    it('backdrop should attach to backdropParent', function() {
+      var parent = angular.element('<div>');
+      var menuEl = angular.element(
+        '<md-menu>' +
+        '  <button ng-click="null">Trigger</button>' +
+        '</md-menu>'
+      );
+      var backdropParent = angular.element('<div>');
+
+      $mdMenu.show({
+        scope: $rootScope,
+        element: menuEl,
+        target: document.body,
+        preserveElement: true,
+        parent: parent,
+        backdropParent: backdropParent,
+      });
+
+      $timeout.flush();
+
+      expect(backdropParent.find('md-backdrop').length).toBe(1);
+
+      // Close the menu and remove the containers
+      $mdMenu.hide();
+      parent.remove();
+      backdropParent.remove();
+    });
+
     it('should warn when the md-menu-content element is missing', function() {
       spyOn($log, 'warn');
 


src/components/menu/js/menuServiceProvider.js
```diff
@@ -59,7 +59,7 @@ function MenuProvider($$interimElementProvider) {
       if (options.hasBackdrop) {
         options.backdrop = $mdUtil.createBackdrop(scope, "md-menu-backdrop md-click-catcher");
 
-        $animate.enter(options.backdrop, $document[0].body);
+        $animate.enter(options.backdrop, options.parent || $document[0].body);
```
mgol, line 62: Backdrop is supposed to catch all clicks and the parent is usually smaller than the whole viewport; I suspect issues if `overflow: hidden` is used, etc. For cases where this is needed, how about a new option `backdropParent`? Then we wouldn't need a breaking change either.

Splaktar:
Since this isn't associated with an issue, can you please provide a StackBlitz or CodePen demo and reproduction steps? This is especially important since it's unlikely that we can effectively unit test this.

Additionally, we're currently in LTS, so we wouldn't accept a breaking change unless it was to fix a security issue, breakage in a major browser, or to maintain compatibility with the latest jQuery.

atfzl:
@Splaktar @mgol I've updated the code to be backward compatible.
If we don't pass `backdropParent`, the old behavior of backdrop rendering in `body` works.

I have also added a UT.

atfzl:
Also, seems like there is some issue with CircleCI https://app.circleci.com/pipelines/github/angular/material/1016/workflows/61f18a04-bd10-480d-8895-63fd90c2bfc3/jobs/6063

```
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.

*** Please tell me who you are.

Run

  git config --global user.email "you@example.com"
  git config --global user.name "Your Name"

to set your account's default identity.
Omit --global to set the identity only in this repository.

fatal: empty ident name (for <circleci@04f0d9a1837b>) not allowed

Exited with code exit status 128
CircleCI received exit code 128
```

Splaktar:
I'm working on fixing the CI issues in https://github.com/angular/material/pull/12093. Thank you for the heads up.

mgol (approved):
Looks good to me. @Splaktar I assume you'd like to fix CI and have this PR pas before approving & merging this?

Splaktar:
@mgol correct. Hopefully I can get that CI fix in today or tomorrow.

Splaktar:
OK, https://github.com/angular/material/pull/12093 landed. Can you please rebase this PR? That should get the CI working properly.

atfzls:
Rebased, tests are passing.

Splaktar (approved):
LGTM. However, when I try to "Approve and run" the workflow for the CI tests, GitHub is giving me a HTTP 500 server error. I'll try to look into this.

Splaktar:
OK, that's just for CodeQL scanning, everything else passed. So I think that this is good to go.

wagnermaciel (approved):
lgtm

PR #12117, opened 2021-07-23
SECU-1152

NirmalaY12:
Integrating CodeQL Static code analysis tool with the github actions to check for the security vulnerabilities in the codebases

### DIFF
--- a/.github/workflows/codeql.yml
+++ b/.github/workflows/codeql.yml
@@ -1,0 +1,46 @@
+name: "CodeQL"
+
+on:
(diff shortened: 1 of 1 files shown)


Splaktar (changes requested):
We already have 
https://github.com/angular/material/blob/fc4e51bc2d9871a5656274aaaf44e3ac6c7738c4/.github/workflows/codeql-analysis.yml

Was this PR targeted at the wrong repository or is this trying to change our config in some way?

